PR #34, opened 2019-02-16
[ashikns] Enable .exe Launching through FullTrustProcess

duke7553:
I tweaked ashikns work to include the click action. I'll leave this open for discussion and review.


ashikns:
Looks good to me 👍 You might want to consider moving the uwp project to it's own folder for organizational purpose since there are 3 projects now.

duke7553:
Thank you!

ashikns:
Tag any issues you need help with in future :)

duke7553:
@ashikns Another question. Since this executable likely cannot be compiled for ARM, in what way should I customize the ARM build to not include or attempt to run this executable?

ashikns:
I have no experience with ARM actually. Never had to deal with it in my career. Does ARM not support executables? What is the equivalent on it?

duke7553:
Windows 10 can only build most executables for x86/x64. They partially solved this problem by introducing UWP which has full ARM support. Since we're bundling your Launcher project inside of this UWP app, I used the Configuration Manager to disable attempting to build it on ARM. 
1e52edc8-3781-410e-8eec-de89ca4d9214 png

My question is: How can we specify a way to only include the Launcher project in the package for x86/x64, and handle its lack of presence in ARM builds?
d783cf42-3a84-4633-81f5-ca539f7fbd54 png

ashikns:
From what I've heard ARM supports .Net core console apps. So all you'll have to do is replace the launcher project (which is based on desktop .Net atm) with .net core console project.

PR #41, opened 2019-02-17
Reorganize Project Into Folders

duke7553:
This will reorganize the project into three main folders: one for the package, exe launcher project, and the regular project.

The way it is now might cause deployment errors, so I will merge only after testing this branch.


duke7553:
Just re-cloned this branch and it deployed successfully.

PR #44, opened 2019-02-17
Fixed visiblity issue for close, minimize, and maximize in light mode.

ATecSolution:
Fixed visiblity issue for close, minimize, and maximize in light mode.


duke7553:
Wow, nice work. Has this been tested to fix the problem when the user changes the theme to differ from the System theme? If so, LGTM

duke7553:
I noticed the same thing happens with the Dark theme, when system is set to Light mode. Would this PR fix that too?
40e8a604-618c-4b90-958e-749c25f05660 png

ATecSolution:
In my testing it did.

duke7553:
Well, in that case...

PR #59, opened 2019-02-20
Greatly Improve Code Architecture

duke7553:
This builds off of @ashikns suggestion for better architecture by removing unnecessary statics. Furthermore, I eliminated the creation of ItemViewModel every single time, in favor of one instance that is referenced over and over.

### DIFF
--- a/Files UWP/AddItem.xaml.cs
+++ b/Files UWP/AddItem.xaml.cs
@@ -34,7 +34,7 @@ namespace Files
         {
 
             GenericFileBrowser.AddItemBox.Hide();
-            var currentPath = ItemViewModel.PUIP.Path;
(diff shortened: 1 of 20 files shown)


ashikns:
Looks much better to me. A couple more good practices: You made the collections non static, but they are still public. Which means a class outside ItemViewModel can modify your observable collections. This is not inherently bad, but usually what you want is only the ItemViewModel should be able to modify the collection and other classes should be able to view the collection only. Try to expose them as readable collections where possible. This has a couple benefits: 1) You can be sure there won't be any accidental modifications to your collection. 2) It reduces the public interface of ItemViewModel so a person who is trying to understand the class has less stuff to worry about.

PR #75, opened 2019-03-03
2/2 abilty to see properties

imthewiseguy:
*see previous*


duke7553:
Hello, thanks for the PR. I would like to see these merged into one.

PR #79, opened 2019-03-06
Auto refresh files and folders

### LINKED ISSUE #6: Auto-Refresh Directories
Currently, making a change Windows Explorer will not reflect in Files until you manually refresh the directory.

### LINKED ISSUE #55: Sometimes, clicking "back" or "forward" during the load of a large directory will include an item from the previous folder in the new one
Large directories are notorious in this. I currently have a cancellation token that triggers when navigation commands are invoked, but it still includes one random item from the previous directory in the directory being navigated to.

ashikns:
This PR aims to fix #6, fix #55.Filesystem changes will be automatically reflected. There are also some code refactors, some were necessary others are just good practices like reducing publicly accessible things.

Performance is good enough for a moderately sized folder. However it's not optimal. There are two broad changes that can be made to improve performance (not just for auto refresh but in general):
1) Expose files and folders as separate lists. UWP treats file and folder as very different types and provides separate api to work with them so it makes sense to split them in the code as well.
2) Use a different observable collection like dictionary or hashset. Of course these don't exist by default so we'll have to write one ourselves.

### DIFF
--- a/Files UWP/AddItem.xaml.cs
+++ b/Files UWP/AddItem.xaml.cs
@@ -42,8 +42,8 @@ namespace Files
                 var userInput = GenericFileBrowser.inputForRename;
                 if (userInput != null)
                 {
-                    await folderToCreateItem.CreateFolderAsync(userInput, CreationCollisionOption.FailIfExists);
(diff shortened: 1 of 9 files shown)


ATecSolution:
Great job!

ashikns:
@duke7553 while testing please keep an eye on the ui to see if any bindings are broken. I don't get along well with xaml bindings -_-

duke7553:
LGTM; Very good work, by the way

PR #129, opened 2019-06-04
Can now reorder tabs.

yair100:


lukeblevins (changes requested):
The reordering would probably work, but to my knowledge it requires extra logic to handle dropping in a tab. Have you tested for this?

yaichenbaum:
Dropping in a tab from another windows does not work, like you said it requires extra logic but reordering tab in the same window works fine with the pr.

duke7553:
It is good practice to disable anything that requires extra work until it’s implemented. Thanks for the PR btw.

lukeblevins (commented):
Doesn’t reordering work with drag and drop disabled?

yaichenbaum:
No.

lukeblevins:
@yair100 Thanks for your work. After testing this PR, I notice that the TabView doesn't yet support proper reordering. (Regardless of the CanReorderItems property) I'll submit a request this property to the appropriate repo.

yaichenbaum:
Great, looking forward, by the way whats the status with the store app?

PR #154, opened 2019-06-30
Added data population in the Properties Window UI

VoidExploiter:
Added data population for properties window. Working on changing file name from the properties window. It will work for properties button click from the Menubar

image


yaichenbaum:
Before this is merged can the text blocks be aligned with each other? 
image

VoidExploiter:
Yeah it can be. i will add that in next PR or i will change that and give you a new PR

PR #161, opened 2019-07-07
Fixed app not detecting the system theme the first time launched

yaichenbaum:
https://github.com/duke7553/files-uwp/projects/1#card-23664632


yaichenbaum:
Else did fire, the problem is it is only saving the setting for the theme then, it is not changing the actual theme, by putting it first if it detects there is no value for the theme then it sets it to default, then once there is a setting the app can actually load it.

duke7553:
Good work. Didn't think of that.

PR #187, opened 2019-07-26
Remove Single Item capability added

RobbyRobbyRobby:
Not sure if it's in line with your UI, but you can easily change the icon from here.


duke7553:
@RobbyRobbyRobby Have you tested that this permanently removes the items? I'll test this change myself momentarily.

RobbyRobbyRobby:
You're right; I missed that.

yaichenbaum:
I believe that all your change does is removes it temporally, to fully remove it, it needs to be removed from the MRU list.

PR #190, opened 2019-07-27
Added remove recent item option from flyout menu

RobbyRobbyRobby:
Same effect as before, but with a confirmation dialog and also using the app's ui theme.


yaichenbaum:
@duke7553 I like the use of the confirmation dialog, can we use that for clearing all items from the list too?

Issue #186, opened 2019-07-25
Show Date/Time based on format Set in system settings.


avknaidu:
The date format right now shows like below.

![image](https://user-images.githubusercontent.com/6171200/61902952-5a7cf200-aef1-11e9-9f50-d6747579e7f6.png)

Instead, it should be in system settings set format. Like below.

![image](https://user-images.githubusercontent.com/6171200/61903005-78e2ed80-aef1-11e9-8111-92e7a9eea3f0.png)

PR #197, opened 2019-07-29
Implement Date Format on File Lists as settings

avknaidu:
This also improves settings page to use Enums instead of string values for comparision.

image

Linked Issue #186


duke7553:
I'll just merge this one now, since I think there is a good consensus around this PR. We can always revert it if something breaks.

Issue #173, opened 2019-07-17
Auto-detect user folders paths


KaKi87:
User folders paths are stored in Windows Registry.

You can find them there : `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders`

| Folder | Windows Registry Key |
| --- | --- |
| Desktop | `Desktop`
| Downloads | `{374DE290-123F-4565-9164-39C4925E467B}`
| Documents | `Personal`
| Pictures | `My Pictures`
| Music | `My Music`
| Videos | `My Video`

*([Source](https://www.winhelponline.com/blog/windows-10-shell-folders-paths-defaults-restore/))*

You should be able to get these by using [Microsoft.Win32.Registry.GetValue](https://docs.microsoft.com/fr-fr/dotnet/api/microsoft.win32.registry.getvalue).

yaichenbaum:
I believe we are already detecting user paths, the only one that is not using that method at moment is the downloads folder, thank you for the registry key.

yaichenbaum:
@KaKi87 would this work?         `public static string DownloadsPath = string)Registry.GetValue(@"HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders", "{374DE290-123F-4565-9164-39C4925E467B}", null);`

yaichenbaum:
@KaKi87 Can you verify #209 helps, if not I will need to know what changes to make, thanks.,

KaKi87:
Send me an APPX and I'll tell you

yaichenbaum:
@KaKi87 Are you still on discord?

KaKi87:
I leaved UWP Community.

yaichenbaum:
Ok, try getting it from here https://1drv.ms/u/s!AsMjD6AbL8azjrsBdZBSzqXv7SdY_w?e=c561Be

KaKi87:
I'm not ARM but x64 and the certificate seems not to work. Error code `0x800B0109`

yaichenbaum:
@KaKi87 I will need to create a new package for you, I will try to do it later today.

duke7553:
@KaKi87 This is fixed for the v0.7 release. Thank you for your patience.

PR #209, opened 2019-08-04
Read registry value for downloads folder

yair100:
Read the registry value for the downloads folder to address #173 if this does not address it please inform me so I can make the necessary adjustments.


avknaidu:
hey @yair100 I see you are using if-else here. I would suggest change that to switch statement because that will be quick and efficient. if-else will hit every if statement to see if the condition is met, switch directly goes to the identifier based on value passed in switch.

yaichenbaum:
@avknaidu I have never done it that way before but would be happy to learn from this, @duke7553 will be finishing the work on this PR when he gets some time, so this might be something for him to do once he is in those files anyways.

PR #241, opened 2019-09-26
Several Tweaks

duke7553:
This PR adds the following:
- Improvements to theme colors and materials, including the removal of Acrylic from the tab strip
- A fix to limit the opening of Settings from ribbon to one instance per app Window
- Changes to Removable Drive recognition

The following is still unimplemented:
- Dynamic updates of detected drive list


duke7553:
After some internal discussion, we're going to put this specific concept on the back burner. Reverted it, and will merge this PR because it contains some other needed changes.

Issue #20, opened 2019-02-12
No way to sort from DataGrid


ZekromInfinity:
(Bugs)
- Feedback doesn't work so here I am.

- Music starts with the 1st name in alphabetical order no matter which I click or what order the folder is sorted on, after the first time, it will work normally. Also, it will always highlight the name or whatever the data is in a white box every time I click something.

- Opening downloads crashed the app.

(A question)
- Is it a common problem with UWP apps to crash or is it just me? I am asking because I have used windows phone and windows store apps on pc and they seem to crash, almost all of them. Please give me some insight into it.

Thank You

ATecSolution:
About your question I find that uwp apps crash more then win 32 ones, more often in larger apps. But I do find that every Windows update improves the quality and decreases the amount of crashes.

duke7553:
Sorting is unimplemented at the moment. This will be added in future release.

carlosrfernandez:
There seems to be an unhandled exception when opening the Downloads folder. I'm also getting this.

duke7553:
Quick question. Have you guys granted the app permission to access your filesystem? In my testing, it displays a prompt upon first access that transfers the user to a Settings page for this permission. Also, what windows version?

duke7553:
(Version number of Windows 10) Example: 1803, 1809, or 1903

carlosrfernandez:
In my case, I'm on 1809, I am compiling from source, I also targeted the SDK that comes with 1809 so it might be related to that. 

To answer your question: yes, I gave the app permission to access the filesystem.

If I have time I'll try to understand what is causing it and provide a fix.

duke7553:
Thanks for the information. I can't seem to replicate this issue, but the app should load every non-pictures directory the same. Does it do this with other directories? Is it a certain type of file? 
Maybe try creating a folder in Windows Explorer and move each file into there until it loads properly in Files. That way, we can narrow things down.

carlosrfernandez:
It happens in other directories as well, so far the exception is being thrown on PNG files, but I can't tell if it's ALL png files, or just some.

Another thing, after opening one of the problematic files' properties using Windows Explorer, it stopped failing, my guess is flag is set at OS level or something. Very weird.

duke7553:
That is weird, but it should be handled now to prevent a crash.

duke7553:
I'll change the title of this now to reflect recent changes.

ZekromInfinity:
> 
> 
[...]

PR #247, opened 2019-10-24
Add sorting function to files list

### LINKED ISSUE #178: Sorting
I would like to be able to sort files by creation/edit date and file type.

Thanks

jeffsieu:
**Files and folders can now be sorted** by clicking on
the column headers or by using the right-click context
menu. The sorting order mimics Windows Explorer,
by grouping the folders and files separately first, then
applying the relevant sort on the specified column,
then finally sorting by name.

Also:
- Improves right-click context menu action grouping
- Removes `RowIndex` from `ListedItem.cs` as it is not a reliable way to access items (indices change) after sorting is implemented
- Refactors code to work around the absence of `RowIndex`
- Other minor code refactoring

## Screenshots
image
image
image

## Related Issues

Closes #178 

Possibly related: #20

### DIFF
--- a/Files UWP/Filesystem/ItemViewModel.cs
+++ b/Files UWP/Filesystem/ItemViewModel.cs
@@ -250,6 +250,58 @@ namespace Files.Filesystem
 
         }
 
+        public void OrderFiles(string orderBy, bool ascending)
+        {
+            Func<ListedItem, object> orderFunc;
+            object orderByNameFunc(ListedItem item) => item.FileName;
+            switch (orderBy)
+            {
+                case "Name":
+                    orderFunc = orderByNameFunc;
+                    break;
+                case "Date":
+                    orderFunc = item => item.FileDateReal;
+                    break;
+                case "Type":
+                    orderFunc = item => item.FileType;
+                    break;
+                case "Size":
+                    orderFunc = item => item.FileSizeBytes;
+                    break;
+                default:
+                    return;
+            }
+
+            // In ascending order, show folders first, then files.
+            // So, we use != "Folder" to make the value for "Folder" = 0, and for the rest, 1.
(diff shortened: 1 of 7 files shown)


yaichenbaum:
Do these changes apply to the photo view as well?

jeffsieu:
No, only on the default list view.

yaichenbaum:
Ok, I tested this and it works as expected, were you planning on doing it for photo view as well?
Side note, in the future we plan on merging the views and changing it to have different layout options so work like this wont need to be done more then once.

jeffsieu:
How does it work on photo view? Is it just via the right click menu, since there are no column headers? I haven't thought about doing it for the photo view as I don't usually use the photo view in Windows Explorer haha. But I guess I could get around to doing it some time soon.

duke7553:
@jeffsieu Great work!

duke7553:
My only concern would be if this breaks any functions that perform an action on a specific DataGrid index. This will require a bit of testing, still great to have regardless.

lukeblevins:
@yair100 @jeffsieu Looks great. After testing, I found several issues which we will need to address before I merge this branch to master:
1.) DataGrid doesn't remember sort options after navigation, but still displays the indicator on the wrong column
2.) Sorting by "Date modified" still needs work to be more accurate
3.) PhotoAlbum needs sort as well.
4.) Add Sort dropdown to ribbon in Layout tab

jeffsieu:
@duke7553 I've added sort by to PhotoAlbum and fixed 1). All that's left is making the sort dropdown in the ribbon.

As to 3), which part of sorting by "Date modified" is inaccurate?

yaichenbaum:
@duke7553 Can you explain what you meant by "Date modified" being inaccurate?

PR #249, opened 2019-10-25
Allow inline, direct renaming in file list

### LINKED ISSUE #189: Improve Item Collision Handling
When renaming files or folders that know how to differentiate between uppercase and lowercase 

1. Go to 'any folder '
2. Click on 'rename, either file or folder'
3. type a lowercase name and then rename but uppercase.
4. The application does not detect the difference between upper and lower case letters.

Be able to rename files or folders with the same name, but have the application detect the difference between upper case letters.

Files-UWP-rename

 - OS Version: [Windows 10 1903 18362.239]
 - App version: [v0.6.2]

jeffsieu:
# Changes
- Renaming dialog replaced with inline renaming in files list and photo album, which is the behaviour on Windows Explorer (also more user-friendly in my opinion)
- File extensions can now be changed when renaming files
- F2 to rename shortcut unchanged

## Screenshot
image

## Known Issues
- ~Editing only works when name cell is focused~
- Attempting to rename files and folders to the same name but in different case will result in unexpected behaviours
  - Since Windows' file path format is case-insensitive, changing the case of a file/folder does not trigger an automatic update in the UI. ~As such, the relevant cache of the directory will have to be updated programmatically to reflect this change~ Seems like this is a bug in the Windows Storage API; files renamed to the same name are not updated (even reopening the app doesn't update renamed files). **Until this issue with the Storage API has been resolved, such same-name-different-case renaming will be disallowed.**
- ~`RenameDialog` still used in `PhotoAlbum`~

# Related Issues
#189


yaichenbaum:
Amazing work, there are no words to describe how helpful all your contributions are! Does this include pressing F2 to rename?

jeffsieu:
Thanks!

Yes, it retains the F2 keybindings. Pressing F2/Rename (whether from the ribbon or from right clicking) triggers `DataGrid.beginEdit()` which puts the name cell into edit mode.

However it doesn't work when selecting the row by clicking on other cells of that row (Date modified, Size, etc) because the DataGrid tries to edit the current cell that is chosen (which will be ReadOnly). A possible workaround may have to be subclassing DataGrid.

jeffsieu:
Fixed the UI problem; now renaming in GenericFileBrowser can be done regardless of how the selected row is brought into focus.

I'll get to doing this for PhotoAlbum

Now, the remaining problem is with renaming files to the same name, but of different case. The Windows Storage API does not reflect the change in file name when the current directory is updated. Even though the change in folder contents is detected, the `StorageFile` has the same `FolderRelativeId` and retains its original name in `StorageFile.DisplayName`. Navigating to a different folder then returning to the original folder causes the renamed file to appear as the original file, even though the case is actually changed.

Steps to reproduce:
1. Create file a.txt
2. Rename it to A.txt
3. Navigate to parent folder
4. Navigate back to original folder
5. File name shows as a.txt (looking at this file with Windows Explorer shows it as A.txt)

Possible solutions:
- Renaming the file to another name, then renaming it back.

yaichenbaum:
Is this a limitation, or is it possible to create a work around?

duke7553:
@jeffsieu That sounds like a limitation of the Windows Storage API

yaichenbaum:
@duke7553 make sure to bring this up on your call next month...

jeffsieu:
@yair100 @duke7553 A possible workaround until the API is updated would be to rename it to a different name, then rename it back. Like a.txt -> b.txt -> A.txt. I've tried doing this and it works (albeit a bit hacky)

yaichenbaum:
@jeffsieu I am all for temporary hacks while we wait for a update so long as it is clearly commented so other developers can work off the changes and understand why its done that way.

jeffsieu:
Just added support for PhotoGrid as well.

With regards to the hack, somehow it doesn't seem to work... I'll have to figure out something else.

jeffsieu:
Maybe the best option for now is to not allow such renaming.

yaichenbaum:
@jeffsieu is this ready for review?

jeffsieu:
@yair100 I'll need to change a few lines to prevent the "same name but different case" renaming, then the PR should be ready.

lukeblevins (approved):
I resolved the merge conflicts with the previous commits. This looks really good. I plan to add a check for the file extension, so a name without a file extension won't overwrite the "Full Name" if file extensions are disabled. Also, I'll ensure rename operations are finalized when row selection changes on the DataGrid. For now, we can merge this. Great work!

PR #267, opened 2019-11-02
Fix double-highlight on PhotoAlbum selection change

jeffsieu:
Currently, changing the selection in PhotoAlbum using arrow keys causes a double-highlight to appear on the newly selected item.

This PR fixes this.

| Before | After |
| ---| --|
| image | image |
| image | image |

### DIFF
--- a/Files UWP/PhotoAlbum.xaml
+++ b/Files UWP/PhotoAlbum.xaml
@@ -523,6 +523,7 @@
                             </MenuFlyout>
                         </Setter.Value>
                     </Setter>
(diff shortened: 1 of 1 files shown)


yaichenbaum:
This may be intentional, let me check with someone who will be able to confirm the expected behavior for using the arrow keys.

yaichenbaum:
Looking at this page [here](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.control.focusstate) it looks like we should keep this focus rectangle to align with the rest of Windows.

jeffsieu:
Hmm, ok let's go with that then

PR #271, opened 2019-11-03
Started work on open file location in recents view

yaichenbaum:
image

I got it set up to open the file location for folders, however I need help getting it to work for files.
In addition we need to handle if the item location was deleted and instead of trying to go to the non existent location and crashing it should tell the user the location no longer exists and ask if they want to remove the item from the list.
We also need to select the right location on the sidebar, for example if the location is in the `c` drive then that drive should be selected in the sidebar.

### DIFF
--- a/Files UWP/YourHome.xaml
+++ b/Files UWP/YourHome.xaml
@@ -606,6 +606,8 @@
                                             <MenuFlyout MenuFlyoutPresenterStyle="{StaticResource MenuFlyoutFluentThemeResources}">
                                                 <MenuFlyoutItem Click="mfi_RemoveOneItem_Click" Icon="Remove" Text="Remove this item" x:Name="mfi_RemoveOneItem"/>
                                                 <MenuFlyoutItem Click="MenuFlyoutItem_Click" Icon="Delete" Text="Clear all items"/>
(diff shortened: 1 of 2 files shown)


yaichenbaum:
It doesn't make sense to have this option for folders since users can just open the folder themselves, how we should move forward is to disable or hide the option for folders and just have it for files.

jeffsieu:
I think it'd make sense to have it for files only.

yaichenbaum:
@jeffsieu That is what I was thinking, I was having trouble getting it working for files because we need to have it go up one directory of the path of the file because otherwise it was trying to open to file.

szaimen:
> It doesn't make sense to have this option for folders

But when searching for specific folders, this option makes sense:
1. Search for a specific folder (based on the name)
2. See the list of results (probably consisting of files and folders)
3. Right klicking on the folder you want to open in his path
(Currently in file explorer the parental directory of this folder when accessing it through the list results, is the search. So here it makes sense to have this option imo.)

yaichenbaum:
This PR is specifically for the resents view on the home page, in search it definitely makes sense but at the moment we are not ready to share anything on search.

Issue #348, opened 2019-12-18
QuickLock Integration


lampenlampen:
# QuickLock integration

https://github.com/QL-Win/QuickLook

I have a early working [prototype](https://github.com/lampenlampen/files-uwp/tree/feat/QuickLockIntegration).

At the moment i can toggle QuickLock by launching an application. So it's basically possible.

## Things to Do

1. Communication between App and Launcher. At the moment i misuse the "Launching Application" signal for toggeling QuickLock. I would appreciate some input on how the communication should be designed.
2. Settings UI, where you can enable the Integration.
3. How QuickLock should be launched (key stroke, context menu, etc.). Most comon way is by pressing spacebar. Is there anything against using spacebar?

yaichenbaum:
By using the space bar it would be by holding it down for a few seconds?

szaimen:
@yair100 IMO, pressing the spacebar once (not holding it down) should open QuickLook. So that it is like on macos. 

(Pressing the spacebar once again while QuickLook is open should close it and refocus the latest FilesUWP Tab. But that is probably an issue that QuickLook has to solve...)

yaichenbaum:
I imagine this will only be if QuickLook is installed so I guess it wont affect regular users.

FelixZacher:
QuickLock provides a NamedPipe, where you can send the toggle-Command.

At first toogle-Command the preview window is shown, at the second-command the preview window is closed again or shows a new file preview, if a new path is send.

@yair100 I would include a settings option, where you have to turn the feature on, if you want to use it and if you have QuickLock installed.

yaichenbaum:
I wouldn't bother with the setting for it, if the user has it installed then it should work without any extra setup from the user.

yaichenbaum:
Thank you @lampenlampen for the amazing job you did with this!

lampenlampen:
# Icon

QuickLook has an [icon](https://github.com/QL-Win/QuickLook/blob/master/QuickLook/Resources/app_black.png) we can use.

I'm no expert in license stuff. For me it seems that QuickLook is under the GPL 3, which gives us the permission to use it.

yaichenbaum:
I am fairly sure we can use it as long as we credit the source, the about page in the app would probably be the best place to reference it.

yaichenbaum:
[...]

PR #353, opened 2019-12-22
QuickLook Integration

lampenlampen:
Solves #348.

At app start it is determined if QuickLook is installed. If yes then the "QuickLook" action becomes available in the "RightClickContextMenu" with the Keyboard shortcut "Space".

# Behaviour
QuickLook opens by selecting an item and pressing "Space". If you select another item and press "Space" the preview window changes it's content to the new item. If "Space" is pressed on the same item again, the preview window closes.

I only tested the behaviour with the Windows Store version of QuickLook, but there should be no difference between them.


Files/GenericFileBrowser.xaml.cs
```diff
@@ -70,6 +70,15 @@ public GenericFileBrowser()
             }
 
             App.OccupiedInstance.instanceViewModel.PropertyChanged += ViewModel_PropertyChanged;
+
+            // QuickLook Integration
+            ApplicationDataContainer localSettings = ApplicationData.Current.LocalSettings;
+            var isQuickLookIntegrationEnabled = localSettings.Values["quicklook_enabled"];
+
+            if (isQuickLookIntegrationEnabled != null && isQuickLookIntegrationEnabled.Equals(false))
```
yaichenbaum: I would have the visibility be a bool property and have the actual item in the context menu bound to it instead of controlling it here.

Files/GenericFileBrowser.xaml
```diff
@@ -147,6 +147,11 @@
                     <FontIcon Glyph="&#xE946;"/>
                 </MenuFlyoutItem.Icon>
             </MenuFlyoutItem>
+            <MenuFlyoutItem Click="{x:Bind AssociatedInteractions.ToggleQuickLook_Click}" Text="QuickLook" x:Name="QuickLook">
```
yaichenbaum: This is where I would add something like `Visibilty="{x:Bind ShowQuickLook}"`

yaichenbaum:
@lampenlampen Great job! There is one change I would like to see before merging in, otherwise everything looks great.

lampenlampen:
Maybe we can use the QuickLock icon in the menu?

yaichenbaum:
@lampenlampen I think that would be a nice addition, do you have a font icon or svg to use?

Issue #236, opened 2019-09-17
Add a way to replace this as the default file explorer


CrankySupertoon:
It can be doen with the registry but UWP apps i dont think have EXEs or do they. I know they are installed by a APPX but that i dont think outputs a EXE

yaichenbaum:
Do you have a link how it can be done from the registry? Its not something possible at the moment since there is no exe but having a link for reference can help for the future.

CrankySupertoon:
https://www.ghisler.ch/board/viewtopic.php?t=47997#

yaichenbaum:
Thank you! I will be keeping this issue open and will update later on if this will be a possibility.

J-Cake:
You could create a *proxy* executable file in like C++ and add that to the registry. Every time you then open the explorer, the exe file opens the app?

yaichenbaum:
If you want you can try adding that, at the moment we have been working on nailing the basics and haven't gotten a chance to look at this yet. This would need to be completely optional and off by default.

yaichenbaum:
@CrankySupertoon With the latest [release](https://github.com/duke7553/files-uwp/releases/tag/v0.7.0) there is an executable file included. If you want you are welcome to try to modify the registry to launch Files UWP, however modifying the registry can be risky and we take no responsibility if something goes wrong.

yaichenbaum:
@CrankySupertoon [Here](https://github.com/duke7553/files-uwp/wiki/Replacing-File-Explorer-with-Files-UWP) is a guide on how to replace File Explorer with Files UWP.

1333481:
> 
> 
> @CrankySupertoon [Here](https://github.com/duke7553/files-uwp/wiki/Replacing-File-Explorer-with-Files-UWP) is a guide on how to replace File Explorer with Files UWP.

Where did it go?

yaichenbaum:
@1333481 It's been moved to [here](https://files-community.github.io/docs/#/articles/replace-file-explorer).

PR #364, opened 2020-01-01
CommandPrompt Activation

FelixZacher:
# Description

I added the "AppExecutionAlias" `files.exe`, so the app can be launched from any command prompt.

In the folder `%localappdata%\Microsoft\WindowsApps` is now `files.exe` created, which helps #236.

I also did some testing with opening files in the current directory from the cmd, but there is currently a race condition between the app startup and the `PopulateDrivesListWithLocalDisks`-method, which does not populate the list fast enough. At a normal start from the cmdline, it always crashes, in the debugger with a breakpoint not. This only occurs if files wants to show the current directory.

# Questions

I would love some input on the cmdline-protocoll: Which commands, ...

PS:
The code as is does not support open in a directory, but works, in that it opens files. Don't know if you want to merge this as whole or as little parts.


lampenlampen:
# Do not merge

Please do not merge. Code is broken.

lampenlampen:
I added a parser and implemented the command:

```
files.exe -Directory C:\Users
```

!!!The crash is not fixed!!!

duke7553:
@lampenlampen Would threading be a way out of the race condition?

lampenlampen:
I'm currently trying to convert the list to a task-based approach, so we can await them.

duke7553:
@lampenlampen The WIP branch for the settings view model might simplify/remove some of the list items.

lampenlampen:
I had a look at the branch and i basically had the same idea. You can have a look at https://github.com/lampenlampen/files-uwp/commit/496a84e2c2f0efd7e3f6f4571d97bd297a13ce01.

I extracted all "drives"-related things to it's own class and added an `Drives`-property to the `SettingsViewModel`.

Please leave a short answer, if you want to refactor these yourself or i can go ahead.

duke7553:
Your implementation looks better than mine! For now, don't worry about my branch, as I can refactor those changes later

duke7553:
@lampenlampen I probably won't have the time to make any refactors myself until Friday at the earliest, so if you can that would be excellent!

lukeblevins (approved):
Looks great. We should now work on ironing out that race condition. @lampenlampen said he would look into it.

lampenlampen:
@duke7553 I saw that your are using "mvvmlight" and `ViewModelBase`. I had a brief look at their github repo, but docu is a bit rare there. Can you point me to something, that explains `ViewModelBase`.

lukeblevins:
This sounds like a question for @yair100 😀

yaichenbaum:
I would check out http://www.mvvmlight.net/

PR #376, opened 2020-01-10
Feat/drives refactoring

lampenlampen:
So i think i'm ready.

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -19,6 +19,7 @@ using Files.Filesystem;
 using System.IO;
 using System.Linq;
 using System.Collections.ObjectModel;
+using System.Threading.Tasks;
 using Windows.Devices.Enumeration;
 using Windows.Devices.Portable;
 using Windows.ApplicationModel.Core;
@@ -63,7 +64,7 @@ namespace Files
 
         public App()
         {
-            this.InitializeComponent();
+	        this.InitializeComponent();
             this.Suspending += OnSuspending;
             consentDialog = new Dialogs.ConsentDialog();
             propertiesDialog = new Dialogs.PropertiesDialog();
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -19,18 +19,9 @@ using Files.Filesystem;
 using System.IO;
 using System.Linq;
 using System.Collections.ObjectModel;
-using System.Threading.Tasks;
-using Windows.Devices.Enumeration;
-using Windows.Devices.Portable;
-using Windows.ApplicationModel.Core;
 using Windows.UI.Core;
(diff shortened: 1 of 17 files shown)


duke7553:
@lampenlampen Startup Crash info: 
Windows.Graphics.Display: GetForCurrentView must be called on a thread that is associated with a CoreWindow.

duke7553:
I'm guessing it has to do with the consent dialog which displays on startup because this exception only occurs when running in a debugger and file system access is disabled.

lampenlampen:
So i can reproduce the crash, if the debugging option "Only my code" is enabled. If disabled, the crash does not occur.

lampenlampen:
A guess:
`consentDialog.ShowAsync()` is somewhere called, before the Application `CoreWindow` is created, thus it fails to display the dialog ui.
And somehow the debugger delays the creation of the `CoreWindow`.

lampenlampen:
I could fix it, by delaying the creation of the `SettingsViewModel`, until the constructor of `App` is called.

duke7553:
@lampenlampen I'm still encountering a crash for some reason on the latest changes

lampenlampen:
Yes there is a `NullReferenceException` at https://github.com/duke7553/files-uwp/blob/69902d4bb3e3f2d46b33c842057d2a982d66aaa5/Files/Filesystem/ListedItem.cs#L45
because `datetimeformat` is `null` at a fresh install.

We need to set a default value at startup, if it is null.
At https://github.com/duke7553/files-uwp/blob/69902d4bb3e3f2d46b33c842057d2a982d66aaa5/Files/View%20Models/SettingsViewModel.cs#L45
we should set a default value, if `datetimeformat` is null.

What should the default value be?

yaichenbaum:
The setting view model should have a default for every setting, the default for date and time format should be friendly date and times. For example "2 days ago".

lampenlampen:
So we can choose between `Application` and `System`. I set it now to `System`.

yaichenbaum:
I think `Application` is the one with friendly dates...

lampenlampen:
They are both with friendly dates.

<table>
<tr>
<th>
Application
</th>
<td>
`4 days ago`, `15 hours ago`, `Saturady, October 12, 2019`
</td>
</tr>
<tr>
<th>
System
</th>
<td>
`4 days ago`, `15 hours ago`, `10/12/2019 5:14 PM`
</td>
</tr>
</table>

yaichenbaum:
Ok, so lets set the default to application. That is how it was in previous versions of Files as well.

yaichenbaum:
Great, is this pull request ready for review now?

lampenlampen:
Yes fixed a last bug, but should be ready now 👍.

Issue #425, opened 2020-02-25
Logger to help debugging


lampenlampen:
We should consider implementing a logger to help debugging bugs that are not reproducible on your own machines like #410.

We could use [NLog](https://nlog-project.org/) and write log statements and all crashes to a file, so we can see what crash hapened and where.

yaichenbaum:
This would be very helpful.

PR #426, opened 2020-02-27
NLog (Logger)

lampenlampen:
# NLog (Logger) (#425)

I implemented a logger and used NLog.

Until now only unhandled Exception are logged and the app start.

I would suggest as development continues we gradually update the app with logging statements.

Edit 1: The output is directed into a file `debug.txt` in the app's local folder.
Edit 2: It seems Exceptions in the App constructor are not handled.


yaichenbaum:
Thank you so much!

PR #459, opened 2020-03-15
Fixed the problem of Settings page when in non-English environment.

creeperlv:
Since File UWP supports multi-language, settings page is broken, so I find a way of fixing it.


yaichenbaum:
@creeperlv Thank you for opening this pull request, this issue has already been fixed in both the `develop` and `master` branches, can you verify this is still an issue with v0.7.1?

yaichenbaum:
What we are currently doing is checking the tag, which is not localized, but after looking over your commit, I believe it is a cleaner solution.

creeperlv:
<del>Yes, I checked the code in `master`, I find that you are using ToString() to determine which page the user wants to view, however, the in non-English environment, it will return other string instead of `Appearance` ...</del>
Thank you.

yaichenbaum:
@creeperlv Before I merge your pull request, would you be able to resolve the merge conflicts?

creeperlv:
I am new of using pull requests, I have no idea of resolving conflicts. Could you help me?

creeperlv:
Should I just overwrite lines in `master` to lines in `develop`?

yaichenbaum:
In this case, since the change is small, I would recommend copying your changes over to the `develop` branch and then to create a new pull request. We recommend changes made to the code base to be based on the latest in `develop` to avoid scenarios like this.

creeperlv:
Okay, thx. I'll do it immediately.

yaichenbaum:
@creeperlv Thank you so much, we appreciate your help to improve Files UWP 🙂

creeperlv:
You are welcome, I like this software. 😆

PR #505, opened 2020-04-04
Perform french translations

KaKi87:
Remaining questions :

- [ ] What are *Flags* (`SettingsFlagsTitle.Text`) ?
- [ ] What is *QuickLook* (`BaseLayoutItemContextFlyoutQuickLook.Text`) ?
- [ ] Why putting `...` at the end of some entries ?
- [ ] Shouldn't placeholder be formulated as affirmative, not imperative ? (`NavigationToolbarVisiblePath.PlaceholderText` is formulated *Enter a path*)

### DIFF
--- a/Files/MultilingualResources/Files.fr-FR.xlf
+++ b/Files/MultilingualResources/Files.fr-FR.xlf
@@ -192,31 +192,31 @@
         </trans-unit>
         <trans-unit id="ConfirmDeleteDialog.Title" translate="yes" xml:space="preserve">
           <source>Delete Item(s)</source>
-          <target state="new">Delete Item(s)</target>
+          <target state="new">Supprimer le(s) élément(s)</target>
         </trans-unit>
         <trans-unit id="ConfirmDeleteDialogCancelButton.Content" translate="yes" xml:space="preserve">
           <source>Cancel</source>
-          <target state="new">Cancel</target>
+          <target state="new">Annuler</target>
         </trans-unit>
         <trans-unit id="ConfirmDeleteDialogDeleteButton.Content" translate="yes" xml:space="preserve">
           <source>Delete</source>
-          <target state="new">Delete</target>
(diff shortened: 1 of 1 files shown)


yaichenbaum:
1. Flags is not yet implemented but the basic idea is to have a settings page to test out experimental features.
2. [QuickLook ](https://github.com/QL-Win/QuickLook) is a 3rd party integration that lets you preview files by pressing the space bar.
3. Each string with `...` most likely has it's own explanation 😃
4. `PlaceholderText` is the property that holds the text so the string needs to include that.

KaKi87:
> 3. Each string with `...` most likely has it's own explanation 😃

`Open in terminal` and `Open with` right-click items usually doesn't have `...` at the end.

> 4. `PlaceholderText` is the property that holds the text so the string needs to include that.

What I mean is that there's usually a difference between this :

2020-04-05 13_40_09-JSitor - NodeJS, JavaScript, HTML, CSS, online editor and 13 more pages - Person

And this :

2020-04-05 13_40_22-JSitor - NodeJS, JavaScript, HTML, CSS, online editor and 13 more pages - Person

yaichenbaum:
3. Many of the strings were added by different people so it is likely this is just because everyone uses their own formats.
4. I agree, but since the property includes the fact that it is a placeholder, it would be a little redundant to add it to the string name.

duke7553:
@KaKi87 You're actually correct about the "Open in Terminal" menu item. I had to re-read the Windows design documentation which states the ellipsis should only be used when more information will be needed for the item's command.

duke7553:
Screenshot_20200405-133608

KaKi87:
> 4. I agree, but since the property includes the fact that it's a placeholder, it would be a little redundant to add it to the string name.

No, what I mean is that "Enter a path" should simply changed by "Path".

Issue #37, opened 2019-02-17
Translations


p0nkr4t:
Hello!
It would be great to add translations to other languages.
If you want (and if it's possible), I can help with Russian translation.
Thanks!

SimonTek27:
Hello,
i offer my help with italian language
thanks for your efforts

duke7553:
Thanks, we can definitely plan this for the initial Microsoft Store release.

Jviix:
Hey, I'd like to help with Spanish translation. So feel free to count with me :)

duke7553:
Wow, I will put it on my list to create a list of strings to be translated. I’ll keep you guys posted.

GuusLieben:
I'll help with Dutch translations once the translations list is available :)

KaKi87:
@PoorPocketsMcNewHold We'll collaborate to the french translation 😉

Margen67:
The title of the issue is misspelled: `Transaltions`

duke7553:
@Margen67 Thanks

themoiFR:
I can help setting up the Globalization and localization if needed and I can translate in french

yaichenbaum:
@themoiFR Thank you for the offer, we are definitely planning on having this done before the launch of Files UWP. I am not sure we are ready for translations at the moment since a lot of strings are going to be changing as we get the app ready for production. I would say that we will be ready for this closer to March.

sladkOy:
If anything, then I can help with the translation into Russian and Ukrainian.

d2dyno006:
I can help with the polish one

lorisobi:
I can help translating to german

yaichenbaum:
### We are going to be starting work on this much-requested option in the next few days 🚀
Adding the strings should take about a week or two and after that, we will be ready for everyone to translate to whatever languages they want.

I will let everyone know when the strings are ready and I will provide steps so you can get started. I would appreciate if everyone who is still interested can let me know. I am looking forward to everyone's contributions 😊

KaKi87:
I am still interested in translating from English to French and I am still waiting for the next release which is supposed to close so much issues and feature requests. xD

R3voA3:
I can provide help with the German translation.

sladkOy:
> ### We are going to be starting work on this much-requested option in the next few days 🚀
> Adding the strings should take about a week or two and after that, we will be ready for everyone to translate to whatever languages they want.
> 
[...]

PR #541, opened 2020-04-08
full translation into Turkish Files.tr-TR.xlf

BuraChuhadar:


BuraChuhadar:
#37

yaichenbaum:
@BuraChuhadar Just a heads up for the future, when creating pull requests, make sure to base it off of the `develop` branch. This time happened to be fine since `develop` is currently even with `master` but that usually is not the case, so it's best to just work off of `develop`.

PR #562, opened 2020-04-09
Improvements to work done by @tsvietOK 

duke7553:
Improvements to work done by @tsvietOK 

In addition to his contributions, we now handle serialization exceptions during the deserialization process by reverting to the default config file's contents.

I'm leaving both PRs open for peer review purposes, but I happen to think mine is more complete unless anyone disagrees.


tsvietOK:
There is possible JsonReaderException in case of manual json editing by user(for example user made a mistake in json structure). So, i think better to unify all exceptions in this part and if for any reason it can't deserialize json, just reload it from assets.

PR #565, opened 2020-04-10
Added some tooltips

MarcAnt01:


duke7553:
@MarcAnt01 Thanks for this PR! I'm going to wait and see if we can localize these.

tsvietOK:
> @MarcAnt01 Thanks for this PR! I'm going to wait and see if we can localize these.

image
We can!

PR #569, opened 2020-04-11
Sort filenames naturally using CompareStringEx

### LINKED ISSUE #352: Natural Sorting

File names should be compared by the Win32Api `StrCmpLogicalW ` function to get the default Windows behavior. It's somehow strange to get the name `abc10` listed between `abc1` and `abc2`.

File names should be shown as:

```
2string
3string
20string
st2ring
st3ring
st20ring
string2
string3
string20
```

but not as (currency)

```
20string
2string
3string
st20ring
st2ring
st3ring
string2
string20
string3
```

File names should be listed like in the first example.

jeffsieu:
Adds a `NaturalStringComparer` class which is used in `ItemViewModel` when sorting by file name.

image

Resolves #352.

### DIFF
--- a/Files/Files.csproj
+++ b/Files/Files.csproj
@@ -150,6 +150,7 @@
       <DependentUpon>ConfirmDeleteDialog.xaml</DependentUpon>
     </Compile>
     <Compile Include="Helpers\ItemsDataTemplateSelector.cs" />
+    <Compile Include="Helpers\NaturalStringComparer.cs" />
     <Compile Include="UserControls\NavigationToolbar\INavigationToolbar.cs" />
     <Compile Include="UserControls\NavigationToolbar\ModernNavigationToolbar.xaml.cs">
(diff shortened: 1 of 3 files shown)


duke7553:
@jeffsieu Thanks for this contribution. My only concern is whether or not this method is allowed in UWP apps.

jeffsieu:
@duke7553 Would using [CompareStringEx](https://docs.microsoft.com/en-sg/windows/win32/api/stringapiset/nf-stringapiset-comparestringex?redirectedfrom=MSDN) from api-ms-win-core-string-l1-1-0 instead be better? As it is [present in all Windows 10 devices](https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-apis)

duke7553:
@jeffsieu Yes, because that one is technically allowed in Universal apps. (I actually overlooked that function on the list, or else I would have fixed this long ago)

jeffsieu:
@duke7553 Here, this should do it.

PR #581, opened 2020-04-12
Properties hash

yaichenbaum:


duke7553:
image

Files/Interacts/Interaction.cs
```diff
@@ -1182,5 +1187,19 @@ public void PushJumpChar(char letter)
         {
             App.CurrentInstance.ViewModel.JumpString += letter.ToString().ToLower();
         }
+
+        public async Task<string> GetHashForFile(ListedItem fileItem, string nameOfAlg)
```
yaichenbaum, line 1191: Can you rename this to `GetMD5HashFromFile`?

duke7553: Actually, if we pass in an alternative second parameter, we can get different types of hashes.

duke7553: That's why I left it as a generic name just in case.

yaichenbaum: OK, that is very good.

PR #582, opened 2020-04-12
Tweaked how navigating to the settings page works

yaichenbaum:


tsvietOK:
Here is how settings page looks at russian language.
image
Last string trimmed, full string is "Боковая панель с эффектом акрила"

tsvietOK:
I want to suggest to put all controls on this page(Appearance) in one grid with first column width =Auto or put combo boxes and toggle buttons under their captions
Second, let's set same width for all combo boxes.

Files/Views/Pages/Settings.xaml
```diff
@@ -40,6 +50,14 @@
                 </ResourceDictionary>
             </muxc:NavigationView.Resources>
 
+            <muxc:NavigationView.PaneHeader>
+                <TextBlock
+                    Margin="10,6"
```
tsvietOK, line 55: Here we need x:Uid for string translation

Files/Views/SettingsPages/Appearance.xaml
```diff
@@ -45,7 +45,7 @@
                         Grid.Column="0"
                         VerticalAlignment="Center"
                         FontSize="16"
-                        Text="Theme" />
+                        Text="Choose an app mode" />
```
yaichenbaum, line 48: I would think the "Theme" is more descriptive, app mode can mean a number of other things such as tablet mode vs desktop mode...

Files/Views/SettingsPages/About.xaml
```diff
@@ -67,8 +70,8 @@
                                 <TextBlock
                                     x:Uid="SettingsAboutSubmitFeedbackDescription"
                                     Foreground="Gray"
-                                    Text="Submit an issue report to provide the developers with information"
-                                    TextWrapping="Wrap" />
+                                    Text="Send the developers an issue report with more information"
```
yaichenbaum, line 73: Unless this is changed in the en-US resource file, the text will show the changes made here.

Files/MultilingualResources/Files.de-DE.xlf
```diff
@@ -575,7 +575,11 @@
           <source>MD5Hash:</source>
           <target state="translated">MD5Hash:</target>
         </trans-unit>
+        <trans-unit id="SettingsPageHeader.Text" translate="yes" xml:space="preserve">
```
tsvietOK, line 578: There is already exists translated string Settings under id "SettingsNav.PaneTitle", so you can change name to "SettingsPageHeader.Text" and no need to translate it again

Files/Views/SettingsPages/Appearance.xaml
yaichenbaum: I am changing it to "App theme", feel free to revert this change if you prefer it the other way.

duke7553: Hi, this change was to follow Microsoft guidelines for App Settings

tsvietOK: @duke7553 Another Microsoft app
image

yaichenbaum: While following the Microsoft Guidelines are good, I don't see any reason to follow this one as it seems to make the description less clear.

PR #584, opened 2020-04-12
Fixed issue with strings

yaichenbaum:


tsvietOK:
Regarding this issue, looks like some users manually editing xliff translation file and it's wrong, because those files regenerates automatically when new strings available in Strings/en-US/Resources.resw file.
We should clarify how to add new strings and how to translate them.
If you want to add new strings, add them to Resources file above (literally en-US folder) and  build project. After that new strings will be added to xliff files with mark "new" and will be available to translation. Here you can translate them manually ot using Multilingual App Toolkit Editor. Make sure that translated strings marked as "translated" instead of "new". Then build project again, this will add translated strings to Strings files automatically. 
Make sure you install toolkits from Translator.md available in project folder.
If you want only translate strings,  use instructions from Translator.md.

tsvietOK:
Oh, this bug caused by merge commit 4dbade5fa2f0bae191751b4967735828d7819ebb. Anyway, maybe above information could be useful for someone.

PR #551, opened 2020-04-08
Extend view into titlebar of properties window

lampenlampen:
I made a change to extend the view into the titlebar of the properties window and added some spacing around the header.

files_properties_titlebar_light

I orientated myself by this [concept](https://i.redd.it/jqmrzvwc7pj41.png).

PR #598, opened 2020-04-14
Redesign Properties window

tsvietOK:
-MD5 field visible only for files properties
-Translated strings
-Added reveal button style to top navigation buttons like Back, Forward...
-Added checkboxes for future implementations of attributes change(currently disabled)


tsvietOK:
image
image

yaichenbaum:
Locking this pull request until #551 is merged.

duke7553:
We can keep the conversation going, but this one must wait for the other PR.

lampenlampen:
Hey,
since this pr includes all the changes of #551, i would close mine and merge this one.

It looks a lot better than mine already and i would work from this one, if i want to change something 👍.

Files/Properties.xaml
```diff
@@ -6,139 +6,220 @@
...
+                    <RowDefinition />
+                    <RowDefinition />
+                    <RowDefinition />
+                    <RowDefinition />
+                </Grid.RowDefinitions>
+                <Grid
+                    x:Name="Icon"
+                    Width="80"
+                    Height="80"
+                    Margin="0,0"
```
yaichenbaum, line 109: Since the margin is 0, I think we can remove it from the xaml.

Files/Properties.xaml
```diff
@@ -6,139 +6,220 @@
...
+                            <StaticResource x:Key="NavigationViewTopPaneBackground" ResourceKey="RibbonBackgroundColor" />
+                        </ResourceDictionary>
+                        <ResourceDictionary x:Key="HighContrast">
+                            <!--  This empty dictionary ensures that the default high contrast resources are used when the user turns on high contrast mode.  -->
+                        </ResourceDictionary>
+                    </ResourceDictionary.ThemeDictionaries>
+                </ResourceDictionary>
+            </ResourceDictionary.MergedDictionaries>
+        </ResourceDictionary>
+    </Page.Resources>
```
yaichenbaum, line 77: What is this template being used for?

Files/Properties.xaml.cs
```diff
@@ -40,18 +39,22 @@ private async void Properties_Loaded(object sender, Windows.UI.Xaml.RoutedEventA
...
-                    // Not a folder, so attempt to get as StorageFile
                     selectedStorageItem = await StorageFile.GetFileFromPathAsync(selectedItem.ItemPath);
                     var hashAlgTypeName = HashAlgorithmNames.Md5;
 
-                    ItemProperties.ItemMD5Hash = await App.CurrentInstance.InteractionOperations.GetHashForFile(selectedItem, hashAlgTypeName); // get file hash
+                    // get file hash
+                    ItemProperties.ItemMD5Hash = await App.CurrentInstance.InteractionOperations.GetHashForFile(selectedItem, hashAlgTypeName);
+
+                    PropertiesMD5Hash.Visibility = Windows.UI.Xaml.Visibility.Visible;
+                    itemMD5HashValue.Visibility = Windows.UI.Xaml.Visibility.Visible;
```
yaichenbaum, line 53: We shouldn't be referencing the xaml directly from code-behind, it would be better to add a property to a view model for this.

Files/Properties.xaml
tsvietOK: For acrylic background of Properties window

yaichenbaum: Can you just use something like `SystemControlAcrylicWindowBrush`?

tsvietOK: After setting Background="{ThemeResource SystemControlAcrylicWindowBrush}" there is no any acrylic blur, it is not working.

yaichenbaum: That is strange, I think we should try to figure out the cause before adding what really shouldn't be necessary here.

PR #604, opened 2020-04-15
Add Invert Selection, icons to status bar, fix folder navigation and remove space before .dll

### LINKED ISSUE #274: Selection options
Please add these ribbon options for selection:

- [x] Select all
- [x] Select none
- [x] Invert selection

Would be really helpful especially for touch input.

This is how it is currently implemented in the file explorer:
image

### LINKED ISSUE #601: The new sorting feature introduced with v0.7.6 crashes the app when navigating to a directory
To reproduce this issue, build the app with release mode and try navigating to a directory.

jeffsieu:
Resolves #601.
Resolves #274.

Previously, when the user navigates into the folder using double click, sometimes it may register as a drag. The program then tries to take the selected files and put it into a `DataPackage` to start the drag process. But when there are no selected files, the `DataPackage` throws an error.

This commit adds a check to see if the selected items list is empty. If it is, cancel the drag gesture.

There's also a bug that causes a folder to delete itself, when it is dragged into itself and "skip" is pressed on the dialog that comes afterwards.

Also adds an Invert Selection and some icons to the menu in the status bar.
image

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -126,6 +126,8 @@ namespace Files
 
         protected abstract void SetSelectedItemsOnUi(List<ListedItem> selectedItems);
 
+        public abstract void FocusSelectedItems();
+        
         protected abstract ListedItem GetItemFromElement(object element);
 
         private void AppSettings_LayoutModeChangeRequested(object sender, EventArgs e)
@@ -321,6 +323,11 @@ namespace Files
                     selectedStorageItems.Add(await StorageFolder.GetFolderFromPathAsync(item.ItemPath));
             }
 
+            if (selectedStorageItems.Count == 0) {
+                e.Cancel = true;
+                return;
+            }
+
             e.Data.SetStorageItems(selectedStorageItems);
             e.DragUI.SetContentFromDataPackage();
         }
@@ -354,7 +361,7 @@ namespace Files
         protected void InitializeDrag(UIElement element)
         {
             ListedItem item = GetItemFromElement(element);
-            if(item != null)
+            if (item != null)
             {
                 element.AllowDrop = false;
                 element.DragStarting -= Item_DragStarting;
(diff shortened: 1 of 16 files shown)


tsvietOK:
You can test it, just build in release mode. For me it is working fine.

jeffsieu:
@tsvietOK Not sure if this is a problem with my computer, but I used to be able to debug Files just fine in the past after enabling the storage permissions in Windows once. But now I've to keep turning the permissions back on as it seems to reset every time I debug the app... Is this also a problem for you?

tsvietOK:
@jeffsieu For some reason when debugging VS every time reinstall app as new, so logically we are have to reenable permission every time. This is uncomfortable for me, but i don't think it is a bug

tsvietOK:
And it is not part of this issue

jeffsieu:
@tsvietOK Okay, I was just wondering if it's my computer because last time I was able to enable it only once...

jeffsieu:
I think I found the problem. When the user navigates into the folder using double click, sometimes it may register as a drag. The program then tries to take the selected files and put it into a `DataPackage` to start the drag process. But when there are no selected files, the `DataPackage` throws an error.

So I have added a check; if the selected items list is empty, cancel the drag gesture.

tsvietOK:
Please, add new string to resources(en-US) and build project(it will add new string to xlf files)

tsvietOK:
@jeffsieu There is another issue with selection, try select folder and unselect it. Now you won't be able to select this folder. To select this folder you have to select another folder and then this.

jeffsieu:
@tsvietOK Yes, I also realised that. But I realised that that is already an existing problem in the current `develop` branch... And I'm not sure why it does that.

duke7553:
I like these changes. We'll get around to reviewing them later this week

duke7553:
@jeffsieu I'm reviewing this now

PR #610, opened 2020-04-16
Fix Download folder location

### LINKED ISSUE #384: Wrong Downloads folder location
**Describe the bug**
Files UWP automatically managed to figure out the location for the Documents, Pictures , Music and Videos folders but for some reason it failed to do so with the Downloads one. All these folders including the Downloads one are straight on my E: drive.
I am aware one can manually change where these folders are located in the settings but since the automatic detection works for the other folders I'd consider it a bug that still should be fixed.

**To Reproduce**
1. Have custom location for downloads folder since before.
2. Install and run Files UWP.
3. Click on "Downloads".
4. See it take you to the default/standard location instead of the custom one.

**Expected behavior**
Expected it to show me the custom location automatically since it worked with all the other folders with custom location.

**Desktop:**
 - OS Version: Windows 10 1909 18363.535
 - App version: v0.6.6.0

tsvietOK:
Get other default folders locations same way
Fixes #384

### DIFF
--- a/Files/View Models/SettingsViewModel.cs
+++ b/Files/View Models/SettingsViewModel.cs
@@ -470,37 +470,37 @@ namespace Files.View_Models
 
         public string DesktopPath
         {
-            get => Get(Environment.GetFolderPath(Environment.SpecialFolder.DesktopDirectory));
+            get => Get(UserDataPaths.GetDefault().Desktop);
             set => Set(value);
         }
 
         public string DocumentsPath
         {
-            get => Get(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments));
+            get => Get(UserDataPaths.GetDefault().Documents);
             set => Set(value);
         }
 
         public string DownloadsPath
         {
-            get => Get(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile) + @"\Downloads");
+            get => Get(UserDataPaths.GetDefault().Downloads);
             set => Set(value);
         }
 
         public string PicturesPath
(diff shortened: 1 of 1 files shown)


duke7553:
@tsvietOK Excellent work! I didn't know UserDataPaths existed

lukeblevins (approved):

yaichenbaum:
This change actually doesn't do anything, we actually set this data from the registry editor and this issue has already been fixed. I think this should be reverted.

lukeblevins:
@yair100 I don't think we necessarily need to revert this because it does improve the detection of the downloads folder path. Before, we simply appended @"\Downloads" to the root user data folder path. 

If it is still a problem after this PR, we can work on improving the registry detection, but for now it looks like a decent improvement. (Unless there are other objections)

tsvietOK:
@yair100 I think you are not right. As i understand, this part with registry detection in Files.Launcher don't even running, try to debug Program.cs.

yaichenbaum:
@tsvietOK If that is the case we have some room for improvements in this area, if your change fixes the issue, then we should be able to remove the code from Files.Launcher.

tsvietOK:
@yair100 it's smth strange, it is workig even if i change this vars to random symbols

image

yaichenbaum:
I think we need to move the startup task to the UWP project, we are changing the value of the settings there but the setting view model isn't updated.

yaichenbaum:
@tsvietOK Can you take a look at #614?

PR #614, opened 2020-04-17
Refactored detecting user locations

yair100:


tsvietOK:
@yair100 The problem was that folder path written in local settings under some key(for example DesktopPath) and from another side(UWP) it has not been read. You trying to get variable DesktopPath instead of local settings key.
What about this PR, it's working fine.

yaichenbaum:
@tsvietOK This PR just changes it that the value doesn't get stored across sessions since there is the possibility that the user changed the path since first running Files UWP.

lukeblevins (approved):
Approved as long as the UWP UserDataPaths account for custom library paths. (Which I believe they do)

PR #619, opened 2020-04-17
Fixed crash when opening folders in GridView

yair100:
Fixed an issue where opening folder in the GridView layout would sometimes crash the app.

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -934,7 +934,7 @@ namespace Files.Interacts
 
             if (acceptedOperation == DataPackageOperation.Move)
             {
-                foreach (IStorageItem item in pastedItems)
+                foreach (IStorageItem item in itemsToPaste)
                 {
                     if (item.IsOfType(StorageItemTypes.File))
                     {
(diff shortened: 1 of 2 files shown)


tsvietOK:
@yair100 this PR not fixes first folder opening, it fixes crash on other folder opening(except first)

tsvietOK:
By the way, dragndrop is not working(drop part).

tsvietOK:
@yair100 dragndrop fixes: this [line](https://github.com/duke7553/files-uwp/blob/e27f216614778dc26c52a7a442ab844c3831ae7a/Files/Interacts/Interaction.cs#L937) replace "pastedItems" with "itemsToPaste". I think we can add this change to this PR.

yaichenbaum:
@tsvietOK Have you tested that change and confirmed that it works?

tsvietOK:
@yair100 Yes. Before this change if you try to move one folder onto another folder it copying it and after removing just copied folder. Now it works

tsvietOK:
Logically, i don't think it is a good idea to delete just moved files:)

lukeblevins (approved):

PR #620, opened 2020-04-18
Fix folder copying to sibling folder with almost the same name

tsvietOK:
To reproduce this bug create folder with name "Some Folder" and another sibling folder with name "Some Folder 1".  Try to copy or move first folder in second.


duke7553:
@tsvietOK Hi thanks for this PR.

The original issue that this block of code tried to solve was users copying a folder into itself. Does this change still prevent all instances of a destination folder being a subfolder of the source folder?

image

tsvietOK:
@duke7553 hmmm, your right, i didn't test this case. I have to find another workaround.

duke7553:
@tsvietOK I'll mark this as a draft for now then.

tsvietOK:
Taken from https://stackoverflow.com/questions/5617320/given-full-path-check-if-path-is-subdirectory-of-some-other-path-or-otherwise
@duke7553  what do you think about it?

duke7553:
@tsvietOK That looks pretty good to me

tsvietOK:
@duke7553 do you have troubles with folder copying?

duke7553:
@tsvietOK Do you want me to test this now? I can mark this as ready for review.

tsvietOK:
@duke7553 If your copy function working fine, of course.

tsvietOK:
@duke7553 did you tested it?

Files/Interacts/Interaction.cs
```diff
@@ -888,7 +889,7 @@ public async Task PasteItems(DataPackageView packageView, string destinationPath
             {
                 if (item.IsOfType(StorageItemTypes.Folder))
                 {
-                    if (destinationPath.Contains(item.Path, StringComparison.OrdinalIgnoreCase))
```
yaichenbaum, line 891: Instead of adding a full class and removing this line, would it work to change `Contains` to `Equals`?

tsvietOK: "D:/Some Folder" copy to "D:/Some Folder/Some Folder another/...". Second folder is subfolder for first folder. But paths are not equals

yaichenbaum: Ok, thanks for explaining.

yaichenbaum:
@tsvietOK Can you merge with the latest from `develop`?

yaichenbaum:
Before merging, this should be tested to make sure that the changes work in release mode.

tsvietOK:
@yair100 Part with subfolder detection working fine for me. Another thing, we currently don't  supporting Windows's path length limit(as i understand 260 characters)

yair100 (approved):

yaichenbaum:
@tsvietOK I would think that is a different issue, right?

tsvietOK:
@yair100 Yes, your right

PR #625, opened 2020-04-20
Added option to open sidebar items in a new tab or window

### LINKED ISSUE #331: "Open in new tab/window" for pins and favorites
We should be able to right click on a folder from the left column and have the option to open it in a new tab or window, just like we can do with folders in the main navigation area

yaichenbaum:
Closes #331.


Files/UserControls/ModernSidebar.xaml.cs
```diff
@@ -118,11 +120,17 @@ private void NavigationViewItem_RightTapped(object sender, RightTappedRoutedEven
         {
             Microsoft.UI.Xaml.Controls.NavigationViewItem sidebarItem = (Microsoft.UI.Xaml.Controls.NavigationViewItem)sender;
             var item = sidebarItem.DataContext as LocationItem;
-            if (!item.IsDefaultLocation)
+            if (item.IsDefaultLocation)
             {
-                SideBarItemContextFlyout.ShowAt(sidebarItem, e.GetPosition(sidebarItem));
-                App.rightClickedItem = item;
+                UnpinItem.Visibility = Visibility.Collapsed;
```
yaichenbaum, line 125: This shouldn't reference the xaml directly, it should instead set a value which is bound to the xaml.

PR #690, opened 2020-04-24
Fixes + checks (when launching from command line)

ADeltaX:
Fixed typos
Fixed path parsing
Added check for path
Added icons in tabs when launching from cmdline


ADeltaX:
This PR should address some (all?) issues of setting this app as default file manager.

yaichenbaum:
@ADeltaX Thank you for your contribution!

Issue #563, opened 2020-04-09
Drives disappear from side menu


Darhagonable:
**Describe the bug**
Sometimes when starting the Files UWP app the the drives section of the side menu is missing.

**To Reproduce**
Not sure. It just happens sometimes when i start the program. Is there anything i could do to test it further?

**Expected behavior**
The section with the drives to exist in the side menu.

**Screenshots**

**Desktop:**
 - OS Version: Windows 10 1909 18363.657
 - App version: v0.7.5

**Additional context**
![image](https://user-images.githubusercontent.com/14127357/78938503-a1783b00-7ab2-11ea-8a32-5d0e892088df.png)

duke7553:
@Darhagonable Yes! Could you please paste your log file? You can get it by going to the folder %localappdata%/Packages, search there for the folder Files and there in the folder LocalState should be a file debug.txt, which contains a stacktrace.

Darhagonable:
I don't seem to have any Files folder that updated recently. I built it myself in VS and didn't install it through MS Store if that could be why?
I don't know much about VS but maybe there's a log in there?

tsvietOK:
@duke7553 Log file is not useful.
Start with disks loaded
`2020-04-11 04:09:09.1492|INFO|Files.Filesystem.DrivesManager|Drive added: C:\, Fixed
2020-04-11 04:09:09.2066|INFO|Files.Filesystem.DrivesManager|Drive added: S:\, Fixed
2020-04-11 04:09:09.2273|INFO|Files.Filesystem.DrivesManager|Drive added: V:\, Fixed
2020-04-11 04:09:09.2471|INFO|Files.App|App launched`

Start without disks loaded
`2020-04-11 04:09:13.1606|INFO|Files.Filesystem.DrivesManager|Drive added: C:\, Fixed
2020-04-11 04:09:13.2205|INFO|Files.Filesystem.DrivesManager|Drive added: S:\, Fixed
2020-04-11 04:09:13.2400|INFO|Files.Filesystem.DrivesManager|Drive added: V:\, Fixed
2020-04-11 04:09:13.4978|INFO|Files.App|App launched`

Darhagonable:
I seem to be able to temporarily fix it by doing this:
- Turning off access to the file system for the app while it's running.
- Start Files UWP again.
- Turning on access to the file system for the app while it's running.
- Start Files UWP again.
- Close it.
- Start it again and it's fixed

yaichenbaum:
[...]

PR #701, opened 2020-04-26
Fix drives loading

tsvietOK:
For me removing changes added in 02971da228fde684d2128b0a0802e0b3def6fc70 by @duke7553 fixing issue #563 . Another way to fix that, just remove "await" from dispatcher. Problem somewhere in dispatcher. And when try-catch catched an error "Element not found", drives did not load.


yaichenbaum:
I think that change was done to fix an issue where the app was crashing for many users. Are you sure this won't bring back that issue?

tsvietOK:
@yair100 May be there was another crash reason?

lukeblevins (changes requested):
While this may fix the issue, we should instead focus on deferring the loading of drives until the UI thread is created instead of always/never loading them.

duke7553:
@tsvietOK Thanks for narrowing down the issue, though-this information can be used to provide a solution for the drives not loading.

tsvietOK:
@duke7553 What do you mean under "instead of always/never loading them"? And why do you want to load drives after UI thread started?

duke7553:
The current implementation will never load the drives list if the UI thread isn't available. We know if an exception is thrown, then the UI thread doesn't exist. iirc the reason we have to populate the drives collection on the UI thread is to prevent occasional crashes on startup.

Therefore, I think the best approach here would be to handle the [catch] by deferring the drives from loading until the page is loaded. Right now, we simply never load the drives if the UI thread doesn't exist yet.

tsvietOK:
@duke7553 I don't agree with current implementation. We can load drives in any time, so when binded collection "sideBarItems" will be changed, ModernSidebar will display drives automatically Try-catch block is unneeded.

duke7553:
@tsvietOK Since we have no definitive evidence that the current implementation fixed any crashes, I'll approve this change because it fixes a bug.

lukeblevins (approved):

PR #706, opened 2020-04-26
Fix properties window crash

### LINKED ISSUE #704: The app crashes when trying to display the properties of an iso file
This happened with a 24 GB and 4 GB iso
I'm not sure if it happens only with large iso files

crash

tsvietOK:
Crash caused by MD5 hash algorithm calculation, it did not support large files.
Now file properties are shown before MD5 calculation. Added indeterminate progress bar.
Fixes #704

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -38,6 +38,7 @@ using Windows.Storage.Streams;
 using GalaSoft.MvvmLight.Command;
 using Files.Helpers;
 using Windows.UI.Xaml.Data;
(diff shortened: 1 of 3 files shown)


xPoppyx:
Is this related to what I had mentioned in #704 ? about the time it takes to show the properties of a slightly large file, or does that need a different fix?

yaichenbaum:
@xPoppyx This pull request should address it.

tsvietOK:
@xPoppyx It working good with 8GB file, so with 24GB it should work too.

PR #708, opened 2020-04-26
Fix file size calculation

tsvietOK:
Before
image
After
image


tsvietOK:
Is it better to move file size calculation to separate (static) class? And which class?

yaichenbaum:
What is this supposed to fix? Comparing the file sizes to what I see in File Explorer it looks like it matches.

tsvietOK:
@yair100 Check files larger then 3GB.

yaichenbaum:
Does this work to detect the size of folders as well?

tsvietOK:
@yair100 No, it is just file size algorithm modification.

yair100 (approved):

yaichenbaum:
:shipit:

PR #717, opened 2020-04-28
Locked the search button since it is not yet implemented

yaichenbaum:


lampenlampen:
I would also disable the search bar in the settings

lampenlampen:
Maybe we should consider to hide features, that are not working, for a better user experience for the official launch.

duke7553:
I plan on starting my work for Search around May 12, but if it isn't satisfactory, then it will live behind a flag on Launch until it's ready.

But yes, generally any feature not ready needs to be disabled or behind a flag.

yaichenbaum:
@duke7553 I would hesitate to add a flag that would imply that some functionality will work if enabled until there is actually something to enable, so I think it should be disabled until then.

lampenlampen:
I would create a code flag for these features and if suitable we can expose those in the experimental settings page, when ready for testing.

PR #729, opened 2020-04-30
Fix file size calculation for var itemSizeBytes

tsvietOK:
This var hasn't been changed in PR #708

### DIFF
--- a/Files/View Models/ItemViewModel.cs
+++ b/Files/View Models/ItemViewModel.cs
@@ -848,28 +848,27 @@ namespace Files.Filesystem
                 systemTimeOutput.Second,
                 systemTimeOutput.Milliseconds);
             long fDataFSize = findData.nFileSizeLow;
-            long fileSize;
+            long itemSizeBytes;
             if (fDataFSize < 0 && findData.nFileSizeHigh > 0)
             {
-                fileSize = fDataFSize + 4294967296 + (findData.nFileSizeHigh * 4294967296);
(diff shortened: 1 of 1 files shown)


Files/View Models/ItemViewModel.cs
```diff
@@ -848,28 +848,27 @@ private void AddFile(WIN32_FIND_DATA findData, string pathRoot)
                 systemTimeOutput.Second,
                 systemTimeOutput.Milliseconds);
             long fDataFSize = findData.nFileSizeLow;
-            long fileSize;
+            long itemSizeBytes;
             if (fDataFSize < 0 && findData.nFileSizeHigh > 0)
             {
-                fileSize = fDataFSize + 4294967296 + (findData.nFileSizeHigh * 4294967296);
+                itemSizeBytes = fDataFSize + 4294967296 + (findData.nFileSizeHigh * 4294967296);
```
duke7553, line 854: What is the significance of the 4294967296 value applied here? Shouldn't we initially store this value as a constant?

tsvietOK: Of course we can use this value as constant. The main algorithm taken from [here](http://www.freevbcode.com/ShowCode.asp?ID=6836)

duke7553: Good. Let me know when you've made that change

tsvietOK: @duke7553 Done.

tsvietOK:
May be in the future we should do global code align, i mean move all const to top of file, check file names for casing, and check fields, constructors and methods order according to code convention

PR #731, opened 2020-05-01
Display already loaded entries while loading files in a directory

hez2010:
- Fix a potential crashing at startup due to null reference
- Display already loaded entries while loading files in a directory (akas. Display As You Go), which got user experience improved a lot
- Sorting entries won't take current position to the top of the list


hez2010:
CI failure is irrelevant with this PR.

yaichenbaum:
@hez2010 I sorted out the issue with the pipeline, I would like to get a review from @duke7553 before merging but from my end everything looks good.

PR #768, opened 2020-05-10
Improved "last tab closing" behavior

gave92:
This pull request changes the way the app is closed when the last tab is closed either with Ctrl+W or tab close button.

In the current implementation the ```Application.Current.Exit()``` method is called which results in an "ugly" closing animation: the app briefly shows an empty window and than the window closes abruptly looking like the app has crashed.

Instead this pull request uses the ```StartTerminateAsync()``` of the [diagnostic API](https://docs.microsoft.com/en-us/uwp/api/windows.system.appresourcegroupinfo.startterminateasync?view=winrt-18362) which results in the standard closing animation (like clicking the exit button on the titlebar) The method exists from the 1809 version and does not require additional app cababilities.

I do find this solution a bit of an overkill but I couldn't find a simpler way 😅

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -439,11 +439,7 @@ namespace Files.Interacts
         {
             if (((Window.Current.Content as Frame).Content as InstanceTabsView).TabStrip.TabItems.Count == 1)
             {
-                IList<AppDiagnosticInfo> infos = await AppDiagnosticInfo.RequestInfoForAppAsync();
-                IList<AppResourceGroupInfo> resourceInfos = infos[0].GetResourceGroups();
-                var pid = Windows.System.Diagnostics.ProcessDiagnosticInfo.GetForCurrentProcess().ProcessId;
-                await resourceInfos.Single(r => r.GetProcessDiagnosticInfos()[0].ProcessId == pid).StartTerminateAsync();
-                //Application.Current.Exit();
+                await InstanceTabsView.StartTerminateAsync();
             }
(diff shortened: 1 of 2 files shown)


yaichenbaum:
What is this supposed to fix?

gave92:
Fix? Nothing :)
It simply improves the app closing behavior: instead of abruptly closing the window it shows the default closing animation

yaichenbaum:
I am not seeing a problem with how it closes, would you be able to show a before and after video?

gave92:
These are the two behaviors I'm getting.
Current behavior (window flashes blue, than closes):
closing_current
With this pull request (window shows closing animation):
closing_new

xPoppyx:
Currently, if I close the last tab and reopen the app, the window is smaller, as if it had been crashed

gave92:
Unfortunately ```CoreApplication.Exit()``` behaves the same as ```Application.Current.Exit()```.

gave92:
No new Capability is needed in this case. The appDiagnostics capability is only needed if you use the Diagnostics API on an app different from the current one. From the (docs)[https://docs.microsoft.com/en-us/uwp/api/windows.system.appdiagnosticinfo.requestinfoforappasync?view=winrt-18362#Windows_System_AppDiagnosticInfo_RequestInfoForAppAsync] `When an app calls this method, it is calling about itself and there is no capability check.`

yaichenbaum:
@ADeltaX I imagine you would have something to say on this 🙂

yaichenbaum:
@duke7553 This would be great to put in a "command" files where we can reuse the code and just call the command when we want to use it.

PR #770, opened 2020-05-10
Add support for uppercase image extension

tsvietOK:
Like .JPG, .PNG etc.
Fixed a crash when you right clicked a file without extension.

### DIFF
--- a/Files/View Models/InteractionViewModel.cs
+++ b/Files/View Models/InteractionViewModel.cs
@@ -53,7 +53,8 @@ namespace Files.Controls
             {
                 string ItemExtension = (App.CurrentInstance.ContentPage as BaseLayout).SelectedItem.FileExtension;
 
-                if (ItemExtension.ToLower() == ".png" 
+                if (!string.IsNullOrEmpty(ItemExtension) 
+                    && ItemExtension.ToLower() == ".png" 
                     || ItemExtension.ToLower() == ".jpg" 
                     || ItemExtension.ToLower() == ".bmp" 
                     || ItemExtension.ToLower() == ".jpeg")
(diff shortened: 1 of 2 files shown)


Files/View Models/InteractionViewModel.cs
```diff
@@ -53,7 +53,10 @@ public void CheckForImage()
             {
                 string ItemExtension = (App.CurrentInstance.ContentPage as BaseLayout).SelectedItem.FileExtension;
 
-                if (ItemExtension == ".png" || ItemExtension == ".jpg" || ItemExtension == ".bmp" || ItemExtension == ".jpeg")
+                if (ItemExtension.ToLower() == ".png"
```
yaichenbaum, line 56: When I tried doing this the application would crash when right clicking on an image file, have you confirmed that this works?

tsvietOK: It is string, ToLower() can't crash app. From my side it is working fine.

duke7553: I would prefer using ItemExtension.Equals(".png", StringComparison.OrdinalIgnoreCase) for this one.

duke7553: Just to keep things consistent.

tsvietOK: Done

yaichenbaum: @tsvietOK Can you test that the functionality still works after doing the changes @duke7553 requested?

tsvietOK: Yeah, i will test it little bit later

tsvietOK: @yair100 It is working

PR #778, opened 2020-05-12
Fix null detection of file extension

tsvietOK:
Remove try-catch block
The problem has not been solved in PR #770 because of try-catch block which is "handle" exception caused app not crashed if `ItemExtension==null`.

### DIFF
--- a/Files/View Models/InteractionViewModel.cs
+++ b/Files/View Models/InteractionViewModel.cs
@@ -48,27 +48,23 @@ namespace Files.Controls
 
         public void CheckForImage()
         {
-            //check if the selected item is an image file
-            try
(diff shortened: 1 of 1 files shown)


Files/View Models/InteractionViewModel.cs
```diff
@@ -48,27 +48,23 @@ public bool IsPageTypeNotHome
 
         public void CheckForImage()
         {
-            //check if the selected item is an image file
```
yaichenbaum, line 51: Can you keep the code commented?

tsvietOK: Done

PR #780, opened 2020-05-13
Fixed file rename with space added opening QuickLook

Dylan-Osborne:
Fixed a bug that caused QuickLook to open when pressing spacebar while renaming a file.

Before:
before
After:
after

**To Reproduce In Current Version**
1. Make sure you have QuickLook installed.
2. Right click on a file and choose "Rename"
3. Give the file any name with a space in it. As soon as spacebar is pressed, QuickLook opens, the file name textbox loses focus and the file name is committed without allowing the user to finish typing.

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -1148,16 +1148,11 @@ namespace Files.Interacts
         {
             try
             {
-                string selectedItemPath = null;
                 int selectedItemCount;
                 Type sourcePageType = App.CurrentInstance.CurrentPageType;
                 selectedItemCount = (CurrentInstance.ContentPage as BaseLayout).SelectedItems.Count;
-                if (selectedItemCount == 1)
-                {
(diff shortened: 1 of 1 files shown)


Files/Interacts/Interaction.cs
```diff
@@ -1148,16 +1148,11 @@ public async void ToggleQuickLook()
         {
             try
             {
-                string selectedItemPath = null;
                 int selectedItemCount;
                 Type sourcePageType = App.CurrentInstance.CurrentPageType;
                 selectedItemCount = (CurrentInstance.ContentPage as BaseLayout).SelectedItems.Count;
-                if (selectedItemCount == 1)
```
yaichenbaum, line 1155: @duke7553 Will removing this cause any side effects?

duke7553: Not that I can think of.

yaichenbaum:
@Dylan-Osborne Thank you for your pull request, I just merged it into the `develop` branch. Your changes should make their way into the next update.

Issue #752, opened 2020-05-07
Deselect with Ctrl + click


xPoppyx:
![crash](https://user-images.githubusercontent.com/61259627/81246803-84b62100-8fe6-11ea-97e6-83aa9ede2783.gif)

duke7553:
Thanks for your feedback. We're having trouble understanding what you're requesting and need additional information to investigate this item. Please re-open when you've filled out the issue template to the best of your ability.

yaichenbaum:
I was able to reproduce this issue. If you ctrl + click on something it should select or deselect it.

xPoppyx:
@duke7553 you can select multiple files with Ctrl + click, but you should also be able to deselect
![deselect](https://user-images.githubusercontent.com/61259627/81882382-9d7f8300-9560-11ea-9967-42d015440755.gif)

yaichenbaum:
This issue has been fixed with #788

duke7553:
Thank you for creating this feedback item. A solution for it is currently being reviewed as a PR (pull request) to our active development branch. We'll let you know once it's merged.

duke7553:
Thank you for creating this feedback item. This issue is currently resolved in our active development branch and will be shipped soon once we've validated that everything works properly. Please subscribe to our releases to be notified when this is available.

Issue #673, opened 2020-04-24
Allow dragging files and folders without selecting them first


XeZrunner:
**Is your feature request related to a problem? Please describe.**
I wouldn't say it's a problem, so I report it as a feature request. Some may consider this annoying or a bug.

**Describe the solution you'd like**
When clicking and holding (or tapping and holding) without selecting the file first, the application should highlight and initiate the drag & drop mechanism.

**Describe alternatives you've considered**
This behavior could be a preference option in the Settings.
Perhaps the highlighting isn't necessary, but Windows normally handles drag & drop this way.

**Additional context**
None.

yaichenbaum:
@XeZrunner Good feedback! This would make drag and drop more user friendly.

xPoppyx:
I want to add something, if you select a folder or file and deselect it, you can't open it

yaichenbaum:
@xPoppyx That sounds more like a bug, can you open a separate issue for it?

yaichenbaum:
@XeZrunner This issue should be resolved with #4626.

PR #788, opened 2020-05-15
Added support for ctrl + click deselection in grid view.

Dylan-Osborne:
Added support for ctrl + click deselection. (#752)

Also changed the behavior of a left click without ctrl to match that of the list view i.e. If you have a few selected items and click one of them without holding ctrl, then that item is selected and the others are deselected. (The end of the below gif is an example of this)

Here's a demo:
GridViewDeselect


xPoppyx:
What about #673 ? can you fix it too? , is another issue about selection

yaichenbaum:
@xPoppyx That issue is unrelated and will probably be a little trickier to fix.

PR #789, opened 2020-05-15
v0.8.1

yaichenbaum:
- Fixed context menu appearing when right clicking on empty space in List view mode
- Added option to set an image as lockscreen background
- Parse command line before App launch
- Added support for uppercase image extension
- Fixed file rename with space added opening QuickLook
- Upgrade WinUI to v2.4
- Fixed grid view selection bugs
- Added support for ctrl + click deselection in grid view


tsvietOK:
Looks like this path caused new app crash when navigating to folder
Error

yaichenbaum:
@tsvietOK What are the steps to reproduce this crash?

tsvietOK:
I can't say exactly, but for me it crashing when i continuously opening folders

yaichenbaum:
@duke7553 Can it be related to the watcher for file changes?

PR #813, opened 2020-05-20
Fix blurred disk drive icons

tsvietOK:
Removed unused vars
Optimized code
Без имени-1 (2)
Attention to drive icons an their size
(Red line is center of column, just for example)

### DIFF
--- a/Files/Filesystem/DriveItem.cs
+++ b/Files/Filesystem/DriveItem.cs
@@ -8,31 +8,36 @@ namespace Files.Filesystem
     public class DriveItem : INavigationControlItem
     {
         public string Glyph { get; set; }
+        public string Text { get; set; }
+        public string Path => Tag;
+        public NavigationControlItemType ItemType { get; set; } = NavigationControlItemType.Drive;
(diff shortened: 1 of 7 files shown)


Files/Filesystem/DriveItem.cs
```diff
@@ -8,31 +8,36 @@ namespace Files.Filesystem
     public class DriveItem : INavigationControlItem
     {
         public string Glyph { get; set; }
+        public string Text { get; set; }
+        public string Path => Tag;
+        public NavigationControlItemType ItemType { get; set; } = NavigationControlItemType.Drive;
         public ulong MaxSpace { get; set; } = 0;
         public ulong SpaceUsed { get; set; } = 0;
-        public string DriveText { get; set; }
-        public string Tag { get; set; }
```
yaichenbaum, line 14: @duke7553 Don't we need the tag for translations purposes?

tsvietOK: @yair100 look at line 17

Issue #773, opened 2020-05-11
Time zone bug


Jaiganeshkumaran:
Windows internally stores timestamps in UTC, however the File Explorer converts it to the current timezone to avoid confusion. Files UWP is not aware of this and shows wrong time for last modfied and created time.

duke7553:
@Jaiganeshkumaran Thank you for opening this feedback item.

Can you please elaborate on why this could be causing the display of negative timestamps? (e.g. -51 seconds)

Jaiganeshkumaran:
I think that this could be because of the calculation. Windows stores timestamps in UTC, not local time.
However Files UWP calculates relative time using DateTime.Now which returns the local time, and that may cause the display of negative timestamps if the time zone is behind UTC.

duke7553:
Thank you for creating this feedback item. This issue is currently resolved in our active development branch and will be shipped soon once we've validated that everything works properly. Please subscribe to our releases to be notified when this is available.

PR #831, opened 2020-05-24
Fix files last modified date

gave92:
**Issue**
Currently Files UWP does not display files last modified date correctly: on my system where the timezone is UTC+2 a newly created file shows "Modified: 2 hours ago" (both in properties dialog and in "Date modified" column).
This pull request fixes this so it shows "Modified: 0 seconds ago".

**Fix**
The issue is due to the fact that the [`FileTimeToSystemTime`](https://docs.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-filetimetosystemtime) function returns an UTC time, but then a new DateTime object is constructed from the returned value without specifying the DateTimeKind as UTC.
```cs
var itemDate = new DateTime(
                systemTimeOutput.Year,
                systemTimeOutput.Month,
                systemTimeOutput.Day,
                systemTimeOutput.Hour,
                systemTimeOutput.Minute,
                systemTimeOutput.Second,
                systemTimeOutput.Milliseconds,
                DateTimeKind.Utc); -----------------> this is needed
```

### DIFF
--- a/Files/View Models/ItemViewModel.cs
+++ b/Files/View Models/ItemViewModel.cs
@@ -852,7 +852,8 @@ namespace Files.Filesystem
                     systemTimeOutput.Hour,
                     systemTimeOutput.Minute,
                     systemTimeOutput.Second,
-                    systemTimeOutput.Milliseconds);
+                    systemTimeOutput.Milliseconds,
+                    DateTimeKind.Utc);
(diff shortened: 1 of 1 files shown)


gave92:
Thanks, this may fix #773

Ein-Tim:
Seems to fix only partly: #2197

PR #833, opened 2020-05-25
Fixed grid view height for tall icons

Dylan-Osborne:
Fixed a bug that caused tall icons to go over the borders and in turn hide the file name.
Thanks @xPoppyx for noticing this👍
Before:
Before
After:
After


yaichenbaum:
:shipit:

PR #840, opened 2020-05-25
Add support for Recycle Bin

gave92:
Hello this pull request adds support for browsing the **Recycle Bin**!
"Support" here means that the recycle bin can be browsed just like any other folder in Files UWP, it's not just a "link" that opens the recycle bin in explorer.

image

#### How it works
As I'm not aware of any way to properly access the recycle bin folder from the UWP app, I'm making use of Files UWP full trust process. The fulltrust process enumerates the contents of the recycle bin using the Shell32 api and send the results to the UWP app. Communication between the fulltrust process and the UWP app is done through AppService.

*NB: This requires to change the fulltrust process lifecyle*: instead of starting a new instance every time to handle the commands to open CMD, toggle QuickLook, etc., a single instance keeps running until the UWP app is closed or suspended.

#### What works
- New entry on the sidebar with localized name of the recycle bin
- Enumerating recycle bin elements
- Monitor filesystem changes and refresh items
- Showing the thumbnails for deleted files
- Properties dialog for deleted files
- Deleting files from the recycle bin

#### What's missing
- A button to empty recycle bin
- A customized context menu with the proper options (e.g "restore")

Let me know what you think of this, if you approve of the "architectural" changes I can keep working to complete what's missing.

### DIFF
--- a/Common/ShellFileItem.cs
+++ b/Common/ShellFileItem.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Files.Common
+{
+    public class ShellFileItem
+    {
+        public bool isFolder;
+        public string recyclePath;
+        public string fileName;
+        public string filePath;
+        public DateTime recycleDate;
+        public string fileSize;
+        public int fileSizeBytes;
+        public string fileType;
+
+        public ShellFileItem()
+        {
+
+        }
+
+        public ShellFileItem(bool isFolder, string recyclePath, string fileName, string filePath, DateTime recycleDate, string fileSize, int fileSizeBytes, string fileType)
+        {
+            this.isFolder = isFolder;
+            this.recyclePath = recyclePath;
+            this.fileName = fileName;
+            this.filePath = filePath;
+            this.recycleDate = recycleDate;
+            this.fileSize = fileSize;
+            this.fileSizeBytes = fileSizeBytes;
+            this.fileType = fileType;
+        }
+    }
+}
--- a/Common/ShellFileItem.cs
+++ b/Common/ShellFileItem.cs
@@ -15,7 +15,7 @@ namespace Files.Common
         public string FilePath;
         public DateTime RecycleDate;
         public string FileSize;
-        public long FileSizeBytes;
+        public ulong FileSizeBytes;
         public string FileType;
 
         public ShellFileItem()
@@ -23,7 +23,7 @@ namespace Files.Common
 
         }
 
-        public ShellFileItem(bool isFolder, string recyclePath, string fileName, string filePath, DateTime recycleDate, string fileSize, long fileSizeBytes, string fileType)
+        public ShellFileItem(bool isFolder, string recyclePath, string fileName, string filePath, DateTime recycleDate, string fileSize, ulong fileSizeBytes, string fileType)
         {
             this.IsFolder = isFolder;
             this.RecyclePath = recyclePath;
--- a/Common/ShellFileItem.cs
+++ b/Common/ShellFileItem.cs
@@ -9,30 +9,30 @@ namespace Files.Common
 {
     public class ShellFileItem
     {
-        public bool isFolder;
-        public string recyclePath;
-        public string fileName;
-        public string filePath;
-        public DateTime recycleDate;
-        public string fileSize;
-        public int fileSizeBytes;
-        public string fileType;
+        public bool IsFolder;
+        public string RecyclePath;
+        public string FileName;
+        public string FilePath;
+        public DateTime RecycleDate;
+        public string FileSize;
+        public long FileSizeBytes;
+        public string FileType;
 
         public ShellFileItem()
         {
 
         }
 
-        public ShellFileItem(bool isFolder, string recyclePath, string fileName, string filePath, DateTime recycleDate, string fileSize, int fileSizeBytes, string fileType)
+        public ShellFileItem(bool isFolder, string recyclePath, string fileName, string filePath, DateTime recycleDate, string fileSize, long fileSizeBytes, string fileType)
         {
-            this.isFolder = isFolder;
-            this.recyclePath = recyclePath;
-            this.fileName = fileName;
-            this.filePath = filePath;
-            this.recycleDate = recycleDate;
-            this.fileSize = fileSize;
-            this.fileSizeBytes = fileSizeBytes;
-            this.fileType = fileType;
+            this.IsFolder = isFolder;
+            this.RecyclePath = recyclePath;
+            this.FileName = fileName;
+            this.FilePath = filePath;
+            this.RecycleDate = recycleDate;
+            this.FileSize = fileSize;
+            this.FileSizeBytes = fileSizeBytes;
+            this.FileType = fileType;
         }
     }
 }
--- a/Files.Launcher/Program.cs
+++ b/Files.Launcher/Program.cs
@@ -207,17 +207,18 @@ namespace FilesFullTrust
                             {
                                 try
                                 {
-                                    //folderItem.Properties.ReadOnly = true;
-                                    //folderItem.Properties.NoInheritedProperties = false;
+                                    folderItem.Properties.ReadOnly = true;
+                                    folderItem.Properties.NoInheritedProperties = false;
                                     string recyclePath = folderItem.FileSystemPath; // True path on disk
-                                    //string fileName = Path.GetFileName(folderItem.FileInfo.DisplayName); // Original file name
-                                    //string filePath = folderItem.FileInfo.DisplayName; // Original file path
-                                    //DateTime recycleDate = folderItem.FileInfo.CreationTime; // This is LocalTime
-                                    //string fileSize = folderItem.Properties.GetPropertyString(Vanara.PInvoke.Ole32.PROPERTYKEY.System.Size);
-                                    //long fileSizeBytes = folderItem.IsFolder ? 0 : folderItem.FileInfo.Length;
-                                    //string fileType = folderItem.FileInfo.TypeName;
-                                    //bool isFolder = folderItem.FileInfo.Attributes.HasFlag(FileAttributes.Directory); //folderItem.IsFolder includes .zip
-                                    folderContentsList.Add(new ShellFileItem(folderItem.IsFolder && Path.GetExtension(folderItem.Name) != ".zip", recyclePath, Path.GetFileName(folderItem.Name), Path.GetDirectoryName(folderItem.Name), DateTime.Now, "0 Kb", 0, "Some stuff"));
+                                    string fileName = Path.GetFileName(folderItem.Name); // Original file name
+                                    string filePath = folderItem.Name; // Original file path + name
+                                    var dt = (System.Runtime.InteropServices.ComTypes.FILETIME)folderItem.Properties[Vanara.PInvoke.Ole32.PROPERTYKEY.System.DateCreated];
+                                    var recycleDate = dt.ToDateTime().ToLocalTime(); // This is LocalTime
+                                    string fileSize = folderItem.Properties.GetPropertyString(Vanara.PInvoke.Ole32.PROPERTYKEY.System.Size);
+                                    ulong fileSizeBytes = folderItem.IsFolder ? 0 : (ulong)folderItem.Properties[Vanara.PInvoke.Ole32.PROPERTYKEY.System.Size];
+                                    string fileType = (string)folderItem.Properties[Vanara.PInvoke.Ole32.PROPERTYKEY.System.ItemTypeText];
+                                    bool isFolder = folderItem.IsFolder && Path.GetExtension(folderItem.Name) != ".zip";
+                                    folderContentsList.Add(new ShellFileItem(isFolder, recyclePath, fileName, filePath, recycleDate, fileSize, fileSizeBytes, fileType));
                                 }
                                 catch (System.IO.FileNotFoundException)
                                 {
--- a/Files.Launcher/Program.cs
+++ b/Files.Launcher/Program.cs
@@ -16,6 +16,8 @@ namespace FilesFullTrust
 {
     internal class Program
     {
+        private static readonly NLog.Logger Logger = NLog.LogManager.GetCurrentClassLogger();
+
         //Put all the variables required for the DLLImports here
         enum RecycleFlags : uint { SHERB_NOCONFIRMATION = 0x00000001, SHERB_NOPROGRESSUI = 0x00000002, SHERB_NOSOUND = 0x00000004 }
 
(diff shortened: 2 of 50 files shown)


duke7553:
@gave92 Have you tried accessing the recycling bin by it's GUID with the existing method? I'd assume it works since we currently use the Win32 FindFirstFileFromApp in the Filesystem view model.

Recycle Bin GUID - `::{645FF040-5081-101B-9F08-00AA002F954E}`

yaichenbaum:
Support for recycle bin would be great, I am not sure if we want to keep it pinned by default. Can you add a feature flag that turns on/off this feature?

gave92:
@yair100 yup I can do that, perhaps also add it to the New Tab page? Anyway is not the UI part that worries me :)

@duke7553 I've tried using the Shell32 api from the UWP side (meaning using the function [SHGetFolderLocation](https://docs.microsoft.com/it-it/windows/win32/api/shlobj_core/nf-shlobj_core-shgetfolderlocation) and other [IShellFolder](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-ishellfolder) related methods and for some reason I cannot enumerate the contents of the recycle bin. The same code works when executed in the fulltrust process.
Is this what you were suggesting by "accessing the recycling bin by it's GUID"?

yaichenbaum:
@gave92 I would add it to the new page, but it should be behind the feature flag as well.

gave92:
@yair100 I'll do that. I'm gonna wait a bit though, until I get some feedback on proposed method for accessing the recycle bin (fulltrust process + appservice).
Perhaps there's a simpler way of doing that that I'm missing.

tsvietOK:
@gave92 
First
Repro steps:
1. Open recycle bin
2. Navigate to folder inside it
3. Press Up button on navigation panel
4. Permission dialog displayed 
Second
I think folders in recycle bin should not have ability to be opened, renamed, pinned to sidebar, shared, opened in new windows/tab.

tsvietOK:
And last, i don't think it is a good idea to merge this PR without supporting restore and clear recycle bin functions.

tsvietOK:
Other way, mark this feature as Experimental

yaichenbaum:
> And last, i don't think it is a good idea to merge this PR without supporting restore and clear recycle bin functions.

We can do this behind a feature flag, this way we are not making any commitment to the feature and the capabilities can be changed.

Files/Program.cs
```diff
@@ -43,10 +41,14 @@ static void Main()
         public static async Task OpenShellCommandInExplorer(string shellCommand, int pid)
         {
             System.Diagnostics.Debug.WriteLine("Launching shell command in FullTrustProcess");
-            ApplicationData.Current.LocalSettings.Values["ShellCommand"] = shellCommand;
-            ApplicationData.Current.LocalSettings.Values["Arguments"] = "ShellCommand";
-            ApplicationData.Current.LocalSettings.Values["pid"] = pid;
-            await FullTrustProcessLauncher.LaunchFullTrustProcessForCurrentAppAsync();
+            if (App.Connection != null)
```
ADeltaX, line 44: App.Connection here is always null because it checks immediately at launch (hence the app will not work as shell replacement).

gave92:
Thanks for the feedbacks.

@tsvietOK
1. Permission dialog displayed => will be fixed when disallowing navigation inside recycle bin

2. I think folders in recycle bin should not have ability to be opened, renamed, pinned to sidebar, shared, opened in new windows/tab => agreed, need to customize the context menu and buttons (e.g add an "Empty recycle bin" button and a "Restore" to context menu)

3. "Recycle Bin" tab text untranslated => solved

4. Button "Copy path" causes app crash => I cannot reproduce this, but it will go away when we change the buttons/context menu (see 2)

Files/Program.cs
gave92: Thanks, I reverted the OpenShellCommandInExplorer function to the original. There's no need to set up the AppService stuff when we're just going to kill the process

Files.Launcher/Program.cs
```diff
@@ -1,85 +1,319 @@
...
+                        }
+                        else if (action == "Enumerate")
+                        {
+                            // Enumerate recyclebin contents and send response to UWP
+                            var response_enum = new ValueSet();
+                            Shell32.FolderItems items = Recycler.Items();
+                            var file_list = new List<ShellFileItem>();
+                            for (int i = 0; i < items.Count; i++)
+                            {
+                                Shell32.FolderItem FI = items.Item(i);
```
tsvietOK, line 215: If this vars are inside loop, so how to use camelCase?

Files/Views/InstanceTabsView.xaml.cs
```diff
@@ -131,6 +134,11 @@ public async void AddNewTab(Type t, string path)
                     tabLocationHeader = ResourceController.GetTranslation("SidebarVideos");
                     fontIconSource.Glyph = "\xE8B2";
                 }
+                else if (path.StartsWith(App.AppSettings.RecycleBinPath, StringComparison.OrdinalIgnoreCase))
+                {
+                    tabLocationHeader = ApplicationData.Current.LocalSettings.Values.Get("RecycleBin_Title", "Recycle Bin");
```
tsvietOK, line 139: What about to use localSettings.Values?

Common/ShellFileItem.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Files.Common
+{
+    public class ShellFileItem
+    {
+        public bool isFolder;
```
tsvietOK, line 12: UpperCamelCase for public fields

tsvietOK (changes requested):

Common/ShellFileItem.cs
gave92: Ok fixed the naming convention of a few variables. I'll be more careful 😅

Files.Launcher/Program.cs
tsvietOK: You forget to change this virus names to camelCase. (Currently they are used CamelCase)

Files/Views/InstanceTabsView.xaml.cs
yaichenbaum: Shouldn't we use `App.AppSettings`?

gave92: Perhaps it would be better just to add the localized recycle bin name to the translation files?

Files/Program.cs
ADeltaX: The `return` is sufficient only for Debug mode. In release mode this does not close the window, hence why we need to kill the "zombie" process

yaichenbaum:
- By default, recycle bin should not be pinned to the sidebar.
- The app crashes when navigating to recycle bin.

gave92:
@yair100 Ok I'll change the default to not pinned.
I'm aware of the crash, I'm waiting for the owner of the [Vanara](https://github.com/dahall/Vanara) library I'm now referencing to publicly release the bug fixed version (I have it locally but it's not on nuget yet).

yaichenbaum:
@gave92 Is there a way I can test this out in the meantime?

gave92:
@yair100 Yes please try again with the latest commit.
Notes:
* auto refresh of recycle bin contents is disabled until the release of the updated Vanara library
* creating/pasting/... a file in the recycle bin crashes the app, I'm not fixing this as the solution will be to have a different context menu and buttons for the recycle bin

yaichenbaum:
@gave92 I can now open recycle bin but it appears to be empty.

gave92:
@yair100 Can you check whether while you are in the recycle bin folder (in Files UWP) there's a process called `FilesFullTrust.exe` running in Task Manager?

yaichenbaum:
image

yaichenbaum:
@gave92 I do not see `FilesFullTrust.exe` in Task Manager.

gave92:
Thanks for the feedback, that means the the process has crashed so you see no results on the UWP app. Please confirm whether the process is there when you start the app (before navigating to the reycle bin). Anyway I'll add some logging to understand the issue.

gave92:
Thanks for the feedback, the issue should now be fixed!

Files/View Models/ItemViewModel.cs
```diff
@@ -274,7 +277,8 @@ public ItemViewModel()
             //(App.CurrentInstance as ProHome).RibbonArea.RibbonViewModel.ShareItems.PropertyChanged += ShareItems_PropertyChanged;
             //(App.CurrentInstance as ProHome).RibbonArea.RibbonViewModel.LayoutItems.PropertyChanged += LayoutItems_PropertyChanged;
             //(App.CurrentInstance as ProHome).RibbonArea.RibbonViewModel.AlwaysPresentCommands.PropertyChanged += AlwaysPresentCommands_PropertyChanged;
-            _cancellationTokenSource = new CancellationTokenSource();
+            _addFilesCTS = new CancellationTokenSource();
+            _semaphoreCTS = new CancellationTokenSource();
```
duke7553, line 281: semaphore?

lukeblevins (approved):
With the guarantee that this feature is behind a flag, and otherwise invisible, I'll approve this now.

gave92:
@duke7553 The feature *is* behind a flag, but anyone can enable it in the settings. So perhaps it should be put under the Experimental setting page? And made disabled like the "Show hidden files and folders" setting?
Note: even if the recycle bin sidebar folder is not shown the modifications to the fulltrust process remain in effect. Is this acceptable? Are you planning to merge this is its current state?

duke7553:
@gave92 Thanks for clarifying this. As long as the recycle bin location is only visible when a flag inside the Experimental Settings page is toggled from its default state, we can merge this when you've finished work on it.

gave92:
A little update on this:
* Added customized context menu with "Empty recycle bin" action
* Added customized context menu for recycle bin items with "Restore" action
* Disabled menu bar "New file" action
* Disabled renaming of items in recycle bin
* Disabled navigation and opening of items in recycle bin
* The option to enable recycle bin support is now in the Experimental settings page

Let me know what you think!

yaichenbaum:
@duke7553 @gave92 I should have clarified the first time that when I said feature flag, I meant the experimental settings page.

yaichenbaum:
@gave92 The app doesn't crash but it still won't load any items I have in recycle bin.

gave92:
Curious. Please check the following:
* That no "FilesFullTrust.exe" process is running after you visit the bin from the uwp app
* In the application data path ("%localappdata%\packages\FilesUWPxxxxx\LocalState" there should be a "debug_fulltrust.txt" with the logged exception

yaichenbaum:
@gave92 FilesFullTrust.exe is running, I see the regular debug file but nothing for debug_fulltrust.exe.

gave92:
Ok so the fulltrust process is not crashing like before which is good... and now I'm confused.
* Does opening files from the uwp work? Does the command "Open in terminal" work?
* Would you be willing to run the app from this PR in debug mode in visual studio?

yaichenbaum:
I don't get a crash, it's just that the progress bar keeps loading in recycle bin and doesn't load any items that are there.

yaichenbaum:
@gave92 It finally loaded after a few minutes, I think we need to figure out a faster way to load recycle bin before shipping it.

gave92:
Ok just how much stuff is in your recycle bin? xD

I'll test with a lot of files as well. Thanks for finding out the issue.

yaichenbaum:
@gave92 It is much faster now.

Common/ShellFileItem.cs
```diff
@@ -0,0 +1,38 @@
...
+        public bool IsFolder;
+        public string RecyclePath;
+        public string FileName;
+        public string FilePath;
+        public DateTime RecycleDate;
+        public string FileSize;
+        public ulong FileSizeBytes;
+        public string FileType;
+
+        public ShellFileItem()
```
yaichenbaum, line 21: Is this necessary?

Common/ShellFileItem.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+
+        }
```
yaichenbaum, line 24: Is this necessary?

gave92: I think so. I'm using Json.Net to serialize/deserialize ShellFileItem objects and as far as I know Json.Net need a parameterless constructor to to this.

Files/BaseLayout.cs
```diff
@@ -189,11 +192,44 @@ private void UnloadMenuFlyoutItemByName(string nameToUnload)
...
+                (menuItem as MenuFlyoutItemBase).Visibility = Visibility.Collapsed;
         }
 
         public void RightClickContextMenu_Opening(object sender, object e)
         {
+            if (App.CurrentInstance.ViewModel.WorkingDirectory.StartsWith(App.AppSettings.RecycleBinPath))
+            {
+                (this.FindName("EmptyRecycleBin") as MenuFlyoutItemBase).Visibility = Visibility.Visible;
+                (this.FindName("OpenTerminal") as MenuFlyoutItemBase).IsEnabled = false;
+                UnloadMenuFlyoutItemByName("NewEmptySpace");
```
yaichenbaum, line 204: You will probably need to use the localized strings to find the right items here.

Files/BaseLayout.cs
```diff
@@ -256,6 +292,28 @@ public void RightClickContextMenu_Opening(object sender, object e)
...
+            // Remove everything except "Delete", "Cut", "Properties"
+            UnloadMenuFlyoutItemByName("UnzipItem");
+            UnloadMenuFlyoutItemByName("OpenItem");
+            UnloadMenuFlyoutItemByName("OpenInNewTab");
+            UnloadMenuFlyoutItemByName("OpenInNewWindowItem");
+            UnloadMenuFlyoutItemByName("ShareItem");
+            UnloadMenuFlyoutItemByName("CopyItem");
+            UnloadMenuFlyoutItemByName("RenameItem");
+            UnloadMenuFlyoutItemByName("SidebarPinItem");
+            UnloadMenuFlyoutItemByName("QuickLook");
```
yaichenbaum, line 306: Being that we are working on adding support for more menu items, it might be a better idea to unload everything through a loop rather than removing the known items, this way we can add more items without updating this list.

yair100 (approved):

Files/BaseLayout.cs
gave92: I'm finding elements based on the "x:Name" attribute. That does not change based on the locale.

Files/BaseLayout.cs
gave92: You mean first removing everything and than adding just what's needed for the recycle bin? Yup I can do that

yaichenbaum: Correct, instead of removing specific ones, remove everything and then add the one we need.

gave92: I did it a little differently, let me know what you think: I simply created a specialized context menu for recycle bin items so there is no need to Hide/Show menu items and there is no issue when more menu items are added to the "standard" menu.

yaichenbaum: Doing it as an individual context menu is a great idea.

lukeblevins (approved):
Looks great! Maybe in the future we can use this AppService to load the extended properties of all StorageItems from the Shell, rather than using the Windows Runtime APIs.

PR #851, opened 2020-05-26
Missing strings added to Files.ta

xPoppyx:


tsvietOK:
@xPoppyx Please, update Resources file too

yaichenbaum:
@tsvietOK I wonder if there is a way to improve the pipeline to automatically update the resource file before merging pull requests.

PR #860, opened 2020-05-27
Fix incorrect path while upward to root disk

hez2010:
Fix incorrect path while upward to root disk

### DIFF
--- a/Files/Views/InstanceTabsView.xaml.cs
+++ b/Files/Views/InstanceTabsView.xaml.cs
@@ -278,7 +278,7 @@ namespace Files
                         if (!remDriveNames.Contains(NormalizePath(currentPathForTabIcon)))
                         {
                             fontIconSource.Glyph = "\xEDA2";
-                            tabLocationHeader = NormalizePath(currentPathForTabIcon);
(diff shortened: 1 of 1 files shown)


tsvietOK:
@hez2010 Can you, please, specify repro steps, when i can get incorrect path while upward to root disk

xPoppyx:
Captura de pantalla (212)

tsvietOK:
This changes fixes a bug, but i think it is better to move it to NormalizePath method like
```csharp
public static string NormalizePath(string path)
        {
            if (path.StartsWith("\\\\"))
            {
                return path.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar).ToUpperInvariant();
            }
            else
            {
                if (!path.EndsWith(Path.DirectorySeparatorChar))
                {
                    path += Path.DirectorySeparatorChar;
                }

                return Path.GetFullPath(new Uri(path).LocalPath)
                .TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)
                .ToUpperInvariant();
            }
        }
```

hez2010:
@tsvietOK 
> This changes fixes a bug, but i think it is better to move it to NormalizePath method like

Done

PR #872, opened 2020-05-27
ToolTip added for selection and layout mode button

xPoppyx:
- Selection options
- Layout mode

**I'm not sure if they are the correct names for these buttons, let me know if it's ok**

### DIFF
--- a/Files/Strings/en-US/Resources.resw
+++ b/Files/Strings/en-US/Resources.resw
@@ -630,7 +630,7 @@
   <data name="StatusBarControlLayoutMode.ToolTipService.ToolTip" xml:space="preserve">
     <value>Layout mode</value>
   </data>
-  <data name="SelectionOptionsButton.ToolTipService.ToolTip" xml:space="preserve">
+  <data name="StatusBarControlSelectionOptions.ToolTipService.ToolTip" xml:space="preserve">
     <value>Selection options</value>
   </data>
 </root>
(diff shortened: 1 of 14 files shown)


Files/Strings/en-US/Resources.resw
```diff
@@ -627,4 +627,10 @@
   <data name="FolderNotFoundDialog.Title" xml:space="preserve">
     <value>Did you delete this folder?</value>
   </data>
+  <data name="LayoutModeButton.ToolTipService.ToolTip" xml:space="preserve">
```
yaichenbaum, line 630:
```suggestion
  <data name="StatusBarControlLayoutMode.ToolTipService.ToolTip" xml:space="preserve">
```

Files/Strings/en-US/Resources.resw
```diff
@@ -627,4 +627,10 @@
...
+    <value>Layout mode</value>
+  </data>
+  <data name="SelectionOptionsButton.ToolTipService.ToolTip" xml:space="preserve">
```
yaichenbaum, line 633:
```suggestion
  <data name="StatusBarControlSelectionOptions.ToolTipService.ToolTip" xml:space="preserve">
```

Files/UserControls/StatusBarControl.xaml
```diff
@@ -24,6 +24,7 @@
             HorizontalAlignment="Right"
             Orientation="Horizontal">
             <Button
+                x:Uid="SelectionOptionsButton"
```
yaichenbaum, line 27: It looks like this has the wrong name for the string.

Files/UserControls/StatusBarControl.xaml
```diff
@@ -58,6 +60,7 @@
             </Button>
 
             <Button
+                x:Uid="LayoutmodeButton"
```
yaichenbaum, line 63: It looks like this has the wrong name for the string.

PR #902, opened 2020-06-04
Added option to open a specific page on startup

yaichenbaum:


xPoppyx:
is it ok without a tooltip for the button?

yaichenbaum:
@xPoppyx That is something we would like to add, for the time being we want to get this out since the first official preview is launching next week.

PR #903, opened 2020-06-04
Auto-Detect and List Windows Terminal

yaichenbaum:


tsvietOK:
Did you try to use [FindPackageForUser](https://docs.microsoft.com/en-us/uwp/api/windows.management.deployment.packagemanager.findpackageforuser?view=winrt-19041)?

yaichenbaum:
@tsvietOK That is a restricted capability and the store didn't certify our app to use that permission.

tsvietOK:
I have found a bug.
Repro:
Revert this patch.
Copy this json to terminal.json
```json
{
  "version": 1,
  "terminals": [
    {
      "id": 1,
      "name": "CMD",
      "path": "cmd.exe",
      "arguments": "/k \"cd /d {0} && title Command Prompt\"",
      "icon": ""
    },
    {
      "id": 2,
      "name": "Windows PowerShell",
      "path": "powershell.exe",
      "arguments": "-WorkingDirectory \"{0}\"",
      "icon": ""
    },
    {
      "id": 3,
      "name": "Windows Terminal",
      "path": "wt.exe",
      "arguments": "-d \"{0}\"",
      "icon": ""
    }
  ]
}
```
Restart app
Set default terminal Windows Terminal.
Close app
Apply this patch.
Start app and try to open terminal in folder.
Crash.
That happen because of localSettings terminal_id store old terminal id value = 3. But current updated terminal.json contain only two items.
So i would suggest to save default terminal id right in terminal.json file.

PR #906, opened 2020-06-04
Fix pasting to recycle bin - Issue 899

### LINKED ISSUE #899: App doesn't show or crashes when moving an object to Recycle Bin
Today, I found the feature of Recycle Bin in the app and then went on to use it. With the Files app open on the side of my desktop, I dropped one of my shortcuts to the bin and when i checked into the recycle bin of the app, there was nothing but the file explorer was showing the deleted shortcut. Next, I cut a folder from the desktop through the app and then pasted it in the recycle bin of the app, the app got stuck for a few mins with high disk usage but then crashed unexpectedly 

I understand that it is an experimental feature. Thus, I didn't give a bad rating. I expect the problem to be fixed as soon as possible.

(The Microsoft Store Version app)
 - OS Version: [Windows 10 2004]
 - App version: [v0.9.1.0]

gave92:
Fixes #899

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -840,6 +840,19 @@ namespace Files.Interacts
 
         public async Task PasteItems(DataPackageView packageView, string destinationPath, DataPackageOperation acceptedOperation)
         {
+            if (!packageView.Contains(StandardDataFormats.StorageItems))
+            {
+                // Happens if you copy some text and then you Ctrl+V in FilesUWP
+                // Should this be done in ModernShellPage?
+                return;
+            }
+            if (CurrentInstance.ViewModel.WorkingDirectory.StartsWith(App.AppSettings.RecycleBinPath))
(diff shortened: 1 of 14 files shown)


Files/Interacts/Interaction.cs
```diff
@@ -842,6 +842,19 @@ public async void PasteItem_ClickAsync(object sender, RoutedEventArgs e)
 
         public async Task PasteItems(DataPackageView packageView, string destinationPath, DataPackageOperation acceptedOperation)
         {
+            if (!packageView.Contains(StandardDataFormats.StorageItems))
```
gave92, line 845: Note: this change is not specific to pasting to the recycle bin but can happen when pasting in any folder

duke7553: Good catch

yaichenbaum:
@gave92 Can you please solve the merge conflicts?

PR #926, opened 2020-06-06
[PROPOSAL] - Access android (MTP) devices

### LINKED ISSUE #779: App crashes when accessing android device's drive
**Describe the bug**
App crashes whenever I try to access the drive of my android phone

**To Reproduce**
1. Plug-in phone
2. Click on phone under drive section
3. App crashes

**Expected behavior**
Root folder of the phone should be displayed

**Desktop**
 - OS Version: [e.g. Windows 10 1909 18363.778]
 - App version: [v0.8.0.0]

**Additional context**
App even crashes even when trying to copy items from my phone to the my PC's local drive (C:).

gave92:
Might help fix #779

This PR contains a proposal of how MTP devices like android phones could be accessed by Files UWP.

What works:
* Navigation inside the MTP device
* Pasting/Deleting/Renaming inside the MTP device
* Creating new ~files~ folders inside the MTP device.
Files can only be copied to, explorer does the same
* Opening files inside the MTP device

@duke7553 what do you think?

### DIFF
--- a/Files/DataModels/SidebarPinnedModel.cs
+++ b/Files/DataModels/SidebarPinnedModel.cs
@@ -87,32 +87,14 @@ namespace Files.DataModels
             {
                 Debug.WriteLine(ex.Message);
             }
-            catch (ArgumentException ex)
+            catch (Exception ex) when (
+                ex is ArgumentException // Pinned item was invalid
+                || ex is FileNotFoundException // Pinned item was deleted
+                || ex is System.Runtime.InteropServices.COMException // Pinned item's drive was ejected
+                || (uint)ex.HResult == 0x800700A1) // The specified path is invalid (usually an mtp device was disconnected)
             {
                 Debug.WriteLine("Pinned item was invalid and will be removed from the file lines list soon: " + ex.Message);
                 RemoveItem(path);
-            }
-            catch (FileNotFoundException ex)
-            {
-                Debug.WriteLine("Pinned item was deleted and will be removed from the file lines list soon: " + ex.Message);
-                RemoveItem(path);
-            }
-            catch (System.Runtime.InteropServices.COMException ex)
-            {
-                Debug.WriteLine("Pinned item's drive was ejected and will be removed from the file lines list soon: " + ex.Message);
-                RemoveItem(path);
-            }
-            catch (Exception ex)
-            {
-                if ((uint)ex.HResult == 0x800700A1) // The specified path is invalid (usually an mtp device was disconnected)
-                {
-                    Debug.WriteLine("Pinned item was invalid and will be removed from the file lines list soon: " + ex.Message);
-                    RemoveItem(path);
-                }
-                else
-                {
-                    throw ex;
-                }
             }
         }
 
--- a/Files/DataModels/SidebarPinnedModel.cs
+++ b/Files/DataModels/SidebarPinnedModel.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Threading.Tasks;
 using Windows.Storage;
 
 namespace Files.DataModels
@@ -35,12 +36,12 @@ namespace Files.DataModels
             return Items;
         }
 
-        public void AddItem(string item)
+        public async void AddItem(string item)
         {
             if (!Items.Contains(item))
             {
                 Items.Add(item);
-                AddItemToSidebar(item);
+                await AddItemToSidebar(item);
                 Save();
             }
         }
@@ -65,12 +66,13 @@ namespace Files.DataModels
             }
         }
 
-        public async void AddItemToSidebar(string path)
+        public async Task AddItemToSidebar(string path)
         {
             try
             {
-                StorageFolder folder = await StorageFolder.GetFolderFromPathAsync(path);
-                int insertIndex = App.sideBarItems.IndexOf(App.sideBarItems.Last(x => x.ItemType == NavigationControlItemType.Location
+                var item = await DrivesManager.GetRootFromPath(path);
+                StorageFolder folder = await StorageFileExtensions.GetFolderFromPathAsync(path, item);
+                int insertIndex = App.sideBarItems.IndexOf(App.sideBarItems.Last(x => x.ItemType == NavigationControlItemType.Location 
                     && !x.Path.Equals(App.AppSettings.RecycleBinPath))) + 1;
                 var locationItem = new LocationItem
                 {
@@ -100,14 +102,26 @@ namespace Files.DataModels
                 Debug.WriteLine("Pinned item's drive was ejected and will be removed from the file lines list soon: " + ex.Message);
                 RemoveItem(path);
             }
+            catch (Exception ex)
+            {
+                if ((uint)ex.HResult == 0x800700A1) // The specified path is invalid (usually an mtp device was disconnected)
+                {
+                    Debug.WriteLine("Pinned item was invalid and will be removed from the file lines list soon: " + ex.Message);
+                    RemoveItem(path);
+                }
+                else
+                {
+                    throw ex;
+                }
+            }
         }
 
-        public void AddAllItemsToSidebar()
+        public async void AddAllItemsToSidebar()
         {
             for (int i = 0; i < Items.Count(); i++)
             {
                 string path = Items[i];
-                AddItemToSidebar(path);
+                await AddItemToSidebar(path);
             }
         }
 
--- a/Files/Filesystem/Drives.cs
+++ b/Files/Filesystem/Drives.cs
@@ -121,7 +121,7 @@ namespace Files.Filesystem
             }
 
             // If drive already in list, skip.
-            if (Drives.Any(x => x.Path == root.Name))
+            if (Drives.Any(x => string.IsNullOrEmpty(root.Path) ? x.Path.Contains(root.Name) : x.Path == root.Path))
             {
                 return;
             }
@@ -286,6 +286,51 @@ namespace Files.Filesystem
             {
                 list.Add(oneDriveItem);
             }
+        }
+
+        public static async Task<StorageFolderWithPath> GetRootFromPath(string devicePath)
+        {
+            if (!Path.IsPathRooted(devicePath))
+            {
+                return null;
+            }
+            var rootPath = Path.GetPathRoot(devicePath);
+            if (devicePath.StartsWith("\\\\?\\"))
+            {
+                // Check among already discovered drives
+                StorageFolder matchingDrive = App.AppSettings.DrivesManager.Drives.FirstOrDefault(x =>
+                    InstanceTabsView.NormalizePath(x.Path) == InstanceTabsView.NormalizePath(rootPath))?.Root;
+                if (matchingDrive == null)
+                {
+                    // Check on all removable drives
+                    var remDevices = await DeviceInformation.FindAllAsync(StorageDevice.GetDeviceSelector());
+                    foreach (var item in remDevices)
+                    {
+                        try
+                        {
+                            var root = StorageDevice.FromId(item.Id);
+                            if (InstanceTabsView.NormalizePath(rootPath).Replace("\\\\?\\", "") == root.Name.ToUpperInvariant())
+                            {
+                                matchingDrive = root;
+                                break;
+                            }
+                        }
+                        catch (Exception)
+                        {
+                            // Ignore this..
+                        }
+                    }
+                }
+                if (matchingDrive != null)
+                {
+                    return new StorageFolderWithPath(matchingDrive, rootPath);
+                }
+            }
+            else
+            {
+                return new StorageFolderWithPath(await StorageFolder.GetFolderFromPathAsync(rootPath), rootPath);
+            }
+            return null;
         }
 
         public void Dispose()
--- a/Files/Filesystem/Drives.cs
+++ b/Files/Filesystem/Drives.cs
@@ -326,10 +326,11 @@ namespace Files.Filesystem
                     return new StorageFolderWithPath(matchingDrive, rootPath);
                 }
             }
-            else
-            {
-                return new StorageFolderWithPath(await StorageFolder.GetFolderFromPathAsync(rootPath), rootPath);
-            }
+            // It's ok to return null here, on normal drives StorageFolder.GetFolderFromPathAsync works
+            //else
+            //{
+            //    return new StorageFolderWithPath(await StorageFolder.GetFolderFromPathAsync(rootPath), rootPath);
+            //}
             return null;
         }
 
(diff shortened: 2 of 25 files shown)


duke7553:
@gave92 I'm impressed by the amount of work you put into this, and will look forward to learning more about this change tomorrow

gave92:
**Description of code changes**
The main issue when accessing MTP devices is that windows does not map them to normal drives. Files inside such devices will not have a filesystem "path", which results in the impossibility of using functions such as `StorageFolder.GetFolderFromPathAsync()` and `FindFirstFileExFromApp`.

The only way to get the contents of a folder is querying the StorageFolder with the `folder.GetFilesAsync()`/`folder.GetFoldersAsync()` methods.
Note: getting the parent of a StorageItem *also* doesn't work (`storageItem.GetParentAsync()` returns null)

In order to navigate inside MTP devices it's necessary to:
1) Keep a reference to the root storage folder of the device -> returned by `StorageDevice.FromId()`
2) Keep a reference to the current storage folder
3) Still have a way to retrieve a specific storage folder from a path (necessary e.g when navigating up, when using the NavigationToolbar, etc)

For point (1) I added the property `Root` to the [INavigationControlItem](https://github.com/files-community/files-uwp/pull/92♠6/files#diff-3aa6f76018bffddc35fd94edc3cac47cR10) interface. The property will be populated for removable drives when creating a [DriveItem](https://github.com/files-community/files-uwp/pull/926/files#diff-31a5e6f57b2f992b1b395d718f8a4ddaR41).

For point (2) I added the `CurrentStorageFolder` property to [ItemViewModel](https://github.com/files-community/files-uwp/pull/926/files#diff-c66272bf7eb1f57b0287c4466394719bR43). The `WorkingDirectory` property now simply returns the `CurrentStorageFolder` path. As operations with storage folders are async, changing the `CurrentStorageFolder` is done in a [method](https://github.com/files-community/files-uwp/pull/926/files#diff-c66272bf7eb1f57b0287c4466394719bR54) and not in a property setter.

Point (3) is the real "fun". To get a storage folder from a path:
- I split the given path into components (the same way it's done for NavigationToolbar)
- For each path component, starting from the Root folder (or the CurrentStorageFolder), I drill down calling folder.GetFolderAsync(component) until I get a reference to the storage item I'm interested in.

This is done in [GetFolderFromRelativePathAsync](https://github.com/files-community/files-uwp/pull/926/files#diff-37400bd801d919a8ce26acb7a209f362R179). This behaves the same as `StorageFolder.GetFolderFromPathAsync` unless we are in a removable drive (where the Root property is not null).
This method returns an object that contains the retrieved StorageFolder and a `Path` that in case of removable drives will be a "virtual" path (= "root_name\folder1_name\folder_2_name...")

Most of the other code changes are simply replacing calls to `StorageFolder.GetFolderFromPathAsync` with `StorageFolder.GetFolderFromRelativePathAsync`

Pros:
- no changes for normal drives
- done in UWP (I discarded the idea of doing this in the fulltrust process like the recycle bin)

tsvietOK:
@gave92 Get an error when trying to access to my OnePlus 5T file system.
Translated exception: Access denied.
image

gave92:
Thanks for the feedback I only tested this on my phone.
Where inside the RefreshItemsAsync method is that exception thrown?

tsvietOK:
In EnumerateItemsFromStandardFolder (ItemViewModel.cs) in foreach:
image
As i understand, problem is in await _rootFolder.GetFilesAsync()

gave92:
@tsvietOK please let me know if latest commit fixes it thanks!

tsvietOK:
@gave92 Ok, now it is working, but much slower then windows explorer. 17sec to load a folder contains 416 elements. 
```
Enumerating items in \\?\ONEPLUS A5010\Внутренний общий накопитель\Music (device) completed in 17396 milliseconds.
```
Now i am getting an exception when trying to create bitmap
image
```
Exception thrown: 'System.UnauthorizedAccessException' in System.Private.CoreLib.dll
An exception of type 'System.UnauthorizedAccessException' occurred in System.Private.CoreLib.dll but was not handled in user code
Access denied. (Exception from HRESULT: 0x80070005 (E_ACCESSDENIED))
```

tsvietOK:
@gave92 Another bug, files are not displayed at all, only folders exist.

gave92:
Ah that's also why you are getting the exception for the bitmap

tsvietOK:
Ok, in this case Create file function should be disabled. 
File paste causing app crash. 
If start app when device already connected, app freezes at splash screen.

gave92:
I believe in your case the cause of file paste causing app crash is the same as crash when creating bitmap which is also caused by the UnauthorizedAccessException when enumerating files.
I'm not sure how to debug (and solve) this.
Ok, in this case Create file function should be disabled. -> yup I'll do that as soon as I know this thing doesn't work just on my phone 😂

tsvietOK:
Menu "Create file" in windows explorer is not exists for my phone too.
Version from store starting fine with device connected.

gave92:
If you are willing to debug the app and see why it freezes at splash screen it would be a huge help as I can't reproduce it.

tsvietOK:
Get another bug. App crash on startup when at least one element pinned to sidebar.
image

tsvietOK:
Update: App starts fine when device connected, app crash caused by issue above.

gave92:
Good catch should be fixed

yaichenbaum:
It looks like CodeFactor found some issues, those will need to be solved either way before merging the pull request, but perhaps they have something to do with the crash.

duke7553:
@gave92 Are we enumerating the storage items 
on the device using the UWP GetItemsAsync() method? Maybe it would be better to somehow get a HANDLE to the StorageFolder and pass it into a Win32 method? It would be a lot faster this was, but I'm not sure if it's possible to get a HANDLE from StorageFolder.

gave92:
@yair100 yeah but they seem to be purely "cosmetic" issue (like multiple classes in the same file..)

@duke7553 yes for mtp devices I'm using GetItemsAsync(UInt32, UInt32) which is slower (it's not so slow on folders so maybe the issue is getting the thumbnail).
I think there's a IStorageFolderHandleAccess method btw, I'll check it out.

gave92:
@duke7553 Turns out the issue causing very slow folder contents enumeration was getting the item thumbnain. It should be now much faster while still using `GetItemsAsync`.

Files/Views/ModernShellPage.xaml.cs
```diff
@@ -152,6 +152,11 @@ private void Page_Loaded(object sender, RoutedEventArgs e)
                         NavigationPath = NavParams;
                         SidebarControl.SelectedSidebarItem = AppSettings.DrivesManager.Drives.First(x => x.Path.ToString().Equals($"{NavParams[0]}:\\", StringComparison.OrdinalIgnoreCase));
                     }
+                    else if(NavParams.StartsWith("\\\\?\\"))
+                    {
+                        NavigationPath = NavParams;
+                        SidebarControl.SelectedSidebarItem = App.AppSettings.DrivesManager.Drives.First(x => x.Path.ToString().Equals($"{System.IO.Path.GetPathRoot(NavParams)}", StringComparison.OrdinalIgnoreCase));
```
yaichenbaum, line 158:
```suggestion
                        SidebarControl.SelectedSidebarItem = AppSettings.DrivesManager.Drives.First(x => x.Path.ToString().Equals($"{System.IO.Path.GetPathRoot(NavParams)}", StringComparison.OrdinalIgnoreCase));
```

gave92: Apologies this may revert some of the refactoring in #1119 
It's hard to keep this updated :)

yaichenbaum: @gave92 No worries, we are really excited about this pull request 🎉.

tsvietOK:
Getting an exception when navigating to my phone storage.
image

lukeblevins (approved):

gave92:
@tsvietOK thanks for the feedback. The exception should not be thrown on that line. It would be great if you could step through the code and get the line.
On a guess it may be [here](https://github.com/files-community/files-uwp/pull/926/files#diff-c66272bf7eb1f57b0287c4466394719bR112). If that's the case, is `_workingRoot` null at that point?

tsvietOK:
@gave92 The problem is in line `return new StorageFolderWithPath(await StorageFolder.GetFolderFromPathAsync(value));` in StorageFileExensions.cs
image
Exception appeared when trying to get storage folder
Value = "\\\\?\\ONEPLUS A5010\\"

gave92:
Is the `rootFolder` parameter null at that point?

tsvietOK:
> Is the `rootFolder` parameter of that function `null` when the exception happens?

Yes

Files/Filesystem/Drives.cs
```diff
@@ -288,6 +288,29 @@ private void GetVirtualDrivesList(IList<DriveItem> list)
             }
         }
 
+        public static async Task<StorageFolderWithPath> GetRootFromPath(string devicePath)
```
gave92, line 291: @tsvietOK could you put a breakpoint on this method (GetRootFromPath) and see why it returns `null`? (when you navigate to the phone storage)

tsvietOK: devicePath = "\\\\?\\ONEPLUS A5010\\"
image
matchingDrive == null, so this method returns null

gave92: Thanks a lot. It's not finding a device with matching name so it returns null. Previous method for getting the drive was more robust, guess I'll revert to that.

gave92: @tsvietOK issue should now be fixed, please let me know.

tsvietOK:
Issue fixed, good job! One small glitch: double separator
image

Files/DataModels/SidebarPinnedModel.cs
```diff
@@ -100,6 +101,18 @@ public async void AddItemToSidebar(string path)
                 Debug.WriteLine("Pinned item's drive was ejected and will be removed from the file lines list soon: " + ex.Message);
                 RemoveItem(path);
             }
+            catch (Exception ex)
+            {
+                if ((uint)ex.HResult == 0x800700A1)
```
tsvietOK, line 106: Can you explain?

Files/Interacts/Interaction.cs
```diff
@@ -433,22 +435,21 @@ private async void OpenSelectedItems(bool displayApplicationPicker)
...
                             {
                                 DisplayApplicationPicker = true
                             };
-                            await Launcher.LaunchFileAsync(file, options);
+                            await Launcher.LaunchFileAsync(childFile.File, options);
                         }
                     }
-                    else
+                    if(!displayApplicationPicker)
+
```
tsvietOK, line 452: New line

Files/Views/Pages/Properties.xaml.cs
```diff
@@ -239,12 +239,12 @@ private async void GetFolderProperties(CancellationTokenSource _tokenSource)
...
                 else
                 {
-                    storageFolder = await StorageFolder.GetFolderFromPathAsync(parentDirectory.ItemPath);
+                    storageFolder = await ItemViewModel.GetFolderFromPathAsync(parentDirectory.ItemPath);
                     ViewModel.ItemCreatedTimestamp = ListedItem.GetFriendlyDate(storageFolder.DateCreated);
                     GetOtherPropeties(storageFolder.Properties);
                     GetFolderSize(storageFolder, _tokenSource.Token);
                 }
-            }
+            }
```
tsvietOK, line 247: Trailing spaces.

Files/Interacts/Interaction.cs
```diff
@@ -1095,6 +1108,12 @@ public async Task PasteItems(DataPackageView packageView, string destinationPath
                         // File was moved/deleted in the meantime
                         continue;
                     }
+                    catch (Exception ex)
```
tsvietOK, line 1111: I think there is better way to detect an exception.

tsvietOK (changes requested):
And why all pinned(by me) items are placed before default pinned items like Music, Downloads etc.?

Files/Interacts/Interaction.cs
gave92: The thrown exception in case of resource in use is of type "System.Exception".

gave92:
@tsvietOK 
> One small glitch: double separator

Fixed
> And why all pinned(by me) items are placed before default pinned items like Music, Downloads etc.?

That's the magic of not awaiting async methods, should be fixed

Files/DataModels/SidebarPinnedModel.cs
```diff
@@ -100,14 +102,26 @@ public async void AddItemToSidebar(string path)
                 Debug.WriteLine("Pinned item's drive was ejected and will be removed from the file lines list soon: " + ex.Message);
                 RemoveItem(path);
             }
+            catch (Exception ex)
+            {
+                if ((uint)ex.HResult == 0x800700A1) // The specified path is invalid (usually an mtp device was disconnected)
```
tsvietOK, line 107: Could we simplify all exceptions into one?

Files/View Models/ItemViewModel.cs
```diff
@@ -52,46 +52,89 @@ public class ItemViewModel : INotifyPropertyChanged, IDisposable
...
+        private StorageFolderWithPath _workingRoot;
 
-            set
-            {
-                if (!string.IsNullOrWhiteSpace(value))
-                {
-                    _WorkingDirectory = value;
+        public async Task SetWorkingDirectory(string value)
+        {
+            if (string.IsNullOrWhiteSpace(value))
```
tsvietOK, line 69:
```csharp
 if (string.IsNullOrWhiteSpace(value))
{
        return;
}
```

tsvietOK (changes requested):
Please, resolve conflicts.

yair100 (approved):

tsvietOK:
@gave92 Seems like there is no option to create a folder inside the MTP device.

gave92:
@tsvietOK 
- Right click on empty space > New > Folder should work
- But the menu "+" button on the menu bar is incorrectly disabled
- On right click menu you can also click "New text file/New bitmap" but those do not work.

Do you confirm all this?

tsvietOK:
1. image
2. "+" button is disabled
3. There is no such menu items at all

tsvietOK:
@gave92 Update: Check in grid view mode

gave92:
@tsvietOK thanks I can confirm there is no New options in Grid View Mode

tsvietOK:
image
There is New Bitmap and New Text document menu item exists in context menu in GridViewBrwoser mode

gave92:
@tsvietOK hopefully it's now fixed

tsvietOK (approved):

yair100 (approved):

yaichenbaum:
@gave92 Great work! :shipit:

Issue #930, opened 2020-06-07
None of my drive shows up


tomyan112:
**Screenshots**
![image](https://user-images.githubusercontent.com/9064351/83962646-afd7ad00-a8d1-11ea-92b2-c7900e239e70.png)

**Desktop (please complete the following information):**
 - OS Version: [e.g. Windows 10 2004 19041.207]
 - App version: [v0.9.1]

Cyberdroid1:
Even I am facing this problem in v0.9.2.

gave92:
I've noticed this problem with removable drives. Sometimes they do not show up when connected/do not disappear when disconnected.
[Note for devs] Changing how you get the `Dispatcher` in `Drives.cs` seems to solve this.
=> `CoreApplication.MainView.Dispatcher` (current) often throws System.Exception
=> `CoreApplication.MainView.CoreWindow.Dispatcher` seems to work consistently
I'm not opening a PR for this as I don't really understand the difference 😅

tsvietOK:
@tomyan112 @Cyberdroid1 Can you give us more information about the disks to be displayed like model, type and connection interface?

gave92:
@tomyan112 also please post the "debug.txt" file you find in `%LocalAppData%\Packages\49306atecsolution.FilesUWP_et10x9a9vyk8t\LocalState`

tomyan112:
> @tomyan112 also please post the "debug.txt" file you find in `%LocalAppData%\Packages\49306atecsolution.FilesUWP_et10x9a9vyk8t\LocalState`

2020-03-15 18:00:12.8600|INFO|Files.Filesystem.DrivesManager|Drive added: E:\, Removable
2020-03-15 18:00:13.0940|INFO|Files.App|App launched
2020-03-15 18:00:56.5690|INFO|Files.Filesystem.DrivesManager|Drive added: H:\, Removable
2020-03-15 18:01:10.3390|INFO|Files.App|App launched
2020-03-15 18:01:11.0980|INFO|Files.Filesystem.DrivesManager|Drive added: E:\, Removable
2020-03-15 18:01:48.7065|INFO|Files.Filesystem.DrivesManager|Drive added: H:\, Removable
[...]

PR #937, opened 2020-06-07
Change how to get Dispatcher

gave92:
This PR:
- may fix the issue of drives not appearing on the sidebar #930 
- *fixes* the issue of removable drives not always appearing/disappearing from the sidebar when connected/disconnected

Basically it changes how to get the UI Dispatcher
- CoreApplication.MainView.Dispatcher -> this often results in `System.Exception` even when app has started
- CoreApplication.MainView.CoreWindow.Dispatcher -> seems ok

Can someone confirm this? (and explain me what the difference is? 😅)

This PR also makes sure that async methods inside a try-catch are always awaited as I don't think that exceptions are caught if the method is not awaited.

```cs
try
{
    StorageFolder.GetFolderFromPathAsync(App.AppSettings.OneDrivePath); ----> async without await
}
catch (Exception) ----> the exception can't be caught
{
    App.AppSettings.PinOneDriveToSideBar = false;
}
```

### DIFF
--- a/Files/Filesystem/Drives.cs
+++ b/Files/Filesystem/Drives.cs
@@ -18,7 +18,7 @@ namespace Files.Filesystem
     {
         private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
 
-        public ObservableCollection<DriveItem> Drives { get; } = new ObservableCollection<DriveItem>();
+        public IList<DriveItem> Drives { get; } = new List<DriveItem>();
         public bool ShowUserConsentOnInit { get; set; } = false;
         private DeviceWatcher _deviceWatcher;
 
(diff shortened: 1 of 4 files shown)


tsvietOK:
To properly test this changes i think we are need the way to reproduce this bug.

gave92:
I usually get it like this:
- I open 3-4 instances of the files app
- I connect my phone
- The phone will not appear in all instances
I'll share a screen of this

duke7553:
CoreApplication.MainView.Dispatcher = The event dispatcher associated with the current view itself.
CoreApplication.MainView.CoreWindow.Dispatcher = Event dispatcher for the Window

My best guess is that the UI dispatcher from MainView is sometimes not ready because the CoreWindow itself is initialized before the ApplicationView is.

PR #938, opened 2020-06-07
fixed auto selection when renaming if extensions are hidden

xPoppyx:
fixed for grid view and list view

**Before:**
before

**After:**
after

### DIFF
--- a/Files/UserControls/LayoutModes/GenericFileBrowser.xaml.cs
+++ b/Files/UserControls/LayoutModes/GenericFileBrowser.xaml.cs
@@ -260,8 +260,17 @@ namespace Files
 
             renamingTextBox = e.EditingElement as TextBox;
             renamingTextBox.Focus(FocusState.Programmatic); // Without this, cannot edit text box when renaming via right-click
-            renamingTextBox.Select(0, SelectedItem.ItemName.Length - extensionLength);
             renamingTextBox.TextChanged += TextBox_TextChanged;
+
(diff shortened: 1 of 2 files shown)


Files/UserControls/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -260,8 +260,17 @@ private void AllView_PreparingCellForEdit(object sender, DataGridPreparingCellFo
 
             renamingTextBox = e.EditingElement as TextBox;
             renamingTextBox.Focus(FocusState.Programmatic); // Without this, cannot edit text box when renaming via right-click
-            renamingTextBox.Select(0, SelectedItem.ItemName.Length - extensionLength);
             renamingTextBox.TextChanged += TextBox_TextChanged;
+
+            if (App.AppSettings.ShowFileExtensions == true)
```
tsvietOK, line 265: I think it is possible to do the same with less code.
```csharp
            renamingTextBox = e.EditingElement as TextBox;
            renamingTextBox.Focus(FocusState.Programmatic); // Without this, cannot edit text box when renaming via right-click

            int selectedTextLength = SelectedItem.ItemName.Length;
            if (App.AppSettings.ShowFileExtensions)
            {
                selectedTextLength -= extensionLength;
            }
            renamingTextBox.Select(0, selectedTextLength);
            renamingTextBox.TextChanged += TextBox_TextChanged;
```
Same with GridViewBrowser.xaml.cs

PR #942, opened 2020-06-07
Dialogs: Hide rename dialog if user pressed Esc button

tsvietOK:


yaichenbaum:
Where are we still using the rename dialog? Ideally all file name edits should be done with inline editing...

tsvietOK:
Rename dialog is used when creating new files

PR #951, opened 2020-06-09
Added option to unpin libraries from the sidebar

yaichenbaum:


Files/View Models/SettingsViewModel.cs
```diff
@@ -95,6 +90,20 @@ private async void PopulatePinnedSidebarItems()
             StorageFolder cacheFolder = ApplicationData.Current.LocalCacheFolder;
             ListFile = await cacheFolder.CreateFileAsync("PinnedItems.txt", CreationCollisionOption.OpenIfExists);
 
+            if (SystemInformation.IsFirstRun || SystemInformation.IsAppUpdated)
```
tsvietOK, line 93: Is it a good idea to pin all items again after app update?

yaichenbaum: This is only going to be for the next update, it will be removed after this update. The reasoning behind this is that without doing this, users will have no pinned libraries unless they reset the app.

PR #960, opened 2020-06-09
Add margin to gridview item

tsvietOK:
Before/After
image


XPoppyX:
Captura de pantalla (11)

PR #977, opened 2020-06-10
Added Japanese translation

hez2010:
- Added Japanese translation

Files on Japanese version of Windows 10 will display Chinese translation, so I added Japanese translation.

I'm not a native Japanese speaker, if there're any mistakes in my translation please point out.


hez2010:
@yair100 Done.

yair100 (approved):

PR #1026, opened 2020-06-12
Updated status bar

yaichenbaum:
- Added status for the selected items count
- Added the size of the selected item (right now it only shows the size if one item is selected)


XPoppyX:
Is it possible to also display how many items are in the current directory?

yaichenbaum:
@XPoppyX It is, that just hasn't been implemented on an level yet so it will come in a future update.

PR #1001, opened 2020-06-11
Prevent the file watching loop from still being alive after we kill it

duke7553:
Under certain conditions, execution of blocks of code inside the file watching loop continues although the loop's condition is false, causing unwanted actions to be performed (e.g. using the closed directory handle) 

This PR aims to:
- Prevent a crash from occurring upon changing the working directory quickly
- Remove unneeded checks before adding/removing files during the watcher operation
- Convert any direct uses of the FilesAndFolders property to use a "snapshot" list instead

Please test this PR by doing the following:
1. Removing/Adding/Renaming a file after refreshing the directory multiple times
2. **RELEASE MODE:** Changing the WorkingDirectory by rapidly invoking various sidebar locations

PR #1048, opened 2020-06-13
Cancel previous watcher

gave92:
Proposal for #1001
\- Use `CancelIoEx` to cancel previous watcher: watcher is started on a background thread so `CancelIo` has no effect.
\- Close overlapped.hEvent handle
\- Close hWatchDir handle

### DIFF
--- a/Files/Helpers/NativeDirectoryChangesHelper.cs
+++ b/Files/Helpers/NativeDirectoryChangesHelper.cs
@@ -20,6 +20,9 @@ namespace Files.Helpers
         [DllImport("api-ms-win-core-io-l1-1-1.dll")]
         public static extern bool CancelIo(IntPtr hFile);
 
+        [DllImport("api-ms-win-core-io-l1-1-1.dll")]
+        public static extern bool CancelIoEx(IntPtr hFile, IntPtr lpOverlapped);
+
         [DllImport("api-ms-win-core-synch-l1-2-0.dll")]
         public static extern uint WaitForMultipleObjectsEx(uint nCount, IntPtr[] lpHandles, bool bWaitAll, uint dwMilliseconds, bool bAlertable);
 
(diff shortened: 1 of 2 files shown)


duke7553:
@gave92 I really like these changes. Thank you for catching them. I added a commit which nulls out the thread to prevent the closeWatcher() block from running more than needed. Let me know what you think about that.

duke7553:
@gave92 The other thing I noticed consistently was holding down the F5 key causes unexpected behavior. I wonder if we can somehow prevent repeating this keyboard accelerator

gave92:
@duke7553 Totally agree on setting `aWatcherAction` to null, closing the handle twice would be bad. Do you know if you can set a handle to a "safe" value? Such that calling `CloseHandle` on that has no effect? Perhaps: `hWatchDir = new IntPtr(-1)`.

duke7553:
@gave92 I feel like I know for sure that setting the handle to -1 would still throw an invalid handle error. I tried that strategy already. Maybe there is another generic value we can set the handle to.

gave92:
@duke7553 I think we still need to handle the case of multiple tabs opened: when you close the tab we should cancel the watcher? There is a `Dispose` method in `ItemViewModel` but it's not called from anywhere.
Also is it a big issue if we don't close the watcher when the app exit?

duke7553:
> @duke7553 I think we still need to handle the case of multiple tabs opened: when you close a tab should we cancel the watcher? There is a `Dispose` method in `ItemViewModel` but it's not called from anywhere.

We do need to handle that case. Good idea. On a side note: I wondered before why memory usage stays the same even after tabs are closed, which is probably a different issue and out of scope for this PR.
> Also is it a big issue if we don't close the watcher when the app exits?
I think it doesn't matter, but we may be registered for the system prefetch, so it may become problematic.

gave92:
@duke7553 Keeping F5 pressed does not causes issues for me (at least in Debug mode)

gave92:
@duke7553 I just discovered the existence of [SafeFileHandle](https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.safehandles.safefilehandle?view=netcore-3.1). Might be of help when dealing with `hWatchDir` cleanup.

duke7553:
Merged this to the branch we're working off of

PR #1055, opened 2020-06-14
Added Support for Date Accessed and File Owner Properties

VigneshVoid:
I have added support for two new properties similar to Windows Explorer. We can have a whole lot of properties from the list in the URL

https://docs.microsoft.com/en-us/windows/win32/properties/core-bumper

Screenshot


yaichenbaum:
Instead of getting the info from code behind in the properties dialog, we should do it in SelectedItemsPropertiesViewModel and bind the TextBlock in the properties dialog to the properties in the SelectedItemsPropertiesViewModel.

VigneshVoid:
Sure will work on that and add in the next commit @yair100

yaichenbaum:
Is there a reason you added `.Properties`? I feel that is a little redundant in this case since `SelectedItemsPropertiesViewModel` is going to be limited to the file properties anyways.

VigneshVoid:
I added to separate Main Item Properties from any visibility properties or any local properties need for the methods within ViewModel. So it will be easy to differentiate from Main and Local Method Properties. 

Mainly for Code Readability. @yair100

Files/Views/Pages/Properties.xaml
```diff
@@ -152,21 +154,24 @@
...
-                    Visibility="{x:Bind ItemProperties.ItemMD5HashVisibility, Mode=OneWay}" />
+                    Text="{x:Bind ViewModel.Properties.ItemMD5Hash, Mode=OneWay}"
+                    Visibility="{x:Bind ViewModel.Properties.ItemMD5HashVisibility, Mode=OneWay}" />
                 <ProgressBar
                     x:Name="ItemMD5HashProgress"
                     Grid.Row="5"
                     Grid.Column="1"
                     Margin="20,0,0,0"
-                    Visibility="{x:Bind ItemProperties.ItemMD5HashProgressVisibility, Mode=OneWay}" />
+                    IsIndeterminate="True"
```
tsvietOK, line 171: IsIndeterminate should be removed

VoidExploiter: Resloved in this [commit ](https://github.com/VoidExploiter/files-uwp/commit/5c602a02968174eeb03bf84ec856261d4e85e837)

Files/Interacts/Interaction.cs
```diff
@@ -1114,7 +1114,7 @@ public void PushJumpChar(char letter)
             App.CurrentInstance.ViewModel.JumpString += letter.ToString().ToLower();
         }
 
-        public async Task<string> GetHashForFile(ListedItem fileItem, string nameOfAlg, CancellationToken token, ProgressBar progress)
+        public async Task<string> GetHashForFile(ListedItem fileItem, string nameOfAlg, CancellationToken token, double progress)
```
tsvietOK, line 1117: Bring back ProgressBar. Otherwise it is not working.

VoidExploiter: Fixed it by revoking back to progressBar [here](https://github.com/VoidExploiter/files-uwp/commit/72ed0807f351df77a62ea6ef85d12d8be872949a)

tsvietOK:
I don't think it is really useful to display item owner in Properties dialog. @yair100 What do you think about it?
Moreover, you displaying "File owner" even if item is folder.
However properties retrieving way you offered could be useful in the future.

yaichenbaum:
@tsvietOK I agree that it's a bit much, what we can do is add a feature flag for this to the experimental page and if a large percentage of users turn it on we can look into making this the default.

XPoppyX:
maybe in the future you could add tabs or a sidebar for more details / options
Anotación 2020-06-16 132524

yaichenbaum:
@XPoppyX At some point we can do that, but I think we should get the basics first, for example, showing multiple dialogs if the user clicked properties when multiple files are selected.

VigneshVoid:
"showing multiple dialogs if the user clicked properties when multiple files are selected."

@yair100  supported the above requirement in this [commit](https://github.com/files-community/files-uwp/pull/1055/commits/5dca9f66f72e0d499cee29d69afc21266d95cd00)

Screenshot

yaichenbaum:
@VoidExploiter I actually made a mistake in the above comment, I should have said support for properties when opening the dialog when multiple items ae selected. We can definitely explore keeping this but my thoughts are that when users select multiple files and hit "properties", they will want to know the size of all of them together and other information once it's extended.

VoidExploiter:
We can provide a settings to open properties in individual windows for multiple files or on same window. 

I will work on that and also opening a window with a combined properties.

image

yaichenbaum:
@VoidExploiter Can you add the setting to the experimental settings page? This pull request is looking very good so far 🙂.

VoidExploiter:
Sure will add multiple windows and also file owner flag in that

VigneshVoid:
@yair100  Moved File Owner and Multiple windows to Experimental Flag

image

XPoppyX:
> they will want to know the size of all of them together and other information once it's extended.

like in file explorer show how many files and folder was selected, location, total size and type

Files/View Models/SettingsViewModel.cs
```diff
@@ -300,7 +300,33 @@ private async void DetectOneDrivePreference()
...
+            get => (bool)GetSettingsValue("OpenPropertiesInMultipleWindows",false);
+            set => localSettings.Values["OpenPropertiesInMultipleWindows"] = (bool)value;
+        }
+        public bool ShowFileOwner
+        {
+            get => (bool)GetSettingsValue("ShowFileOwner", false);
+            set => localSettings.Values["ShowFileOwner"] = (bool)value;
+        }
 
+        protected TResult GetSettingsValue<TResult>(string name, TResult defaultValue)
```
yaichenbaum, line 314: Is there a reason you built your own GetSettingsValue when one is already implemented in the SettingsViewModel?

yaichenbaum:
@VoidExploiter Can you add strings for the new text that you added?

yair100:
> I added to separate Main Item Properties from any visibility properties or any local properties need for the methods within ViewModel. So it will be easy to differentiate from Main and Local Method Properties.
> 
> Mainly for Code Readability. @yair100

I don't think this is necessary, as anything this ViewModel is already for item properties.

VoidExploiter:
Okay will add the strings and move properties Model back into ViewModel

Files/View Models/SettingsViewModel.cs
VoidExploiter: Resolved in this [commit](https://github.com/files-community/files-uwp/pull/1055/commits/adaff7ea13068ec6aa50d07d255b341c2b3e132f)

VigneshVoid:
> > I added to separate Main Item Properties from any visibility properties or any local properties need for the methods within ViewModel. So it will be easy to differentiate from Main and Local Method Properties.
> > Mainly for Code Readability. @yair100
> 
> I don't think this is necessary, as anything this ViewModel is already for item properties.

Removed in this [commit](https://github.com/files-community/files-uwp/pull/1055/commits/636f7f33b0387a90fb8079c91ce349605d2919c7)

VoidExploiter:
> @VoidExploiter Can you add strings for the new text that you added?

Will add for english

Files/View Models/SettingsViewModel.cs
```diff
@@ -300,7 +300,16 @@ private async void DetectOneDrivePreference()
                 PinOneDriveToSideBar = false;
             }
         }
-
+        public bool OpenPropertiesInMultipleWindows
+        {
+            get => Get<bool>(false,"OpenPropertiesInMultipleWindows");
+            set => localSettings.Values["OpenPropertiesInMultipleWindows"] = (bool)value;
```
yaichenbaum, line 306:
```suggestion
            get => Get(false);
            set => Set(value);
```

Files/View Models/SettingsViewModel.cs
```diff
@@ -300,7 +300,16 @@ private async void DetectOneDrivePreference()
...
+        }
+        public bool ShowFileOwner
+        {
+            get => Get<bool>(false, "ShowFileOwner");
+            set => localSettings.Values["ShowFileOwner"] = (bool)value;
```
yaichenbaum, line 311:
```suggestion
            get => Get(false);
            set => Set(value);
```

Files/View Models/SelectedItemsPropertiesViewModel.cs
```diff
@@ -82,110 +92,143 @@ public bool IsItemSelected
...
+            set => Set(ref _ItemMD5HashProgressVisibiity, value);
+        }
+        #endregion
+        #region Properties
+        public ProgressBar ItemMD5HashProgress
+        {
+            get; set;
+        }
+        public ListedItem Item { get; }
+        public Visibility FileOwnerVisibility { get => App.AppSettings.ShowFileOwner ? Visibility.Visible : Visibility.Collapsed; }
```
yaichenbaum, line 141: Why don't we just bind the TextBlock in the property dialogs xaml to this setting instead of checking it in the view model?

yair100 (changes requested):
Please address the above comments.

Files/View Models/SelectedItemsPropertiesViewModel.cs
VoidExploiter: Resolved [here](https://github.com/files-community/files-uwp/pull/1055/commits/ceaea013f1dc474a46170806cd0314f934bde7c2)

yair100 (approved):

yair100 (approved):

PR #1063, opened 2020-06-14
Improve performance on selection operations

hez2010:
- Added BulkObservableCollection: support bulk operations based on ObservableCollection, and make sorting process faster
- Optimize SelectAll: call non-public method `DataGrid.SelectAll()` to make Select All operation as fast as Ctrl+A

A further question:
How to optimize Invert Selection? DataGrid used `SetRowsSelection()` internally for fast selection operation, I think we can make it use for implementing faster Invert Selection.

https://github.com/windows-toolkit/WindowsCommunityToolkit/blob/4aca25975f3cec90d4f86d5b5239aa6bc185880b/Microsoft.Toolkit.Uwp.UI.Controls.DataGrid/DataGrid/DataGridRows.cs#L797

### DIFF
--- a/Files/Files.csproj
+++ b/Files/Files.csproj
@@ -167,6 +167,7 @@
     <Compile Include="Helpers\NaturalStringComparer.cs" />
     <Compile Include="Helpers\PackageHelper.cs" />
     <Compile Include="Helpers\StringExtensions.cs" />
+    <Compile Include="Helpers\BulkObservableCollection.cs" />
     <Compile Include="Helpers\ThemeHelper.cs" />
     <Compile Include="Program.cs" />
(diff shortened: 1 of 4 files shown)


yaichenbaum:
@hez2010 Will this work for the grid layouts as well?

yaichenbaum:
@hez2010 Great work! I am not sure about the invert selection but it's a already a great improvement for selection and clearing the selection.

hez2010:
I have not tested this with grid layout yet, but I think it should work.

hez2010:
Update: Grid layout uses a separated logic for `SelectAllItems`, it's already fast enough. This PR only affects the behavior of list layout.

PR #1066, opened 2020-06-14
feat: Uri activation with path

FelixZacher:
allows to activate the app via uri and specify a path.

```
files-uwp:///C:/windows
```

1. The drive letter must be uppercase
2. forward slashes `/` must be used.

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -308,9 +308,14 @@ namespace Files
                 case ActivationKind.Protocol:
                     var eventArgs = args as ProtocolActivatedEventArgs;
 
-                    if (eventArgs.Uri.AbsoluteUri == "files-uwp:")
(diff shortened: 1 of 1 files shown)


tsvietOK:
@lampenlampen Seems like your patch get function "Open in new windows" broken. Please, check it.

yaichenbaum:
Don't we already support it using `files`?

FelixZacher:
@tsvietOK Will have a look. Thanks 👍
@yair100 We only support the commandline with `files.exe -Directory ""` and to start the app with the uri `files-uwp:`. This adds support to open the app in a folder via uri `files-uwp:///C:/users`.

This is useful for other uwp apps, because they can then start files with `Launcher.LaunchUriAsync()`, because starting another process is prohibited by default.

lampenlampen:
Okay now i learned, we already support this 🤣.

PR #1072, opened 2020-06-15
Added ability to cancel hash computation

hez2010:
- Cancel hash computation after property window closed

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -12,6 +12,7 @@ using System.IO.Compression;
 using System.Linq;
 using System.Reflection;
 using System.Text.RegularExpressions;
+using System.Threading;
 using System.Threading.Tasks;
 using Windows.ApplicationModel.Core;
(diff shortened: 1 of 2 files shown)


lampenlampen:
I see, that you implemented `IDisposable`. Does the runtime/uwp-framework call `Properties.Dispose()` on a page, if it's implemented?

hez2010:
Added some logic.
`CancellationTokenSource` will finally disposed by GC, but it can also be disposed by calling `Dispose()`

lampenlampen:
Why do we even need to implement it? If the dialog is closed, we can handle it in the unloaded event.

hez2010:
Done.

> If the dialog is closed, we can handle it in the unloaded event.

Sounds reasonable.

lampenlampen:
Okay you still saw my answer ;)

My understanding is either the `unloaded`-event fires or the app crashes, so in both cases we are safe.

In the end i decided you option doesn't harm.

PR #1081, opened 2020-06-15
Launch file (with other files containing in folder as param) to support apps like Photos and Movies & TV + Include Files.Extensions (unused at the moment)

ADeltaX:


ADeltaX:
Files.Extensions is unused at the moment. This will be used in the near future for handling desktop/tablet mode changes with relatively preferred user input type (mouse/touch).

gave92:
Can we use the same approach to open multiple selected files in the same app? This would be a better way of doing it instead of what I did here #1062

PR #1084, opened 2020-06-16
Added pin path to sidebar to right click context menu

### LINKED ISSUE #255: Add option to right click context menu to pin the current directory to the sidebar
I want to add a whole smb-server to the sidebar. Since the network address section isn't shown, I cannot view network addresses and therefore pin e.g. whole SMB-servers through right click to the sidebar. (e.g. such addresses: **\\\smb-server**)

Please enable us to pin the current path to the sidebar.

Maybe right-clicking on the current path (or anywhere inside the current folder) and showing the option to pin the current path to the sidebar could be the UX for that. In the file explorer you can do this through pulling the current path into the sidebar section.

yaichenbaum:
Closes #255.

### DIFF
--- a/Files/DataModels/SidebarPinnedModel.cs
+++ b/Files/DataModels/SidebarPinnedModel.cs
@@ -145,6 +145,10 @@ namespace Files.DataModels
             {
                 iconCode = "\uE8B2";
             }
+            else if(Path.GetPathRoot(path).Equals(path, StringComparison.OrdinalIgnoreCase))
+            {
+                iconCode = "\uEDA2";
+            }
             else
             {
                 iconCode = "\uE8B7";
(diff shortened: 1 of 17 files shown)


XPoppyX:
Wouldn't it be better if it were between "refresh and "paste"?
Captura de pantalla (26)

yaichenbaum:
@XPoppyX I haven't thought about that, do you have any reason why it would be better higher up?

XPoppyX:
I think "properties" should always be the last. it is not entirely necessary to put it between "refresh" and "paste", above "properties" is fine too, that depends on you, but it definitely shouldn't be under "properties"

yaichenbaum:
@XPoppyX I don't think moving the location will be an issue, is this something you are interested in working on?

XPoppyX:
in the context menu of folders, "pin to sidebar" is above "properties", just put them in the same way and it will be fine

PR #1091, opened 2020-06-17
Bug fixes for handling network path

hez2010:
Fixes:
- Tag will lost prefixed `\\` if it's a network path, and will crash if try to navigate by clicking PathItemBox
- Path.GetRootPath() may return null or empty string, which leads to crash in NormalizePath()


hez2010:
Ping. Any updates on this PR? It blocks network device access on files-uwp.

PR #1109, opened 2020-06-18
Fix bug with Status Bar control not displaying information through multiple tabs

duke7553:
This is fixed by making the properties specific to the tab instances themselves. We also rename IShellPage.ViewModel to IShellPage.FilesystemViewModel to help with the distinction between these view model properties.


yaichenbaum:
@duke7553 It looks like some merge conflicts came up, can you resolve them?

PR #1136, opened 2020-06-19
Set WorkingDirectory for terminal

hez2010:
- Set WorkingDirectory for terminal so that cmd no longer need to execute `cd` command for switching directory.

### DIFF
--- a/Files/Assets/terminal/terminal.json
+++ b/Files/Assets/terminal/terminal.json
@@ -12,13 +12,13 @@
     {
       "name": "PowerShell Core 6",
       "path": "pwsh.exe",
-      "arguments": "-WorkingDirectory \"{0}\"",
+      "arguments": "",
       "icon": ""
     },
     {
       "name": "Windows Terminal",
       "path": "wt.exe",
-      "arguments": "-d \"{0}\"",
+      "arguments": "-d .",
       "icon": ""
     }
     */
--- a/Files/Assets/terminal/terminal.json
+++ b/Files/Assets/terminal/terminal.json
@@ -5,7 +5,7 @@
     {
       "name": "CMD",
       "path": "cmd.exe",
-      "arguments": "/k \"cd /d {0} && title Command Prompt\"",
+      "arguments": "",
       "icon": ""
     }
     /*
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -142,7 +142,8 @@ namespace Files.Interacts
                 {
                     { "WorkingDirectory", CurrentInstance.ViewModel.WorkingDirectory },
                     { "Application", terminal.Path },
-                    { "Arguments", string.Format(terminal.Arguments, CurrentInstance.ViewModel.WorkingDirectory) }
+                    { "Arguments", string.Format(terminal.Arguments,
+                        InstanceTabsView.NormalizePath(CurrentInstance.ViewModel.WorkingDirectory)) }
                 };
                 await App.Connection.SendMessageAsync(value);
             }
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -138,9 +138,12 @@ namespace Files.Interacts
 
             if (App.Connection != null)
             {
-                var value = new ValueSet();
-                value.Add("Application", terminal.Path);
-                value.Add("Arguments", string.Format(terminal.Arguments, CurrentInstance.ViewModel.WorkingDirectory));
+                var value = new ValueSet
+                {
+                    { "WorkingDirectory", CurrentInstance.ViewModel.WorkingDirectory },
+                    { "Application", terminal.Path },
+                    { "Arguments", string.Format(terminal.Arguments, CurrentInstance.ViewModel.WorkingDirectory) }
+                };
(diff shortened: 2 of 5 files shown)


lampenlampen:
Do you know, if this works for the other Terminal apps as well?

hez2010:
> Do you know, if this works for the other Terminal apps as well?

Tested with cmd, powershell and wsl. 
Major shells will set WorkingDirectory as current directory such as cmd, powershell, bash and etc.

Windows Terminal needs explicit argument "-d {0}".

Files/Views/SettingsPages/Appearance.xaml.cs
```diff
@@ -25,7 +25,7 @@ public Appearance()
             ThemeChooser.SelectedIndex = (int)Enum.Parse(typeof(ElementTheme), ThemeHelper.RootTheme.ToString());
             ThemeChooser.Loaded += (s, e) =>
             {
-                ThemeChooser.SelectionChanged += async (s1, e1) =>
```
tsvietOK, line 28: What purpose of this change?

hez2010: Without `await` in the closure, `async` here is unnecessary but adds overhead. VS also produces warnings for it.

tsvietOK: Did you test this change? Is changing theme working properly?

hez2010: Sorry for the out-of-scope change, but I've tested it and it works fine.
image
image

Files/Assets/terminal/terminal.json
```diff
@@ -5,7 +5,7 @@
     {
       "name": "CMD",
       "path": "cmd.exe",
-      "arguments": "/k \"cd /d {0} && title Command Prompt\"",
```
tsvietOK, line 8: This change will not update user's `terminal.json`. So will it work with old arguments?

hez2010: Yes, old arguments still work.

Files/Interacts/Interaction.cs
```diff
@@ -138,9 +138,12 @@ public async void OpenDirectoryInTerminal(object sender, RoutedEventArgs e)
...
             if (App.Connection != null)
             {
-                var value = new ValueSet();
-                value.Add("Application", terminal.Path);
-                value.Add("Arguments", string.Format(terminal.Arguments, CurrentInstance.ViewModel.WorkingDirectory));
+                var value = new ValueSet
+                {
+                    { "WorkingDirectory", CurrentInstance.ViewModel.WorkingDirectory },
+                    { "Application", terminal.Path },
+                    { "Arguments", string.Format(terminal.Arguments, CurrentInstance.ViewModel.WorkingDirectory) }
```
tsvietOK, line 145: To fix Windows Terminal opening after app launch:
```suggestion
                    { "Arguments", string.Format(terminal.Arguments, InstanceTabsView.NormalizePath(CurrentInstance.ViewModel.WorkingDirectory)) }
```

hez2010: Done.
Also change default arguments for other terminals and tested ok.

Files/View Models/SelectedItemsPropertiesViewModel.cs
```diff
@@ -150,6 +150,13 @@ public string ItemMD5Hash
                 }
             }
         }
+        private bool _ItemMD5HashCalcError;
```
tsvietOK, line 153: Remove it from this PR.

Files/Assets/terminal/terminal.json
```diff
@@ -12,13 +12,13 @@
     {
       "name": "PowerShell Core 6",
```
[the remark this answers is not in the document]
tsvietOK: It is not needed, this profile here is just for example.

Files/View Models/SelectedItemsPropertiesViewModel.cs
hez2010: Sorry but I don't know why my changes on another branch be commit to this branch....

hez2010: Done

Files/Assets/terminal/terminal.json
```diff
@@ -5,20 +5,20 @@
     {
       "name": "CMD",
       "path": "cmd.exe",
-      "arguments": "/k \"cd /d {0} && title Command Prompt\"",
+      "arguments": "",
       "icon": ""
     }
     /*
```
yaichenbaum, line 11: By removing the comment, what would happen if Powershell is not installed?

duke7553: It is an example profile which users can derive new profiles from.

yaichenbaum: Correct, but by removing the `*/` it adds it is actually adding it as one of the profiles.

lukeblevins: @yair100 I just viewed the raw file, and they don't appear to be removed

duke7553: image

PR #1141, opened 2020-06-19
Fix issue where terminal.json has been overwritten on every app launch

tsvietOK:


Files/DataModels/TerminalFileModel.cs
```diff
@@ -50,17 +50,19 @@ public async Task<bool> AddOrRemoveTerminal(TerminalModel terminal, string packa
             }
             else if (!isInstalled)
             {
-                if (string.IsNullOrWhiteSpace(DefaultTerminalPath))
+                if (Terminals.Remove(Terminals.FirstOrDefault(x => x.Path.Equals(terminal.Path, StringComparison.OrdinalIgnoreCase))))
```
duke7553, line 53: What does Terminals.Remove() return here?

tsvietOK: image

duke7553: Aha. Thanks.

PR #1145, opened 2020-06-19
Properties: Fix some properties loading on older systems

tsvietOK:
Optimize code

### DIFF
--- a/Files/View Models/SelectedItemsPropertiesViewModel.cs
+++ b/Files/View Models/SelectedItemsPropertiesViewModel.cs
@@ -173,10 +173,9 @@ namespace Files.View_Models
         }
         #endregion
         #region Properties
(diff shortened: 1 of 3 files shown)


Files/View Models/SelectedItemsPropertiesViewModel.cs
```diff
@@ -332,7 +331,6 @@ public async Task GetPropertiesAsync(CancellationTokenSource _tokenSource)
                 catch (Exception ex)
                 {
                     NLog.LogManager.GetCurrentClassLogger().Error(ex, ex.Message);
-                    ItemMD5HashProgress.Value = ItemMD5HashProgress.Maximum;
```
hez2010, line 335: If ShowError is true, the color of ProgressBar will become yellow to indicate that there're errors occurred.
However, if the value of ProgressBar is 0, user won't know that the progress of calculation failed because nothing becomes yellow.

PR #1167, opened 2020-06-20
Fix open in new window

gave92:
This PR fixes the right click + open in new window functionality.
Probably after 53ba1a87f2675a5dfd9a01a6c24ed44b3ed5341a I'm getting NullPointerException on `ThemeHelper.Initialize()`

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -284,7 +284,6 @@ namespace Files
                 Window.Current.Content = rootFrame;
             }
 
-            ThemeHelper.Initialize();
             var currentView = SystemNavigationManager.GetForCurrentView();
             switch (args.Kind)
             {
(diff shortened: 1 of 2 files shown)


Files/App.xaml.cs
```diff
@@ -300,6 +299,7 @@ protected override async void OnActivated(IActivatedEventArgs args)
                         var trimmedPath = eventArgs.Uri.OriginalString.Split('=')[1];
                         rootFrame.Navigate(typeof(InstanceTabsView), @trimmedPath, new SuppressNavigationTransitionInfo());
                     }
+                    ThemeHelper.Initialize();
```
tsvietOK, line 302: Don't need to duplicate command. Just move it to `InstanceTabsView` at the and of main constructor. And remove This command at all from App.xaml.cs.

PR #1169, opened 2020-06-20
Properties: Display how many folders and files a folder contains

### LINKED ISSUE #1134: Properties dialog: Display how many folders and files a folder contains
Anotación 2020-06-18 191650

tsvietOK:
Calculate size of hidden, system and link files too
Changed Properties lines spacing to avoid spacing on the hidden line:
image
Fixes #1134 
Final:
image


yaichenbaum:
@tsvietOK Instead of putting the new properties in `SelectedItemsPropertiesViewModel`, they would fit better in `DirectoryPropertiesViewModel.cs`.

tsvietOK:
@yair100 The new properties are part of `SelectedItemsPropertiesViewModel` as files and folders count calculation is happening here too.

yaichenbaum:
@tsvietOK That must have gotten past in a different pull request. In order to keep everything decoupled, all calculations should be done **outside** of the view models. The view models should be limited to holding the properties, the properties can be set from classes in the commands folder.
`SelectedItemsPropertiesViewModel` Should be limited to properties related to the selected item. The item might be a folder so there can an item count property for the items within the selected item, however the calculations should be done outside the view model.
`DirectoryPropertiesViewModel` Should be limited to properties related to the current directory, this can be properties such as the item count within the current directory, the layout mode for the directory and other related properties.

yair100 (approved):

lukeblevins (approved):

PR #1186, opened 2020-06-21
1185: "Open", "Open with", "open in new tab / window" and "pin to sidebar" no longer appear in the item context menu

BuraChuhadar:
Fixing an issue with assigning the wrong method to the right click menuflyout.


yaichenbaum:
@XPoppyX

yaichenbaum:
:shipit:

Issue #1182, opened 2020-06-21
the app doesn't launch at all.


Badr-Zh:
as the title suggests, the app never launches. I've had this problem since earlier versions. I had the app store version before and I just uninstalled it today and installed the latest version from Github but I sill have the same problem. I click on launch, the logo appears then crashes immediately.
app version: 0.10.2.0
windows version: 2004 os build 19041.329

hez2010:
Could you please paste your logs which located in %localappdata%\Packages\49306atecsolution.FilesUWP_et10x9a9vyk8t\LocalState ?
There're log files debug.log and debug_fulltrust.txt which could help us investigate this problem.

Jaiganeshkumaran:
@Badr-Zh I was having the same issue. I fixed it by changing the Build Action for the file "Files\Assets\terminal\terminal.json" from "None" to "Content" in my local branch (plus reinstalling the app afterwards).

Kaedin:
Same issue here with app version 0.10.4.0.

Badr-Zh:
I've fixed the issue. My issue was a Microsoft store issue, all Microsoft store apps suffered the same problem. now the app works fine. here is how I fixed it, I just typed these two commands:
 CMD: wsreset
 and in PowerShell I typed this:  
 Get-AppXPackage | Foreach {Add-AppxPackage -DisableDevelopmentMode -Register "$($_.InstallLocation)\AppXManifest.xml"}

Kaedin:
> I've fixed the issue. My issue was a Microsoft store issue, all Microsoft store apps suffered the same problem. now the app works fine. here is how I fixed it, I just typed these two commands:
> CMD: wsreset
> and in PowerShell I typed this:
> Get-AppXPackage | Foreach {Add-AppxPackage -DisableDevelopmentMode -Register "$($_.InstallLocation)\AppXManifest.xml"}

Tried the same and 'Files UWP - Preview' main UI opened for 2 seconds but then closed immediately. And subsequent tries forcibly close again before reaching the main UI.

[
[debug.txt](https://github.com/files-community/files-uwp/files/4817142/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/files-uwp/files/4817143/debug_fulltrust.txt)

](url)

chuchanpr:
Same issue with 0.10.4.0. #1201 

[debug_fulltrust.txt](https://github.com/files-community/files-uwp/files/4821259/debug_fulltrust.txt)
[...]

PR #1190, opened 2020-06-22
Fixed app launch

tsvietOK:
Fixed self reference loop
Fixed NullReferenceException as mentioned [here](https://github.com/files-community/files-uwp/issues/1185#issuecomment-647189413)


XPoppyX:
maybe related to #1182 ?

tsvietOK:
> maybe related to #1182 ?

Version 0.10.2.0 should work fine.

PR #1200, opened 2020-06-22
Fix NullReferenceException for some scenarios

tsvietOK:
`new SidebarPinnedModel()` should be called before using.


lukeblevins (approved):

yaichenbaum:
While this would work, the better solution would be to load the view models from app.cs. Additionally, loading the sidebar items should be done from the sidebar, not from SettingsViewModel.

tsvietOK:
@yair100 Just want to note, to initialize sidebar model without problems, we have to do it after `App.AppSettings` initializing is done in `InstanseViewModel`

Issue #876, opened 2020-05-28
Properties window shows wrong date modified and created date for system root directories


Jaiganeshkumaran:
**Describe the bug**
The properties window seems to show that the system root directories were created and modified on January 1, 1601. 

**To Reproduce**
1. Go to any drive other than network drives and OneDrive.
2. Right-click on the empty area and hit properties.
3. It'll show that the folder was modified and created on January 1, 1601.

**Expected behavior**
It shouldn't show the date modified and created as Windows doesn't store the date modified and created date for drives.

**Screenshots**
[![2020-05-28.png](https://i.postimg.cc/TPNCrrb0/2020-05-28.png)](https://postimg.cc/v19ryx3x)

**Desktop:**
 - OS Version: Windows 10 Pro Insider Preview Build 19624
 - App version: The issue exists on both v0.8.2 and v0.9.0

Issue #1097, opened 2020-06-17
Properties dialog for drives


XPoppyX:
![Anotación 2020-06-17 131048](https://user-images.githubusercontent.com/61259627/84941167-f934b300-b0ae-11ea-8135-f51dd8457eee.jpg)

XPoppyX:
#1211

PR #1211, opened 2020-06-23
Properties: Add support for combined and drive properties

### LINKED ISSUE #259: Refinements to Properties
I suggest these refinements to the properties especially useful when comparing folders/files.
Currently I just can see the size of one file but not for folders and or when selecting more than on folder/file.

I would like to see the overall size of a folder and/or files (and the count of folders and files inside this folder).

Also when choosing more than one file and/or folder, I would like to see the count of files and the overall size.

**Additional context**
This is how it is handled in file explorer:
image

tsvietOK:
New combined properties:
image
New drive properties:
image
Thanks: #1097 
Fixes #259

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -82,7 +82,7 @@ namespace Files
                         if (SelectedItems.Count == 1)
                         {
                             SelectedItemsPropertiesViewModel.SelectedItemsCount = SelectedItems.Count.ToString() + " " + ResourceController.GetTranslation("ItemSelected/Text");
-                            SelectedItemsPropertiesViewModel.ItemsSize = SelectedItem.FileSize;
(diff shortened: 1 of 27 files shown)


Jaiganeshkumaran:
Will this fix #876?

tsvietOK:
@Jaiganeshkumaran It will not show dates for drive properties at all.

XPoppyX:
is it possible to add the "Disk cleanup" option?

yaichenbaum:
@XPoppyX Being that Disk Cleanup is being deprecated in favor of Storage Sense, it would make sense to keep along the same path for Files UWP.

XPoppyX:
on all my disks the type shows as "fixed", what does it mean?

Jaiganeshkumaran:
I think it means that the drive isn’t removable.

XPoppyX:
mm, but it also shows that on my external disk

Issue #710, opened 2020-04-27
Reset terminal json when loading the app if the file is invalid


yaichenbaum:
If the terminal json file has an error inside it then the app will crash when loading, instead of crashing we should reset the json file to be proper so that the app can load.

tsvietOK:
As i told in #562 we should unify all exceptions(replace JsonSerializationException with Exception or add JsonReaderException), so if there will be any json parsing error, just reset this json to default.

lampenlampen:
But if we reset the json file, the user loses all his profiles and we will have to store a default json file. If the terminal loading process goes wrong for whatever reason, i would simply default to cmd and log, that there was an error. So the user can investigate his json file, if he want's to fix it and if not, he gets the cmd terminal.

yaichenbaum:
@lampenlampen That sounds like a better idea.

tsvietOK:
@yaichenbaum we can add an interface, so user can edit terminals through settings, instead of modifying json file manually.

yaichenbaum:
That would be a nice addition, however it would still make sense to include a json file for someone who wants to use the json format.

lampenlampen:
Maybe we should get rid of the json file, especially if we autodetect the common terminals (powershell (core), Winodws Terminal).

And for the few users, who want's to use a custom one, they can choose `custom` and then we present them an input box to specify the path.

I choosed json, because i thought then they can specify multiple terminals and icons and so on, but i don't think we will use that customization.
If a user want's control over multiple terminal, he will install Windows Terminal or terminus, ...

yaichenbaum:
> Maybe we should get rid of the json file, especially if we auto-detect the common terminals (powershell (core), WindowsTerminal).
> 
> And for the few users, who want's to use a custom one, they can choose `custom` and then we present them an input box to specify the path.
> 
> I choosed json, because i thought then they can specify multiple terminals and icons and so on, but i don't think we will use that customization. If a user want's control over multiple terminal, he will install Windows Terminal or terminus, ...

@lampenlampen that is a possibility, and once support is added for importing and exporting the settings json file, users can technically add a custom terminal path even it's not detected automatically by Files.

yaichenbaum:
@d2dyno1 can you confirm that this issue was resolved in your recent PR?

d2dyno1:
It's not been resolved yet. I'm working on this. Simple `try-catch` should be enough.

hecksmosis:
[...]

PR #1224, opened 2020-06-24
Fix JSON loading when it's empty

tsvietOK:
It could happen, when previous app version crashing leaving json file empty.
When deserializing an empty file, it returns null object, which is caused an exception.


Jaiganeshkumaran:
Will this fix #710?

tsvietOK:
@Jaiganeshkumaran It is not supposed to fix it.

PR #1250, opened 2020-06-26
Settings: Remove transition animations

tsvietOK:
Before:
Settings_before
After:
Settings_after


Jaiganeshkumaran:
Why? The animations look very nice.

tsvietOK:
@Jaiganeshkumaran I think we are don't need several animations(from bottom to top + from right to left), we are need one, which is already exists(from bottom to top)

PR #1282, opened 2020-06-28
Fix for issue 306

rjnb50:
This is a proof of concept fix for issue 306.

It's a little bit rough and ready, only implemented in GenericFileBrowser and not also in GridViewBrowser and will need a bit of tidying up, but it provides a good and workable fix for the issue.

### DIFF
--- a/Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
+++ b/Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
@@ -10,6 +10,7 @@ using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using Windows.System;
 using Windows.UI.Core;
 using Windows.UI.Xaml;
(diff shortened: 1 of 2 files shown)


rjnb50:
This is issue 306: https://github.com/files-community/files-uwp/issues/306

yaichenbaum:
@rjnb50 The issue is only present in GenericFileBrowser so it shouldn't be necessary to implement it in GridViewBrowser. Let us know if this is ready for review.

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -436,9 +440,26 @@ protected override ListedItem GetItemFromElement(object element)
             return row.DataContext as ListedItem;
         }
 
-        private void RightClickContextMenu_Opening(object sender, object e)
+        [DllImport("user32.dll", CharSet = CharSet.Auto, ExactSpelling = true)]
```
duke7553, line 443: iirc user32.dll is not available to UWP apps.

duke7553:
@rjnb50 Would it be a better idea to set a Task.Delay() here than fiddling with DateTime ?

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
tsvietOK: It is [available](https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-extension-apis#apis-from-user32dll), but there is no method which is used in this PR.

duke7553: @tsvietOK You're right, it's just very limited.

rjnb50: Thank you; you are both correct. I should have used

```
private readonly Windows.UI.ViewManagement.UISettings uiSettings = new Windows.UI.ViewManagement.UISettings();
uiSettings.DoubleClickTime
```

I'm working on a new version of this fix with a couple of other small tweaks to the code and then I'll update the PR in a few days.

tsvietOK: @rjnb50 You can use `GetDoubleClickTime()` from Vanara library

yaichenbaum:
@rjnb50 I am still getting the same issue as before after testing your changes...

yaichenbaum:
@rjnb50 I am closing this pull request as we have not heard back from you, please reopen the pull request when you get a chance to review the comments.

PR #1287, opened 2020-06-28
Removed docs in favor of the Wiki

yair100:


lukeblevins:
@yair100 Good idea. Maybe we can promote the wiki from these docs also?

lukeblevins (approved):

PR #1331, opened 2020-07-02
Add recently opened directories to the taskbar jumplist

### LINKED ISSUE #1238: Add recently opened directories to the taskbar jumplist

Jaiganeshkumaran:
Implements and closes #1238.

**Know issues:**
- If you remove a jumplist item while the app is running, the folder will be no longer added until you restart the app.
- Drives aren't being added to the jumplist.


Jaiganeshkumaran:
Should we pin recycle bin to the taskbar jumplist?

yaichenbaum:
@Jaiganeshkumaran I think so, the goal would be to allow the user to jump back to recent directories.

Jaiganeshkumaran:
Can we just use the same FolderIcon for all the folders? I've converted the SVG to PNG since the jumplist API only supports bitmaps.

Jaiganeshkumaran:
Should we localize the text displayed for library folders like Documents?

Files/Helpers/JumpListManager.cs
```diff
@@ -0,0 +1,114 @@
...
+                    displayName = localSettings.Values.Get("RecycleBin_Title", "Recycle Bin");
+                }
+                else
+                {
+                    displayName = Path.GetFileName(path);
+                }
+
+                var jumplistItem = JumpListItem.CreateWithArguments(path, displayName);
+                jumplistItem.Description = jumplistItem.Arguments;
+                jumplistItem.GroupName = "ms-resource:///Resources/JumpListRecentGroupHeader";
```
tsvietOK, line 80: You don't need to translate group name as it should be translated by default. Look at this [link](https://github.com/microsoft/Windows-universal-samples/blob/a58c5d9b84456196889edb1f4a6e844c7a729b8b/Samples/JumpList/cs/Scenario4_ChangeSystemGroup.xaml.cs#L49).

Files/Helpers/JumpListManager.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    displayName = "ms-resource:///Resources/SidebarPictures";
+                }
+                else if (path.Equals(App.AppSettings.MusicPath, StringComparison.OrdinalIgnoreCase))
+                {
+                    displayName = "ms-resource:///Resources/SidebarMusic";
+                }
+                else if (path.Equals(App.AppSettings.VideosPath, StringComparison.OrdinalIgnoreCase))
+                {
+                    displayName = "ms-resource:///Resources/SidebarVideos";
```
tsvietOK, line 66: You can use `ResourceController.GetTranslation("SidebarVideos")` instead of this

Files/Helpers/JumpListManager.cs
Jaiganeshkumaran: This is for the automatic recent list which doesn't work with Files UWP since this app isn't not associated with any file type.

Files/Helpers/JumpListManager.cs
Jaiganeshkumaran: I know however when you assign the ms-resource:/// protocol to a jumplist item, Windows will automatically retrieve the string for the current language. This way we don't need to refresh the jumplist after the user changes their language. However we need to refresh the jumplist after an update since the ms-resource:/// protocol is version dependent.

Files/View Models/SettingsViewModel.cs
```diff
@@ -324,6 +324,8 @@ public bool PinRecycleBinToSideBar
                                 App.sideBarItems.Remove(item);
                             }
                         }
+
+                        App.JumpList.RemoveFolder(RecycleBinPath);
```
yaichenbaum, line 328: I don't think it's necessary to remove recycle bin from the recent items list. The only thing disabling recycle bin should do right now, is to unpin it from the sidebar.

Issue #985, opened 2020-06-10
Apps are launching behind the Files UWP app window


Taras-Parfeniuk:
**Describe the bug**
If Windows Terminal set as default terminal app, after a click on 'Open in Terminal' context menu option the terminal window opens behind the Files UWP window, and the Files UWP window remains in focus.

**To Reproduce**
1. Go to 'Settings / Preferences'
2. Select 'Windows Terminal' in 'Terminal Applications' dropdown
3. Go to any directory
4. Open right click context menu
5. Click on 'Open in Terminal' option
6. The Windows Terminal window opens, but the Files UWP window remains in focus

**Expected behavior**
A newly opened terminal window should be focused like it is for CMD terminal application.

**Screenshot**
![FilesUWP-WinTerminal](https://user-images.githubusercontent.com/12896598/84291263-85922380-ab4d-11ea-9747-10692927f0dc.gif)

**Desktop (please complete the following information):**
 - OS Version: Windows 10 1909 (Build: 18363.900)
 - Files UWP Version: 0.9.2.0
 - Windows Terminal Version: 1.0.1401.0

Jaiganeshkumaran:
@XPoppyX Same for me. I think it's a bug.

yaichenbaum:
When opening the app we need to set it to be focused, I haven't looked into it but I would assume it's a simple parameter.

lampenlampen:
for me only for the first or second time, launching the terminal, the terminal is in the background. After the third time it always opens in the foreground

yaichenbaum:
I have not been able to reproduce this issue, @XPoppyX @Jaiganeshkumaran @lampenlampen what version of Windows are you running? Perhaps the behavior changed with 2004 and that is why I cannot reproduce this.

jaigak:
@yair100 I am using Windows 10 Insider Preview Fast Ring.

XPoppyX:
in the microsoft store version it does happen, but with the last changes to the master branch I can no longer reproduce this issue

Jaiganeshkumaran:
If you're using Windows 10 Insider Preview, the issue of apps opening in the background has been fixed with [Build 20150](https://blogs.windows.com/windowsexperience/2020/06/17/announcing-windows-10-insider-preview-build-20150/).

FibYar:
> 
> 
[...]

Issue #1270, opened 2020-06-27
App opening behind the main app.


jrwut:
**Describe the bug**
A clear and concise description of what the bug is when I open files or even I want to submit feedback in about section, the app like Photo and Edge is opening in the background, not show up in front. I am not sure this is the bug of windows or not that will always show the current app on top.

**To Reproduce**
1. Go to 'folder or about in setting'
2. Click on 'photos or submit feedback'
3. Then 'the apps are opening in the background'
4. This is not an error, but it should appear in the front.

**Expected behavior**
A clear and concise description of what you expected to happen. It should appear on top of the current uwp app.

**Screenshots**
![image](https://user-images.githubusercontent.com/67519322/85930186-e6ab4c80-b8ec-11ea-9ef1-2dc23d2f6d69.png)
![image](https://user-images.githubusercontent.com/67519322/85930191-fcb90d00-b8ec-11ea-9ef2-4954c929092a.png)

**Desktop (please complete the following information):**
 - OS Version: [e.g. Windows 10 20H2]
![image](https://user-images.githubusercontent.com/67519322/85930221-338f2300-b8ed-11ea-826c-c355b5105e5a.png)

 - App version: [v0.10.2.0]

**Additional context**
it happens to other UWP app too for example Twitter app, when I open the link from the Twitter app, the new Edge will open in the background.

Jaiganeshkumaran:
It's a Windows bug and not Files UWP bug since it also affects other applications.

duke7553:
Duplicate of #985

PR #1336, opened 2020-07-03
POC to fix issue #985

gave92:
@duke7553 This is what I was trying to do to solve the issue of Window Terminal appearing in the background.

### DIFF
--- a/Files.Launcher/Win32API.cs
+++ b/Files.Launcher/Win32API.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Linq;
@@ -7,11 +6,11 @@ using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Vanara.Windows.Shell;
+using Vanara.PInvoke;
 
 namespace FilesFullTrust
 {
-    internal class Win32API
+    internal static class Win32API
     {
         public static Task<T> StartSTATask<T>(Func<T> func)
         {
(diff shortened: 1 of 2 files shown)


Files.Launcher/Win32API.cs
```diff
@@ -47,97 +47,29 @@ public static async Task<string> GetFileAssociation(string filename)
...
-        private static extern bool FreeLibrary(IntPtr hModule);
-
         public static string ExtractStringFromDLL(string file, int number)
         {
-            IntPtr lib = LoadLibrary(file);
+            var lib = Vanara.PInvoke.Kernel32.LoadLibrary(file);
             StringBuilder result = new StringBuilder(2048);
-            LoadString(lib, number, result, result.Capacity);
-            FreeLibrary(lib);
+            Vanara.PInvoke.User32.LoadString(lib, number, result, result.Capacity);
```
tsvietOK, line 54: Add `using Vanara.PInvoke;` instead of using full namespace path.

Files.Launcher/Win32API.cs
```diff
@@ -165,5 +97,34 @@ public static void UnlockBitlockerDrive(string drive, string password)
                 // If user cancels UAC
             }
         }
+
+        public static void Foreground(this Process process)
```
tsvietOK, line 101: Now window is opening foreground, but it seems like this window is not focused(i can't write command in terminal) and i need to click to focus

gave92: Ok I'm removing the not working code.
- I'm not sure why `SetForegroundWindow()` is not working, according to the docs you are allowed to bring to the foreground a process you started yourself.
- `SetFocus` and `SetActiveWindow` do not work as they require `AttachThreadInput` to succeed (and it's not)
- `AttachThreadInput` does not succeed: in the above that's normal because the thread does not have a message queue

tsvietOK:
Still not working, let me know when you will be ready for review.

gave92:
@tsvietOK unfortunately I'm in the dark with this as I'm not able to reproduce the original issue consistently.
What happens with the latest commit?
- Window Terminal is brought to the foreground but not "activated" (like before)
- Window Terminal is *not* brought to the foreground and the icon on the taskbar flashes
- Something else?

tsvietOK:
@gave92 Window Terminal is not brought to the foreground without any flash.

gave92:
@tsvietOK If you have time, could you check if you have any luck with the latest commit?

tsvietOK:
Windows Terminal is brought to the foreground but not "activated" (like before)

gave92:
@tsvietOK thanks for the help, I'm absolutely not able to reproduce this.
I can either close this PR and let someone else try to have better luck or we can say that having the window in the foreground but not "activated" is better than nothing (in which case I'll revert to the previous code that was much simpler.
Let me know what you think.

tsvietOK:
I will try to fix that by myself as i can reproduce it. I think we will close this PR as it is fixes nothing. Anyway, we could reopen it in any time.

gave92:
Ok good luck!

tsvietOK:
It looks like it is known Windows Terminal [bug](https://github.com/microsoft/terminal/issues/6611). But i can reproduce the same issue with Fluent Terminal.

gave92:
Interesting, and judging from other issues (e.g. #1270) this is not limited to Terminals either.

Issue #1323, opened 2020-07-01
Pressing space key while focused on the "permanently delete" checkbox in the "delete item(s)" modal dialog breaks keyboard navigation


stamminator:
**Describe the bug**
When the modal dialog for deleting a folder/file pops up, pressing `Space` while focused on the "Permanently delete" checkbox causes the focus to leave the dialog and go back to the main window, even though the dialog is still present. There appears to be no way to return focus to the dialog window with keyboard navigation alone, forcing the user to use the cursor.

**To Reproduce**
1. Right click, then select Delete on a file/folder, or press the `Delete` key.
2. Press `Tab` to set focus on the "permanently delete" checkbox field.
3. Press `Space`.
4. Observe that the focus has now been moved to the main window instead of the modal.

**Expected behavior**
1. Pressing the space key should toggle the checked status of the checkbox.
2. It should be impossible for a modal to lose focus and still be visible. Losing focus should dismiss the modal.

**Screenshots**

![sZ0XYLcr5J](https://user-images.githubusercontent.com/10633170/86276969-e5f01000-bba3-11ea-9f67-05f1cfa54cc3.gif)

**Desktop:**
 - OS Version: Windows 10 2004 19041.329
 - App version: v0.11.0

gave92:
Guess this might be related to the method `Page_CharacterReceived` in `GenericFileBrowser` where we have the line `AllView.Focus(FocusState.Keyboard);` that steals the focus from the displayed dialog. We should do this only if no dialog is displayed.

Thomas1664:
@yair100 Since this issue is fixed, you can go ahead and close this.

yaichenbaum:
@stamminator Can you confirm if this is fixed on your end?

stamminator:
@yair100 Well it appeared to be fixed in the initial release that contained the fix several weeks ago, but the .16 and .17 builds are unresponsive for me, so I can't confirm.

EDIT: It looks like all functionality except for the fancy cursor hover effects is completely frozen for about the first 40 seconds after startup, as if it's waiting for some synchronous task to complete. Once I waited for it to become responsive, I was able to confirm the modal issue no longer takes place.

Issue #1322, opened 2020-07-01
Pressing 'Space' to activate QuickLook only works in List View


generalguy41:
**Describe the bug**
Pressing 'Space' to activate QuickLook only works in List View, not Tile or Grid view (of any size).

**To Reproduce**
1. Go to a folder
2. Click on a file
3. Press 'Space' in List View
   a. See that it works
4. Press 'Space' in Tile or Grid View
   a. Doesn't work

**Expected behavior**
QuickLook should work in all views.

**Screenshots**
![ezgif-4-0e134ca30722](https://user-images.githubusercontent.com/8879244/86273830-d15d4900-bb9e-11ea-91b2-c199818e655b.gif)

**Desktop (please complete the following information):**
 - OS Version: Windows 10 2004 20152.1000
 - App version: v0.11.0.0 (Store Version)

lampenlampen:
I have still version 0.10.2.0 and it works there.

gave92:
Could be related to this change 35679a189eb1bb0b0d58695a779bd4c653dd5e87?
The keyboard accelerator for Space key is not firing, instead it seems the GridView is "swallowing" it

Talisman39ar:
This is now working for me (using Quicklook v3.6.9 and the v 0.12.0.0 of Files UWP

generalguy41:
Yea, this was fixed in #1349. Closing now.

PR #1264, opened 2020-06-27
RecycleBin: Remove Properties menu item for the folder itself

tsvietOK:
Fixes #1263

Issue #1263, opened 2020-06-27
Properties window not working for recycle bin


Jaiganeshkumaran:
**Describe the bug**
Properties option isn't working for the recycle bin. This only occurs for the folder itself and not deleted files on it.

**To Reproduce**
1. Enable recycle bin support
2. Navigate to recycle bin
3. Right-click and hit navigate
4. Files UWP will crash 

**Expected behavior**
It should show the properties of the recycle bin.

**Desktop:**
 - OS Version: Windows 10 Insider Preview Build 20152.1000
 - App version: v0.11.0

gave92:
@Jaiganeshkumaran if you're experiencing a crash when opening recycle bin properties, could you post the contents of debug.txt and debug_fulltrust.txt? (possibly right after the problem occours)
Thanks!

Jaiganeshkumaran:
Here you go.

[debug.txt](https://github.com/files-community/files-uwp/files/4841523/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/files-uwp/files/4841524/debug_fulltrust.txt)

PR #1349, opened 2020-07-04
Fix dialog-related issues

gave92:
This PR fixes a few issues related to dialogs.
In particular:
- Pressing 'Space' to activate QuickLook only works in List View #1322
-> GridView seems to "swallow" the Space key, for `GridViewBrowser` the Space key is now handled in `FileList_PreviewKeyDown`

- Properties window not working for recycle bin #1263
-> Properties dialog for recycle bin was removed in #1264, this removes the code causing the exception in #1263

- Pressing space key while focused on the "permanently delete" checkbox in the "delete item(s)" modal dialog breaks keyboard navigation #1323
-> When a key is pressed the file list is focused only if no dialog is currently displayed

- After canceling a permanent deletion, if you press "shift + del" again in the same file, "permanently delete" is no longer checked again #1253
-> Fix a bug with "{x:Bind}" that was causing multiple `ConfirmDeleteDialog` to persist after being closed
-> Fix the "permanently delete" option not being checked if the dialog was canceled and reopened quickly

Let me know if you wish this split into separate PRs for each item (each commit in this fixes a single issue)

### DIFF
--- a/Files/Dialogs/ConfirmDeleteDialog.xaml.cs
+++ b/Files/Dialogs/ConfirmDeleteDialog.xaml.cs
@@ -1,4 +1,5 @@
-﻿using Windows.UI.Xaml;
+﻿using Windows.Storage;
+using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 
 // The Content Dialog item template is documented at https://go.microsoft.com/fwlink/?LinkId=234238
@@ -7,6 +8,7 @@ namespace Files.Dialogs
 {
     public sealed partial class ConfirmDeleteDialog : ContentDialog
     {
+        public StorageDeleteOption PermanentlyDelete { get; set; }
         public MyResult Result { get; set; }
 
         public enum MyResult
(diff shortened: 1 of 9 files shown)


Files/Dialogs/ConfirmDeleteDialog.xaml.cs
```diff
@@ -16,14 +18,26 @@ public enum MyResult
...
+            this.Unloaded += ConfirmDeleteDialog_Unloaded;
+        }
+
+        private void ConfirmDeleteDialog_Unloaded(object sender, RoutedEventArgs e)
+        {
+            // x:Bind "forgets" to unregister bindings when unloading
+            System.Reflection.FieldInfo field = typeof(ConfirmDeleteDialog).GetField("Bindings", 
+                System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
+            dynamic bindings = field.GetValue(this);
+            bindings.StopTracking();
```
duke7553, line 40: I've never done something like this before, so I'm curious why it's necessary in this case.

gave92: I noticed the issue by putting a breakpoint [here](https://github.com/files-community/files-uwp/blob/c16cb7a728f7e3d5fa8fea9227c39aefd84c6dce/Files/View%20Models/InteractionViewModel.cs#L38) and seeing it was getting called more and more times when canceling and reopening the dialog.
This change is probably now unnecessary as I'm now binding `PermanentlyDelete` to a dialog local property instead of a static app property.
I guess it's worth checking if a similar issue exists for other Pages/Dialogs as it could be the source of memory leaks?

PR #1350, opened 2020-07-04
Remove/Reduce Unnecessary Resources in App.xaml

duke7553:
Hopefully this has a pronounced effect on startup time for some users, but I wouldn't get my hopes up.

### DIFF
--- a/Files/Views/ModernShellPage.xaml
+++ b/Files/Views/ModernShellPage.xaml
@@ -95,7 +95,7 @@
                 BlurRadius="4"
                 Canvas.ZIndex="100"
                 OffsetX="-2"
-                OffsetY="10"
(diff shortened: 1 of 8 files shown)


Files/Views/ModernShellPage.xaml
```diff
@@ -95,7 +95,7 @@
                 BlurRadius="4"
                 Canvas.ZIndex="100"
                 OffsetX="-2"
-                OffsetY="10"
+                OffsetY="0"
```
duke7553, line 98: This actually fixed one of the issues with the drop shadow

yaichenbaum, line 98:
```suggestion
```

PR #1356, opened 2020-07-05
Show OneDrive sync status

### LINKED ISSUE #1314: Show OneDrive/iCloud sync status
It'd be helpful if Files UWP shows sync status for OneDrive. These are the different possible sync status:

- Available when online
- Available on this device
- Syncing...
- Can't sync
- Not synced (excluded)

gave92:
Show sync status for files and folders for Onedrive.
Fixes #1314

This PR adds support for showing Onedrive sync status.
Currently the underlying stuff works and it features a ~hideous~beautiful UI, but it can still be improved :)

**To be checked (help needed)**
- Does the sync status appear on files and folders when you navigate to the Onedrive folder?
- Are there any items without icons?
- Are there inconsistencies between windows explorer sync status and status in the app?
- Is performance ok?

**How it works**
Files and folders inside the OneDrive folder have the following file properties:
- [System.FileOfflineAvailabilityStatus](https://docs.microsoft.com/en-us/windows/win32/properties/props-system-fileofflineavailabilitystatus) -> for folders it has the following values:
online: 0
partially offline: 1
offline: 2
offline pinned: 3
empty folder: 5

- System.FilePlaceholderStatus -> for files it has the following values (checked on my pc, could not find documentation):
not onedrive: 6
online: 8
syncing: 9
offline: 14
offline pinned: 15


yaichenbaum:
It looks like the basic functionality works. I would recommend moving the status to it's own column in the list view layout.

gave92:
@yair100 happy to hear that.
So you suggest to mimic windows explorer correct?
- an additional column in datagridview
- an fonticon overlay (like now but smaller and on a corner of item icon) for gridview

yair100:
> @yair100 happy to hear that.
> So you suggest to mimic windows explorer correct?
> 
> * an additional column in datagridview
> * an fonticon overlay (like now but smaller and on a corner of item icon) for gridview

That sounds like a good way to do it.

gave92:
Below the current status of UI, still need polish.

DataGrid
<a href="url"></a>
TileView
<a href="url"></a>
GridView
<a href="url"></a>

Files/Views/LayoutModes/GenericFileBrowser.xaml
```diff
@@ -634,12 +639,29 @@
...
+                    x:Uid="syncStatusColumn"
+                    CanUserResize="False"
+                    CanUserSort="False"
+                    Visibility="{x:Bind local:App.CurrentInstance.InstanceViewModel.IsPageTypeOnedrive, Converter={StaticResource BoolToVisibilityConverter}, Mode=OneWay}"
+                    Binding="{Binding SyncStatusGlyph}"
+                    DisplayIndex="2"
+                    FontSize="14"
+                    FontFamily="Segoe MDL2 Assets"
+                    IsReadOnly="True"
+                    Header="Sync"
```
yaichenbaum, line 653:
```suggestion
                    Header="Status"
```

yaichenbaum:
@gave92 Should we add a color to the sync status glyphs depending on the status?

gave92:
@yair100 I think it would be nice, but I don't know if it would play well when shown above another icon in GridView.
Please also look at the comment above for a few other doubts I have regarding UI.

Files/Views/LayoutModes/GenericFileBrowser.xaml
gave92: Yup I'll change it in the Resources as well

yaichenbaum:
> Is the status icon positioning ok? (Explorer puts the icons near the text in "GridView")

If possible, we should use a similar placement.

> Should the icons in Tile/GridView have a background? Like a filled circle or something?

I think that would look good, if not, it shouldn't be too hard to tweak that later.

> Should the icons in GridView scale when making elements bigger/smaller?

I don't think that's necessary.

> Should the icons be colored (i.e. light blue/filled white for cloud and green for the checkmark)?

I think so, this would help the icons stand out a little bit mor.

gave92:
@yair100 Thank for the feedback, I'll let you know how it coming along.
Edit: added color to icons, I think they're nice! But almost invisible in dark mode :/

lukeblevins (approved):
This is really good considering I wasn't aware we could find this information through the core property system. Maybe change all references of Onedrive -> OneDrive

gave92:
@duke7553 yeah only it won't probably be extendable to other cloud services.
>  Maybe change all references of Onedrive -> OneDrive

Sure I'll do

lampenlampen:
For me it also works for the Synology Drive Client. Maybe we can detect somehow, if a directory or file supports sync functionality or expose a setting, where directories can be added.

lampenlampen:
We should also rename it to a neutral name without `OneDrive`.

gave92:
@lampenlampen Do you mean that with this PR you get sync status icons also for Synology Drive Client?
Perhaps OneDrive -> CloudStorage or similar

lampenlampen:
Yes. I needed to tweak 2 statements to enable sync status for these folders, but then yes. I remember slightly reading in one of the OneDrive-blog posts some years ago, that the underlying infrastructure is available to other cloud-providers as well

Maybe othrs with other Providers can test this as well, so we get a list with which providers it works. (Dropbox, Google, ...?)

gave92:
@lampenlampen Wow that's very nice to hear! For detecting whether a folder is synced perhaps we can check if the `System.FileOfflineAvailabilityStatus` folder property is present or not?
In any case I guess totally worth to check if this also works with other cloud providers.

duke7553:
@lampenlampen @gave92 I noticed there is a way we can test this through https://docs.microsoft.com/en-us/windows/win32/properties/props-system-storageproviderid . Examples of returned values include Local, Network, OneDrive, or Computer.

Secondly, I would prefer if we ensure the root StorageFolder of each directory has a `StorageFolder.Provider != Local`, etc. before assuming we should get a IStorageItem for each item. This is because doing so has a performance penalty.

gave92:
@duke7553 Thanks for the info, a couple of questions:
- Where did you get the values of the property (Local/Onedrive/Network/...) from?
- Would this help to check whether a folder is synced to a cloud storage (different from Onedrive)?

duke7553:
@gave92 Sorry. I sent the wrong link.

Assuming these have the same return values, it really doesn't matter which one you use.

Correct Link: https://docs.microsoft.com/en-us/uwp/api/windows.storage.storageprovider.id?view=winrt-19041#Windows_Storage_StorageProvider_Id

duke7553:
_The ID that is returned will depend on your provider._ 

Additionally, I assume this way works with different providers, as they would return their own value for this. We'll need to have someone who uses another cloud provider test this for us.

gave92:
@duke7553 Ah thanks for the clarification, I thought it was a fixed enumeration value.
I guess I could check this with Google Drive at least.
But we would also have to check if the method I'm using for getting the status works with each provider.

gave92:
@lampenlampen Coud you check if the latest commit shows the sync status for both OneDrive and Synology Drive Client (without modifications)?
What were your edits? Just the lines that were comparing the `WorkingDirectory` with `OneDrivePath` or something more?

lampenlampen:
Yes it works 👍. I only edited these two lines.

yaichenbaum:
@gave92 I tested the latest changes and I like the direction it's going in, one suggestion I have is to update the status when it changes without requiring the user to refresh the directory.

gave92:
@yair100 Latest commit adds support for auto-refreshing the icon when the sync status changes.

@duke7553 Could you check if you agree with the implementation of this? To get notified of sync status changes I added the FILE_NOTIFY_CHANGE_ATTRIBUTES filter to ReadDirectoryChangesW. Do you think this could lead to an overflow of notifications (e.g in case search indexing or antivirus scanning the watched folder or something)?

duke7553:
@gave92 That is a valid concern, but would an antivirus scan actually change the attributes of an item?

gave92:
@duke7553 I was thinking about the Last Access Time property which I think is update any time the file is opened (even if contents are not changed)

yair100:
> @yair100 Latest commit adds support for auto-refreshing the icon when the sync status changes.

@gave92 It looks great!

Files/Views/LayoutModes/GenericFileBrowser.xaml
```diff
@@ -637,12 +641,34 @@
...
+                <controls:DataGridTemplateColumn
+                    x:Name="syncStatusColumn"
+                    x:Uid="syncStatusColumn"
+                    CanUserResize="False"
+                    CanUserSort="False"
+                    DisplayIndex="2"
+                    IsReadOnly="True"
+                    Header="Status"
+                    Tag="Status"
+                    Visibility="{x:Bind local:App.CurrentInstance.InstanceViewModel.IsPageTypeCloudDrive, Converter={StaticResource BoolToVisibilityConverter}, Mode=OneWay}">
```
yaichenbaum, line 653: Instead of referencing the view model from the xaml, would it make more sense to create the reference in the code behind of GenericFile Browser?

gave92: Do you mean putting in BaseLayout something like:
`public InstanceViewModel Instance => App.CurrentInstance.InstanceViewModel;`
and than doing:
`Visibility="{x:Bind Instance.IsPageTypeCloudDrive`}` in GenericFileBrowser?

duke7553: @gave92 I don't think it really matters that much in this case.

lukeblevins: @yair100 As long as they're consistent in the future

yaichenbaum:
@gave92 Have you looked at aligning the status icon with the item name in the grid layouts?

gave92:
@yair100 still have too try that. 
I was looking into the possibility to add the OneDrive specific file actions to the context menu (like "Free space", "Always keep on this device",...). Seems doable but the changes are extensive and better put in a separate PR that adds support for missing context menu items in general.
I'll try tweaking the UI as you suggest soon.

*Edit* I actually like the appearance of the status icon in the Tile View (as shown above). In Grid View it's perhaps better to put it next to the filename so it's consistent when you make the grid items bigger.

yaichenbaum:
> I was looking into the possibility to add the OneDrive specific file actions to the context menu (like "Free space", "Always keep on this device",...). Seems doable l but the changes are extensive and better put in a separate PR that adds support for missing context menu items in general.
I'll try tweaking the UI as you suggest soon.

@gave92 Looking forward to those UI tweaks, we should definitely look at adding those context menu items, but I agree it's probably better for a separate PR.

gave92:
Grid View
image
Tile View
image
@yair100 What do you think?

yaichenbaum:
@gave92 It looks great!

yair100 (approved):

lukeblevins (approved):

PR #1359, opened 2020-07-05
Fix Chinese Translation

ChihweiLHBird:


tsvietOK:
Could you add generated after this changes Resource file?

PR #1360, opened 2020-07-06
Create State class and make PasteState and AlwaysPresentCommandsState inherit the State class

ChihweiLHBird:
… inherit the State class


ChihweiLHBird:
I am a beginner and new to the community, so my code may not be the best way to do something. Please feel free to decline my PR if you have a better idea about it. ^_^

ChihweiLHBird:
> Thanks for your contribution! You certainly have the right idea in this one, but I think we could do even better in this particular change. Perhaps, we can move all of the properties into one class which derives from INotifyPropertyChanged rather than keep them separate.
> 
> For instance, is there a better ViewModel that these can be moved into?

Since each State will be instanced to an object, if we put all properties in the class, then each it is possible that each object will need a larger memory space. Unless we can change they to static...

ChihweiLHBird:
> Thanks for your contribution! You certainly have the right idea in this one, but I think we could do even better in this particular change. Perhaps, we can move all of the properties into one class which derives from INotifyPropertyChanged rather than keep them separate.
> 
> For instance, is there a better ViewModel that these can be moved into?

Oh, I found `InteractionViewModel` can be a good place for them, isn't? We don't have to instance the state but get a value of the state from InteractionViewModel in `App`. What do you think?

ChihweiLHBird:
> Thanks for your contribution! You certainly have the right idea in this one, but I think we could do even better in this particular change. Perhaps, we can move all of the properties into one class which derives from INotifyPropertyChanged rather than keep them separate.
> 
> For instance, is there a better ViewModel that these can be moved into?

After exploring everything, I still think the version in this PR is relatively good because there is not a ViewModel that can store these states cross the application (global). Am I missing anything?

duke7553:
> I found `InteractionViewModel`

InteractionViewModel is not instance-specific, meaning that ViewModel is stored as a static reference in App.cs rather than being updated with each new tab. That is why CurrentInstanceViewModel is ideal for this situation

ChihweiLHBird:
Thanks @duke7553 for giving advises. After the discussion on discord, we figured out a better way to do it, and a commit was added to make those changes.

PR #1395, opened 2020-07-11
Use the .log extension for log files instead of .txt

jaigak:


seltinusjulio (approved):
Good idea

yaichenbaum:
@Jaiganeshkumaran Is there any reason to make this change?

jaigak:
@yair100 The only reason I changed the extension was to make the log file more identifiable for non-technical users.

yaichenbaum:
@Jaiganeshkumaran Unless users are asking for this I don't think there is much of a reason to do it...

PR #1417, opened 2020-07-13
Implemented tab drag and drop functionality

soumyamahunt:
Implemented tab drag and drop functionality across app instances, dragging tab outside creates a new instance of the app.

4mvKFW6UCI

PR #1432, opened 2020-07-13
Implemented tab drag and drop functionality

### LINKED ISSUE #828: Support drag to move to another tab
One of the main uses of multiple tabs for me would be to move a file/set of files from one folder to another, in this case, tabs. 

When I drag selected files/folders to another tab, the tab should automatically switch to the one the cursor is on top of. I should then be able to put it in, which cuts the files from the previous location and pastes into the new location. Perhaps alt + drag could be used to copy instead. (or if there's a standard implementation in the default explorer, that could be used)

**Additional context**
:) Great app!

### LINKED ISSUE #1453: Make tabs more intuitive
I believe that doing the following would help make this application feel more intuitive:

- Dragging a tab out of the window opens that tab in a new window
- Dragging a tab when that is the only tab in the window moves the window instead of the tab

I think it'd be great if dragging a tab out of the window will open that tab in it's own window as users will likely do this instinctively anyways (as this is how tabs work in browsers like Chrome, Edge & Firefox).

By having it so when the user drags the tab out of the window, letting go of the tab will open that tab in a new window at the point where the user let go of the tab. There should only be one exception to this feature being that dragging a tab when it's the only tab in the window (in this case dragging the single tab) should revert to moving the window around.

For reference, open a browser like **Edge, Firefox or Chrome** and **drag a tab out of the window** when there are **multiple tabs present** and when there is only **one tab present**; that is the experience I believe that would help make this application feel **more intuitive for the end-user**.

soumyamahunt:
Implemented tab drag and drop functionality across app instances, dragging tab outside creates a new instance of the app.

gPm3JA3lLr

Also fixed opening new window by providing custom path.

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -97,9 +97,9 @@ namespace Files.Interacts
             await LockScreen.SetImageFileAsync(file);
         }
 
-        public void OpenNewTab()
+        public async void OpenNewTab()
         {
-            MainPage.AddNewTab(typeof(ModernShellPage), ResourceController.GetTranslation("NewTab"));
+            await MainPage.AddNewTab(typeof(ModernShellPage), ResourceController.GetTranslation("NewTab"));
         }
 
         public async void OpenInNewWindowItem_Click(object sender, RoutedEventArgs e)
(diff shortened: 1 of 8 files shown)


soumyamahunt:
@yair100  @duke7553 As requested #1417 is implemented here.

yair100 (approved):

yaichenbaum:
:shipit:

PR #1462, opened 2020-07-15
Open taskbar-pinned folders in existing instance

### LINKED ISSUE #1448: Open pinned folders in existing windows of F-UWP instead of creating a new one
Hi there,

**Problem:**
I often open new folders by right-clicking the explorer and opening a shortcut. One of my favourite "features" of F-UWP is that you don't open new windows for every instance, but instead have tabs like in a browser - except from when you open a pinned folder from the taskbar.

**Solution:**
I would love to have a setting which allows me to open pinned folders from the taskbar in a already existing window of Files-UWP instead of opening a new window.

image

gave92:
Fixes #1448

With this change folders pinned on the taskbar will be opened in the last used Files instance (in a new tab).
Folders launched from the command line will always open in a new instance.
Clicking the Files tile or start menu icon will always open a new instance.

TODO:
- ~add a setting to activate the new behaviour~
- [maybe] if the selected tab in the target instance is "new tab" page, open folder in that tab

2020-07-15-08-14-34-2-Trim2


soumyamahunt:
How does this work if app is minimized??

Jaiganeshkumaran:
> How does this work if app is minimized??

The app will be activated by Windows.

PR #1469, opened 2020-07-15
Fixed language options

soumyamahunt:
## 1. By default the language of the app syncs with system language. 

ApplicationFrameHost_0u48bWLEgI

## 2. "Restart Required" is only displayed when the language of app is changed. 

ApplicationFrameHost_LnxoiueGkJ
## 3.  The name of the languages are shown in their native name and script.

ApplicationFrameHost_FjFVuyvarp

### DIFF
--- a/Files/DataModels/DefaultLanguageModel.cs
+++ b/Files/DataModels/DefaultLanguageModel.cs
@@ -9,9 +9,18 @@ namespace Files.DataModels
 
         public DefaultLanguageModel(string id)
         {
-            var info = new CultureInfo(id);
-            ID = info.Name;
-            Name = info.DisplayName;
+            if (!string.IsNullOrEmpty(id))
+            {
+                var info = new CultureInfo(id);
+                ID = info.Name;
+                Name = info.NativeName;
+            }
+            else
+            {
+                ID = string.Empty;
+                var systemDefaultLanguageOptionStr = ResourceController.GetTranslation("SettingsPreferencesSystemDefaultLanguageOption");
+                Name = string.IsNullOrEmpty(systemDefaultLanguageOptionStr) ? "System Default" : systemDefaultLanguageOptionStr;
+            }
         }
 
         public override string ToString()
--- a/Files/MultilingualResources/Files.de-DE.xlf
+++ b/Files/MultilingualResources/Files.de-DE.xlf
@@ -993,6 +993,10 @@
           <source>Remove Page</source>
           <target state="new">Remove Page</target>
         </trans-unit>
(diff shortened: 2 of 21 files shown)


Jaiganeshkumaran:
How did you get all the supported language? Did you install them from Windows Settings?

soumyamahunt:
> How did you get all the supported language? Did you install them from Windows Settings?

`Windows.Globalization.ApplicationLanguages.ManifestLanguages` will do the trick.

> Instead of System Default use Windows default and don't make the D capital.

Ok.

Jaiganeshkumaran:
@soumyamahunt It doesn't work that way. Windows.Globalization.ApplicationLanguages.ManifestLanguages will return the list of supported languages however you can only set the language if it's installed. By default, only resource files for the languages in your language list would be downloaded by the Microsoft Store. Setting a language that is not in the language list won't work. You can try it for yourself.

tsvietOK:
@soumyamahunt Please, check "Restart required" text visibility on light theme.
@Jaiganeshkumaran All languages are visible only in debug mode, when you build in release mode, only installed languages will be visible

soumyamahunt:
> @soumyamahunt It doesn't work that way. Windows.Globalization.ApplicationLanguages.ManifestLanguages will return the list of supported languages however you can only set the language if it's installed. By default, only resource files for the languages in your language list would be downloaded by the Microsoft Store. Setting a language that is not in the language list won't work. You can try it for yourself.

`Windows.Globalization.ApplicationLanguages.ManifestLanguages` will return all the supported languages that are currently installed with the app. If during packaging of the app “Create App bundle” isn't set to "Never", when published to store Windows will only install the language which are currently installed by user in Windows settings. If devs want to make all the supported languages available to user they can just set “Create App bundle” to "Never". In any case, if you are wondering there is any chance of error there won't be any exception thrown.

> @soumyamahunt Please, check "Restart required" text visibility and readability on light theme.

Will push a fix on that.

> @Jaiganeshkumaran All languages are visible only in debug mode, when you build in release mode, only installed languages will be visible

Not true actually, you can set “Create App bundle” to "Never" to show all supported languages even that are not installed in Windows settings.

Jaiganeshkumaran:
> All languages are visible only in debug mode, when you build in release mode, only installed languages will be visible

> Not true actually, you can set “Create App bundle” to "Never" to show all supported languages even that are not installed in Windows settings.

But we shouldn't do that since it will increase the download size.

yaichenbaum:
> How did you get all the supported language? Did you install them from Windows Settings?

When you build the app from the source you will see all the languages.

PR #1478, opened 2020-07-16
Added relative path support and environmental variable support to NavigationToolBar

soumyamahunt:
Travelling to folders with relative path are now allowed also added support for entering any environment variables in NavigationToolBar

### DIFF
--- a/Files/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
+++ b/Files/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
@@ -126,7 +126,17 @@ namespace Files.Filesystem
                     return new StorageFolderWithPath(folder, path);
                 }
             }
-            return new StorageFolderWithPath(await StorageFolder.GetFolderFromPathAsync(value));
+
+            try
+            {
+                var folder = await parentFolder.Folder.GetFolderAsync(value);
+                var path = Path.Combine(parentFolder.Folder.Path, value);
+                return new StorageFolderWithPath(folder, path);
+            }
+            catch
+            {
+                return new StorageFolderWithPath(await StorageFolder.GetFolderFromPathAsync(value));
+            }
         }
 
         public async static Task<StorageFolder> GetFolderFromPathAsync(string value, StorageFolderWithPath rootFolder = null, StorageFolderWithPath parentFolder = null)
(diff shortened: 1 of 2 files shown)


Files/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
```diff
@@ -126,7 +126,17 @@ public async static Task<StorageFolderWithPath> GetFolderWithPathFromPathAsync(s
                     return new StorageFolderWithPath(folder, path);
                 }
             }
-            return new StorageFolderWithPath(await StorageFolder.GetFolderFromPathAsync(value));
+
+            try
+            {
+                var folder = await parentFolder.Folder.GetFolderAsync(value);
```
gave92, line 132: @soumyamahunt could you explain me the purpose of this change?
`var folder = await parentFolder.Folder.GetFolderAsync(value);` will I think always throw either `NullPointerException` exception or `ArgumentException` beacuse either `parentFolder` is null or `value` is a full path so `GetFolderAsync` fails.

soumyamahunt: If `value` is a relative path to the `parentfolder.Folder` then it will succeed otherwise fail. That's why I put it in a try/catch block.

gave92: @soumyamahunt And when will a relative path be passed to the function?

soumyamahunt: Suppose you are currently in Users folder in files-uwp, and there is a Test folder in your Documents folder, you can type `Documents\Test` in your navigationbar to go into that Test folder. Also, in future you can pass relative path as command line arguments too and use this(not implemented yet).

gave92: @soumyamahunt Thanks for the clarification (can you already type a relative path in the navigationbar or is this still to be implemented too)?
In any case this made me realize a small bug (`parentfolder` will be always be null when called from `ItemViewModel.LoadExtendedItemProperties()`)

soumyamahunt: Yes, currently you can type relative path in navigationbar, for other cases it will just handle the error and default to `StorageFolder.GetFolderFromPathAsync`.

PR #1482, opened 2020-07-16
Better exception handling and tracking

soumyamahunt:
Added tracking of unhandled and unobserved exceptions.

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -26,6 +26,7 @@ using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Media.Animation;
 using Windows.UI.Xaml.Navigation;
+using System.Threading.Tasks;
 
 namespace Files
 {
@@ -56,14 +57,15 @@ namespace Files
 
         public App()
         {
+            UnhandledException += OnUnhandledException;
+            TaskScheduler.UnobservedTaskException += OnUnobservedException;
+
             InitializeComponent();
             Suspending += OnSuspending;
             LeavingBackground += OnLeavingBackground;
             // Initialize NLog
             Windows.Storage.StorageFolder storageFolder = Windows.Storage.ApplicationData.Current.LocalFolder;
             NLog.LogManager.Configuration.Variables["LogPath"] = storageFolder.Path;
-
-            RegisterUncaughtExceptionLogger();
 
 #if !DEBUG
             AppCenter.Start("682666d1-51d3-4e4a-93d0-d028d43baaa0", typeof(Analytics), typeof(Crashes));
@@ -128,14 +130,6 @@ namespace Files
             // Complete the deferral so that the platform knows that we're done responding to the app service call.
             // Note for error handling: this must be called even if SendResponseAsync() throws an exception.
             messageDeferral.Complete();
-        }
-
-        private void RegisterUncaughtExceptionLogger()
-        {
-            UnhandledException += (sender, args) =>
-            {
-                Logger.Error(args.Exception, args.Message);
-            };
         }
 
         private void CoreWindow_PointerPressed(CoreWindow sender, PointerEventArgs args)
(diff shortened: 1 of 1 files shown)


yaichenbaum:
@soumyamahunt Great work! In the future, try to make sure your pull requests are linked to approved issues in order to prevent duplicate work 🙂

Jaiganeshkumaran:
@soumyamahunt IIRC, older versions of Files UWP would show a BSOD-like page when an unhandled exception is thrown. Could you make something similar?

yaichenbaum:
> @soumyamahunt IIRC, older versions of Files UWP would show a BSOD-like page when an unhandled exception is thrown. Could you make something similar?

That is a little out of scope for this PR, however it would be great to add in the future.

Files/App.xaml.cs
```diff
@@ -394,6 +388,64 @@ private void OnSuspending(object sender, SuspendingEventArgs e)
...
+                $"Message: {e.Message}, " +
+                $"InnerException: {e.Exception?.InnerException}, " +
+                $"InnerExceptionMessage: {e.Exception?.InnerException?.Message}",
+                "UnhandledException");
+
+            Analytics.TrackEvent("OnUnhandledException", diagnosticInfo);
+            Crashes.TrackError(e.Exception, diagnosticInfo, attachment);
+
+            // suppress and handle it manually.
+            e.Handled = true;
```
tsvietOK, line 418: Don't suppress crash. We should fix app crash by fixing bugs instead of handling all possible exceptions. This PR should add only extended crash log info.

Jaiganeshkumaran: Right. If we handled all the exception it will make it harder to find bugs. Let it crash so that we can know that there's an issue without seeing logs.

soumyamahunt: I am making changes so that app will crash only in debug version and not in release version. Is that acceptable??

tsvietOK: No, it is not acceptable. If smth wrong, app should crash even in release mode.

soumyamahunt: Testers and normal users shouldn't experience crashes if something goes wrong, that is why I added the handling logic. @yair100 @duke7553  What do you guys think??

yaichenbaum: @soumyamahunt If something is wrong, we would like to get the extended crash info logged, however, we should still let the app crash so that we get the report and crash error rate info in AppCenter and the Partner Center dashboard.

soumyamahunt: @yair100 I am using `AppCenter.Crashes.CrashError` to upload error report to AppCenter dashboard. Also, I am using `AppCenter.Analytics.TrackEvent` too. App doesn't need to crash to get error rate info in AppCenter and the Partner Center dashboard.

yaichenbaum: @soumyamahunt Tracking the error rate is very important in order to keep track of the amount of errors across versions. Please address the changes @tsvietOK requested 🙂.

soumyamahunt: > @soumyamahunt Tracking the error rate is very important in order to keep track of the amount of errors across versions. Please address the changes @tsvietOK requested 🙂.

Addressed those changes.

Files/App.xaml.cs
```diff
@@ -394,6 +388,48 @@ private void OnSuspending(object sender, SuspendingEventArgs e)
...
+        {
+            Logger.Error(e.Exception, e.Exception.Message);
+
+            var appVersion = Package.Current.Id.Version;
+
+            var diagnosticInfo = new Dictionary<string, string>()
+            {
+                { "Message", e.Exception?.Message },
+                { "Exception", e.Exception?.ToString() },
+                { "InnerException", e.Exception?.InnerException?.ToString() },
```
hez2010, line 425: InnerException itself may have its inner exception, I think this should be expanded recursively.

soumyamahunt: Extended log will be available in AppCenter dashboard, hence it is not needed.

Files/App.xaml.cs
```diff
@@ -394,6 +388,48 @@ private void OnSuspending(object sender, SuspendingEventArgs e)
...
+                { "Exception", e.Exception?.ToString() },
+                { "Culture", SystemInformation.Culture.EnglishName },
+                { "AvailableMemory", SystemInformation.AvailableMemory.ToString("F0") },
+                { "FirstUseTimeUTC", SystemInformation.FirstUseTime.ToUniversalTime().ToString("MM/dd/yyyy HH:mm:ss") },
+                { "OSArchitecture", SystemInformation.OperatingSystemArchitecture.ToString() },
+                { "OSVersion", SystemInformation.OperatingSystemVersion.ToString() },
+                { "AppVersion", string.Format("{0}.{1}.{2}.{3}", appVersion.Major, appVersion.Minor, appVersion.Build, appVersion.Revision) }
+            };
+
+            Analytics.TrackEvent("OnUnhandledException", diagnosticInfo);
```
yaichenbaum, line 410: Instead of logging the crash as an event in AppCenter, perhaps it's better to save this in the local log file. AppCenter already catches the crash info and logging it as an event will make tracking our other analytics harder.

soumyamahunt: Made changes.

PR #1485, opened 2020-07-17
Different layout mode and sorting for each folder

### LINKED ISSUE #530: Different layout mode and sorting for each folder
I would like to be able to use different layout modes and sorting in each folder, like in file explorer.

Jaiganeshkumaran:
Implements and closes #530


soumyamahunt:
I think it would be a good idea to keep two values for sorting and layout mode each, one for app wide and one for current directory. Also, when changing sorting or layout mode, user would be given option whether to save for directory only while default being app wide change.

Jaiganeshkumaran:
> I think it would be a good idea to keep two values for sorting and layout mode each, one for app wide and one for current directory. Also, when changing sorting or layout mode, user would be given option whether to save for directory only while default being app wide change.

I don't think that would be a good idea since the user can change the sort mode just by clicking on a category on the DataGrid and showing a dialog will not be a good experience. Instead, we can have a switch in the settings page to make sorting and layout mode app wide or specific to each directory,

soumyamahunt:
> I don't think that would be a good idea since the user can change the sort mode just by clicking on a category on the DataGrid and showing a dialog will not be a good experience.

Instead of showing dialog, what my thinking was, instead of showing flyout for changing layout and sorting, we can show a teaching tip for changing these options. The bottom of the teaching tip would be a checkbox checking which saves the sorting and layout option for the directory. Kind of like this:

soumyamahunt:
Also isn't there size limit for `ApplicationDataContainer`??

yaichenbaum:
> I think it would be a good idea to keep two values for sorting and layout mode each, one for app wide and one for current directory. Also, when changing sorting or layout mode, user would be given option whether to save for directory only while default being app wide change.

Perhaps we can add a setting to preferences later on, however let's keep the scope of this pull request limited to what's already done.

Jaiganeshkumaran:
@tsvietOK Are you trying this on a root directory? It works on other directories for me.

Jaiganeshkumaran:
> Also isn't there size limit for `ApplicationDataContainer`??

There is a limit on how much data you can store per key but no there's no overall limit as far as I know.

tsvietOK:
@Jaiganeshkumaran Changing layout type should work for any directory.

soumyamahunt:
> There is a limit on how much data you can store per key but no there's no overall limit as far as I know.

I could only find limit on key length and key data limit, couldn't find any limit on for `ApplicationDataContainer`.

Jaiganeshkumaran:
> @Jaiganeshkumaran Changing layout type should work for any directory.

This issue has been fixed.

Jaiganeshkumaran:
> Don't working at all. When i change layout type for one folder, it changes layout type for all folders.

I've made some changes and it should work now.

tsvietOK:
@Jaiganeshkumaran Fix build, please.

tsvietOK:
@Jaiganeshkumaran Can't change layout type again. Please, properly test your changes!

Jaiganeshkumaran:
@tsvietOK For some reason, I can't run this on Windows 10 version 1903.

Jaiganeshkumaran:
Now I'm on my 2004 PC and I am going to test.

tsvietOK:
@Jaiganeshkumaran Can't open folder through sidebar.
image
Translated exception: "invalid path specified"

Jaiganeshkumaran:
@tsvietOK It should work now.

tsvietOK:
App still crashing, but now without any exceptions.

Jaiganeshkumaran:
Should we remember the grid view size as well?

Jaiganeshkumaran:
Doesn't work

coolmyll:
This still doesnt work. Changing the layout or sort changes it for all folders.

Issue #1409, opened 2020-07-12
Properties window not closing when pressing "Esc" key.


osieltorres:
**Describe the bug**
Currently the file or folder Properties window is not closed when pressing "Esc" key on keyboard, although this is a natural behaviour in default windows explorer.

**To Reproduce**
1. Go to any file or folder.
2. Right click and go to menu Properties option .
3. Press the "Esc" key to close the window.
4. The Properties window does not close.

**Expected behavior**
The Properties window to close when "Esc" key is pressed.

**Desktop (please complete the following information):**
 - OS Version: [e.g. Windows 10 2004 19041.329]
 - App version: [v0.11.0.0]

XPoppyX:
@yaichenbaum

yaichenbaum:
This issue has been fixed with #1504 and should roll out to users in a couple of weeks.

PR #1504, opened 2020-07-19
Added the ability to use the escape key to exit the properties window

winiston2212:
Added the ability to use the escape key to exit the file properties window as requested in issue #1409.

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -535,14 +535,25 @@ namespace Files.Interacts
         {
             if (ApiInformation.IsApiContractPresent("Windows.Foundation.UniversalApiContract", 8))
             {
+
                 AppWindow appWindow = await AppWindow.TryCreateAsync();
                 Frame frame = new Frame();
+                frame.KeyDown += Frame_KeyDown;
                 appWindow.TitleBar.ExtendsContentIntoTitleBar = true;
                 frame.Navigate(typeof(Properties), item, new SuppressNavigationTransitionInfo());
+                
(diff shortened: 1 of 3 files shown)


Files/Interacts/Interaction.cs
```diff
@@ -535,15 +535,26 @@ private async Task OpenPropertiesWindow(object item)
...
+                frame.KeyDown += Frame_KeyDown;
                 appWindow.TitleBar.ExtendsContentIntoTitleBar = true;
                 frame.Navigate(typeof(Properties), item, new SuppressNavigationTransitionInfo());
+                
                 WindowManagementPreview.SetPreferredMinSize(appWindow, new Size(400, 500));
 
                 appWindow.RequestSize(new Size(400, 475));
                 appWindow.Title = ResourceController.GetTranslation("PropertiesTitle");
 
+                async void Frame_KeyDown(object sender, KeyRoutedEventArgs e)
```
tsvietOK, line 550: I think it is better to do it inside `Properties.xaml.cs` as it is n't a part of interaction.

Files/Views/Pages/Properties.xaml.cs
```diff
@@ -131,5 +133,13 @@ private async void Button_Click(object sender, RoutedEventArgs e)
                 propertiesDialog.Hide();
             }
         }
+
+        async void Page_KeyDown(object sender, KeyRoutedEventArgs e)
+        {
+            if (e.Key.Equals(VirtualKey.Escape))
+            {
+                await propWindow.CloseAsync();
```
tsvietOK, line 141: Add check for `UniversalApiContract`

Issue #1379, opened 2020-07-09
Drag A Folder And Drop It On Itself Cause Error


ChihweiLHBird:
Dragging and dropping a folder on itself do not have to cause the pop up error message.

PR #1506, opened 2020-07-19
Cancel Action Without Error Message When Drag One Or More Folders and Drop on itself (themselves) 

ChihweiLHBird:
Attempted to fix issue #1379.
Make it do nothing when drag a folder and drop it on itself and when drag some files and folders and drop them on one of the dragged folder.

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -508,9 +508,27 @@ namespace Files
 
         protected async void Item_Drop(object sender, DragEventArgs e)
         {
+            IReadOnlyList<IStorageItem> draggedItemsList = await e.DataView.GetStorageItemsAsync();
             e.Handled = true;
             ListedItem rowItem = GetItemFromElement(sender);
-            await App.CurrentInstance.InteractionOperations.PasteItems(e.DataView, rowItem.ItemPath, e.AcceptedOperation);
+            bool droppedOnDraggedItem = false;
+            foreach(IStorageItem draggedItem in draggedItemsList)
+            {
+                if (rowItem.ItemPath == draggedItem.Path)
+                {
+                    droppedOnDraggedItem = true;
+                    break;
+                }
+                    
+            }
+            if (!droppedOnDraggedItem)
+            {
+                await App.CurrentInstance.InteractionOperations.PasteItems(e.DataView, rowItem.ItemPath, e.AcceptedOperation);
+            }
+            else
+            {
+                e.DataView.ReportOperationCompleted(e.AcceptedOperation);
+            }
         }
 
         protected void InitializeDrag(UIElement element)
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -519,7 +519,6 @@ namespace Files
                     droppedOnDraggedItem = true;
                     break;
                 }
-                    
             }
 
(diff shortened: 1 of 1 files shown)


Files/BaseLayout.cs
```diff
@@ -508,9 +508,27 @@ protected async void Item_DragOver(object sender, DragEventArgs e)
...
+            }
+
+            if (!droppedOnDraggedItem)
+            {
+                await App.CurrentInstance.InteractionOperations.PasteItems(e.DataView, rowItem.ItemPath, e.AcceptedOperation);
+            }
+            else
+            {
+                e.DataView.ReportOperationCompleted(e.AcceptedOperation);
+            }
```
soumyamahunt, line 531: Instead of making changes here, make changes in `PasteItems` function. What you are doing is canceling the hole operation when one folder matches the path. Although, this is fine for now since only one item drag-drop is supported but that might change in future with multiple items drag-drop.

tsvietOK: It is OK to filter dragged items in Drop method, as it is part of it. PasteItems method should be responsible only for pasting items. 
We are already supporting drag-n-drop of several files.
It is OK to cancel the whole operation if dragged items which contains folder trying to paste to this folder itself

ChihweiLHBird: > Instead of making changes here, make changes in `PasteItems` function. What you are doing is canceling the hole operation when one folder matches the path. Although, this is fine for now since only one item drag-drop is supported but that might change in future with multiple items drag-drop.

If we handle this situation in the `PasteItems`function, then it will pops up an error message and I think that's not what we want. Dragging an item and dropping it on itself means the user want to cancel this operation and put the item back. By the way, Files UWP has supported multiple files drag and drop for a while.

soumyamahunt: Just checked in Windows Explorer, it handles changes in DragOver method, handling changes in DragOver method will be a better idea.

ChihweiLHBird: > Just checked in Windows Explorer, it handles changes in DragOver method, handling changes in DragOver method will be a better idea.

Okay, but may I know how can I handle it in DragOver method? The DeagOver method seems doesn't start any action like move or copy the dragged items. Thank you.

soumyamahunt: You can use the same logic to check any of the storage item's path matches the path of folder on which user trying to drop, if any match is abort the action. You have to do this before `DragEventArgs.AcceptedOperation` is set to any value. Doing that will no more fire the drop event.

ChihweiLHBird: Just get it done, thank you very much for the hint. Can you help me to review it again? Thanks!

ChihweiLHBird: > You can use the same logic to check any of the storage item's path matches the path of folder on which user trying to drop, if any match is abort the action. You have to do this before `DragEventArgs.AcceptedOperation` or `DragEventArgs.Handled` is set to any value. Doing that will no more fire the drop event.

Just get it done, thank you for your hint. Can you please help me to review it again? Thanks!

soumyamahunt: > Just get it done, thank you for your hint. Can you please help me to review it again? Thanks!

Just tested, works fine for me.

Issue #634, opened 2020-04-21
Drag File/Folder To A Previous Directory In Path To Move


eraychumak:
**Is your feature request related to a problem? Please describe.**
Cannot move a file/folder to a previous directory within the current path.

**Describe the solution you'd like**

Example of windows file explorer:

![image](https://user-images.githubusercontent.com/44278658/79837098-d9cc2300-83a8-11ea-8ed7-e280778f5312.png)

Current result of Files UWP:

![image](https://user-images.githubusercontent.com/44278658/79837275-1009a280-83a9-11ea-858d-1b06bcfefa3a.png)

**Describe alternatives you've considered**
...

**Additional context**
...

yaichenbaum:
This has been implemented with #1507.

Issue #1052, opened 2020-06-13
Sidebar items' pressed effect persists when released while outside of control


XeZrunner:
**Describe the bug**
Holding a menu item in the sidebar, then moving the cursor/pointer outside of the control and letting go will result in the highlight effect persisting until the control is pressed again.

**To Reproduce**
1. Open the app.
2. Click or tap, and keep holding an item in the left sidebar.
3. Move the mouse or your finger/pointer outside of the menu item control.
4. Let go of the mouse button or your finger/pointer.

**Expected behavior**
The highlight effect should not persist when letting go of it.

**Screenshots**
![A GIF showing off the issue.](https://i.imgur.com/EJVmTL5.gif)

**Desktop (please complete the following information):**
 - OS Version: Windows 10 2004 19041.329
 - App version: v0.9.2.0

**Additional context**
This issue is also present within the Settings sidebar.
This issue does not occur with other controls in the application.

duke7553:
This is reproducible in v0.10.2

tsvietOK:
This is Navigation View bug, same bug exists in XAML Controls gallery app.

yaichenbaum:
@tsvietOK Is there an issue on their repo?

tsvietOK:
@yair100 As mentioned [here](https://github.com/microsoft/microsoft-ui-xaml/issues/3085#issuecomment-671054773) this bug should be resolved in pre-release version of microsoft-ui-xaml library.

yaichenbaum:
Thank you for the update, I am closing this issue since it will be fixed in the new release of WinUI.

PR #1507, opened 2020-07-19
Added drop support to the NavigationBar

soumyamahunt:
As requested in #634, you can drop files/folders in navigationbar items to move that file/folder to that path. Demo:

ILBqCV5Bkx


soumyamahunt:
Currently, after I drop on an item the item is highlighted, I am working on a fix for it. May be related to #1052.

soumyamahunt:
> When moving file to the same directory it create a new file with the same name with added '- copy". I think this operation should not be available.

Fixed that, already existing items are skipped.

tsvietOK:
I am still able to drop item to last navigation bar path item

soumyamahunt:
> I am still able to drop item to last navigation bar path item

Does dropping create a duplicate file as you stated before??

tsvietOK:
No, it is doing nothing. But text "Move to" still exists, so it could mislead user.
image

soumyamahunt:
> No, it is doing nothing. But text "Move to" still exists, so it could mislead user.
> image

That is the default Windows Explorer behavior, when you drag items to any path items it displays the move to text.
explorer_qweO4voqXM
May be I should add a dialog to user explaining the move operation cannot complete.

tsvietOK:
Or just remove this operation from last path item? If it is not possible, add a dialog.

soumyamahunt:
> Or just remove this operation from last path item? 

The reason I am deciding against that is because user might want to drop files to the pathitems from another instance or another app in case if their gridview is blocked by another window. This is the default behavior in Windows Explorer.
> If it is not possible or too hard, add a dialog.

I tried to add a dialog but the method failed with `RPC_E_WRONG_THREAD` exception. Queuing the task in ui thread using dispatcher also failed. Similar behavior implemented for dragging a parent folder to child folder also fails. So I have disabled both methods for now until a solution is found. 

Currently, I have implemented to check if item already exist in the folder in DragOver method.

soumyamahunt:
> If check of an item already exist in the folder now implemented, user dialog is no longer needed. Please, remove it.

I have commented out the dialog part, if any workaround for the issue is found, then it can be reenabled.

Files/Interacts/Interaction.cs
```diff
@@ -1081,12 +1082,47 @@ public async Task PasteItems(DataPackageView packageView, string destinationPath
...
+                    }
+
                     if (item.IsOfType(StorageItemTypes.Folder))
                     {
                         if (!string.IsNullOrEmpty(item.Path) && destinationPath.IsSubPathOf(item.Path))
                         {
                             ImpossibleActionResponseTypes responseType = ImpossibleActionResponseTypes.Abort;
-                            Binding themeBind = new Binding();
+
+                            /// Currently following implementation throws exception until it is resolved keep it disabled
```
tsvietOK, line 1124: What exactly throws an exception? What type of exception? Why there is no issue about it? How to reproduce it? If there is no issue, please bring it back.

Files/Interacts/Interaction.cs
```diff
@@ -1081,12 +1082,47 @@ public async Task PasteItems(DataPackageView packageView, string destinationPath
             {
                 foreach (IStorageItem item in itemsToPaste)
                 {
+                    if (!copyDuplicate && !string.IsNullOrEmpty(item.Path) &&
```
tsvietOK, line 1085: Remove this block, as long as this dialog is not used.

Files/Interacts/Interaction.cs
```diff
@@ -1048,7 +1048,8 @@ public async void PasteItem_ClickAsync(object sender, RoutedEventArgs e)
             await PasteItems(packageView, destinationPath, packageView.RequestedOperation);
         }
 
-        public async Task PasteItems(DataPackageView packageView, string destinationPath, DataPackageOperation acceptedOperation)
+        public async Task PasteItems(DataPackageView packageView, string destinationPath, DataPackageOperation acceptedOperation,
+            bool copyDuplicate = true)
```
tsvietOK, line 1052: Remove `copyDuplicate`

Files/Interacts/Interaction.cs
soumyamahunt: Created an issue subjecting that in #1545.

Files/Interacts/Interaction.cs
soumyamahunt: I created that block to address #1546, but since this isn't the scope of the pr I will remove it for now and address in a next pr.

Files/Interacts/Interaction.cs
soumyamahunt: I created tit to address #1546, but since this isn't the scope of the pr I will remove it for now and address in a next pr.

soumyamahunt: Made requested changes.

Issue #672, opened 2020-04-24
Ability to resize the sidebar


XeZrunner:
**Is your feature request related to a problem? Please describe.**
It isn't related to a problem.

**Describe the solution you'd like**
When hovering between the sidebar and content edge, the cursor should change into the horizontal resizing cursor and let us change the sidebar width. The application should also save this width so we don't have to keep adjusting the sidebar width every time.

**Describe alternatives you've considered**
Increase the default sidebar width, to match Fluent Design / Windows guidelines.

**Additional context**
[Here's a quick mockup done in Paint.](https://i.imgur.com/au7iSLE.png)

chuchanpr:
Add option to collapse the side bar, for cleaner look

yaichenbaum:
This has been implemented with #1512.

PR #1512, opened 2020-07-19
Added the ability to resize the sidebar

soumyamahunt:
This pr tries to implement feature requested in #672.


soumyamahunt:
Currently sidebar can only be resized up to certain width.

yaichenbaum:
> Currently sidebar can only be resized up to certain width.

@soumyamahunt That would be the expected behavior. Does the app keep track of the size across sessions?

soumyamahunt:
> > Currently sidebar can only be resized up to certain width.
> 
> @soumyamahunt That would be the expected behavior. Does the app keep track of the size across sessions?

No not now, I will add it.

duke7553:
@soumyamahunt If you need any inspiration, check out the appropriate branch for my past, unfinished work on this. Looks really good so far.

soumyamahunt:
@yair100 @duke7553 what should we keep minimum and maximum sidebar width??

yair100:
> @yair100 @duke7553 what should we keep minimum and maximum sidebar width??

I would recommend keeping the current size as the minimum.

jaigak:
@yair100 To match Fluent Design Guidelines, the minimum width must be at least 20% of the container.

Jaiganeshkumaran:
@soumyamahunt Could you share a screenshot on the changes?

soumyamahunt:
> @soumyamahunt Could you share a screenshot on the changes?

Here is how it behaves with a minimum and maximum value.
R7Mjnsy4Dx

soumyamahunt:
@yair100, @duke7553 minimum width is kept at 200(as was before) maximum is kept at 500. Also, width is remembered across instances.

R7Mjnsy4Dx

Jaiganeshkumaran:
Can you increase the minimum width to match the default WinUI styles?

soumyamahunt:
> Can you increase the minimum width to match the default WinUI styles?

I can but @yair100 suggested to keep it same as before, @yair100 what do you think of this?? Also what is the default width value in WinUI style??

duke7553:
@soumyamahunt Nope, we're happy with the existing minimum.

Files/Strings/de-DE/Resources.resw
```diff
@@ -666,4 +723,43 @@
...
+  </data>
+  <data name="SettingsOnStartupLaunchNewTab.Content" xml:space="preserve">
+    <value>Ordner in neuem Tab öffnen</value>
+  </data>
+  <data name="SettingsPreferencesSystemDefaultLanguageOption" xml:space="preserve">
+    <value>Windows standard</value>
+  </data>
+  <data name="TabStripDragAndDropUIOverrideCaption" xml:space="preserve">
+    <value>Tab hierhin verschieben</value>
+  </data>
```
yaichenbaum, line 764: Please revert the changes to the strings.

Files/Views/ModernShellPage.xaml.cs
```diff
@@ -26,6 +26,26 @@ public sealed partial class ModernShellPage : Page, IShellPage
...
+                }
+
+                return new GridLength(sidebarWidth, GridUnitType.Pixel);
+            }
+            set
+            {
+                ApplicationData.Current.LocalSettings.Values["SidebarWidth"] = value.Value;
+            }
+        }
+
```
yaichenbaum, line 48: Please use this format for saving the settings.
```
public bool OpenASpecificPageOnStartup
{
        get => Get(false);
        set => Set(value);
}
```

Jaiganeshkumaran:
> @soumyamahunt Nope, we're happy with the existing minimum.

Now the settings sidebar looks out of place. Could you use the same size for both NavigationView controls and also make Setting sidebar resizable?

Files/Views/ModernShellPage.xaml.cs
soumyamahunt: Made changes.

soumyamahunt:
> Now the settings sidebar looks out of place. Could you use the same size for both NavigationView controls and also make Setting sidebar resizable?

I don't think making settings sidebar resizable gonna improve any usability since the strings used there are static, but I can make it sync with the main sidebar. @yair100, is this something you would like too??

yaichenbaum:
The settings sidebar should not be resizable.

yair100 (approved):

PR #1522, opened 2020-07-20
Auto complete folder names when typing in the navigation bar

### LINKED ISSUE #1511: Address bar improvements - Auto suggest + shortcuts
**Proposed features**
It would be nice if when typing in the navigation bar it would suggest folder names to complete the path.
Also the navigation bar should go into editable mode when pressing "Ctrl + L" (or "Alt + D" for english keyboards)

Would you consider something like the following?

ezgif com-gif-maker

gave92:
Implements and closes #1511

- Auto complete folder names when typing in the navigation bar
- Navigation bar goes into editable mode when pressing "Ctrl + L" (or "Alt + D" for english keyboards)
- Fix an issue of current navigation bar that causes a crash on android/mtp devices

ezgif com-gif-maker


yaichenbaum:
@gave92 It looks like switching to the AutoSuggestBox got rid of the rounded corners, would you be able to add them back?

yair100 (dismissed):

tsvietOK (changes requested):
Nothing happens when clicking on AutoSuggestBox item. Only Enter button works fine.
When typing C: it suggests relevant folders.
image
When typing V: (where is debug version of this app is actually installed) it suggests folders which is inside app folder.
image
The same scenario is possible when release version of the app will be installed on drive C:

gave92:
@yair100 Meaning rounded corners around suggestbox/textbox?
I still have rounded corners on top, but not on bottom when the suggestion list is open (which i guess shows continuity between suggestbox and suggestion list)

Below what I'm getting (old vs new) and a show of my considerable artistic abilities :)
If you're getting the same, should I add corners on bottom too?

Cattura

@tsvietOK Thanks for the feedbacks, should be able to address them soon

soumyamahunt:
Does this support environment variables too??

yaichenbaum:
@gave92 What if the auto suggest box had the same background color as the dropdown list?

gave92:
@tsvietOK both points above should be addressed
@soumyamahunt great suggestion, it does now!
@yair100 I'll try that, but perhaps the acrylic effect is "too much" when typing/with the blinking cursor and affects readability

yaichenbaum:
@gave92 I would remove the acrylic background in this case.

tsvietOK:
@yair100 I like acrylic background in AutoSuggestBox.

yair100 (approved):

tsvietOK (approved):

yaichenbaum:
:shipit:

PR #1531, opened 2020-07-22
Updated translation.

KiriKirby:


Jaiganeshkumaran:
Please change the title to include "Chinese (Simplified)". Always mention the language name when updating translations.

PR #1549, opened 2020-07-23
Fixed tab title not showing correctly in some cases

### LINKED ISSUE #1548: The folder path in the tab title is wrong when folders opened in a new tab
**Describe the bug**
When right clicking on any folder and opening it in new tab before clicking on the tab, the tab displays path for the parent of the folder.

**To Reproduce**
1. Open files-uwp.
2. Right click on any folders except root folders(i.e. C:\).
3. Click on "Open in new tab" in context flyout.
4. Open the Tabview flyout.
5. Check the title of the tab you recently opened.
6. See error.

**Expected behavior**
The tab should display either full path or the name of the folder.

 - OS Version:  Windows 10 1903 18362.239
 - App version: v0.12.0

**Additional context**
Add any other context about the problem here. Does this problem occur again after restarting the app?
This issue doesn't crash the app, but it persists across restarts. Also, clicking on the tab changes the title to the name of the folder.

soumyamahunt:
Fixes #1548.


tsvietOK:
image
Bug still don't fixed.

soumyamahunt:
> image
> Bug still don't fixed.

The path displayed on the Tabtitle not the absolute path for the directory??

tsvietOK:
What do you mean? My screenshot describes the problem still exists.

soumyamahunt:
> What do you mean? My screenshot describes the problem still exists.

The fix I have pushed will show the full path of the terminal tab until it is opened, then it switches to the folder title. So can you describe if you can see the full path as title of the tab??

tsvietOK:
It is better to show screenshots before/after.  Until tab is opened it anyway should display in tab title correct folder name, not full path.

soumyamahunt:
> It is better to show screenshots before/after.

| Before tab loaded      |         After tab loaded |
| :--------------------------------- | :-------------------- |
| ApplicationFrameHost_lul8lATNjQ | ApplicationFrameHost_9UAbtTpYxr |

> Until tab is opened it anyway should display in tab title correct folder name, not full path.

After tab is loaded the title is synonymous with folder name, hence I opted for full folder path so that user can distinguish which tabs has been loaded or not. Now that I think about it, it would be better to use accent color to indicate which tabs hasn't been loaded yet. What do you think about this??

tsvietOK:
As i mentioned before, it doesn't matter tab has been loaded or not. Tab title should contain only folder name without full path.

soumyamahunt:
> As i mentioned before, it doesn't matter tab has been loaded or not. Tab title should contain only folder name without full path.

Made changes to reflect that.

PR #1563, opened 2020-07-24
Update string resources [pl-PL]

d2dyno1:
Some strings were fixed and not translated ones were translated.


tsvietOK:
Why there is two translation files(Files.pl.xlf and Files.pl-PL.xlf)?

d2dyno1:
> Why there is two translation files(Files.pl.xlf and Files.pl-PL.xlf)?

Sorry, I didn't notice the second file (Files.pl.xlf). It's been removed now.

Issue #632, opened 2020-04-21
Improve the Speed and Reliability of Storage Item Operations


duke7553:
**Is your feature request related to a problem? Please describe.**
The cut, copy, and paste operations aren't as fast as we'd like.

**Describe the solution you'd like**
Use the Win32 *FromApp APIs available to UWP apps.

**Describe alternatives you've considered**
N/A

**Additional context**
TBA

duke7553:
Notes: Someone really needs to take a good look at improving the reliability of these operations especially when multitasking with different tabs.

I consider this issue a prerequisite for the Status UI.

lorisobi:
When I move some bigger files and switch the tab files uwp crashes

duke7553:
@lorisobi Is this on the latest preview? Also, can you upload your debug log so I can look into this?

d2dyno1:
Are all .dll functions imports supported in Files? ("Kernel32.dll", "win32.dll")

duke7553:
@d2dyno1 They have to be from this list: https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-and-com-for-uwp-apps

d2dyno1:
How do I import them?
Is it done by `[DllImport()]`?

```
[return: MarshalAs(UnmanagedType.Bool)]
[DllImport("Kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
internal static extern bool CopyFileExW(...);
```
?

tsvietOK:
@d2dyno1 
```csharp
[return: MarshalAs(UnmanagedType.Bool)]
[DllImport("api-ms-win-core-file-l2-1-0.dll", CharSet = CharSet.Unicode, SetLastError = true)]
public static extern bool CopyFileExW(...);
```

d2dyno1:
Thanks, @tsvietOK !

I'll see what I can do then...

yaichenbaum:
This issue should be resolved in `main`.

Issue #177, opened 2019-07-19
Add support for View Alternate Data Streams


zbalkan:
I have been looking for ways to view Alternate Data Streams of NTFS in Explorer. I have used [Streams ](https://docs.microsoft.com/en-us/sysinternals/downloads/streams)from [Sysinternals ](https://docs.microsoft.com/en-us/sysinternals/) for CLI and [AlternateStreamView](https://www.nirsoft.net/utils/alternate_data_streams.html) from [NirSoft](https://www.nirsoft.net/) for UI. But I could not manage to do it in a proper way.

I'd like to have an alternative like "Show Hidden Files" but a "Show Alternate Data Streams" so that I can view them as casual files.

Although hidden files are shown with some transparency, alternate data streams can be shown as semitransparent like them but with some tricks like a different color in titles or any icon identifier like shortcuts does. That would be awesome.

duke7553:
Unfortunately the current Windows.Storage APIs don't support displaying these yet. Keeping open for now while I investigate ways to work around this limitation.

zbalkan:
I believe [AlphaFS](https://github.com/alphaleonis/AlphaFS) would suit better for all file system needs instead of System.IO and Windows.Storage in many cases. For future releases, that would probably accelerate your features.

yaichenbaum:
@zbalkan does AlphaFS work with UWP applications?

zbalkan:
@yair100 It's a drop in replacement for `System.IO`, so it won't be a problem, I assume. I have never used it with a UWP project.

yaichenbaum:
It looks promising, @duke7553 can you investigate if this will work in a UWP app? If it does then we can solve many open issues by switching to this.

zbalkan:
At this point, I believe "developing things that work" is better than "developing thing that does more". My feature suggestion may not be a must, but using AlphaFS might be helpful. I hope it would accelerate your efforts.

Jaiganeshkumaran:
If it's not supported in UWP, we can use the full trust component similar to other features that aren't possible in UWP.

gave92:
AlphaFS "works" in UWP but has all the same limitations as System.IO functions (AccessDenied everywhere).
For Alternate Data Streams we could use `FindFirstStreamW` in the fulltrust process to enumerate streams and than `OpenFileFromApp` from uwp *may* work to open them.

duke7553:
@gave92 Do you mean `CreateFileFromApp`?

[...]

Issue #843, opened 2020-05-26
Tiles View: file type not displaying correctly in my language


XPoppyX:
![82855662-e5e05e80-9ed9-11ea-957b-9ab4aee35515](https://user-images.githubusercontent.com/61259627/82856250-7ff4d680-9edb-11ea-99d0-d958467017d7.png)

PR #1573, opened 2020-07-25
Add support for shortcut files

gave92:
This PR adds support for shortcut files (".lnk", ".url")
Related issue: #177

**What works**
- Listing shortcuts with thumbnail
- Opening file shortcuts
- Navigating to destination when opening a shortcut to a folder (also open in new tab/window)
- Deleting shortcuts (but can't send to recycle bin)
- Copy/Move shortcut files
- Renaming shortcuts
- Dropping a file inside a folder shortcut

**What doesn't work**
- Drag shortcuts --> *I have no clue how to make this work*

**What's missing**
- ~Properties window could allow to edit shortcuts~
- ~Creating new shortcuts~ (from context menu)
- ~Share a .url file as link~
- ~A decent icon on lnk/url files~ --> *thanks @XPoppyX!*

*Edit*

@duke7553 We are now using 3 separate methods for supporting all file operations:
- StorageFile API
- UWP ***FromApp methods
- Win32 functions in the fulltrust process

This leads to a bit of complication, also the `ItemViewModel` starts to look a bit messy/hackish. What do you think?
Btw have you already made some progress on #632?


Jaiganeshkumaran:
@gave92 You can't share a shortcut file even in Explorer but for .URL files we could share the link of the shortcut.

Jaiganeshkumaran:
Is is possible to create new shortcut files with this PR?

gave92:
@Jaiganeshkumaran I have yet to look into creating new shortcuts, currently you can only view and use existing ones. Ideally you should be able to create new shortcuts and edit existing ones in the Property window.

Jaiganeshkumaran:
@gave92 Don't share anything for LNK files since they mostly contain a local path and sharing it doesn't make sense unless the path is accessible from other PCs.

gave92:
@Jaiganeshkumaran now you can create shortcuts to files and folders from right-click context menu :)

Jaiganeshkumaran:
Can you create a shortcut from a selected folder? Please add an option to create shortcut from selected folder.

gave92:
@Jaiganeshkumaran yes you can already do that

gave92:
Would you consider something like this for the Property windows?

image

duke7553:
@gave92 Sorry I couldn't get to this one earlier. 

To answer your questions:
- We would not want to use a TabView for Static navigation. A Pivot-like control is probably more suited for this scenario.
- Our goal is to switch over to exclusively use the *FromApp methods for File operations, which I haven't got around to yet.

Thank you for your work on this.

yaichenbaum:
> Would you consider something like this for the Property windows?
> 
> 

As @duke7553 mentioned, we would probably want to use something like a horizontal navigation menu here.

gave92:
@yair100 So something like [Pivot](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/pivot) or [horizontal NavigationView](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview) correct?

XPoppyX:
@gave92 what about this icon?

Anotación 2020-07-26 122526

yair100:
> @yair100 So something like [Pivot](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/pivot) or [horizontal NavigationView](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview) correct?

If I recall correctly, it's recommended to use the NavigationView over the Pivot control.

XPoppyX:
steam game shortcuts are shown with this icon, instead of the game icon
Anotación 2020-07-26 125627

gave92:
@XPoppyX I can confirm that at the moment ".url" files won't display the proper icon

XPoppyX:
I have tested the new icon but with reduced size:

**GridView:** 28 => 18

**TilesView:** 22 => 15

Anotación 2020-07-26 130725

I think the current size is somewhat big

XPoppyX:
@gave92 Check your repository, I added a PR

gave92:
@XPoppyX thanks a lot!

XPoppyX:
what about md5hash?
Anotación 2020-07-26 144251

gave92:
I take it doesn't work? Only on shortcuts or everywhere?

XPoppyX:
only on shortcuts

gave92:
@XPoppyX should be fixed, note that it will be the hash of the shortcut `target` and not of the shortcut itself

XPoppyX:
icon proposal
Anotación 2020-07-26 171619

XPoppyX:
- in "Shortcut" tab, for exe files it should display as "Application" type instead of just "file"
- in "General" tab, the type does not display correctly in my language:
Anotación 2020-07-26 185500
_this also happens in tiles view, see:_ #843
- Is it okay to display the type as "Ink File" instead of "Shortcut (.Ink)"?

Jaiganeshkumaran:
> * Is it okay to display the type as "Ink File" instead of "Shortcut (.Ink)"?

No, it's not. Not everyone knows what the INK extension means.

gave92:
@XPoppyX most of your notes should now be fixed:
- icon proposal --> done!
- in "Shortcut" tab, for exe files it should display as "Application" type instead of just "file" --> done!
- in "General" tab, the type does not display correctly in my language: --> done (needs resource translation/only for shortcuts)
- Is it okay to display the type as "Ink File" instead of "Shortcut (.Ink)"? --> done (needs resource translation)

XPoppyX:
nice

Files/Views/Pages/Properties.xaml.cs
```diff
@@ -78,11 +51,7 @@ private void Properties_Loaded(object sender, RoutedEventArgs e)
                 TitleBar.ButtonInactiveBackgroundColor = Colors.Transparent;
                 AppSettings.UpdateThemeElements.Execute(null);
             }
-
-            if (BaseProperties != null)
-            {
-                BaseProperties.GetSpecialProperties();
-            }
+            AppSettings.ThemeModeChanged += AppSettings_ThemeModeChanged;
```
tsvietOK, line 54: To fix black window controls (minimize, maximize, close) move `AppSettings.ThemeModeChanged += AppSettings_ThemeModeChanged;` before if in this method.

tsvietOK:
Can you increase Properties window height from 475 to 550?

tsvietOK (changes requested):
Disable MD5 has calculation for link item, as it is hash of the real file, not the link file.

tsvietOK:
@gave92 Why did you removed "Created" and "Accessed" fields for link files?

gave92:
@tsvietOK both were of the link target and not of the link itself (same as the md5 hash)

tsvietOK:
Is it possible to to don't display link file extension independently from "Show file extension" setting?

Files/Interacts/Interaction.cs
```diff
@@ -600,13 +633,38 @@ private async void Manager_DataRequested(DataTransferManager sender, DataRequest
...
+                {
+                    var value = new ValueSet
+                    {
+                        { "Arguments", "FileOperation" },
+                        { "fileop", "CreateLink" },
+                        { "targetpath", selectedItem.ItemPath },
+                        { "arguments", "" },
+                        { "workingdir", "" },
+                        { "runasadmin", false },
+                        { "filepath", Path.Combine(CurrentInstance.FilesystemViewModel.WorkingDirectory, selectedItem.ItemName + " - shortcut" + ".lnk") }
```
tsvietOK, line 661: Add ability to translate " - shortcut"

tsvietOK (changes requested):

tsvietOK (dismissed):

yair100 (changes requested):

yair100 (approved):

PR #1575, opened 2020-07-26
Implemented dragging storage items from other apps into Files UWP

### LINKED ISSUE #829: I'd like to drag storage items into/outside the app
**Describe the bug**
Drag and drop only works for moving files and not for uploading files to websites like OneDrive.

**Expected behavior**
File should be uploaded to the website.

 - OS Version: Windows 10 Pro Insider Preview Build 19624.1000
 - App version: v0.8.2.0

soumyamahunt:
Fixes #829.

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -438,30 +438,47 @@ namespace Files
 
         protected async void List_DragOver(object sender, DragEventArgs e)
         {
+            var deferral = e.GetDeferral();
+
             ClearSelection();
             if (e.DataView.Contains(StandardDataFormats.StorageItems))
             {
+                e.Handled = true;
                 IReadOnlyList<IStorageItem> draggedItems = await e.DataView.GetStorageItemsAsync();
+                e.DragUIOverride.IsCaptionVisible = true;
+
+                var folderName = Path.GetFileName(App.CurrentInstance.FilesystemViewModel.WorkingDirectory);
                 // As long as one file doesn't already belong to this folder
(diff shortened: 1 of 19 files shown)


yaichenbaum:
@soumyamahunt Instead of copying the items, it would be better to move them.

soumyamahunt:
> @soumyamahunt Instead of copying the items, it would be better to move them.

The default behaviour in Windows explorer in drag and drop is: If the item is in same drive as the destination folder then it will move the item otherwise it will just copy. Shouldn't we keep the existing behaviour as in explorer?? Wouldn't implementing this behaviour confuse users??

XPoppyX:
What do you think about adding the feature of showing a dialog to choose if you want to copy or move if you drag with the right click?

soumyamahunt:
> What do you think about adding the feature of showing a dialog to choose if you want to copy or move if you drag with the right click?

May be in a future pr, can you file an issue for this feature!!

yaichenbaum:
> > @soumyamahunt Instead of copying the items, it would be better to move them.
> 
> The default behaviour in Windows explorer in drag and drop is: If the item is in same drive as the destination folder then it will move the item otherwise it will just copy. Shouldn't we keep the existing behaviour as in explorer?? Wouldn't implementing this behaviour confuse users??

Sorry about that, I saw the new strings and automatically assumed you defaulted to this behavior.

soumyamahunt:
> I am not sure is this a part of this PR, but when you dragging an image from Google Chrome to app it is crashing.
> image

Fixed that, should work now.

PR #1597, opened 2020-07-28
Tiles and List View: Gray text for file properties

### LINKED ISSUE #1170: UI improvement for file properties in list view / tile view

When a user browses through the files in a file explorer, they try to identify the file based on two things:
1. Name of the file
2. Icon/thumbnail of the file

In the current app, all the properties shown are of the same importance (name, size, file type...)
While yes, the information is helpful and is displayed at all times, they are only necessary when the user wants to know it.

Except for the name, other properties can be of a lesser opacity, so that the name and icon/thumbnail stand out better.

Current
Current implementation

Proposed
Proposed implementation

While the exact color values are up for further discussion, I immediately notice that there is a huge improvement in the visual quality of the screen.

Current
Current implementation

Proposed
Proposed implementation

The same applies to the light theme as well, though the difference it makes seems lesser compared to the dark theme. 

Image 2
Current implementation

Group 5
Proposed implementation

This is a generic UI improvement and there are no alternatives I could think of.
Instead of making secondary information less prominent (reducing opacity/grey out), the primary information can be made more prominent (by making the name bolder/bigger) but that doesn't seem visually elegant and results in unnecessarily complicated UI.

### LINKED ISSUE #1585: File Type in Grid View should have different font
image

The file type and the folder name have the same font size, type and color and it's hard to distinguish them from one another at first glance.
I'd suggest to maybe use a light grey for the item type and size.

XPoppyX:
Closes #1170 
Closes #1585 

----------------------

**Tiles view:**

Anotación 2020-07-28 123250
Anotación 2020-07-28 123306

----------------------------------------

**List view / Details view:**

Anotación 2020-09-05 115329
Anotación 2020-09-05 115340


soumyamahunt:
Instead of providing alpha value in color itself try something like this:
`<SolidColorBrush x:Key="GrayTextColor" Color="#000000" Opacity="0.6"/>`

Files/App.xaml
```diff
@@ -17,20 +17,23 @@
                             <SolidColorBrush x:Key="CloudDriveSyncStatusExcludedColor" Color="#AAAAAA" />
                             <SolidColorBrush x:Key="ApplicationPageBackgroundThemeBrush" Color="White" />
                             <SolidColorBrush x:Key="SystemControlPageBackgroundMediumAltMediumBrush" Color="#99000000" />
+                            <SolidColorBrush x:Key="GrayTextColor" Color="#99000000" />
```
soumyamahunt, line 20: Instead of providing alpha value in color itself try something like this:
`<SolidColorBrush x:Key="GrayTextColor" Color="#000000" Opacity="0.6"/>`

XPoppyX: tested, still not working

XPoppyX:
I have received this reply about the color change problem, I need your help to make the necessary changes, since I don't have much experience.
-------------------------------------

The behavior sounds right to me. See how the TextBlock/TextBox's Foreground brush gets set in the DataGridTextColumn's cells, via the RefreshForeground method. It uses the DataGridTextColumn.Foreground property or DataGridRow.ComputedForeground property (which uses DataGridRow.Foreground or DataGrid.RowForeground or DataGrid.AlternatingRowForeground). It's a pretty complicated inheritance mechanism, right?

In your case even the light mode must be broken. Check what happens after you edit some cells (and a TextBox is used instead of a TextBlock). The Foreground starts to be black and because of row recycling random rows will show incorrectly as you scroll down. By the way if you started your app in Dark mode by default, you'd think the Light mode is broken.

Anyways this is how I made it work for me in my application:

In app.xaml I declared custom theme colors:

image

(see how my default RequestedTheme is "Light")

I added a combobox to my MainPage.xaml:

<ComboBox x:Name="cmbAppTheme" SelectedIndex="0" SelectionChanged="cmbAppTheme_SelectionChanged">
  <ComboBoxItem>Default</ComboBoxItem>
  <ComboBoxItem>Light</ComboBoxItem>
  <ComboBoxItem>Dark</ComboBoxItem>
</ComboBox>
Here's what I do when the combobox changes (dg is my DataGrid):

        private void cmbAppTheme_SelectionChanged(object sender, SelectionChangedEventArgs e)
        {
            switch (cmbAppTheme.SelectedIndex)
            {
                case 0:
                    if (this.RequestedTheme != ElementTheme.Default)
                    {
                        this.RequestedTheme = ElementTheme.Default;
                    }
                    break;
                case 1:
                    if (this.RequestedTheme != ElementTheme.Light)
                    {
                        this.RequestedTheme = ElementTheme.Light;
                    }
                    break;
                case 2:
                    if (this.RequestedTheme != ElementTheme.Dark)
                    {
                        this.RequestedTheme = ElementTheme.Dark;
                    }
                    break;
            }

            if (dg != null)
            {
                SolidColorBrush grayTextColorBrush = this.ThemeDictionary["GrayTextColorBrush"] as SolidColorBrush;

                foreach (DataGridColumn dataGridColumn in dg.Columns)
                {
                    DataGridTextColumn dgtc = dataGridColumn as DataGridTextColumn;

                    if (dgtc != null)
                    {
                        dgtc.Foreground = grayTextColorBrush;
                    }
                }
            }
        }
Also whenever a DataGridTextColumn is added to the DataGrid, its Foreground property needs to be set:

        private void Dg_Columns_CollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
        {
            ...
                if (e.NewItems != null)
                {
                    SolidColorBrush grayTextColorBrush = this.ThemeDictionary["GrayTextColorBrush"] as SolidColorBrush;

                    foreach (DataGridColumn dataGridColumn in e.NewItems)
                    {
                        DataGridTextColumn dgtc = dataGridColumn as DataGridTextColumn;

                        if (dgtc != null)
                        {
                            dgtc.Foreground = grayTextColorBrush;
                        }
                    }
                }
            ...
        }
That worked for me. I hope it works for you. Thanks

Jaiganeshkumaran:
@XPoppyX Instead of changing the colour itself, change the opacity to something like .6 for the text block. This should work for any theme and accent colour.

XPoppyX:
> @XPoppyX Instead of changing the colour itself, change the opacity to something like .6 for the text block. This should work for any theme and accent colour.

i will test it

XPoppyX:
@yaichenbaum i changed the font size to how it was before, is this ok?

XPoppyX:
@Jaiganeshkumaran thanks, just changing opacity instead of foreground worked

yaichenbaum:
> @yaichenbaum i changed the font size to how it was before, is this ok?

@XPoppyX It's best to keep it how it is until we add customization options for it.

XPoppyX:
> @XPoppyX It's best to keep it how it is until we add customization options for it.

ok, I thought that reducing the size to 12 was a temporary solution until the color change is done, and I increased it back to 14

yaichenbaum (approved):

yaichenbaum:
@XPoppyX Great work! :shipit:

PR #1613, opened 2020-07-30
Adds a context menu item to open the file location of a shortcut

### LINKED ISSUE #1606: Shortcuts: Add "Open file location" in context menu
it is faster to open the file location from the context menu, it would be a bit annoying to have to open the properties window first
Anotación 2020-07-29 215651

tobias-mayer:
Fixes #1606


Files/View Models/SelectedItemsPropertiesViewModel.cs
```diff
@@ -517,10 +525,18 @@ public void CheckForImage()
                     IsSelectedItemImage = true;
                     return;
                 }
+                else if (ItemExtension.Equals(".lnk", StringComparison.OrdinalIgnoreCase)
```
gave92, line 528: Instead of checking the extension you can probably use `App.CurrentInstance.ContentPage.SelectedItem..IsShortcutItem`

Files/Interacts/Interaction.cs
```diff
@@ -315,6 +315,13 @@ public void OpenItemWithApplicationPicker_Click(object sender, RoutedEventArgs e
             OpenSelectedItems(true);
         }
 
+        public void OpenFileLocation_Click(object sender, RoutedEventArgs e)
+        {
+            var item = CurrentInstance.ContentPage.SelectedItem as ShortcutItem;
+            var folderPath = Path.GetDirectoryName(item.TargetPath);
```
gave92, line 321: Shortut `TargetPath` can somatimes be null or empty. Please check if this causes issues.
The most notable example of this is the "Vault" link in the OneDrive folder

Files/View Models/SelectedItemsPropertiesViewModel.cs
gave92: Also If you are using this to show "Open File Location" you may not want to do this for ".url" files

Files/Interacts/Interaction.cs
tobias-mayer: I could only get the TargetPath empty/null by manually setting it to empty/null during debugging.
But when it is null/empty nothing happens. Do you have any suggestions on how we should handle this?

Files/View Models/SelectedItemsPropertiesViewModel.cs
tobias-mayer: I just added a new commit which makes use of `App.CurrentInstance.ContentPage.SelectedItem.IsShortcutItem`.
Url files are now also supported. This should now be the same behaviour as in the `PropertiesShortcut` Tab

XPoppyX: @tobias-mayer  i don't think it makes sense to show "open file location" for .url files, the only thing that option would do is open the shortcut, which by the way should be disabled in the properties window

tobias-mayer: Ok yes I also think so. I'll change this. Should I also change it in the properties window or should I open another issue therfore?

XPoppyX: it would be better to add an issue and fix it in a new PR

XPoppyX:
"open file location" only appears in list view

Files/View Models/SelectedItemsPropertiesViewModel.cs
tobias-mayer: Now the option only shows up for ".lnk" files in the context menu

tobias-mayer:
> "open file location" only appears in list view

Sorry, I totally forgot about that. It should now show up in the GridView as well.

XPoppyX:
Is it possible to auto-select the folder or file after opening the file location?

tobias-mayer:
> Is it possible to auto-select the folder or file after opening the file location?

I think we could handle this in the `OnNavigatedTo` in the `BaseLayout`. But I´m new to the project and not to sure about that. 
Does anyone else have an idea?

Files/Views/LayoutModes/GenericFileBrowser.xaml
```diff
@@ -251,6 +251,16 @@
                     <FontIcon FontFamily="{StaticResource CustomGlyph}" Glyph="&#xF108;" />
                 </MenuFlyoutItem.Icon>
             </MenuFlyoutItem>
+            <MenuFlyoutItem
+                x:Name="OpenFileLocation"
+                x:Uid="BaseLayoutItemContextFlyoutOpenFileLocation"
```
tsvietOK, line 256: Add ability to translate this text.

tsvietOK (changes requested):

Files/Views/LayoutModes/GenericFileBrowser.xaml
tobias-mayer: I added entries for de-DE and en-US in the corresponding Resource files.

tsvietOK (approved):

yaichenbaum:
@tobias-mayer Will this work for the recent item section on the home page as well?

tobias-mayer:
> @tobias-mayer Will this work for the recent item section on the home page as well?

@yair100 Opening the file location is already implemented for the recent item section.

yair100 (approved):

PR #1666, opened 2020-08-06
Added a push notification asking for feedback when the app crashes

### LINKED ISSUE #1663: Show a notification when the app crashes

jaigak:
Implements and closes #1663


jaigak:
@yair100 Please don't merge this yet. I'm planning for adding buttons to the notification so that the user can easily access the log file and open an issue.

Jaiganeshkumaran:
Here's an idea. Before showing the notification, we could try to save all the tabs to a file. When the user launches the app again, tabs would be recovered and everything will be ready to go. It's like 'Continue where you left off' but when an exception is thrown.

yaichenbaum:
> I'm planning for adding buttons to the notification so that the user can easily access the log file and open an issue.

The ideal use of the notification would be to ask the user to open an issue on GitHub.

> Here's an idea. Before showing the notification, we could try to save all the tabs to a file. When the user launches the app again, tabs would be recovered and everything will be ready to go. It's like 'Continue where you left off' but when an exception is thrown.

This is a great idea.

jaigak:
@yair100 Should we add a button to restart the app? It doesn't make sense because the notification is sent after the app crashes. Instead, we could open the when the user clicks on the notification itself. That's the default behavior.

Jaiganeshkumaran:
Here's how it would look like:
திரைப்பிடிப்பு (11)

generalguy41:
Bit of feedback; While I think this is a great idea and looks well implemented, should it maybe not be combined to one button or something else? As in, if the user clicks "Report this issue", it brings them to the github issue page, but then they have to go look for the log location, which is a couple extra steps. And on the other side, if they click "Open log location", then it opens the log location, they have the log, but have to figure out where to report the issue. Unless there's some way to not dismiss the notification without manually clicking the "x", but I'm not sure if that's a thing...

Jaiganeshkumaran:
@generalguy41 The notification is always in the action center until the users clears it. A single button can't open GitHub and the log file.

generalguy41:
@Jaiganeshkumaran Does clicking one of those buttons not dismiss the notification entirely? Or does it only commit the action and then dismiss it to the action center?

Jaiganeshkumaran:
The screenshot is just a prototype. The latest commit does not include the buttons yet. Only the text.

generalguy41:
Yea, I agree. That would probably be the best solution.

Jaiganeshkumaran:
Since UWP background tasks don't support launching URLs or folders, these buttons will activate the app but also will do the action.

Jaiganeshkumaran:
For some reason, the notification is fired at the starting because of two exceptions occur in DrivesManager. One exception is caught thus preventing from crash but the notification still appears. This also occurs in the debugger. I'm working on rewriting the block to prevent sending a notification. Until then no notification.

Jaiganeshkumaran:
I fixed it. Getting 'System.UsedSpace' doesn't working for OneDrive so it threw an exception. Now I have restored the notification.

Jaiganeshkumaran:
@generalguy41 Unfortunately, I learned that clicking the buttons will remove it from action center for ever. Instead we could remove those buttons and instead clicking on it will open a window that offers these options.

generalguy41:
Perhaps another alternative would be just to have the button open the url to the github bug page, and reference where to get the log file from there in the placeholder text?

Jaiganeshkumaran:
We could tell users how to access the log location when creating a new issue. Maybe Luke can modify the issue templates to teach users how to upload the log file because most users don't know and they don't upload the log file.

yaichenbaum:
@Jaiganeshkumaran The notification should have an option to restore tabs and an option to report the issue. The report issue button should link to [here](https://github.com/files-community/files-uwp/issues/new/choose).

Jaiganeshkumaran:
> @Jaiganeshkumaran The notification should have an option to restore tabs and an option to report the issue. The report issue button should link to [here](https://github.com/files-community/files-uwp/issues/new/choose).

I already added to button to open the issue chooser. Instead of having a button to restore why not make it the default behaviour. Clicking the report button itself starts a new instance. I've not worked on saving the tabs yet but I'll do that soon.

Jaiganeshkumaran:
When clicking on the report button or simply opening a new instance will show a notification if it could restore the tabs.

Files.Package/Package.appxmanifest
```diff
@@ -48,14 +48,14 @@
...
-				<uap5:Extension
+        <uap5:Extension
 				  Category="windows.appExecutionAlias"
 				  Executable="files.exe"
 				  EntryPoint="files.App">
-					<uap5:AppExecutionAlias>
-						<uap5:ExecutionAlias Alias="files.exe"/>
-					</uap5:AppExecutionAlias>
-				</uap5:Extension>
+          <uap5:AppExecutionAlias>
```
tsvietOK, line 55: Don't change this file.

Files/Views/MainPage.xaml.cs
```diff
@@ -106,6 +107,60 @@ protected override async void OnNavigatedTo(NavigationEventArgs eventArgs)
...
+                            {
+                                Visual = new ToastVisual()
+                                {
+                                    BindingGeneric = new ToastBindingGeneric()
+                                    {
+                                        Children =
+                                        {
+                                            new AdaptiveText()
+                                            {
+                                                Text = "We've restored your tabs"
```
tsvietOK, line 133: Add translation.

Files/Views/MainPage.xaml.cs
```diff
@@ -120,6 +175,56 @@ protected override async void OnNavigatedTo(NavigationEventArgs eventArgs)
...
+            {
+                Visual = new ToastVisual()
+                {
+                    BindingGeneric = new ToastBindingGeneric()
+                    {
+                        Children =
+                        {
+                            new AdaptiveText()
+                            {
+                                Text = "We couldn't restore your tabs"
```
tsvietOK, line 193: Add translation.

tsvietOK (changes requested):
Resolve conflicts, please.

Jaiganeshkumaran:
Although it seem to save open tabs, it prevents from the app from crashing. I will remove that and make it for continue where you left off.

Files/Views/MainPage.xaml.cs
Jaiganeshkumaran: This notification has been removed since tab's are no longer restored.

Jaiganeshkumaran:
I have modified the issue templates in my branch to emphasize to upload the log file. It should be live after merging this pull request.

Files/Views/SettingsPages/About.xaml.cs
```diff
@@ -13,19 +12,16 @@ public About()
...
         }
+        
+        private void OpenLogLocationButton_Click(object sender, RoutedEventArgs e) => View_Models.SettingsViewModel.OpenLogLocation();
 
         private async void FeedbackListView_DoubleTapped(object sender, Windows.UI.Xaml.Input.DoubleTappedRoutedEventArgs e)
         {
             if (FeedbackListView.SelectedIndex == 0)
             {
-                await Launcher.LaunchUriAsync(new Uri(@"https://github.com/files-community/files-uwp/issues/new/choose"));
+                View_Models.SettingsViewModel.ReportIssueOnGitHub();
```
yaichenbaum, line 24: Why did you move the code to SettingsViewModel?

Jaiganeshkumaran: To keep it same across both the notification and about. In future if the link changes, we could change it easily from one place.

Files/App.xaml.cs
```diff
@@ -359,6 +362,18 @@ protected override async void OnActivated(IActivatedEventArgs args)
...
+                    var eventArgsForNotification = args as ToastNotificationActivatedEventArgs;
+                    rootFrame.Navigate(typeof(MainPage), "tryRecoverTabs", new SuppressNavigationTransitionInfo());
+                    // Ensure the current window is active.
+                    Window.Current.Activate();
+                    Window.Current.CoreWindow.PointerPressed += CoreWindow_PointerPressed;
+                    Window.Current.CoreWindow.Activated += CoreWindow_Activated;
+                    currentView.BackRequested += Window_BackRequested;
+                    // Launch the URI now
+                    SettingsViewModel.ReportIssueOnGitHub();
+                    return;*/
```
yaichenbaum, line 376: We shouldn't add code that is not being used, it would be better to add this later when it's ready to be used.

Files/Filesystem/DriveItem.cs
```diff
@@ -49,20 +49,27 @@ public DriveItem(StorageFolder root, DriveType type)
...
+                    SpaceUsed = MaxSpace - FreeSpace;
+                    SpaceText = string.Format(
+                        ResourceController.GetTranslation("DriveFreeSpaceAndCapacity"),
+                        FreeSpace.ToBinaryString().ConvertSizeAbbreviation(),
+                        MaxSpace.ToBinaryString().ConvertSizeAbbreviation());
+                }
+                catch (NullReferenceException)
+                {
+                    SpaceText = ResourceController.GetTranslation("DriveCapacityUnknown");
+                }
```
yaichenbaum, line 68: Are these changes necessary for this pull request?

Files/Views/MainPage.xaml.cs
yaichenbaum: It looks like the code is still in here, can you remove the code or add the translations?

yair100 (changes requested):

Files/Filesystem/DriveItem.cs
Jaiganeshkumaran: Yes because actually two exceptions are thrown for certain drives which causes a error notification on startup.

Files/Views/MainPage.xaml.cs
```diff
@@ -119,6 +121,7 @@ protected override async void OnNavigatedTo(NavigationEventArgs eventArgs)
                 var mainView = rootFrame.Content as MainPage;
                 mainView.SelectedTabItem = AppInstances[App.InteractionViewModel.TabStripSelectedIndex];
             }
+
```
yaichenbaum, line 124:
```suggestion
```

yair100 (dismissed):

yair100 (dismissed):

tsvietOK (changes requested):

tsvietOK (approved):

PR #1667, opened 2020-08-06
Fixed a bug where "empty recycle bin" was not grayed out although the bin is empty

### LINKED ISSUE #1420: "Empty Recycle Bin" isn't greyed out in the sidebar right-click menu when the bin is empty
**Describe the bug**
The "Empty Recycle Bin" button in the right-click sidebar menu isn't greyed out when the Bin is empty, unlike the right-click menu in the Bin folder

**To Reproduce**
Right click on the Recycle Bin, of course when it's empty

**Expected behavior**
The button should be greyed out, just like the right-click menu of the Bin folder

**Screenshots**
image

 - OS Version: Windows 10 2004
 - App version: v0.12

tobias-mayer:
Fixes #1420

### DIFF
--- a/Files/UserControls/SidebarControl.xaml
+++ b/Files/UserControls/SidebarControl.xaml
@@ -295,6 +295,7 @@
                                 x:Name="EmptyRecycleBin"
                                 x:Uid="BaseLayoutContextFlyoutEmptyRecycleBin"
                                 x:Load="{x:Bind ShowEmptyRecycleBin, Mode=OneWay}"
+                                IsEnabled="{x:Bind RecycleBinHasItems, Mode=OneWay}"
                                 Click="{x:Bind local1:App.CurrentInstance.InteractionOperations.EmptyRecycleBin_ClickAsync}"
                                 Text="Empty recycle bin">
                                 <MenuFlyoutItem.Icon>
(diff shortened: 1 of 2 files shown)


yaichenbaum:
@tobias-mayer There is already a boolean that holds the the value if recycle bin has items in it.

tobias-mayer:
> @tobias-mayer There is already a boolean that holds the the value if recycle bin has items in it.

@yair100 Do you mean `ShowEmptyRecycleBin`? This only indicates whether the right clicked element was the recycle bin, not if it is empty or has items.
And if you mean `IsFolderEmptyTextDisplayed` in the ItemViewmodel: This does not work from the sidebar because it belongs to the currently opened folder and not to the clicked sidebar item.
Otherwise where is that property?

yaichenbaum:
@tobias-mayer I do not recall where the property is saved, however I know it was used for the right click menu within recycle bin.

tobias-mayer:
> @tobias-mayer I do not recall where the property is saved, however I know it was used for the right click menu within recycle bin.

@yair100 Yes that is the `IsFolderEmptyTextDisplayed` I mentioned above. I actually tried that one but that does not work in this case, because the currently opened folder could be anything when you right click the sidebar item.

yair100 (approved):

PR #1717, opened 2020-08-12
Fix crash on invalid shortcuts

### LINKED ISSUE #1716: App crashes when parsing/opening some shortcuts
**Describe the bug**
App may crash with a "KeyNotFoundException" when browsing a folder that contains invalid shortcuts.
App may crash when clicking "Open File Location" for a shortcut and the destination does not exist.
App may crash when viewing properties of a shortcut and the destination does not exist.

 - OS Version: [e.g. Windows 10 1903 18362.175]
 - App version: [v0.13.0]

gave92:
Fixes #1716

Related AppCenter reports: [1](https://appcenter.ms/orgs/Files-UWP/apps/Files/crashes/errors/3529088089u/overview)


Files/Strings/en-US/Resources.resw
```diff
@@ -951,4 +951,13 @@
   <data name="SettingsPreferencesShowMultitaskingControl.Text" xml:space="preserve">
     <value>Show multitasking flyout in the navigation toolbar</value>
   </data>
+  <data name="InvalidItemDialogContent" xml:space="preserve">
+    <value>The item referenced is either invalid or inaccessible.
+Message:
+
```
tsvietOK, line 957: Why is there an empty line?

gave92: Those new lines are preserved when showing the dialog.
I kept the very same message that was shown before which contained a couple of "\n"

tsvietOK: It is better to contain \n, instead of empty lines.

tsvietOK: Or, if possible, enable word wrap for dialog message.

gave92: I'm not sure if "\n" works in resource files

tsvietOK: And another way without \n, is `The item referenced is either invalid or inaccessible.{0}Message:{1}{2}` and in `string.Format` change 0 and 2 to `Enviroment.NewLine`

PR #1733, opened 2020-08-13
replaced hardcoded property names with nameof()

### LINKED ISSUE #1729: Use nameof() instead
There's nothing wrong with using `NotifyPropertyChanged("SomeProperty")`. The problem starts when the property gets renamed. You then must remember (and know where) to rename all the strings within `NotifyPropertyChanged("SomeProperty")` to avoid any hidden errors. With the new solution, VS wouldn't let you complile the app - pointing out the problem immediately.

I'd like to propose a little code improvement - changing `NotifyPropertyChanged("SomeProperty")` -> to `NotifyPropertyChanged(nameof(SomeProperty))`

tobias-mayer:
Fixes #1729 
This is just an addition to the PR from @d2dyno1.
This class was probably forgotten because `RaisePropertyChanged` is used instead of `NotifyPropertyChanged`.
But we should use nameof() wherever possible

### DIFF
--- a/Files/View Models/CurrentInstanceViewModel.cs
+++ b/Files/View Models/CurrentInstanceViewModel.cs
@@ -13,9 +13,9 @@ namespace Files.View_Models
             set
             {
                 Set(ref _IsPageTypeNotHome, value);
-                RaisePropertyChanged("IsCreateButtonEnabledInPage");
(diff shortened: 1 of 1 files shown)


yaichenbaum:
@tobias-mayer can you take a look at #7625?

PR #1754, opened 2020-08-18
Added support for Dynamic Shell Context Menus

### LINKED ISSUE #1397: Bring Support Of Dynamic Shell Context Menus
Alright I have finally finished my investigation around dynamic shell context menus. I found it very difficult overall with incomplete Windows API documentations and unanswered questions under Stack Overlow regarding bringing support of dynamic shell context menus. So instead of full support of dynamic shell context menus, I think it can be supported as scanning all CLSIDs under the registry and then provide the user to select an application to execute under the settings page. This feature would bring support of shell context menus for applications like 7-zip, notepad++ and other types of applications which doesn't use static shell context menu.

I would like to start implementing dynamic shell content menus. This will allow files UWP to fully support of Windows Explorer's context menu. The idea I am proposing is first known dynamic shell context menus can be stored in an XML file as a database and implement them manually for each application then the unknown ones can be added later or provide the user under the settings tab to let them choose automatically to create additional context menu items per file or folder item.

I investigated to query dynamic shell context menu by using PowerShell. That didn't give enough information to parse and execute the shell context menu. I found out that not all the items are being listed under the PowerShell script. I also used COM implementation under Windows Forms and that however can be implemented only under a Windows Forms application. I also tried using Shell32.DLL to parse the context menu. There is no documentation at least from Microsoft's sites how to parse submenu items from different applications such as 7-Zip. Stack overflow site is also full of unanswered questions around this.

**Additional context**
Please assign this to me so that I can work on it. Any feedback is welcome.

gave92:
Implements and closes #1397

Here is some work in progress to add (close to) full support for right-click menu.
If you could try this out and let me know what you think it would be great!

Screenshot (6)


Jaiganeshkumaran:
@gave92 Consider adding an option to disable these. Although they are useful, they can clutter the experience at times.

gave92:
@Jaiganeshkumaran yeah I guess items are much taller in our menu wrt to explorer.
We could also only show them when holding the Shift key maybe.
Or put all the stuff from explorer under a SubItem of the regular menu

Jaiganeshkumaran:
@gave92 We could have a more menu that shows all these options in a submenu.

yaichenbaum:
The default should be to show the context menu items, however, if there is enough interest, we can add an option to hide them.

generalguy41:
I like seeing the context menu items, but I think the menu itself is definitely too tall. Maybe we reduce the padding slightly?

gave92:
Menu will now show up to 4 menu items from explorer at the top of the list. The rest is put inside the "More" menu item so the menu won't become too tall but the top elements will be immediately visible. If the user is pressing Shift it shows more items

Screenshot (7)

yaichenbaum:
@gave92 This is how I think it should work, by default it should show all the items without the overflow menu, if the user would like the menu to be cleaner, there can be a setting "Use overflow menu for the right click context menu", this setting should then follow the behavior you mentioned above with showing 4 items and putting the rest in the overflow menu,

yaichenbaum:
@gave92 Showing 4 items by default sounds good. A setting can definitely be added, but it doesn't need to be part of this pull request.

yaichenbaum:
I did some testing and I found a couple of things that need to be addressed,
- The pin to quick access option should be removed.
- The run as administrator option shows twice because of the implementation we added ourselves.
- Padding for the menu items should be reduced when opening the context menu from a mouse, it should still show the large menu when using a tablet or touch screen.

jaigak:
@yair100 We can remove our implementation and use the built-in one since it can change in the future.

MadafakinRio:
Please make this optional, one of the reasons why people wanted a UWP explorer for so long was the clutter and ugliness.

yaichenbaum:
@MadafakinRio Please add your feedback to #1397.

gave92:
@yair100, @Jaiganeshkumaran now Run as Admin and Run as User call the shell implementation 👍
@Jaiganeshkumaran For ".msi" files the "Open" option is now hidden and the "Install" option is shown -> is this what you meant?

yair100 (changes requested):

yair100 (dismissed):

gave92:
Uops sorry @yair100 it should be good now

yair100 (dismissed):

yair100 (dismissed):

gave92:
@yair100 crash in release mode appears to be fixed, I'll test again with the azure build -> it works

~Edit: tomorrow I'm going to try a different solution as the last commit adds a resourcedictionary to App.xaml which might have some impact on startup performance~ -> done, changed to UserControl

Files/BaseLayout.cs
```diff
@@ -259,78 +273,141 @@ private void UnloadMenuFlyoutItemByName(string nameToUnload)
...
+                await App.Connection.SendMessageAsync(new ValueSet() {
+                    { "Arguments", "ExecAndCloseContextMenu" },
+                    { "Handle", shellContextMenuTag.menuHandle } });
             }
         }
 
         public void RightClickItemContextMenu_Opening(object sender, object e)
         {
-            SetShellContextmenu();
+            var selectedFileSystemItems = App.CurrentInstance.ContentPage.SelectedItems;
```
tsvietOK, line 398: `SelectedItems` already is a part of `BaseLayout`, so you don't need to use full path to it.

gave92: Right, there is probably no need to create a temp variable at all

Common/ContextMenu.cs
```diff
@@ -0,0 +1,33 @@
...
+        MFT_MENUBARBREAK = 32,
+        MFT_MENUBREAK = 64,
+        MFT_OWNERDRAW = 256,
+        MFT_RADIOCHECK = 512,
+        MFT_SEPARATOR = 2048,
+        MFT_RIGHTORDER = 8192,
+        MFT_RIGHTJUSTIFY = 16384
+    }
+
+    public class IWin32ContextMenu
```
tsvietOK, line 19: Why class name starts with 'I'? It is not an interface. Same for `IWin32ContextMenuItem`

Files.Launcher/Program.cs
```diff
@@ -233,17 +239,32 @@ private static async Task parseArguments(AppServiceRequestReceivedEventArgs args
...
-                    responseArray.Add("ParsedArguments", JsonConvert.SerializeObject(resultArgument));
-                    await args.Request.SendResponseAsync(responseArray);
+                case "ExecAndCloseContextMenu":
+                    var menuKey = (string)args.Request.Message["Handle"];
+                    var execThreadWithMessageQueue = handleTable.GetValue<Win32API.ThreadWithMessageQueue<ValueSet>>(menuKey);
+                    if (execThreadWithMessageQueue != null)
+                    {
+                        await execThreadWithMessageQueue.PostMessage(args.Request.Message);
+                    }
+                    //handleTable.RemoveValue(menuKey);
```
tsvietOK, line 258: Could it be removed?

Files.Launcher/Program.cs
```diff
@@ -276,6 +297,52 @@ private static async Task parseArguments(AppServiceRequestReceivedEventArgs args
...
+                    var split = filePath.Split('|').Where(x => !string.IsNullOrWhiteSpace(x));
+                    var cMenuLoad = Win32API.ContextMenu.GetContextMenuForFiles(split.ToArray(),
+                        extendedMenu ? Shell32.CMF.CMF_EXTENDEDVERBS : Shell32.CMF.CMF_NORMAL, FilterMenuItems(showOpenMenu));
+                    table.SetValue("MENU", cMenuLoad);
+                    return cMenuLoad;
+
+                case "ExecAndCloseContextMenu":
+                    var cMenuExec = table.GetValue<Win32API.ContextMenu>("MENU");
+                    cMenuExec?.InvokeItem(message.Get("ItemID", -1));
+                    //table.RemoveValue("MENU");
```
tsvietOK, line 318: Could it be removed?

Files.Launcher/Win32API_ContextMenu.cs
```diff
@@ -0,0 +1,425 @@
...
+                    {
+                        Marshal.ReleaseComObject(cMenu);
+                        cMenu = null;
+                    }
+
+                    disposedValue = true;
+                }
+            }
+
+            // TODO: eseguire l'override di un finalizzatore solo se Dispose(bool disposing) include il codice per liberare risorse non gestite.
```
tsvietOK, line 365: Can you write a comment in English so other developers can understand it too.

Common/ContextMenu.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public List<IWin32ContextMenuItem> Items { get; set;  }
```
tsvietOK, line 21:
```suggestion
        public List<IWin32ContextMenuItem> Items { get; set; }
```

tsvietOK (changes requested):

Files.Launcher/Win32API_ContextMenu.cs
gave92: Those comments are created automatically :) I'll translate them

Common/ContextMenu.cs
gave92: Right, it used to be

Files.Launcher/Program.cs
gave92: Do you mean the commented line? I'd like to keep it: that call is needed cleanup resources when menu is closed. Unfortunately if you uncomment it some menu items will randomly stop working.
Resource cleanup is currently done on app closing, but I hope to find a solution for the issue above and cleanup as soon as a menu is closed.
I'll add a comment about this.

Files.Launcher/Program.cs
gave92: Same as above

tsvietOK (dismissed):

gave92:
Sorry just noticed it was crashing on android/mtp devices

yair100 (approved):

tsvietOK (approved):

PR #1810, opened 2020-08-26
Fixed an issue where external devices cannot be removed while app is opened

### LINKED ISSUE #1767: External device cannot be removed while app is opened
**Describe the bug**
If you open an external device in app, you cannot remove device unless you close the app

**To Reproduce**
1. Plug in any external device
2. Open it in app
3. Change disk to any other and try to remove external device from taskbar option
4. Not possible to remove device, windows keeps telling that device is in use, after closing the app, then the device is removed without problem.

**Expected behavior**
App should close any link to external device when selecting the Remove device option in taskbar to allow safe extraction.

 - OS Version: Windows 10 2004 19041.450
 - App version: v0.13.1.0

gave92:
Fix #1767

This PR allows to eject a connected USB drive without closing the app. It's sufficient to close the tab or navigate away.

Please check if you agree on how to detect when a tab was closed, I don't know much about how tabs are handled.

### DIFF
--- a/Files/Views/MainPage.xaml.cs
+++ b/Files/Views/MainPage.xaml.cs
@@ -80,7 +80,7 @@ namespace Files.Views
                 foreach (var removedTab in e.OldItems)
                 {
                     // Cleanup resources for the closed tab
-                    ((((removedTab as TabItem).Content as Grid).Children[0] as Frame).Content as IShellPage).FilesystemViewModel?.Dispose();
+                    ((((removedTab as TabItem).Content as Grid).Children[0] as Frame).Content as IShellPage)?.FilesystemViewModel?.Dispose();
                 }
             }
         }
--- a/Files/Views/MainPage.xaml.cs
+++ b/Files/Views/MainPage.xaml.cs
@@ -69,6 +69,20 @@ namespace Files.Views
             }
 
             AllowDrop = true;
+
+            AppInstances.CollectionChanged += AppInstances_CollectionChanged;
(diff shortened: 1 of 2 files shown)


tsvietOK:
Getting an exception when trying to close tab
image
Steps: 
Open app
Open one more tab(don't select it)
Close created tab

Files/Views/MainPage.xaml.cs
```diff
@@ -69,6 +69,20 @@ public MainPage()
             }
 
             AllowDrop = true;
+
+            AppInstances.CollectionChanged += AppInstances_CollectionChanged;
+        }
+
+        private void AppInstances_CollectionChanged(object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs e)
+        {
+            if (e.OldItems != null)
```
tsvietOK, line 78: What do you think about possibility to move this to tab close event?

gave92: I think the `TabStrip_TabCloseRequested` event is not called when the tab is closed by Ctrl+W
But perhaps I could move the logic to `horizontalTabView_TabItemsChanged/verticalTabView_TabItemsChanged` which are triggered by the `TabItemsChanged` event?

tsvietOK: I think it is possible to find a better solution, but now we can keep this solution.

tsvietOK (approved):

PR #1832, opened 2020-08-27
Add an option to display dotfiles.

SketchNI:
It always seemed weird to me that the various file explorers (Windows, Nautilius, and whatever MacOS's is called) allow you to show and hide File Extensions but never give consideration for Dotfiles and leave an empty slot.

Before Patch: 
image

After Patch:
image


yaichenbaum:
@SketchNI Thank you for opening this pull request, this change is a good one, however to prevent extra work, in the future, please open an issue so the change can be disccussed before any time is spent on it.

yaichenbaum:
The implmetnation looks fine, however, the option should only be enabled when the option to show file extensions is set to false.

SketchNI:
> @SketchNI Thank you for opening this pull request, this change is a good one, however to prevent extra work, in the future, please open an issue so the change can be disccussed before any time is spent on it.

Whoops, will do!

> The implmetnation looks fine, however, the option should only be enabled when the option to show file extensions is set to false.

I'll do this now. Thanks

SketchNI:
Requested changes have been made.
image

image

yaichenbaum:
@SketchNI I think the toggle switch should still be visible even when disabled.

tsvietOK:
I think there is shouldn't be an option at all. Dot files should be displayed independently from ShowFileExtensions setting. @yair100 agreed with me.

SketchNI:
> I think there is shouldn't be an option at all. Dot files should be displayed independently from ShowFileExtensions setting. @yair100 agreed with me.

So I should revert the behaviour so it doesn't matter if none, one or both of ShowFileExtensions and ShowDotfileNames are seleceted?

tsvietOK:
Right. Dot files should be visible like files with name. ShowDotfileNames is no more needed.

SketchNI:
Okay. I'll sort that tomorrow.

jaigak (changes requested):
Windows 10 version 1903 and later allows you to have files starting with a dot (with or without extension) but previous versions don't allow. Please check the version to prevent crash.

SketchNI:
> Windows 10 version 1903 and later allows you to have files starting with a dot (with or without extension) but previous versions don't allow. Please check the version to prevent crash.

@Jaiganeshkumaran I've added a condition to check the version. It reads the ReleaseID from the registry which feels like a cheap hack imo. Advise if there is a better way?

Jaiganeshkumaran:
@SketchNI You can check whether the APIs for 1903 are present using ApiInformation.IsApiContractPresent("Windows.Foundation.UniversalApiContract", versionCount).
Set versionCount as 8 because 1903 is the 8th feature update of Windows 10.

SketchNI:
Thanks. This is the first time I've worked with the UWP and only the third time I've ever used C# so I am open to any and all feedback regarding my commits.

yair100 (dismissed):

Files/Converters/InverseBoolConverter.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using Windows.UI.Xaml.Data;
+
+namespace Files.Converters
```
tsvietOK, line 4: Is it needed?

Files/View Models/ItemViewModel.cs
```diff
@@ -1112,7 +1129,7 @@ public void AddFileOrFolderFromShellFile(ShellFileItem item)
                 });
             }
 
-            IsFolderEmptyTextDisplayed = false;
+            IsFolderEmptyTextDisplayed = true;
```
tsvietOK, line 1132: Is it needed?

Files/Views/SettingsPages/FilesAndFolders.xaml
```diff
@@ -4,10 +4,13 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:local="using:Files"
+    xmlns:ivb="using:Files.Converters"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     Background="{ThemeResource ApplicationPageBackgroundThemeBrush}"
     mc:Ignorable="d">
-
+    <Page.Resources>
```
tsvietOK, line 11: Is it needed?

tsvietOK (changes requested):

Files/Converters/InverseBoolConverter.cs
SketchNI: The namespace or the Converter itself?

The Converter was being used before I was asked to remove the ShowDotfileNames setting. It's not being used anymore but I left it alone incase it's useful in future.

Files/Views/SettingsPages/FilesAndFolders.xaml
SketchNI: Nope. Removed.

Files/View Models/ItemViewModel.cs
SketchNI: Apparently not. Reverted.

Files/Converters/InverseBoolConverter.cs
tsvietOK: You can remove it as there is default BoolNegationConverter exists.

SketchNI: Removed.

Files/View Models/SettingsViewModel.cs
```diff
@@ -419,7 +419,7 @@ public bool DoubleTapToRenameFiles
         public bool ShowFileExtensions
         {
             get => Get(true);
-            set => Set(value);
+            set => Set(value);
```
tsvietOK, line 422:
```suggestion
            set => Set(value);
```

tsvietOK:
> Windows 10 version 1903 and later allows you to have files starting with a dot (with or without extension) but previous versions don't allow. Please check the version to prevent crash.

Can you post a source of your statement? As i remember, any Windows 10 version allows you to have dot files, but before W10 1903 you couldn't **create** a dot file.

Jaiganeshkumaran:
@tsvietOK You can always have dot files in Windows but previous Win32 and UWP file APIs didn't support them properly so we shouldn't show them as it will cause issues.

tsvietOK:
Instead of hiding them i think we have to block creating and renaming such files in <1903

Jaiganeshkumaran:
Also sharing

tsvietOK:
@Jaiganeshkumaran  So all Windows version checks could be removed?

Jaiganeshkumaran:
@tsvietOK No don't do that. I meant that share option doesn't work before 1903 for files starting with a dot.

tsvietOK:
I understand, but it is not a part of this PR. And we have to display dot files for all W10 versions, so all checks here should be removed. To disable creating, renaming and sharing dotfiles we have to create another PR.

tsvietOK (approved):

PR #1833, opened 2020-08-27
Added settings to choose multitasking preferences

lukeblevins:
We add a new page in Settings for users to opt out of the default adaptive multitasking behavior:
image


Files/App.xaml.cs
```diff
@@ -34,11 +34,14 @@
 using System.Linq;
 using Newtonsoft.Json;
 using Files.Common;
+using Files.UserControls.MultiTaskingControl;
 
 namespace Files
 {
     sealed partial class App : Application
     {
+        public static IMultitaskingControl multitaskingControl = null;
```
tsvietOK, line 43:
```suggestion
        public static IMultitaskingControl MultitaskingControl = null;
```

duke7553:
Some of the fixes from newer versions of Files aren't in this branch, so there are some bugs present which will be fixed when this is merged.

duke7553:
I also found a crash which occurs on resizing the app window. It only occurs in a debugger though. Can't repro elsewhere. If anyone has any ideas please let me know.

Files/View Models/SettingsViewModel.cs
```diff
@@ -434,16 +436,58 @@ public bool AreLinuxFilesSupported
...
+                Set(value);
+                if (value)
+                {
+                    Window.Current.SizeChanged += Current_SizeChanged;
+                    if (App.InteractionViewModel.IsWindowCompactSize)
+                    {
+                        App.InteractionViewModel.IsVerticalTabFlyoutVisible = true;
+                        App.InteractionViewModel.IsHorizontalTabStripVisible = false;
+                    }
+                    else if (!App.InteractionViewModel.IsWindowCompactSize)
```
tsvietOK, line 453:
```suggestion
                    else
```

tsvietOK (dismissed):

Jaiganeshkumaran:
@duke7553 It might be a XAML issue because I have seen exceptions with XAML thrown when running inside the debugger.

yair100 (dismissed):

tsvietOK:
@yair100 Changing the type of multitasking layout from horizontal to vertical does not change it.

yair100:
> @yair100 Changing the type of multitasking layout from horizontal to vertical does not change it.

I will take a look.

yair100:
> @yair100 Changing the type of multitasking layout from horizontal to vertical does not change it.

I cannot reproduce this issue.

yair100 (approved):

PR #1854, opened 2020-09-01
Improve codebase quality in properties

### LINKED ISSUE #1836: Avoid duplicate code in properties
**Describe the bug**
The section inside the properties dialog which contains the file icon and the file name is implemented in the general view as well as in the shortcut view.
The result of this doubled implementation is that future changes must be implemented twice which may lead to errors.

**Expected behaviour**
Implement this section once e.g. in a custom control

I could fix it with a PR by myself.

Thomas1664:
resolves #1836 
Changes:
- Refactor icon into a custom control
- Add common stylings for property title (left column), property value (right column) & for the property tabs

I tried to create a base style for property values since there are TextBoxes as well as TextBlocks, but I couldn't find a way to use a style for multiple target types. If there is a way to do this, please let me know.

The icon control contains a property called LargerItemSize which is 2 px larger than the original ItemSize because 2 images (CombinedItemsIcon and DriveItemIcon) had a 2px larger FontSize. Since I'm not sure if that is intentional I made its size 2px larger as well.

### DIFF
--- a/Files/UserControls/FileIcon.xaml.cs
+++ b/Files/UserControls/FileIcon.xaml.cs
@@ -1,0 +1,29 @@
+﻿using Files.View_Models;
+using Windows.UI.Xaml.Controls;
+
+namespace Files.UserControls
+{
+    public sealed partial class FileIcon : UserControl
+    {
+        private double itemSize;
+
+        public SelectedItemsPropertiesViewModel ViewModel { get; set; }
+        public double ItemSize { get; set; }
+
+        public double Size
+        {
+            get => itemSize;
+            set
+            {
+                itemSize = value;
+                LargerItemSize = itemSize + 2.0;
+            }
+        }
+        private double LargerItemSize { get; set; }
+        
+        public FileIcon()
+        {
+            this.InitializeComponent();
+        }
+    }
+}
--- a/Files/UserControls/FileIcon.xaml.cs
+++ b/Files/UserControls/FileIcon.xaml.cs
@@ -5,12 +5,10 @@ namespace Files.UserControls
 {
     public sealed partial class FileIcon : UserControl
     {
-        private double itemSize;
-
         public SelectedItemsPropertiesViewModel ViewModel { get; set; }
-        public double ItemSize { get; set; }
 
-        public double Size
+        private double itemSize;
+        public double ItemSize
         {
             get => itemSize;
             set
@@ -20,7 +18,7 @@ namespace Files.UserControls
             }
         }
         private double LargerItemSize { get; set; }
-        
+
         public FileIcon()
         {
             this.InitializeComponent();
--- a/Files/Files.csproj
+++ b/Files/Files.csproj
@@ -443,6 +443,10 @@
       <Generator>MSBuild:Compile</Generator>
       <SubType>Designer</SubType>
     </Page>
+    <Page Include="ResourceDictionaries\PropertiesStyles.xaml">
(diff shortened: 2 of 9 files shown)


gave92:
Not only the xaml but also the code behind is duplicated between PropertiesGeneral.xaml.cs and PropertiesShortcut.xaml.cs. Perhaps you could address this too in the PR?

Thomas1664:
> Not only the xaml but also the code behind is currently duplicated between PropertiesGeneral.xaml.cs and PropertiesShortcut.xaml.cs. Do you plan to address this too in the PR?

@gave92 I dressed it by creating a class which both pages instantiate. My original plan was to let the pages derive from that class but it didn't work because it's a sealed class.

gave92:
@Thomas1664 nice work. You should be able to create a base class deriving from Page and then make PropertiesGeneral and PropertiesShortcut derive from that. 

An example of this is the BaseLayout class from which GenericFileBrowser and GridViewBrowser both derive

Thomas1664:
@gave92 When I try to let PropertiesGeneral derive from PropertiesTab I get the following error message:
CS0263: Partial declarations of "PropertiesGeneral" may not derive from different base classes
I looked at BaseLayout and I can't find out what I did wrong

Thomas1664:
> @Thomas1664 You have to change the class in the xaml file as well. Look at the beginning of GenericFileBrowser.xaml

Thanks @gave92 !

Thomas1664:
@yair100 would you please review this PR?

tsvietOK:
image
Why corners of TextBox is not rounded?

Thomas1664:
I fixed the issue concerning rounded corners. The issue occurred because rounded corners were introduced in [WinUI 2.2](https://docs.microsoft.com/en-us/windows/uwp/design/style/rounded-corner) and because of some reason the new styling has not been implemented correctly. That is also the reason why the default `CornerRadius` inside the properties window in VS says `0,0,0,0`.

tsvietOK:
Progress is not updating when calculating MD5 hash of large file.

Files/UserControls/FileIcon.xaml.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Files.UserControls
+{
+    public sealed partial class FileIcon : UserControl
+    {
+        private double itemSize;
+
+        public SelectedItemsPropertiesViewModel ViewModel { get; set; }
+        public double ItemSize { get; set; }
+
+        public double Size
```
tsvietOK, line 13: Seems like this property is not used, so could you remove it?

tsvietOK:
Getting crash when trying to open properties window for disk root folder.

Files/UserControls/FileIcon.xaml.cs
Thomas1664: Of curse

Thomas1664:
> Getting crash when trying to open properties window for disk root folder. Same for combined properties.

@tsvietOK I can't reproduce the issue. Maybe it is related to a particular version of Windows 10. I'm using version 2004. What version do you use?

tsvietOK:
> > Getting crash when trying to open properties window for disk root folder. Same for combined properties.
> 
> @tsvietOK I can't reproduce the issue. Maybe it is related to a particular version of Windows 10. I'm using version 2004. What version do you use?

The same version.

Thomas1664:
In this case I suggest that you rebuild the project

tsvietOK (approved):

PR #1856, opened 2020-09-01
Show dialog to restart app after language changed

### LINKED ISSUE #1387: Ability to restart automatically when changing language
Currently changing Files UWP's language requires me to manually restart. Would it not be better if it automatically restarts for you?

Thomas1664:
I implemented a dialog that shows up when a restart is required because the language has changed.
Fixes #1387.


Thomas1664:
I'm done with the implementation. 
But the restart doesn't work properly: After the app is restarted, the splash screen shows up and the app closes.
Do I need to restart the app with some special arguments or is this issue related to something else?

yaichenbaum:
@Thomas1664 Check the screenshot in #1387, instead of using a dialog, it's better to go with the design in that screenshot.

Thomas1664:
@yair100 I'm not sure if a user would recognize if we added a bar at the bottom of the page to restart the app. Maybe a better solution would be to display it at the top of the page.

Jaiganeshkumaran:
@Thomas1664 Then you'll need to move all those UI elements. Instead why not send a push notification.

Thomas1664:
I worked on the notification but when I click on a button, a new instance gets opened. But I want the current view to be used. How can I achive this?

Jaiganeshkumaran:
@Thomas1664 You can pass some arguments which will use the existing instance. I don't really how it works.

Thomas1664:
First Draft:

Screenshot (37)

Thomas1664:
I'm done with the UI. But restarting doesn't work properly. The splash screen appears and the app closes for some reason. And there's nothing in the logs about an error. I'm using the following code to perform the restart:
`AppRestartFailureReason failureReason = await CoreApplication.RequestRestartAsync("");`
Can someone help me with that?

Jaiganeshkumaran:
@Thomas1664 You should wait. It will reload all DLLs and other stuff from scratch. It will come by itself.

Thomas1664:
> @Thomas1664 You should wait. It will reload all DLLs and other stuff from scratch. It will come by itself after it has reloaded fully.

I don't think that's the problem. After the splash screen appears, the entire app closes and nothing happens. The doesn't appear in task manager.

Jaiganeshkumaran:
@Thomas1664 Don't do anything. When I tried it in my app, it minimizes to the taskbar and then it comes after a while.

Thomas1664:
@Jaiganeshkumaran It's not related to my impatience that restart doesn't seem to work since the debugging session is closed and nothing - not even in the logs - happens.
However, opening a new instance using the uri `files-uwp://` doesn't work either. I get following exception;
`2020-09-09 17:37:44.9924|ERROR|Files.App|Index was outside the bounds of the array.|System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Files.App.OnActivated(IActivatedEventArgs args)
   at System.Threading.WinRTSynchronizationContextBase.Invoker.InvokeCore()`

Jaiganeshkumaran:
@Thomas1664 Something is wrong with Files UWP. This shouldn't happen. Can you try `LaunchUriAsync(new Uri("files-uwp://" + ResourceController.GetTranslation("NewTab")));` This will open the new tab in the new instance.

Thomas1664:
@Jaiganeshkumaran this throws a `System.UriFormatException`

Thomas1664:
@yair100 Ready for review

Jaiganeshkumaran:
@Thomas1664 Does it work now? From the run box I needed to put the new tab string in my language to open a new instance.

Thomas1664:
@Jaiganeshkumaran Yes. It works now

Files/Strings/en-US/Resources.resw
```diff
@@ -975,6 +975,15 @@
...
     <value>Show all context menu items</value>
   </data>
+  <data name="ButtonNo.Content" xml:space="preserve">
+    <value>No</value>
+  </data>
+  <data name="ButtonYes.Content" xml:space="preserve">
+    <value>Yes</value>
+  </data>
+  <data name="RestartNotificationText.Text" xml:space="preserve">
+    <value>You need to restart the application to apply the changed language settings. Do you want to restart the application now?</value>
```
yaichenbaum, line 985:
```suggestion
    <value>The application needs to be restarted in order to apply the langague setting, would you like to restart the app?</value>
```

yair100 (changes requested):

Files/Dialogs/RestartDialog.xaml.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        public void Dismiss()
+        {
+            RestartNotification.Dismiss();
+        }
+
+        private async void YesButton_Click(object sender, RoutedEventArgs e)
+        {
+            await Launcher.LaunchUriAsync(new Uri("files-uwp://home/page=home"));
```
yaichenbaum, line 28: It is important to restore the session so that the user does not lose their flow.

Files/Views/SettingsPages/Preferences.xaml
```diff
@@ -137,5 +130,7 @@
                 </StackPanel>
             </StackPanel>
         </ScrollViewer>
+
+        <dialogs:RestartDialog x:Name="RestartDialog"/>
```
yaichenbaum, line 134: Can you add a light dismiss to the notification?

Thomas1664: @yair100 I'm gonna file an issue in the WindowsCommunityToolkit repo to fix dark dismiss button on dark theme.

yaichenbaum: I meant light dismiss as in dismissing the notification if you click outside of the notification.

Files/Dialogs/RestartDialog.xaml.cs
Thomas1664: Is there a way to navigate to settings via uri?

Thomas1664:
@yair100 Actually, I don't think it's a good idea to close the notification when a user clicks outside because a user may click outside unintentionally. In that case the notification would be gone.

yair100:
> @yair100 Actually, I don't think it's a good idea to close the notification when a user clicks outside because a user may click outside unintentionally. In that case the notification would be gone.

An in-app notification should be transient and should go away if the user doesn't click on it.

Thomas1664:
> An in-app notification should be transient and should go away if the user doesn't click on it.

@yair100 I could add a timeout so the notification closes after a while

Thomas1664:
@yair100  I tried to dismiss the notification when the control lost its focus. But if you try to click a button on the control the notification gets dismissed as well because the focus is now on the scroll view. Any ideas to fix that?

yair100:
> @yair100 I tried to dismiss the notification when the control lost its focus. But if you try to click a button on the control the notification gets dismissed as well because the focus is now on the scroll view. Any ideas to fix that?

Is there a property called LightDismiss?

Thomas1664:
> Is there a property called LightDismiss?

Unfortunately not.

yaichenbaum:
> > Is there a property called LightDismiss?
> 
> Unfortunately not.

In that case it's best to just hide the notification after about 10 seconds

Files/Dialogs/RestartDialog.xaml.cs
Thomas1664: implemented feature

Files/Views/SettingsPages/Preferences.xaml
Thomas1664: Notification closes after 10 seconds

yair100 (dismissed):

Files/Views/SettingsPages/Preferences.xaml
```diff
@@ -34,14 +35,6 @@
                         x:Uid="SettingsPreferencesAppLanguage"
                         FontSize="17"
                         Text="App Language" />
-                    <TextBlock
-                        x:Name="RestartRequiredPrompt"
-                        x:Uid="SettingsPreferencesAppLanguageRestartRequired"
```
tsvietOK, line 39: Can you remove no more needed translations for this UID?

tsvietOK (approved):

Issue #1424, opened 2020-07-13
OneDrive sync status should appear instantly


Jaiganeshkumaran:
Currently OneDrive sync status only appears after the page is loaded. Instead, it should be handled before it's loaded and before the progress bar is dismissed.

jaigak:
@yair100 This ain't a feature request, so please remove the feature_request tag, added by the bot.

PR #1860, opened 2020-09-02
Display localised names for system folders

### LINKED ISSUE #1226: Show localised (localized) names for system folders
Not really. In File Explorer, it will show the localized name instead of the real name for system folders such as Users folder.

2020-06-25 11-13-34_Trim

Files UWP should show to localized name instead of the real name for system folders.

jaigak:
Implements and closes #1226

திரைப்பிடிப்பு (25)


yair100 (dismissed):

lukeblevins (changes requested):
Unfortunately, retrieving a StorageFolder for each item will cause a massive performance hit. I don't think we should pursue this method because we don't rely solely on StorageItems for a reason.

gave92:
@duke7553 could we put this in the `LoadExtendedItemProperties` method? It won't hurt performance, only issue is that it will briefly show the untranslated name (maybe acceptable)

Jaiganeshkumaran:
@gave92 Actually we can do this from the Win32 component. That's what a tried first.

gave92:
@Jaiganeshkumaran that would be pretty slow as well I think

Jaiganeshkumaran:
@gave92 What should I do then? SHGetLocalizedName is not available for UWP apps.

gave92:
@Jaiganeshkumaran I think the current method (StorageFolder) is good, and performance issues can be solved by moving this to the LoadExtendedItemProperties() method

Jaiganeshkumaran:
@gave92 But then it will not be shown in the folder list.

Jaiganeshkumaran:
We can have it as a optional feature

Jaiganeshkumaran:
@gave92 Since the name is shown on the files list rather than something like the properties, I'm not going to put this in extended properties. Instead we can have it as an option feature.

gave92:
@Jaiganeshkumaran of you put it on LoadExtendedItemProperties it *will* be shown on the file list

Jaiganeshkumaran:
@gave92 But what's the difference?

Jaiganeshkumaran:
Here's an issue with localization. In English (United States) it's 'localized' but in other regions it is 'localised'. I've added a resource file for en-gb however it will use the en-us version even if it's set to some other en other than en-us or en-uk because Windows provides many en regions. Should I add resource files for each one of en region using MAT?

gave92:
@Jaiganeshkumaran nope, it's called from GenericFileBrowser and GridViewBrowser to load some file info (e.g. thumbnail, OneDrive sync status, etc)

Jaiganeshkumaran:
@gave92 But now the names appear after the page has fully loaded which makes it look bad.

gave92:
@Jaiganeshkumaran yeah I don't particularly mind that (better that no translated name and better that slow file list loading) but that's my opinion :)
@duke7553 do you have any thoughts on this?

Jaiganeshkumaran:
#1424

gave92:
@Jaiganeshkumaran yeah this is the cause of that issue as well.

Files/View Models/ItemViewModel.cs
```diff
@@ -1429,10 +1433,14 @@ public async Task AddFolder(StorageFolder folder)
                     return;
                 }
 
+                string itemName = string.Empty;
+                if (!AppSettings.ShowLocalisedName) itemName = folder.Name;
+                else // ItemName will be loaded later to improve perfomance
+
```
gave92, line 1439: Here the translated name is already available -> folder.DisplayName

Files/View Models/ItemViewModel.cs
```diff
@@ -1251,11 +1252,14 @@ private void AddFolder(WIN32_FIND_DATA findData, string pathRoot)
                 systemTimeOutput.Milliseconds,
                 DateTimeKind.Utc);
             var itemPath = Path.Combine(pathRoot, findData.cFileName);
+            string itemName = string.Empty;
+            if (!AppSettings.ShowLocalisedName) itemName = findData.cFileName;
+            else  // ItemName will be loaded later to improve perfomance
```
gave92, line 1257: You can fill the name field with the untranslated name. It will automatically update ones the LoadExtendedItemProperties is called. Might look much better that an empty name.

yaichenbaum:
> We can have it as a optional feature

@Jaiganeshkumaran We should not add an optional feature in this case since it will slow down the app for users who turn it on who might not be aware of the side affects.

jaigak:
@yair100 Then can I close this pull request?

yair100:
> @yair100 Then can I close this pull request?

We would love to include this if you can find a way to implement it without any side effects.

PR #1871, opened 2020-09-04
Update Resources.resw

pixl-8:


yaichenbaum:
Please update the strings in the xaml files as well in order to keep a proper reference to the localized string.

PR #1877, opened 2020-09-06
Removed duplicate extract all menu item

yair100:


Jaiganeshkumaran:
So no built-in ZIP support for Files UWP?

tsvietOK:
@Jaiganeshkumaran It is better to install 7-zip and extract archives from context menu.

Jaiganeshkumaran:
@tsvietOK Why download extra software if you don't use all the features? Those can be installed if you need a lot of features but Explorer built-in only supports ZIP file and Files UWP is designed to be what an enthusiast thinks about the File Explorer and shouldn't be a downgrade.

tsvietOK (approved):

yaichenbaum:
> So no built-in ZIP support for Files UWP?

@Jaiganeshkumaran Built in support is still there, this just removes the method we implemented in favor of the system one since the system version is more reliable. In the future we may add back our own implementation in order to have more control ad add support for more archive types but at the moment the system version does a better job.

jaigak:
@yair100 No the system version uses a outdated UI. Instead use Windows.Storage.Compression for now which does the same thing but we can have a modern UI

yair100:
> @yair100 No the system version uses a outdated UI. Instead use Windows.Storage.Compression for now which does the same thing but we can have a modern UI

The previous implementation had multiple issues however, I am open to reverting the change and switching to  Windows.Storage.Compression to get the fluent UI. However, until that work is done, we should hide the duplicate item to remove confusion.

Jaiganeshkumaran:
Instead of reverting the change, create a new PR for updated implementation.

Jaiganeshkumaran:
Here's a sample on how to use it. You may even make it permanent however it only supports ZIP: https://github.com/microsoft/Windows-universal-samples/tree/master/Samples/Compression

yaichenbaum:
> Instead of reverting the change, create a new PR for updated implementation later

I don't plan to work on this for a while, so I figured to revert the change, so the foundation would be there in case someone wants to work on it. Either way is fine since it shouldn't be so hard to build the dialog again.

PR #1881, opened 2020-09-06
Fixed a crash that would sometimes occur when selecting some options on the right click context menu

### LINKED ISSUE #1869: Shell context menu items vanished after hitting 'Pin to Start'
**Describe the bug**
After hitting 'Pin to Start' shell context menu item, all shell context items gets removed for all items.

**To Reproduce**
1. Right-click on a folder
2. Click on 'Pin to Start'
3. Right-click on any other item
4. Shell context menu will be removed until you restart the app

**Expected behavior**
Items shouldn't get removed. Also pin to start option isn't working

**Desktop:**
 - OS Version: Windows 10 1903 18362.175
 - App version: v0.14.0

gave92:
Fixes #1869

This PR hides the "pin to start" context menu option which is not working. Also it avoids fulltrust process crash in case the same happens with other menu items.

### DIFF
--- a/Files.Launcher/Program.cs
+++ b/Files.Launcher/Program.cs
@@ -334,7 +334,7 @@ namespace FilesFullTrust
         {
             var knownItems = new List<string>() {
                 "opennew", "openas", "opencontaining", "opennewprocess",
-                "runas", "runasuser", "pintohome",
+                "runas", "runasuser", "pintohome", "PinToStartScreen",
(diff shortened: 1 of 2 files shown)


gave92:
@Jaiganeshkumaran I can confirm that builds from Azure display the issue you reported. I changed the log to use NLog instead of Debug.WriteLine as you suggested. This should give us some info.

yaichenbaum (approved):

PR #1962, opened 2020-09-22
Added a details page to the properties dialog

winston-de:
This pull request adds a details page to the properties window. 
Screencaps:
image
image


builds/azure-pipelines.yml
```diff
@@ -21,6 +21,12 @@ steps:
       $xmlDoc.Package.Applications.Application.VisualElements.DisplayName="Files UWP - Preview"
       $xmlDoc.Save('$(Build.SourcesDirectory)\Files.Package\Package.appxmanifest')
     failOnStderr: true
+    
+- task: DownloadSecureFile@1
+  name: mapsDevKey
+  displayName: 'Download Bing Maps Dev Key'
+  inputs:
+    secureFile: 'BingMapsKey.json'
```
yaichenbaum, line 29:
```suggestion
    secureFile: 'BingMapsKey.json'
    
- task: PowerShell@2
  inputs:
    targetType: 'inline'
    script: |
      Copy-Item -Path $(mapsDevKey.secureFilePath) -Destination $(Build.SourcesDirectory)\Files
```

Files.Package/Package.appxmanifest
```diff
@@ -9,7 +9,7 @@
          xmlns:uap4="http://schemas.microsoft.com/appx/manifest/uap/windows10/4"
          xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5"
          IgnorableNamespaces="uap uap5 mp rescap desktop4 desktop">
-    <Identity Name="FilesDev" Publisher="CN=53EC4384-7F5B-4CF6-8C23-513FFE9D1AB7" Version="0.14.0.0" />
+    <Identity Name="FilesUWPDev" Publisher="CN=53EC4384-7F5B-4CF6-8C23-513FFE9D1AB7" Version="0.14.6.0" />
```
tsvietOK, line 12: I guess this file should not be changed.

yaichenbaum: This was just for testing purposes and it can be reverted before merging the pull request.

winston-de:
Although functional, this feature is not necessarily finished. I intend to add more properties and features in the near future, I'm just busy at the moment and didn't want this branch to far to far behind. 

@yair100 If you have any more design suggestions please let me know.

**Current property groups supported:**
- Core
- Image
- Media
- Photo
- GPS
- Media
- Audio
- Music

**Features:**
- Read and write capabilities
- Clear all personal properties
- Overview page for image files
- GPS coordinates from properties are shown as an address

**Planned features**
- Support for properties that require combo boxes
- Support for more controls, like date-time pickers
- Warning when properties failed to save
- Design improvements
- More properties
- Windows narrator support*

Screencaps:
image*

*Edit

yaichenbaum:
Editing the properties and clicking "OK" doesn't seem to be saving them.

tsvietOK (dismissed):
image
Wrong photo properties: Camera, shot.
Here is a right one:
image

tsvietOK:
@winiston2212 I think it is better to move details properties to DataGrid control. It is not good to define all these properties inside both xaml and xaml.cs files.
image
I would suggest to do the following things:
Create class like `DetailsProperty` with properties `Name`, `Value` and `IsReadOnly`. 
Create class with lists of `DetailsProperty`: `CoreProperties`, `PhotoProperties`, `ImageProperties`, `GpsProperties`, `AudioProperties`, `MusicProperties`, `MediaProperties`, `PersonalProperties`.
When you get values for these properties just bind lists to the `DataGrid` control. If it is not possible to create one DataGrid for all types of properties, create several controls of this type.

winston-de:
Thanks for the feedback! 

@tsvietOK 
The main reason I went this method is to have flexibility with controls. Many properties use combo boxes, date time pickers, list editors, number boxes, and other custom controls. Many number properties also rely on separate numerator and denominator properties for editing, which would require more customized controls. I was afraid that by relying on a DataGrid, ListView, or similar control it would be significantly harder to add editing capability to these properties in the future .

> Wrong photo properties: Camera, shot.
For the camera, what is displayed is both the camera manufacturer and the camera model properties (see below). It appears that your camera puts "Canon" in both the manufacturer and model properties. 

image
image

I fixed the way the values for the shot properties were displayed (they were mixed). However, I don't think the numbers aren't entirely accurate. I believe this is an issue with the way Windows Storage calculates properties that have numerator and denominator properties. For example: 
> This property is calculated from System.Photo.ApertureNumerator and System.Photo.ApertureDenominator.

https://docs.microsoft.com/en-us/windows/win32/properties/props-system-photo-aperture

I believe this can be worked around by calculating the values manually instead of relying on Windows storage to do it.

@yair100 I switched to a more reliable (but slower) property save system, saving should work properly now.

tsvietOK:
image
Example of different elements inside DataGrid

mdtauk:
image

The design for the Properties Dialog is pretty closely mirroring the existing Win32 one.

Did you consider re-thinking what the design **should** be like?  Or did you just want to make it familiar and recognisable?

winiston2212:
@tsvietOK
>Example of different elements inside DataGrid

Yeah, but you'd end up with a giant list defining all the properties, like this: 

image

which I didn't think would be much better than doing it in the XAML file. If it's better to do it programmatically, I can.

@mdtauk 
> The design for the Properties Dialog is pretty closely mirroring the existing Win32 one.
> Did you consider re-thinking what the design should be like? Or did you just want to make it familiar and recognisable?

I wasn't the one who designed the page you showed, so I wouldn't know. The only design change that was made (in this PR), is transparent backgrounds for text boxes when they aren't selected or hovered.

image

tsvietOK:
@winiston2212 Yes this list is quite big, but at least all properties are grouped in one place, instead of a lot of elements in both xaml and cs files. So yes, it is better to do it programmatically. 
That's what i get for now:
image

duke7553:
@tsvietOK Good idea. Though, perhaps we can use a ListView instead of DataGrid as we're now trying to distance future experiences from the Toolkit DataGrid.

duke7553:
@winiston2212 I think it's not important at this point to use a List control whether ListView or DataGrid, as we can iterate here later on in another PR.

duke7553:
@winiston2212 Again, I really appreciate your work on this. 🥇 I plan to conduct my testing and code review on this PR tomorrow.

winiston2212:
@duke7553 Alright, sounds good.

For the future, I'm working on a program to generate a .cs file with the list of properties and their information, making it easier to switch.
https://github.com/winiston2212/PropertiesListGenerator

tsvietOK:
Do we really need to display all [available](https://docs.microsoft.com/en-us/windows/win32/properties/props) file properties? I think the main goal is creating ListView with different controls inside it with ability to edit properties. So if you create this control the right way, there will not a problem to add more file properties.

winiston2212:
@tsvietOK Just the ones that the current File Explorer currently shows, no?

tsvietOK:
@winiston2212 I just want to say, that first of all  we have to create a control, which will hold, for example, 10 properties. So we can improve control design and avoid unnecessary actions (like creating your code generator). It is not hard to add all properties to the code.

winiston2212:
@tsvietOK There are only like half a dozen different property types: String, Double, UInt32, Multivalue Strings, DateTime, the occasional enum (which are normally UInt32s), and Bytes. Wouldn't it be easier if separate controls and converters were created for each type, then the generator determined which ones to use based off of the type of property? The generator can also fetch the text to use for enums. Then all that would have to be done is correct any mistakes the generator made. A generator is how I did most of the XAML code.

yaichenbaum:
@mdtauk Right now the goal is to implement something functional with a touch of fluent. There was an issue about rethinking the properties dialog but I can't find it, if you have any ideas, feel free to start a new issue so everyone can chime in.

mdtauk:
> @mdtauk Right now the goal is to implement something functional with a touch of fluent. There was an issue about rethinking the properties dialog but I can't find it, if you have any ideas, feel free to start a new issue so everyone can chime in.

Nothing especially comes to mind, I was just bringing up the question - as this issue was talking about the details, and that could spiral out of control depending on what metadata is populated.

winiston2212:
> Nothing especially comes to mind, I was just bringing up the question - as this issue was talking about the details, and that could spiral out of control depending on what metadata is populated.

@mdtauk If you think of anything, please tell me.

Files/Filesystem/ListedItem.cs
```diff
@@ -150,6 +150,16 @@ public DateTimeOffset ItemDateAccessedReal
 
         private DateTimeOffset _itemDateAccessedReal;
 
+        public bool IsImage()
+        {
+            if (FileExtension != null)
+            {
+                string lower = FileExtension.ToLower();
+                return lower.Contains("png") || lower.Contains("jpg") || lower.Contains("png") || lower.Contains("gif") || lower.Contains("jpeg");
```
yaichenbaum, line 158: It looks like you included `png` twice.

yair100 (dismissed):

yair100 (approved):

lukeblevins (approved):

yaichenbaum:
:shipit:

PR #1974, opened 2020-09-27
Fix properties navigation background

### LINKED ISSUE #1689: Acrylic background is not shown for the navigation view pane in properties window
**Describe the bug**
Acrylic background is not shown for the navigation view pane in file properties window. It instead shows the theme background.

**To Reproduce**
1. Go to the properties window of a file or folder
2. Drag window over a color that contrasts with your theme
3. The non-acrylic navigation view will be obvious

**Expected behavior**
The background of the navigation view pane should be acrylic to match the rest of the window.

**Screenshots**
image

**Desktop:**
 - OS Version: [Windows 10 2004 20185.1000]
 - App version: [v12.0.0]

Thomas1664:
This PR doesn't fix #1689 immediately because `NavigationView` doesn't allow you to override the background if `PaneDisplayMode` is set to `Top`. This bug is related to WinUI. For more information, see this [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/3299). However, you can see that this fix works if you set `PaneDisplayMode` to `Left`.

### DIFF
--- a/Files/Views/Pages/Properties.xaml
+++ b/Files/Views/Pages/Properties.xaml
@@ -43,7 +43,7 @@
             PaneDisplayMode="Top"
             SelectionChanged="NavigationView_SelectionChanged"
             SelectionFollowsFocus="Enabled">
-            
+
             <!--  Tabs  -->
             <muxc:NavigationView.MenuItems>
                 <muxc:NavigationViewItem
--- a/Files/Views/Pages/Properties.xaml
+++ b/Files/Views/Pages/Properties.xaml
@@ -14,6 +14,11 @@
     Unloaded="Properties_Unloaded"
     mc:Ignorable="d">
 
+    <Page.Resources>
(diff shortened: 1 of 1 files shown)


yaichenbaum:
@Thomas1664 Starting with #1962, we are going to use a pivot instead, @winiston2212 can you confirm if switching to a pivot will solve this issue?

winston-de:
@yair100 I was in the process of switching to a Pivot View, but stopped after realizing that tabs cannot be hidden using a Pivot View (this would cause a crash if the user tried to access a tab like shortcut settings after selecting multiple files). Sorry for the confusion.

yair100 (approved):

marcelwgn:
@yair100 @Thomas1664 NavigationView uses the "NavigationViewTopPaneBackground" resource for the background of the NavigationView when display mode is set to top. It is possible to customize the background through that resource similar to "NavigationViewDefaultPaneBackground".

yaichenbaum:
@chingucoding Thank you for letting us know, also great to see you here 😃.

PR #2007, opened 2020-09-30
Improved the codebase quality in AddItemDialog

Thomas1664:
- Refactoring of the following methods: `AddItemsToList`, `ListView_ItemClick`, `CreateFile`
- Added translation strings & translated them to German
- Fixed an issue that didn't adjust the background of the dialog after a different theme was selected

Maybe we should consider to create some common stylings for all dialogs?


lukeblevins:
@yair100 Were we still planning to make improvements to the add item dialog or deprecate it entirely? I wasn't aware of this still being in use.

yaichenbaum:
@duke7553 I believe the plan is to completely deprecate it, however, I don't think there is an issue making improvements for the current solution until that happens.

Thomas1664:
@tsvietOK Thanks for your assistance on fixing merge errors!

tsvietOK (approved):

yair100 (approved):

PR #2030, opened 2020-10-05
Switch to inactive tab when files are dragged over for a certain duration

### LINKED ISSUE #1968: Ability to drag files between tabs in the same window
Currently it's impossible to drag a file from one tab to another in the same window.

Make inactive tabs into valid drop targets, and also activate the target tab after hovering over it for a certain period of time (similar to how you can hover a file over an opened application on the taskbar and the window will activate).

jeffsieu:
Dragging files over a tab will switch to it after 0.5s

Uses a `DispatcherTimer` with a 500ms interval. The timer is reset if the files are dropped into the tab, or if the tab is not longer dragged over.

Resolves #1968

### DIFF
--- a/Files/UserControls/MultitaskingControl/VerticalTabViewControl.xaml.cs
+++ b/Files/UserControls/MultitaskingControl/VerticalTabViewControl.xaml.cs
@@ -25,14 +25,19 @@ namespace Files.UserControls
 
         public const string TabPathIdentifier = "FilesTabViewItemPath";
 
-        public VerticalTabViewControl()
-        {
-            this.InitializeComponent();
-        }
-
         public void SelectionChanged() => TabStrip_SelectionChanged(null, null);
 
         public ObservableCollection<TabItem> Items => MainPage.AppInstances;
+
+        private readonly DispatcherTimer tabHoverTimer = new DispatcherTimer();
+        private TabViewItem hoveredTabViewItem = null;
+
+        public VerticalTabViewControl()
+        {
+            this.InitializeComponent();
+            tabHoverTimer.Interval = TimeSpan.FromMilliseconds(500);
+            tabHoverTimer.Tick += TabHoverSelected;
+        }
 
         public async Task SetSelectedTabInfo(string text, string currentPathForTabIcon = null)
         {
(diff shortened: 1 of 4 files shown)


yaichenbaum:
@jeffsieu Are you sure this resolves #1968?

jeffsieu:
@yair100 I am able to use an inactive tab as a drag target, and also able to switch to it after hovering over it. I just added some changes that disables the tab reordering when hovering files over. It seems to have eliminated that bug mentioned in #1968.

yaichenbaum:
@jeffsieu Is this for the vertical tab strip as well?

jeffsieu:
@yair100 Just added support for the vertical strip as well. But I notice there's a lot of duplicated code in horizontal and vertical tab views.

yaichenbaum:
> But I notice there's a lot of duplicated code in horizontal and vertical tab views.

I was not aware of this, but it's something to work on.

yair100 (dismissed):

tsvietOK:
@jeffsieu Would it better to disable tab reorder when you drag a `StorageItem`?
```
private void TabViewItem_DragOver(object sender, DragEventArgs e)
        {
            if (e.DataView.AvailableFormats.Contains(StandardDataFormats.StorageItems))
            {
                HorizontalTabView.CanReorderTabs = false;
                e.AcceptedOperation = DataPackageOperation.Move;
            }
            else
            {
                HorizontalTabView.CanReorderTabs = true;
                e.AcceptedOperation = DataPackageOperation.None;
            }
        }
```

jeffsieu:
@tsvietOK Hmm, isn't that what I'm currently doing in the PR? Or did I miss something out?

tsvietOK:
@jeffsieu Here is a difference: i am changing HorizontalTabView.CanReorderTabs value only in one method.

jeffsieu:
@tsvietOK Oh, I've tried this, but it doesn't work as dragging the items above the tab bar causes the reordering to happen. So `TabViewItem_DragOver` won't even trigger, but the re-ordering animation will have started. We need to disable the dragging the moment the cursor enters the tab bar.

Files/UserControls/MultitaskingControl/VerticalTabViewControl.xaml.cs
```diff
@@ -234,20 +239,39 @@ private void TabViewItem_Drop(object sender, DragEventArgs e)
...
+                tabHoverTimer.Start();
+                hoveredTabViewItem = sender as TabViewItem;
             }
             else
             {
                 e.AcceptedOperation = DataPackageOperation.None;
             }
         }
 
+        private void TabViewItem_DragLeave(object sender, DragEventArgs e)
```
tsvietOK, line 261: Should this method be used?

jeffsieu: Are you suggesting that we can use only `TabViewItem_DragEnter`?

jeffsieu: But won't that cause problems if an item is dragged over, then away from the TabViewItem? This would disable the reordering of the tabs even though it would be a valid operation.

tsvietOK: No, i am saying, that in `HorizontalMultitaskingControl.xaml` `TabViewItem_DragLeave` is used, but in `VerticalTabViewControl.xaml` this method is missing.

jeffsieu: Oh I see, sorry I missed that out. Thanks for pointing it out!

tsvietOK (approved):

yaichenbaum:
@jeffsieu Great work!

PR #2037, opened 2020-10-05
Fixed two issues with keyboard navigation

### LINKED ISSUE #2020: Unable to invoke certain focused buttons with Enter
**Describe the bug**
Focusing most buttons in the app with the tab key works, but users have trouble selecting/invoking some buttons with the enter key. 

**To Reproduce**
1. Press tab until a status bar button on the bottom right side are highlighted
2. Attempt to press enter to invoke the attached Flyout

**Expected behavior**
Users should be able to change the app layout, access selection options, and view status of file operations without using the mouse.

 - OS Version: Windows 10
 - App version: v15.4

**Additional context**
This incorrect behavior is caused by the layout modes intercepting key presses including the Enter key.

### LINKED ISSUE #2036: Unable to use TAB key to select all sidebar items
**Describe the bug**
Some items are not registered as a tab stop

**To Reproduce**
Reproducible by trying to use the TAB key to focus some items

**Expected behavior**
All items should be invokable via keyboard

duke7553:
Fixes #2020 
Fixes #2036

### DIFF
--- a/Files/UserControls/SidebarControl.xaml
+++ b/Files/UserControls/SidebarControl.xaml
@@ -2,21 +2,14 @@
     x:Class="Files.Controls.SidebarControl"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:Windows10version1903="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract, 8)"
-    xmlns:contract4NotPresent="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractNotPresent(Windows.Foundation.UniversalApiContract,4)"
-    xmlns:contract4Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,4)"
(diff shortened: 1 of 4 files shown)


yaichenbaum (dismissed):

Thomas1664 (changes requested):
In line 77 in the file `SidebarControl.xaml` `IsTabStop` is also set to `False`. Can you fix that as well?

yaichenbaum:
> In line 77 in the file SidebarControl.xaml IsTabStop is also set to False. Can you fix that as well?

@duke7553 Can you look at this?

duke7553:
@Thomas1664 This property value doesn't affect functionality, does it?

Jaiganeshkumaran:
@duke7553 It does. Disabling tab stop will prevent keyboard tab navigation for the UI element

duke7553:
@Jaiganeshkumaran @Thomas1664  Sorry, but what you said is incorrect for this case.

The changes I already made allow for using tab because tab stop is enabled for the items themselves already. Unless you want to have a discussion about removing the custom template for these sidebar items altogether (which I'd support), there is no need to change the template value because it already works fine with my changes.

Would enabling the template value for the tab stop give us any more benefits?  In my testing, the sidebar items already work as intended.

Thomas1664:
@duke7553 Although your changes already work, it would be better for code style if you removed the remaining `IsTabStop`. Furthermore, it may save us from future bugs.

duke7553:
@Thomas1664 Since the template itself was redundant, I just got rid of it so we use the standard resources now.

yaichenbaum (approved):

Thomas1664:
> @Thomas1664 Since the template itself was redundant, I just got rid of it so we use the standard resources now.

Sonds great

PR #2058, opened 2020-10-09
Added AutomationProperties for the preferences page in settings

yaichenbaum:


Jaiganeshkumaran:
Why not use the header property instead of a separate text block. You can actually customize the font size. Setting as header creates a connection between the box and the text. Or you can use AutomationProperties.DescribedBy and fill the describing text block. Don't set AutomationProperties.Name for the combo box itself which may cause confusion since it will read both the value without seperation.

yaichenbaum:
@Jaiganeshkumaran I would like to get this merged as soon as possible but afterwards, I will take a look at doing what you suggested in order to simplify the code.

duke7553 (approved):

PR #2069, opened 2020-10-11
Make optional/move repetitive checks of property values outside item enumeration loop

duke7553:
Slightly improve CPU usage by making some property checks less repetitive in the enumeration loop by being optional arguments.

### DIFF
--- a/Files/Filesystem/ListedItem.cs
+++ b/Files/Filesystem/ListedItem.cs
@@ -160,21 +160,73 @@ namespace Files.Filesystem
             return false;
         }
 
-        public ListedItem(string folderRelativeId)
+        /// <summary>
+        /// Create an item object, optionally with an explicitly-specified dateReturnFormat.
+        /// </summary>
+        /// <param name="folderRelativeId"></param>
+        /// <param name="dateReturnFormat">Specify a date return format to reduce redundant checks of this setting.</param>
+        public ListedItem(string folderRelativeId, string dateReturnFormat = null)
         {
             FolderRelativeId = folderRelativeId;
+            if (dateReturnFormat != null)
+            {
+                DateReturnFormat = dateReturnFormat;
+            }
+            else
+            {
+                ApplicationDataContainer localSettings = ApplicationData.Current.LocalSettings;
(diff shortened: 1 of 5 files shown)


Files/Filesystem/ListedItem.cs
```diff
@@ -160,21 +160,73 @@ public bool IsImage()
...
-        public static string GetFriendlyDate(DateTimeOffset d)
+        private string DateReturnFormat { get; }
+
+        private string GetFriendlyDate(DateTimeOffset d)
         {
-            ApplicationDataContainer localSettings = ApplicationData.Current.LocalSettings;
             var elapsed = DateTimeOffset.Now - d;
 
-            string returnformat = Enum.Parse<TimeStyle>(localSettings.Values[LocalSettings.DateTimeFormat].ToString()) == TimeStyle.Application ? "D" : "g";
+            if (elapsed.TotalDays > 7)
```
yaichenbaum, line 189: Can you use a switch statement here?

duke7553: This is the way the code already was. And switch statements lack support for conditionals

Files/View Models/ItemViewModel.cs
```diff
@@ -306,7 +306,7 @@ public ItemViewModel()
             FilesAndFolders = new ReadOnlyObservableCollection<ListedItem>(_filesAndFolders);
             _addFilesCTS = new CancellationTokenSource();
             _semaphoreCTS = new CancellationTokenSource();
-
+            shouldDisplayFileExtensions = App.AppSettings.ShowFileExtensions;
```
tsvietOK, line 309: Why you changed one bool to another one?

duke7553: Because it is apparently redundant to check the value of the property for each file being added. What this does is check the value of that property from AppSettings only when it would need to be updated.

duke7553: A lot of CPU time was being spent for each file by resolving the Get() method for App.AppSettings.ShowFileExtensions. This stores that property's value and updates it once per enumeration rather than once per file.

yaichenbaum (approved):

tsvietOK (approved):

PR #2088, opened 2020-10-14
Hungarian translate

KristofKekesi:
## Files to hungarian
I've made a folder called 'hu-HU' inside Strings folder. I copied the Resources.resw from the english translation's folder. I translated most of the strings. I will finish it soon.

### Things to do
I saw some settings like values (at the beginning of the file) and I don't know what to do with them. 
I don't know how to program in C so I would be glad is someone would connect it to the backend.

I hope my translations will be at Files someday.
It is getting way better, congratulations to all.


KristofKekesi:
I used an old version of Resources.resw. I will fix this as soon as possible.

tsvietOK:
Don't translate app through `Resources.resw` file, this file will be generated automatically from file in `MultilingualResources/Files.hu-HU.xlf`. So translate it in `Files.hu-HU.xlf` when it will be added to the main branch.

KristofKekesi:
Oke, thanks :D

tsvietOK:
Now you can open `MultilingualResources/Files.hu-HU.xlf` and translate strings by changing value of `target` tag and don't forget to change state from `new` to `translated` for translated strings.

Issue #1212, opened 2020-06-23
Sidebar: add "properties" in context menu of drives


XPoppyX:
![Captura de pantalla (33)](https://user-images.githubusercontent.com/61259627/85471740-e292ce00-b57e-11ea-852f-3fcabb303a45.jpg)

tobias-mayer:
What about displaying it for every item in the Sidebar, just like in Windows explorer?

Thomas1664:
@yaichenbaum Can you assign me?

yaichenbaum:
@Thomas1664 Done 👍.

Thomas1664:
What about adding (almost) every context menu item to each sidebar item (except recycle bin and home)?

yaichenbaum:
> What about adding (almost) every context menu item to each sidebar item (except recycle bin and home)?

That should work.

Thomas1664:
@yaichenbaum Adding properties to drives works. However, it does not work for pinned folders because no properties are shown.

PR #2089, opened 2020-10-14
Added properties to the context menu for items in the sidebar

Thomas1664:
This PR implements #1212 
In this issue, it was suggested to add properties for pinned items as well. I implemented this capability but the `FolderProperties` view model causes several issues. I fixed a thrown exception when the use is on the home page. But I was not able to fix an issue where the properties of the current folder in the file browser are shown.
As a result, implementing properties context menu for pinned items requires breaking to the `FolderProperties` view model which are beyond the range of this PR.

### DIFF
--- a/Files/View Models/Properties/FolderProperties.cs
+++ b/Files/View Models/Properties/FolderProperties.cs
@@ -80,7 +80,17 @@ namespace Files.View_Models.Properties
             }
 
             StorageFolder storageFolder;
-            var isItemSelected = await CoreApplication.MainView.ExecuteOnUIThreadAsync(() => App.CurrentInstance.ContentPage.IsItemSelected);
+            bool isItemSelected;
+
+            try
+            {
+                isItemSelected = await CoreApplication.MainView.ExecuteOnUIThreadAsync(() => App.CurrentInstance.ContentPage.IsItemSelected);
+            }
+            catch
+            {
+                isItemSelected = true;
+            }
+
             if (isItemSelected)
             {
                 storageFolder = await ItemViewModel.GetFolderFromPathAsync(Item.ItemPath);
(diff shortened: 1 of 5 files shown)


Files/View Models/Properties/FolderProperties.cs
```diff
@@ -80,7 +80,17 @@ public async override void GetSpecialProperties()
...
 
             StorageFolder storageFolder;
-            var isItemSelected = await CoreApplication.MainView.ExecuteOnUIThreadAsync(() => App.CurrentInstance.ContentPage.IsItemSelected);
+            bool isItemSelected;
+
+            try
+            {
+                isItemSelected = await CoreApplication.MainView.ExecuteOnUIThreadAsync(() => App.CurrentInstance.ContentPage.IsItemSelected);
+            }
+            catch
```
gave92, line 89: @Thomas1664 I'm curious, what kind of exception is thrown here?

Thomas1664: @gave92 It throws a `System.NullReferenceException` because `ContentPage` is `null` if you are on `Home`.

duke7553 (approved):
LGTM

gave92:
@Thomas1664 What was the issue with showing the properties for the pinned folders?
If I enable it on master it looks like it works just fine.

Thomas1664:
> @Thomas1664 What was the issue with showing the properties for the pinned folders?
> If I enable it on master (`ShowProperties = true;`) it looks like it works just fine.

@gave92 
Does it also work when you are on homepage?
Are the properties for the correct folder displayed?

gave92:
@Thomas1664 
Does it also work when you are on homepage? -> yeah, no crash
Are the properties for the correct folder displayed? -> looks like it
Perhaps you can confirm this

Thomas1664:
@gave92 Unfortunately, it doesn't work. Although the app doesn't crash, there's still an exception thrown which isn't ideal and the properties which are displayed are from the current folder and not from the selected folder in sidebar

gave92:
@Thomas1664 argh it works for me xD

Issue #1452, opened 2020-07-14
Application date format not corresponded with selected format in settings


osieltorres:
**Describe the bug**

**To Reproduce**
1. Go to Appearance settings
2. Click on Date format and select "System" option
4. When return to list of files and folders, some appears with date base in "Application" option, and some on selected "System" option.

**Expected behavior**
Date for files and folders should be formatted to selected option in Appearance settings.

**Screenshots**
![2020-07-14_115148](https://user-images.githubusercontent.com/3835402/87447522-af18f180-c5c8-11ea-9881-0a2916401925.png)
![2020-07-14_115321](https://user-images.githubusercontent.com/3835402/87447526-b04a1e80-c5c8-11ea-8de6-f337e719716d.png)

**Desktop (please complete the following information):**
 - OS Version: [e.g. Windows 10 2004 19041.329]
 - App version: [v0.12.0]

Thomas1664:
I can confirm that issue. The changed date format is applied to none of the list items. The same bug appears in properties as well

yaichenbaum:
This issue should be fixed with #2092.

PR #2092, opened 2020-10-14
Fixed an issue with the way dates were displayed in the app

### LINKED ISSUE #2091: Wrong date format for dates that are more than a week ago
**Describe the bug**
When date format is set to `System`, the date format is everywhere wrong for dates that are more than a week ago. 

**To Reproduce**
1. Set date format to `System` in settings
2. Create a new item or view items that are created less than a week ago

**Expected behavior**
Dates should be in correct format

Thomas1664:
Fixes #2091 , #1452


duke7553 (approved):
@Thomas1664 Excellent work!

PR #2130, opened 2020-10-21
Added a setting to open items with a single click

### LINKED ISSUE #261: Enable us to open everything with just one click
Especially for touchscreens it would be nice beeing able to open every file/folder with just one click.

This and other UWP apps are ment to be also used by devices with touchscreen, so I think the option to open files and folders with just one click would be really nice.

Maybe for this also selecting through pulling would make sense?
image

**Additional context**
In file explorer the option to open everything with just one click can be found in file explorer settings:
image

yair100:
Closes #261.


lukeblevins (changes requested):
Is the goal here to make this experiment only work with mouse input? We should make this flag universal because users will have a greater propensity to turn it on for touch screen devices. Double tapping every time isn't exactly practical on touch, so that is what this PR should accomplish imo

Files/Interacts/Interaction.cs
```diff
@@ -58,6 +58,28 @@ public void List_ItemClick(object sender, DoubleTappedRoutedEventArgs e)
             OpenSelectedItems(false);
         }
 
+        public async void List_ItemPress(object sender, PointerRoutedEventArgs e)
+        {
+            if (e.Pointer.PointerDeviceType == Windows.Devices.Input.PointerDeviceType.Mouse)
+            {
```
duke7553, line 64: We'd want to have this condition handle touch input instead of just Mouse.

Files/Views/LayoutModes/GenericFileBrowser.xaml
```diff
@@ -614,6 +615,7 @@
             IsRightTapEnabled="True"
             ItemsSource="{x:Bind AssociatedViewModel.FilesAndFolders, Mode=OneWay}"
             LoadingRow="AllView_LoadingRow"
+            PointerPressed="{x:Bind local:App.CurrentInstance.InteractionOperations.List_ItemPress}"
```
duke7553, line 618: I think PointerPressed on DataGrid will register the event for the entire control rather than just the rows themselves. We should try to catch situations where the user may invoke something other than a Row this way.

yaichenbaum: I cannot reproduce any issues around this.

Files/Interacts/Interaction.cs
yaichenbaum: Thank you for the feedback, this should be resolved.

gave92:
A couple of curiosities and notes:
- Is the plan to have selection on hover?
- Will the user still be able to drag the files?
- I would suggest to test this after merging with master as the new selection rectangle also uses DataGrid's PointerPressed event
- Would it be better trigger file open on PointerReleased instead (same as Explorer)?

yaichenbaum:
> Is the plan to have selection on hover?

Yes.

> Will the user still be able to drag the files?

The user should be able to, but this needs to be tested to make sure.

> Would it be better trigger file open on PointerReleased instead (same as Explorer)?

Do you know if this would have any benefits?

gave92:
> Would it be better trigger file open on PointerReleased instead (same as Explorer)?
I think the current implementation might have issues with the selection rectangle and/or the ability to drag files (not tested, just a guess)
If this is the case, triggering file open on Release might help to differentiate between a click to open file, a file drag or rectangle selection

yaichenbaum:
@gave92 After checking File Explorer it seems that there is no selection on hover when the open item with a single click option is turned on.
> Will the experimental option "Double Tab to rename" remain when we add this?

It will for now but we will probably remove it in the future as it's only a temporary workaround for a bug in the DataGrid.

gave92:
@yair100 
>After checking File Explorer it seems that there is no selection on hover 

Actually I think there is:
- if you hover over a file with the mouse and press "F2" you will be able to rename that file
- if you hover over a file with the mouse and press "Enter" that file will open

This does not happen when single click option is not enabled

yaichenbaum:
@gave92 You are correct, I was thinking about the selection rectangle on hover. The plan is to include selection on hover however I don't plan on it being part of this pull request.

gave92:
> I was thinking about the selection rectangle on hover

This would be triggered when the user pressed e.g the Ctrl key and moves the mouse?

yaichenbaum:
> This would be triggered when the user pressed e.g the Ctrl key and moves the mouse?

Yes, it needs to be implemented as well but that is out of scope for this pull request.

gave92:
Ok I ended up doing more changes than I expected. ~Let me know if you require this to be split into different PRs.~
As per @yair100 request this PR now only contains changes specifically related to the "open files with one click" feature
- e62c62336cb40f24f4cbe3ab950686408ecc91b5 fixes the few issues I found with the "open files with one click" feature
    - Fixes multiple selection by Cltr/Shift + click
    - Fixes files opening during a "rectangle selection" operation
    - Fixes files opening instead of dragging

yaichenbaum:
@gave92 Lets separate the changes into separate pull requests in case one of them needs to be reverted later.

gave92:
I've found two further issues in ListView:
- If you drag an item quickly over a folder, the folder will open
- If you click on the column headers to reorder, the selected items will open

lukeblevins (approved):
Approved as per contributor discussions

gave92 (approved):

PR #2196, opened 2020-11-01
Added a drives widget

### LINKED ISSUE #2188: Add a drives widget to the home page
I have a feature request, so i am bypassing template.

I cant select "This PC"  in startup settings, it says "you cant open this location in this app...." so please add option to use this location on startup, because sometimes i want to see how much space is used or free in each drive.

yair100:
Fixes #2188.

### DIFF
--- a/Files/Files.csproj
+++ b/Files/Files.csproj
@@ -258,6 +258,9 @@
     <Compile Include="UserControls\MultitaskingControl\VerticalTabViewControl.xaml.cs">
       <DependentUpon>VerticalTabViewControl.xaml</DependentUpon>
     </Compile>
+    <Compile Include="UserControls\Widgets\DrivesWidget.xaml.cs">
+      <DependentUpon>DrivesWidget.xaml</DependentUpon>
+    </Compile>
     <Compile Include="UserControls\Widgets\LibraryCards.xaml.cs">
       <DependentUpon>LibraryCards.xaml</DependentUpon>
     </Compile>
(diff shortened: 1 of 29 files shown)


gave92:
Looking good @yair100 :)

image

Any reason for defining a new class for `DrivesLocationItem` instead of just using the existing `DriveItem`?

yaichenbaum:
@gave92 I started with a copy of the library cards to have something to go off of so that's probably why. Did you get the progress bar working?

gave92:
@yair100 Yeah, I just tried using the existing `DriveItem` class which has `MaxSpace` and `SpaceUsed` properties, like this:
```cs
<muxc:ProgressBar
        MinWidth="160"
        Maximum="{x:Bind MaxSpace.GigaBytes, Mode=OneWay}"
        Value="{x:Bind SpaceUsed.GigaBytes, Mode=OneWay}" />
```
I also made the `itemsAdded` list an ObservableCollection (so drives should disappear when disconnected)

yaichenbaum:
@gave92 Great work! I am going to work on a settings page to turn on and off widgets, so hold off on approving the changes.

gave92:
Just a small UI tweak, I noticed the progressbar got cut if the app's window was small

gave92:
Couple of notes:
1) Settings are not applied immediately, you have to navigate away from homepage and back again
2) Drive with long names will not be displayed correctly (text is cut)
image
3) As an alternative to settings we could show each widget under an expander (and save the expanded state). Just an idea, not sure if it's better
image
4) Don't think it's related to this but.. this started happening, is it just me? Clicked sidebar items now look like they are permenently selected
image

yaichenbaum:
> Couple of notes:
> 
> 1. Settings are not applied immediately, you have to navigate away from homepage and back again

I cannot reproduce this, are you testing with the last commit?

> 2. Drives with long name will not be displayed correctly (text is cut)
>    image

I will look at making this more responsive.

> 3. As an alternative to settings we could show each widget under an expander (and save the expanded state). Just an idea, not sure if it's better
>    image

The reason I added a settings page is because there will be more widgets in the future and users may not want to display all of them. It doesn't need to be a settings page but let's stick with one for now.

> 4. Don't think it's related to this but.. this started happening, is it just me? Clicked sidebar items now look like they are permenently selected
>    image

Do you have the experimental setting to reorder pinned items turned on? If so, that would be why it's happening, there is a bug with the navigation menu, and we are waiting for it to be fixed in the WinUI library.

gave92 (approved):

seohyun1108:
Very good!

Issue #2190, opened 2020-10-31
Hidden Folder Bug


RomRodry:
Hello, before I describe the bug I encountered I would like to say that I really admire your work and look forward to the release of the final version of File. 

**Describe the bug**
Whenever I hide a folder using the General Tab in the Properties of that same folder and then hide it using the Hidden items option in the View tab of the default File Explorer, File automatically closes.

**To Reproduce**
(some of this steps were taken from https://www.windowscentral.com/how-hide-files-and-folders-windows-10) 
1. Open File Explorer.
2. Navigate to the file or folder you want to hide.
3. Right-click the item and click on Properties.
4. On the General tab, under Attributes, check the Hidden option.
5. Click Apply.
6. Select the Apply changes to this folder, subfolders, and files option.
7. Click OK.
8. Click OK again to complete the task.
9. Open File Explorer
10. Click on the View tab 
11. Check the Hidden items option

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS Version: Windows 10.0.19041 
 - App version: 0.18.0.0

**Additional context**

**Log file**
[debug.txt](https://github.com/files-community/Files/files/5469210/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5469211/debug_fulltrust.txt)

igoragm:
I can confirm this. Steps that I took to reproduce it:
1) Opened a folder
2) In default File Explorer set the folder to hidden 
3) Got back to Files, right click withing the folder and selected Refresh.

Does Files even support hiding items and displaying those that are already hidden? Under Properties both Read-only and Hidden attributes are disabled for me (screenshot attached).
Second question related to the first one, what would be the desired behavior in this particular scenario?

Thomas1664:
> Does Files even support hiding items and displaying those that are already hidden?

No, but maybe in the future.

> Second question, what would be the desired behavior in this particular scenario?

Don't show the hidden folder.

igoragm:
> 
> 
> > Does Files even support hiding items and displaying those that are already hidden?
> 
> No, but maybe in the future.
> 
> > Second question, what would be the desired behavior in this particular scenario?
> 
> Don't show the hidden folder.

Thanks for the answer.
I'd like to contribute to this. Could you please assign me to the issue?

Thomas1664:
I can't. But @yaichenbaum can.

igoragm:
[...]

PR #2197, opened 2020-11-01
[#2190] - Checked for UnauthorizedAccessException on a selected directory

igoragm:
[#2190] - Checked for UnauthorizedAccessException on a selected directory. When the exception occurs, GetFolderWithPathFromPathAsync will be called recursively each time with a parent path of the folder for which we don't have necessary access rights.


tsvietOK (approved):

yaichenbaum:
@igoragm can you take a look at #7625?

PR #2210, opened 2020-11-02
Fix for Files crashing when attempting to access unmounted ISO

igoragm:
Expanded the catch clause in StorageFileExtension to include a wider range of exceptions & logged out the error message.

### DIFF
--- a/Files/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
+++ b/Files/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
@@ -128,8 +128,9 @@ namespace Files.Filesystem
                 {
                     return new StorageFolderWithPath(await StorageFolder.GetFolderFromPathAsync(value));
                 }
-                catch (UnauthorizedAccessException)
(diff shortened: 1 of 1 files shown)


gave92:
@igoragm thanks for your contributions!
Please note that the proposed solution (as it is) won't solve the linked issue:
`Directory.GetParent(value).FullName` will throw NullReferenceException in that case because you are already at the root of the drive (`Directory.GetParent` will return null)

igoragm:
@gave92 thanks for your reply. 
You are right, what I've handled was accessing folders within an unmounted ISO but I haven't handled accessing files which does result in an exception of its own. In fact I'm not even sure this component handles opening the files at all, I will have to check and make the appropriate changes.

yaichenbaum:
@igoragm Thank you for your help on this.

gave92:
@igoragm thanks for the help! The original issue has been fixed in #2269 as part of a larger rework of exception handling.

igoragm:
Great! Thanks for the info.

PR #2213, opened 2020-11-03
Added animations to the widgets

yaichenbaum:


gave92:
Nice! A couple of (very small) visual issues:
- In some conditions the zoomed element appears slightly cut
image
- Depending on the window size the elements zoom appears "off-center" (margin larger on one side vs the other
image

gave92 (approved):

PR #2215, opened 2020-11-03
Overhaul Codebase to Avoid Static App.CurrentInstance

### LINKED ISSUE #16: Add support for multiple Onedrive instances
Hey. I have both my work Onedrive and personal Onedrive on my computer. Could I have separate instances for it?

Also, SharePoint support would be great.

Great app!

duke7553:
This PR is the culmination of lots of work I put into improving the app's codebase by removing/replacing our use of static App.CurrentInstance. This makes certain controls inside the project more modular as they no longer have inside knowledge of the current selected app instance. While this is a large step forward, this PR currently exists for review purposes and has no guarantee of being merged. Please test as much of the app's functionality as possible to find edge cases, because things may not be what they seem with these changes alone.

What I'm aware of:
- FIXED: ~~**Broken workaround for ModernNavigationToolbar and StatusBarControl**: Functionality on these controls is broken until we can implement the Commands properly.~~
- FIXED: ~~**Displayed path briefly flickers an incorrect value**: This only seems to happen when navigating back/forward~~
- OK: ~~**Opening or closing many tabs at once will lead to an unhandled exception**: We need to figure our a way to throttle tab creation/removal to be one at a time.~~
- FIXED: ~~**Opening the properties dialog will crash the app**~~

Fixes #16.

I will also look forward to hearing your feedback on the more technical aspects of this codebase change. If something seems out of place, we'd ideally want to document/correct it before this PR is merged.


Files/App.xaml.cs
```diff
@@ -38,26 +38,8 @@ namespace Files
 {
     sealed partial class App : Application
     {
-        public static IMultitaskingControl MultitaskingControl = null;
-
-        private static IShellPage currentInstance;
         private static bool ShowErrorNotification = false;
 
-        public static IShellPage CurrentInstance
```
duke7553, line 46: Removing this once and for all felt sooooo good.

Files/Views/ModernShellPage.xaml.cs
```diff
@@ -103,21 +555,21 @@ protected override void OnNavigatedTo(NavigationEventArgs eventArgs)
...
+        AppServiceConnection Connection = null;
         private void Page_Loaded(object sender, RoutedEventArgs e)
         {
-            viewModel = new ItemViewModel();
-            interactionOperation = new Interaction();
-
+            Connection = new AppServiceConnection();
+            FilesystemViewModel = new ItemViewModel(this, ref Connection);
+            FilesystemViewModel.InitializeAppServiceConnection();
+            InteractionOperations = new Interaction(this, ref Connection);
```
duke7553, line 564: Should the AppInstance be passed by reference here also?

gave92:
Noticed one issue:
- if you minimize and than restore the app anything that invokes the fulltrust process will crash the app with a `System.ObjectDisposedException` on `AppServiceConnection.SendMessageAsync`
Before the AppServiceConnection was recreated in the `OnLeavingBackground` when restoring the app

duke7553:
@gave92 Thanks for pointing this out. I noticed this also, and assumed that I fixed it by registering the OnLeavingBackground to an event handler inside ItemViewModel.cs. Maybe we should check if the AppService is available a better way?

gave92:
@duke7553 I think the only issue is that you added this check `if (this.Connection == null)` which is only true the first time.
I think somewhere we forgot to put `Connection = null` when the connection is being closed

duke7553:
@gave92 I added that check because the AppServiceConnection would occasionally display an exception that the Connection.OpenAsync() was executed at the wrong time. I assume that was because we tried to initialize the connection when one already exists.

Files/Views/ModernShellPage.xaml.cs
Jaiganeshkumaran: Pass by reference if you want the reference change to apply in other places. Technically already everything is a reference but changing it will create a new reference without the ref keyword.

duke7553: Gave92's recent commits make this question obsolete.

yaichenbaum (approved):

d2dyno1:
:eyes:

d2dyno1:
Wooooo!!!! :confetti_ball: :tada:

PR #2267, opened 2020-11-08
Fixed the right click context menu

yaichenbaum:


duke7553 (approved):
We should be on the lookout for other broken libraries we have dependencies on in case they aren't compatible with the new optimizations applied.

PR #2289, opened 2020-11-09
Refactored properties: details page

winston-de:
These changes were requested in my original PR, #1962.
 
### Changes made:
- Moved properties to a list view, bound to a grouped data set
- Removed dialog header
- Removed overview section, as it is unnecessary with a "view all" feature
- Support for other languages with property names
- Fix issue with text box size
- Dialog for failed save

### Screenshots:
<image src="https://user-images.githubusercontent.com/59544401/100019882-a6fd6980-2d93-11eb-9e77-3e7f917d0f46.png" width="400"/>
<image src="https://user-images.githubusercontent.com/59544401/100018077-bcbd5f80-2d90-11eb-8b37-eb6cef8cd35a.png" width="400"/>


yaichenbaum:
> Push notification for failed save

@winston-de Instead of a push notification, it's probably better to display a dialog.

yaichenbaum:
@winston-de Is this ready for review yet?

winston-de:
@yair100 Yes

yaichenbaum:
@winston-de It looks like there are several merge conflicts, can you solve those?

winston-de:
@yair100 Sure.
I do that by merging with upstream/master in VS, right?

yair100:
> @yair100 Sure.
> I do that by merging with upstream/master in VS, right?

Yeah

yair100 (changes requested):

Files/View Models/Properties/FileProperty.cs
```diff
@@ -0,0 +1,269 @@
...
+        /// </summary>
+        public void InitializeNames()
+        {
+            Name = ResourceController.GetTranslation(NameResource);
+            Section = ResourceController.GetTranslation(SectionResource);
+        }
+
+        public static readonly List<FileProperty> PropertyListItemsBase = new List<FileProperty>()
+        {
+            new FileProperty() { NameResource = "PropertyAddress", SectionResource = "PropertySectionGPS", ID = "address" },
```
d2dyno1, line 184: Assign the properties in a constructor?

d2dyno1: Actually instead of hardcoding it like this, I'd create a factory class that'd instantiate FileProperty from databse

Files/Strings/en-US/Resources.resw
```diff
@@ -1122,16 +1122,22 @@
...
-  <data name="PropertySaveErrorDialogCloseAnyway" xml:space="preserve">
+  <data name="PropertySaveErrorDialog.SecondaryButtonText" xml:space="preserve">
     <value>Close anyway</value>
   </data>
-  <data name="PropertySaveErrorDialogCancel" xml:space="preserve">
+  <data name="PropertySaveErrorDialog.CloseButtonText" xml:space="preserve">
+    <value>Cancel</value>
+  </data>
+  <data name="PropertySaveErrorMessage" xml:space="preserve">
     <value>Cancel</value>
```
yaichenbaum, line 1141: Is the message meant to be "cancel"?

winston-de: @yair100 Yeah, because it just cancels the attempt to save (and closes the dialog).

yaichenbaum: I thought the message was meant to say "There was an issue saving some properties..."

winston-de: @yair100 Wait no you're right, I see what you're talking about. Sorry about that.

yaichenbaum:
@winston-de It looks like the resource files just got messed up.

winston-de:
@yair100 Can you please tell me what you mean? I can't seem to figure out what the issue is.

Files/Strings/de-DE/Resources.resw
```diff
@@ -1002,94 +1002,4 @@
...
-  </data>
-  <data name="HorizontalMultitaskingControlDuplicateTab.Text" xml:space="preserve">
-    <value>Tab duplizieren</value>
-  </data>
-  <data name="HorizontalMultitaskingControlMoveTabToNewWindow.Text" xml:space="preserve">
-    <value>Tab in neuem Fenster öffnen</value>
-  </data>
-  <data name="HorizontalMultitaskingControlNewTab.Text" xml:space="preserve">
-    <value>Neuer Tab</value>
-  </data>
```
yaichenbaum, line 1094: @winston-de Here is an exmaple, it looks like some of the translated strings were deleted.

d2dyno1:
```c#
foreach (string line in database)
{
    string[] lineSplit = line.Split(separator);
    string NameResource = lineSplit[0];
    string SectionResource = lineSplit[1];
    string ID = lineSplit[2];
    props.Add(new FileProperty(NameResource, SectionResource, ID));
}
```

Files/Strings/de-DE/Resources.resw
winston-de: @yair100 Thanks. I think I messed up the translation resource files when merging.
Anyway, it should be fixed now.

Files/Views/Pages/PropertiesDetails.xaml
```diff
@@ -2,501 +2,97 @@
...
+                x:Name="ClearPropertiesButton"
+                Margin="0,10,0,0"
+                HorizontalAlignment="Stretch"
+                x:Uid="ClearPropertiesButton">
                     <Button.Flyout>
-                        <Flyout x:Name="ClearPersonalInformationFlyout">
+                        <Flyout x:Name="ClearPropertiesFlyout">
                             <StackPanel Orientation="Vertical" Spacing="5">
-                                <TextBlock Text="Confirm?" />
+                                <TextBlock x:Uid="ClearPropertiesFlyoutText"/>
```
yaichenbaum, line 87:
```suggestion
                                <TextBlock x:Uid="ClearPropertiesFlyoutText"  Text="Confirm?"/>
```

Files/Views/Pages/PropertiesDetails.xaml
```diff
@@ -2,501 +2,97 @@
...
                                 <StackPanel Orientation="Horizontal" Spacing="10">
-                                    <Button Click="ClearPersonalInformation_Click" Content="Clear" />
+                                    <Button Click="ClearPropertiesConfirmation_Click" x:Uid="ClearPropertiesFlyoutConfirmation"/>
```
yaichenbaum, line 89:
```suggestion
                                    <Button Click="ClearPropertiesConfirmation_Click" x:Uid="ClearPropertiesFlyoutConfirmation" Content="Clear"/>
```

Files/Dialogs/PropertySaveError.xaml
```diff
@@ -0,0 +1,20 @@
...
+    x:Uid="PropertySaveErrorDialog"
+    PrimaryButtonClick="ContentDialog_PrimaryButtonClick"
+    CornerRadius="3"
+    SecondaryButtonClick="ContentDialog_SecondaryButtonClick"
+    PrimaryButtonStyle="{StaticResource ButtonRevealStyle}"
+    SecondaryButtonStyle="{StaticResource ButtonRevealStyle}"
+    CloseButtonStyle="{StaticResource ButtonRevealStyle}">
+
+    <StackPanel Spacing="5">
+        <TextBlock Grid.Row="1" x:Uid="PropertySaveErrorMessage"/>
```
yaichenbaum, line 18: Can you add the text in the xaml as well?

yair100 (dismissed):

winston-de:
> ```cs
> foreach (string line in database)
> {
>     string[] lineSplit = line.Split(separator);
>     string NameResource = lineSplit[0];
>     string SectionResource = lineSplit[1];
>     string ID = lineSplit[2];
>     props.Add(new FileProperty(NameResource, SectionResource, ID));
> }
> ```

@d2dyno1 Do you want me to move the base list of properties to a separate file?

d2dyno1 (commented):
> @d2dyno1 Do you want me to move the base list of properties to a separate file?

Yea, a database file would remove hardcoding. It has also the benefit of - if the constructor was changed, then you'd only need to update it in one place

winston-de:
@d2dyno1 I think this could be accomplished pretty easily with a [Json deserializer](https://www.newtonsoft.com/json/help/html/DeserializeCollection.htm) and a separate Json file.

yair100 (dismissed):

d2dyno1 (dismissed):

Files/View Models/SelectedItemsPropertiesViewModel.cs
```diff
@@ -639,132 +646,11 @@ public Uri FolderIconSource
...
-        public Visibility DetailsSectionVisibility_Music
-        {
-            get => detailsSectionVisibility_Music;
-            set => SetProperty(ref detailsSectionVisibility_Music, value);
-        }
-
-        private Visibility detailsSectionVisibility_Media;
-
-        public Visibility DetailsSectionVisibility_Media
+        private ObservableCollection<FilePropertySection> _PropertySections = new ObservableCollection<FilePropertySection>();
```
tsvietOK, line 649: Make sure you are following our code style, you can find it on our wiki page.

tsvietOK (changes requested):

Files/View Models/SelectedItemsPropertiesViewModel.cs
yaichenbaum, line 649:
```suggestion
        private ObservableCollection<FilePropertySection> propertySections = new ObservableCollection<FilePropertySection>();
```

Files/View Models/SelectedItemsPropertiesViewModel.cs
```diff
@@ -639,132 +646,11 @@ public Uri FolderIconSource
...
+        public ObservableCollection<FilePropertySection> PropertySections
         {
-            get => detailsSectionVisibility_Media;
-            set => SetProperty(ref detailsSectionVisibility_Media, value);
+            get => _PropertySections;
+            set => SetProperty(ref _PropertySections, value);
```
yaichenbaum, line 653:
```suggestion
            get => propertySections;
            set => SetProperty(ref propertySections, value);
```

yair100 (dismissed):

Thomas1664:
I don't think we need an extra dialog page. The dialog can be created in codebehind.

Files/Views/Pages/PropertiesDetails.xaml.cs
```diff
@@ -31,71 +31,41 @@ protected override void Properties_Loaded(object sender, RoutedEventArgs e)
...
-        }
-
-        private void SetStringArray(string val, string key)
-        {
-            ViewModel.SystemFileProperties_RW[key] = val.Split("; ");
-        }
+                switch (await dialog.ShowAsync())
+                {
+                    case ContentDialogResult.Primary:
+                        SaveChangesAsync();
```
tsvietOK, line 52: Recursive call?

tsvietOK:
@winston-de Can you check, is Details tab working for you while file hash calculation is not complete?

Files/Views/Pages/PropertiesDetails.xaml.cs
winston-de: Yeah. That's if the user presses the retry button.

tsvietOK: Should we await this call?

winston-de:
> Can you check, is Details tab working for you while file hash calculation is not complete?
@tsvietOK It works on my machine, although it does seem to cause a slight delay.
I'll have access to another device in a little bit, I'll test it on that, too.

tsvietOK:
@winston-de I testing with 14Gb file, it takes 1 minute to get details properties.

Files/Views/Pages/PropertiesDetails.xaml.cs
winston-de: @tsvietOK Is there a better way to do this?

tsvietOK: What about Infinite loop?

tsvietOK: Kind of
```
public async Task<bool> TrySaveChangesAsync()
{
    while (true)
    {
        try
        {
            await (BaseProperties as FileProperties).SyncPropertyChangesAsync();
        }
        catch
        {
            var dialog = new PropertySaveError();

            switch (await dialog.ShowAsync())
            {
                case ContentDialogResult.Primary:
                    break;
                case ContentDialogResult.Secondary:
                case ContentDialogResult.None:
                    return true;
                default:
                    break;
            }
        }
        break;
    }

    return true;
}

```

winston-de:
@tsvietOK Oh ok. I tested it with larger files, and I have the same issue. I'll add some code to cancel the task when the user switches tabs.

Files/Views/Pages/PropertiesDetails.xaml.cs
```diff
@@ -31,71 +31,41 @@ protected override void Properties_Loaded(object sender, RoutedEventArgs e)
...
-            {
-                OverviewImage.Visibility = Visibility.Visible;
-            }
-        }
-
-        private string GetStringArray(object array)
+        /// <summary>
+        /// Returns false if the operation was cancelled
+        /// </summary>
+        /// <param name="item"></param>
```
tsvietOK, line 37: There is no such parameter in this method.

Files/Views/Pages/PropertiesDetails.xaml.cs
```diff
@@ -31,71 +31,44 @@ protected override void Properties_Loaded(object sender, RoutedEventArgs e)
...
-            {
-                OverviewImage.Visibility = Visibility.Visible;
-            }
-        }
-
-        private string GetStringArray(object array)
+        /// <summary>
+        /// Tries to save changed properties to file.
+        /// </summary>
+        /// <returns>Returns true if properties has been saved successfully.</returns>
```
yaichenbaum, line 37:
```suggestion
        /// <returns>Returns true if properties have been saved successfully.</returns>
```

tsvietOK:
@winston-de When opening Details tab and pressing Ok button, it shows error dialog, when trying to save changes without an actual chnages.

Files/View Models/Properties/FileProperty.cs
```diff
@@ -0,0 +1,271 @@
...
+
+            return list;
+        }
+
+        /// <summary>
+        /// Since you can't serialize lambdas from a json file, define them here
+        /// </summary>
+        private static readonly Dictionary<string, Func<object, string>> DisplayFuncs = new Dictionary<string, Func<object, string>>()
+        {
+            { "DivideBy100", input => (((uint) input)/1000).ToString() },
```
tsvietOK, line 267: DivideBy100 or 1000?

Files/View Models/Properties/FileProperty.cs
```diff
@@ -0,0 +1,271 @@
...
+            {
+                DisplayFunction = displayFunction;
+            }
+        }
+
+        /// <summary>
+        /// Sets the Value property
+        /// </summary>
+        /// <param name="file">The file to get the property of</param>
+        public async Task SetValueFromFile(StorageFile file)
```
tsvietOK, line 139: No more used.

Files/View Models/Properties/FileProperty.cs
```diff
@@ -0,0 +1,271 @@
...
+        {
+            IsReadOnly = isReadOnly;
+        }
+
+        /// <summary>
+        /// Use this function when taking a file property from the base list and adding it to the view model list.
+        /// </summary>
+        /// <param name="fileProperty"></param>
+        /// <returns></returns>
+        public async Task InitializeProperty(StorageFile file)
```
tsvietOK, line 126: No more async.

tsvietOK:
@winston-de When changing Title field, this change applies to Subject too.

tsvietOK:
@winston-de What about to remove read only fields if they are empty?

tsvietOK:
@winston-de For some reason some of the properties are not retrieved.
image

winston-de:
@tsvietOK 

> When changing Title field, this change applies to Subject too.

What type of file are you testing on? I am unable to reproduce this issue. 

> What about to remove read only fields if they are empty?

Good idea.

> For some reason some of the properties are not retrieved.

Does this happen with all image files?

tsvietOK:
> 
> What type of file are you testing on? I am unable to reproduce this issue.

Photo(,jpg).

> Does this happen with all image files?

Hmmm, now it shows horizontal and vertical resolutions, but without Color space. And yes, for all image files.

winston-de:
@tsvietOK 
> Photo(,jpg).

I tried it on a jpg file, and I had the issue once, then it stopped happening. I think it might be a bug with the storage API.

> Hmmm, now it shows horizontal and vertical resolutions, but without Color space. And yes, for all image files.

I don't see color space either, however I don't think it's an issue with Files. I tried using an online metadata viewer and it didn't show anything.

Files/View Models/Properties/FileProperty.cs
```diff
@@ -0,0 +1,259 @@
...
+
+            return list;
+        }
+
+        /// <summary>
+        /// Since you can't serialize lambdas from a json file, define them here
+        /// </summary>
+        private static readonly Dictionary<string, Func<object, string>> DisplayFuncs = new Dictionary<string, Func<object, string>>()
+        {
+            { "DivideBy1000", input => (((uint) input)/1000).ToString() },
```
tsvietOK, line 255: I guess here should be added null check.

winston-de: @tsvietOK Good point, though I think it might be simpler to add a null check in the ConvertToString function, so that no null values get passed to any converter.

tsvietOK:
@winston-de Looks like Color space property is not empty.
image
Value 1 means it is sRGB: https://www.awaresystems.be/imaging/tiff/tifftags/privateifd/exif/colorspace.html

winston-de:
@tsvietOK Thanks. I think the reason it was showing as null is because it's a Uint16, which doesn't work with the current system. I'll fix that, and also add support for enumerated lists so that it displays the text instead of just the number value.

tsvietOK:
@winston-de Do you plan to add more supported properties and when? (https://docs.microsoft.com/ru-ru/windows/win32/properties/props)

winston-de:
@tsvietOK I do intend to add some more later this month, however there are a lot of properties and I'm a bit busy. It's also very easy to expand the json file, so if anyone has some they want to add they can.

tsvietOK:
> @tsvietOK I do intend to add some more later this month, however there are a lot of properties and I'm a bit busy. It's also very easy to expand the json file, so if anyone has some they want to add they can.

It is not a problem, just asking.

tsvietOK (approved):

yair100 (approved):

d2dyno1 (approved):

yaichenbaum:
@winston-de Great work!

winston-de:
@yair100 Thanks :)

PR #2323, opened 2020-11-12
Include language resources in the app package

yaichenbaum:
This change will allow users to change the language even if the user did not install the region from the Windows settings app.


duke7553 (commented):
@yaichenbaum What benefits would this change deliver?

yaichenbaum:
@duke7553 I updated the description.

duke7553 (approved):

PR #2327, opened 2020-11-13
CloudProvider: Code refactor

tsvietOK:


yaichenbaum (dismissed):

Files/Filesystem/Cloud/Providers/GoogleDriveCloudProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+                string appDataPath = UserDataPaths.GetDefault().LocalAppData;
+                string dbPath = @"Google\Drive\user_default\sync_config.db";
+                var configFile = await StorageFile.GetFileFromPathAsync(Path.Combine(appDataPath, dbPath));
+                await configFile.CopyAsync(ApplicationData.Current.TemporaryFolder, "google_drive.db", NameCollisionOption.ReplaceExisting);
+                var syncDbPath = Path.Combine(ApplicationData.Current.TemporaryFolder.Path, "google_drive.db");
+
+                // Build the connection and sql command
+                SQLitePCL.Batteries_V2.Init();
+                using (var con = new SqliteConnection($"Data Source='{syncDbPath}'"))
+                using (var cmd = new SqliteCommand("select * from data where entry_key='root_config__0'", con)) //local_sync_root_path
```
tsvietOK, line 26: I changed "local_sync_root_path" to "root_config__0", because it didn't find Google Drive sync folder, now it is working for me. @gave92 If you use Google Drive sync, can you check does any of this keys exists in your sync_config.db? Another problem here, is when in google drive sync you select two different folders, Files shows only second folder.
image

gave92: Hi @tsvietOK yeah I tested with google drive and or me it works using "local_sync_root_path"
I'll check with "root_config__0" as well
This is what I'm using:
image
I don't this I even have the option to synchronize more than one folder

tsvietOK: I use the same version. You can open Drive sync settings and select other folders to synchronize them.

gave92: This is my db, I can confirm that "root_config__0" works as well
image

gave92: I just realized you are syncing folders from your pc to google drive. I was syncing stuff I had on my drive to my pc instead.
I think the first uses "root_config__0" the second "local_sync_root_path"

tsvietOK: About Mega, where did you get an algorithm of sync folder decryption?

gave92: Did it myself, MEGA is [open source](https://github.com/meganz/MEGAsync) which helped :)

Files/Filesystem/Drives.cs
```diff
@@ -42,6 +40,8 @@ public bool ShowUserConsentOnInit
         public DrivesManager()
         {
             EnumerateDrives();
+
+            cloudProviderController = new CloudProviderController();
```
gave92, line 44: This should be called before EnumerateDrives right? Since you call GetVirtualDrivesListAsync which uses it inside EnumerateDrives

gave92 (changes requested):

Files/Filesystem/Drives.cs
tsvietOK: Yeah, your right. But it is still working...

gave92: Magic of `async void` methods

gave92 (dismissed):

gave92 (approved):

PR #2337, opened 2020-11-14
[Hidden files] - Enable setting file/folder as hidden in Properties page

gave92:
This PR adds support for setting a file/folder as hidden and read-only from properties page (it enables the checkboxes)


duke7553 (approved):
LGTM. Great use of the *FromApp method

yaichenbaum (approved):

PR #2343, opened 2020-11-15
Added a rightclick context menu to the drives widget

yair100:
- [x] Improve the code quality

### DIFF
--- a/Files/UserControls/Widgets/DrivesWidget.xaml
+++ b/Files/UserControls/Widgets/DrivesWidget.xaml
@@ -75,6 +75,23 @@
                             Style="{StaticResource ButtonRevealStyle}"
                             Tag="{x:Bind Path}"
                             ToolTipService.ToolTip="{x:Bind Text, Mode=OneWay}">
+                            <Button.ContextFlyout>
+                                <MenuFlyout>
+                                    <MenuFlyout.Items>
+                                        <MenuFlyoutItem
+                                            x:Name="EjectDevice"
+                                            Tag="{x:Bind Path}"
+                                            x:Uid="SideBarEjectDevice"
+                                            x:Load="{x:Bind IsRemovable, Mode=OneWay}"
+                                            Click="EjectDevice_Click"
+                                            Text="Eject">
+                                            <MenuFlyoutItem.Icon>
+                                                <FontIcon FontFamily="{StaticResource CustomGlyph}" Glyph="&#xF10B;" />
+                                            </MenuFlyoutItem.Icon>
+                                        </MenuFlyoutItem>
+                                    </MenuFlyout.Items>
+                                </MenuFlyout>
+                            </Button.ContextFlyout>
                             <Grid
                                 Margin="12"
                                 HorizontalAlignment="Stretch"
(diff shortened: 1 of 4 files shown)


Files/UserControls/Widgets/DrivesWidget.xaml
```diff
@@ -75,6 +75,23 @@
                             Style="{StaticResource ButtonRevealStyle}"
                             Tag="{x:Bind Path}"
                             ToolTipService.ToolTip="{x:Bind Text, Mode=OneWay}">
+                            <Button.ContextFlyout>
+                                <MenuFlyout>
+                                    <MenuFlyout.Items>
+                                        <MenuFlyoutItem
+                                            x:Name="EjectDevice"
+                                            Tag="{x:Bind Path}"
```
yaichenbaum, line 83: We should find a better way to do this so that we don't need to pass in the path.

Files/UserControls/Widgets/DrivesWidget.xaml.cs
```diff
@@ -24,6 +26,11 @@ public DrivesWidget()
             InitializeComponent();
         }
 
+        private async void EjectDevice_Click(object sender, RoutedEventArgs e)
+        {
+            await Interaction.EjectDeviceAsync(((MenuFlyoutItem)sender).Tag.ToString());
```
yaichenbaum, line 31: Ideally we should derive from DriveItem here.

gave92: Hi @yair100 I don't find anything particularly wrong with the current approach.
Anyway an alternative method is in #2347, if you like it more

Files/UserControls/Widgets/DrivesWidget.xaml
gave92: Hi @yair100 I don't find anything particularly wrong with the current approach.
Anyway an alternative method is in #2347, if you like it more

Files/UserControls/Widgets/DrivesWidget.xaml.cs
```diff
@@ -24,6 +26,24 @@ public DrivesWidget()
             InitializeComponent();
         }
 
+        private async void EjectDevice_Click(object sender, RoutedEventArgs e)
+        {
+            var item = ((MenuFlyoutItem)sender).DataContext as DriveItem;
```
tsvietOK, line 31: What about to do the same in SidebarControl.xaml.cs?

yaichenbaum: @tsvietOK We already do.

yaichenbaum: It's a little out of scope for this pull request but we can.

gave92:
@yair100 are you still planning to add a Properties context menu item for drives?

yair100:
> @yair100 are you still planning to add a Properties context menu item for drives?

I ran into an issue where I couldn't just call ShowProperties since it needed an object passed in. I am going to keep playing around but I decided to leave that out for now.

gave92:
I'm seeing that for drives we simply pass a DriveItem so you should be able to do that if you want
``cs
await OpenPropertiesWindowAsync(App.AppSettings.DrivesManager.Drives
                        .Single(x => x.Path.Equals(AssociatedInstance.FilesystemViewModel.CurrentFolder.ItemPath)));`
``

gave92 (approved):

Issue #2008, opened 2020-09-30
Multiselect


yanjingtui:
**Is your feature request related to a problem? Please describe.**
Please add multi select features for touch screen and mouse

**Describe the solution you'd like**
Long press a file, hold your finger and use another finger to select another file. Just like this at 1:34 https://youtu.be/I5EA8il5ojQ
For mouse users, just select file by clicking and dragging. 

**Describe alternatives you've considered**
Just add a select and modify mode

**Additional context**
![image](https://user-images.githubusercontent.com/18437653/94722053-08d2c180-0357-11eb-9cdf-37fc8902a35a.png)

gave92:
An alternative way for touch would be a long press that triggers a "multiple selection mode" where checkboxes appear next to items
![image](https://user-images.githubusercontent.com/9673091/99309862-66b15f00-285a-11eb-8629-ba2859d2d2c2.png)

yaichenbaum:
This feature was recently implemented in `main`.

PR #2356, opened 2020-11-16
Do not trigger selection rectangle with touch

### LINKED ISSUE #2349: Don't show selection rectangle on touch input
**Describe the bug**
the selection rectangle shows up while using the touchscreen to scroll in the items view

**To Reproduce**
1. Open a folder with many items or decrease the window size until the scrollbar appears
2. Use the touch screen to scroll up and down

**Expected behavior**
Don't show selection rectangle while using the touchscreen in scrollable folders

 - OS Version: Windows 10 20H2 19042.630
 - App version: v0.20.0.0

gave92:
Fixes #2349 

This PR disables drag selection with touch input to not interfere with list scrolling

### DIFF
--- a/Files/UserControls/Selection/RectangleSelection_DataGrid.cs
+++ b/Files/UserControls/Selection/RectangleSelection_DataGrid.cs
@@ -140,9 +140,9 @@ namespace Files.UserControls.Selection
             originDragPoint = new Point(e.GetCurrentPoint(uiElement).Position.X, e.GetCurrentPoint(uiElement).Position.Y); // Initial drag point relative to the topleft corner
             var verticalOffset = (scrollBar?.Value ?? 0) - 38; // Header height
             originDragPoint.Y += verticalOffset; // Initial drag point relative to the top of the list (considering scrolled offset)
-            if (!e.GetCurrentPoint(uiElement).Properties.IsLeftButtonPressed)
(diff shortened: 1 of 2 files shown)


yaichenbaum:
@gave92 Is it posible to keep the selection rectangle on touch devices but only when the user holds down for a few seconds?

gave92:
Perhaps, but in general it may be a bit difficult to trigger the selection rectangle on touch as you have to click outside listed items. I was so thinking that for now it's better to disable it.
(I don't own a touchscreen device though, I'm not able to test this)

yaichenbaum:
@gave92 In that case it might be better to disable it for now, but eventually we will want to implement long touch + drag since touch users will want a way to select multiple items.

yaichenbaum (approved):

gave92:
"eventually we will want to implement long touch + drag since touch users will want a way to select multiple items"
An alternative way for touch would be a long press that triggers a "multiple selection mode" where checkboxes appear next to items
image

yaichenbaum:
@gave92 That's a nice idea, can you add it to #2008?

yair100:
> Perhaps, but in general it may be a bit difficult to trigger the selection rectangle on touch as you have to click outside listed items. I was so thinking that for now it's better to disable it. (I don't own a touchscreen device though, I'm not able to test this) What do you think?

@gave92 I don't recall the exact setup we had when this was last discussed, but in the latest version of Files, there are scenarios where there is enough room outside the file area. Do you have any pointers to be able to implement #17961.

Issue #260, opened 2019-10-31
Context menu for tabs


szaimen:
**Is your feature request related to a problem? Please describe.**
I would like to change the sort order of tabs and open already open tabs in new windows

**Describe the solution you'd like**
I would like to pull one tab out of the tab row to let it open in a new window. (you can do this in edge) and change the sort order of tabs through pulling

**Describe alternatives you've considered**
Maybe you could also enable us to right-click on tabs to get e.g. the following options (most of the are also seen e.g. in Edge):
- [x] duplicate tab
- [x] **move tab to new window** (most important in my opinion)
- [x] close tabs to the right
- [ ] pin tab
- [x] new tab

**Additional context**
![image](https://user-images.githubusercontent.com/42591237/67966370-155d1000-fc04-11e9-9062-46b734c1b77b.png)

yaichenbaum:
You can already change the order of tabs through pulling if you build the application of the master branch.
The other stuff we plan on doing, thanks fro creating this issue to keep track of it.

yaichenbaum:
I updated the original pitch with some more information, I also realized that you might have meant pulling tabs for touch which is not implemented at the moment. I had originally tested with a mouse. I think this pitch now includes what we need so I will go ahead and approve this.

szaimen:
Maybe "move tab to new window" could get updated to "move tab to new window or to an already open other one"? 
I can do this if you agree.

yaichenbaum:
I don't want to add too many options to the menu but if enough people would find it helpful then we can explore adding a second option "Move to an open window" and that would have a sub menu with the other open windows.

yaichenbaum:
Improvements to touch and pulling tabs should come in a future update of WinUI.

Jaiganeshkumaran:
This is no longer relevant since the latest version has removed TabView.

szaimen:
> This is no longer relevant since the latest version has removed TabView.

Why, do you have a screenshot of the new tab layout?

Jaiganeshkumaran:
Here's the new layout. Tabs are no longer stacked on the tab like a browser. Instead, there's now a button for switching between and adding new tabs.
![திரைப்பிடிப்பு (7)](https://user-images.githubusercontent.com/45227503/87798950-3cc73c00-c86a-11ea-9281-34825c99a6c5.png)

szaimen:
Thanks for the screenshot!

generalguy41:
[...]

PR #2363, opened 2020-11-16
Added a right click context menu to the horizontal tab control

yair100:
image
Part of #260.

### DIFF
--- a/Files/MultilingualResources/Files.de-DE.xlf
+++ b/Files/MultilingualResources/Files.de-DE.xlf
@@ -1461,6 +1461,18 @@
           <source>Eject</source>
           <target state="new">Eject</target>
         </trans-unit>
+        <trans-unit id="HorizontalMultitaskingControlDuplicateTab.Text" translate="yes" xml:space="preserve">
+          <source>Duplicate tab</source>
+          <target state="new">Duplicate tab</target>
+        </trans-unit>
+        <trans-unit id="HorizontalMultitaskingControlMoveTabToNewWindow.Text" translate="yes" xml:space="preserve">
+          <source>Move tab to new window</source>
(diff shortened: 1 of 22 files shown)


d2dyno1:
Looks good 👍

gave92 (dismissed):
This adds a new tab at the end of the tab row right?
Would be nice if it added the new tab right next to the clicked tab

yaichenbaum:
> This adds a new tab at the end of the tab row right?
> Would be nice if it added the new tab right next to the clicked tab

That would be a nice enhancement, do you have any ideas how we can find out which tab was right clicked on?

gave92:
@yair100 The following should do the job
```cs
private async void NewTabMenuItem_Click(object sender, RoutedEventArgs e)
{
    var tabItem = ((FrameworkElement)sender).DataContext as TabItem;
    var index = MainPage.AppInstances.IndexOf(tabItem);
    await MainPage.AddNewTabByPathAsync(typeof(ModernShellPage), "NewTab".GetLocalized(), index + 1);
}
```
!EDIT!
This will easily trigger an (existing) bug that makes the app crash, to reproduce:
1) right click on the first tab and select "add new tab"
2) right click on the first tab and select "add new tab" (again)
3) navigate to a folder
4) crash with NullReferenceException in `SetSelectedTabInfoAsync`

gave92:
New options work fine but I can still repro the bug described above with the "new tab" menu
ezgif com-gif-maker

yaichenbaum:
> This will easily trigger an (existing) bug that makes the app crash, to reproduce:

@gave92 I removed the part that inserts the tab next to the one that was right clicked. This should solve the issue while we continue to look for a better solution.

gave92 (approved):

PR #2365, opened 2020-11-17
Added a tooltip with the path for pinned items on the sidebar

### LINKED ISSUE #1005: Sidebar tooltips
Adding an item to the sidebar merely adds its name (last element of the full path?).
However, apart from clicking it, its path is unknown (for us with short memories!). Add a tooltip to show full path.
Items with the identical name, but located on a different drive (e.g. NAS) can't be added at the moment. The Desktop explorer allows duplicate names. The tooltip helps identify them.

yaichenbaum:
Closes #1005

### DIFF
--- a/Files/UserControls/SidebarControl.xaml
+++ b/Files/UserControls/SidebarControl.xaml
@@ -55,7 +55,8 @@
                     Drop="NavigationViewLocationItem_Drop"
                     IsRightTapEnabled="True"
                     RightTapped="NavigationViewLocationItem_RightTapped"
-                    Tag="{x:Bind Path}">
+                    Tag="{x:Bind Path}"
+                    ToolTipService.ToolTip="{x:Bind Path}">
                     <muxc:NavigationViewItem.Icon>
                         <FontIcon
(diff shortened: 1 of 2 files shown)


Files/UserControls/SidebarControl.xaml
```diff
@@ -79,7 +80,7 @@
                     IsRightTapEnabled="True"
                     RightTapped="NavigationViewDriveItem_RightTapped"
                     Tag="{x:Bind Path}"
-                    ToolTipService.ToolTip="{x:Bind SpaceText, Mode=OneWay}"
+                    ToolTipService.ToolTip="{x:Bind Path, Mode=OneWay}"
```
lukeblevins, line 83: @yair100 Have we confirmed that users would prefer Path over Space Text here?

yaichenbaum: The feedback we got was that the path is more relevant than the space when a user is looking to navigate to a drive.

duke7553 (approved):

PR #2366, opened 2020-11-17
Increased the delay required to trigger double tap to rename

yaichenbaum:
This fixes an issue where double clicking an item would sometimes trigger the rename option.


gave92 (approved):

gave92:
With 1000ms the delay is a bit noticeable... 700ms? :)

gave92 (approved):

PR #2382, opened 2020-11-18
Fixed an issue where the eject drive option would sometimes show for pinned folders

### LINKED ISSUE #2373: Eject shown for pinned folders
**Describe the bug**
Sometimes eject option is shown for pinned folder

**To Reproduce**
1. Right-click on a secondary drive
2. Now right click on a pinned folder

If you right-click on the primary drive (Windows) you will not see the eject option for pinned folders next time

**Expected behavior**
Eject option should not be shown for pinned folders.

**Screenshots**
திரைப்பிடிப்பு (120)

**Desktop:**
 - OS: [e.g. Windows 10 Pro Version 20H2]
 - App version: [v0.20.1]

**Additional context**
Add any other context about the problem here. For instance, are you building from source or using a precompiled build/snapshot?
From the Store

**Log file**
Please post the log file here so that we can understand your problem better. You can access it from Settings->About->Open log location.
Nothing useful

yaichenbaum:
Fixes #2373


Files/UserControls/SidebarControl.xaml.cs
```diff
@@ -229,6 +229,7 @@ private void NavigationViewLocationItem_RightTapped(object sender, RightTappedRo
             // Additional check needed because ShowProperties is set to true if not recycle bin
             if (item.IsDefaultLocation)
             {
+                ShowEjectDevice = false;
```
gave92, line 232: Should we put this outside the "if"?

yaichenbaum: It shouldn't make a difference, the main thing is that the eject option shouldn't show for a regular pinned item.

gave92: You sure? Because user pinned folders IsDefaultLocation is false I think

gave92 (dismissed):

gave92 (approved):

PR #2391, opened 2020-11-19
Updated acrylic tints

yaichenbaum:


mdtauk:
> * The sidebar is less transparent now in dark mode
> * The title bar is less transparent now in dark mode
> * The TintColor is now the same as the FallbackColor in dark mode

Have you looked at this issue from the WinUI repo?
[Update acrylic theme resources](https://github.com/microsoft/microsoft-ui-xaml/issues/3478)

yaichenbaum:
@mdtauk I just took a look at it, it looks like those are in app brushes but perhaps I saw that wrong.

mdtauk:
> @mdtauk I just took a look at it, it looks like those are in app brushes but perhaps I saw that wrong.

The ones with **HostBackdrop** are Background Acrylic brushes

PR #2418, opened 2020-11-22
Grouping Contents

mbeps:
Adding the ability to group common contents in a folder. For example, group common file types.


yaichenbaum:
@mbeps It looks like the commit history got a little messed up.

PR #2425, opened 2020-11-24
[Delay bug] - Move sorting to background task

gave92:
This PR:

- Changes the logic so the items loading operation happens in a single background task
This PR wraps the whole file loading loop with a `Task.Run()` and once every 64 items it updates the UI
Also it tries to do the sorting in the background thread instead of the UI thread

- Revert edits to LoadExtendedItemProperties
The changes currently made in the delay bug branch do not appear to give any performance benefits (fulltrust process is still called for each file) but it breaks thumbnails for hidden and shortcut items

### DIFF
--- a/Files/View Models/ItemViewModel.cs
+++ b/Files/View Models/ItemViewModel.cs
@@ -795,20 +795,18 @@ namespace Files.Filesystem
                         && response.Message.ContainsKey("Enumerate"))
                     {
                         var folderContentsList = JsonConvert.DeserializeObject<List<ShellFileItem>>((string)response.Message["Enumerate"]);
+                        var tempList = new List<ListedItem>();
                         for (int count = 0; count < folderContentsList.Count; count++)
                         {
                             var item = folderContentsList[count];
                             var listedItem = AddFileOrFolderFromShellFile(item, returnformat);
                             if (listedItem != null)
                             {
-                                await CoreApplication.MainView.ExecuteOnUIThreadAsync(() =>
-                                {
-                                    _filesAndFolders.Add(listedItem);
-                                });
+                                tempList.Add(listedItem);
                             }
-                            if (count % 300 == 0)
+                            if (count == 32 || count % 300 == 0 || count == folderContentsList.Count - 1)
                             {
-                                var orderedList = OrderFiles2(_filesAndFolders.ToList());
+                                var orderedList = OrderFiles2(tempList);
                                 await CoreApplication.MainView.ExecuteOnUIThreadAsync(() =>
                                 {
                                     OrderFiles(orderedList);
@@ -979,6 +977,8 @@ namespace Files.Filesystem
                 {
                     await Task.Run(async () =>
                     {
+                        var tempList = new List<ListedItem>();
+                        var hasNextFile = false;
                         do
                         {
                             if (((FileAttributes)findData.dwFileAttributes & FileAttributes.System) != FileAttributes.System)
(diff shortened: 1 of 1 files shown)


yaichenbaum:
@gave92 This PR seems to bring back the delay with displaying items in the UI.

Files/View Models/ItemViewModel.cs
```diff
@@ -803,34 +784,50 @@ public async Task EnumerateItemsFromSpecialFolderAsync(string path)
...
+                        for (int count = 0; count < folderContentsList.Count; count++)
                         {
-                            OrderFiles();
+                            var item = folderContentsList[count];
+                            var listedItem = AddFileOrFolderFromShellFile(item, returnformat);
+                            if (listedItem != null)
+                            {
+                                tempList.Add(listedItem);
+                            }
+                            if (count % 64 == 0)
```
yaichenbaum, line 814: Did you change the number to 64 on purpose?

gave92: No could be "300" i haven't tested if there's an improvement

gave92: No, it could be "300". But since I wait at least X elements to begin showing them on the UI that number shouldn't be too high.
I'll test if there's any improvement one way or the other.

gave92:
@yaichenbaum from my testing it should reduce the overall time to load at the expense of a slightly higher delay before showing the first items (I wait 64 items, while in the base branch each item is shown when it comes)
Do you confirm this? I guess it's a question of finding the right balance

Files/View Models/ItemViewModel.cs
gave92: My comment deleted as no longer relevant

gave92:
Ok with the latest commit I'm getting pretty good results both in overall loading time and delay before first items are shown.
Please let me know if you're tests confirm this ;)

Recap of the changes in this PR:
- Revert edits to LoadExtendedItemProperties
- Fixed an issue where clicking the "up" button while a folder is loading would not list the contents of the parent folder
- Changed loading logic which in my testing has this benefits
    - Better looking loading for small folders -> in the base branch at the beginning user will see files and folders mixed/unordered, with this PR items shown to the user are always ordered
    - Better overall loading time

Note: you'll see that I commented `//_filesAndFolders.BeginBulkOperation();`. While this is not required it looks like it's not hurting performance and allows the user to start scrolling down the list while it's loading

yaichenbaum:
@gave92 Things are really fast now! For some reason `C:\` doesn't load but sub folders seem to work fine.

gave92:
@yair100 there may be something wrong
- Could you check if it shows items at all if you turn the "show hidden files" option off?
- Could you check if the items count is correct/does it show all the items?

yair100:
> @yair100 there may be something wrong
> 
> * Could you check if it shows items at all if you turn the "show hidden files" option off?

It looks like all items come up when the setting is off. When it's on nothing come up.

> * Could you check if the items count is correct/does it show all the items?

The item count matches however many items loaded.

gave92:
Thanks
(1) It looks like all items come up when the setting is off. -> It' the reverse for me xD but so yes there's something wrong
(2) The item count matches however many items loaded. -> sorry I meant if the count matches explorer/if the same items are shown but this is secondary wrt to (1)

duke7553:
@gave92 Thanks. You just beat me to pushing that same fix. 🙂

lukeblevins (approved):
LGTM

PR #2440, opened 2020-11-25
Fix CloseHandle definition

gave92:
Hi @tsvietOK 
In the current master version I get an immediate crash when trying to eject drives from Files.
Crash only happens in Release mode.
This PR solves this by defining the correct DLL for the `NativeIoDeviceControlHelper.cs->CloseHandle` function, dll taken from [here](https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-apis).
Do you agree with this solution?


tsvietOK (approved):
Good catch! Thank you!

PR #2446, opened 2020-11-26
Implemented the search feature

### LINKED ISSUE #333: Implement search options
When searching nothing happens.

duke7553:
For code review purposes. We implement search which allows for recursive queries inside any normal directory. We also aim to support hidden items with this PR.

Limitations: Only indexed locations are queried for performance reasons. We'll add a toggle to change this in a future PR

Fixes #333


yaichenbaum:
- It looks like the search box cuts off other items in the navigation bar when you make the windows size smaller.
- We should hide and clear the text in the search box if it loses focus.

Files/BaseLayout.cs
```diff
@@ -242,46 +242,61 @@ private void NotifyPropertyChanged([CallerMemberName] string propertyName = "")
             PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
         }
 
+        bool IsSearchResultPage = false;
```
d2dyno1, line 245: Move this variable to the top?

d2dyno1:
Hmmm.. It seems like it does not display any search results for me no matter what I search for

Sank30:
Search logo shown as a greyed out for my every location.

Files/BaseLayout.cs
duke7553: @d2dyno1 Thanks! Good catch.

yaichenbaum (approved):

yaichenbaum (approved):

yaichenbaum:
@duke7553 Great work!

PR #2459, opened 2020-11-27
Fixed an issue where clicking the back button on the settings page in tablet mode wouldn't work

### LINKED ISSUE #1682: Tablet mode back button doesn't work on settings page
I don't know if this is a bug or enhancement because I didn't implement it but ideally the back button when used in the settings page should go to the main page with tabs where you can access your directories.

yaichenbaum:
Fixes #1682

### DIFF
--- a/Files/Views/Settings.xaml.cs
+++ b/Files/Views/Settings.xaml.cs
@@ -2,6 +2,7 @@
 using Files.View_Models;
 using Windows.ApplicationModel.Core;
 using Windows.ApplicationModel.Resources.Core;
+using Windows.UI.Core;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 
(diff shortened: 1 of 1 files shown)


gave92 (approved):

Files/Views/Settings.xaml.cs
```diff
@@ -29,6 +33,16 @@ public Settings()
             SettingsPane.SelectedItem = SettingsPane.MenuItems[0];
         }
 
+        private void OnBackRequested(object sender, BackRequestedEventArgs e)
+        {
+            Frame rootFrame = Window.Current.Content as Frame;
+            if (rootFrame.CanGoBack)
+            {
+                rootFrame.GoBack();
+                e.Handled = true;
```
gave92, line 42: Will throw null exception

gave92 (changes requested):

Files/Views/Settings.xaml.cs
d2dyno1: `e?.Handled = true;` should do it

gave92 (approved):

Issue #1838, opened 2020-08-29
Properties: Move "Cancel" button to the right


XPoppyX:
The "ok" button should always be on the left, and "Cancel" on the right

![Anotación 2020-08-28 220552](https://user-images.githubusercontent.com/61259627/91626160-a9aa1780-e97a-11ea-863f-0b97cabde0a8.png)

PR #2484, opened 2020-11-29
Fixed an issue where the text in the properties dialog would get cut off in some languages

### LINKED ISSUE #2410: Make text in the properties dialog responsive
When using the German localisation, the right attribute text is cut off due to the window being too narrow.

image

yair100:
- Fixed an issue where the text in the properties dialog would get cut off in some languages.
- Changed the "ok" button in the properties dialog to "apply"

Fixes #2410

### DIFF
--- a/Files/MultilingualResources/Files.de-DE.xlf
+++ b/Files/MultilingualResources/Files.de-DE.xlf
@@ -575,6 +575,10 @@
           <source>Tiles View</source>
           <target state="translated">Kacheln</target>
         </trans-unit>
+        <trans-unit id="PropertiesDialogOKButton.Content" translate="yes" xml:space="preserve">
+          <source>OK</source>
+          <target state="translated">OK</target>
+        </trans-unit>
         <trans-unit id="AccessDeniedDeleteDialog.Text" translate="yes" xml:space="preserve">
           <source>We weren't able to delete this item</source>
           <target state="translated">Löschen der Datei oder des Ordners nicht möglich</target>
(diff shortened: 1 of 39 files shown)


d2dyno1:
"Apply" is inconsistent with Windows design guidlines

yaichenbaum:
> "Apply" is inconsistent with Windows design guidlines

Do you have a link?

d2dyno1:
@yair100 #1838 (#1851)

yair100:
> @yair100 #1838 (#1851)

Do you have any links regarding the guidelines? The links you shared are regarding the order, not the text.

d2dyno1:
https://docs.microsoft.com/en-us/windows/win32/uxguide/win-property-win#guidelines

> OK buttons
> For owner property windows, the OK button means apply the pending changes (made since the window was opened or the last Apply), and close the window

-

> Apply buttons
> For owner property sheets, the Apply button means apply the pending changes (made since the window was opened or the last Apply), but leave the window open

If we want to keep the Apply button, it must not close the window, it should only save the changes.

yaichenbaum:
> https://docs.microsoft.com/en-us/windows/win32/uxguide/win-property-win#guidelines
> 
> > OK buttons
> > For owner property windows, the OK button means apply the pending changes (made since the window was opened or the last Apply), and close the window
> 
> > Apply buttons
> > For owner property sheets, the Apply button means apply the pending changes (made since the window was opened or the last Apply), but leave the window open
> 
> If we want to keep the Apply button, it must not close the window, it should only save the changes.

That's a good point, we should probably have all three buttons.

lukeblevins (approved):
changes LGTM

PR #2487, opened 2020-11-30
Fixed an issue where opening files wouldn't work when the directory was sorted by size or by file type

### LINKED ISSUE #2475: selected file can not be opend with a double click, only on the context Menue, with selecting open with.
Hey, I'm using Files 0.21.0. Its a very cool tool, thanks. 
Sinc the last update I have the impression, that you can not open a fiel on double-Clicking on it. YOu have to go the contextmenu an select ther "open with-..."

Lubieerror:
DefaultQuery doesn't work because it's always overwritten(?) with null property* from sortEntry.PropertyName.
Might be necessary to write custom replacement in the future to achieve full functionality.

*(or just adding null causes problems)

Fixes #2475

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -506,8 +506,10 @@ namespace Files.Interacts
                                     //We can have many sort entries
                                     SortEntry sortEntry = new SortEntry()
                                     {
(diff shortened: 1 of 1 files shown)


Lubieerror:
Should solve #2475 issue.

yaichenbaum (approved):

yaichenbaum:
@Lubieerror Great job!

PR #2497, opened 2020-11-30
Reduce unnecessary work during extended property initialization and use new item icon in GridView

duke7553:
For the DataGrid layout (ListView) we no longer fire the EffectiveViewportChanged event on every item whenever they are scrolled into view. Since we now use the built-in LoadingRow event instead, this has the implication of reducing unneeded calculations for each item. It wouldn't be unreasonable to predict this change has a small, albeit positive impact on scrolling in directories with lots of rows loaded. 

Additionally, we reduce unnecessary casts with both layouts and use the FluentUI asset for the uninitialized thumbnail icon in GridView.


yaichenbaum (approved):

yaichenbaum:
Great work!

PR #2505, opened 2020-12-01
Right Click Animation Direction

Olcay91:
If the mouse is on the upper half the context menu is above the mouse cursor, but the animation is starting at the top instead of at the position where the cursor is. It would feel more natural if the animation would start from the cursors position upwards.

Image of Yaktocat


yaichenbaum:
@Olcay91 You can open an issue [here](https://github.com/files-community/Files/issues/new/choose).

PR #2508, opened 2020-12-01
Install updates from Files

### LINKED ISSUE #1513: Ability of checking updates actively
An app can check for updates actively and request silent downloads and installation, see 
https://docs.microsoft.com/en-us/windows/uwp/packaging/self-install-package-updates.
It can keep users who disabled Automatically Update in Microsoft Store up-to-date, and it can also inform users about the up-coming update contents.

I think it's helpful for improving user experience.

Thomas1664:
Fixes #1513 
I implemented the feature to install updates via a button in the about settings page.


yaichenbaum:
@Thomas1664 Instead of adding a button to check for updates, it should autmaticly check for udpates when launching the app.

Thomas1664:
I didn't want to implement it yet because if the code is wrong the app crashes

yaichenbaum:
> I didn't want to implement it yet because if the code is wrong the app crashes

That's fine, we will test it in a flight group to make sure it works correctly first, and if needed, we can wrap it in a try/catch.

Jaiganeshkumaran:
> @Thomas1664 Instead of adding a button to check for updates, it should autmaticly check for udpates when launching the app.

No need. It will slow down the app startup. Also this app is available on the Store so it will check for updates automatically. Users will click the update button only if the store doesn't check,.

Files/Views/SettingsPages/About.xaml
```diff
@@ -188,6 +188,8 @@
                             VerticalAlignment="Center"
                             Text="License:" />
                         <HyperlinkButton NavigateUri="https://github.com/QL-Win/QuickLook/blob/master/LICENSE-GPL.txt">GPL</HyperlinkButton>
+                        <Button Content="Check for updates" Click="UpdateCheck_Click">
```
yaichenbaum, line 191: Please remove the button and have the code checking for updates run when the app is first launched.

yair100 (changes requested):

Files/Views/SettingsPages/About.xaml
tsvietOK: @yair100 I don't think it is a good idea to show update dialog on every app launch.

yaichenbaum: There is no need to show a dialog every time the app is opened, the only time a dialog will be displayed is if a mandatory update is detected.

boscorelly: i don't think it must be at startup by default. Maybe an option. Noting more, as store is already updating apps

yaichenbaum: @boscorelly This pull request is regarding requiring mandatory updates.

yaichenbaum:
image
@Thomas1664 Is this because the app is not connected to the Store in our debug builds?

Thomas1664:
Yes

yair100 (dismissed):

yair100 (dismissed):

yair100 (approved):

tsvietOK (approved):

myokeeh:
> Yes

How did you get around this?

Thomas1664:
> > Yes
> 
> How did you get around this?

See the attatched issue. We used a flight group.

myokeeh:
> See the attached issue. We used a flight group.

interesting. i've been struggling with this. I have a flight group also, but this still happens.

yaichenbaum:
@myokeeh We ended up wrapping this in a try/catch, for some reason we still get the error every time and updates are not detected, but at least the app isn't crashing.

myokeeh:
@yair100, so `GetAppAndOptionalStorePackageUpdatesAsync` is in fact still broken--no updates detected. 🤷‍♂️I hope somebody at Microsoft gets on this and get it fixed.

Issue #2540, opened 2020-12-05
Default tab focus move to explorer region.


EricChen1248:
**Is your feature request related to a problem? Please describe.**
The default tab focus on the explorer screen (and probably others too, I've noticed the properties as well) are not very useful, and unhelpful for keyboard based navigation. Currently you need 6 "tab presses" to switch to the actual explorer region before being able to navigate with arrow keys to select folders/drives.

**Describe the solution you'd like**
Change the default tab focus to either the explorer region. Default tab focus being in the top menu/new tab area is rather pointless, as you cannot use enter key there, and all actions can already completed with other keyboard shortcuts (ctrl + t and ctrl + w).

**Describe alternatives you've considered**
Default in the side panel is also an option I think, at least still much more useful than the current default.

**Additional context**
None

PR #2541, opened 2020-12-05
Remove tab stop to top navigation bar

EricChen1248:
The "New Tab" containers and the close buttons are part of the tab stop cycle, however, being in that cycle is pretty pointless, as they do not allow for any other keyboard access (enter key doesn't work, all actions can be completed with ctrl + t / ctrl + w anyway).
I removed them from being set as tab stops to make it more friendly for keyboard users to access.

This is related to feature-request #2540 that I submitted earlier. There are other parts that would be more similar to what I wish to do in the submission, but involves more parts that I'm afraid might cause some unforeseen side effects that at my skill level I wouldn't notice.
Signed-off-by: Eric Chen <ericchen1248@gmail.com>

### DIFF
--- a/Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml
+++ b/Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml
@@ -107,6 +107,7 @@
                                                 CornerRadius="2"
                                                 HighContrastAdjustment="None"
                                                 IsTextScaleFactorEnabled="False"
+                                                IsTabStop="False"
                                                 Style="{StaticResource TabViewButtonStyle}"
                                                 Visibility="{Binding IsAddTabButtonVisible, RelativeSource={RelativeSource TemplatedParent}}" />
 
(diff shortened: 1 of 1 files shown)


Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml
```diff
@@ -107,6 +107,7 @@
                                                 CornerRadius="2"
                                                 HighContrastAdjustment="None"
                                                 IsTextScaleFactorEnabled="False"
+                                                IsTabStop="False"
```
yaichenbaum, line 110: The add tab button should still have the tab stop option.

Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml
```diff
@@ -834,6 +835,7 @@
                         FontFamily="{StaticResource FluentUIGlyphs}"
                         Header="{x:Bind Header, Mode=OneWay}"
                         IconSource="{x:Bind IconSource, Mode=OneWay}"
+                        IsTabStop="False"
```
yaichenbaum, line 838: Can you confirm that the first tab can still be focused using the tab key?

EricChen1248: No, this disables it. If the plan is still keep them on the tab cycle then this PR isn't useful. I wasn't thinking anyone would want to tab onto these. I'm closing the PR then, thanks for your time.

Issue #2479, opened 2020-11-28
Add a smaller version of the GridView layout mode


BooBerry:
I'll keep this simple, I feel Layout mode > Grid View (Small)'s icon size is still too large, even on a 1440p screen (with Windows' scaling unchanged at 100%). Grid View (Small) would be more-or-less equal to File Explorer's View > Large icons in this case, and there's no size smaller than this (I use File Explorer's View > Medium icons by default in Windows 10).

**Is your feature request related to a problem? Please describe.**
Even with layout mode set to Grid Mode (Small) the icons still are too large, in my opinion. It's an aesthetic type of problem, but a problem nonetheless.

**Describe the solution you'd like**
A new option for an even smaller Grid View, e.g. **Grid View (Smaller)**. Grid View (Smaller)'s icons should be more-or-less equal to the size of File Explorer's View > Medium icons option.

**Describe alternatives you've considered**
The only alternative right now is Grid View (Small), but it leaves a lot to be desired as it's too large.

**Additional context**
Since this app's goal is to reach parity with File Explorer, adding this new Grid View option would make sense.

yaichenbaum:
This issue should be resolved with #2550.

PR #2550, opened 2020-12-06
Scaled down the Layout Sizes for GridView

d2dyno1:
It's also a codebase quality improvement with addition of class `Constatnts` where all (current and future) constant values SHOULD BE stored for maintainability purposes.

Resloves #2479

### DIFF
--- a/Files/Constants.cs
+++ b/Files/Constants.cs
@@ -6,9 +6,9 @@
         {
             public static class GridViewBrowser
             {
-                public const int GRIDVIEW_SIZE_LARGE = 275;
+                public const int GRIDVIEW_SIZE_LARGE = 225;
 
-                public const int GRIDVIEW_SIZE_MEDIUM = 175;
+                public const int GRIDVIEW_SIZE_MEDIUM = 150;
 
                 public const int GRIDVIEW_SIZE_SMALL = 100;
             }
--- a/Files/Constants.cs
+++ b/Files/Constants.cs
@@ -1,0 +1,21 @@
+﻿namespace Files
+{
+    public static class Constants
+    {
+        public static class Browser
+        {
+            public static class GridViewBrowser
+            {
+                public const int GRIDVIEW_SIZE_LARGE = 275;
+
+                public const int GRIDVIEW_SIZE_MEDIUM = 175;
+
+                public const int GRIDVIEW_SIZE_SMALL = 100;
+            }
+
+            public static class GenericFileBrowser
+            { 
+            }
+        }
+    }
+}
--- a/Files/Constants.cs
+++ b/Files/Constants.cs
@@ -6,9 +6,9 @@
         {
             public static class GridViewBrowser
             {
-                public const int GRIDVIEW_SIZE_LARGE = 275;
+                public const int GRIDVIEW_SIZE_LARGE = 225;
 
-                public const int GRIDVIEW_SIZE_MEDIUM = 175;
+                public const int GRIDVIEW_SIZE_MEDIUM = 160;
 
                 public const int GRIDVIEW_SIZE_SMALL = 100;
             }
--- a/Files/Constants.cs
+++ b/Files/Constants.cs
@@ -6,11 +6,11 @@
         {
             public static class GridViewBrowser
             {
-                public const int GRIDVIEW_SIZE_LARGE = 220;
+                public const int GridViewSizeLarge = 220;
 
-                public const int GRIDVIEW_SIZE_MEDIUM = 160;
+                public const int GridViewSizeMedium = 160;
 
-                public const int GRIDVIEW_SIZE_SMALL = 100;
+                public const int GridViewSizeSmall = 100;
             }
 
             public static class GenericFileBrowser
--- a/Files/Constants.cs
+++ b/Files/Constants.cs
@@ -6,9 +6,9 @@
         {
             public static class GridViewBrowser
             {
-                public const int GRIDVIEW_SIZE_LARGE = 225;
+                public const int GRIDVIEW_SIZE_LARGE = 220;
 
-                public const int GRIDVIEW_SIZE_MEDIUM = 150;
+                public const int GRIDVIEW_SIZE_MEDIUM = 160;
 
                 public const int GRIDVIEW_SIZE_SMALL = 100;
             }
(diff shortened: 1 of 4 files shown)


yaichenbaum:
@d2dyno1 Can you share before and after pictures for the different sizes?

d2dyno1:
Small:
BEFORE:
image
AFTER:
image

Medium:
BEFORE:
image
AFTER:
image

Large:
BEFORE:
image
AFTER:
image

d2dyno1:
@yair100 
I actually think Large can be scaled down even more as well as Medium

(For some reason images didn't load on current version)

Files/Constants.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public static class Constants
+    {
+        public static class Browser
+        {
+            public static class GridViewBrowser
+            {
+                public const int GRIDVIEW_SIZE_LARGE = 275;
+
+                public const int GRIDVIEW_SIZE_MEDIUM = 175;
```
yaichenbaum, line 11:
```suggestion
                public const int GRIDVIEW_SIZE_MEDIUM = 160;
```

Files/Constants.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Files
+{
+    public static class Constants
+    {
+        public static class Browser
+        {
+            public static class GridViewBrowser
+            {
+                public const int GRIDVIEW_SIZE_LARGE = 275;
```
yaichenbaum, line 9:
```suggestion
                public const int GRIDVIEW_SIZE_LARGE = 240;
```

d2dyno1:
I've made them smaller

yaichenbaum:
> I've made them smaller

Please make sure the numbers are divisible by 4.

Files/Constants.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Files
+{
+    public static class Constants
+    {
+        public static class Browser
+        {
+            public static class GridViewBrowser
+            {
+                public const int GRIDVIEW_SIZE_LARGE = 220;
```
tsvietOK, line 9: Fix naming, please. `GridViewSizeLarge`

d2dyno1: I intentionally named it with uppercase letters. The constant `const` variables should be distinguishable from other non-constants. We should update the naming scheme for constant values

tsvietOK: @d2dyno1 Do not use underscores (_) in between words in variables. ([Source](https://www.c-sharpcorner.com/blogs/naming-conventions-in-c-sharpnet))
[Code style](https://github.com/files-community/Files/wiki/Code-style)

d2dyno1: The uppercase with `_` is a C/C++ naming convention which I think is superior to the provided sources above. It removes any misconceptions between variables and `const` variables. Obviously, by default constants in c# are PascalCase but ignoring that, we would remove any naming ambiguities.

tsvietOK: If you use C# code in your PR, you have to follow C# rules and our local code style.

Files/View Models/SettingsViewModel.cs
```diff
@@ -593,10 +593,10 @@ public Type GetLayoutType()
 
         private void DetectGridViewSize()
         {
-            _GridViewSize = Get(125, "GridViewSize"); // Get GridView Size
+            _GridViewSize = Get(Constants.Browser.GridViewBrowser.GRIDVIEW_SIZE_SMALL, "GridViewSize"); // Get GridView Size
         }
 
-        private int _GridViewSize = 125; // Default Size
+        private int _GridViewSize = Constants.Browser.GridViewBrowser.GRIDVIEW_SIZE_SMALL; // Default Size
```
tsvietOK, line 599: Can you change name of field too? `gridViewSize`

tsvietOK (changes requested):

Files/View Models/SettingsViewModel.cs
d2dyno1: I have not named it like that but I'll change it

Files/Constants.cs
d2dyno1: Too much PascalCases are making the code confusing and I'm sure uppercase constant is not a thing that any developer would find new

d2dyno1: @yair100 what's your final word on that?

yaichenbaum: @d2dyno1 The goal is to have a consistent codebase, this is an area where everyone has their own preference and habits, therefore it's something we discussed a while back and this is what we settled on.

d2dyno1: As you wish

lukeblevins: @yair100 Sorry, but it should be the convention to have all constant values as UPPERCASE_NAMES.

duke7553: There must be a way to easily distinguish between normal global variables and constants, and I've always heard the convention is to have constants declared as uppercase.

yaichenbaum: > @yaichenbaum Sorry, but it should be the convention to have all constant values as UPPERCASE_NAMES.

In that case, can you update the Wiki page because right now it doesn't specify that.

lukeblevins: @yair100 Actually, scratch that. Uppercase isn't consistent with StyleCop and C# specifically. This is an area where C language and C# diverge.

lukeblevins (dismissed):

Dylan-Osborne:
@yair100 @d2dyno1 I understand that some users prefer much smaller icon sizes and I don't have an issue with the default sizes being made smaller. But I do think that the sizes for GridView should be able to scroll a larger than the max in this pull request (GRIDVIEW_SIZE_LARGE is 220). I personally prefer and have used a size of around 275 up to now and I'm sure that there are others who also used sizes that will no longer be supported.

If I'm correct, Windows Explorer allows the thumbnail view to go up to 256x256 (with the padding and name comes to about 270x275) so surely Files should also allow to somewhere around there?

This shouldn't be difficult to implement, it would basically be an added constant e.g. GRIDVIEW_SIZE_MAX that would be the upper limit for ctrl+scroll instead of using GRIDVIEW_SIZE_LARGE as the upper limit.

yair100 (approved):

tsvietOK (approved):

tsvietOK:
@Dylan-Osborne If you think that grid view items are too small, you can create issue about it(if such issue is not exist). It is not a part of this PR.

yaichenbaum:
@Dylan-Osborne These sizes are just the defaults, you should still be able to increase the size by using the `ctrl + +` shortcut.

PR #2561, opened 2020-12-08
Increased max GridView size

### LINKED ISSUE #2560: Increase max GridView size for ctrl + scroll
In the current master branch, the max GridView size achievable by _ctrl + scroll_ is 220. It would be really appreciated if that max could be increased without modifying the default GridView layout sizes in the master branch ie. leaving the large, medium and small defaults as 220, 160 and 100 as set in #2479

Increase the max GridView size achievable by _ctrl + scroll_. Windows Explorer's "Extra large icons" are approximately 270x275, so somewhere around there should be good, but it may be worth setting the max around 320 just to give users a little extra flexibility.

**Additional context**
Max GridView of 220 (as in master)
Max 220

Proposed Max GridView of 320 (only achievable by _ctrl + scroll_ or _ctrl + +_ ie. not a default in the layout options flyout)
Max 320

If you're happy with this proposal, can you assign me to it?

Dylan-Osborne:
Fixes #2560

Allows a user to use _ctrl + scroll_ or _ctrl + +_ to resize the GridView to larger than the "Grid View (Large)" default. No new menu option is added and no changes are made to the existing menu sizes.

Max before:
Max 220

Max after:
Max 320

### DIFF
--- a/Files/Constants.cs
+++ b/Files/Constants.cs
@@ -6,6 +6,8 @@
         {
             public static class GridViewBrowser
             {
+                public const int GridViewIncrement = 20;
+
                 public const int GridViewSizeMax = 300; // Max achievable ctrl + scroll, not a default layout size.
 
                 public const int GridViewSizeLarge = 220;
--- a/Files/Constants.cs
+++ b/Files/Constants.cs
@@ -6,6 +6,8 @@
         {
             public static class GridViewBrowser
             {
+                public const int GridViewSizeMax = 320; // Max achievable ctrl + scroll, not a default layout size.
+
                 public const int GridViewSizeLarge = 220;
 
                 public const int GridViewSizeMedium = 160;
--- a/Files/Constants.cs
+++ b/Files/Constants.cs
@@ -6,7 +6,7 @@
         {
             public static class GridViewBrowser
             {
-                public const int GridViewSizeMax = 320; // Max achievable ctrl + scroll, not a default layout size.
+                public const int GridViewSizeMax = 300; // Max achievable ctrl + scroll, not a default layout size.
 
                 public const int GridViewSizeLarge = 220;
 
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -745,7 +745,7 @@ namespace Files
 
         public void GridViewSizeIncrease(KeyboardAccelerator sender, KeyboardAcceleratorInvokedEventArgs args)
         {
-            AppSettings.GridViewSize = AppSettings.GridViewSize + 25; // Make Larger
+            AppSettings.GridViewSize = AppSettings.GridViewSize + Constants.Browser.GridViewBrowser.GridViewIncrement; // Make Larger
             if (args != null)
             {
                 args.Handled = true;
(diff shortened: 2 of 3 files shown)


Files/Constants.cs
```diff
@@ -6,6 +6,8 @@ public static class Browser
         {
             public static class GridViewBrowser
             {
+                public const int GridViewSizeMax = 320; // Max achievable ctrl + scroll, not a default layout size.
```
yaichenbaum, line 9:
```suggestion
                public const int GridViewSizeMax = 300; // Max achievable ctrl + scroll, not a default layout size.
```

yair100 (changes requested):

Files/Constants.cs
Dylan-Osborne: Updated to 300 as requested

Dylan-Osborne:
@yair100 If I remember correctly, you wanted the sizes to be multiples of 4, do you also want the GridView size increments to be a multiple of 4 e.g. 20 instead of the 25 it currently is?
https://github.com/files-community/Files/blob/e0da5d6482a4e35d7bdfd23cb49d82c1c054d3b4/Files/BaseLayout.cs#L748
https://github.com/files-community/Files/blob/e0da5d6482a4e35d7bdfd23cb49d82c1c054d3b4/Files/BaseLayout.cs#L757

yair100:
> @yair100 If I remember correctly, you wanted the sizes to be multiples of 4, do you also want the GridView size increments to be a multiple of 4 e.g. 20 instead of the 25 it currently is?
> https://github.com/files-community/Files/blob/e0da5d6482a4e35d7bdfd23cb49d82c1c054d3b4/Files/BaseLayout.cs#L748
> 
> 
> https://github.com/files-community/Files/blob/e0da5d6482a4e35d7bdfd23cb49d82c1c054d3b4/Files/BaseLayout.cs#L757

Can you change that to `24`?

Dylan-Osborne:
@yair100 I can do so, but is there a specific reasoning for 24? Remember that the allowed sizes range from 100 to 300 and 24 isn't a multiple of any of them which means if you choose a default and then _crtl + scroll_ to adjust you may end up with an odd final adjustment like this:

whereas an increment of 20 is smooth to the largest possible size achievable by _ctrl + scroll_


Basically the sizing with 24 works as follows:
220 (default size) --> 244 --> 268 -> 292 --> 316 --> 320 (max is capped at 320)
while 20 is consistent to the end:
220 --> 240 --> 260 --> 280 --> 300 --> 320 (max)

yaichenbaum:
@Dylan-Osborne Good catch, let's go with 20.

mdtauk:
Will there be consistent naming with the various set values of Icon Sizes, now they have changed.

What about providing a slider in the context menus, with ticks for the set sizes?

Dylan-Osborne:
@yair100 Thanks, I've set the increment to 20

@mdtauk The changes in this PR are only accessible by using _ctrl + +_ and _ctrl + scroll_ to fine tune the GridView size. The existing default GridView sizes accessible from the layout flyouts remain the same (small, medium and large). So there's no naming that should need to change.

yaichenbaum:
> Will there be consistent naming with the various set values of Icon Sizes, now they have changed.
> 
> What about providing a slider in the context menus, with ticks for the set sizes?

@mdtauk Can you open a new proposal for this? Luke started work on a concept a couple of months ago that's similar to what you are describing.
image

yair100 (approved):

mdtauk:
> > Will there be consistent naming with the various set values of Icon Sizes, now they have changed.
> > What about providing a slider in the context menus, with ticks for the set sizes?
> 
> @mdtauk Can you open a new proposal for this? Luke started work on a concept a couple of months ago that's similar to what you are describing.
> image

#2575

PR #2594, opened 2020-12-13
Enabled parallel builds to improve the build time

yair100:


gave92:
@yair100 ~Is there a way to check whether it's actually building in parallel? The build still took 44 minutes, pretty much like all the others~
Edit: well it *is* adding the `/m` switch when building, I just was expecting a bigger difference

yaichenbaum:
@gave92 I am not sure, this doesn't guarantee that it will build in parallel as it depends on how many other builds are happening at the same time from other pull request. I previously tested the change when no other builds were running, and the build time was 35 minutes.

gave92 (approved):

PR #2595, opened 2020-12-13
Fixed an issue where the keyboard focus would get stuck within the file browser page, status bar and grid splitter

### LINKED ISSUE #2477: Keyboard focus navigation gets stuck within file browser page, status bar and grid splitter
https://app.box.com/s/d6izg7ga9gn5bvbz0sn0zqwr22iz09yl

yaichenbaum:
Fixes #2477


gave92 (approved):
I can confirm the changes fix the issue 👍 
Curious: why is adding `TabFocusNavigation="Cycle"` necessary? Why is the tab getting stuck otherwise?

yaichenbaum:
> I can confirm the changes fix the issue 👍
> Curious: why is adding `TabFocusNavigation="Cycle"` necessary?

I think it's because if we didn't do that, the focus would get stuck in the status bar.

PR #2606, opened 2020-12-15
Added the option to list and sort directories alongside files

### LINKED ISSUE #1929: List and sort directories alongside files
Directories contain files and sub-directories. When listing a directory, I expect the contents to be sorted by date or alphabetically. However, the typical design pattern on Windows is to have a list of directories sorted then list of files. Here is an illustration of a situation where this causes a problem: when I unzip a directory from a zip flie in my downloads folder, I need to remember the zip files name, scroll up and find the name in the list of directories, 

An option in settings to list directories alongside files

I'm sure this isn't to everyone's taste. But this pattern of separating files and folder only happens on Windows. Mac and Linux typically don't use this pattern. I'm not sure why people like this Windows pattern. Maybe because Windows doesn't have this feature: #1928

**Additional context**
Could you point me to the code where this would be implemented? I'd like to try make a PR for learning purposes.

yaichenbaum:
Fixes #1929

### DIFF
--- a/Files/Views/SettingsPages/FilesAndFolders.xaml
+++ b/Files/Views/SettingsPages/FilesAndFolders.xaml
@@ -44,6 +44,12 @@
                     Header="Open items with a single click"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{x:Bind AppSettings.OpenItemsWithOneclick, Mode=TwoWay}" />
+
+                <ToggleSwitch
+                    x:Uid="SettingsListAndSortDirectoriesAlongsideFiles"
+                    Header="List and sort directories alongside files"
+                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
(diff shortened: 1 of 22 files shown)


Files/Views/SettingsPages/FilesAndFolders.xaml
```diff
@@ -44,6 +44,12 @@
                     Header="Open items with a single click"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{x:Bind AppSettings.OpenItemsWithOneclick, Mode=TwoWay}" />
+
+                <ToggleSwitch
+                    x:Uid="SettingsListAndSortDirectoriesAlongsideFiles"
+                    Header="List and sort directories alongside files"
+                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
+                    IsOn="{x:Bind AppSettings.ListAndSortDirectoriesAlongsideFiles, Mode=TwoWay}" />
```
duke7553, line 52: I'd place it here, because I'm not sure where else it belongs.

duke7553 (approved):
Great work!

Issue #2537, opened 2020-12-04
Opening the properties window from the sidebar crashes the app after updating to WinUI 2.5


yaichenbaum:
`2020-12-07 00:55:31.1217|ERROR|.ctor|The application called an interface that was marshalled for a different thread.`

lukeblevins:
@yair100 Does the log file contain sufficient information for this crash?

yaichenbaum:
@duke7553 I updated the description.

osieltorres:
I have to say that the crash happens on every possible partition, not only for system, and not only from the sidebar.

winston-de:
I believe the source of the issue is the NavigationView in Properties.xaml as the window will launch properly when it is removed.

winston-de:
Using Windows.Ui.NavigationView instead of Microsoft.Ui.Xaml.NavigationView seems to be a workaround to this issue.

PR #2612, opened 2020-12-16
Fixed an issue where opening the properties window would crash the app

winston-de:
I am mostly certain that issue #2537 is caused by a bug in the Microsoft.Ui.Xaml NavigationView.
Switching to the Windows.Ui NavigationView seems to be an effective workaround to this bug.

### DIFF
--- a/Files/Views/Pages/Properties.xaml
+++ b/Files/Views/Pages/Properties.xaml
@@ -31,7 +31,7 @@
             VerticalAlignment="Center"
             Text="Properties" />
 
-        <muxc:NavigationView
+        <NavigationView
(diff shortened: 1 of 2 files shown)


yair100 (approved):

yaichenbaum:
@winston-de Great work!

winston-de:
@yair100 Thanks!
Do you think I should file a bug report in the WinUi repository for this?

yair100:
> @yair100 Thanks!
> Do you think I should file a bug report in the WinUi repository for this?

Definitely, this change should be considered a work around until the issue is fixed on the WinUI end.

PR #2615, opened 2020-12-16
Increase CoreDispatcherPriority of adding drive sidebar items to "Normal"

duke7553:
This may/may not have an impact and requires testing.


yaichenbaum:
@duke7553 It looks like there is still a small delay but it's less than what there was before.

yaichenbaum (approved):

PR #2622, opened 2020-12-17
Added the option to hide columns in the details layout

### LINKED ISSUE #380: Add the option to hide columns in the details layout mode
We should add the option to edit which columns show.

yaichenbaum:
Please note, hiding a column won't work unless you toggle the item and then hide it again afterwards, this is a known issue with the setup of the SettingsViewModel and it shouldn't hold up merging this pull request.
Fixes #380


duke7553 (approved):

gave92 (approved):
Works good with the exception of that known bug

PR #2623, opened 2020-12-17
Show localized names for system folders

### LINKED ISSUE #1226: Show localised (localized) names for system folders

Jaiganeshkumaran:
Closes #1226

### DIFF
--- a/Files.Extensions/Files.Extensions.vcxproj
+++ b/Files.Extensions/Files.Extensions.vcxproj
@@ -105,11 +105,19 @@
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
     <ClCompile>
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <ForcedIncludeFiles Condition="'$(Configuration)|$(Platform)'=='Debug|ARM'">pch.h</ForcedIncludeFiles>
+      <ForcedIncludeFiles Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">pch.h</ForcedIncludeFiles>
+      <ForcedIncludeFiles Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">pch.h</ForcedIncludeFiles>
+      <ForcedIncludeFiles Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">pch.h</ForcedIncludeFiles>
     </ClCompile>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Release'">
     <ClCompile>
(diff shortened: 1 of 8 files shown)


duke7553 (approved):
Looks good to me 👍

gave92:
@duke7553 doesn't this slow down folder loading? it's calling the fulltrust process for every folder.
(I haven't tried though, the performance hit may be negligible)

duke7553:
@gave92 Isn't this what we already do for loading thumbnails on files?

gave92:
@duke7553 the difference is that thumbnails are loaded in `LoadExtendedItemProperties` while this calls the fulltrsut process during `AddFolder`

duke7553:
@gave92 Hmm... you're certainly correct about that.

Would it be appropriate to load the localized names on an ad-hoc basis through LoadExtendedItemProperties() ?

gave92:
If it's moved to `LoadExtendedItemProperties` you don't even need the fulltrustprocess, but can just use the StorageFile `DisplayName`.
Downside is that you briefly see the untranslated name

duke7553:
@Jaiganeshkumaran Can you make the changes based on what @gave92 suggested?

gave92:
@Jaiganeshkumaran do you have any update on this? We are getting close to releasing V1 and it would be nice if this was included :)

d2dyno1:
@Jaiganeshkumaran

gave92:
@duke7553 moving loading the localized name to LoadExtendedItemProperties unfortunately means that folders may end up not sorted in alphabetical order ("**P**ictures" becomes "**I**mmagini" in italian e.g.). Not sure if it's less bad:
- to slow down loading a bit and get the name during AddFolder
- to reorder items during LoadExtendedItemProperties
- to have items not sorted by display name

Jaiganeshkumaran:
@gave92 Maybe instead of adding directly to the main observable collection, we could add it to a temp list during enumeration, then LoadExtendedItemProperties and then put them in the main observable collection

gave92:
@Jaiganeshkumaran that would delay showing the contents to the user a lot :(
Plus 99%+ of folders have a DisplayName identical to the non-localized name so we'd be slowing down most folders for nothing.

Jaiganeshkumaran:
@gave92 Then what to do?

gave92:
Good question :)

Jaiganeshkumaran:
We could have a list of predefined folders with localized names like built-in folders and only load for them

gave92:
Yeah that would work or we could cache the association between Name and DisplayName (#2655 could help with that)

PR #2655, opened 2020-12-21
Add file list caching

jakoss:
Added cache to store visited paths. Successful read from cache **WILL NOT** cancel standard load from disc. It will just trick user to see cached results right away while still refreshing files in background.

For now i integrated this functionality as experimental and turned off be default

Further work to be done:
- [x] Use some persistent store for cache (SQLite)
- [ ] Cache preemptively - if user navigates to some directory we can check if parent directory and all children directories are cached as well. If not - we can read them in background and store in cache. This has the potential to smooth navigation out significantly **This will be introduced in another Pull Request, along with more options to customize cache size and behavior**


yaichenbaum:
@jakoss Great work! This change makes everything much faster and it will a significant improvement over the current behavior.

jakoss:
Awesome. So i'll try and make this persistent. Also the preemptive loading will make it much faster on rarely visited paths

yaichenbaum:
@jakoss It would be best if we could use one of the existing dependencies such as SQLite as opposed to adding additional ones.

jakoss:
Sure, i can try using SQLite for that. I can even use memory cache before db

yaichenbaum:
> Sure, i can try using SQLite for that. I can even use memory cache before db

If you use a memory cache, there should probably be some sort of limit to ensure that it doesn't get out of hand.

Files/Helpers/FileListCache.cs
```diff
@@ -0,0 +1,48 @@
+﻿using Files.Filesystem;
+using Microsoft.Extensions.Caching.Memory;
```
duke7553, line 2: Are .NET Platform extensions allow in UWP apps?

jakoss: As i wrote earlier - i have no prior experience with UWP so i'm not completely sure, but i don't see any reason why shouldn't it be allowed. It's targeting .NET Standard 2.0, so i believe it should be working practically everywhere. And for sure it's working on my machine :)

jakoss:
@yair100 There is a serious issue with persisting cache in current form. The `ListedItem` class is completely unserializable. To make this work we would have to do quite heavy code refactoring to introduce some intermediate file representation that can be serialized and then converted to `ListedItem` after deserialization. Currently the bitmap fields are making that impossible.

Right now, the file listing code is too tightly coupled with UI model which is `ListedItem`. I think file listing code should be completely separated and it should return some generic file/folder description, that can be serialized or used to create `ListedItem`. I hope it make sense for you.

Maybe we should release the in-memory cache as it is right now and make persisting cache a next milestone for that? I can see that it requires a lot of work to do that right

gave92:
Hi @jakoss the only ListedItem's fields difficult to serialize should be `FileIcon` and `IconOverlay`, right?
There is no need to persist such fields in the cache as they are filled after the folder enumeration has completed, so they do not impact loading performance too much.
If I were to serialize them using Json.Net I'd simply add the `[JsonIgnore]` attribute on those two field.

jakoss:
@gave92 So then i'll have to trigger loading after retrieving from the cache, but i cannot do that based on the information in `ListedItem` (i think that's the case, maybe i'm wrong?). Whole point of the cache is to display something **before** the folder enumeration

gave92:
@jakoss no need to trigger the loading of the file thumbnail as it's already handled in `LoadExtendedPropertiesAsync`.
If you save a ListedItem to the cache setting it's 'IsItemPropertiesLoaded' property to false, this is enough to trigger the thumbnail loading.
Tha cache will still make the items appear much faster, only with no icon

jakoss:
@gave92 Ok, that would make it so easy, will try it out.

I think persisting images would make the cache bloated and will impact the deserialization process, and that would make the whole cache pointless. I think default icon might do, it will still make app appear faster. And, the most imporant part, i'm keeping memory cache before db. In memory cache all bitmaps are stored. So the icons loading would happen only on first cache load, then it will be stored in memory (with reasonable limits of course)

jakoss:
@gave92 @yair100 I pushed SQLite implementation of persistent cache adapter. It's working fine, but i'll do a experimental implementation based on DBreeze just to compare performance between those. SQLite is not really optimized for big payloads and random read performance, so i expect DBreeze to be significantly faster.

I'll post my findings and measurements here and we will have some ground for discussion if it's worth it to use additional dependency

Files/Helpers/FileListCache.cs
Jaiganeshkumaran: @duke7553 They are allowed

jakoss:
I run the numbers and found out that DBreeze is indeed faster, but both databases were really fast. Results from BenchmarkDotNet:
```
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042

Intel Core i5-9600K CPU 3.70GHz (Coffee Lake), 1 CPU, 6 logical and 6 physical cores

.NET Core SDK=5.0.101

  [Host] : .NET Core 5.0.1 (CoreCLR 5.0.120.57516, CoreFX 5.0.120.57516), X64 RyuJIT


|          Method |       Mean |    Error |   StdDev |   Gen 0 |   Gen 1 |   Gen 2 | Allocated |
|---------------- |-----------:|---------:|---------:|--------:|--------:|--------:|----------:|
|  TestSQLiteRead |   983.7 μs | 17.14 μs | 16.04 μs | 49.8047 | 49.8047 | 49.8047 | 274.84 KB |
| TestDBreezeRead | 1,316.8 μs | 12.30 μs | 11.51 μs | 48.8281 | 48.8281 | 48.8281 | 383.68 KB |


// * Legends *

  Mean      : Arithmetic mean of all measurements

  Error     : Half of 99.9% confidence interval

  StdDev    : Standard deviation of all measurements

  Gen 0     : GC Generation 0 collects per 1000 operations

  Gen 1     : GC Generation 1 collects per 1000 operations

  Gen 2     : GC Generation 2 collects per 1000 operations

  Allocated : Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)

  1 μs      : 1 Microsecond (0.000001 sec)
```

DBreeze has one more limitation, and that could be probably a major one - it only supports single process per database. SQLite do not have such restrictions. In this case i believe it's quite safe to stick with SQLite as it is now.

I'm posting this pull request as is for review. I'll add preemptive cache and custom cache configuration later on in another Pull Request

gave92:
@jakoss thanks a lot for your work. I too agree to stick with SQLite (also we avoid adding another dependency)

Files/Views/SettingsPages/Experimental.xaml
```diff
@@ -38,6 +38,12 @@
                     Header="Show Owner in properties"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{x:Bind AppSettings.ShowFileOwner, Mode=TwoWay}" />
+
+                <ToggleSwitch
+                    x:Uid="SettingsUseFileListCache"
+                    Header="Cache file list for performance boost (only memory cache for now)"
```
yaichenbaum, line 44:
```suggestion
                    Header="Cache files and folders for better performance"
```

Can you add this string to the en-US resource file?

Files/View Models/SettingsViewModel.cs
```diff
@@ -228,6 +228,12 @@ public bool ShowFileOwner
             set => Set(value);
         }
 
+        public bool UseFileListCache
+        {
+            get => Get(false);
+            set => Set(value);
+        }
```
yaichenbaum, line 235:
```suggestion
        /// <summary>
        /// Gets or sets a value indicating whether or not to cache files and folders.
        /// </summary>
        public bool UseFileListCache
        {
            get => Get(false);
            set => Set(value);
        }
```

yair100 (changes requested):

Files/Helpers/FileListCache/FileListCacheController.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        private readonly IMemoryCache filesCache = new MemoryCache(new MemoryCacheOptions
+        {
+            SizeLimit = 1_000_000
+        });
+
+        public Task SaveFileListToCache(string path, CacheEntry cacheEntry)
+        {
+            if (!App.AppSettings.UseFileListCache) return Task.CompletedTask;
```
yaichenbaum, line 28: @tsvietOK Should this be changedd to `ExperimentalViewModel`?

tsvietOK: `ExperimentalViewModel` is not available until it is launched.

yair100 (approved):

yaichenbaum:
@jakoss can you reproduce an issue where files randomly disappear and reload after around a second when the setting to use the cache is turned on?

jakoss:
@yair100 I encountered this twice, but i could not replicate that, so i could not check what's going on in debugger

yaichenbaum:
@jakoss Great work! LGTM

yaichenbaum:
@jakoss It looks like these changes are crashing the app in release mode.
image

jakoss:
StackTrace:
```
   at System.Reflection.IntrospectionExtensions.GetTypeInfo(Type type) in f:\dd\ndp\fxcore\CoreRT\src\System.Private.CoreLib\shared\System\Reflection\IntrospectionExtensions.cs:line 14
   at Microsoft.Data.Sqlite.Utilities.BundleInitializer.Initialize()
   at Microsoft.Data.Sqlite.SqliteConnection..cctor()
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext* pContext) in f:\dd\ndp\fxcore\CoreRT\src\System.Private.CoreLib\src\System\Runtime\CompilerServices\ClassConstructorRunner.cs:line 92
```
That's most certainly issue with SQLite binary file. Seems like architecture missmatch: https://stackoverflow.com/questions/37849668/value-cannot-be-null-r-nparameter-name-s

Have you tried if `GoogleDriveCloudProvider` works in release mode? Because there's exact same code, but with try/catch that swallow any exception. I'm pretty sure that won't work either.

PR #2673, opened 2020-12-23
Added a resize animation to the grid layout mode

yaichenbaum:


Files/Views/LayoutModes/GridViewBrowser.xaml.cs
```diff
@@ -411,5 +437,36 @@ private async void FileList_ItemClick(object sender, ItemClickEventArgs e)
                 ParentShellPageInstance.InteractionOperations.OpenItem_Click(null, null);
             }
         }
+
+        #region Animation
+
+        private CompositionAnimationGroup createOffsetAnimation()
```
duke7553, line 443:
```suggestion
        private CompositionAnimationGroup CreateOffsetAnimation()
```

duke7553 (dismissed):
Assuming this was validated for performance & memory consumption in large collections, I don't think it would break the item container virtualization of GridView. Though, we may need to test with users to make sure the duration is right. Animations should be quick and not over the top. It LGTM, but just something to keep in mind. We can always revert this PR if users report jank when resizing the window.

Files/Views/LayoutModes/GridViewBrowser.xaml.cs
yaichenbaum: This should be resolved wtih a4237e4c5e8e80d29e735bd911951973d54cf3e2.

duke7553 (approved):

Issue #2447, opened 2020-11-26
Feature: Add support for searching in recycle bin


yaichenbaum:

gave92:
Searching inside a hidden folder crashes the app with UnauthorizedAccessException
`2020-12-15 19:55:53.7967|ERROR|Files.App|Accesso negato. (Exception from HRESULT: 0x80070005 (E_ACCESSDENIED))|System.UnauthorizedAccessException: Accesso negato. (Exception from HRESULT: 0x80070005 (E_ACCESSDENIED))
   at Files.Filesystem.Search.FolderSearch.SearchForUserQueryTextAsync(String userText, String WorkingDirectory, Int32 maxItemCount)
   at Files.Views.Pages.ModernShellPage.ModernShellPage_SearchTextChanged(AutoSuggestBox sender, AutoSuggestBoxTextChangedEventArgs args)
   at System.Threading.WinRTSynchronizationContextBase.Invoker.InvokeCore()`

Hefaistos68:
Shouldnt Search show up in the "Ongoing Tasks" window?

Hefaistos68:
[x] show somewhere what the search was about (currently it shows only "search results in (location)" )
[x] support for patterns, regex, history

tech189:
Support for saving and opening already Saved Searches (.search-ms) files would be nice too. Pinning them to the sidebar would be awesome.

eraychumak:
When searching in e.g. `Downloads` folder and pressing enter, pressing on `Downloads` in the sidebar should remove the search results override and show `Downloads` items as normal instead of having to first click on e.g. `Desktop` and then back to `Downloads` to see everything again.

winston-de:
I noticed that you can't select search results from the drop down with arrow keys. Attempting to use the down-arrow key will open the first item. 
I'd like to add that to the list.

PR #2679, opened 2020-12-23
Enabled search inside hidden directories

gave92:
Fir #2447

This PR:
- handles errors that can be thrown by StorageFolder.GetFolderFromPathAsync()
- allows to search inside hidden folders (shallow search only)
- fix a bug for which returned hidden items were not filtered by search term
- fix a bug for which results were ordered by file name instead of search rank
- return at most *maxItemsCount` results
- small code refactor


yaichenbaum:
> fix a bug for which results are ordered by file name instead of search rank

@gave92 This was done intentionally so that the most relevant items would show on the quick search drop down. Is there a way to have the drop-down sort by name and have the search results page sort by rank?

gave92:
@yair100 mmm I'm not getting the most relevant results atm.

Example search:
I'm looking for files named "test". If in the current path I have a folder also named "test" in the dropdown I get all the files contained inside the "test" folder. Instead I'd like to see files matching "test" first.

Doesn't sorting by rank guarantee that I get the most relevant results?
If they are sorted by filename I get a lot of uninteresting stuff instead

yair100 (approved):

yaichenbaum:
@gave92 Great work!

PR #2681, opened 2020-12-24
Added an option to search unindexed items

yaichenbaum:
For #2447

### DIFF
--- a/Files/Filesystem/Search/FolderSearch.cs
+++ b/Files/Filesystem/Search/FolderSearch.cs
@@ -118,9 +118,18 @@ namespace Files.Filesystem.Search
             QueryOptions options = new QueryOptions()
             {
                 FolderDepth = FolderDepth.Deep,
-                IndexerOption = IndexerOption.OnlyUseIndexerAndOptimizeForIndexedProperties,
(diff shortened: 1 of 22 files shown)


gave92:
Should we use `UseIndexerWhenAvailable` instead of `DoNotUseIndexer`?

Searching without the indexer is obviously veeery slow, we should probably (not necessarili in this PR):
1- Always use the indexer for drop down results regardless of this new setting
2- Start showing results as soon as they are available
3- Offer a way to cancel search when in progress like explorer does

yaichenbaum:
> Should we use `UseIndexerWhenAvailable` instead of `DoNotUseIndexer`?

This didn't seem to make a difference with regards to showing items that were not yet indexed.

> Searching without the indexer is obviously veeery slow, we should probably (not necessarily in this PR):
> 1- Always use the indexer for drop down results regardless of this new setting

This setting is meant to bypass the indexer, the user is clearly informed beforehand that it will be slower. I am not sure the drop down should have different behavior.

> 2- Start showing results as soon as they are available (e.g every 10 results or so)
> 3- Offer a way to cancel search when in progress like explorer does

Yes to both of these!

gave92 (approved):

PR #2690, opened 2020-12-24
Show dialog when renaming a file and the destination already exists

### LINKED ISSUE #2689: No dialog is shown when renaming an item and a files with the same name already exists
**Describe the bug**
When renaming an item, if a file with the same name exists in the folder the rename operation silently fails. The item is renamed in the list with the new name but reverts back when refreshing the folder.

**To Reproduce**
1. Navigate to a folder that contains files "file1.txt" and "file2.txt"
2. Rename file2.txt to file1.txt
3. file2.txt is renamed in the file list but the name reverts back when refreshing the folder

**Expected behavior**
App should show a popup and let user decide whether to overwrite the existing file or select a different name

 - OS Version: [Windows 10 1903 18362.175]
 - App version: [v0.23.1]

gave92:
Fixes #2689

This PR:
- Fixes pasting files to hidden folders
- Fix rename dialog not showing when an item with the same name already exists
- Fix file list not updating in recycle bin when new files are deleted
- Fix a case in which the "empty folder" text was visible after pasting files into a folder with a redo operation

### DIFF
--- a/Files/Filesystem/FilesystemOperations/FilesystemOperations.cs
+++ b/Files/Filesystem/FilesystemOperations/FilesystemOperations.cs
@@ -210,39 +210,33 @@ namespace Files.Filesystem
                     progress?.Report((float)(itemSize * 100.0f / itemSize));
                 }
 
-                FilesystemResult<StorageFolder> fsResult = await associatedInstance.FilesystemViewModel.GetFolderFromPathAsync(Path.GetDirectoryName(destination));
+                FilesystemResult<StorageFolder> destinationResult = await associatedInstance.FilesystemViewModel.GetFolderFromPathAsync(Path.GetDirectoryName(destination));
+                var sourceResult = await source.ToStorageItemResult(associatedInstance);
+                var fsResult = (FilesystemResult)(sourceResult.ErrorCode | destinationResult.ErrorCode);
 
                 if (fsResult)
                 {
-                    StorageFile file = (StorageFile)await source.ToStorageItem(associatedInstance);
-
-                    FilesystemResult<StorageFile> fsResultCopy = new FilesystemResult<StorageFile>(null, FilesystemErrorCode.ERROR_GENERIC);
-
-                    if (file != null)
-                    {
-                        fsResultCopy = await FilesystemTasks.Wrap(() =>
-                        {
-                            return file.CopyAsync(fsResult.Result, Path.GetFileName(file.Name), NameCollisionOption.GenerateUniqueName).AsTask();
-                        });
-                    }
-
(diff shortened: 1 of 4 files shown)


yaichenbaum:
@gave92 Is it posible to use the the availible buttons in the status center instead of the dialog?

gave92:
@yair100 should I have buttons in the status center? could you share a screenshot?

yair100:
> @yair100 should I have buttons in the status center? could you share a screenshot?

They are hidden by default, but you can display them and choose an action when updating a banner.

gave92:
Aha found them. I wasn't aware of the possibility, I don't think they are used anywhere atm, right?

If it's acceptable I'd keep this PR as is and tackle the possibility to use the status banner instead of the dialogs another time

yaichenbaum:
> Aha found them. I wasn't aware of the possibility, I don't think they are used anywhere atm, right?

There is one error that will cause them to show but it's not common.

tsvietOK:
@gave92 Should it work when creating a new file with the same name?

gave92:
@tsvietOK no just in case of renaming for now. Creating a new file with the same name should result in "File (2)" at the moment

yaichenbaum:
> If it's acceptable I'd keep this PR as is and tackle the possibility to use the status banner instead of the dialogs for rename and the other operations another time.

I got confused with copy/paste with the same file name, a dialog is the proper solution for solving rename conflicts.

tsvietOK:
@yair100 What about to add button like "Cancel", "Do nothing" or "Skip" to the dialog?

yaichenbaum:
> @yaichenbaum What about to add button like "Cancel", "Do nothing" or "Skip" to the dialog?

It looks like there is already one for skip and cancel, a "Do nothing" option would be helpful.

tsvietOK:
@yair100 There is no "Cancel" option
image

gave92:
Yeah, cancel currently can only be done by pressing Escape key

tsvietOK:
> Yeah, cancel currently can only be done by pressing Escape key (possibly not intuitive for every user)

By the way, it is not always working.

gave92:
@tsvietOK Meaning that sometimes the dialog does not close upon pressing Escape? Or that is closes but some operation is done anyway?

tsvietOK:
@gave92 dialog does not close when pressing Escape/

gave92:
@tsvietOK does it seem something specific to the rename dialog? Or have you seen this behavior on other dialogs?

tsvietOK:
@gave92 That's strange, now it is working.

tsvietOK (approved):

PR #2698, opened 2020-12-25
Shortcut: Fixed selected range and renaming

tsvietOK:
- Fixed an issue where when renaming a shortcut, its extension is cleared.
- Fixed selected name range when trying to rename shortcut.
Before:
image
After:
image


yaichenbaum (approved):

yaichenbaum:
:shipit:

PR #2712, opened 2020-12-26
Show a dialog to resolve conflicts when transferring an item and the destination already exits

### LINKED ISSUE #1893: Support overwrite experience during paste operation
When you're pasting a file to a folder that already has a file with the same name, there is no choice to overwrite it - the app always creates a copy.

Offer a dialog with the choice of either overwriting or creating a copy.

**Additional context**
App version: v0.14.0

gave92:
Fixes #1893

This PR adds a dialog (identical to the one already existing for the rename operation) when moving/copying an item and the destination already exists.
Dialog options: Generate new name, Overwrite, Do nothing (escape key)
This is for both files and folders, there is no "merge folders" option

In addition:
- enables the dialog shown when pasting a folder inside itself
- fixes an issue for which a banner was left in the status center after canceling a delete operation
- fixes an issue for which the app crashed when pressing delete key -> [2010984911u](https://appcenter.ms/orgs/Files-UWP/apps/Files/crashes/errors/2010984911u/overview)


Files/App.xaml.cs
```diff
@@ -39,10 +39,7 @@ sealed partial class App : Application
     {
         private static bool ShowErrorNotification = false;
 
-        private readonly static CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
-        public static CancellationToken CancellationToken = cancellationTokenSource.Token;
```
d2dyno1, line 43: :(

gave92:
> :(
I simply moved it to ModernShellPage :)
It's not used atm, right? In the future we could use it to stop a particular operation, but would not make sense for it to be global i guess

tsvietOK:
@gave92 Maybe i am doing something wrong, but when i copying file or folder and pasting in the same place, it shows dialog with 2 buttons instead of 3.

gave92:
@tsvietOK that's expected, currently the only way to cancel without doing anything is to press the escape key
You're right we need to add the "CloseButtonText" to both this and the rename dialog

tsvietOK:
@gave92 Oh, ok. Sorry, my fault. Would you want to add this button in this or another PR?

gave92:
@yair100 do you think I should add a third button to the rename and copy dialogs to cancel the operation?
Or is pressing the escape key to cancel sufficient?

yair100:
> @yair100 do you think I should add a third button to the rename and copy dialogs to cancel the operation?
> Or is pressing the escape key to cancel sufficient?

A cancel button is especially important for tablet users so we should probably add one.

gave92:
Cancel button added to dialog 👍

yair100 (approved):

PR #2713, opened 2020-12-26
Fixed layout selection when navigating using the address bar

gave92:
Fixes an issue for which the wrong layout mode is selected when navigating using the address bar (clicking on the path bar components)


yaichenbaum (approved):

yaichenbaum:
:shipit:

Issue #2137, opened 2020-10-24
Selection rectangle: Don't clear selection if ctrl is pressed


Thomas1664:
**Describe the bug**
When starting selection rectangle and holding CTRL, the current selected item gets deselected.

**To Reproduce**
1. Select an item
2. Hold `CTRL`
3. Trigger selection rectangle

**Expected behavior**
Don't clear selection

**Desktop (please complete the following information):**
 - OS Version: [Windows 10 2009]
 - App version: [master]

yaichenbaum:
@Thomas1664 This issue should be resolved with #2716.

PR #2716, opened 2020-12-26
Fixed an issue where holding the ctrl key when using the selection rectangle would clear the current selection

Don-Vito:
A fix for https://github.com/files-community/Files/issues/2137.
In addition to the above, preserves existing selection 
(when ctrl is pressed) if clicking at no item (similar to explorer).

Please be gentle - this is my first commit in this repo :blush:

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -67,7 +67,8 @@ namespace Files.Interacts
 
         public void List_ItemDoubleClick(object sender, DoubleTappedRoutedEventArgs e)
         {
-            if (!AppSettings.OpenItemsWithOneclick)
+            // Skip opening selected items if the double tap doesn't capture an item
+            if ((e.OriginalSource as FrameworkElement)?.DataContext is ListedItem && !AppSettings.OpenItemsWithOneclick)
             {
                 OpenSelectedItems(false);
             }
--- a/Files/UserControls/Selection/RectangleSelection_DataGrid.cs
+++ b/Files/UserControls/Selection/RectangleSelection_DataGrid.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Windows.Foundation;
+using Windows.System;
 using Windows.UI.Xaml;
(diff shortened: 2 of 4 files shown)


gave92:
Hi @Don-Vito thanks a lot for this contribution! This changes look good to me.
One small issue is that with
> preserving existing selection (when ctrl is pressed) if clicking at no item (similar to explorer)

is that is the user selects a bunch of files and then double clicks outside *with control pressed*, all selected files get opened.
This may or may not be an acceptable behaviour.. @yair100? @tsvietOK?

yaichenbaum:
@gave92 The issue with double clicks opening all selected files should probably be fixed before this is merged.

Don-Vito:
@gave92, @yair100  - thanks for the feedback! On it.

Don-Vito:
@gave92, @yair100 - not sure how to re-request review 😊 

BTW, just wondering if you considered GitHub review process:
* instead of adding "changes requested" label, you could finish the review with "changes requested" status
* the contributor could re-request the review, once fixed

Don-Vito:
@gave92 - I also have a PoC for inverting the selection of previously selected items (hope to get there later on) - please let me know if you want to handle this in the scope of this PR or separately.

gave92:
>I also have a PoC for inverting the selection of previously selected items

Meaning that dragging the rectangle on previously selected items while pressing ctrl will deselect them like in explorer?
If yes, I see no problem including it here as it's related, but I'd wait on a confirmation by @yair100 that it's a feature we want (I personally don't mind the current behaviour of only adding to selection, I don't think I ever used that explorer feature)
>instead of adding "changes requested" label, you could finish the review with "changes requested" status

We usually do both :)

yaichenbaum:
> @gave92 - I also have a PoC for inverting the selection of previously selected items (hope to get there later on) - please let me know if you want to handle this in the scope of this PR or separately.

That's an interesting feature, it's probably better to split it into another pull request.

yair100 (approved):

yaichenbaum:
@Don-Vito Great work and congrats on your first pull request 🚀

Issue #1674, opened 2020-08-07
Dual-pane view


compl:
Guys,

It would be awesome to have dual-pane view as in good old Total Commander

Thnx

![image](https://user-images.githubusercontent.com/9084162/89675656-f5751e00-d8f2-11ea-8600-96d59052058a.png)

gabrieldelaparra:
Would it be possible that a pane works as in visual studio? You have a tab and you can drag it to split the grid view (horizontally or vertically). This is such a cool feature of VSCode, but I am not sure if this can be done here (?)

compl:
Indeed, as in VSCode would be perfect.

And I didn't mean the constant dual-pane view, there should be a choice, of course, as in VSCode again.

Starbuck1991:
Dual pane is a much needed feature. Please, make it so!

yaichenbaum:
This feature was implemented with #2727.

PR #2500, opened 2020-12-01
ColumnView

TyJOrtiz:
Added a columnViewBrowser.

Very barebones, only the "Layout Mode" Context menu item works.

PR #2727, opened 2020-12-27
Implemented a dual pane feature

gave92:
Implements #1674

image

The changes to remove the App.CurrentInstance (thanks @duke7553!) allow to implement this with minimum changes.

What's missing:
- [x] The second pane looks like it's not the same width as the first pane
- [x] The focus indicator should only show up if more than one pane is open
- [x] The "New Pane" menu item should be hidden when two panes are already open
- [x] The shadow between the panes should be removed in favor of a thin border
- [x] The keyboard shortcut to open a new pane should be <kbd>Alt</kbd>+<kbd>shift</kbd>+<kbd>+</kbd>
- [x] The keyboard shortcut to close a pane should be <kbd>Ctrl</kbd>+<kbd>shift</kbd>+<kbd>w</kbd>
- [x] The vertical flyout and the status center should be shown only in the right pane
- [x] The tab title reflects the working directory of the pane that was last navigated (should the title be like "Documents | Downloads" instead?)
- [x] The resume where you left off option reopens only one pane with the working directory of the pane that was last navigated before closing
- [x] Maybe an option to always open the right pane when opening a new tab
- [x] An option to open a folder or a sidebar item in the right pane
- [x] An button to open the right pane from the navigation bar menu
- [x] An option to disable this feature


yaichenbaum:
@gave92 Nice work! How does the dual pane work with the sidebar?

gave92:
@yair100 the sidebar always controls the left pane. Seems better since the sidebar is visually connected to the left pane only. What do you think? At most I'd add a "Open in the right pane" context menu option to the sidebar items

yair100:
> @yair100 the sidebar always controls the left pane. Seems better since the sidebar is visually connected to the left pane only. What do you think? At most I'd add a "Open in the right pane" context menu option to the sidebar items

An "Open in new pane" option would be helpful. However, being that this feature is for power users and many users would prefer to hide these options, I would recomend adding a setting to enable/disable the dual pane feature.

d2dyno1:
WOOOOOOOOOOOOOOOOOOO
POGCHAMP

:shipit: :shipit: :shipit:

Files/MultilingualResources/Files.hi-IN.xlf
```diff
@@ -1948,6 +1948,22 @@
           <source>Enable individual preferences for individual directories</source>
           <target state="new">Enable individual preferences for individual directories</target>
         </trans-unit>
+        <trans-unit id="AddPaneTeachingTip.Subtitle" translate="yes" xml:space="preserve">
+          <source>Click to open the right pane</source>
+          <target state="new">Click to open the right pane</target>
+        </trans-unit>
+        <trans-unit id="AddPaneTeachingTip.Title" translate="yes" xml:space="preserve">
```
yaichenbaum, line 1955: I don't think we need this teaching tip since the feature will be for powerusers and most users won't find it very helpful.

d2dyno1:
> The tab title reflects the working directory of the pane that was last navigated (should the ta title be like "Documents | Downloads" instead?)

Maybe we indicate to user (somehow) the selected pane and update path for it when it's focused.

Files/MultilingualResources/Files.hi-IN.xlf
d2dyno1: I don't consider dual pane view as poweruser feature

d2dyno1: I suggest we add more teaching tips AROUND the dual pane view

Files/Views/SettingsPages/Multitasking.xaml
```diff
@@ -76,6 +76,18 @@
...
+                              Header="Enable dual pane view"
+                              HeaderTemplate="{StaticResource CustomHeaderStyle}"
+                              IsOn="{x:Bind AppSettings.IsDualPaneEnabled, Mode=TwoWay}" />
+
+                <ToggleSwitch x:Name="SettingsMultitaskingAlwaysOpenDualPane"
+                              x:Uid="SettingsMultitaskingAlwaysOpenDualPane"
+                              Header="Always open new tabs in dual pane mode"
+                              HeaderTemplate="{StaticResource CustomHeaderStyle}"
+                              x:Load="{x:Bind AppSettings.IsDualPaneEnabled, Mode=OneWay}"
+                              IsOn="{x:Bind AppSettings.AlwaysOpenDualPaneInNewTab, Mode=TwoWay}" />
```
yaichenbaum, line 90: Instead of reworking the open in new tab menu item, should there be a new menu item to open items in a new pane?

Files/MultilingualResources/Files.hi-IN.xlf
yaichenbaum: @d2dyno1 Perhaps that's because you are a poweruser 😄
The design of an app should be usable by the majority of users without having to explain how everything works. If you find the need to explain how a feature works, then most of the time, you need to rethink the feature. Teaching tips are meant to be used when there is no other way for a user to learn how to use a feature however, sometimes, some features need a guide or more information and that is where the wiki can fill the gap.

gave92: I added so the user will know he/she has to hover near the right edge of the window to popup the "add new pane" button. Otherwise it may not be obvious

yaichenbaum: Adding on to my comment about the `AlwaysOpenDualPaneInNewTab` setting, I think that it won't be as necessary to inform the user of this method since there will be a right click option as well.

Files/Views/SettingsPages/Multitasking.xaml
gave92: Not sure if I got what you mean here, the above setting is for automatically opening both panes when opening a new tab
I'll for sure add an option to the right click to open in the right pane

yaichenbaum: I misunderstood what this setting does 😂

Files/Views/PaneHolderPage.xaml.cs
```diff
@@ -0,0 +1,256 @@
...
+using System;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+using Windows.ApplicationModel.AppService;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Input;
+using Windows.UI.Xaml.Navigation;
+
+// Il modello di elemento Pagina vuota è documentato all'indirizzo https://go.microsoft.com/fwlink/?LinkId=234238
```
d2dyno1, line 15: Italian?

gave92: Aha uops, default comment

Jaiganeshkumaran:
@gave92 Instead of two panes, could it support variable size panes?

gave92:
@Jaiganeshkumaran yeah it could support any number of panes (the panes are already resizable).
I don't know if I'll manage to include that in this PR, also it would overlap a lot with ColumnView #2500

d2dyno1:
Do we have any progress report on #2500?

duke7553:
@gave92 This is great to see! Looking forward to finally addressing user feedback with this PR.

d2dyno1:
bruh

yaichenbaum:
@gave92 This feature looks great! Here are a couple of things I noticed so far
- The second pane looks like it's not the same width as the first pane
- The focus indicator should only show up if more than one pane is open
- The "New Pane" menu item should be hidden when two panes are already open
- The shadow between the panes should be removed in favor of a thin border

mdtauk:
Not entirely comfortable with there being two panes for a single tab, with two totally unrelated locations.  Do you need two tab bars, or does the tab UI change to indicate two panes side by side?  Could the second pane be a flyout on the right that can be brought into view temporarily for file transfers or comparisons, then dismissed?  So a transient UI and not a full IconView in the same way as the tab would navigate to?

There is also a question to be asked with future support for dual-screened devices.  Assuming Files plans to be ready for 10X, how will the dual screen form factor be supported?  This feature could complicate that.

yaichenbaum:
> Not entirely comfortable with there being two panes for a single tab, with two totally unrelated locations. Do you need two tab bars, or does the tab UI change to indicate two panes side by side? Could the second pane be a flyout on the right that can be brought into view temporarily for file transfers or comparisons, then dismissed? So a transient UI and not a full IconView in the same way as the tab would navigate to?

@mdtauk This dual pane feature is commonly found in other file managers and its targeted towards power users who want to work with more than one directory in the same tab. You can think of it as split screen, but the difference is that it's inside the tab and you can even restore your session when restarting the app. A transient UI is an interesting idea, but the dual pane view is something that users have been asking for us to add for quite some time and the functionality is different.

> There is also a question to be asked with future support for dual-screened devices. Assuming Files plans to be ready for 10X, how will the dual screen form factor be supported? This feature could complicate that.

As of now Files does not run on Windows 10X but if we can get it working in the future, the dual pane feature will give the user more choices in how to use the app.

mdtauk:
> @mdtauk This dual pane feature is commonly found in other file managers and its targeted towards power users who want to work with more than one directory in the same tab. You can think of it as split screen, but the difference is that it's inside the tab and you can even restore your session when restarting the app. A transient UI is an interesting idea, but the dual pane view is something that users have been asking for us to add for quite some time and the functionality is different.
> 
So I think what people have wanted, is the ability to drag a tab to the side of the window, and dock it as a split view.  That is a bit different to how you are implementing the feature.

I think an option to have two tab views, split into two columns, would solve the problem.  Or visually distinguishing the second pane as a flyout (without light dismiss behaviour, but the ability to close it) so there is no confusion for navigation or tab contents.

Having split tab bars, would also solve the problem with supporting Dual Screens.

Then the only thing you may need to implement is a toggle for the Sidebar, so you can close it to give you more space on both screens.

yaichenbaum:
@mdtauk That use case is already covered by the built-in split screen in Windows. I think that both of them are valid use cases but as of now, we have only gotten feedback regarding the use case being implemented here.

mdtauk:
> @mdtauk That use case is already covered by the built-in split screen in Windows. I think that both of them are valid use cases but as of now, we have only gotten feedback regarding the use case being implemented here.

Except that is not two windows side by side, but a single window split over both screens.

gave92:
@mdtauk is your idea close to what you get if you open two instances of the app and stick them to the screen edges? You get two independent tab bars and you can drag tabs from one to the other.

image

mdtauk:
> @mdtauk is your idea close to what you get if you open two instances of the app and stick them to the screen edges? You get two independent tab bars and you can drag tabs from one to the other.
> 
> image

That should already be possible, I think I am talking more about the best way to have a split pane layout for the app.

The idea as demonstrated in the screenshot, is a single window, but the two icon views are under one tab.

I think this could be confusing to users, as two views _should_ be two tabs.  But the purpose is for dragging between views, or visually comparing right?  So rather than present the second icon view in the same way, if it were in a right hand sidebar/flyout - it would look distinctive enough to not need a second tab.

But if you are thinking ahead to Windows 10X and Dual Screen form factors, then two tab bars beside each other would work better.

image

gave92:
> @gave92 This feature looks great! Here are a couple of things I noticed so far
> 
> * The second pane looks like it's not the same width as the first pane
> * ...

@yair100 everything in the list is now implemented 👍 let's start with the next list xD

yair100 (approved):

PR #2728, opened 2020-12-27
Teach rectangle selection to invert previous selection 

Don-Vito:
Stacked upon https://github.com/files-community/Files/pull/2716.
(that is why the change looks bigger than it is :blush:)

In the previous PR when CTRL is clicked, rectangle selection preserves
previous selection.

Here we extend the logic to invert selection on items that were
previously selected, like in Windows Explorer:

InvertSelection


Don-Vito:
Resolving conflicts.

Files/UserControls/Selection/ISelectedItems.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Files.UserControls.Selection
+{
+    public interface ISelectedItems
```
gave92, line 3: Is this interface necessary? It's identical to `IList`

Files/UserControls/Selection/GenericSelectedItems.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+
+namespace Files.UserControls.Selection
+{
+    public class GenericSelectedItems : ISelectedItems
```
gave92, line 5: Same comment as ISelectedItems

Files/UserControls/Selection/RawSelectedItems.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Collections;
+
+namespace Files.UserControls.Selection
+{
+    public class RawSelectedItems : ISelectedItems
```
gave92, line 5: What's the difference with GenericSelectedItems?

Files/UserControls/Selection/ISelectedItems.cs
Don-Vito: @gave92 - I know it looks crazy :smile:

I didn't touch C# for a while, but I wanted to apply the same logic to both `Collections.IList` and `Collections.Generics.IList<object>`.

The reason for this is that I wanted to reuse the same strategy for both list and grid, but their respective Selected items have different types.

Files/UserControls/Selection/GenericSelectedItems.cs
Don-Vito: If there is a better solution from the problem above I will happily get rid of this :blush:

Files/UserControls/Selection/RawSelectedItems.cs
Don-Vito: and of this :blush:

Files/UserControls/Selection/ISelectedItems.cs
gave92: Suggestion: `uiElement.SelectedItems.Cast<object>().ToList()` will convert an `IList` to an `IList<object>`
The reverse should be possible as well

Don-Vito: @gave92  - I cannot do it as I need a side-effect on the original collection

gave92: And the reverse? var generic = (System.Collections.IList)uiElement.SelectedItems;

Don-Vito: Crashes the COM object :smile:

gave92:
@Don-Vito you should find a PR on your repo with an slightly different approach, let me know if you agree with it.
Also great job with this, everything else seems to be working good 👍

Don-Vito:
> 
> 
> @Don-Vito you should find a PR on your repo with an slightly different approach, let me know if you agree with it.
> Also great job with this, everything else seems to be working good 👍

@gave92  - Yeah, considered this as well, but was concerned with implementing a very yide interface.
(thus invented one of my own).

WDYT of implementing ICollection rather than IList - it should be sufficient for the strategy and it somewhat smaller
(I can do it not to waste your time)? 

Otherwise let's merge yours :blush:

gave92:
Yup that's even better, if you can do that, then we can merge this. Thank you

gave92 (approved):

Don-Vito:
@gave92 - sorry for the delay.. unrelated to this PR noticed that shift selection sometimes clashes with ctrl selection (when you move and click very fast). I will open a separate PR for fixing this.

Issue #2341, opened 2020-11-15
MVVM


d2dyno1:
**Is your feature request related to a problem? Please describe.**
WPF and UWP technologies were intentionally designed to work with MVVM pattern. Files follows it very poorly as majority of the code is hidden in the code-behind files (MainPage.xaml.cs)

**Describe the solution you'd like**
A loosely coupled code achieved with MVVM would be more maintainable and would make the entire codebase more professional. The code would also be more efficient and more flexible. Less boilerplate, copied-over code. Not only that, project with implemented MVVM, if designed well, could work on multiple platforms.
Implementing MVVM would mean a huge rewrite of Core Files functionalities and would take a huge amount of time, but in the end I believe the change would be worth it. 

(This would mean we'd had to delete the code-behind (e.g. MainPage.xaml.cs) and move it over to ViewModel class (e.g. MainPageViewModel.cs))

**Additional context**
I always code my applications following MVVM pattern and I don't regret doing so.

yaichenbaum:
@d2dyno1 Moving towards MVVM is a long-term goal and we have made several steps in this direction with some of the recent refactoring we have done. Is this something you are interested in helping with?

d2dyno1:
I'll make sure my next PRs follow the MVVM pattern

yaichenbaum:
@d2dyno1 That sounds great, one of the things we can already start doing is moving some of the core functions out of the code-behind into a separate "commands" folder for features such as copying, pasting etc.

d2dyno1:
@yair100  Are you referring to new changes in #2358? In the means of copying, moving, deleting items, the (new, reworked) `IFilesystemOperations` needed its own special place. I intentionally didn't keep it in the Commands directory because `Commands` are specifically reserved for UI elements functionality implementation (`ICommand`). The new `IFilesystemOperations` is a part of Files' underlying FS.

yaichenbaum:
@d2dyno1 No, I was just saying in general.

duke7553:
@d2dyno1 We do need to rename the "Commands" folder to something less confusing.

d2dyno1:
I'd create a ViewModel folder for all View Models. I'll open a PR regarding this - How I'd do it, how I see it done - as soon as #2358 is merged, and #1763, #2117 is completed

yaichenbaum:
@d2dyno1 We already have a ViewModel folder.

d2dyno1:
👀

yaichenbaum:
[...]

PR #2733, opened 2020-12-27
Settings: Use MVVM pattern

tsvietOK:
For #2341 
This also improves the UI on the "On Startup" page.
Before:
image
After:
image


yaichenbaum (approved):

d2dyno1:
Nice nice :shipit:

Files/ViewModels/SettingsViewModels/FilesAndFoldersViewModel.cs
```diff
@@ -0,0 +1,135 @@
...
+        private bool searchUnindexedItems = App.AppSettings.SearchUnindexedItems;
+
+        public bool AreHiddenItemsVisible
+        {
+            get
+            {
+                return areHiddenItemsVisible;
+            }
+            set
+            {
```
d2dyno1, line 21: Since this class inherits from `ObservableObject` we can drop the check and `OnPropertyChanged()` and replace it with `SetProperty()`

Files/Views/SettingsPages/About.xaml
```diff
@@ -43,126 +84,30 @@
...
+                                Command="{Binding ClickAboutFeedbackItemCommand}" />
+                        </ic:EventTriggerBehavior>
+                    </i:Interaction.Behaviors>
                 </ListView>
 
                 <Button
                     x:Name="OpenLogLocationButton"
                     x:Uid="SettingsAboutOpenLogLocationButton"
-                    Click="OpenLogLocationButton_Click"
+                    Command="{Binding OpenLogLocationCommand}"
```
d2dyno1, line 110: `x:Bind` ?

Files/ViewModels/SettingsViewModels/FilesAndFoldersViewModel.cs
tsvietOK: As i understand, only Binding is working with DataContext.
image

d2dyno1: Yeah, sorry my bad, though it could be solved with `x:Bind` 😅.

tsvietOK: Like this?
```csharp
if (SetProperty(ref areHiddenItemsVisible, value))
{
    App.AppSettings.AreHiddenItemsVisible = areHiddenItemsVisible;
}

```

d2dyno1: Yes :100:

d2dyno1: > `App.AppSettings.AreHiddenItemsVisible = areHiddenItemsVisible;`

`App.AppSettings.AreHiddenItemsVisible = value;`
To make it a bit clearer

tsvietOK: Thank you!

gave92:
@tsvietOK I like this changes a lot :)
Curious: why have you resorted to classing Bindings instead of x:Bind? Was it just to be able to declare the view model in Xaml and not in the code behind? Or where there other reasons?
(It's not an issue, it's not like we need the performance gains of x:Bind in the settings pages)

d2dyno1:
As @tsvietOK explained to me, with Commands, `x:Bind` does not work. Only `Binding` is allowed.

@gave92

gave92:
> As @tsvietOK explained to me, with Commands, `x:Bind` does not work. Only `Binding` is allowed.

? `Command="{x:Bind NewTabInvokedCommand, Mode=OneWay}"` --> one of the many examples in the code

d2dyno1:
:eyes:

d2dyno1:
Maybe that's because the command is in code-behind (?)

yaichenbaum:
> Curious: why have you resorted to classing Bindings instead of x:Bind? Was it just to be able to declare the view model in Xaml and not in the code behind?

That was the reason here, for some reason `x:Bind` isn't working with `DataContext`.

gave92:
Yeah, the alternative would have been to declare the viewmodel in the codebehind.
That said it doesn't make much difference here (with the exception of missing compile time checking and allowing visual studio to autocomplete the binding properties)

tsvietOK:
"Because {x:Bind} uses generated code to achieve its benefits, it requires type information at compile time. This means that you cannot bind to properties where you do not know the type ahead of time. Because of this, you cannot use {x:Bind} with the DataContext property, which is of type Object, and is also subject to change at run time."

gave92:
> Yeah, the alternative would have been to declare the viewmodel in the codebehind.
Would this be so bad? :)

gave92 (approved):

PR #2734, opened 2020-12-27
Teach rectangle selection to extend existing selection when the Shift key is pressed

Don-Vito:
This PR contains several modifications:
1. Shift + Select Rectangle - will extend the existing selection as in Windows Explorer
(unlike Ctrl + Select Rectangle that inverts the existing selection).
If both Shift and Ctrl are selected, Control wins (like in Windows Explorer)
2. The decision on the strategy is done upon mouse click:
you can release Shift, Ctrl afterwards
3. The logic of Clearing Selection was removed from GridViewBrowserViewer,
so the selection strategy can make the relevant decision

### DIFF
--- a/Files/Files.csproj
+++ b/Files/Files.csproj
@@ -187,6 +187,7 @@
     <Compile Include="UserControls\Selection\ExtendPreviousItemSelectionStrategy.cs" />
     <Compile Include="Helpers\GenericItemsCollection.cs" />
     <Compile Include="UserControls\Selection\IgnorePreviousItemSelectionStrategy.cs" />
+    <Compile Include="UserControls\Selection\InvertPreviousItemSelectionStrategy.cs" />
     <Compile Include="UserControls\Selection\ItemSelectionStrategy.cs" />
     <Compile Include="View Models\FolderSettingsViewModel.cs" />
     <Compile Include="View Models\Properties\FileProperty.cs" />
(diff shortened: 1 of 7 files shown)


XPoppyX:
Rectangle selection works for you in grid view?

Don-Vito:
> 
> 
> Rectangle selection works for you in grid view?

@XPoppyX - works randomly 😄  Actually I am exactly trying to understand why. 

There is some logic that looks for a scrollbar UIElement upon initialization of Rectangle Selection. If the scrollbar is not found the selection is disabled. Right now I am trying to understand:
* Why do we need this fabulous check
* Why is it failing randomly

gave92:
> Why do we need this fabulous check

It's needed to scroll the list when the rectangle is dragged near the top or bottom of the page

Don-Vito:
> 
> 
> > Why do we need this fabulous check
> 
> It's needed to scroll the list when the rectangle is dragged near the top or bottom of the page

Absolutely!  Thanks!

> > Why is it failing randomly
> 
> This is a very good question xD

Addressed this here: https://github.com/files-community/Files/pull/2736.
The reason is the race where the uiElement loaded is triggered before the children getting loaded.

gave92:
@Don-Vito it looks like pressing both shift and control will reset the selection every time

Don-Vito:
> 
> 
> @Don-Vito it looks like pressing both shift and control will reset the selection every time
> This check: `e.KeyModifiers == VirtualKeyModifiers.Control` will be false if both control and shift are pressed I think
> Something like `e.KeyModifiers.HasFlag(...)` may work

Absolutely. Thanks!

gave92 (approved):

yaichenbaum:
@Don-Vito Great work!

PR #2736, opened 2020-12-28
Make ScrollViewer lookup in Rectangle Selection robust

Don-Vito:
There is a race in RectangleSelection_ListViewBase where we try to
find ScrollViewer in the VisualTree upon uiElement.Loaded.

Unfortunately, Loaded might be handled before all the children
of uiElement are loaded, and in this case ScrollViewer is still not found.

When this happens we actually don't have rectangle selection available.

The idea in this PR is to register for LayoutUpdated if scroll viewer is
not found upon Loaded (and to unregister once found).


tsvietOK (approved):

yaichenbaum (approved):

yaichenbaum:
@Don-Vito Great work!

Issue #1546, opened 2020-07-23
Cut-pasting item in the same folder creates a copy of the item


soumyamahunt:
**Describe the bug**
Cut-pasting an item from the same folder replaces the item with a copy of it and attaches "-Copy" to its original name.

**To Reproduce**
1. Open files-uwp app.
2. Cut an item from a folder
3. Try pasting the item in the same folder.

**Expected behavior**
A dialog should be prompted with option to either skip or cancel the operation.

**Screenshots**
![EwoK2txTc6](https://user-images.githubusercontent.com/51203900/88255564-4e7b6a00-ccd6-11ea-9066-d0fd311eef98.gif)

**Desktop (please complete the following information):**
 - OS Version: Windows 10 1903 18362.239
 - App version: v0.12.0

**Additional context**
Add any other context about the problem here. Does this problem occur again after restarting the app?
Although it doesn't crash the app, the problem persists across restarts.

yaichenbaum:
@soumyamahunt Would you like to work on this issue?

yaichenbaum:
@soumyamahunt This issue should be resolved with #2742.

Issue #1545, opened 2020-07-23
Cut-pasting a parent folder to a child folder crashes the app


soumyamahunt:
**Describe the bug**
Cut-pasting a parent folder to a child folder crashes the app.

**To Reproduce**
1. Open files-uwp app.
2. Cut a folder which has children folders, i.e. cut "C:\Users\Documets\Test"
3. Try pasting it to a child folder, i.e. paste in "C:\Users\Documents\Test\ChildTest"

**Expected behavior**
The operation should prompt a dialog giving user to either skip or cancel the operation. 

**Screenshots**

**Desktop (please complete the following information):**
 - OS Version: Windows 10 1903 18362.239
 - App version: v0.12.0

**Additional context**
Although there exists logic for handling this type of cases by showing dialog, but the dialog creation throws exception. Error log:
```
2020-07-23 10:56:09.3884|ERROR|Files.App|The application called an interface that was marshalled for a different thread. (Exception from HRESULT: 0x8001010E (RPC_E_WRONG_THREAD))|System.Exception: The application called an interface that was marshalled for a different thread. (Exception from HRESULT: 0x8001010E (RPC_E_WRONG_THREAD))
   at Windows.UI.Xaml.Data.Binding..ctor()
   at Files.Interacts.Interaction.<>c__DisplayClass60_0.<<PasteItems>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Files.Interacts.Interaction.PasteItems(DataPackageView packageView, String destinationPath, DataPackageOperation acceptedOperation)
   at Files.Interacts.Interaction.PasteItem_ClickAsync(Object sender, RoutedEventArgs e)
   at System.Threading.WinRTSynchronizationContextBase.Invoker.InvokeCore()
```

yaichenbaum:
@soumyamahunt This issue should be resolved with #2742.

PR #2742, opened 2020-12-28
Fixed an issue where trying to move a parent directory to a child directory would crash the app

Don-Vito:
Closes: https://github.com/files-community/Files/issues/1546
Closes: https://github.com/files-community/Files/issues/1545

Breaking change: do not proceed to Deletion during Move
if Copy failed (to prevent data loss)

### DIFF
--- a/Files/Filesystem/FilesystemOperations/FilesystemOperations.cs
+++ b/Files/Filesystem/FilesystemOperations/FilesystemOperations.cs
@@ -280,6 +280,12 @@ namespace Files.Filesystem
                                                      IProgress<FilesystemErrorCode> errorCode,
                                                      CancellationToken cancellationToken)
         {
+            if (Path.GetFileName(source.Path) == destination)
+            {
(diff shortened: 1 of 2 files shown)


Don-Vito:
@gave92 - I was really hesitating if to put the check only there or rather in `FilesystemOperations.MoveAsync` so all the cases are covered.

WDYT?  If I move it to `FilesystemOperations` I guess I will need to return a null storage history.

gave92:
I agree it's better to move it to FilesystemOperations.
This is an example taken from FilesystemOperations.RenameAsync() which looks similar to what you're doing here
```cs
if (Path.GetFileName(source.Path) == newName)
{
    errorCode?.Report(FilesystemErrorCode.ERROR_ALREADYEXIST);
    return null;
}```

Don-Vito:
> example taken from FilesystemOperations.RenameAsync(

Thanks! Doing it right now :smile:

gave92 (approved):

yaichenbaum:
@Don-Vito Great work! LGTM

PR #2750, opened 2020-12-29
Fix release build issues with SQLite binaries

jakoss:
This is a fix for an issue reported in https://github.com/files-community/Files/pull/2655


gave92:
@jakoss Could you verify that this does not add an extra SqLite binary to the package? An SQLBinary is already included due to AppCenter

jakoss:
@gave92 It seems like it does, but i'm not sure what to do about that. SQLitePCLRaw clearly does not work in current setup.

I can see binaries: `e_sqlite3.dll` and `sqlite3.dll`

gave92:
@jakoss what happens if you update Microsoft.Data.Sqlite from 1.1.1 to 5.0.1?

gave92:
@jakoss try updating Microsoft.Data.Sqlite from 1.1.1 to 5.0.1, seems to get rid of the extra dll

jakoss:
It compiles without extra dll, but it does not work with the exact exception from PR #2655

gave92:
> It compiles without extra dll, but it does not work with the exact exception from PR #2655

@jakoss dammit :) I'd really like to avoid adding a second sqlite binary. I'm going to a try few experiments too

jakoss:
I wonder if AppCenter can use this library or is it crashing silently in background. Because there's not much one can do if native library won't match..

gave92:
@jakoss maybe you're right. Do you think the `Default.Rd.xml` could have somthing to do with this?

yaichenbaum:
> @jakoss maybe you're right. Do you think the `Default.Rd.xml` could have something to do with this?
> I'm gonna try adding `<Assembly Name="*Application*" Dynamic="Required All" />` to it. Is it something you tried already?

That's a possibility. Adding sqlite to `rd.xml` shouldn't be too difficult.

jakoss:
@gave92 Even though i worked with .NET for 10 years now - i don't have any prior experience with .NET Native or UWP. So i'm lost here more then you do :)

gave92:
@jakoss @yair100 I can confirm that changing the Default.Rd.xml fixes this 🚀🚀🚀
Just need to figure out what namespace is needed, currently I just added `<Assembly Name="*Application*" Dynamic="Required All" />` which is everything

yaichenbaum:
@gave92 The namespace is probably `Microsoft.Data.Sqlite`.

gave92:
@yair100 I hope so, that's what I'm trying right now. Release build is just sooo slow 😃

yaichenbaum:
@gave92 I just took a look and it looks like it might actually be `Microsoft.Data.Sqlite.Core`.

gave92:
@yair100 I'll try that next. How did you guess?
[Btw, since we are doing this we could also try to fix empty the exception notification]

yair100:
> @yair100 I'll try that next. How did you guess?

That's what it comes up as under the references section in solution explorer.

gave92:
Of course it was neither xD It was `SQLitePCL`
Going to open a PR with this

gave92:
@jakoss I hope you don't mind if I close this PR. Thank you for your amazing work on #2655

jakoss:
No worries

Issue #2450, opened 2020-11-26
Feature: Details Pane 


mbeps:
**Is your feature request related to a problem? Please describe.**
Viewing details about specific files already exists. It should be possible to add a details pane on the side similar to the built-in File Explorer. This will allow the user to view file details without going to the properties. 

**Describe the solution you'd like**
In the settings, add an option to show the details pane similar to File Explorer. Once enabled, the details pane should be on the right side showing the details about the file selected. If nothing is selected, the pane should show the details about the current folder. The details pane should not disappear when nothing is selected as this will cause confusion and cause the icons in the folder to move around unnecessarily. 

**Describe alternatives you've considered**
Add a shortcut key that would automatically open a flyout showing the details. Possibly open the properties and automatically go the details tab automatically. 

**Additional context**
Add any other context or screenshots about the feature request here. Ex: What device type/UI layout and app version?
![image](https://user-images.githubusercontent.com/58662575/100360994-86cae600-2ff1-11eb-9dcd-e710c1b8188d.png)

winston-de:
@yair100 Can you please assign me to this?

yair100:
> @yair100 Can you please assign me to this?

Sure! Do you have any designs in mind before you start?

winston-de:
@yair100 My plan is to use the design and most of the code from the properties: details page, but without the section headers and the clear properties button.
I was also planning on adding the file icon, an "open properties window" button, and ok/cancel buttons (that are revealed when the user makes a change) like Windows Explorer has.

yaichenbaum:
What do you think about adding a grid splitter so users can resize the pane?

winston-de:
@yair100 Good idea.

mdtauk:
I think this is a space where some thought of information design could be put in.

Text hierarchy

![image](https://user-images.githubusercontent.com/7389110/102031143-f6d8ba80-3dac-11eb-986f-31da1a10c116.png)


Headers and body text, and ways to visualise the non-text based data.

Grouping of information such as permissions and status, and ordering these groups based on what typical file information a user wants to see.

Different filetypes may also display their file details in different ways.  Photos may focus on dimensions, the preview, as well as file size and formats.

winston-de:
@mdtauk 
I agree with everything you said.

Personally I like the idea of using icons (like the one in the google photos sample). It would make it much more compact when the details pane is narrowed.

[...]

Issue #130, opened 2019-06-04
Selected Item Preview Pane


eraychumak:
By the default, there's a lot of empty space on the right section of the explorer, it would be a nice addon to be able to preview your selected file _(very rough example below...)_ so you don't have to install applications like QuickLook to do the same thing.

![file_uwp](https://user-images.githubusercontent.com/44278658/58911049-e99a3480-870e-11e9-81c5-dcb8b96102e4.png)

Poopooracoocoo:
File Explorer calls it Details Pane and Preview Pane fyi

yaichenbaum:
Files UWP works with QuickLook, is there a benefit of including similar functionality when QuickLook already does this?

duke7553:
This would be a new experience powered by QuickLook (if possible)

EukalyptenHD:
> Files UWP works with QuickLook, is there a benefit of including similar functionality when QuickLook already does this?

Why download an entirely separate app for this one feature when you can just support it natively?

yaichenbaum:
@EukalyptenHD The reason would be is the QuickLook provides additional functionality, of course having a built in option would be nice as well, but for the meantime, I would recommend trying out QuickLook.

duke7553:
Considering a separate app exists (and is now integrated with Files), I don't think this will be an area of focus for near future unless we hear more about how something like this can add unique value to the experience.

callmehanu:
Yep, for me **preview pane** is also really necesary. Quicklook works greate but need an extra trigger. For example if I looking for a photo, I need to select photos one by one and after hit spacebar, this is really tiresome. Yep, Quicklook is pretty and fancy but for "real work" lack of convenience. 

On other hand, [Metro Ccommander](https://www.microsoft.com/store/productId/9WZDNCRFJB30) offers a **full screen** preview pane with many powerfull shorcuts but with this we loose the focus. 

I think a preview pane can really be a greate tool for hard users and the most simple way to find folders and files on the fly,

And if you need help maybe I can work the design, I'm ux designer!

yaichenbaum:
@callmehanu A preview pane is on our backlog. We would love to work with the community towards creating a design to base the feature off of. Concepts are welcome, but keep in mind that the final decision is up to the maintainers of the app.

winston-de:
I've been working on making a preview pane the past couple of weeks (#2756).
Since I don't have the time to cover all file types, what would you guys consider the most useful? 

[...]

PR #2756, opened 2020-12-29
Added a details/preview pane

winston-de:
This PR adds a details/preview pane for the selected items.

#2450 #130

### To-do:
- [x] Text file preview
- [x] Markdown file preview
- [x] Resizable pane
- [x] Image file preview
- [x] Support for previews through extensions
- [x] PDF previews 
- [x] Audio/video file preview 
- [x] .html preview
- [x] code source files (.cs, .xml. .json) preview (w/ colors!)
- [x] Bottom-aligned pane for narrow windows
- [x] Display file properties
- [x] Pane visibility toggle button and shortcut
- [x] Opening and closing animations
- [x] Support for all layouts
- [x] Add option to disable and edit adaptive preview pane positioning
- [x] Use x:Load instead of visibility
- [x] Fix issue where preview pane cuts of part of file browser while collapsed
- [x] Hide pane on home page
- [x] Allow details to be provided by extensions and preview controls
- [x] Display preview text files regardless of extension

### For next PR
- Prevent preview from being refreshed when same file is re-selected
- Add setting to not load preview for online-only files
- Different methods of displaying values for details
- Investigate using icons
- Fix issue where rectangle selection thing does not stop at preview pane bounds
- Fix issue where preview pane is refreshed when same file is re-selected
- Add setting to not load preview for online-only files

https://user-images.githubusercontent.com/59544401/104795770-a49d0780-5765-11eb-9300-7f0b4c98a9f3.mp4

Feedback is appreciated! :)


Files/Converters/BooleanToVisibilityConverter.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Data;
+
+namespace Files.Converters
+{
+    class BooleanToVisibilityConverter : IValueConverter
```
tsvietOK, line 11: You can use WCT built-in [converter](https://docs.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.ui.converters.booltovisibilityconverter?view=win-comm-toolkit-dotnet-stable&viewFallbackFrom=uwp-toolkit-dotnet).

Files/UserControls/DetailsPane.xaml.cs
```diff
@@ -0,0 +1,227 @@
...
+                return;
+            }
+
+            PreviewNotAvaliableText.Visibility = Visibility.Visible;
+
+        }
+
+        UserControl GetBuiltInPreviewControl(ListedItem item)
+        {
+            if (MediaPreview.Extensions.Contains(item.FileExtension))
```
tsvietOK, line 154: I guess comparing should be case insensitive?

Files/Converters/BooleanToVisibilityConverter.cs
winston-de: I didn't realize there was one. Thanks!

yaichenbaum:
@winston-de Can you try moving the preview pane toggle button to the navigation toolbar?

Files/Views/SettingsPages/FilesAndFolders.xaml
```diff
@@ -71,6 +71,12 @@
                     Header="Enable individual preferences for individual directories"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{Binding AreLayoutPreferencesPerFolder, Mode=TwoWay}" />
+                
+                <ToggleSwitch
+                    x:Uid="SettingsFilesAndFoldersEnableAdaptivePreviewPane"
+                    Header="Enable adaptive preview pane"
+                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
+                    IsOn="{Binding EnableAdaptivePreviewPane, Mode=TwoWay}" />
```
yaichenbaum, line 79: Should this go under the preference settings page?

Files/ViewModels/SettingsViewModel.cs
```diff
@@ -730,6 +730,18 @@ public SortOption DefaultDirectorySortOption
...
+        {
+            get => Get(false);
+            set => Set(value);
+        }
+
+        public bool EnableAdaptivePreviewPane
+        {
+            get => Get(true);
+            set => Set(value);
+        }
```
yaichenbaum, line 743: Can you move these to the preferences region and add xml to improve IntelliSense?

Files/UserControls/StatusBarControl.xaml
```diff
@@ -47,6 +47,23 @@
...
             Orientation="Horizontal"
             Spacing="2">
+            
+            <Button 
+                x:Name="PreviewPane"
+                x:Uid="PreviewPaneToggle"
+                FontFamily="{StaticResource FluentUIGlyphs}" Content="&#xEA8D;"
+                Width="36"
+                Height="32"
+                Margin="2,0,2,0"
```
yaichenbaum, line 57: We made some changes to the spacing an margins in `master`, can you try synchronizing with `master` to make sure this is styled correctly? Do you have any thoughts on making this a toggle button?

yair100 (changes requested):

Files/UserControls/StatusBarControl.xaml
winston-de: @yair100 
I'll update the margins now.
I didn't make this a toggle button because I couldn't get the custom styling used for the other buttons to work properly.
Instead of turning the accent color, it would just show the hover state when pressed.

winston-de:
> @winston-de Can you try moving the preview pane toggle button to the navigation toolbar?

@yair100 Like this?

image

yaichenbaum:
@winston-de Kind of, I meant to put it next to the search icon.

winston-de:
@yair100 

image

winston-de:
@yair100 Here it is to the right of the search box and button
image

yair100 (dismissed):

yair100 (dismissed):

eraychumak:
- I would
-

eraychumak:
Suggestions based on the video above:

- The spacing between the property names and values is quite wide, I would reduce it by half or even less so that you can see more of the value because most of it overflows ruining the point of glancing at it quickly.
- The line separator between the preview and properties should be aligned and be the same width (when the panel is on the right) and height (when the panel is on the bottom).
- The spinner should also be centered vertically and the style should be more subtle and be more like the Windows 10 tail spinner that uses a bunch of spaced circles.

tsvietOK:
@winston-de When app light theme enabled and dark system theme enabled:
image
Can you remove background from this code preview control?

tsvietOK (changes requested):
Text "No preview available" center horizontally and vertically in view. By the way, here is mistake in this text "No preview avaliable".

tsvietOK:
I think ScrollViewer could be removed from ImagePreview, because it is impossible to scroll wide images. 
Preview of .doc files is not working, app just freezes.

winston-de:
> Resize mouse icon is not always visible when hovering over vertical resize line.

@tsvietOK I believe this is an issue with the GridSplitter control. If you go the the Toolkit sample app and change the GripperCursor property, the cursor doesn't show on mouse hover.

winston-de:
> Optionally you can add name of selected file to preview pane (above or below item preview).

> The spacing between the property names and values is quite wide, I would reduce it by half or even less so that you can see more of the value because most of it overflows ruining the point of glancing at it quickly.

> The line separator between the preview and properties should be aligned and be the same width (when the panel is on the right) and height (when the panel is on the bottom).

@tsvietOK @eraychumak I made these changes, what do you guys think?

image

I also added a tooltip for the property name in the event it gets cut off.

winston-de:
> @winston-de When app light theme enabled and dark system theme enabled:
> image
@tsvietOK 

I'm not sure why that's happening. It only seems to occur with xml, so I'll remove xml support from the code block for the time being.

> Can you remove background from this code preview control?

Fixed.

winston-de:
> The spinner should also be centered vertically and the style should be more subtle and be more like the Windows 10 tail spinner that uses a bunch of spaced circles.

@eraychumak 
The progress ring is only shown for PDFs, as they take much longer to load than other file types, and is shown at the bottom of the page list.
As for the style, the style used is the one provided by the Windows UI library, and represents the latest design standards for Windows. 
However, I could replace it with a progress bar, which is more subtle

<image src="https://user-images.githubusercontent.com/59544401/104959929-1ce31300-5988-11eb-9301-538cef26d6b8.png" Width="500"/>

eraychumak:
> @tsvietOK @eraychumak I made these changes, what do you guys think?

The name of the file in the preview pane is quite squished, add more spacing on the top and bottom. 

However, I'd say it's quite redundant since you can see the name on the left already and the previewed file is highlighted for you so it's just visual clutter.

winston-de:
> > @tsvietOK @eraychumak I made these changes, what do you guys think?
> 
> The name of the file in the preview pane is quite squished, add more spacing on the top and bottom.
> 
> However, I'd say it's quite redundant since you can see the name on the left already and the previewed file is highlighted for you so it's just visual clutter.

I think it could be useful, for example if the user wants to copy the name of the file.

I could put it as a file property instead of in the header, so that it feels less cluttered.
image

eraychumak:
> I could put it as a file property instead of in the header, so that it feels less cluttered.

Ah there you go, yeah that's okey, whats up with the date and that `-08:00`? It could be formatted better like this:

image

yaichenbaum:
Is it possible to use a TextBlock for the properties instead of a TextBox?

eraychumak:
> Is it possible to use a TextBlock for the properties instead of a TextBox?

People will not be able to select the filename to copy as @winston-de intended for it.

tsvietOK:
@eraychumak 
> However, I'd say it's quite redundant since you can see the name on the left already and the previewed file is highlighted for you so it's just visual clutter.

File name could be useful if it is too long to show it in explorer window.

yaichenbaum:
> > Is it possible to use a TextBlock for the properties instead of a TextBox?
> 
> People will not be able to select the filename to copy as @winston-de intended for it.

You can still select text in a TextBlock.

winston-de:
> Is it possible to use a TextBlock for the properties instead of a TextBox?

@yair100 
Yes, though I think the read-only TextBox is better as it makes it easy to view the full value by scrolling if it gets cut off.

> Ah there you go, yeah that's okey, whats up with the date and that -08:00? It could be formatted better like this:

@eraychumak Thanks, fixed.

yaichenbaum:
> Yes, though I think the read-only TextBox is better as it makes it easy to view the full value by scrolling if it gets cut off.

@winston-de What about removing the background?

mdtauk:
Just sharing a design for the details pane
image

yaichenbaum:
@mdtauk I like the idea of using a different color background for the pane, can you share the brushes that you used?

eraychumak:
> 
> 
> @mdtauk I like the idea of using a different background color for the pane, can you share the brushes that you used?

It's #eeeeee

mdtauk:
> @mdtauk I like the idea of using a different background for the pane color, can you share the brushes that you used?

I'm using the new brushes, so you would have to do something custom, or wait until you move to winUI 2.6 (which I think will use the new values maybe?)

image

image

winston-de:
I think in order for the separate background color to look good, there would have to be some visual separation with the status bar, navigation toolbar, and preview pane. Otherwise it would look something like this: 

image


Anyone have any advice for this?

winston-de:
Also @yair100 here's what the text boxes look like now.
image

tsvietOK:
I think it is better to leave original style of GridSplitter.
image

yaichenbaum:
> I think it is better to leave original style of GridSplitter.
> image

I am not sure we even need to show the GridSplitter if the background has a different color.

winston-de:
@tsvietOK  Like this?
image

tsvietOK:
> > I think it is better to leave original style of GridSplitter.
> > image
> 
> I am not sure we even need to show the GridSplitter if the background has a different color.

GridSplitter needed because user should know, that there is possible to resize this area.

d2dyno1:
The gridspliiter is not visible in the default File Explorer and users get around this and know it can be resized

tsvietOK:
@winston-de Left one definitely yes. I am not sure with right one (horizontal).

mdtauk:
> > > I think it is better to leave original style of GridSplitter.
> > > image
> > 
> > 
> > I am not sure we even need to show the GridSplitter if the background has a different color.
> 
> GridSplitter needed because user should know, that there is possible to resize this area.

I would replace those two lines in the GridSplitter, and go for a lozenge to match the more modern grab indicator.
image
image

winston-de:
Like this?
image

d2dyno1:
> Like this?
> 
> 

Maybe without the background?

mdtauk:
> > Like this?
> > 
> 
> Maybe without that gray background?

If the pill shape has good contrast, the background can be more subtle (with good PointerOver and Pressed colours)

winston-de:
> > Like this?
> > 
> 
> Maybe without that gray background?

mdtauk:
> > > Like this?
> > > 
> > 
> > 
> > Maybe without that gray background?
> 
> 

If the pill shape has good contrast, the background can be more subtle (with good PointerOver and Pressed colours)

winston-de:
> > > > Like this?
> > > > 
> > > 
> > > 
> > > Maybe without that gray background?
> > 
> > 
> > 
> 
> If the pill shape has good contrast, the background can be more subtle (or Transparent, with good PointerOver and Pressed colours)

@mdtauk The background is transparent in this screenshot.

d2dyno1:
> @winston-de Can you increase height of this pill and do it less bright, please? User should have ability to easy click and drag this element.

Instead making it bigger, increase the click area

mdtauk:
> > @winston-de Can you increase height of this pill and do it less bright, please? User should have ability to easy click and drag this element.
> 
> Instead of making it bigger, increase the click area

The pill doesn't need to be bigger, if you include a good Hover/PointerOver state which takes up the full width

winston-de:
@tsvietOK @d2dyno1 The click area is the same as before. All that has changed is the icon thing and the background.
The click area is highlighted here
image

winston-de:
@tsvietOK How does this look for the color?
image

It's just a bit darker than it is now.

tsvietOK:
@winston-de Good. But now it is floating inside view when no preview available. 
image

winston-de:
@tsvietOK I suppose I could add text for "no details available"
That would fix that issue.

tsvietOK:
@winston-de Is it possible to hide details (bottom part) when no preview available, so text No preview available will be centered vertically?

gave92:
@winston-de sorry to add one thing more :) A small issue is present when triggering the preview pane in double-pane mode: a thick grey bar appears at the top-right like this:

image

mdtauk:
image
At Rest and Hover

winston-de:
> @winston-de Is it possible to hide details (bottom part) when no preview available, so text No preview available will be centered vertically?

It's possible, but I think the details should not be bound to whether or not a preview is present. I intend to allow file properties to be shown even if there's no preview available in the future.

Files/UserControls/PreviewPane.xaml.cs
```diff
@@ -0,0 +1,241 @@
...
+            }
+
+            return null;
+        }
+
+        async void LoadPreviewControlFromExtension(ListedItem item, Helpers.Extension extension)
+        {
+            var file = await StorageFile.GetFileFromPathAsync(item.ItemPath);
+
+            var buffer = await FileIO.ReadBufferAsync(file);
```
gave92, line 196: Should we just pass the filepath to the extension? Loading the whole file could be really big.

mdtauk:
Rest - Hover - Pressed states

image

Files/UserControls/PreviewPane.xaml.cs
winston-de: @gave92 I think the issue with that is then the extensions would have to have the full filesystem access capability.

winston-de: But it's probably a good idea to put a file size limit on that and just send the path.

tsvietOK:
There is a lot place to the right of details values, but some keys (like Camera Manufacturer) are cut off.
image

Files/UserControls/PreviewPane.xaml.cs
gave92: @winston-de right I didn't think of that 🤦‍♂️ There's no way to "share" the access to a StorageFile with an extension is there? The extensions would have to open a file picker

winston-de:
> There is a lot place to the right of details values, but some keys (like Camera Manufacturer) are cut off.

I added a tooltip for the titles. If that isn't enough I guess I could put it in a styled textbox.

yaichenbaum:
@winston-de What do you think about hiding the "pill" to resize the pane, and to only show the one inside the pane between the details and preview?

Files/UserControls/PreviewPane.xaml.cs
gave92: Perhaps you can actually do it: https://docs.microsoft.com/en-us/uwp/api/Windows.ApplicationModel.DataTransfer.SharedStorageAccessManager?view=winrt-19041

winston-de:
@yair100 I didn't put a pill for the pane resize.

winston-de:
I think putting a pill for the pane resize would be too confusing with the scrollbar right there.

mdtauk:
> I think putting a pill for the pane resize would be too confusing with the scrollbar right there.

I am finding this to be the case as I toy around with UI ideas

mdtauk:
image

you can compensate with extra padding above and to the other sides

yair100:
> @yair100 I didn't put a pill for the pane resize.
> 

Great! What about removing the border as well? Perhaps a shadow would look better...

mdtauk:
> > @yair100 I didn't put a pill for the pane resize.
> > 
> 
> Great! What about removing the border as well? Perhaps a shadow would look better...

You can keep a border, and use a ThemeShadow - at least until you find a solution for the grabber

mdtauk:
@winston-de @yair100  Have you planned for reflowing the content of the details pane as you resize the pane?  Headers from the left to above maybe?

What about Minimum widths, and maximum widths?

winston-de:
There are maximum and minimum widths in place.

winston-de:
> > @yair100 I didn't put a pill for the pane resize.
> > 
> 
> Great! What about removing the border as well? Perhaps a shadow would look better...

@yair100 I think shadows would look great, but I have no clue how to do that 😂
If someone who knows how to do that sort of thing could contribute, it would be greatly appreciated.

Files/UserControls/PreviewPane.xaml.cs
winston-de: @gave92 Thanks, that's a much better solution! I probably won't be able to implement it for this PR though. For the time being I just put a file size limit.

winston-de: @gave92 I was able to get it working with no issues. Thanks again for the tip!

yaichenbaum: A bit out of scope for this PR, but we should probably use that API when launching files as well.

callmehanu:
I'm running 1.1 version but I can't see the preview icon. 
Screenshot 2021-01-18 204502
Screenshot 2021-01-18 204555

winston-de:
@callmehanu This feature is not available yet, as it has not been merged.

winston-de:
> @winston-de sorry to add one more thing :) A small issue is present when triggering the preview pane in double-pane mode: a thick grey bar appears at the top-right like this:
> 
> image
> 
> Only happens if second pane is on home.

@gave92 
I changed it so the preview pane option is no longer app wide, so this issue should be fixed now. :)

tsvietOK:
@winston-de Looks like .iso files are opening as TextPreview.

winston-de:
> @winston-de Looks like .iso files are opening as TextPreview.

@tsvietOK 
What happens? I don't have an .iso file to test with.
It's supposed to detect when a file is plain text, and if so display it as text preview.

tsvietOK:
@winston-de 
image

winston-de:
@tsvietOK Does this happen with all .iso files, or just that one?

tsvietOK:
@winston-de This happens with all .iso files. By the way, it takes 20 sec for showing details (even wrong) for 2.26Gb file.

winston-de:
Hmmmmm. 
I could add an exclusion for .iso files I guess, but I'm not sure what's causing the program to think it's a text file though.

tsvietOK:
@winston-de I think the problem is in check:
```cs
if (text.Contains("\0\0\0\0"))
{
    return null;
}

```

winston-de:
> @winston-de I think the problem is in check:
> 
> ```cs
> if (text.Contains("\0\0\0\0"))
> {
>     return null;
> }
> ```
I don't think that's it, as that's just a backup check if the file contains binary.
The `FileIO.ReadTextAsync` normally throws an error when a file isn't a text file.

winston-de:
Anyway, I added a list of excluded extensions, and if any other similar cases pop up they can be added.

tsvietOK:
@winston-de For some reason ReadTextAsync doesn't throw an exception and loads an empty string.
image

winston-de:
@tsvietOK Yeah, that's what I figured was happening.
Out of curiosity, what happens when you open the file with WordPad or another text editor?

tsvietOK:
@winston-de File is too large to be opened in text editor.

winston-de:
Odd. Maybe it's an issue with the `FileIO.ReadTextAsync` function?
Anyway, it shouldn't be an issue anymore.

winston-de:
Also, this is what it would look like if the pane had a pill for the resize bar.

image

yaichenbaum:
> Also, this is what it would look like if the pane had a pill for the resize bar.
> 
> Vertical:
> image
> 
> Horizontal:
> image
> 
> I'm using 5px margins on either side of the resize bar.

I think it looks a little out of place, I personally would prefer if the vertical resize bar didn't have any pill.

winston-de:
> > Also, this is what it would look like if the pane had a pill for the resize bar.
> 
> I think it looks a little out of place, I personally would prefer if the vertical resize bar would only show the pill on hover.

I think that would be a usability issue for touch devices.

tsvietOK:
> > > Also, this is what it would look like if the pane had a pill for the resize bar.
> > 
> > 
> > I think it looks a little out of place, I personally would prefer if the vertical resize bar would only show the pill on hover.
> 
> I think that would be a usability issue for touch devices.

Yes, but vertical pill looks like secondary scroll bar.

yaichenbaum:
@winston-de The resize bar for the sidebar and dual pane view don't display any indicator, and it hasn't been an issue for touch devices, so I would imagine it's the same over here.

winston-de:
> @winston-de The resize bar for the sidebar and dual pane view don't display any indicator, and it hasn't been an issue for touch devices, so I would imagine it's the same over here.

It could just be me, but I have to press half a dozen times to find the touch point to resize either of those.

Also, I think it would look less out of place if the background color was slightly different for the pane, or if there was some sort of shadow effect.

eraychumak:
> Also, I think it would look less out of place if the background color was slightly different for the pane, or if there was some sort of shadow effect.

I'd remove the resize bars entirely and rely on the background being slightly different to distinguish the edge of the preview panel, then clicking/dragging anywhere on the edge will allow you to resize otherwise it all looks weird, there's like 3 different bars that look like scrollbars

winston-de:
Maybe we could just make the pill optional?

Or perhaps even detect if the device has a touch screen, and if it does enable it by default?

winston-de:
The pill icon is just a TextBlock, I could add a visual state that just makes it blank when the option is disabled.

eraychumak:
If you look at the native file explorer, they simply use the entire edge to resize the sidebar, so it should be familiar/intuitive (maybe even expect) for users to have the details pane resize work the same way

winston-de:
> If you look at the native file explorer, they simply use the entire edge to resize the sidebar, so it should be familiar/intuitive (maybe even expected) for users to have the details pane resize work the same way

Resizing the preview pane on a touch screen is next to impossible in Windows Explorer.

winston-de:
I removed the background, as requested.
The grid splitter now shows only on hover.

Vertical:


Horizontal:

Files/Strings/en-US/Resources.resw
```diff
@@ -1536,6 +1536,9 @@
   <data name="SettingsFilesAndFoldersHideSystemItems.Header" xml:space="preserve">
     <value>Hide protected operating system files (Recommended)</value>
   </data>
+  <data name="SettingsFilesAndFoldersEnableAdaptivePreviewPane.Header" xml:space="preserve">
```
yaichenbaum, line 1539:
```suggestion
  <data name="SettingsEnableAdaptivePreviewPane.Header" xml:space="preserve">
```

yair100 (changes requested):

Files/Filesystem/ListedItem.cs
```diff
@@ -178,6 +180,14 @@ public DateTimeOffset ItemDateAccessedReal
 
         private DateTimeOffset itemDateAccessedReal;
 
+        private ObservableCollection<FileProperty> _itemProperties;
```
d2dyno1, line 183: CodeStyle, remove `_`

Files/Helpers/ExtensionManager.cs
```diff
@@ -0,0 +1,531 @@
...
+using Windows.UI.Core;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Media.Imaging;
+
+namespace Files.Helpers
+{
+    class ExtensionManager
+    {
+        //private CoreDispatcher _dispatcher; // used to run code on the UI thread for code that may update UI
+        private AppExtensionCatalog _catalog; // the catalog of app extensions available to this host
```
d2dyno1, line 24: :eyes:

Files/Helpers/ExtensionManager.cs
```diff
@@ -0,0 +1,531 @@
...
+        /// <summary>
+        /// Sets up handlers for package events
+        /// Be sure to call this from the UI thread.
+        /// </summary>
+        /// <param name="dispatcher"></param>
+        public void Initialize()
+        {
+            #region Error Checking & Dispatcher Setup
+            // verify that we haven't already been initialized
+            //if (_dispatcher != null)
```
d2dyno1, line 62: Remove unnecesary code

Files/Helpers/ExtensionManager.cs
```diff
@@ -0,0 +1,531 @@
...
+                var text = await FileIO.ReadTextAsync(file);
+                FileExtensions = JsonConvert.DeserializeObject<List<string>>(text);
+            } catch
+            {
+                Debug.WriteLine("Unable to get extensions");
+            }
+
+            Loaded = true;
+            Visible = Visibility.Visible;
+            RaisePropertyChanged("Visible");
```
d2dyno1, line 469: `nameof(Visible)`

Files/Helpers/ExtensionManager.cs
```diff
@@ -0,0 +1,531 @@
...
+                Unload();
+            }
+        }
+        #region PropertyChanged
+
+        /// <summary>
+        /// Typical property changed handler so that the UI will update
+        /// </summary>
+        /// <param name="name"></param>
+        private void RaisePropertyChanged(string name)
```
d2dyno1, line 522: You could inherit from `ObservableObject` - it already has `OnPropertyChanged()`

Files/UserControls/FilePreviews/TextPreview.xaml.cs
```diff
@@ -0,0 +1,106 @@
...
+using Windows.Storage;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Documents;
+
+// To learn more about WinUI, the WinUI project structure,
+// and more about our project templates, see: http://aka.ms/winui-project-info.
+
+namespace Files.UserControls.FilePreviews
+{
+    public sealed partial class TextPreview : PreviewControlBase
```
d2dyno1, line 26: MVVM 😶

Files/UserControls/NavigationToolbar.xaml
```diff
@@ -849,7 +849,7 @@
             HorizontalAlignment="Right"
             Orientation="Horizontal"
             Spacing="2">
-
+
```
d2dyno1, line 852: indentation, empty spaces?

Files/ViewModels/SettingsViewModel.cs
```diff
@@ -536,6 +536,23 @@ public bool PinRecycleBinToSideBar
             }
         }
 
+        ///// <summary>
+        ///// Gets or sets the value indicating whether the preview pane should be shown.
+        ///// </summary>
+        //public bool PreviewPaneEnabled
```
d2dyno1, line 542: unnecessary code?

winston-de: Yeah, I forgot to delete that after moving the setting to InstanceViewModel.

Files/UserControls/FilePreviews/TextPreview.xaml.cs
winston-de: @d2dyno1 Can you tell me specifically what you want me to change?

d2dyno1: We want to merge into MVVM pattern sometime in the future. It'd be better if new PRs came with MVVM

Files/Helpers/ExtensionManager.cs
winston-de: @d2dyno1 I think this code should be kept as it's for updating the UI (say, a page for managing extensions), but isn't currently implemented.

Files/UserControls/FilePreviews/TextPreview.xaml.cs
winston-de: @d2dyno1 What exactly do I need to change?

d2dyno1: Ideally you'd have to convert all `.xaml.cs` (code-behind) files to View Models (`somefileViewModel`). If you have any problems with it, check out #2881, it follows MVVM pattern and it can help answer some of your questions :)

Files/Helpers/ExtensionManager.cs
winston-de: @d2dyno1 Is this change necessary? It would require rewriting a decent amount of code.

tsvietOK:
Sometimes No details preview is visible. 
Steps:
1. Select archive (or another file without preview)
2. Fast select video file (or another file with preview)

image

Files/UserControls/PreviewPane.xaml.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        public PreviewPane()
+        {
+            this.InitializeComponent();
+            RegisterPropertyChangedCallback(Grid.RowProperty, GridRowChangedCallback);
+        }
+
+        public event PropertyChangedEventHandler PropertyChanged;
+        void RaisePropertyChanged(string name)
```
tsvietOK, line 103: Seems like it is not used.

Files/Views/ModernShellPage.xaml
```diff
@@ -244,17 +315,38 @@
...
+                        <Grid>
+                            <TextBlock HorizontalAlignment="Center" IsHitTestVisible="False" VerticalAlignment="Center" x:Name="GripperGlyphTextBlock"
+                               Text="" FontSize="36" Foreground="{StaticResource SystemChromeHighColor}" FontFamily="{StaticResource FluentUIGlyphs}">
+                            </TextBlock>
+                        </Grid>
+                    </Custom:GridSplitter.Element>
+                </Custom:GridSplitter>
+
+                <controls:PreviewPane Grid.Row="1" Grid.Column="2" x:Name="PreviewPane" 
+                                    SelectedItems="{x:Bind ContentPage.SelectedItems, Mode=OneWay}"
```
tsvietOK, line 333: `BaseLayout` contains SelectedItem property, so what about to use it?

winston-de: The SelectedItem property doesn't update after the selection changes.

tsvietOK: It does.
image

winston-de: Odd. I had tried using it at first, but I switched to SelectedItems because it wasn't updating.
I guess I'll change it back.

tsvietOK: Optionally you can add `NotifyPropertyChanged(nameof(SelectedItem));` inside.

winston-de: @tsvietOK SelectedItem will always show the first item in the list, no matter how many items are selected. This will not work with the PreviewPane, as it should not show anything when multiple items are selected. 

image

Files/Helpers/Extension.cs
```diff
@@ -0,0 +1,265 @@
...
+        {
+            if (this.properties.ContainsKey("Service"))
+            {
+                PropertySet serviceProperty = this.properties["Service"] as PropertySet;
+                serviceName = serviceProperty["#text"].ToString();
+            }
+        }
+        #endregion
+
+        UniqueId = ext.AppInfo.AppUserModelId + "!" + ext.Id; // The name that identifies this extension in the extension manager
```
tsvietOK, line 61:
```suggestion
        UniqueId = $"{ext.AppInfo.AppUserModelId}!{ext.Id}"; // The name that identifies this extension in the extension manager
```

tsvietOK (approved):
LGTM

yair100 (approved):

yaichenbaum:
:shipit:

PR #2760, opened 2020-12-30
CodeMaid

yair100:


tsvietOK (dismissed):

tsvietOK (approved):

Don-Vito:
@yair100, @tsvietOK - have you considered a github action that could require correct formatting.

E.g., we could use [this action](https://github.com/marketplace/actions/dotnet-format) that runs a [formatting tool](https://github.com/dotnet/format) to check if there are code style issues.
We can instruct the developers to run this tool manually before creating PR as well to prevent action failures.

Additional advantage of enforcing formatting prior to merging is that commits like this might create conflicts.

WDYT?

yair100:
> @yair100, @tsvietOK - have you considered a github action that could require correct formatting.
> 
> E.g., we could use [this action](https://github.com/marketplace/actions/dotnet-format) that runs a [formatting tool](https://github.com/dotnet/format) to check if there are code style issues.
> We can instruct the developers to run this tool manually before creating PR as well to prevent action failures.
> 
> Additional advantage of enforcing formatting prior to merging is that commits like this might create conflicts.
> 
> WDYT?

It's something we wanted to do for a while, is this something you would like to work on?

Don-Vito:
Why not - sounds like a fun learning task! :smile:. 
My major disclaimers are:
1. I was not coding a lot in the last years (besides constantly killing my raspberry PI and developing Windows Terminal for the last few months)
2. I was not coding in .net/C# for 8 years
So there is a risk I won't pick the best tools.
If you are not discouraged by those - let's do it :smile:

yaichenbaum:
@Don-Vito Files is a learning experience for all of us 😃
We appreciate all the help and bug fixes from you over the past few days.

tsvietOK:
@yair100 Is it possible to save your CodeMaid config file inside project, so all devs before submitting PR should have to use CodeMaind extension with this config file.

yaichenbaum:
@tsvietOK I can save it inside the project, but I am not sure I want to make having the extension a prerequisite.

Don-Vito:
@tsvietOK  - CodeMaid extension can be problematic with those using VS Code / Rider, etc. That is why external tool usually is more applicable.

Issue #2231, opened 2020-11-04
Selecting an item, deselecting it and then selecting it again triggers rename


XPoppyX:
When you want to rename with double click you must first select the item and then click again. But if you cancel and try to do it again with **the same item**, the rename function starts with a single click.

As you can see in the gif, in the first attempt it works correctly, the rename function starts in the second click after selecting first.
**But from the second time, it starts with a single click**

![rename3](https://user-images.githubusercontent.com/61259627/98165025-ed158a80-1ec3-11eb-93fc-8d9e2cbf131b.gif)

yaichenbaum:
@XPoppyX Is this still an issue in v0.21.1?

generalguy41:
@yaichenbaum I can reproduce it in 0.21.1.0

XPoppyX:
@yaichenbaum yes, it is still an issue with the latest changes to "master"

yaichenbaum:
@XPoppyX This issue should be solved with #2813.

PR #2762, opened 2020-12-30
Prevent rename after single click

Don-Vito:
Closes: https://github.com/files-community/Files/issues/2231

Disclaimer: I really hate this solution, and myself.. for not finding something better :smile:

Root cause:
* When the user clicks outside of the AllView datagrid we manually clean the selection
* But from the datagrid perspective the "current cell" did not change
* Hence clicking on this "current cell" again triggers editing
(as per data grid specification)

I didn't find any reasonable way to convince the datagrid that the current
slot is invalid, hence managed this manually by "remembering" the click
outside of the grid. If such click was performed we consider there is
no current cell, unless the user actively clicks on one of the cells.


gave92:
@Don-Vito thanks for this. With this PR all the stuff related to `GenericFileBrowser.xaml.cs -> tapDebounceTimer` can probably be remove

gave92 (commented):
@Don-Vito thanks for this. With this PR all the stuff related to [GenericFileBrowser.xaml.cs -> tapDebounceTimer](https://github.com/files-community/Files/pull/2762/files#diff-84c054f6ded97591cf9ef1d64a6c0d5348723d507fe5b89d4d8e80e8d9d59a34R275) can probably be removed (it was another ugly solution for this issue)

Don-Vito:
@gave92 - I would avoid nominating this PR to the beauty contest as well 😆 
I will check if debouncer can be removed.

Probably there is a better solution, or probably we need to extend the data grid / contribute to it.
Since this issue is just a "one of" example. There are more, e.g., if you start shift selection in the grid
and click very fast we identify this as rectangle selection (because there is both mouse move and the button is pressed).
Our rectangle selection resets the selected items, effectively canceling the shift selection of the grid.

This could be avoided if we would extend the grid and implement the rectangle selection and current cell deselection "natively".
WDYT?

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -262,34 +259,6 @@ private void AllView_PreparingCellForEdit(object sender, DataGridPreparingCellFo
                 return;
             }
 
-            // Only cancel if this event was triggered by a tap
-            // Do not cancel when user presses F2 or context menu
-            if (e.EditingEventArgs is TappedRoutedEventArgs)
-            {
-                if (AppSettings.OpenItemsWithOneclick)
```
gave92, line 255: This is still neded I think, try testing with `OpenItemsWithOneclick` option enabled

Don-Vito: Sorry

Don-Vito: @gave92 - before I fix this, I never worked single-click mode. Is there an option at all to rename based on mouse interaction?

gave92: In single-click mode rename is only with "F2" or mouse right click (context menu) <- confirm @yair100?

Don-Vito: @gave92 - in this case I am not sure why we need this logic in cell editing as it should be already handled in ItemClick but I guess it was added for reason.

gave92: The DataGrid would put the cell in rename mode, can you point me to the ItemClick you are referring to? It should be only in GridView not Listview

Don-Vito: I was actually referring to `AllView_ItemPress`. 
After few minutes of playing I was able to get to PrepairingEdit in one-click mode.. but it might originate from some other defect...

Don-Vito: @gave92 - I think there are several issues here:
1. `AllView_ItemPress` should mark event as handled if it decides to invoke ItemClick, so it won't get routed to the Grid  (to avoid editing from being invoked).
2. There RectangleSelection creates another issue - it unregisters `All_ViewItemPress` and.. it is too sensitive and is started even if from my perspective I didn't move the mouse after clicking.  So sometimes when I click, the rectangle selection kicks in, unregisters AllView_ItemPressed and nothing happens upon click.

Don-Vito: I can return the protection, to edit - but it should never be invoked in normal conditions.

gave92: > I think there are several issues here

Mouse/Pointer/Keyboard event handling is a mess, yeah
1- Ok, if it prevents AllView_PreparingCellForEdit from firing in case of single click mode, then we can remove the whole code section as you did
2- Selection rectangle's SelectionStarted should be only called if the mouse was moved at least a bit

yaichenbaum: > In single-click mode rename is only with "F2" or mouse right click (context menu) <- confirm @yaichenbaum?

That's correct.

Don-Vito: @gave92 - I returned the protection and I am changing the sensitivity of selection in a separate PR

duke7553: @Don-Vito @gave92 Looking at the [docs](https://docs.microsoft.com/en-us/windows/communitytoolkit/controls/datagrid_guidance/editing_inputvalidation#editing-methods-and-events), it is actually unnecessary to handle the `PreparingCellForEdit` event. We should evaluate whether handling the `BeginningEdit` event will suffice for our special processing since the `BeginningEdit` event does occur **before** a cell or row enters editing mode instead of when a cell enters the editing mode.

Basically since `PreparingCellForEdit` is technically a method as per the docs, it may be bad practice to override it with our own logic.

duke7553 (commented):
@Don-Vito @gave92 Looking at the docs, it is actually unnecessary to handle the PreparingCellForEdit event. We should evaluate whether handling the BeginningEdit event will suffice for our special processing since the BeginningEdit event does occur before a cell or row enters editing mode instead of when a cell enters the editing mode.

Basically since PreparingCellForEdit is technically a method as per the docs, it may be bad practice to override it with our own logic.

duke7553 (dismissed):

Don-Vito:
> 
> 
> @Don-Vito @gave92 Looking at the docs, it is actually unnecessary to handle the PreparingCellForEdit event. We should evaluate whether handling the BeginningEdit event will suffice for our special processing since the BeginningEdit event does occur before a cell or row enters editing mode instead of when a cell enters the editing mode.
> 
> Basically since PreparingCellForEdit is technically a method as per the docs, it may be bad practice to override it with our own logic.

Sounds really good - I can try to move it! Though, probably, it is better to do it in a separate PR, once this one gets merged - to preserve the scope of this one.

gave92:
@Don-Vito sry I was wrong the debounce is still needed. To reproduce:
- select an item
- click outside datagrid
- double click on the file name -> instead of opening it will start renaming

Don-Vito:
@gave92 - the double click rename reproduced to me randomly. But after rebasing and moving the logic to BeginningEdit, it doesn't happen to me anymore.. Can you please recheck?

gave92:
@Don-Vito looks like that issue is resolved 👍 I have a new one though: when single click mode is enabled, sometimes it doesn't work and opening an item requires two clicks, is this something you can reproduce?

Don-Vito:
@gave92 - I will try to repro. But probably we should merge the "calm down rs" first, as it sometimes "steals" clicks by unregistering itempressed

gave92:
> I really hate this solution

Apparently there is an alternative solution to the rename issue. The following code will reset the current datagrid cell.
```cs
uiElement.GetType().GetProperty("CurrentColumnIndex", BindingFlags.NonPublic | BindingFlags.Instance).SetValue(uiElement, -1);
```
So if you added it [here](https://github.com/files-community/Files/blob/master/Files/UserControls/Selection/RectangleSelection_DataGrid.cs#L173) it should work without the need of RootGrid_Tapped/isAllViewCurrentSlotValid/debounce timer

Not sure if you like this better :)

Don-Vito:
> > I really hate this solution
> 
> Apparently there is an alternative solution to the rename issue. The following code will reset the current datagrid cell.
> ```cs
> typeof(DataGrid).GetProperty("CurrentColumnIndex", BindingFlags.NonPublic | BindingFlags.Instance).SetValue(uiElement, -1);
> ```
> So if you added it [here](https://github.com/files-community/Files/blob/5d93f493869902f3e6e0e3e5ada9acac6412ed4c/Files/UserControls/Selection/RectangleSelection_DataGrid.cs#L184) it should work without the need of RootGrid_Tapped/isAllViewCurrentSlotValid/debounce timer
> 
> Not sure if you like this better :)

Cool! This should be much better. I will test it later on. Wondering if we can do it with CurrentItem/Slot.

Don-Vito:
Closing this one in favor of https://github.com/files-community/Files/pull/2813

PR #2764, opened 2020-12-30
Fixed a couple of issues with detecting and saving terminal profiles

### LINKED ISSUE #1441: Terminals with same path can't be saved as default terminal in settings
**Describe the bug**
If have 2 terminals with same path in terminal.json, the second terminal can't be saved as default terminal in settings.

**To Reproduce**
1.  Config 2 terminals with same path in terminal.json like that
```json
{
  "version": 1,
  "defaultTerminalPath": "wt.exe",
  "terminals": [
    {
      "name": "Windows Terminal",
      "path": "wt.exe",
      "arguments": "-d .",
      "icon": ""
    },
    {
      "name": "WSL",
      "path": "wt.exe",
      "arguments": "-d . wsl.exe",
      "icon": ""
    }
  ]
}
```
2. Go to 'Settings -> Preference'.
3. Set 'Terminals' to 'WSL'.
3. Leave settings.
4. Go to 'Settings -> Preference'.
5. The 'Terminals' became 'Windows Terminal' and can't be saved as 'WSL'.

**Expected behavior**
The 'Terminals' should be 'WSL'.

 - OS Version: [Windows 10 2004 19041.329]
 - App version: [v0.11.0]

### LINKED ISSUE #2009: Windows Terminal Preview is not recognized by Files
**Describe the bug**
Basically, Files doesn't recognize WT Preview. You may try to manyally add it to the terminal.json file but it gets removed on startup. Even changing the title from "Windows Terminal" to "Windows Terminal Preview" still makes Files remove that JSON object all together.

**To Reproduce**
1. Go to Settings > Preferences
2. Edit the file to add Windows Terminal Preview
3. Go back to Settings > Preferences after a restart of the app
4. See that the entry you added is deleted and not there.

**Expected behavior**
The expected behavior is that my entry for Windows Terminal Preview would continue to be there after a restart.

**Desktop:**
 - OS Version: Windows 10 1909 18363.1082
 - App version: v0.15.4.0

**Additional context**
 The workaround I have found for this is to change the command for the Windows Terminal JSON object from `wt.exe` to just `wt`.
EDIT:
For Clarity, this bug is only present when **ONLY** Windows Terminal Preview is installed and not also the Stable version.

yair100:
Fixes #2009
Fixes #1441

### DIFF
--- a/Files/Controllers/TerminalController.cs
+++ b/Files/Controllers/TerminalController.cs
@@ -1,5 +1,6 @@
 ﻿using Files.DataModels;
 using Files.Filesystem;
+using Files.Helpers;
 using Newtonsoft.Json;
 using System;
 using System.IO;
@@ -115,12 +116,29 @@ namespace Files.Controllers
                 Icon = ""
             };
 
-            bool isWindowsTerminalAddedOrRemoved = await Model.AddOrRemoveTerminalAsync(windowsTerminal, "Microsoft.WindowsTerminal_8wekyb3d8bbwe");
-            bool isFluentTerminalAddedOrRemoved = await Model.AddOrRemoveTerminalAsync(fluentTerminal, "53621FSApps.FluentTerminal_87x1pks76srcp");
-            if (isWindowsTerminalAddedOrRemoved || isFluentTerminalAddedOrRemoved)
+            bool isWindowsTerminalInstalled = await PackageHelper.IsAppInstalledAsync("Microsoft.WindowsTerminal_8wekyb3d8bbwe");
+            bool isWindowsTerminalPreviewInstalled = await PackageHelper.IsAppInstalledAsync("Microsoft.WindowsTerminalPreview_8wekyb3d8bbwe");
+            bool isFluentTerminalInstalled = await PackageHelper.IsAppInstalledAsync("53621FSApps.FluentTerminal_87x1pks76srcp");
+
+            if (isWindowsTerminalInstalled || isWindowsTerminalPreviewInstalled)
             {
-                SaveModel();
+                Model.AddTerminal(windowsTerminal);
             }
+            else
+            {
+                Model.RemoveTerminal(windowsTerminal);
+            }
+
+            if (isFluentTerminalInstalled)
+            {
+                Model.AddTerminal(fluentTerminal);
+            }
+            else
+            {
+                Model.RemoveTerminal(fluentTerminal);
+            }
+
+            SaveModel();
         }
 
         public void SaveModel()
--- a/Files/DataModels/TerminalFileModel.cs
+++ b/Files/DataModels/TerminalFileModel.cs
@@ -10,15 +10,15 @@ namespace Files.DataModels
         [JsonProperty("version")]
         public int Version { get; set; }
 
-        [JsonProperty("defaultTerminalPath")]
-        public string DefaultTerminalPath { get; set; }
+        [JsonProperty("DefaultTerminalName")]
+        public string DefaultTerminalName { get; set; }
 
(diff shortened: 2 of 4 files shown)


Don-Vito:
@yair100 - I considered exactly this when looking at https://github.com/files-community/Files/issues/2009. My major issue was how to preserve backward compatibility (for those who already configured custom settings). What is your plan around this?

BTW when working on Terminal itself we have introduced a Guid for everything (profile, color scheme, etc.). While this complicates a stuff a bit, it is highly future compatible.

yair100:
> @yair100 - I considered exactly this when looking at #2009. My major issue was how to preserve backward compatibility (for those who already configured custom settings). What is your plan around this?
>

@Don-Vito Custom settings from previous versions are fully compatible and they are not overridden by this change.

Don-Vito:
@yair100 - oh.. I was sure that this change will forget what users configured in DefaultTerminalPath. I guess I still don't fully understand the flow.. sorry..

yair100:
> @yair100 - oh.. I was sure that this change will forget what users configured in DefaultTerminalPath. I guess I still don't fully understand the flow.. sorry..

It's posible that the default terminal itself will reset to `cmd` but I would consider that ok in this case. The main thing is for the terminal list to carry over properly and that appears to work corectly.

Files/Controllers/TerminalController.cs
```diff
@@ -115,12 +116,29 @@ public async Task GetInstalledTerminalsAsync()
                 Icon = ""
             };
 
-            bool isWindowsTerminalAddedOrRemoved = await Model.AddOrRemoveTerminalAsync(windowsTerminal, "Microsoft.WindowsTerminal_8wekyb3d8bbwe");
-            bool isFluentTerminalAddedOrRemoved = await Model.AddOrRemoveTerminalAsync(fluentTerminal, "53621FSApps.FluentTerminal_87x1pks76srcp");
-            if (isWindowsTerminalAddedOrRemoved || isFluentTerminalAddedOrRemoved)
+            bool isWindowsTerminalInstalled = await PackageHelper.IsAppInstalledAsync("Microsoft.WindowsTerminal_8wekyb3d8bbwe");
```
duke7553, line 119: giphy

lukeblevins (approved):
LGTM :shipit:

Don-Vito:
> 
> 
> > @yair100 - oh.. I was sure that this change will forget what users configured in DefaultTerminalPath. I guess I still don't fully understand the flow.. sorry..
> 
> It's posible that the default terminal itself will reset to `cmd` but I would consider that ok in this case. The main thing is for the terminal list to carry over properly and that appears to work corectly.

Yeah.. This is what I was addressing.. I am not familiar with the community - if this doesn't get us yelled at, then it is fine :smile:.

Last time suggesting to use GUIDs (as tomorrow the users will want to change the name of the terminal, or something weird like that), and another compatibility change might be required..

yaichenbaum:
@Don-Vito We can definitely make that change, in fact, we should probably try do that before we ship v1. Is this something you would like to work on?

Don-Vito:
I can definitely do that! I will open the tasks for what we discussed and then let's assign them to me! :blush:

Don-Vito:
> 
> 
> @Don-Vito We can definitely use GUIDs, in fact, we should probably try do that before we ship v1. Is this something you would like to work on?

Another question - is there a roadmap for 1.0 (not necessarily with dates) published somewhere?

yaichenbaum:
@Don-Vito The goal is the have v1 ready for testing this week. We don't have a timeline since everyone is working on their own time, what you see today is basically feature complete for v1.

PR #2768, opened 2020-12-30
Calm down rectangle selection

Don-Vito:
It seems that our Rectangle Selection is not aware of the fact
that brain and fingers are not connected.
So if you click and move a mouse a tiny bit, it already gets triggered.
This is something not really pleasant, as it might cancel shift selection
of the grid, or to unregister `AllView_ItemPressed`, actually preventing
activation in one-click mode.

This PR verifies that the mouse has moved sufficiently to indicate
rectangle selection (the value is hard-coded right now - please
let me know if you have better ideas).

BTW, probably the better solution is radius, rather than checking
the dimensions, but I doubt the ROI :smile:


Files/UserControls/Selection/RectangleSelection_DataGrid.cs
```diff
@@ -37,17 +37,28 @@ public RectangleSelection_DataGrid(DataGrid uiElement, Rectangle selectionRectan
 
         private void RectangleSelection_PointerMoved(object sender, PointerRoutedEventArgs e)
         {
+            if (scrollBar == null)
```
d2dyno1, line 40: More fancier way 😜 `if (scrollBar is not ScrollBar)`

Don-Vito: This definitely looks more philosophical this way :smile:
I preserved the original semantics (that were in line 48)

gave92: @d2dyno1 how could you! you are missing the brackets, please follow Files code-style!! 😂

d2dyno1: NO_I_WILL_NOT

gave92:
Also here this issue is back #2749?

Don-Vito:
> 
> 
> Also here this issue is back #2749?

@gave92 - doesn't happen to me. Also see no reason why it would. 🤔

Don-Vito:
@gave92 - fixed.. by rebasing to master :smile:

gave92:
@Don-Vito uops xD I just assumed it already was. Thanks

gave92 (approved):
@Don-Vito great work 👍

Thomas1664:
> BTW, probably the better solution is radius, rather than checking
> the dimensions, but I doubt the ROI 😄

@Don-Vito If you want radius you can use the vector-distance-formula like so:
```c#
private double CalcVectorDistance(int x1, int y1, int x2, int y2)
{
    return Math.Sqrt(Math.Pow((x2 - x1), 2) + Math.Pow((y2 - y1), 2));
}
```

PR #2777, opened 2020-12-30
Application crashes when clicking sorting by name twice on a plain drive (no subfolder)

### LINKED ISSUE #2775: Application crashes when clicking sorting by name twice on a plain drive (no subfolder)
**Describe the bug**
When having navigated directly onto a drive (like C:, D:, ...) the whole file manager crashes when I click on the "Name" tab in the header twice in order to sort the files by name in ascending and then descending order.

**To Reproduce**
1. Go to any drive (like C:). It doesn't matter whether you do this by using the Drive button in the sidebar or if you use the one on the main page
2. In the header of the file contents table click on "Name". Your drive contents will get sorted.
3. Click again on the "Name" header button....
4. ...Visual Studio will stop the debugging of the application because in the file "ViewModels/FolderSettingsViewModel.cs" an exception is thrown in the method: `WriteLayoutPreferencesToSettings(string folderPath, LayoutPreferences prefs)`. Exception is shown in Screenshot below.

**Expected behavior**
I would expect Files to simply change the sorting order of my files and folders on my (external) hard drive.

**Screenshots**
image
Despite using an english version of Windows and Visual Studio somehow the error still seems to be in german... I am really sorry for that it translates into: "The given path is invalid".

 - OS Version: [e.g. Windows 10 20H2 19042.685]
 - App version: [v0.11.0]??? Current master branch on files-community/Files

**Additional context**
The issue seems to be that whenever a naked drive is sorted instead of some folder the path ends with '\\' so e.g. "C:\\". If a folder is opened up there would be no such trailing '\\', e.g. "C:\someFolder". Every path which ends with a trailing backslash or encoded as '\\' is not a valid path?!
A quick fix seems to modify the property "public SortDirection DirectorySortDirection" as follows:
`
        public SortDirection DirectorySortDirection
        {
            get => LayoutPreference.DirectorySortDirection;
            set
            {
                if (SetProperty(ref LayoutPreference.DirectorySortDirection, value, nameof(DirectorySortDirection)))
                {
                    // Workaround for the case that the directory path is a drive. Then the trailing '\' needs to be removed
                    String workingDirectoryPath = associatedInstance.FilesystemViewModel.WorkingDirectory.EndsWith('\\') ? associatedInstance.FilesystemViewModel.WorkingDirectory.TrimEnd('\\') : associatedInstance.FilesystemViewModel.WorkingDirectory;
                    UpdateLayoutPreferencesForPath(workingDirectoryPath, LayoutPreference);
                    SortDirectionPreferenceUpdated?.Invoke(this, new EventArgs());
                }
            }
        }
`
I do know the codebase not that well yet. But I think a better solution might be to modify the method "public async Task<FilesystemResult> SetWorkingDirectoryAsync(string value)" in "ItemViewModel.cs" as it sets the directory path used in "DirectorySortDirection"... Is this approach correct? Was somebody able to reproduce the issue?

**Log file**
Please post the log file here so that we can understand your problem better. You can access it from Settings->About->Open log location.
[debug.txt](https://github.com/files-community/Files/files/5755725/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5755726/debug_fulltrust.txt)

### LINKED ISSUE #2814: Application crashes when clicking on drives without a letter
**Describe the bug**
When connected to my PC, my mobile phone also gets enumerated as a disk drive. But when I click on it the application crashes because the path is invalid?! My assumption is that this might be because my mobile phone does not a have a drive letter assigned.

**To Reproduce**
1. Connect an Android mobile with your PC
2. A new "stick drive" appears
3. Try to access it by clicking on it
4. See error as it can be seen on the screenshot below.

**Expected behavior**
No crash and an empty folder should appear (as it does in explorer)

**Screenshots**
image
The error says that the given pathname is invalid.

 - OS Version: [e.g. Windows 10 20H2 19042.685]
 - App version: Current master branch

**Additional context**
As it seems to be a path problem as well it might be connected with issue #2777 

**Log file**
[debug.txt](https://github.com/files-community/Files/files/5759268/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5759269/debug_fulltrust.txt)

heistema:
Fixes #2775
Fixes #2814


gave92:
@heistema thanks for this PR
> That's because "TrimEnd(..)" returns null if not successful

Are you sure about this? Doesn't seem to happen in my testing, is this something you noticed happening (if so with which string for example?) or is it something from the documentation? (we use TrimEnd() in other places, this is omething important to confirm)

gave92 (commented):
Can you remove the TrimEnd from [here](https://github.com/files-community/Files/pull/2777/files#diff-68a49b148b94057c697cfb2ee2afeb34cb5f1f899f56f87a75cb6d7db61ee4dfR48) and move it to `GetLayoutPreferencesForPath`? So we keep the Save and Get functions coherent

gave92:
@heistema sadly the fix is not enough, the app *also* crashes when the path *starts* with a backslash (network share, phone storage)
So it's necesasry something like the following:
```cs
if (App.AppSettings.AreLayoutPreferencesPerFolder)
{
	if (!WriteLayoutPreferencesToAds(folderPath.TrimEnd('\\'), prefs)) ---> must use TrimEnd() here
	{
		WriteLayoutPreferencesToSettings(folderPath.Replace('\\', '_'), prefs); ---> replace all backslashed to underscore
	}
}
```

heistema:
I also found other issues, especially concerning drives... They might be connected as well. I am currently unable to answer further on the issue (no tech except mobile phone). But I really would like to tomorrow

Holen Sie sich Outlook für Android<https://aka.ms/ghei36>

________________________________

heistema:
I posted an other issue which also seems to be path related. The question is shall I correct this bug by making it as suggested by gave92 or shall I introduce a new class like PathSanitizer where all our paths have to run through?

The latter might also be a fix for #2814 .

I will also give TrimEnd another try maybe gave92 was right with his first suggestion...

gave92:
@heistema Looking at #2814, I'd still suggest the above approach which will (should) fix both #2775 and #2814:

In `GetLayoutPreferencesForPath`:
```cs
if (App.AppSettings.AreLayoutPreferencesPerFolder)
{
    var layoutPrefs = ReadLayoutPreferencesFromAds(folderPath.TrimEnd('\\')); ---> must use TrimEnd() here as it needs a valid filepath
    return layoutPrefs ?? ReadLayoutPreferencesFromSettings(folderPath.Replace('\\', '_')); ---> replace all backslashes to underscore
}

In `UpdateLayoutPreferencesForPath`:
```cs
if (App.AppSettings.AreLayoutPreferencesPerFolder)
{
    if (!WriteLayoutPreferencesToAds(folderPath.TrimEnd('\\'), prefs)) ---> must use TrimEnd() here as it needs a valid filepath
    {
        WriteLayoutPreferencesToSettings(folderPath.Replace('\\', '_'), prefs); ---> replace all backslashes to underscore
    }
}
```

heistema:
You're right. I updated the solution in my repository... It should be attached to this PR?!

gave92:
@heistema looks good 👍 
Can you remove the TrimEnd from [here](https://github.com/files-community/Files/blob/master/Files/ViewModels/FolderSettingsViewModel.cs#L48)? With this changes it's not needed anymore.
Then we're ready to merge this.

heistema:
@gave92 And it's done...

gave92 (approved):

PR #2791, opened 2020-12-31
Fixed suggested keyboard shortcut in navbar menu

gave92:
The add pane menu in the navigation bar currently displays "ctrl + right" to open the right pane, the correct shortcut is instead *alt + shift + "+"* (or *ctrl + shift + right*)

image


yaichenbaum (approved):

yaichenbaum:
@gave92 Good catch!

PR #2798, opened 2020-12-31
Update LICENSE

yaichenbaum:


duke7553 (commented):
Can we make this change without receiving approval from every previous contributor?

yaichenbaum:
> Can we make this change without receiving approval from every previous contributor?

I think we can since these are compatible with each other, however, we need to confirm this.

duke7553 (approved):

PR #2813, opened 2021-01-01
Ensure current cell is deselected when clicking outside of DataGrid

Don-Vito:
Closes: https://github.com/files-community/Files/issues/2231

Root cause:
* When the user clicks outside of the AllView datagrid we manually clean the selection
* But from the datagrid perspective the "current cell" did not change
* Hence clicking on this "current cell" again triggers editing
(as per data grid specification)

Some implementation notes:
* To deselect the current cell we are forced to use reflection
as the relevant method is internal
* The PR removes "edit protections" that are not needed anymore
* The remaining protections are moved to BeginningEdit

### DIFF
--- a/Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
+++ b/Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
@@ -217,50 +217,18 @@ namespace Files.Views.LayoutModes
         private TextBox renamingTextBox;
 
         private DispatcherTimer tapDebounceTimer;
-
-        private void AllView_PreparingCellForEdit(object sender, DataGridPreparingCellForEditEventArgs e)
+        private void AllView_BeginningEdit(object sender, DataGridBeginningEditEventArgs e)
         {
             if (ParentShellPageInstance.FilesystemViewModel.WorkingDirectory.StartsWith(AppSettings.RecycleBinPath))
             {
                 // Do not rename files and folders inside the recycle bin
-                AllView.CancelEdit(); // Cancel the edit operation
+                e.Cancel = true;
                 return;
             }
+        }
 
-            // Only cancel if this event was triggered by a tap
-            // Do not cancel when user presses F2 or context menu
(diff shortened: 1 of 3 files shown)


yaichenbaum (dismissed):

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -217,50 +217,18 @@ private void ViewModel_PropertyChanged(object sender, PropertyChangedEventArgs e
...
                 // Do not rename files and folders inside the recycle bin
-                AllView.CancelEdit(); // Cancel the edit operation
-                return;
-            }
-
-            // Only cancel if this event was triggered by a tap
-            // Do not cancel when user presses F2 or context menu
-            if (e.EditingEventArgs is TappedRoutedEventArgs)
-            {
-                if (AppSettings.OpenItemsWithOneclick)
```
gave92, line 234: This stuff is still needed: if OpenItemsWithOneclick is enabled and you click twice on the filename it starts renaming

gave92 (changes requested):

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
Don-Vito: @gave92 - I can add this back, but it really should not happen - we should never enter begin the edit as ItemPressed should perform the navigation.

gave92: @Don-Vito i know but i can still trigger it 100% of times :(

Don-Vito:
@gave92 - do you mean it happens to you know, after I returned the protection as well? As I cannot reproduce it.

gave92:
> @gave92 - do you mean it happens to you know, after I returned the protection as well? As I cannot reproduce it.

Nono, I meant that I was able to trigger it before. I'll test your latest commit but the issue should be solved 👍

gave92 (approved):

PR #2817, opened 2021-01-01
Select first DataGrid column when pressing F2

gave92:
This PR sets the file name column as the active DataGrid column on folder navigation, so pressing "F2" works without the need to click the list with the mouse first.
(thanks @Don-Vito for the SetCurrentCellCore method)

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -91,7 +91,7 @@ namespace Files
                 if (value != selectedItems)
                 {
                     selectedItems = value;
-                    if (selectedItems.Count == 0)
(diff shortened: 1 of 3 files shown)


tsvietOK:
@gave92 When selecting an item via date/other column and pressing F2 still nothing happens, can you check it from your side?

gave92:
@tsvietOK that's expected. This PR only selects the first column on navigation, so it's the default one.

tsvietOK:
@gave92 Ok, so what is the purpose of your changes and why do we need to select first column on navigation?

gave92:
@tsvietOK I've just changed this to do as you suggested :)

yaichenbaum (approved):

PR #2829, opened 2021-01-03
Fix crash when clipboard changes before InteractionViewModel is set

Don-Vito:
During debugging I often encounter the problem when copy-paste crashes the app.
This happens when InteractionViewModel is not initialized.
I didn't want to postpone the registration of Clipboard changed,
as in the future we might want to add handling even if InteractionViewModel
is still not initialized. So simply added a safe-check.

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -176,6 +176,11 @@ namespace Files
 
         private void Clipboard_ContentChanged(object sender, object e)
         {
+            if (App.InteractionViewModel == null)
+            {
+                return;
+            }
+
             try
             {
                 // Clipboard.GetContent() will throw UnauthorizedAccessException


gave92 (approved):

gave92:
Alternative would be to initialize InteractionViewModel in App constructor. But I don't now if that would have impact on the perceived app startup time, so this is a good solution 👍

yaichenbaum:
@gave92 Would it be posible to move the clipboard changed event to InteractionViewModel?

gave92:
> During debugging I often encounter the problem when copy-paste crashes the app.
> This happens when InteractionViewModel is not initialized.

How can this happen? InteractionViewModel is initialized before anything is shown, if you see enough that you can copy-paste from Files, InteractionViewModel should have long been initialized..

Don-Vito:
> 
> 
> > During debugging I often encounter the problem when copy-paste crashes the app.
> > This happens when InteractionViewModel is not initialized.
> 
> How can this happen? InteractionViewModel is initialized before anything is shown, if you see enough that you can copy-paste from Files, InteractionViewModel should have long been initialized..

@gave92 - the InteractionViewModel  is initialized during OnLauched, while the handler is registered in ctor. When debugging your startup is slow enough to crash if you copy something in VS.

gave92:
If it's during startup, and *not* while copying stuff inside Files, it makes sense 👍

@yair100 Registering to Clipboard.Content_Changed inside the InteractionViewModel constructor would also work I guess, but a similar check would be necessary anyway when calling Clipboard_ContentChanged(null, null) from CoreWindow_Activated

PR #2836, opened 2021-01-04
Automatically refresh view when an item visibility is changed

### LINKED ISSUE #2785: When marking a folder as hidden the folder view does not update itself automatically
**Describe the bug**
When any folder is set to a hidden folder by using the properties window the folder view does not update itself automatically. So after closing the properties window the folder icon still is shown as if the folder is being visible.

**To Reproduce**
1. Go to any folder and right click on it
2. Click on 'Properties' and enable the checkbox on 'Hidden'
3. Click on 'OK'. This closes the dialog and applies the hidden property.
4. As can be seen the folder icon still gives the impression of the folder being a visible one. But in fact it's not. Manual reload (e.g. by changing folder back and forth) solves the issue

**Expected behavior**
Right after closing the properties dialog the folder icon should change to its respective transparent brother

**Screenshots**

 - OS Version: [e.g. Windows 10 20H2 19042.685]
 - App version: Current master

**Additional context**
None. As mentioned above a manual refresh of the folder view solves the problem.

**Log file**
[debug.txt](https://github.com/files-community/Files/files/5756177/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5756178/debug_fulltrust.txt)

heistema:
The PR works for single files only so far... Still no refresh when setting multiple files to hidden at once (or vice versa).

Another issue is also not fixed yet. Deletion of hidden files. The .NET library cannot directly delete hidden files and / or folders, we would need WIN32 API for that. But we can remove the hidden flag prior deletion ;-).

Fixes #2785

### DIFF
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -870,14 +870,15 @@ namespace Files.Interacts
             return false;
         }
 
-        public void SetHiddenAttributeItems(ListedItem item, bool isHidden)
+        /// <summary>
+        /// Set a single file or folder to hidden or unhidden an refresh the
+        /// view after setting the flag
+        /// </summary>
+        /// <param name="item"></param>
+        /// <param name="isHidden"></param>
+        public void SetHiddenAttributeItem(ListedItem item, bool isHidden)
         {
-            // When multiple files are selected to be made hidden at once item gets null
-            if (item != null)
-            {
-                item.IsHiddenItem = isHidden;
-            }
-            // The view area needs a refresh when the hidden flag had been set
+            item.IsHiddenItem = isHidden;
             AssociatedInstance.ContentPage.ResetItemOpacity();
         }
 
--- a/Files/Interacts/Interaction.cs
+++ b/Files/Interacts/Interaction.cs
@@ -870,22 +870,11 @@ namespace Files.Interacts
             return false;
         }
 
-        public bool SetHiddenAttributeItems(ListedItem item, bool isHidden)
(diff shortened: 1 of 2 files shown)


Files/Interacts/Interaction.cs
```diff
@@ -870,6 +870,24 @@ public async Task<bool> RenameFileItemAsync(ListedItem item, string oldName, str
             return false;
         }
 
+        public bool SetHiddenAttributeItems(ListedItem item, bool isHidden)
```
gave92, line 873: I believe this can be simplified to:
```cs
public void SetHiddenAttributeItems(ListedItem item, bool isHidden)
{
	item.IsHiddenItem = isHidden;
	AssociatedInstance.ContentPage.ResetItemOpacity();
}
```

heistema: Yes, that's indeed a huge simplification ;-)
I updated it accordingly... Deletion of hidden files and folders also seems to work (but I think that was not me fixing this behavior)

gave92:
@heistema thanks for this PR. Looks like the branch got a little messed up (49 file modified), could you check?

heistema:
@gave92 Yes, I just pulled main into my branch... I took it back.

gave92 (approved):

gave92 (approved):

yaichenbaum:
@heistema Great work!

yaichenbaum:
@heistema can you take a look at #7625?

PR #2844, opened 2021-01-05
Improved widgets loading priority

yaichenbaum:


Files/UserControls/Widgets/DrivesWidget.xaml
```diff
@@ -132,6 +132,7 @@
                                 </Grid.ColumnDefinitions>
                                 <FontIcon
                                     Grid.Column="0"
+                                    x:Phase="2"
```
gave92, line 135: I believe x:Phase makes more sense for big lists that get scrolled, drive list is not big and all the items are always visible at once on screen. Does adding x:Phase make sense here?
[We should really to do it on the GridViewBrowser/GenericFileBrowser template though]

Files/UserControls/Widgets/RecentFiles.xaml
```diff
@@ -102,33 +103,41 @@
...
+                                        x:Phase="2"
                                         Source="{x:Bind FileImg}"
-                                        Stretch="UniformToFill"
-                                        Visibility="{x:Bind FileIconVis}" />
+                                        Stretch="UniformToFill" />
                                 </Grid>
                                 <TextBlock
                                     Grid.Column="1"
                                     VerticalAlignment="Center"
+                                    x:Phase="1"
```
gave92, line 132: Any element without  x:Phase is considered "Phase 0". So you can probably remove the phase from TextBlock and decrease the phase on Images to "1".

Files/UserControls/Widgets/DrivesWidget.xaml
yaichenbaum: The drives list can get a little long especially when there are a lot of network drives 😃. We should work on this for the GridViewBrowser/GenericFileBrowser templates as well.

Files/UserControls/Widgets/DrivesWidget.xaml
```diff
@@ -144,17 +145,20 @@
                                         x:Name="ItemLocationName"
                                         Margin="0,0,0,8"
                                         VerticalAlignment="Center"
+                                        x:Phase="1"
```
gave92, line 148: Could you give this the same treatment as RecentFileList? Remove "Phase 1" and drecrease "Phase 2"?

gave92 (approved):

PR #2854, opened 2021-01-05
Improved the layout flyout to include common display options

yaichenbaum:
The updated flyout lets users quickly toggle common options such as the layout mode, the option to show/hide file extensions and the option to show or hide hidden items.
image

### DIFF
--- a/Files/UserControls/StatusBarControl.xaml
+++ b/Files/UserControls/StatusBarControl.xaml
@@ -143,7 +143,7 @@
                                     ToolTipService.ToolTip="Details View">
                                     <FontIcon
                                         FontFamily="{StaticResource CustomGlyph}"
-                                        FontSize="15"
+                                        FontSize="16"
                                         Glyph="&#xF101;" />
                                 </Button>
                                 <Button
@@ -155,7 +155,7 @@
                                     ToolTipService.ToolTip="Tiles View">
                                     <FontIcon
                                         FontFamily="{StaticResource CustomGlyph}"
-                                        FontSize="15"
+                                        FontSize="16"
                                         Glyph="&#xF100;" />
                                 </Button>
                                 <Button
@@ -167,7 +167,7 @@
                                     ToolTipService.ToolTip="Grid View (Small)">
                                     <FontIcon
                                         FontFamily="{StaticResource FluentUIGlyphs}"
-                                        FontSize="15"
+                                        FontSize="16"
                                         Glyph="&#xEBA1;" />
                                 </Button>
                                 <Button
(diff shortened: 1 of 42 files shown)


Files/UserControls/StatusBarControl.xaml
```diff
@@ -99,65 +99,109 @@
...
+                                Text="Layout Mode" />
+                            <StackPanel Orientation="Horizontal" Spacing="4">
+                                <Button
+                                    x:Uid="StatusBarControlDetailsViews"
+                                    Height="32"
+                                    Command="{x:Bind FolderSettings.ToggleLayoutModeDetailsView}"
+                                    CornerRadius="2"
+                                    Style="{StaticResource ToolBarButtonStyle}"
+                                    ToolTipService.ToolTip="Details View">
+                                    <StackPanel Orientation="Horizontal">
```
gave92, line 144: Extra StackPanel?

gave92:
Looks good. Have you considered using ToggleSwitch instead of Checkboxes? This would be the only case for which we use checkboxes for boolean settings

yaichenbaum:
@gave92 Thanks for catching those! They should be resolved with my latest commit.

mdtauk:
> Looks good. Have you considered using ToggleSwitch instead of Checkboxes? This would be the only case for which we use checkboxes for boolean settings

RadioButtons would probably be the correct control behaviour, if only one option can be active at a time.

yaichenbaum:
> > Looks good. Have you considered using ToggleSwitch instead of Checkboxes? This would be the only case for which we use checkboxes for boolean settings
> 
> RadioButtons would probably be the correct control behaviour, if only one option can be active at a time.

@mdtauk In this case the options are separate settings so they should use the toggle switch. There is however a separate issue that someone opened about using radio buttons to indicate the layout mode.

mdtauk:
> > > Looks good. Have you considered using ToggleSwitch instead of Checkboxes? This would be the only case for which we use checkboxes for boolean settings
> > 
> > 
> > RadioButtons would probably be the correct control behaviour, if only one option can be active at a time.
> 
> @mdtauk In this case the options are separate settings so they should use the toggle switch. There is however a separate issue that someone opened about using radio buttons to indicate the current layout mode.

I was specifically talking about the Layout Mode buttons.

gave92:
Yeah it would be nice to have the selected layout mode button highlighted bit we can do it in a different PR

gave92 (approved):

yaichenbaum:
Yeah, that's a nice change but it requires a little more work so it will come in a different PR.

PR #2855, opened 2021-01-05
Changed default sorting method for properties

### LINKED ISSUE #2808: Properties: Titles of detail sections are sorted alphabetically
The property section titles are sorted alphabetically. This makes the **"Image"** section, for example, be positioned after **"Core"**.

### **So what is the problem?**

Well, i think the **"Core"** section should always be in the last place, because **"Image" / "Audio" / "Video"** are more important sections.

I already tested using **A** and **B**, and they were definitely sorted alphabetically

Captura de pantalla 2021-01-01 100758 Captura de pantalla 2021-01-01 100919

winston-de:
Fixes #2808
Also improved syntax on some of my old code


tsvietOK:
Is something changed?
image

tsvietOK (approved):

PR #2864, opened 2021-01-06
fr-fr

boscorelly:
fr-fr


tsvietOK:
@boscorelly Do you want to add another translation or is it ready?

boscorelly:
hi,

i always check for new translations, and i also need to see whre 
non-translated sentences are use to understand them and translate 
correctly :)

tsvietOK (approved):

PR #2880, opened 2021-01-07
Use temporary list for storage folder enumeration

### LINKED ISSUE #2878: Duplicate items in SMB drive after "Cache files and folders for better performance" enabled
**Describe the bug**
While "Cache files and folders for better performance" is enabled (seems default on), the file/folder list in a mounted SMB drive will duplicate after every refresh. 

**To Reproduce**
1. Enable "Cache files and folders for better performance"
2. Open a mounted SMB drive
3. refresh

**Screenshots**
屏幕截图 2021-01-07 162642

 - OS Version: Windows 10 21286
 - App version: 1.0.0.0

jakoss:
Fixes #2878

@gave92 I'm not convinced about the enumeration method right now. Enumerating via file system one by one seems completly unnecessary, and i believe it's inefficient. Is this by design?
image

I believe that it would be the best to create load request for all items or pretty large batches. Operating system should take care about the caching etc.

That being said - are you open for Pull Requests with refactoring? I'm pretty sure i'll have to do some refactoring to implement preemptive caching anyway


gave92:
> I'm not convinced about the enumeration method right now

I'm not convinced it's the best way either. This was done to:
- Immediately show results to user -> can be done anyway by making the first batch "small" and the rest larger
- Handle the UnauthorizedAccessException and FileNotFoundException exceptions -> this check was added after seeing bug reports but I was unable to reproduce. What happens if you ask GetItemsAsync for 100 items and only one throws UnauthorizedAccessException?

jakoss:
The solution might be to do 1by1 enumeration on Exception encounter on batch call, making it optional only for those cases. I'll deal with that during preemptive cache refactoring if you don't mind

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1227,13 +1228,15 @@ private async Task EnumFromStorageFolderAsync()
                 }
                 if (item.IsOfType(StorageItemTypes.Folder))
                 {
-                    await AddFolderAsync(item as StorageFolder, returnformat);
+                    var folder = await AddFolderAsync(item as StorageFolder, returnformat);
+                    tempList.Add(folder);
```
gave92, line 1232: Add: `if (folder != null) tempList.Add(folder);`

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1227,13 +1228,15 @@ private async Task EnumFromStorageFolderAsync()
...
                     ++count;
                 }
                 else
                 {
                     var file = item as StorageFile;
-                    await AddFileAsync(file, returnformat, true);
+                    var fileEntry = await AddFileAsync(file, returnformat, true);
+                    tempList.Add(fileEntry);
```
gave92, line 1239: Add: if (fileEntry != null) tempList.Add(fileEntry);

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1242,14 +1245,23 @@ private async Task EnumFromStorageFolderAsync()
                 }
                 if (count % 300 == 0)
```
gave92, line 1246: Maybe do the same as standard enumeration: `count == 32 || count % 300 == 0` so the user starts seeing items sooner

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1242,14 +1245,23 @@ private async Task EnumFromStorageFolderAsync()
...
                 {
-                    OrderFiles();
+                    var orderedList = OrderFiles2(tempList);
+                    await CoreApplication.MainView.ExecuteOnUIThreadAsync(() =>
```
gave92, line 1249: This should be unnecessary, enumeration from storagefolder is done on UI thread

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1878,17 +1890,13 @@ public void AddItemsToCollectionAsync(string path, string previousDir)
             RapidAddItemsToCollectionAsync(path, previousDir);
         }
 
-        public async Task AddFolderAsync(StorageFolder folder, string dateReturnFormat)
+        public async Task<ListedItem> AddFolderAsync(StorageFolder folder, string dateReturnFormat)
         {
             var basicProperties = await folder.GetBasicPropertiesAsync();
 
-            if ((AssociatedInstance.ContentFrame.SourcePageType == typeof(GenericFileBrowser)) || (AssociatedInstance.ContentFrame.SourcePageType == typeof(GridViewBrowser)))
+            if (!_addFilesCTS.IsCancellationRequested && ((AssociatedInstance.ContentFrame.SourcePageType == typeof(GenericFileBrowser)) || (AssociatedInstance.ContentFrame.SourcePageType == typeof(GridViewBrowser))))
```
gave92, line 1897: Could you write this check as (personal preference && coherence with other similar checks):
```cs
if (_addFilesCTS.IsCancellationRequested)
{	
     return null;	
}
if ((AssociatedInstance.ContentFrame.SourcePageType == typeof(GenericFileBrowser)) || (AssociatedInstance.ContentFrame.SourcePageType == typeof(GridViewBrowser)))
{
    return new ....
}
```

gave92 (changes requested):

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1904,11 +1912,12 @@ public async Task AddFolderAsync(StorageFolder folder, string dateReturnFormat)
...
                     FileSizeBytes = 0
                     //FolderTooltipText = tooltipString,
-                });
+                };
             }
+            return null;
         }
 
-        public async Task AddFileAsync(StorageFile file, string dateReturnFormat, bool suppressThumbnailLoading = false)
+        public async Task<ListedItem> AddFileAsync(StorageFile file, string dateReturnFormat, bool suppressThumbnailLoading = false)
```
gave92, line 1920: Could you make this `private`?

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1878,17 +1890,13 @@ public void AddItemsToCollectionAsync(string path, string previousDir)
             RapidAddItemsToCollectionAsync(path, previousDir);
         }
 
-        public async Task AddFolderAsync(StorageFolder folder, string dateReturnFormat)
+        public async Task<ListedItem> AddFolderAsync(StorageFolder folder, string dateReturnFormat)
```
gave92, line 1893: Could you make this `private`?

gave92 (approved):

gave92 (approved):
Looks good. Thanks a lot 👍

yaichenbaum:
@jakoss How will this effect the time it takes to load large directoreis when the setting to cache items is turned off?

jakoss:
It still flushes files list to UI when loaded 32 files or when `count % 300 == 0`. So basically it works the same as the native loading right now

jakoss:
Also, until now, i believe that items loaded with storage enumeration weren't properly ordered until count of 300. There was no Order after finished loading, so smaller lists might be not ordered. But i'm not 100% sure about that

gave92:
@jakoss i think you're right. Except GetItemsAsync() may already be returning items sorted by name by default (need to verify this). In any case the changes make it work even if the user changes the sort mode/direction

yaichenbaum:
@jakoss Looks good then. By the way, the issue where directories "load" twice when using the cache options seems to be happening consistently now.

jakoss:
That might be the issues with updating the observable collection of files, but i didn't observe that today once. I'll try and dig more about that later on

PR #2883, opened 2021-01-07
Preserve selection when sorting items

Don-Vito:


gave92:
@Don-Vito I could swear it doesn't work for big folders , could you check?

Don-Vito:
> 
> 
> @Don-Vito I could swear it doesn't work for big folders, could you check?

Yeah.. I see what your are talking about... it works for me quite well in a large folder when selecting only few items.. but when I select like 50 items, then selection disappears.. taking this back to draft :blush:

gave92:
@Don-Vito no issue :) but it happens all the time to me in say c:\windows\system32 when selecting one item and changing sort mode ascending to descending e.g.

Don-Vito:
> 
> 
> @Don-Vito no issue :) but it happens all the time to me in say `c:\windows\system32` when selecting even one item and changing sort mode ascending to descending e.g.
> I wonder if changing how we sort the items we could get the datagrid to automatically remember selection.. e.g trying to reorder the items calling the `Move` method of ObservableCollection?

This could be really good. I will check, but I am not too much optimistic, as MSFT examples for sort contain removing and adding.
It is super annoying that `CollectionViewSource` in UWP / XAML UI doesn't support the sorting / filtering capabilities that exist in WPF.

Don-Vito:
@gave92 - btw tested it in system 32 and the selection was preserved.. until the application became fully irresponsive.. writing tons of FileNotFoundExceptions and Access Denied.. 🤕

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -340,6 +339,7 @@ private void AllView_SelectionChanged(object sender, SelectionChangedEventArgs e
 
         private void AllView_Sorting(object sender, DataGridColumnEventArgs e)
         {
+            var previouslySelectedItems = SelectedItems;
```
gave92, line 342: `var previouslySelectedItems = SelectedItems.ToList();` should fix the issue

gave92:
@Don-Vito It looks like the selected items get cleared by the selection rectangle

Don-Vito:
> @Don-Vito It looks like the selected items get cleared by the selection rectangle, possibly when you change sort option while the list is scrolled
> Also something like this may be necessary -> `var previouslySelectedItems = SelectedItems.ToList();`

@gave92 - sounds good directions! I changed ordering to use Move and it already looks better. Hope to get to it later on, but with no day care due to another lockdown I am mostly trying to survive :smile:

yaichenbaum:
@Don-Vito Hi Don, can we get an update on this?

Don-Vito:
@yair100 - hey! I didn't work on the project.. or open source in general lately.. :( I miss the action and really hope to get back to it soon. I will close my PRs to cleanup the tracker.

yaichenbaum:
@Don-Vito Thanks for the update, we hope to see you active again in the future 🙂. Your contributions have been very helpful.

PR #2887, opened 2021-01-07
Improve layout mode switching

gave92:
Avoids reloading the folder contents when switching from ListView to Tiles or Grid view and between different GridView sizes.
Also fixes changing layout mode in search page

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -60,6 +60,8 @@ namespace Files
 
         public bool IsRenamingItem { get; set; } = false;
 
+        private NavigationArguments navigationArguments;
+
         private bool isItemSelected = false;
 
         public bool IsItemSelected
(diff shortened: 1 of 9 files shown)


yair100 (dismissed):

gave92:
@yair100 with the animations when switching from datagrid to tilesview I see items with the wrong icon (e.g files with a folder icon) then it fades to the correct one

yair100:
> @yair100 with the animations when switching from datagrid to tilesview I see items with the wrong icon (e.g files with a folder icon) then it fades to the correct one

Do you think this can be fixed by only showing the "add item" animation?

gave92:
Perhaps I'd just wait to add animations, when switching layout there's an ugly moment in which you see a gray background to the items. We'd probably need to fix that anyway before thing to animations.

ezgif com-gif-maker

mdtauk:
That odd refresh, perhaps you could have a placeholder icon until the grid item content has loaded

gave92:
@yair100 sorry this turned out bigger than I thought.
Switching layout mode or gridview size doesn't trigger a refresh anymore even between Datagrid and TilesView

yair100 (approved):

yaichenbaum:
@gave92 Amazing job!

PR #2900, opened 2021-01-10
Properly clear and disconnect ItemSource upon navigating away from layout mode Page

lukeblevins:
We also disable caching of the layout mode pages which should cause them to be recreated every time. It also may prevent unnecessary caching of content.

**Don't merge this change. This is for testing whether memory is freed more frequently now compared to master.**


lukeblevins:
Ignore the branch name.

@yair100 When you get a chance, I'm curious as to whether or not this change causes memory usage to return to somewhat normal levels after loading large directories and navigating away from them.

We may never merge this, but it's useful to know if this was one of the culprits.

Issue #1769, opened 2020-08-20
Improve performance


cristiangerardo:
Very good UI, but the performance is very bad. When accessing any folder, the system takes a long time to read and display the results.
In addition, the properties of disks such as folders or files are very basic.

yaichenbaum:
@cristiangerardo Can you provide information about the number of items in the location, the drive type, and how long it takes to load the items? It would also be helpful if you shared how the speeds compare to File Explorer.

perplexityjeff:
I continue to monitor the project with great interest but this is a lot of the reason that I am not using it as my daily driver as well. For example I use Tablacus now and it speed is amazing but the UI is outdated. 

Specs:
I have a 256 Samsung 8XX Pro SSD with 16GB RAM and a Ryzen 5.

jakoss:
I'm prety sure that native file explorer does some heavy indexing/caching internally, that's why it's so fast. I'm not sure how files are doing that now, but i think that caching paths in some super-fast embedded database (for example LMDB is extremly fast - https://github.com/kwaclaw/KdSoft.Lmdb). I'd love to take a shot at this, but a newborn baby limits my time a little bit. I'll keep that in mind and if nobody will take this - i might later on

yaichenbaum:
@jakoss The latest update should improve the performance, but if you have idea's to improve it further, feel free to take a look when you get a chance 😃

jakoss:
I did some measurement and experiments with memory cache and unfortunetely - i think the "load performance" issue is somewhere else (probably the rendering itself). All code that listed files performed under 800ms (average 350 ms for 20 measurements on my machine) which is in my opinion acceptable. I'm not sure how to measure perceivable performance here, but i'm pretty sure that ui (time that took from clicking the folder to displaying it on screen) is slower than all the code i was able to measure (methods in `ItemViewModel` that updates the `_filesAndFolders`).

The only weird thing in my measurements was line `var syncStatus = await CheckCloudDriveSyncStatusAsync(_rootFolder);` in method `ItemViewModel.EnumerateItemsFromStandardFolderAsync`. This method sometimes performed under 20 ms, but a few times it jumped even to 300 ms, and i could not correlate this to anything - the folders were all similar in size. I did all the testing on local folders, so i believe this can be avoided somehow.

Now, since i'm far from UWP expert (or any UI in particular) i probably cannot help you with UI optimizations, sorry

jakoss:
[...]

PR #2944, opened 2021-01-12
Preemptive caching

### LINKED ISSUE #3169: Crashes Upon Load
[debug.txt](https://github.com/files-community/Files/files/5836456/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5836457/debug_fulltrust.txt)

jakoss:
Adds preemptive caching to #1769
Fixes #3169

I had to do some refactoring to implement this feature. I had to make the "enumeration" functions more independent of `ItemViewModel`. I moved out some methods out of viewmodel, just for sake of readability and just to be a little more SOLID. I managed to lower `ItemViewModel` line count by 200.

There's one question from me. What is the reason to keep Win32 files enumeration along with Windows.Storage? I measured a little bit and i used application with only Windows.Storage enumeration and i didn't found significant performance difference. All gains from using direct C++ apis are minimal compared to other required operations in `EnumerateItemsFromStandardFolderAsync`. So i believe that using that introduces a lot of complicated and doubled logic without any real gain.  @yair100 @gave92 What do you think about that?

I'll tackle the "double loading" issue in next Pull Request.


gave92:
Windows.Storage does not lost shortcut files and hidden items

Files/Filesystem/StorageEnumerators/UniversalStorageEnumerator.cs
```diff
@@ -0,0 +1,268 @@
...
+                {
+                    break;
+                }
+                catch (Exception ex) when (
+                    ex is UnauthorizedAccessException
+                    || ex is FileNotFoundException
+                    || (uint)ex.HResult == 0x80070490) // ERROR_NOT_FOUND
+                {
+                    // If some unexpected exception is thrown - enumerate this folder file by file - just to be sure
+                    items = await EnumerateFileByFile(rootFolder, count, maxItemsToRetrieve);
```
jakoss, line 64: @gave92 As per our discussion - i'm using batch file loading and falling back to file by file only in case of exception here

Files/ViewModels/ItemViewModel.cs
```diff
@@ -806,10 +802,45 @@ public async void RapidAddItemsToCollectionAsync(string path, string previousDir
...
+                                    if (Path.IsPathRooted(path))
+                                    {
+                                        var res = await FilesystemTasks.Wrap(() => StorageFileExtensions.DangerousGetFolderWithPathFromPathAsync(path, null, parentFolder: currentStorageFolderSnapshot));
+                                        if (res)
+                                        {
+                                            storageFolder = res.Result;
+                                        }
+                                    }
+                                    await EnumerateItemsFromStandardFolderAsync(path, storageFolder, sourcePageType, _addFilesCTS.Token, cacheOnly: true);
+                                }, maxDegreeOfParallelism: 5);
```
jakoss: @gave92 @yair100 I setup 5 as a start, but i believe this should be an option in settings?

yaichenbaum: @jakoss What does this number control?

jakoss:
> Windows.Storage does not list shortcut files and hidden items
> And Win32 does not work on phone storage (and network?)

Ok, so removing that is not an option

gave92:
And none of those work on recycle bin xD yeah it's complicated

Files/ViewModels/ItemViewModel.cs
jakoss: On every navigation i'm running preemptive caching on child folders. This numbers controls the number of folders to cache in parallel. The larger the number - the more disc and cpu usage user can see

yaichenbaum: 5 seems like a good start, there should probably be a setting though.

jakoss: I added new setting. PR is ready for review

Files/Views/SettingsPages/Experimental.xaml
```diff
@@ -45,6 +46,16 @@
...
+
+                <controls:NumberBox 
+                    x:Uid="PreemptiveCacheParallelLimits"
+                    HorizontalAlignment="Left"
+                    Header="Preemptive cache parallel limit (0 turns off the preemptive cache)" 
+                    FontSize="17"
+                    SpinButtonPlacementMode="Inline"
+                    Width="300"
+                    Value="{Binding PreemptiveCacheParallelLimit, Mode=TwoWay}" 
+                    Minimum="0" />
```
yaichenbaum, line 58:
```suggestion
                    Minimum="0" 
                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
                    IsEnabled="{Binding UseFileListCache, Mode=OneWay}"/>
```

yair100 (commented):
@jakoss It seems like the NumberBox Header is skipped over by the screen reader, can you check that out?

Files/Views/SettingsPages/Experimental.xaml
jakoss: The issue with screen reader seems to be rooted in WinUI library. I didn't found such bug for NumberBox specificaly, but there are quite such bugs for other controls, like radio buttons: https://github.com/microsoft/microsoft-ui-xaml/issues/3183

I'm not quite sure what to do here

yaichenbaum: If there was one person who would know, it would probably be @chingucoding 😂

Files/Views/SettingsPages/Experimental.xaml
```diff
@@ -45,7 +46,20 @@
...
+                    x:Uid="PreemptiveCacheParallelLimits"
+                    HorizontalAlignment="Left"
+                    Header="Preemptive cache parallel limit (0 turns off the preemptive cache)" 
+                    SpinButtonPlacementMode="Inline"
+                    Width="450"
+                    Value="{Binding PreemptiveCacheParallelLimit, Mode=TwoWay}" 
+                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
+                    IsEnabled="{Binding UseFileListCache, Mode=OneWay}"
+                    Minimum="0" 
+                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
```
jakoss: @yair100 I already did that change, so we have a duplicates here

Files/Views/SettingsPages/Experimental.xaml
chingucoding: What exactly is the question here?

Files/Views/SettingsPages/Experimental.xaml
yaichenbaum: Whoops, sorry about that!

Files/Views/SettingsPages/Experimental.xaml
yaichenbaum: @chingucoding for some reason the header of the number box isn't read by the screen reader.

Files/Views/SettingsPages/Experimental.xaml
```diff
@@ -45,7 +46,20 @@
...
+
+                <controls:NumberBox 
+                    x:Uid="PreemptiveCacheParallelLimits"
+                    HorizontalAlignment="Left"
+                    Header="Preemptive cache parallel limit (0 turns off the preemptive cache)" 
+                    SpinButtonPlacementMode="Inline"
+                    Width="450"
+                    Value="{Binding PreemptiveCacheParallelLimit, Mode=TwoWay}" 
+                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
+                    IsEnabled="{Binding UseFileListCache, Mode=OneWay}"
```
yaichenbaum, line 58:
```suggestion
```

Files/Views/SettingsPages/Experimental.xaml
chingucoding: This is a known issue with NumberBox and the fact that it wraps around a TextBox which conflict with each other. See https://github.com/microsoft/microsoft-ui-xaml/issues/3505 and the fixing PR https://github.com/microsoft/microsoft-ui-xaml/pull/3775 for more context.

yaichenbaum: Thanks! I guess this will be fixed in one of the next WinUI releases so it's nothing to worry about.

Files/Views/SettingsPages/Experimental.xaml
```diff
@@ -45,7 +46,18 @@
...
+                <controls:NumberBox 
+                    x:Uid="PreemptiveCacheParallelLimits"
+                    HorizontalAlignment="Left"
+                    Header="Preemptive cache parallel limit (0 turns off the preemptive cache)" 
+                    SpinButtonPlacementMode="Inline"
+                    Width="450"
+                    Value="{Binding PreemptiveCacheParallelLimit, Mode=TwoWay}" 
+                    Minimum="0" 
+                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
+                    IsEnabled="{Binding UseFileListCache, Mode=OneWay}"/>
```
yaichenbaum, line 59: Would you be able to create a page on the [documentation ](https://github.com/files-community/files-community.github.io/blob/main/docs) site explaining how this works?

jakoss: Sure, i'll create performance article there

jakoss: https://github.com/files-community/files-community.github.io/pull/23

yair100 (dismissed):

Files/ViewModels/SettingsViewModel.cs
```diff
@@ -607,6 +607,15 @@ public bool UseFileListCache
             set => Set(value);
         }
 
+        /// <summary>
+        /// Gets or sets a value indicating the limit of parallel preemptive cache loading limit.
+        /// </summary>
+        public int PreemptiveCacheParallelLimit
+        {
+            get => Get(5);
```
gave92, line 615: Enabled by default is ok @yaichenbaum?

gave92 (changes requested):
I love the ItemViewModel refactoring :)

An issue I'm seeing is that with this option enabled folder switching is sometimes delayed (I'll try to post a video). Is it just me?

Files/Strings/en-US/Resources.resw
```diff
@@ -1554,6 +1554,9 @@
   <data name="SettingsUseFileListCache.Header" xml:space="preserve">
     <value>Cache files and folders for better performance</value>
   </data>
+  <data name="PreemptiveCacheParallelLimit.Header" xml:space="preserve">
```
gave92, line 1557: Why am I not seeing this resource added to all .xlf files?

yaichenbaum: It's actually preferred to let one of us do it when before submitting it to the store, this will prevent potential merge conflicts that are difficult to solve.

Files/ViewModels/SettingsViewModel.cs
yaichenbaum: That's fine, after testing it out, I think we should lower it a bit since it's a little heavy on system resources.

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1096,203 +1134,101 @@ public async Task<bool> EnumerateItemsFromStandardFolderAsync(string path)
                     FileSize = null,
                     FileSizeBytes = 0
                 };
+                if (!cacheOnly)
```
gave92, line 1137: Missing "{" for code style

Files/ViewModels/SettingsViewModel.cs
jakoss: I think it might be good to consider different values for different drive type

jakoss:
Wow, commit `c38ef5cee5dd72b0aab11fbd15f0e43d662e6006` was a killer merge for me

yaichenbaum:
@jakoss What's the current status with this PR?

jakoss:
Sorry, I had busy week. I need to merge hez2010 changes in here and that will take me some time. I'll try to do this sometime next week

yaichenbaum:
> Sorry, I had busy week. I need to merge hez2010 changes in here and that will take me some time. I'll try to do this sometime next week

It's no rush, thank you for the update.

jakoss:
@yair100 @gave92 I merged the changes from main branch, it was quite a ride after @hez2010 PR. But i think i got it right. Not sure about the end result after all. It does speed up things on my computer, but i have only nvme drives to test that on. And even on those the speedup is not really that mindblowing. Try it out on your machines.

Also there are minor changes to `PersistentSQLiteCacheAdapter` that should resolve the https://github.com/files-community/Files/issues/3169

I also made the adjustement to cache only the first 32 loaded items, since it's all we need to "visually" improve loading time.

Files/ViewModels/SettingsViewModel.cs
yaichenbaum: > I think it might be good to consider different values for different drive type

That might be a good way to do it, I would also consider changing this option into an on/off switch instead of letting the user choose the actual value.

jakoss: Do you have some tested way to check for the drive type in UWP?

yaichenbaum: @jakoss Not that I know of, perhaps we should stick with a lower number for now and if we find a way to detect the drive type, we can adjust it later on.

jakoss: Ok, i lowered this to 2 (i think it's better to target hard drives for now)

jakoss: I adjusted the settings (new on/off switch). I think PR is ready for review and merge

yaichenbaum: Should we remove the setting to adjust the limit and just have the option to turn it on or off?

jakoss: I think it's fine the way it is now. On/off switch is independent of limit setting. Limit selector will be enabled only if caching and preemptive caching are on. Users with fast ssd drives can bump this up. Minimum is 1

yair100 (dismissed):

gave92:
[DO NOT MERGE] Latest commit solves merge conflicts with main but looses change from (at least)  #3123, #3561

gave92 (dismissed):

gave92:
@jakoss I updated this PR with the latest changes from main. I hope I haven't screwed anything up.

Files/ViewModels/SettingsViewModel.cs
```diff
@@ -671,10 +671,28 @@ public bool ShowFileOwner
         /// </summary>
         public bool UseFileListCache
         {
-            get => Get(true);
+            get => Get(false);
```
yaichenbaum, line 674:
```suggestion
            get => Get(true);
```

yair100 (changes requested):

yair100 (approved):

yaichenbaum:
@jakoss Thank you!

Issue #2843, opened 2021-01-04
Bug report: Closing tabs don't free up memory


yaichenbaum:
**To Reproduce**
1. Open a couple tabs
2. Close them

**Expected behavior**
Memory goes down

winston-de:
It looks like there are some circular refences to the ModernShellPage that may be preventing proper garbage collection.
Files.ViewModels.ItemViewModel and Files.BaseLayout both contain circular references to ModernShellPage.

d2dyno1:
I feel like a lot of core logic has to be rewritten

ziv-commentor:
I don't feel like this is a problem: The file explorer merely takes up 100 MB of RAM, opening up 20 tabs doesn't increase RAM usage either.

yaichenbaum:
@ziv-commentor The tab itself takes around 1mb but if you navigate to large directory such as System 32, you will see the memory usage increase and closing the tab doesn't free up the memory.

winston-de:
Related: 
https://github.com/microsoft/microsoft-ui-xaml/issues/3597
#2065

duke7553:
@d2dyno1 Merely saying it needs to be re-written without specific solutions won't help anything. 

@winston-de Re-working ItemViewModel is quite a task, and I avoided it with the overhaul I did several months ago. We should aim to remove unwanted references from inside ItemViewModel. 

The minimal sample project I created on that WinUI issue demonstrates the issue is not entirely related to Files logic. I'd be interested in what is using the most memory after tabs are closed. TabView control uses ListView under the hood, yet clearing the data source doesn't free old memory (item containers?) either.

Old issue I found for reference: https://social.msdn.microsoft.com/Forums/en-US/92f76519-f139-4c94-ae5e-630f0abda66e/uwpcmemory-leak-with-listview-ui-virtualization

winston-de:
> I'd be interested in what is using the most memory after tabs are closed. TabView control uses ListView under the hood, yet clearing the data source doesn't free old memory (item containers?) either.

@duke7553 

These classes were the top 5 biggest users of memory after opening 7 tabs of System32 and closing it again.

Object Type | Count | Count Diff. | Size (Bytes) | Total Size Diff. (Bytes) | Inclusive Size (Bytes) | Inclusive Size Diff. (Bytes) | Module
-- | -- | -- | -- | -- | -- | -- | --
[...]

PR #2962, opened 2021-01-13
Un-initialize Drag and Drop events on items

duke7553:
We may never ship this, but these changes are reasonable and won't hurt in my opinion. This PR shouldn't break anything, but memory usage and scrolling performance should still be compared to master for potential regressions.

Addresses the part of #2843 that doesn't involve TabView at all, but item instances instead.

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -392,11 +392,6 @@ namespace Files
             }
         }
 
-        //protected override void OnNavigatedFrom(NavigationEventArgs e)
-        //{
-
-        //}
-
         private void UnloadMenuFlyoutItemByName(string nameToUnload)
         {
             if (FindName(nameToUnload) is MenuFlyoutItemBase menuItem) // Prevent crash if the MenuFlyoutItem is missing
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -972,6 +972,15 @@ namespace Files
             }
         }
 
+        protected void UninitializeDrag(UIElement element)
+        {
+            element.AllowDrop = false;
+            element.DragStarting -= Item_DragStarting;
+            element.DragOver -= Item_DragOver;
+            element.DragLeave -= Item_DragLeave;
+            element.Drop -= Item_Drop;
+        }
+
         // VirtualKey doesn't support / accept plus and minus by default.
         public readonly VirtualKey PlusKey = (VirtualKey)187;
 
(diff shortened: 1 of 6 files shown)


duke7553:
@tsvietOK Were you going to review this one?

tsvietOK:
@duke7553 sorry, didn't get a chance to review it.

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -112,8 +112,10 @@ private void SelectionRectangle_SelectionEnded(object sender, EventArgs e)
         protected override void OnNavigatedTo(NavigationEventArgs eventArgs)
         {
             base.OnNavigatedTo(eventArgs);
+            AllView.ItemsSource = ParentShellPageInstance.FilesystemViewModel.FilesAndFolders;
```
tsvietOK, line 115: What benefits will we get on setting ItemsSource in code behind?

Files/BaseLayout.cs
```diff
@@ -384,17 +384,19 @@ protected override void OnNavigatingFrom(NavigatingCancelEventArgs e)
...
-        protected override void OnNavigatedFrom(NavigationEventArgs e)
-        {
             var parameter = e.Parameter as NavigationArguments;
             if (!parameter.IsLayoutSwitch)
             {
                 ParentShellPageInstance.FilesystemViewModel.CancelLoadAndClearFiles();
             }
         }
 
+        //protected override void OnNavigatedFrom(NavigationEventArgs e)
```
tsvietOK, line 395: Is it needed?

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
duke7553: We set the ItemsSource to null after each navigation because it frees memory unlike simply using Clear(). This is because ListView builds an internal list for caching purposes and setting ItemsSource to null properly clears it. 

> ListView makes the distinction between hard and soft resets.
> A hard reset is when you set the ItemsSource to a new value (or to null). In that case, ListView will throw away all its containers.
> A soft reset is when you keep the same ItemsSource but you clear the collection. In that case, ListView treats it as any other collection change (e.g. add/remove) and put the containers in the recycle pool so that those containers can be reused in the future. Those containers don’t leave the visual tree and FrameworkElement.Unloaded won’t be raised because of that.

Files/BaseLayout.cs
duke7553: @tsvietOK Nope. Good catch

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
tsvietOK: I mean, you can keep binding ItemsSource in xaml, but set it to null (when needed) in code behind. This PR does both this operations in code behind.

duke7553: @tsvietOK Wouldn't setting it to null break the existing XAML binding?

tsvietOK (approved):

PR #2969, opened 2021-01-13
Docs: Fix capitalization and font weight

duke7553:


duke7553:
This also removes "explorer" from being used twice in a row.

yaichenbaum (approved):

PR #3031, opened 2021-01-14
Improved performance when scrolling in large directories

gave92:
Do you find scrolling on large folders to be smoooother with this PR?
Do you see issues with icons? E.g not loading?


yaichenbaum:
> Do you see issues with icons? E.g not loading?

Icons don't seem to be loading beyond the first 15 or so.

gave92:
~@yair100 Never? Sometimes? After switching folder? After switching layout? Only in listview/gridview..?~
Slower it's expected but not too much

yaichenbaum:
@gave92 Other than icons loading a little slower it seems to perform better.

yair100 (dismissed):

yair100 (approved):

Issue #2527, opened 2020-12-03
Context Menu Slow to Open


sguergachi:
**Is your feature request related to a problem? Please describe.**

When I right click on a file item the right click context menu takes at least a few seconds, which is considerably slower than native File Explorer and frequently interrupts my flow of using this app, making it unusable for me.

**Describe the solution you'd like**

I want it to show context menu as soon as possible, and if need be, load additional context menu items asynchronously, that way items that should be readily available  (cut, copy paste) don't need to wait for context items that need more time to load in.

**Describe alternatives you've considered**

Using native file explorer.

**Additional context**
Add any other context or screenshots about the feature request here. Ex: What device type/UI layout and app version?

This is my context menu for reference 
![image](https://user-images.githubusercontent.com/615386/101061125-25a99200-355e-11eb-98d4-7e1496e3f96e.png)

sguergachi:
Basically if we can have an instant access to these items on right click and load in the rest as they come in, that would improve UX by a lot

![image](https://user-images.githubusercontent.com/615386/101061370-6d301e00-355e-11eb-847e-0896d0748e46.png)

ShankarBUS:
How about using a CommandBarFlyout instead of a MenuFlyout?

Would that cause any problems?

Chris2011:
I can reproduce it. But IMHO it is more a bug or misbehaviour than a new feature.

gave92:
Some findings on this.

I've added a stopwatch in a couple of checkpoints to see what is taking the longest:
- [QueryContextMenu](https://github.com/files-community/Files/blob/master/Files.Launcher/Win32API_ContextMenu.cs#L231), 1 line
- [LoadContextMenu](https://github.com/files-community/Files/blob/master/Files.Launcher/Program.cs#L262) function in fulltrust process, which includes [1]
- [SetShellContextMenu](https://github.com/files-community/Files/blob/master/Files/BaseLayout.cs#L218) function, which includes [1] and [2] and is the total time taken to show the menu items

| QueryContextMenu [ms] | LoadContextMenu [ms] | SetShellContextmenu [ms] |
|:-----------------:|:-----------------: |:--------------------: |
|57                       |70                       |93                            |
|334                     |368                     |420                          |
|1576                   |1609                   |1681                        |
|2208                   |2232                   |2285                        |

[...]

PR #3045, opened 2021-01-15
Significantly improve overall performance and UI responsiveness

hez2010:
With this PR, UI responsiveness should be significantly improved. For test, just open `C:\Windows\System32` and UI almost no longer freezes.

Before (UI freeze everywhere, and after entering `System32` the UI froze for several seconds and cannot operate UI):

before

After (Even using debug build is smoother than before, and there's almost no UI freezes even after entering `System32` and `WinSxS`, you can operate UI immediately):

after

**Change Notes**
- Remove unnecessary `Task.Run`
- Use BulkObservableCollection and suppress CollectionChanged event while loading entries heavily
- Change the workflow of loading entries and applying UI updates:
  - Files and folders list is manipulated on `filesAndFolders`, which is a `List<ListedItem>`
  - After manipulation, call `ApplyFilesAndFoldersChangesAsync` to update UI, which calculate changes incrementally to avoid unnecessary list entry additions as many as possible
  - The `BulkObservableCollection<ListedItem> FliesAndFolders` should be only used by Bindings and `ApplyFilesAndFoldersChangesAsync`
  - This change also fixes the unexpected list flash while loading a directory
- Use `Task.Run` to make list sorting as a background task to avoid UI freezing
- Increase the max value of `loadExtendedPropsSemaphore` to match the number of CPU processors
- Remove `matchingItem` lookup in `LoadExtendedItemProperties` to avoid the possibility of enumerating on a modified list which may crash the whole program. Also `matchingItem` is unnecessary because the parameter `item` already comes from `filesAndFolders` and it's ok if the item doesn't exist in `filesAndFolders`
- Use an interval sampler to update DataGrid and improve loading performance
- Use a concurrent queue to optimize file monitor, preventing UI lags and possible dead locks when lots of file changes occurred

Note: 
- The xlf changes are generated by compilers, I didn't change any language resources
- Cache is preventing files from incremental loading (which can update UI while loading files and won't freeze UI), and it introduces incorrect caching result if user navigate too fast between directories, so I removed it as a temporary workaround.

Contributes to #1769 and #2527 and #3062


yaichenbaum:
@jakoss @gave92

jakoss:
The UI responsiveness is awesome, but file caching removal brings back the most pressing (for me) part that i'm waiting for files to load on every navigation. So i'd argue that it should be refined and integrated, not removed

hez2010:
> The UI responsiveness is awesome, but file caching removal brings back the most annoying (for me) part that i'm waiting for files to load on every navigation. So i'd argue that it should be refined and integrated, not removed

In my test, loading 5000 entries only need less than 2 seconds after PR, which is almost equal to loading those entries from SQLite cache. 
Also a cancelled enumeration could result in incorrect cache and there's no way to reset the cache.

I think we may improve the implementation of cache in the future.

jakoss:
5000 entries is very unlikely scenario, i think we can talk about 20-30 files in folder on average i think? Loading 30 entries from SQLite takes 80 ms at my machine, while loading them from drive takes around 800-1000. Also - SQLite is only second line of cache. Memory cache is significantly faster, it also stores extended properties. Renavigating to visited path is so much faster with this. I know that UI responsiveness is great, but the basic "fast navigation" flow is crucial for productiveness.

You can take a look at this PR: https://github.com/files-community/Files/pull/2944 . It implement preemptive caching, which on fast drives preloads child folders into memory cache.

There is no problem with clearing incorrect cache on enumeration, it's just one event on `CancellationToken`.

jakoss:
@hez2010 I think we could store cache entries as 1 row per file/folder. This way we could even get rid of json serialization. Then on cache load we could "preload" 30 entries and then load the rest, just like during ordinal loading.

Just to confirm - DataGrid have issues with loading a lot of data at once? Maybe simple partitioning of list would be enough? Something like: `load 30 entries, then add 300, then add 300 etc`

hez2010:
@jakoss For example, if 10k entries are added into ObservableCollection at once, even with my BulkObservableCollection DataGrid will stuck for several seconds in the end. Therefore there'll be a period of time of blank screen before list items being displayed. 
You can simply open `C:\Windoes\WinSxS` (which has at least 20k items) to test this.

jakoss:
I think we can do similar strategy for preloaded list from cache somewhere here: https://github.com/files-community/Files/blob/a7479a3f6c62375bd7029beb500f6c1b6d7127e4/Files/ViewModels/ItemViewModel.cs#L815 . It would "simulate" gradual loading

hez2010:
@jakoss Cache fixed via 105d793f296dd835afe5a53d7aaf1dcfc02ec55d

hez2010:
I've rebased this PR from master

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1171,28 +1254,34 @@ public async Task<bool> EnumerateItemsFromStandardFolderAsync(string path)
...
+                    });
 
+                    if (!addFilesCTS.IsCancellationRequested)
+                    {
                         await fileListCache.SaveFileListToCache(path, new CacheEntry
                         {
                             CurrentFolder = CurrentFolder,
-                            FileList = tempList
+                            // since filesAndFolders could be mutated, memory cache needs a copy of current list
+                            FileList = filesAndFolders.ToArray().ToList()
```
jakoss, line 1269: 'ToList()' on a list will create copy too, no need for double allocation here

jakoss (dismissed):
Great work, app is now amazingly responsive

Files/ViewModels/ItemViewModel.cs
hez2010: Fixed via c70cf14

hez2010:
I've added an interval sampler for updating UI which managed to improve the loading performance

Files/ViewModels/ItemViewModel.cs
```diff
@@ -800,38 +865,71 @@ public async void RapidAddItemsToCollectionAsync(string path, string previousDir
...
+                (bool LoadedFromCache, ListedItem CurrentFolder) cacheResult = (false, null);
+
+                if (useCache)
                 {
-                    CurrentFolder = cacheEntry.CurrentFolder;
-                    var orderedList = OrderFiles2(cacheEntry.FileList);
-                    OrderFiles(orderedList);
-                    Debug.WriteLine($"Loading of items from cache in {WorkingDirectory} completed in {stopwatch.ElapsedMilliseconds} milliseconds.\n");
-                    IsLoadingIndicatorActive = false;
+                    cacheResult = await Task.Run(async () =>
```
jakoss, line 899: So if I understand correctly - we are waiting for results to load from cache and only after that we're loading fresh data from file system? I think that's totally fine for small folders, but on large ones that might make everything longer. I wonder if maybe cache should be for small folders only? Or maybe only first 300 items should be cached, even if folder is bigger? @gave92 @yair100 what do you think about that?

Or maybe there's a way to load cache and filesystem in parallel? Although this might be really hard to synchronize. 
I think we should define equality check for 'ListedItem' so that we can compare items from filesystem and cache effectively (right now comparison is done by reference, which will be different when loading from sqlite)

hez2010: In my experiment, loading from the persistent cache for large folders is slow, and the user experience is lower than loading them from filesystem directly ("loading many entries at once but takes several seconds" vs "gradually loading them from filesystem but some entries can be immediately listed").
However, the memory cache does speed up a lot in both small and large folder cases, so I committed f76c98c which loads persistent cache into memory cache if memory cache is absent.

jakoss: Ok, so that leaves only one issue - the first time load of a big folders from sqlite might be slower then gradually loading them from filesystem. Hence my proposal for limiting cache only for smaller folders or to cache only first 32 files for initial user interaction and loading the rest from filesystem. I believe that's a decision for maintainers

Files/ViewModels/ItemViewModel.cs
```diff
@@ -460,143 +475,193 @@ public void CancelExtendedPropertiesLoading()
...
+                await Task.Run(() =>
                 {
-                    filesAndFolders.Add(orderedList[i]);
-                }
+                    for (var i = 0; i < filesAndFolders.Count; i++)
+                    {
+                        if (addFilesCTS.IsCancellationRequested) return;
+                        if (i < FilesAndFolders.Count)
+                        {
+                            if (FilesAndFolders[i] != filesAndFolders[i])
```
jakoss, line 517: Here we should use value equality, reference won't match with list from cache

hez2010: I think reference equality is needed here, because the only thing `ApplyFilesAndFoldersChangesAsync` needs to do is to reflect changes from `filesAndFolders`, and all entries in `FilesAndFolders` are referenced from `filesAndFolders`.

jakoss: Yes, but what about changes in 'filesAndFolders'? I mean - when some item is loaded from cache and then the same file is loaded from filesystem their reference won't match. It would be ideal for the 'FilesAndFolder' not to refresh lost in that case

hez2010: As for now the case "some item is loaded from cache and then the same file is loaded from filesystem" doesn't exists. Change this may cause unexpected behavior in `LoadExtendedItemProperties` for it uses item from `FilesAndFolders` directly.

jakoss: I'm pretty sure at some point you have to check if cache entries are up to date. In that place value equality could potentially be better. 

I don't have access to bigger screen right now, will take a look at this tomorrow

Files/ViewModels/ItemViewModel.cs
yaichenbaum: > Ok, so that leaves only one issue - the first time load of a big folders from sqlite might be slower then gradually loading them from filesystem. Hence my proposal for limiting cache only for smaller folders or to cache only first 32 files for initial user interaction and loading the rest from filesystem. I believe that's a decision for maintainers

I like the idea of loading only the first batch from the cache on the first interaction.

gave92:
@hez2010 I've noticed an issue for which the app will sometimes hang indefinitely while loading items. For some reasons the hang only happens in Grid View.

https://user-images.githubusercontent.com/9673091/104848119-ccb27500-58e3-11eb-860d-065684e9d843.mp4

When I look in VS the app is stuck at EndBulkOperation():
```cs
await CoreApplication.MainView.ExecuteOnUIThreadAsync(() =>
                {
                    // trigger ObservableCollection.CollectionChanged once
                    // loading is completed so that UI can be updated
                    FilesAndFolders.EndBulkOperation(); --------------> here
                    IsFolderEmptyTextDisplayed = FilesAndFolders.Count == 0;
                    UpdateDirectoryInfo();
                });
```

hez2010:
@gave92 I can sometimes reproduce the hang but it is caused by FilesFullTrust crash or WinRT Storage APIs timeout. Under this circumstance, you just need to wait a period from several seconds to a minute.

hez2010:
Rebased from main branch.

Files/ViewModels/ItemViewModel.cs
jakoss: So we can just cache first 32 items (or less) and ignore the rest. The gradual loading is fast enough that user won't be able to "go through" cached items while new ones will appear from filesystem. @hez2010 Can you add that in this PR?

Files/ViewModels/ItemViewModel.cs
jakoss: I went through the code and here's what i got:
- First - you are loading files from cache (if any)
- Second - you are running enumeration though a filesystem to load files from drive

At any point you do not clean `filesAndFolders` list between those 2 steps, so when files are loaded from drive at some point they are compared to the files already loaded from cache (in function `ApplyFilesAndFoldersChangesAsync`. Since files loaded from cache are different reference then those loaded from filesystem - it will always call `FilesAndFolders.Insert(i, filesAndFolders[i]);` forcing an unnecessary refresh of this particular item (i think, again i'm not an UWP guy, so here's where i can be dead wrong). And i think this can (and should probably?) be avoided by using value equality - do not replace item read from cache if it's the same item (i think value equality can be based on path + type). @gave92 @yair100 do you think that's reasonable?

Files/ViewModels/ItemViewModel.cs
jakoss: @hez2010 I think some rebase reverted that change

Files/ViewModels/ItemViewModel.cs
hez2010: Done in 06b47f4.

jakoss: Awesome!

Files/ViewModels/ItemViewModel.cs
hez2010: Fixed via 77503ed

Files/ViewModels/ItemViewModel.cs
hez2010: In 06b47f4 I just skipped entries which already loaded from cache, so I think it's not a issue anymore. Also, comparing by reference is quicker than comparing by value.

jakoss: Ok, i think that's fine. 

Comparing by reference is basically a comparison on int/long so it is faster, but i'm pretty sure we're talking about nanoseconds here ;)

Files/Views/LayoutModes/GridViewBrowser.xaml
```diff
@@ -15,7 +15,7 @@
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
     x:Name="PageRoot"
     Background="{ThemeResource ApplicationPageBackgroundThemeBrush}"
-    NavigationCacheMode="Disabled"
+    NavigationCacheMode="Enabled"
```
yaichenbaum, line 18: Is this change intentional?

hez2010: I didn't touch any XAML file, I think there's something wrong with rebase.

hez2010: Fixed in c63a9b7

yaichenbaum (approved):

Files/ViewModels/ItemViewModel.cs
```diff
@@ -800,22 +870,47 @@ public async void RapidAddItemsToCollectionAsync(string path, string previousDir
...
+                        {
+                            cacheEntry = null;
+                        }
+
+                        if (cacheEntry != null)
+                        {
+                            for (var i = 0; i < cacheEntry.FileList.Count; i++)
+                            {
+                                filesAndFolders.Add(cacheEntry.FileList[i]);
+                                if (addFilesCTS.IsCancellationRequested) break;
```
tsvietOK, line 903: [Code style 11](https://github.com/files-community/Files/wiki/Code-style#11)

tsvietOK (changes requested):

Files/Helpers/FileListCache/FileListCacheController.cs
```diff
@@ -27,21 +28,36 @@ private FileListCacheController()
         public Task SaveFileListToCache(string path, CacheEntry cacheEntry)
         {
             if (!App.AppSettings.UseFileListCache) return Task.CompletedTask;
```
tsvietOK, line 30: [Code style 11](https://github.com/files-community/Files/wiki/Code-style#11)

Files/Helpers/FileListCache/FileListCacheController.cs
```diff
@@ -27,21 +28,36 @@ private FileListCacheController()
...
             // save entry to persistent cache in background
-            Task.Run(async () => await persistentAdapter.SaveFileListToCache(path, cacheEntry));
-            return Task.CompletedTask;
+            return persistentAdapter.SaveFileListToCache(path, cacheEntry);
         }
 
-        public async Task<CacheEntry> ReadFileListFromCache(string path)
+        public async Task<CacheEntry> ReadFileListFromCache(string path, CancellationToken cancellationToken)
         {
             if (!App.AppSettings.UseFileListCache) return null;
```
tsvietOK, line 47: [Code style 11](https://github.com/files-community/Files/wiki/Code-style#11)

gave92:
> @gave92 I've made LoadExtendedItemProperties as a background task, could you test again?

@hez2010 looks like the issue is solved 👍
Minor issue remaining is that on loading, the gridview "flashes" twice which does not seem to happen on master:

https://user-images.githubusercontent.com/9673091/104963723-74f73500-59db-11eb-9a45-a8ace6c03a91.mp4

hez2010:
@gave92 This is expected because the collection is changing from time to time while loading entries, which causes GridView to rerender. In master the UI just freeze while loading so there is no flash.

gave92:
This is due to the fact that bulk observable collection fires a CollectionReset event right? Which causes the gridview to rerender completely

tsvietOK (approved):

gave92 (approved):

yaichenbaum:
@hez2010 @jakoss Fantastic work on this pull request! This will have a significant impact for everyone who uses Files.

hez2010:
@gave92 Yes, the `CollectionReset` event indicates the collection changed dramatically and it's supposed to be, so DataGrid re-renders it completely. There's no `AddRange` method for `ObservableCollection` so I think it has to be that.

Issue #3073, opened 2021-01-15
Using `/` as a path separator sometimes results in odd display


eamodio:
**Describe the bug**
Using `/` as a path separator sometimes results in odd display

**To Reproduce**
1. Type a full path using `/` rather than `\`
2. Observe that it goes to the right place
3. :bug: Observe that it displays incorrectly

Note, it works if you allow auto-complete (as it will fix the path)

**Expected behavior**
It work the same as `\`

**Screenshots**
![image](https://user-images.githubusercontent.com/641685/104787065-97c7e600-575c-11eb-8ec1-cacfba5daa65.png)

![image](https://user-images.githubusercontent.com/641685/104787038-841c7f80-575c-11eb-9e65-4d9c60524416.png)

**Desktop (please complete the following information):**
 - OS Version: [Windows 10.0.19042]
 - App version: [v1.1.0.0]

yaichenbaum:
@eamodio This issue should be resolved with #3096.

PR #3093, opened 2021-01-16
Fix paths separated by '/' not displaying properly.

BrianAllred:
Fixes/implements #3073.


BrianAllred:
There's still an issue with going up when you use '/'.

PR #3096, opened 2021-01-16
Fixed an issue where paths separated by '/' were not displaying properly

### LINKED ISSUE #3073: Using `/` as a path separator sometimes results in odd display

BrianAllred:
Fixes #3073. Also fixes going up when using '/' path separators.

### DIFF
--- a/Files/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
+++ b/Files/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
@@ -64,7 +64,7 @@ namespace Files.Filesystem
 
             for (var i = 0; i < value.Length; i++)
             {
-                if (value[i] == '\\' || value[i] == '?')
+                if (value[i] == '\\' || value[i] == '/' || value[i] == '?')
                 {
                     if (lastIndex == i)
                     {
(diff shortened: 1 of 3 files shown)


gave92:
@BrianAllred thanks for this PR. Could you verify if thumbnails in grid view are still shown when writing a path with forward slash?

BrianAllred:
My only concern with that is forcefully changing the user's use of '/' to '\\'. We'd have to be careful with that once WSL support is implemented and other filesystems come into play. But you're right, thumbnails are broken with my current changes, so changing it in `GetPathWithoutEnvironmentVariable()` is probably the better solution, at least for now.

gave92 (approved):

Issue #3095, opened 2021-01-16
Support for multiple OneDrive "Commercial" accounts


AndyMDoyle:
I have Office 365 accounts with multiple organisations so Explorer shows these all in the sidebar clearly with the company name.
Contrast this with Files that shows a single "OneDrive Commercial" entry as shown in this screenshot:

![image](https://user-images.githubusercontent.com/307719/104821629-2f990300-5835-11eb-94e1-0ae9289be063.png)

I would really like to see "OneDrive Commercial" replaced with however many "OneDrive - Tenant name" entries shown in Explorer.

Thanks.

gave92:
[Internal comment]
To detect onedrive we currently check the "%onedrive%", "%onedrivecommercial%" evironment variables, but this doesn't work in case of multiple account.
Onedrive lists all synced accounts in the registry under `HKCU\SOFTWARE\Microsoft\OneDrive\Accounts`
For example the path is save din `HKCU\Software\Microsoft\OneDrive\Accounts\Business1` -> `UserFolder`
I don't have a onedrive for business account unfortunately.

AndyMDoyle:
@gave92, thanks for the quick reply.

The following code _**would**_ correctly identify and add all OneDrive accounts (Personal and Business), however due to the sandbox environment a universal app runs in, access to these registry keys in not allowed.

```csharp
using Files.Enums;
using Microsoft.Win32;
using System.Collections.Generic;
using System.Security;
using System.Threading.Tasks;

namespace Files.Filesystem.Cloud.Providers
{
    public class OneDriveCloudProvider : ICloudProviderDetector
    {
        public async Task DetectAsync(List<CloudProvider> cloudProviders)
        {
            try
            {
                await Task.Run(() =>
                {
                    try
                    {
                        var oneDriveAccountsKey = Registry.CurrentUser.OpenSubKey(@"SOFTWARE\Microsoft\OneDrive\Accounts", false);

                        foreach (var account in oneDriveAccountsKey.GetSubKeyNames())
                        {
                            var accountKeyName = @$"{oneDriveAccountsKey.Name}\{account}";
                            var displayName = (string)Registry.GetValue(accountKeyName, "DisplayName", null);
                            var userFolder = (string)Registry.GetValue(accountKeyName, "UserFolder", null);

                            if (!string.IsNullOrWhiteSpace(displayName) && !string.IsNullOrWhiteSpace(userFolder))
                            {
                                cloudProviders.Add(new CloudProvider()
                                {
[...]

PR #3099, opened 2021-01-16
[Draft] Attempt to load OneDrive accounts from the Windows Registry

AndyMDoyle:
In relation to #3095 

Locally I see the SecurityException raised when attempting to open the registry key, however I believe this may be a local issue for me as @gave92 confirmed the code works okay for him (https://github.com/files-community/Files/issues/3095#issuecomment-761670885)

I am opening this PR to see if the resulting build artifact does indeed work for myself and anyone else who would like to test it.

If it does work, you should see "OneDrive - Personal" in the sidebar, plus additional entries for any business/education/government OneDrive accounts you have as "OneDrive - TENANT NAME".


Files/Filesystem/Cloud/Providers/OneDriveCloudProvider.cs
```diff
@@ -13,16 +14,28 @@ public async Task DetectAsync(List<CloudProvider> cloudProviders)
             {
                 await Task.Run(() =>
                 {
-                    var onedrivePersonal = Environment.GetEnvironmentVariable("OneDriveConsumer");
-                    if (!string.IsNullOrEmpty(onedrivePersonal))
+                    try
                     {
-                        cloudProviders.Add(new CloudProvider()
+                        var oneDriveAccountsKey = Registry.CurrentUser.OpenSubKey(@"SOFTWARE\Microsoft\OneDrive\Accounts", false);
```
gave92, line 19: `using var oneDriveAccountsKey = ...`

Files/Filesystem/Cloud/Providers/OneDriveCloudProvider.cs
```diff
@@ -13,16 +14,28 @@ public async Task DetectAsync(List<CloudProvider> cloudProviders)
...
                         {
-                            ID = CloudProviders.OneDrive,
-                            Name = "OneDrive",
-                            SyncFolder = onedrivePersonal
-                        });
+                            var accountKeyName = @$"{oneDriveAccountsKey.Name}\{account}";
+                            var displayName = (string)Registry.GetValue(accountKeyName, "DisplayName", null);
+                            var userFolder = (string)Registry.GetValue(accountKeyName, "UserFolder", null);
+
+                            if (!string.IsNullOrWhiteSpace(displayName) && !string.IsNullOrWhiteSpace(userFolder))
```
gave92, line 27: For me (Personal account) displayName in null so its not added

Files/Filesystem/Cloud/Providers/OneDriveCloudProvider.cs
```diff
@@ -13,16 +14,28 @@ public async Task DetectAsync(List<CloudProvider> cloudProviders)
...
+                            {
+                                cloudProviders.Add(new CloudProvider()
+                                {
+                                    ID = CloudProviders.OneDrive,
+                                    Name = $"OneDrive - {displayName}",
```
gave92, line 32: `Name = string.IsNullOrWhiteSpace(displayName) ? "OneDrive" : $"OneDrive - {displayName}",`

Files/Filesystem/Cloud/Providers/OneDriveCloudProvider.cs
```diff
@@ -13,16 +14,28 @@ public async Task DetectAsync(List<CloudProvider> cloudProviders)
             {
                 await Task.Run(() =>
                 {
-                    var onedrivePersonal = Environment.GetEnvironmentVariable("OneDriveConsumer");
-                    if (!string.IsNullOrEmpty(onedrivePersonal))
+                    try
```
gave92, line 17: The outer try/catch should be enough

gave92 (commented):
@AndyMDoyle posting some notes on the code, feel free to ignore them till you've verified that this approach works for you

AndyMDoyle:
@gave92, thanks for the feedback.

I am still unsure how it seems to work for you but locally I get the security exception. From what I know, UWP apps can't read from most parts of the registry unless they use an app service connection to a non-UWP app (such as Files.Launcher). I can read HKEY_CLASSES_ROOT, and HKEY_CURRENT_USER at the top level and \Environment keys, but nothing below HKCU\Software\Microsoft.

If anyone else can confirm if this works, it might point to an issue with my registry, or possibly I am doing something wrong when launching from VS.

#3099 has been updated to address the personal OneDrive account bug so it is ready for further testing.


In the meantime I have gone down the route of using Files.Launcher to read the registry keys and I have an working sample. Well I say working.... it works perfectly if debugging and stepping through code. But if launched without any breakpoints, the main UI appears but no OneDrive entries show up. From a lot of testing, I believe setting up the AppService connection has some initial delay, or a timing issue when not using the AppService connection from a UI component. Adding an artificial delay to simulate stepping through code between opening the connection and sending a message over the connection demonstrates it works.

I am not sure if this could be related to the main UI itself establishing a connection and my extra connection from ```OneDriveCloudProvider``` interferring; but I see no easy way to obtain a connection other than constructing my own. I have considered that it might be beneficial to have a single class responsible for maintaining and establishing an app-wide AppServiceConnection, but that may be a bad direction and require quite a bit of a rewrite.

If have this alternative method on a branch in case you'd like to see it: https://github.com/AndyMDoyle/Files/commit/29fe6329afbaf161be7a5c0777178aab1c56ee3d

gave92:
@AndyMDoyle thanks I've just tested your latest changes and Onedrive - Personal happily shows on the sidebar in Release mode as well. @yair100, @tsvietOK could you test the azure build of this PR and see if it works for you?
If the fulltrust process method gets too complicated you could also take a look at reading the onedrive path from the .ini files here `%localappdata%\Microsoft\OneDrive\settings\Personal`. This should be free of permission and timing issues

yaichenbaum:
Will this pull request help for displaying multiple OneDrive accounts or is it just an improvement over the current code?

yaichenbaum:
@gave92 Is just tested this PR in debug and OneDrive isn't coming up on the sidebar.

gave92:
@yair100 Thanks, I guess I'm the lucky one then. @AndyMDoyle ok, sorry this method is a no-go.

> Will this pull request help for displaying multiple OneDrive accounts or is it just an improvement over the current code?

It would show multiple business accounts with the company name

AndyMDoyle:
@yair100, this method of using the registry to load the list of configured OneDrive accounts will result in any configured OneDrive accounts being listed exactly as Explorer displays them.

Sadly it seems the UWP sandbox is a problem if it doesn't work for you either.

Using Files.Launcher (as with my branch above) is one workaround. I have had a look at the ini files on disk and there doesn't appear to be consistency between the file naming or file structure. Each account has files with a random filename, and the contents vary so pinpointing the path is going to be a string parsing nightmare.

I am going to try and improve the Files.Launcher situation regarding the delay when I have a little more time, unless anyone has any other ideas?

gave92:
The ini should have a line like this `library = 1 4 AD184A76A70E8923!122 1483211540 "SkyDrive" Me personal "D:\Users\Marco\OneDrive" `. Is it different for you / business?
When I've time I'll take a look at your branch and see if anything comes to mind.

AndyMDoyle:
For business accounts the only line containing the path is this:

```libraryScope = 0 c1c456ef4XXXXXXXXXXXXXXXX7c1 5 "MySite" "ODB" 2 "https://TENANT-my.sharepoint.com/personal/andy_doyle_DOMAIN" "b2dfXXXX-XXXX-XXXX-XXXX-XXXXXXXXb21e" 6fd6XXXXXXXXXXXXXXXXXXXXXXXf28b 513fXXXXXXXXXXXXXXXXXXXXXXXb0e0 f2b4XXXXXXXXXXXXXXXXXXXXXXX399f 161XXXXXX24 "C:\Users\andydoyle\OneDrive - COMPANY NAME" 1 ad50XXXX-XXXX-XXXX-XXXX-XXXXXXXX2f31  - ```

It feels a little dirty relying on this when Microsoft could break it at any moment simply by slightly changing one line in a text file.

gave92:
@AndyMDoyle well they could change the registry too I guess.. but I hoped the ini file would have been a bit more uniform

AndyMDoyle:
I'm going to propose a middle ground to give the enhancement without reduced risk of OneDrive not appearing at all.

I can update the OneDriveCloudProvider by enhancing it by parsing the INI file. We do both methods of using the environment settings, but attempt to add additional accounts by parsing the INI files. This way if there is a problem loading or parsing the files, users will still see their personal and first business account in the UI.

Does that sound worthwhile?

yaichenbaum:
It's strange that it's giving issues reading from the registry, we do it in other areas and it seems to work fine.

AndyMDoyle:
@yair100, I agree. From what I can tell inside the sandbox it can read certain parts of the registry, but not everything.

I don't know if Microsoft protect HKCU\Software\Microsoft completely from being read; or if there is some other logic behind it.

yaichenbaum:
@AndyMDoyle In that case we should probably do what Gave92 suggested and use the FullTrustProcess for this.

AndyMDoyle:
@yair100, I have that almost working but as above, there seems to be some delay in establishing the connection if I create a connection from OneDriveCloudProvider. As OneDriveCloudProvider is constructed and used early in the application startup, it seems the startup finishes and shows before the connection is available.

I'm going to try a few more things to make that more reliable though.

gave92:
@AndyMDoyle Yeah it looks like the fulltrust process takes some time (few seconds) to load and establish the connection. 
For reference I'm posting below a clear and artistic depiction of the appservice communication in Files:

image

The UWP app does not speak directly to the fulltrust process, so when "OpenAsync()" returns it only means that you ghave connected to "MessageRelay", no guarantee that the fulltrust process has started and connected to "MessageRelay" itself

AndyMDoyle:
@gave92 @yair100 Here's a larger update to get your teeth around when you have time. It consists of the following changes:

- ```ICloudProviderDetector``` no longer accepts an existing list and only concerns itself with returning a new list of ```CloudProvider``` objects it cares about
- ```CloudProviderController``` fires each ```ICloudProviderDetector``` off with a task continuation so all detectors can go about their job in parallel
- An event handler in ```CloudProviderController``` is fired when an ```ICloudProviderDetector``` returns any results
- ```Drives``` listens for this event handler being called, and adds the resulting ```CloudProvider``` objects to ```drivesList```, and calls ```DeviceWatcher_EnumerationCompleted``` to update the list of drives displayed in the UI
- ```AppServiceConnectionHelper``` is a single place to go to obtain an app service connection. It is initialised in App.xaml.cs with ```AppServiceConnectionHelper.Initialize();```, and all views, and ```OneDriveCloudProvider``` now access and share this single connection instead of constructing multiple connections and passing them around.
- ```AppServiceConnectionHelper``` has a utility method to check if the connection is in place. Calling ```AppServiceConnectionHelper.IsConnected()``` will return a boolean confirming if it was able to send a message and get a matching response back from FullTrust (a simple Marco-Polo shout out to check if the sender gets the expected response back)
- ```OneDriveCloudProvider``` waits for the app service connection to be established before attempting to load the registry settings. I have added a basic timeout of 60 seconds after which it will give up waiting for the connection to be available and will fail silently

So far, everything I have tested works. All OneDrive accounts appear as they should, folders show data, icons load, and if you have multiple tabs open, a change that utilises FullTrust (such as changing a drive label), updates every tab.

I have also simulated a very long running ```ICloudProviderDetector``` by adding ```await Task.Delay(30000);``` at the start of ```DetectAsync()```. With this in place, the UI loads straight away as you would want, everything is responsive/usable, and after 30 seconds those cloud provider entries are added to the sidebar.

I know this is rather big change and I have touched a lot of things that aren't specifically OneDrive related. I hope this isn't treading on anyone's toes, and that I haven't inadvertently broken something.

If this is of any use then great; or if you have any issues with it and want to use any of this as a starting point then please feel free to do so!

AndyMDoyle:
Closing and replacing with #3177

PR #3120, opened 2021-01-17
[Do not merge] - Show localized names for folders

gave92:
Opening this PR just to gather some suggestions.

This PR shows the localized name for system folders (e.g "Program Files" -> "Programmi" in Italian)
To not slow down file loading, the localized name is fetched in LoadExtendedItemProperties() but this has another problem: the file list won't be sorted by name anymore. Ordering the list again works but it's quite ugly.

@jakoss how hard would it be to cache the localized names? I'd cache them separately from the file list cache and save the filesystem path + localized name

### DIFF
--- a/Files/ViewModels/ItemViewModel.cs
+++ b/Files/ViewModels/ItemViewModel.cs
@@ -492,14 +492,14 @@ namespace Files.ViewModels
         {
             if (orderedList == null)
             {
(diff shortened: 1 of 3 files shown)


jakoss:
I see no problem in separate cache for localised names only. It will be saved on `LoadExtentedItemProperties` and loaded during file loading

PR #3156, opened 2021-01-18
Tweaked settings font sizes and strings

yaichenbaum:


mdtauk:
Posting these here by way of illustrating the more modern settings design used by Windows

image

image

PR #3157, opened 2021-01-18
Properties: Change "OK" button to Accent button style

### LINKED ISSUE #3130: Change "OK" action in properties dialog to Accent style
Currently in the properties dialog, there are two options: "OK" and "Cancel". These both use reveal style, however I think that it would both look better and improve the user experience by distinguishing these two buttons better.

Change "OK" button in properties dialog to `AccentButtonStyle` to highlight the primary action of the dialog.

None

**Additional context**
image

XPoppyX:
Closes #3130 

- ----------

button1
button2


Files/Views/Pages/Properties.xaml
```diff
@@ -99,7 +99,7 @@
                 HorizontalAlignment="Right"
                 Click="OKButton_Click"
                 Content="OK"
-                Style="{ThemeResource ButtonRevealStyle}" />
+                Style="{ThemeResource AccentButtonStyle}" />
```
tsvietOK, line 102: You can use `{StaticResource AccentButtonStyle}` here.

XPoppyX: ok

mdtauk:
The accent button style is traditionally used for "default actions", so the action that will be performed when you press Enter, because that item has focus.

In modern UI, it is also used to distinguish destructive actions, or actions that will change something.

So its important to not only change the style, but to examine the behaviour of the button too.

Files/Views/Pages/Properties.xaml
tsvietOK: In that case `AccentButton.xaml` and `packages.config` no more needed.

XPoppyX: removed

tsvietOK (approved):

yaichenbaum:
@XPoppyX Can you remove the reveal style from the cancel button as well to make it consistent with the "OK" button?

yaichenbaum (approved):

tsvietOK (approved):

yaichenbaum:
@XPoppyX Great work!

PR #3159, opened 2021-01-18
Indicate current layout mode in the layout flyout on the status bar

### LINKED ISSUE #1023: Indicate the current layout mode
concept:
Captura de pantalla 2021-01-05 184523

lukeblevins:
image

This should work properly when layout mode is changed or on navigations. My main issue is likely code quality, so please look for any regressions. Warning: hacks are in use 😄

Fixes #1023


yaichenbaum:
@duke7553 Everything looks good! LGTM

yair100 (approved):

mdtauk:
Should the layout button itself, display the icon of the currently chosen layout mode?

yaichenbaum:
> Should the layout button itself, display the icon of the currently chosen layout mode?

That was one of the options we considered decided against since the flyout now includes options besides for changing the layout mode.

mdtauk:
> > Should the layout button itself, display the icon of the currently chosen layout mode?
> 
> That was one of the options we considered decided against since the flyout now includes options besides for changing the layout mode.

I wasn't saying only indicate on the status bar button, I mean in addition to the toggled style in the flyout itself.

Like how a ComboBox displays the current text in the control, as well as highlighted in the flyout.

yaichenbaum:
@mdtauk I am not sure I understand your proposal correctly, do you mind coming up with a rough sketch of your idea?

mdtauk:
> @mdtauk I am not sure I understand your proposal correctly, do you mind coming up with a rough sketch of your idea?

image

@yair100  Here you can see the icon is different between the two panes.  Pressing the button would still bring up the new flyout

yaichenbaum:
@mdtauk Ok, that's what I originally though you meant 😄. The issue with that is that it might confuse users if the icon keeps changing, it's one thing if there is an icon to indicate the layout mode, but it's another to keep changing the icon they are meant to use to trigger some actions.

mdtauk:
> @mdtauk Ok, that's what I originally thought you meant 😄. The issue with that is that it might confuse users if the icon keeps changing, it's one thing if there is an icon to indicate the layout mode, but it's another to keep changing the icon they are meant to use to trigger some actions.

The tooltip would still say **Layout**, and the existing icon is not immediately obvious as to it's purpose - but I am just making the suggestion.

yaichenbaum:
@mdtauk The tooltip is good, but users still connect the icon to the action. That's also why it doesn't really matter what the icon is, users will learn what the icon does the first time that they click on it. Yes, they might eventually learn that the icon indicates the layout mode when the icon keeps changing, but I think it's better to reduce any potential confusion here.

mdtauk:
> @mdtauk The tooltip is good, but users still connect the icon to the action. That's also why it doesn't really matter what the icon is, users will learn what the icon does the first time that they click on it. Yes, they might eventually learn that the icon indicates the layout mode when the icon keeps changing, but I think it's better to reduce any potential confusion here.

I would argue the individual view mode icons better represent the layout, but I wont belabour the point :)

PR #3191, opened 2021-01-20
Move cloud drive detection out of DrivesManager and improve sidebar population

AndyMDoyle:
DrivesManager was pretty complicated and had to deal with enumerating local and cloud drives, and rendering them both in the sidebar. Keeping this all in sync was leading to duplication and performance issues.

This PR introduces ```CloudDrivesManager``` which takes over responsibility of identifying cloud providers on a system, and rendering these in a section of the sidebar titled "Cloud Drives".

The original ```DrivesManager``` is now a little simpler, and has a rewritten method to add/remove drives in the sidebar.

I have identified that some of the sync and duplication issues appeared to be caused by timing issues likely introduced by various async calls being made from constructors (which obviously are not async methods). To improve this situation, ```CloudDrivesManager```, ```DrivesManager```, and ```SettingsViewModel``` have been tweaked to use a factory constructor pattern. Now instead of ```new```ing up those objects and having async methods called and forgotten about, a new instance can be obtained by calling ```await ClassName.CreateInstance();``` which allows async methods to be called when these objects are setup.

I have seen some areas; particularly stemming from ```SettingsViewModel``` where I feel classes have more responsibility than they need, and have complicated dependencies. For example, ```SettingsViewModel``` is responsible for loading all drives, cloud drives, and a lot of other system information - making it more than just settings. By being responsible for loading all drives, it has a hard dependency on ```DrivesManager```, which in one or more methods directly calls methods on views. ```DrivesManager``` also depends on ```DriveItem``` which also calls the UI to perform updates or run code on the UI thread.

I started going down a deep whole trying to refactor some of this to make changes easier to test with unit tests, but these dependencies make it almost impossible. So I worked my way back out and have done this with the minimal changes. I have some ideas that might help based on what I've done with some Xamarin apps, but I don't know how feasible they will be with a UWP app.

@gave92, please give this a try and see if it stops you duplicated drive issue.

### DIFF
--- a/Files/Filesystem/CloudDrivesManager.cs
+++ b/Files/Filesystem/CloudDrivesManager.cs
@@ -103,7 +103,21 @@ namespace Files.Filesystem
                         {
                             Text = "SidebarCloudDrives".GetLocalized()
                         };
-                        MainPage.SideBarItems.Add(drivesSection);
+
+                        //Get the last location item in the sidebar
+                        var lastLocationItem = MainPage.SideBarItems.LastOrDefault(x => x is LocationItem);
+
+                        if (lastLocationItem != null)
+                        {
+                            //Get the index of the last location item
+                            var lastLocationItemIndex = MainPage.SideBarItems.IndexOf(lastLocationItem);
+                            //Insert the drives title beneath it
+                            MainPage.SideBarItems.Insert(lastLocationItemIndex + 1, drivesSection);
+                        }
+                        else
+                        {
+                            MainPage.SideBarItems.Add(drivesSection);
+                        }
                     }
 
                     var sectionStartIndex = MainPage.SideBarItems.IndexOf(drivesSection);
--- a/Files/Filesystem/CloudDrivesManager.cs
+++ b/Files/Filesystem/CloudDrivesManager.cs
@@ -5,14 +5,9 @@ using Microsoft.Toolkit.Uwp.Extensions;
 using NLog;
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Windows.ApplicationModel.Core;
-using Windows.Devices.Enumeration;
-using Windows.Devices.Portable;
-using Windows.Storage;
 using Windows.UI.Core;
 
 namespace Files.Filesystem
@@ -68,9 +63,9 @@ namespace Files.Filesystem
             {
                 await SyncSideBarItemsUI();
             }
-            catch (Exception)       // UI Thread not ready yet, so we defer the pervious operation until it is.
+            catch (Exception) // UI Thread not ready yet, so we defer the pervious operation until it is.
             {
-                Debug.WriteLine($"RefreshUI Exception");
+                System.Diagnostics.Debug.WriteLine($"RefreshUI Exception");
                 // Defer because UI-thread is not ready yet (and DriveItem requires it?)
                 CoreApplication.MainView.Activated += RefreshUI;
             }
@@ -84,7 +79,6 @@ namespace Files.Filesystem
 
         private async Task SyncSideBarItemsUI()
         {
-            Debug.WriteLine("SyncSideBarItemsUI");
             await CoreApplication.MainView.CoreWindow.Dispatcher.RunAsync(CoreDispatcherPriority.Normal, () =>
             {
                 lock (MainPage.SideBarItems)
--- a/Files/Filesystem/CloudDrivesManager.cs
+++ b/Files/Filesystem/CloudDrivesManager.cs
@@ -1,0 +1,177 @@
+using Files.Filesystem.Cloud;
+using Files.Views;
+using Microsoft.Toolkit.Mvvm.ComponentModel;
+using NLog;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Windows.ApplicationModel.Core;
+using Windows.Devices.Enumeration;
+using Windows.Devices.Portable;
+using Windows.Storage;
+using Windows.UI.Core;
+
+namespace Files.Filesystem
+{
+    public class CloudDrivesManager : ObservableObject
+    {
+        private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
+        private List<DriveItem> drivesList = new List<DriveItem>();
+
+        public IReadOnlyList<DriveItem> Drives
+        {
+            get
+            {
+                lock (drivesList)
+                {
+                    return drivesList.ToList().AsReadOnly();
+                }
+            }
+        }
+
+        private async Task EnumerateDrivesAsync()
+        {
+            var cloudProviderController = new CloudProviderController();
+            await cloudProviderController.DetectInstalledCloudProvidersAsync();
+
+            foreach (var provider in cloudProviderController.CloudProviders)
+            {
+                var cloudProviderItem = new DriveItem()
+                {
+                    Text = provider.Name,
+                    Path = provider.SyncFolder,
+                    Type = DriveType.CloudDrive,
+                };
+                lock (drivesList)
+                {
+                    drivesList.Add(cloudProviderItem);
+                }
+            }
+
+            await RefreshUI();
+        }
+
+        public static async Task<CloudDrivesManager> CreateInstance()
+        {
+            var drives = new CloudDrivesManager();
+            await drives.EnumerateDrivesAsync();
+            return drives;
+        }
+
+        private async Task RefreshUI()
+        {
+            try
+            {
+                await SyncSideBarItemsUI();
+            }
+            catch (Exception)       // UI Thread not ready yet, so we defer the pervious operation until it is.
+            {
+                Debug.WriteLine($"RefreshUI Exception");
+                // Defer because UI-thread is not ready yet (and DriveItem requires it?)
+                CoreApplication.MainView.Activated += RefreshUI;
+            }
+        }
+
+        private async void RefreshUI(CoreApplicationView sender, Windows.ApplicationModel.Activation.IActivatedEventArgs args)
+        {
+            await SyncSideBarItemsUI();
+            CoreApplication.MainView.Activated -= RefreshUI;
+        }
+
+        private async Task SyncSideBarItemsUI()
+        {
+            Debug.WriteLine("SyncSideBarItemsUI");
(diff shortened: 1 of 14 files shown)


Files/Filesystem/CloudDrivesManager.cs
```diff
@@ -0,0 +1,177 @@
...
+                    {
+                        //No drives - remove the header
+                        MainPage.SideBarItems.Remove(drivesSection);
+                    }
+
+                    if (drivesSection == null)
+                    {
+                        drivesSection = new HeaderTextItem()
+                        {
+                            Text = "Cloud Drives"
```
yaichenbaum, line 103: Is it posible to add this string to the en-US resource file?

AndyMDoyle: Of course. Does it matter if a string does not have translations in all of the other resource files?

yaichenbaum: @AndyMDoyle I will take care of generating translations in the other files.

AndyMDoyle: Done

yaichenbaum:
@AndyMDoyle Can you put the cloud section above the drive section?

yaichenbaum:
@gave92 Once we are changing how this works, handling the drives should probably moved completly out of SettingsViewMode, what do you think?

AndyMDoyle:
> @AndyMDoyle Can you put the cloud section above the drive section?

@yair100 Almost finished testing the SharePoint integration so will send over screenshot of it both ways around so you can get an idea of what it looks like with a few more cloud drives listed.

yaichenbaum:
@AndyMDoyle That sounds great! The goal is to eventually make the groups collapsible so I wouldn't be too worried about how many items are in each section.

Files/ViewModels/SettingsViewModel.cs
```diff
@@ -28,17 +29,31 @@ public class SettingsViewModel : ObservableObject
     {
         private readonly ApplicationDataContainer localSettings = ApplicationData.Current.LocalSettings;
 
-        public DrivesManager DrivesManager { get; }
+        public DrivesManager DrivesManager
```
gave92: I agree @yair100 we should probably remove the drives from settings, they don't have much business here

Files/ViewModels/SettingsViewModel.cs
```diff
@@ -53,6 +68,26 @@ public SettingsViewModel()
...
+        {
+            var settings = new SettingsViewModel(false);
+            return settings.Initialize();
+        }
+
+        private SettingsViewModel() : this(true)
+        {
+        }
+
+        private SettingsViewModel(bool initialize)
```
gave92, line 85: At the end I'd keep only the constructor that's actually used

Files/Views/MainPage.xaml.cs
```diff
@@ -75,7 +73,7 @@ protected override async void OnNavigatedTo(NavigationEventArgs eventArgs)
         {
             if (eventArgs.NavigationMode != NavigationMode.Back)
             {
-                App.AppSettings = new SettingsViewModel();
+                App.AppSettings = await SettingsViewModel.CreateInstance();
```
gave92, line 76: Mmm this adds 2 seconds of white screen before anything is loaded, not good.
If we remove drives from settings then loading drives should be done either at the end

Files/Views/YourHome.xaml.cs
```diff
@@ -85,6 +85,18 @@ private async void RecentFilesWidget_RecentFileInvoked(object sender, UserContro
                             return;
                         }
                     }
+                    foreach (DriveItem drive in AppSettings.CloudDrivesManager.Drives)
```
gave92, line 88: Since the two for loops look identical we can just loop on the two lists together

gave92 (changes requested):

gave92:
Wait is it just me or the tab bar disappears with this PR? 😮

AndyMDoyle:
Argh, I thought I had dealt with that tab bar issue by having settings available a little earlier.

It's again a timing issue because HorizontalMultitaskingControl relies on theme info from AppSettings. And while debugging to simulate a delay in loading settings, the MainPage view could possibly load before AppSettings (and the theme info) was populated, and if HorizontalMultitaskingControl can't get the settings it needs it doesn't appear properly.

I had a partial fix by moving the generation of AppSettings right to the startup of the app; but this caused issues because AppSettings depends on things that update the UI; and at that stage of the lifecycle the UI wasn't yet ready.

I will do some more on it this evening after work.

Files/Filesystem/CloudDrivesManager.cs
```diff
@@ -0,0 +1,178 @@
...
+                    foreach (var drive in Drives.OrderBy(o => o.Text))
+                    {
+                        MainPage.SideBarItems.Insert(insertAt, drive);
+                        insertAt++;
+                    }
+                }
+            });
+        }
+
+        public static async Task<StorageFolderWithPath> GetRootFromPathAsync(string devicePath)
```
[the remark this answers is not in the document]
AndyMDoyle: Correct

Files/Filesystem/Drives.cs
```diff
@@ -71,53 +85,91 @@ private async void EnumerateDrives()
...
+                            .Where(x => x.ItemType == NavigationControlItemType.Drive)
+                            .ToList())
                         {
                             MainPage.SideBarItems.Remove(item);
                             DrivesWidget.ItemsAdded.Remove(item);
                         }
+
+                        //Add all drives to the sidebar
+                        var insertAt = sectionStartIndex + 1;
+                        foreach (var drive in Drives.OrderBy(o => o.Text))
```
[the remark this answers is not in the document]
AndyMDoyle: Fixed

Files/ViewModels/SettingsViewModel.cs
AndyMDoyle: Maybe a challenge for another day :)

Files/ViewModels/SettingsViewModel.cs
AndyMDoyle: Removed

Files/Views/YourHome.xaml.cs
AndyMDoyle: One is looping over DrivesManager, and the other CloudDrivesManager.

AndyMDoyle:
@yair100 @gave92, please give this update a try to see if it addresses the tab bar issue on your system.

The population of AppSettings, InteractionViewModel, and SidebarPinnedController are now performed from App.xaml.cs when the ```OnLaunched``` and ```OnActivated``` events occur, and when the app is resumed from a suspended state. If these objects are already populated the work isn't done again to improve performance.

InteractionViewModel, SidebarPinnedController, and TerminalController have also been refactored slightly to allow async initialization to ensure these sort of dependencies are fully constructed before the application startup gets to the point where they are required.

Other changes in this update...

- Cloud Drives have been moved above Drives in thee sidebar
- Local drives retain the original order
- PR comments addressed

yair100 (dismissed):

yaichenbaum:
[CodeFactor](https://www.codefactor.io/repository/github/files-community/files) found an issue: A closing curly bracket must not be preceded by a blank line.

It's currently on:
[Files\Filesystem\CloudDrivesManager.cs:191
](https://github.com/files-community/files/blob/11c40efe64690c5cd71eb82b5915dfca81c2233a/Files/Filesystem/CloudDrivesManager.cs#L191)Commit 11c40efe64690c5cd71eb82b5915dfca81c2233a

yaichenbaum:
[CodeFactor](https://www.codefactor.io/repository/github/files-community/files) found an issue: A closing curly bracket must not be preceded by a blank line.

It's currently on:
[Files\Filesystem\Drives.cs:160
](https://github.com/files-community/files/blob/11c40efe64690c5cd71eb82b5915dfca81c2233a/Files/Filesystem/Drives.cs#L160)Commit 11c40efe64690c5cd71eb82b5915dfca81c2233a

AndyMDoyle:
> [CodeFactor](https://www.codefactor.io/repository/github/files-community/files) found an issue: A closing curly bracket must not be preceded by a blank line.
> 
> It's currently on:
> [Files\Filesystem\CloudDrivesManager.cs:191 ](https://github.com/files-community/files/blob/11c40efe64690c5cd71eb82b5915dfca81c2233a/Files/Filesystem/CloudDrivesManager.cs#L191)Commit [11c40ef](https://github.com/files-community/Files/commit/11c40efe64690c5cd71eb82b5915dfca81c2233a)

Resolved by removing an unused method earlier.

AndyMDoyle:
> [CodeFactor](https://www.codefactor.io/repository/github/files-community/files) found an issue: A closing curly bracket must not be preceded by a blank line.
> 
> It's currently on:
> [Files\Filesystem\Drives.cs:160 ](https://github.com/files-community/files/blob/11c40efe64690c5cd71eb82b5915dfca81c2233a/Files/Filesystem/Drives.cs#L160)Commit [11c40ef](https://github.com/files-community/Files/commit/11c40efe64690c5cd71eb82b5915dfca81c2233a)

Fixed

Files/Views/MainPage.xaml.cs
AndyMDoyle: @gave92 does the latest commit give you the same white screen?
Creatings SettingsViewModel is expensive, and I am already testing some ideas to split DrivesManager and CloudDrivesManager out and make them properties of App; with all dependencies being constructed in parallel to minimise startup time.

AndyMDoyle:
@gave92 @yair100, is there anything else you need from me for this PR?

Once merged, I have a new PR to raise based on this that has improved startup performance a lot. On my system it's probably less than 2 seconds from clicking the icon to seeing everything in Release mode (debug mode is still around 8 seconds and most of that is down to XAML rendering). I have done this by moving DrivesManager and CloudDrivesManager out of SettingsViewModel which shaved off a little time, but the big improvement came from async instantiating CloudDriveManager and async loading cloud drive providers.

gave92:
@AndyMDoyle looks good, no duplicates, no white screen, tabs are back.
Opened a small PR with some cleanup, let me know what you think.

AndyMDoyle:
@gave92, your cleanup PR looks good. I have added one minor fix in addition as I found if the app suspends there was an issue with the drive watcher when resuming.

gave92 (approved):

yaichenbaum:
@AndyMDoyle Thank you for your contribution 🚀

PR #3235, opened 2021-01-23
Added ability to paste images from clipboard directly into folder

### LINKED ISSUE #3029: Paste images from clipboard into folder
Windows Explorer (only in preview builds though) allows you to paste images copied from your clipboard directly into a folder, and saves it as an image. This is a useful feature, especially when working with screen snips.

It would be nice if Files could have a similar feature. 

Save from clipboard button separate from paste functionality.

**Additional context**
Add any other context or screenshots about the feature request here. Ex: What device type/UI layout and app version?

winston-de:
Closes #3029

Images pasted are saved as pngs with the current date and time as the name.

### DIFF
--- a/Files/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
+++ b/Files/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
@@ -539,7 +539,7 @@ namespace Files.Filesystem
                     using var imageStream = await imgSource.OpenReadAsync();
                     var folder = await StorageFolder.GetFolderFromPathAsync(destination);
                     // Set the name of the file to be the current time and date
-                    var file = await folder.CreateFileAsync($"{DateTime.Now:mm-dd-yy-HHmmss}.png");
+                    var file = await folder.CreateFileAsync($"{DateTime.Now:mm-dd-yy-HHmmss}.png", CreationCollisionOption.GenerateUniqueName);
 
                     using var stream = await file.OpenAsync(FileAccessMode.ReadWrite);
                     await imageStream.AsStreamForRead().CopyToAsync(stream.AsStreamForWrite());
--- a/Files/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
+++ b/Files/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
@@ -13,7 +13,9 @@ using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Windows.ApplicationModel.DataTransfer;
+using Windows.Graphics.Imaging;
 using Windows.Storage;
+using Windows.Storage.Streams;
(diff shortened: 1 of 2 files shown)


Files/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
```diff
@@ -505,33 +507,53 @@ public async Task<ReturnResult> CopyItemAsync(IStorageItemWithPath source, strin
...
+            if (packageView.Contains(StandardDataFormats.Bitmap))
             {
-                destinations.Add(Path.Combine(destination, item.Name));
+                try
+                {
+                    var imgSource = await packageView.GetBitmapAsync();
+                    using var imageStream = await imgSource.OpenReadAsync();
+                    var folder = await StorageFolder.GetFolderFromPathAsync(destination);
+                    // Set the name of the file to be the current time and date
+                    var file = await folder.CreateFileAsync($"{DateTime.Now:mm-dd-yy-HHmmss}.png");
```
TheXamlGuy, line 542: If I were you I would be doing a filename check after generating the DateTime stamp to be use for the file name in event where the user may end up pressing ctrl + v in repetitive succession. If there was a case of a duplicated filename then it would be ideal to append a numeric index depending on the number of duplicates at the end of the filename, i.e. 01-23-2021 (1).

gave92: Or just `var file = await folder.CreateFileAsync($"{DateTime.Now:mm-dd-yy-HHmmss}.png", CreationCollisionOption.GenerateUniqueName);`

TheXamlGuy: > Or just `var file = await folder.CreateFileAsync($"{DateTime.Now:mm-dd-yy-HHmmss}.png", CreationCollisionOption.GenerateUniqueName);`

Nice and simpler then 😁

Files/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
```diff
@@ -505,33 +507,53 @@ public async Task<ReturnResult> CopyItemAsync(IStorageItemWithPath source, strin
...
+
+                returnStatus = await CopyItemsAsync(source, destinations, registerHistory);
+
+                return returnStatus;
             }
-            ReturnResult returnStatus = ReturnResult.InProgress;
 
-            List<string> destinations = new List<string>();
-            foreach (IStorageItem item in source)
+            if (packageView.Contains(StandardDataFormats.Bitmap))
```
gave92, line 534: Perhaps this should be changed to `else if` in case the clipboard contains both file and image formats (not sure in any program does that when copying images)

gave92 (approved):

yaichenbaum:
:shipit:

PR #3245, opened 2021-01-24
Fixed issue where scrolling in details view was broken

winston-de:
The bug was caused by the DataGrid being placed in the wrong row.
Also removed row/column definitions, as they are unused.


Files/Views/LayoutModes/GenericFileBrowser.xaml
```diff
@@ -673,7 +663,7 @@
 
         <controls:DataGrid
             x:Name="AllView"
-            Grid.Row="3"
+            Grid.Row="0"
```
duke7553, line 666: Wouldn't we remove this altogether if there are no rows/columns set on the Grid?

duke7553 (approved):
LGTM

PR #3257, opened 2021-01-24
Improve startup performance

AndyMDoyle:
This PR improves startup performance and gets the user to an interactive state quicker by making the following changes:

- `CloudDrivesManager` and `DrivesManager` have been removed from `SettingsViewModel` to allow app settings to be loaded and made available to launch the app as early as possible.
- `CloudDrivesManager` and `DrivesManager` are instantiated using an async singleton pattern with them being constructed as early as possible, while not delaying the launch of the UI should one of them require longer to complete.
- Certain initial dependencies are created in parallel to shave off a little more startup time.
- `ICloudProviderDetector`'s are loading asynchronously in parallel so that `CloudDrivesManager` only takes as long as the slowest `ICloudProviderDetector`. This is mainly where `OneDriveCloudProvider` relies on FullTrust and can be delayed by FullTrust loading slowly. With this design, all other `ICloudProviderDetector`'s can complete while waiting for the slowest one.

In future there would be further opportunities to speed up the initial load by adding cloud drives to the UI individually as they load so that fast `ICloudProviderDetectors` appear straight away, and slower ones will gracefully appear in the UI a little later. However for now I don't feel any of the `ICloudProviderDetectors`'s are slow enough to require even more changes.

### DIFF
--- a/Files/ViewModels/SettingsViewModel.cs
+++ b/Files/ViewModels/SettingsViewModel.cs
@@ -838,10 +838,5 @@ namespace Files.ViewModels
         private delegate bool TryParseDelegate<TValue>(string inValue, out TValue parsedValue);
 
         #endregion ReadAndSaveSettings
-
-        public void Dispose()
-        {
-            App.DrivesManager?.Dispose();
-        }
     }
 }
--- a/Files/ViewModels/SettingsViewModel.cs
+++ b/Files/ViewModels/SettingsViewModel.cs
@@ -29,8 +29,6 @@ namespace Files.ViewModels
     {
         private readonly ApplicationDataContainer localSettings = ApplicationData.Current.LocalSettings;
 
-        public DrivesManager DrivesManager { get; private set; }
-
         public CloudDrivesManager CloudDrivesManager { get; private set; }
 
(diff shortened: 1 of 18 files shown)


Files/ViewModels/SettingsViewModel.cs
```diff
@@ -847,7 +841,7 @@ public TValue Get<TValue>(TValue defaultValue, [CallerMemberName] string propert
 
         public void Dispose()
         {
-            DrivesManager?.Dispose();
+            App.DrivesManager?.Dispose();
```
yaichenbaum, line 844: Do we still need this reference?

gave92: Agreed, this Dispose method can go away. In `OnSuspending` change `AppSettings?.Dispose();` to `DrivesManager?.Dispose()`

AndyMDoyle: Updated

yaichenbaum (approved):

gave92 (approved):

yaichenbaum:
:shipit:

Issue #3259, opened 2021-01-24
Excessive whitespace above the files list when a folder contains a small number of items


AndyMDoyle:
If a folder contains a fewer items than would require the file list to scroll, the file list appears docked to the bottom of the UI and a lot of whitespace appears above it.

![image](https://user-images.githubusercontent.com/307719/105633499-d80c2000-5e50-11eb-92e5-df5dff8d42cb.png)

Likewise, if the folder is empty, the file list is at the bottom of the UI.

![image](https://user-images.githubusercontent.com/307719/105633505-e3f7e200-5e50-11eb-8602-9f5349782345.png)

I am seeing this issue in the latest code in the main branch.

AndyMDoyle:
Addressed in #3245

PR #3261, opened 2021-01-24
Fix: Excessive whitespace above the files list when a folder contains a small number of items

AndyMDoyle:
This fix is to address issue #3259


XPoppyX:
#3245 This is not the same fix?

AndyMDoyle:
@XPoppyX, looks like it.

I didn't go as far as removing the grid row and column definitions completely though as they were added recently as part of a bigger PR.

PR #3297, opened 2021-01-26
Tab header icon: Implemented checks for all drive types.

### LINKED ISSUE #3195: Tab header icon:  Improve drive type check method
Currently to show the drive icon in the tab header, the type is checked only by the letter.

**Drive letter: A or B** = Floppy disk icon

**Any other letter** = Hard disk icon

suvrat29:
Implemented checks for all the drive types and assigned the relevant icons to them. Closes #3195

### DIFF
--- a/Files/Views/MainPage.xaml.cs
+++ b/Files/Views/MainPage.xaml.cs
@@ -5,6 +5,7 @@ using Files.UserControls.MultitaskingControl;
 using Files.ViewModels;
 using Microsoft.Toolkit.Uwp.Extensions;
 using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.IO;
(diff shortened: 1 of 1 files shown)


tsvietOK (dismissed):

tsvietOK (approved):

Files/Views/MainPage.xaml.cs
```diff
@@ -492,5 +501,54 @@ private void HorizontalMultitaskingControl_Loaded(object sender, RoutedEventArgs
         {
             MultitaskingControl = HorizontalMultitaskingControl;
         }
+
+        private static string GetDriveTypeIcon(DriveInfo drive)
```
yaichenbaum, line 505: Does this has to be in MainPage.cs?

suvrat29: It can be moved to another page since it is a static function.  Please let me know if the function needs to be moved to another page.

yaichenbaum: That would be best.

suvrat29: Okay. I'll move it somewhere else. Is there a specific file where you would have it placed? Or is the Interaction.cs file okay?

yaichenbaum: @tsvietOK Any ideas?

tsvietOK: I think we could keep it as it is, as it is used here.  By the way, SetSelectedTabInfoAsync should be moved too.

yaichenbaum: It's best if it's moved so that we can reuse it, I meant ideas where to move it to.

yaichenbaum:
@suvrat29 Great work!

suvrat29:
Thank you! 🙂

PR #2881, opened 2021-01-07
Bundles!!

d2dyno1:
This PR brings Bundles Widget - Grouped containers of files/folders for quick access from the home page.

**Example usage**
- Group your useful programs or games with a little help from bundles.
- Access your workload files quickly without browsing a single folder!

<br/>

**Features of this PR**
- MVVM Ready
- `IDisposable` Ready
- Flexible `BaseJsonSettingsViewModel` for serialization!
- `DynamicDialog` for dynamic dialogs - any dialog could be created off of it! (MVVM Ready)
- Overhauled `Interaction.OpenSelectedItems` -> `Interaction.OpenPath`
- Minor Improvements
- Bundles UI

**Things yet to do**
- [x] Improve UI
- [x] Setting for disabling/enabling Bundles
- [x] Add "Add Bundle" UI
- [x] Remove, Rename Bundles
- [x] Remove items from Bundles
- [x] Folder Icons
- [x] Icons don't show bruh
- [x] Right click option to open folders in new tab
- [x] View does not update when performing actions with bundles
- [x] Drag and Drop does not cover entire Bundle area
- [x] Show big add button when there are no Bundles
- [x] Big Add Bundle button is not being hidden
- [x] Replace `Interaction.OpenSelectedItems()` with `Interaction.OpenPath()` [Help wanted]
- [x] Add Translation strings [Help wanted]
- [x] Add option to export bundles configuration file
- [x] Add option to import bundles
- [x] Fix columns plz  @yair100 
- [x] Remove the reveal border from bundles [Help wanted]
- [x] Remove the reveal border for items in bundles [Help wanted]
- [x] Fix code style 🤯

**Screenshots (constantly updated)**
image
image

**Future Bundles PR #3637**
- Option to rearrange Bundle items by dragging
- Select items upon `OpenPath` navigation
- Ability to rearrange Bundles
- TBD: Support for adding shourtcut files
- Display warning when creating a Bundle if the name already exists #3449
<br/>
Closes #1465

PR #3301, opened 2021-01-26
Replaced some dialogs with DynamicDialog

d2dyno1:
#2881 introduced `DynamicDialog` - a flexible dialog whose content can be adjusted dynamically from code. This PR replaces simple, existing dialogs with DynamicDialog. These new dynamically created dialogs work just as good but with less code, less files and with more consistent MVVM design.


d2dyno1:
Ready to be reviewed

yaichenbaum:
@d2dyno1 Can you post before and after screenshots of the dialogs?

d2dyno1:
After:
image
image
image
image

@yair100

d2dyno1:
@yair100 this PR focuses on codebase quality and utilizes consistent MVVM design

tsvietOK:
@d2dyno1 What are the advantages of this DynamicDialog?

d2dyno1:
DynamicDialog can be implemented to replace any other dialog. DynamicDialog aims to remove additional files and gets rid of duplicated code. It's also desined with MVVM design as opposed to existing dialogs. Furthermore, since we have the control of the base [DynamicDialog] any changes made to it will be reflected with other dialogs.
DynamicDialog was originally implemented with #2881 and used throughout Bundles' classes. This PR further improves it and implements it in other places.

@tsvietOK

tsvietOK:
I am not sure about this "improvement". As example, some UI elements generating inside DynamicDialogFactory.cs. It is better to add them via xaml, as it is now.
Before:
image
image
After: size of the dialog changes
image
image
 In case of more complicated Dialog structure (a lot of buttons, data checks, etc.), it will be pretty hard to add it via DynamicDialog. @yair100 what do you think ?

yaichenbaum:
@tsvietOK Complex dialogs can still be custom, but for most dialogs, this is probably a better solution. However, it looks like some more work is needed to make sure the primary buttons use the accent color on the new dialogs before this can be merged.

d2dyno1:
@yair100 All instances of DynamicDialog already use accent color for primary button
image

(I'll work on fixing the unexpected resize with DynamicDialog(RenameDialog) soon) (as @tsvietOK pointed out)
And yes, DynamicDialog is intended to be used on a smaller scale, complex dialogs should still be created separately

d2dyno1:
@tsvietOK the resize has been resolved -> there's now a fixed size

d2dyno1:
@yair100 ready!

Files/ViewModels/Dialogs/DynamicDialogViewModel.cs
```diff
@@ -219,16 +385,19 @@ public DynamicDialogViewModel()
         {
             // Create default implementation
             TitleText = "DynamicDialog";
-            PrimaryButtonText = "OK";
+            PrimaryButtonText = "Ok";
```
tsvietOK, line 388: I guess it is should be translated.

d2dyno1: It's a fallback value and should always be assigned when creating the dialog. This value should not be used, and I've only kept it here so the button is visible

Files/Views/Pages/PropertiesDetails.xaml.cs
```diff
@@ -36,23 +37,24 @@ public async Task<bool> SaveChangesAsync()
...
                     return true;
                 }
                 catch
                 {
-                    switch (await dialog.ShowAsync())
+                    await dialog.ShowAsync();
+                    switch (dialog.DynamicResult)
                     {
-                        case ContentDialogResult.Primary:
+                        case ViewModels.Dialogs.DynamicResult.Primary:
```
tsvietOK, line 51:
```suggestion
                        case ViewModels.Dialogs.DynamicDialogResult.Primary:
```

Files/ViewModels/Dialogs/DynamicDialogViewModel.cs
```diff
@@ -10,31 +11,59 @@ namespace Files.ViewModels.Dialogs
 {
     public enum DynamicResult
```
tsvietOK, line 12: Move it to Enums folder.

Files/ViewModels/Dialogs/DynamicDialogViewModel.cs
```diff
@@ -10,31 +11,59 @@ namespace Files.ViewModels.Dialogs
...
     {
-        Primary = 0,
-        Secondary = 1,
-        Cancel = 2
+        Primary = 1,
+        Secondary = 2,
+        Cancel = 4,
     }
 
     public enum DynamicButtons
```
tsvietOK, line 19: Move it to Enums folder.

d2dyno1: kk, done!

d2dyno1:
@tsvietOK anything left todo? ✅

tsvietOK (approved):

Issue #2577, opened 2020-12-09
A11y: Invoking "Ongoing tasks" button does not read out of opened flyout


chingucoding:
**Describe the bug**
When invoking the ongoing tasks button, users are not notified about the content of the opened flyout. Generally, when invoking a button, users should be notified of changes, e.g. reading out something or switching focus if applicable

**To Reproduce**
1. Visit a folder
2. Invoke ongoing task button
3. Observe nothing gets announced

**Expected behavior**
Ideally, narrator should announce the changes. That is something that the WinUI InfoBar control already does: https://github.com/microsoft/microsoft-ui-xaml/blob/50f29e79da9b457d9b136293d82ff21b8340bf2a/dev/InfoBar/InfoBarAutomationPeer.cpp#L24

So in the case of Files, you should raise a [notification event](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer.raisenotificationevent?view=winrt-19041) with [processingkind](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationnotificationprocessing?view=winrt-19041) ImportantMostRecent.

**Screenshots**

**Desktop (please complete the following information):**
- OS: Insider Build 10.0.20262 Build 20262
- App version: 0.21.1.0

yaichenbaum:
@chingucoding This issue should be resolved with #3312.

marcelwgn:
Thank you @winston-de and @yair100 for fixing this!

Issue #2570, opened 2020-12-09
A11y: File/folder list is missing UIA names, faulty controltype & localizedcontroltype 


chingucoding:
**Describe the bug**
The default file/folder list is missing UIA names for multiple things:
1. The image in the first column
2. The second hidden image which get's reported by accessibility insights and narrator
3. The value for the "Size" column if there is no size present

In addition to that, the first column of the file/folder view is using the "Custom" controltype and is not providing a localized control type, resulting in users having no idea what the element actually is.

Also, the status column is also affected by this:
1. Control with controltype "custom" and no localized control type

Here are screenshots to clarify what I mean:

![Folder view with accessibility issues highlighted](https://user-images.githubusercontent.com/16122379/101672035-a0d6e080-3a55-11eb-9f24-39f1011d59b8.png)

![Image showing the control tree of the accessibility issues in question](https://user-images.githubusercontent.com/16122379/101672107-bcda8200-3a55-11eb-8998-a926622b765d.png)

![Image of folder with status column issue pointed out](https://user-images.githubusercontent.com/16122379/101676603-aa634700-3a5b-11eb-932b-a6800f88abae.png)

**To Reproduce**
1. Open Files
2. Open a folder in Files
3. Use narrator item mode to narrate the entire line of an item
4. Observe the narrated content

**Expected behavior**
In the case of the size column, it might make sense to announce something like "not calculated" instead of nothing.
In the case of the custom controltype without localized controltype. either use non custom controltype, provide a localized contorltype or even remove it from the control tree if it does not add information to users.

Not sure why there are two images in the control tree, this seems like a bug. Maybe remove them from the control tree too if it makes no sense do discern the first column as multiple parts?

**Desktop (please complete the following information):**
- OS: Insider Build 10.0.20262 Build 20262
- App version: 0.21.1.0

**Additional context**
Add any other context about the problem here. For instance, are you building from source or using a precompiled build/snapshot?

**Log file**
Please post the log file here so that we can understand your problem better. You can access it from Settings->About->Open log location.
[debug.txt](https://github.com/files-community/Files/files/5667956/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5667957/debug_fulltrust.txt)

yaichenbaum:
@winston-de Can you confirm if this is resolved in `main`?

winston-de:
[...]

Issue #2563, opened 2020-12-09
Banners in the status center are missing AutomationProperties


yaichenbaum:
The banners in the status center don't have AutomationProperties set, causing the screen reader to read out unhelpful information.

PR #3312, opened 2021-01-27
Accessibility improvements

### LINKED ISSUE #2919: A11y: Tab order lands on element "Pane" with no focus indicator
**Describe the bug**
When opening a folder and tabbing through the UI, I noticed that focus can land on an element that is being announced "Pane" without a more descriptive name that also has no focus visual:

**To Reproduce**
1. Open folder in Files (in my case desktop)
2. Tab through entire app
3. Start narrator
4. Notice elements visited and announced
4. Notice pane without name or focus indicator

**Expected behavior**
Either give it a proper name and add focus indicator or skip element in tab order
**Screenshots**
Gif showing faulty tab order where focus rect moves through app and then suddenly disappears

 - OS Version: Win10 10.0.19042
 - App version: [1.0.0.0]

**Additional context**
Add any other context about the problem here. Does this problem occur again after restarting the app?

**Log file**
Please upload the log file here so that we can understand your issue better. You can access it from Settings->About->Open log location. You can also find the log file manually by following [this](https://github.com/files-community/Files/wiki/Opening-the-log-file) guide.
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5797991/debug_fulltrust.txt)
[debug.txt](https://github.com/files-community/Files/files/5797992/debug.txt)

winston-de:
This PR closes a few accessibility issues from the to-do list.
Closes #2919, #2577, and #2563.

Also added "not calculated" text when there is no file size, as requested in #2570.
I also did the groundwork for labeling the sync status, but I have been unable to get automation properties to work with the DataGrid.


Files/UserControls/StatusBarControl.xaml
```diff
@@ -46,7 +46,7 @@
             x:Uid="StatusCenterTeachingTip"
             Title="Status Center"
             Subtitle="Check the status of file operations here"
-            Target="{x:Bind StatusCenter}" />
```
yaichenbaum, line 49: Was this property removed on purpose?

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -337,11 +337,13 @@ private void TextBox_TextChanged(object sender, TextChangedEventArgs e)
 
             if (FilesystemHelpers.ContainsRestrictedCharacters(textBox.Text))
             {
+                FileNameTeachingTip.Visibility = Visibility.Visible;
                 FileNameTeachingTip.IsOpen = true;
             }
             else
             {
                 FileNameTeachingTip.IsOpen = false;
+                FileNameTeachingTip.Visibility = Visibility.Collapsed;
```
tsvietOK, line 346: What is the purpose of this change? I think setting `IsOpen` to false is enough to hide tip.

Files/Views/LayoutModes/GenericFileBrowser.xaml
```diff
@@ -980,7 +983,7 @@
                     x:Uid="sizeColumn"
                     Width="Auto"
                     MinWidth="100"
-                    Binding="{Binding FileSize}"
+                    Binding="{Binding FileSizeDisplay}"
```
yaichenbaum, line 986: Perhaps the AutomationProperties name should say "Not Calculated" and the size should remain blank.

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
winston-de: @tsvietOK If you don't collapse it, it shows up on the tab stack. This change fixes #2919

Files/Views/LayoutModes/GenericFileBrowser.xaml
winston-de: @yair100 I was unsuccessful in getting automation properties to work with the DataGrid. The screen reader wouldn't read out the name when the column was selected. I had the same issue with trying to add a name to the sync status glyph. If you want, I could switch back and we can try to implement this at a later date (Perhaps when the DataGrid is modernized?).

yaichenbaum: Let's to that and wait on the new DataGrid control to fix this issue.

Files/Filesystem/ListedItem.cs
```diff
@@ -138,6 +138,7 @@ public string ItemType
 
         public string FileExtension { get; set; }
         public string FileSize { get; set; }
+        public string FileSizeDisplay => FileSize ?? "ItemSizeNotCalcluated".GetLocalized();
```
yaichenbaum, line 141: Can you remove this string?

Files/Strings/en-US/Resources.resw
```diff
@@ -1788,4 +1788,28 @@
...
+  </data>
+  <data name="CloudDriveSyncStatus_Sync" xml:space="preserve">
+    <value>Syncing</value>
+  </data>
+  <data name="CloudDriveSyncStatus_Excluded" xml:space="preserve">
+    <value>Excluded from sync</value>
+  </data>
+  <data name="ItemSizeNotCalcluated" xml:space="preserve">
+    <value>Not calculated</value>
+  </data>
```
yaichenbaum, line 1811: Can you remove this string?

winston-de: @yair100 Shouldn't this be kept for when #2570 is eventually able to be implemented?

yaichenbaum (approved):

yaichenbaum:
@winston-de Great work! Accessibility is a high priority for Files, so these kinds of pull requests help Files stay true to its vision.

PR #3319, opened 2021-01-27
Localize tab headers from desktop.ini / Support user level folder name localization for tabs.

### LINKED ISSUE #3313: Tab title does not follow the localization settings
**Describe the bug**
The title of the opened folder does not follow the localized folder name set in the desktop.ini file, even if it has been pinned to the sidebar and its localized name has been correctly identified in the sidebar.

**To Reproduce**
1. Go to a sidebar folder
2. Edit desktop.ini file
3. Modify `LocalizedResourceName` entry
4. See that its sidebar name is different from the tab title name

**Expected behavior**
See the localized name of this folder in the tab title.

**Screenshots**
image

 - OS Version: Windows 10 20H2 19042.746
 - App version: v1.1

**Log file**
Nothing useful.

suvrat29:
Implement the ability to pick up user localized names from desktop.ini for the tabs' text. Closes #3313 .


suvrat29:
@yair100 I created a test code block for that issue, but I'm not really confident that this is feasible. In essence, it just parses the desktop.ini and grabs the string that is to be assigned as the tab name for localization. Can someone take a look at this an let me know if it is okay or not?

suvrat29:
There is a native function (although in the native 32 bit library) but I'm not sure if I can implement it/can be implemented.

`[DllImport("kernel32")]
private static extern int GetPrivateProfileString(string section, string key,string def, StringBuilder retVal, int size,string filePath);`

suvrat29:
image
Translated tab name

fxzxmicah:
@suvrat29 Why can the sidebar get the localized name normally?

suvrat29:
> @suvrat29 Why can the sidebar get the localized name normally?

The reason for that is sidebar items already have their translated names present in the resource files. We don't have to read those from  any external files (like desktop.ini)

gave92:
@suvrat29 sidebar shows the localized name by getting a StorageFolder for the path and showing its "DisplayName" property.

fxzxmicah:
> @suvrat29 sidebar shows the localized name by getting a StorageFolder for the path and showing its "DisplayName" property. You could do the same for the tabs.

@suvrat29 Yes, I think it should be similar to what he said, because when I added a new folder to the sidebar after adding a desktop.ini file, the sidebar automatically recognized the localized name I set in the desktop.ini file (more precisely, a custom display name, not necessarily a translated name).

suvrat29:
> @suvrat29 sidebar shows the localized name by getting a StorageFolder for the path and showing its "DisplayName" property. You could do the same for the tabs.

Okay, I'll look into that. Does StorageFolder's DisplayName support INI translations?

gave92:
@suvrat29 the DisplayName corresponds to the name Window Explorer shows for a folder, so it should support everything explorer does (desktop.ini included)

suvrat29:
> @suvrat29 the DisplayName corresponds to the name Window Explorer shows for a folder, so it should support everything explorer does (desktop.ini included)

Ah, that makes sense. I'll implement those changes later today.

suvrat29:
@gave92 I implemented the StorageFolder implementation, but it seems that it has some issues with encoding.

With StorageFolder implementation
i2

With my old implementation
i1

gave92:
@suvrat29 could you try to set the default windows language to match that file name? I believe you have now set english, could you try setting japanese? (If that's not japanese I apologise xD)

suvrat29:
> @suvrat29 could you try to set the default windows language to match that file name? I believe you have now set english, could you try setting japanese? (If that's not japanese I apologise xD)
> Windows Settings > Time & language > Language

Sure, no problem

suvrat29:
First of all, apologies for the late responses, I've been a bit preoccupied with some other stuff.

@gave92 I tried what you said, but my windows license is single language only. But, I was able to the language from control panel as Japanese. While the situation was a bit better, it was not accurate. 
tt

However, I believe it is because of my language settings/windows setup. So with your permission, I think we should go with the display name route and if you're okay with it, I'll go ahead and commit the code with display name changes.

gave92:
@suvrat29 sure go ahead 👍

suvrat29:
> @suvrat29 sure go ahead 👍
> 
> Edit: @suvrat29 remember to test the changes with hidden folders, for which StorageFolder methods don't work

Oh, okay. I'll tetst.

gave92:
@suvrat29 nevermind, I just saw your changes and the try/catch covers it.

```cs
var item = await FilesystemTasks.Wrap(() => DrivesManager.GetRootFromPathAsync(path));
var res = await FilesystemTasks.Wrap(() => StorageFileExtensions.DangerousGetFolderFromPathAsync(path, item));
if (res)
{
}
```

fxzxmicah:
> First of all, apologies for the late responses, I've been a bit preoccupied with some other stuff.
> 
> @gave92 I tried what you said, but my windows license is single language only. But, I was able to the language from control panel as Japanese. While the situation was a bit better, it was not accurate.
> tt
> 
> However, I believe it is because of my language settings/windows setup. So with your permission, I think we should go with the display name route and if you're okay with it, I'll go ahead and commit the code with display name changes.

@suvrat29 For Chinese, desktop.ini files only support GBK encoding, while UTF-8 encoding also shows messy code in Explorer on the Chinese system. So it should not be a problem as long as the user is set up properly.

suvrat29:
@gave92 Ah, that does looks more elegant. I'll integrate these changes.

suvrat29:
@fxzxmicah Oh, that makes sense. That should not be a problem at all then. Thanks for the info ^_^

suvrat29:
Changes are implemented.

suvrat29:
@yair100

yair100 (dismissed):

Files/Views/MainPage.xaml.cs
```diff
@@ -354,12 +355,33 @@ private static async Task SetSelectedTabInfoAsync(TabItem selectedTabItem, strin
                     tabLocationHeader = currentPath.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar).Split('\\', StringSplitOptions.RemoveEmptyEntries).Last();
                 }
             }
+
             if (tabHeader != null)
             {
                 tabLocationHeader = tabHeader;
             }
+
+            FilesystemResult<StorageFolderWithPath> currentItem = await FilesystemTasks.Wrap(() => DrivesManager.GetRootFromPathAsync(currentPath));
```
gave92, line 364: Should this be put inside the `else` at line 352? Here it may overwrite the header set for special folders.

gave92:
@suvrat29 I pushed a commit here that adds support for showing the localized name also when using dual-pane mode. Hope you don't mind ;)

suvrat29:
> @suvrat29 I pushed a commit here that adds support for showing the localized name also when using dual-pane mode. Hope you don't mind ;)

Not at all :) It has been pending for almost a month. I'm not able to work/contribute since my last commit due to other obligations. Thanks for adding the support.

suvrat29:
@gave92 If that's all that is needed, maybe we can finally merge it?

gave92 (approved):

gave92:
@suvrat29 For me it's a yes. Thanks for your contribution.

suvrat29:
> @suvrat29 For me it's a yes. Thanks for your contribution.

Always happy to help

PR #3344, opened 2021-01-29
Improve performance while occurring a large set of entry changes

hez2010:
Instead of calling `ApplyFilesAndFoldersChangesAsync` directly after each filesystem change, accumulate those changes and update them one time with an interval sampler to improve performance while occurring a large set of entries changes.


yaichenbaum (approved):

yaichenbaum:
@hez2010 Nice change!

PR #3346, opened 2021-01-29
Create codeql-analysis.yml

yaichenbaum:


h1dden-da3m0n:
sorry to bump this closed PR but I got CodeQL [working](
https://github.com/h1dden-da3m0n/Files/actions) for this project and could open a PR for it if you want me to (after some workflow cleanup).
HOWEVER, I am unsure if it would not be better to install a static code analysis workflow such as [Codacy](https://github.com/codacy/codacy-analysis-cli-action) since that should be capable to give also maintainability suggestions. (CodeQL has such a mode too, however it failed to produce any meaningful output when I tried it and referred to generated code classes instead)

yaichenbaum:
@h1dden-da3m0n Thanks for reaching out, a PR for CodeQL would be great! I would like to try CodeQL first to see how it works, and if we feel a need for something more, than we can look into alternatives.

h1dden-da3m0n:
Just keep in mind this will run for ~25min each PR and Push to `Main`

I'll clean up the workflow then and fire u a PR :wink:

yaichenbaum:
@h1dden-da3m0n We already have a separate pipeline that takes around 20 minutes to run so an extra 5 minutes shouldn't be a big deal.

Issue #3231, opened 2021-01-22
Drives not showing up on home screen


KibSquib48:
**Describe the bug**
Drives do not appear on home screen until opening one of the library folders, and sometimes not even that works.

**To Reproduce**
1. Open files
2. Go to click on the C: Drive
3. wait where did it go

**Expected behavior**
The buttons for the C: drive and all other drives should be there

**Screenshots**
![image](https://user-images.githubusercontent.com/33189136/105526413-5f6e5d80-5cb0-11eb-8672-1239934b84db.png)

**Desktop (please complete the following information):**
 - OS Version: Windows 10 20H2 19042.782
 - App version: 1.1.5.0

**Additional context**
Terminating, Resetting, and Repairing from the Settings app does not fix it, and neither does a simple close and reopen.

**Log file**
[debug.txt](https://github.com/files-community/Files/files/5857617/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5857619/debug_fulltrust.txt)

gave92:
@KibSquib48 does it work when you type a drive path e.g. "C:\\" in the navigationbar?

SimonCDalzell:
Typing in the path of one of the non-showing drives works and contents is displayed correctly.

gave92:
@SimonCDalzell do you experience the same? Is all your drives missing or just some? Could you post the log files a screenshot of the sidebar?
Are the drive missing from the sidebar only or from the home page too?

KibSquib48:
> @KibSquib48 does it work when you type a drive path e.g. "C:\" in the navigationbar?
> Did the missing drive by chance reappear after installing OneDrive?

Installing OneDrive fixed it, and typing C:\ worked too

SimonCDalzell:
> @SimonCDalzell do you experience the same?
> 
> * are all your drives missing or just some?
> * could you post the log files and a screenshot of the sidebar?
> * are the drive missing from the sidebar only or from the home page too?

Just some. I believe the two drives that appear correctly have AHCI Hotswap enabled in the bios, while the ones that do not show have it disabled. I can test this in a sec. 
I've attached the logs and a screenshot
The drives are missing from the sidebar and from the home page.

Probably worth mentioning that I'm on Windows 1809.

![Capture](https://user-images.githubusercontent.com/19229302/105580129-3ffc3100-5d93-11eb-9957-00e14bb93121.JPG)
[...]

PR #3350, opened 2021-01-29
Improve drive detection and UI sync to deal with duplicate entries

AndyMDoyle:
#3231 

Primary change is making the app service connection a singleton which means multiple messages can be sent without issues launching additional processes. A lot of this has been taken from a @gave92's network drives PR and should make that PR a smaller update.

The logic to add local drives and cloud drives to the side bar has also been refactored to ensure this is done on the UI thread and that a lock is obtained to prevent multiple updates to the side bar items at the same time. Now a lot of the code is async, this is done using a `SemaphoreSlim` to ensure one thread at a time updates the side bar items.

So far this update has been 100% reliable in debug mode, release mode, and with SharePoint sync folder support on top which results in an additional concurrent full trust call at startup.


Files/Views/MainPage.xaml.cs
```diff
@@ -51,6 +53,7 @@ public TabItem SelectedTabItem
 
         public static ObservableCollection<TabItem> AppInstances = new ObservableCollection<TabItem>();
         public static ObservableCollection<INavigationControlItem> SideBarItems = new ObservableCollection<INavigationControlItem>();
+        public static SemaphoreSlim sideBarItemsSemaphore = new SemaphoreSlim(1, 1);
```
gave92, line 56: `SideBarItemsSemaphore ` since it's public

Files/Filesystem/Drives.cs
```diff
@@ -118,29 +110,39 @@ private async Task RefreshUI()
                 // Defer because UI-thread is not ready yet (and DriveItem requires it?)
                 CoreApplication.MainView.Activated += RefreshUI;
             }
+            Logger.Info("RefreshUI() complete");
```
gave92, line 113: I'd probably remove all this logs

gave92 (changes requested):

Files/Filesystem/Drives.cs
AndyMDoyle: Should be able to do without them. Easy enough to add back later for troubleshooting should they be needed for testing.

gave92:
Still duplicates :(
image
Wrong drive order:
image

AndyMDoyle:
@gave92, PR updated to remove drive ordering. I've realised I didn't see that because my drives are named in a way they end up in order of the drive letters.

Can you try and repro the duplication issue by swapping

```
_ = Task.Factory.StartNew(async () =>
{
    await DrivesManager.EnumerateDrivesAsync();
    await CloudDrivesManager.EnumerateDrivesAsync();
});
``` 

for 

```
await DrivesManager.EnumerateDrivesAsync();
await CloudDrivesManager.EnumerateDrivesAsync();
```

Task.Factory.StartNew doesn't preserve the synchronization context and has caused me issues in the past. This way the drive lists should be fully populated before MainPage is constructed.

gave92:
@AndyMDoyle yeah but I really don't want to wait for the sidebar to be fully populated before showing MainPage :)
I'm pushing here what fixes the duplicate issue *for me*.

AndyMDoyle:
Neither do I really but doing that did get rid of duplication issue for me.

Your commit looks great and I can't see any issues on my side. One question though, the existing usage of `BulkConcurrentObservableCollection` in `ItemViewModel` calls the `BeginBulkOperation()` method before working on the collection, followed by `EndBulkOperation()`. Should `BeginBulkOperation()` be called from these new places?

And in `ItemViewModel()`, `EndBulkOperation()` is called on the UI thread, so is that also needed instead of that happening on a background thread?

gave92:
Yes, EndBulkOperation() must be called on the UI thread, I should be already doing it, if not please correct :)
BeginBulkOperation() is not really mandatory in this case. It's used to pause firing events for collection changes but here I just want to call an additional `NotifyCollectionChangedAction.Reset` event when I'm done adding items to workaround the duplicate issue.
> Your commit looks great and I can't see any issues on my side
I can't believe that xD If you minimize and restore the app *withou being attached to the debugger* are you really not getting a crash?

AndyMDoyle:
@gave92, yes I am seeing a crash related to BulkConcurrentObservableCollection.OnCollectionChanged if I run without the debugger.

Trying the latest commit now.

gave92:
Hi @hez2010. In this PR we were trying to switch the MainPage SidebarItems list from ObservableCollection to the BulkConcurrentObservableCollection. Unfortunately we were getting a crash anytime the `Remove` method was called with the following error: `Parameter name: index|System.ArgumentOutOfRangeException: This collection cannot work with indices larger than Int32.MaxValue - 1 (0x7FFFFFFF - 1).`.

Do you understand why?

If you wish to test you can checkout commit 5eceb15df6c7a78e25bb654d157637cb2d5c02cf (not the latest one). To reproduce the bug you can suspend/resume the app (or minimize/restore it with no debugger attached).

AndyMDoyle:
@gave92 everything looks okay when using an ObservableCollection. Release mode starts up nice an quick and the local drives and cloud drives sections appear when ready.

hez2010:
@AndyMDoyle I've proposed a fix to `BulkConcurrentObservableCollection`, could I push the fix to your branch?

hez2010:
I think this PR is ready for merge. @gave92 could you test again to see if there're still issues remaining or performance degradation about `BulkConcurrentObservableCollection`?

hez2010:
I think this PR is ready for merge. @gave92 could you test again to see if there're still issues remaining or performance degradation about `BulkConcurrentObservableCollection`?

gave92 (approved):

yaichenbaum:
@AndyMDoyle Great work!

Issue #290, opened 2019-11-10
Toggle for Bulk Selection


szaimen:
**Is your feature request related to a problem? Please describe.**
The current Selection UX isn't working without a keyboard and is therefore not touch-optimized: So you cannot make a custom selection of folders/files with only touch input.

**Describe the solution you'd like**
Maybe you could add a "Bulk" button to the "Layout" Ribbon Section to enable Bulk selection.

**Describe alternatives you've considered**
After Pressing the button every file/folder, that you touch is getting added or removed from the selection with one click. Additionally, if https://github.com/duke7553/files-uwp/issues/94 was implemented, you than could also add a bunch of files to the selection which would be very handy. 

**Additional context**
Maybe if you already have selected a few files/folders and use the rectangle to select the same files and one additional one, the currently selected files remain selected + the additional one. Only if you select only files/folders with the rectangle, that are already selected, they should get removed from the current selection imo.

yaichenbaum:
I feel like #94 would address the issue about touch input. One of the things we are doing while creating this app is rethinking which features are needed for a file explorer app that is both powerful and simple. Just because a feature is in File Explorer it does not mean that we will add it to Files, in this case if the selection rectangle solves the problem on hand I don't think we need an extra option.
In any case thank you for bringing this up and regardless of if your ideas are implemented or not please keep them coming! They are helping us understand the direction this app needs to go in and is reminding us of all the small details that need to be addressed.

szaimen:
Actually this feature isn't implemented into file explorer to my knowledge.
It is something I thought about, that would make sense and would improve the UX for selecting files with touch input imho.

yaichenbaum:
How about check boxes at the left of every file? Once you check one then you will be able to tab on any file to select or deselect it.

szaimen:
> How about check boxes at the left of every file? Once you check one then you will be able to tab on any file to select or deselect it.

That sounds like another good way to do this. If you like this better, I am okay with it.

But I think my Proposal is design-wise cleaner (but maybe not UI- or UX-wise), because it doesn't need the checkboxed at the left on every file/folder.

yaichenbaum:
For a cleaner design how does it sound if the check boxes are hidden until the mouse is near by or someone touches near there?

szaimen:
That sounds already better to me but then you would be in the situation of having an empty column between the navigation bar on the left and the title column...

szaimen:
[...]

PR #3362, opened 2021-01-31
Added experimental multiselect feature for GridView layouts

winston-de:
Part of #290.
This pr adds an experimental feature to show a multiselect toggle in the selection options flyout. Currently, only GridView based layout modes are supported.

https://user-images.githubusercontent.com/59544401/106374581-7c9ece00-6339-11eb-851b-89680e2d04b8.mp4


yaichenbaum:
@winston-de Is it possible to make the check boxes visible on every item so users know they can select them?

winston-de:
@yair100 They're supposed to show, but it looks like the checkboxes had the same default background color as the app. 
How does this look?
image

yaichenbaum:
@winston-de That's better, is it possible to make the background transparent and show a border instead?

winston-de:
@yair100 It doesn't look like it. The only brushes available to change are the background brush and checkmark color.
https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Controls.GridViewItem?view=winrt-19041#control-style-and-template

However, it looks like that will change with the WinUI visual update. 
https://github.com/microsoft/microsoft-ui-xaml/blob/8477c990e56f379102bf3195234464092c665cf0/dev/CommonStyles/GridViewItem_themeresources_21h1.xaml#L28

Files/Views/SettingsPages/Experimental.xaml
```diff
@@ -45,6 +45,12 @@
                     Header="Cache files and folders for better performance"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{Binding UseFileListCache, Mode=TwoWay}" />
+                
+                <ToggleSwitch
+                    x:Uid="SettingsShowMultiselectOption"
+                    Header="Show multiselect in selection options menu"
```
yaichenbaum, line 51:
```suggestion
                    Header="Enable multiselect options"
```

Files/ViewModels/SettingsViewModel.cs
```diff
@@ -760,6 +769,12 @@ public SortOption DefaultDirectorySortOption
             set => Set((byte)value);
         }
 
+        public bool MultiselectEnabled
+        {
+            get => Get(false);
+            set => Set(value);
+        }
```
yaichenbaum, line 776: Should this be part of the SettingsViewModel?

winston-de: @yair100 I think it should be. I had trouble getting a "enable multiselect by default" option working reliably, so the multiselect option is persistent instead.

yaichenbaum: That's fine, my question was regarding if this is the correct place to declare the variable, shouldn't this be in a different view model?

winston-de: @yair100 I thought it was fine, since it allows the setting to be persistent and is accessible to both the status center and the layouts, but I could be wrong. Where do you think it should go?

yaichenbaum: This setting should really be temporary since it's meant to toggle the selection option, I would recommend putting it in `InteractionViewModel`. Technically this should be called `MainViewModel` but that's beyond the scope of this PR.

Files/UserControls/StatusBarControl.xaml.cs
```diff
@@ -3,13 +3,15 @@
 using System.ComponentModel;
 using System.Runtime.CompilerServices;
 using System.Windows.Input;
+using Windows.UI.Xaml;
```
yaichenbaum, line 6: Is this extra?

yair100 (dismissed):

tsvietOK:
@winston-de What about to use SelectionMode `Multiple` and bind `MultiselectEnabled` to `IsMultiSelectCheckBoxEnabled`?

Files/BaseLayout.cs
```diff
@@ -177,6 +179,8 @@ public virtual void InvertSelection()
 
         public abstract void ScrollIntoView(ListedItem item);
 
+        //public abstract bool MultiselectEnabled { get; set; }
```
tsvietOK, line 182: Is it needed?

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -83,6 +83,7 @@ public DataGridColumn SortedColumn
             }
         }
 
+        //public override bool MultiselectEnabled { get; set; }
```
tsvietOK, line 86: Is it needed?

winston-de:
@tsvietOK I don't think that's a good idea because the multiple selection mode behaves differently than extended mode. For example, when using the multiple selection mode, clicking an item toggles that item's selection, while clicking an item in extended mode clears the rest of the selection. This would be confusing for users who are used to the current behavior.

Files/Views/LayoutModes/GridViewBrowser.xaml
```diff
@@ -958,8 +961,17 @@
...
             PreviewKeyDown="FileList_PreviewKeyDown"
             SelectionChanged="FileList_SelectionChanged"
-            SelectionMode="Extended"
+            SelectionMode="{x:Bind InteractionViewModel.MultiselectEnabled, Mode=OneWay, Converter={StaticResource BoolToSelectionModeConverter}}"
             Visibility="{x:Bind FolderSettings.IsLayoutModeChanging, Mode=OneWay, Converter={StaticResource NegatedBoolToVisibilityConverter}}">
+            <GridView.Resources>
+                <ResourceDictionary>
+                    <ResourceDictionary.ThemeDictionaries>
+                        <ResourceDictionary x:Key="Default">
+                            <SolidColorBrush x:Key="GridViewItemCheckBoxBrush"  Color="{StaticResource SystemChromeMediumLowColor}"/>
```
tsvietOK, line 970:
```suggestion
                            <SolidColorBrush x:Key="GridViewItemCheckBoxBrush"  Color="{ThemeResource SystemChromeMediumLowColor}"/>
```

tsvietOK: This will fix colors with different app and system themes enabled.

Files/Strings/en-US/Resources.resw
```diff
@@ -1812,4 +1812,10 @@
   <data name="CloudDriveSyncStatus_Unknown" xml:space="preserve">
     <value>Unknown</value>
   </data>
+  <data name="StatusBarControlMultiselect.Text" xml:space="preserve">
+    <value>Multiselect</value>
+  </data>
+  <data name="SettingsShowMultiselectOption.Header" xml:space="preserve">
+    <value>Enable multiselect options</value>
```
tsvietOK, line 1819: What about to specify, that for the moment Multiselect is working only in GridView mode?

yaichenbaum: @tsvietOK It's under the experimental features so they don't really need any extra information, but it can't hurt.

tsvietOK (approved):

PR #3366, opened 2021-01-31
Added BitLocker module to cpp_fulltrust

d2dyno1:
This is PR adds powershell BitLocker module to cpp_fulltrust.

### DIFF
--- a/Files/ViewModels/ItemViewModel.cs
+++ b/Files/ViewModels/ItemViewModel.cs
@@ -1010,9 +1010,9 @@ namespace Files.ViewModels
                     {
                         var value = new ValueSet();
                         value.Add("Arguments", "Bitlocker");
-                        value.Add("action", "Unlock");
-                        value.Add("drive", Path.GetPathRoot(WorkingDirectory));
-                        value.Add("password", userInput);
+                        value.Add("Action", "Unlock");
+                        value.Add("FilePath", Path.GetPathRoot(WorkingDirectory));
+                        value.Add("Password", userInput);
                         await Connection.SendMessageAsync(value);
 
                         if (await CheckBitlockerStatusAsync(_rootFolder))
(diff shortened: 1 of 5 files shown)


d2dyno1:
:eyes:

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1010,9 +1010,9 @@ public async Task<bool> EnumerateItemsFromStandardFolderAsync(string path)
                     {
                         var value = new ValueSet();
                         value.Add("Arguments", "Bitlocker");
-                        value.Add("action", "Unlock");
-                        value.Add("drive", Path.GetPathRoot(WorkingDirectory));
-                        value.Add("password", userInput);
+                        value.Add("Action", "Unlock");
```
gave92, line 1013: I think it's better to minimize changes outside the Cpp Fulltrustprocess unless there's a compelling reason, since it will be a long time before merging to master

d2dyno1: hmm... This is the only instance where Bitlocker call is used. Since I had a chance, I wanted to unify the names of parameters

gave92: I think every other call uses uppercase for Arguments and lowercase for everything else

d2dyno1:
@gave92 could you check this PR, please! 😃

gave92:
Thanks a lot for this @d2dyno1 👍. In cbd60ad4c0394202230a45aa653cc7c7812a1f51 I made couple of changes:
- Changed `<< *password` to `<< std::wstring(password)` because only the first character was being added to the stringstream
- Made so the FTP waits until the Powershell script completes before sending a response
- Registered the bitlocker module in `main.cpp`

d2dyno1:
@gave92 Thanks for correcting me! Learned some things too!

PR #3349, opened 2021-01-29
Preview pane improvements

winston-de:
Improvements:
- [x] remember pane sizing
- [x] load file details even when no preview is available
- [x] cap file size for TryLoadAsText as attempting to load large files would cause performance issues
- [x] refine list of file details to load to improve performance and reduce visual clutter
- [x] fix issue where details are loaded twice when selected file is quickly switched.
- [x] improve text preview word count
- [x] make rich text preview control transparent

PR #3381, opened 2021-01-31
Fixed issue where shortcut items would not be cached properly

winston-de:
This pr fixes an issue where Shortcut items would not be cached properly.

The issue was caused by the Json converter serializing and de-serializing the shortcut as a ```ListedItem```, not as a ```ShortcutItem```. When the cache was then read, Files would treat the shortcut like a normal file, instead of as a shortcut. 

This would cause the details page instead of the shortcut page to be shown in the properties window when a shortcut was selected, and also caused a minor issue with #3349. 

Attempting to navigate to the details page would then cause a crash, since StorageFile doesn't work with shortcuts.

This PR fixes that issue by removing the ```ShortcutItem``` child class, and moving all of it's properties into the ```ListedItem``` parent class. ```ListedItem.IsShortcutItem``` is now set to true when a shortcut is initialized. This way, all of the shortcut-specific properties can be serialized and de-serialized, therefore allowing the shortcut's information to be cached.

I have tested these changes with creating, deleting, opening, and changing the settings of shortcuts, with and without file caching enabled, and have not noticed any issues thus far.

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -845,7 +845,7 @@ namespace Files
 
             foreach (ListedItem item in ParentShellPageInstance.ContentPage.SelectedItems)
             {
-                if (item is ShortcutItem)
+                if (item.IsShortcutItem)
                 {
                     // Can't drag shortcut items
                     continue;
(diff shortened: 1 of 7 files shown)


yaichenbaum (approved):

gave92:
I'm pretty sure you can set the json serializer to keep the type when serializing and thus deserialize to the correct type with the TypeNameHandling option

winston-de:
@gave92 That seems to work, and is much simpler. Thanks. I'll make a new branch and PR. #3434

PR #3386, opened 2021-02-01
Reduce scrolling stutters

### LINKED ISSUE #3371: Scrolling stutters/jumps in v1.2
**Describe the bug**
Scrolling seems worse than in 1.1 -- even in a directory that only has a handful of folders and files. It kind of seems to get stuck at certain points when scrolling and then continues (or jumps). After a restart (well actually a crash where the window just disappeared which I will file in a separate bug), things are better, but still stuttery/jumpy. 

From watching how it stutters/jumps, it is somewhat related to showing the highlighted item under the mouse, or at least it seems worse when that happens. Although it still seems to happen without that happening, but its far less stuttery.

**To Reproduce**
1. Open a folder with like 20-50 folders or files
2. Scroll with the mouse wheel -- note, I am using a Logitech MX Master 3 mouse with a very fast wheel, so you might need to scroll quite quickly.

**Expected behavior**
Smooth scrolling

 - OS Version: [10.0.19042 Build 19042]
 - App version: [v1.2.0.0]

**Log file**
[debug.txt](https://github.com/files-community/Files/files/5900628/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5900629/debug_fulltrust.txt)

hez2010:
Make use of SyncRoot to reduce unnecessary semaphores in BulkConcurrentObservableCollection.

I found there's no need to implement lock manually, there's already built-in `SyncRoot` for this (inspired by `SynchronizedCollection<T>`).

Fixes #3371

Note: it's no longer safe if you try to enumerate it without neither `lock (list.SyncRoot)` nor `.ToList()`

### DIFF
--- a/Files/Helpers/BulkConcurrentObservableCollection.cs
+++ b/Files/Helpers/BulkConcurrentObservableCollection.cs
@@ -10,106 +10,19 @@ namespace Files.Helpers
 {
     public class BulkConcurrentObservableCollection<T> : INotifyCollectionChanged, INotifyPropertyChanged, ICollection<T>, IList<T>, ICollection, IList
     {
-        private volatile bool isBulkOperationStarted;
-        private volatile int readerCount;
-        private volatile bool needSnapshot;
-        private readonly SemaphoreSlim writerLock = new SemaphoreSlim(1, 1);
-        private readonly SemaphoreSlim mutex = new SemaphoreSlim(1, 1);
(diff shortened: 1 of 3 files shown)


yaichenbaum (approved):

yaichenbaum:
@hez2010 Thank you!

PR #3434, opened 2021-02-03
Fixed issue where shortcut items would not be cached properly

winston-de:
New PR for the bug in #3381 with @gave92's suggested alternative changes.

This pr fixes an issue where Shortcut items would not be cached properly.

The issue was caused by the Json converter serializing and de-serializing the shortcut as a ```ListedItem```, not as a ```ShortcutItem```. When the cache was then read, Files would treat the shortcut like a normal file, instead of as a shortcut. 

This would cause the details page instead of the shortcut page to be shown in the properties window when a shortcut was selected, and also caused a minor issue with #3349. 

Attempting to navigate to the details page would then cause a crash, since StorageFile doesn't work with shortcuts.

This PR fixes that by adding the preserve type setting to the Json serializer.


winston-de:
Also, fyi @jakoss, I want to make sure this doesn't interfere with #2944.

jakoss:
> Also, fyi @jakoss, does this interfere with #2944?

I think it should be merged automatically. I modified other methods in this file

winston-de:
@jakoss Got it, I just wanted to make sure there aren't any major conflicts.

gave92:
@winston-de I haven't tested this but I thought TypeNameHandlig had be specified also when Deserializing, could you check?

winston-de:
> @winston-de I haven't tested this but I thought TypeNameHandlig had to be specified also when Deserializing, could you check?

@gave92 I tested it, and it deserializes to the correct type without issue, so I don't think that's the case.

gave92 (approved):

PR #3461, opened 2021-02-06
Fixes an issue that prevented some folders from opening

### LINKED ISSUE #3460: Can't Open Invisible Folder
Hello Everyone,
I maked an invisble folder without name just space with named alt+0160 on numpad technique to easily access my apps from toolbar .But when I open that folder with double click or context menu's open, it not opens. It successfully open in new pane, tab, window
I think its a bug so I reported it.

gave92:
Fixes #3460


tsvietOK (approved):

yaichenbaum:
@gave92 Thank you!

PR #3510, opened 2021-02-09
Update french translation

LoganTann:
Just a translation of some strings that have got the `state="new"` label

A lot needs a review, because it depends on the context of the sentence / what the feature does, and some strings don't contain the exact words of the original strings to improve the readability.


tsvietOK:
I think we need someone who knows French.

yaichenbaum:
@LoganTann The French translation has been updated in another pull request.

PR #3527, opened 2021-02-10
📓 Updated HUN

KristofKekesi:
For reference here is the win 10 popup with the hungarian translation.

image


d2dyno1:
HUN update is so fast it's already translated before a feature is merged 😆

yaichenbaum (approved):

PR #3580, opened 2021-02-15
Fix ObjectDisposedException in CancellationTokenSource

hez2010:
After some investigation, there's no need to dispose CancellationTokenSource manually if `LinkedTokenSource` was not used, so just let GC dispose CancellationTokenSource to prevent ObjectDisposedException.

### DIFF
--- a/Files/Files.csproj
+++ b/Files/Files.csproj
@@ -170,7 +170,6 @@
     <Compile Include="Converters\UInt32ToString.cs" />
     <Compile Include="Converters\VisiblityInvertConverter.cs" />
     <Compile Include="Converters\LayoutModeToBoolConverter.cs" />
-    <Compile Include="Converters\LayoutModeToBoolConverter.cs" />
     <Compile Include="DataModels\ShellNewEntry.cs" />
     <Compile Include="Dialogs\DynamicDialog.xaml.cs">
       <DependentUpon>DynamicDialog.xaml</DependentUpon>
--- a/Files/ViewModels/ItemViewModel.cs
+++ b/Files/ViewModels/ItemViewModel.cs
@@ -752,7 +752,7 @@ namespace Files.ViewModels
                 {
                     await loadExtendedPropsSemaphore.WaitAsync(loadPropsCTS.Token);
                 }
-                catch (Exception ex) when (ex is OperationCanceledException || ex is ObjectDisposedException)
+                catch (OperationCanceledException)
                 {
                     return;
                 }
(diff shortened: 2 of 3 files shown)


gave92:
@hez2010 this is cool. Could you post here the source/documentation that suggests this?

hez2010:
I followed https://stackoverflow.com/questions/6960520/when-to-dispose-cancellationtokensource
Since we don't use "linked cancellationtoken", I think there's no need to dispose it manually.

gave92:
Reading the linked post I understand that you can skip disposing the CancellationToken if:
- you're not calling CancelAfter() -> we're not
- you're not using LinkedTokenSource -> we're not
- you're not using WaitHandle -> ???

> The m_kernelHandle internal field mentioned above is the synchronization object backing the WaitHandle property in both the CTS and CT classes. It is only instantiated if you access that property. So, unless you are using WaitHandle for some old-school thread synchronization in your Task calling dispose will have no effect.

Problem is:
> It may not be you (i.e. the one making the async request) who uses the wait handle, it may be the listener (i.e. the one answering the request). Which means you, as the one responsible for disposing, effectively have no control over whether the wait handle is used or not.

hez2010:
How about implementing a wrapper for `CancellationTokenSource` to dispose it on finalizer?

```csharp
class AutoFinalizedCancellationTokenSource : CancellationTokenSource
{
    ~AutoFinalizedCancellationTokenSource()
    {
        base.Dispose();
    }
}
```

gave92:
Hi @hez2010 I've checked with the following code if the "_kernelEvent" of the CancellationTokenSource is actually being initialized or not:
`Debug.WriteLine("NOT NULL WAITHANDLE: {0}", typeof(CancellationTokenSource).GetField("_kernelEvent", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance).GetValue(loadPropsCTS) != null);`
I've added this after the call to `Cancel` but before the call to `Dispose` and in all cases I could see it says that the `_kernelEvent` has not been initialized.

So I think we can go with your solution "1" and simply remove the calls to Dispose where we're just going to create the CTS again (and remove the ObsectDisposedException handling to).

hez2010:
Done in befa355

gave92:
Thanks @hez2010 looks good.

In case we find issues with this an alternative approach would have been caching the Token and using that instead of the CTS:
https://www.rocksolidknowledge.com/articles/working-with-cancellationtoken-and-dispose/

gave92 (approved):

PR #3638, opened 2021-02-19
Made PathBox ignore double backslashes

d2dyno1:


gave92:
What about network shares ("\\server\folder")?

d2dyno1:
It always frustrated me when I copied a string path with double `\\` this PR is a quality of life change.

> And what about network shares ("\\server\folder")?

That's a good point! I'll add a check for that.

d2dyno1:
@gave92 should be fine now

gave92 (approved):

Issue #3502, opened 2021-02-09
Automatically preview readme files when no file is selected


andrew-stclair:
**Is your feature request related to a problem? Please describe.**
Not a problem, just an idea i had for implementing something from the XP days.

**Describe the solution you'd like**
When in a folder with no file or files selected, the preview pane should preview a file (if it exists) with a specific name, like README.md

**Describe alternatives you've considered**
Only alternative i have come up with is to just click on README.md with the preview pane open

**Additional context**
(This but without me having to select the file)
![image](https://user-images.githubusercontent.com/4944499/107349032-d8c4d780-6b1b-11eb-861e-4604db7926af.png)

yaichenbaum:
@andrew-stclair This is an interesting idea, and I can see it being useful for a lot of users.

R3voA3:
Love the idea. Not sure if useful, but maybe a setting where the user can define the name of the file?

d2dyno1:
How about it previewed the current directory you're in?

winston-de:
What if, instead, the file was automatically selected when the user navigated to a folder that contains a readme file? This way, if the file is quickly accessible to open (or right click) and would show the preview. The user would also be able to set names/extensions (eg: .sln file, .csproj) that they would want to quickly be able to access/preview when opening a folder.

d2dyno1:
e

yaichenbaum:
This issue should be resolved with #5393.

Issue #3400, opened 2021-02-01
Preview pane for folders


KristofKekesi:
**Describe the solution you'd like**
Some additional informations about the folder (like how much sotrage does it take, how much files does it contain, or how much files with x file extension).

yaichenbaum:
@KristofKekesi This enhancement was implemented with #3639.

winston-de:
Unfortunately, Folder previews were not finished in #3639, and were disabled.

I made a new PR where I will continue work on implementing this, #3773.

Also, here's what folder previews look like so far. 
![image](https://user-images.githubusercontent.com/59544401/109431243-5ad94b00-79ba-11eb-9627-e6a1a804d240.png)

PR #3639, opened 2021-02-19
Preview pane improvements 2

### LINKED ISSUE #3446: Preview pane should say no item selected when there's no selected item
Today preview pane says no details available / no preview available but when there's no selected item it should say "No item selected. Please select one to view it's details and preview"

winston-de:
### to-do
- [ ] Folder previews #3400 
- [x] Shortcut previews
- [x] Move preview pane code-behind to a view model
- [x] Show thumbnails when no preview control/extension exists
- [x] Fix issue where preview pane doesn't show in search results
- [x] Allow attempting to load as text file even if file has no extension
- [ ] Reduce text cutoff
- [ ] Automatically preview readmes when no file is selected (#3502)
- [ ] Add option to open preview pane by default
- [ ] Cache file property list from json file to reduce inefficiency 
- [ ] Improve code quality for adaptive pane
- [ ] Enable opening links from markdown and web preview
- [x] Try to fix and re-implement code previews, if not possible remove altogether
- [x] Attempt to add shadow for pane
- [x] Don't try and load hidden items
- [x] Show different message when no item is selected #3446 

Closes #3446

Thumbnail preview:


d2dyno1:
Item tiles with names for folder preview would be perfect!

winston-de:
I was able to get the shadow effect working, which was requested in the original PR.
Should I keep it?

image

yaichenbaum:
@winston-de Can you try casting the shadow on the preview pane?

winston-de:
@yair100 From the content pane, right?

yair100:
> @yair100 From the content pane, right?

Right

winston-de:
@yair100 
image

winston-de:
Also, I'm trying to load the thumbnail for shortcut previews, but currently the function for sending the request to the full trust process is located in ItemViewModel. Would it be alright if I moved it to it's own helper class?

yaichenbaum:
@winston-de I know it's not directly connected to the preview pane, but what if shadows were added to the navigation and status bar?

winston-de:
Yes, though right now it doesn't work well at all if the preview pane also has a shadow because the elevations for the preview pane and content would be inconsistent. I'll work on using a drop shadow instead.

For now, here's what the shadows for the nav toolbar and status bar would look.
image

yaichenbaum:
@winston-de Once the paddings/margins and margins are sorted out, I think would look pretty good.

winston-de:
@yair100 I remember you saying something about changing the properties icon so it wasn't the same as the icon for the preview pane. I found one that I think would work, but it doesn't seem to be in the current sheet.

https://raw.githubusercontent.com/microsoft/fluentui-system-icons/master/assets/Wrench/SVG/ic_fluent_wrench_24_regular.svg

yair100:
> @yair100 I remember you saying something about changing the properties icon so it wasn't the same as the icon for the preview pane. I found one that I think would work, but it doesn't seem to be in the current sheet.
> 
> https://raw.githubusercontent.com/microsoft/fluentui-system-icons/master/assets/Wrench/SVG/ic_fluent_wrench_24_regular.svg

That looks good.

winston-de:
I think I'll have to update the fluent ui glyphs font currently in Files to use it though.

yaichenbaum:
@winston-de That's probably out of scope for this PR, but it would be great if you could make that change.

winston-de:
Alright, I redid the shadows with ```DropShadowPanel```s instead, because they work much better and require a lot less code. However, I'm not certain if the opacity, blur radius, and offset values are the best.

@yair100 Do you have any input?

image

yaichenbaum:
@winston-de Can you try copying the values from the sidebar? By the way, you might want to merge the latest from `main` to see how the shadows will work with the recent changes.

winston-de:
Here it is with the values from the sidebar shadow.

image

yaichenbaum:
@winston-de That looks good, I am looking into using different colors there as well so that should help the shadow stand out a little more.

yair100 (approved):
Great work! LGTM

Issue #3184, opened 2021-01-20
Extend the left sidebar to the title bar (tab page bar) area, and limit the tab page bar to the right area of ​​the window (as shown in the figure)


Shomnipotence:
**Is your feature request related to a problem? Please describe.**
Please build a more holistic interactive interface. In this interface, there are two levels of navigation, one is the left sidebar navigation, and the other is the tab navigation. How to deal with the relationship between the two is a very troublesome thing.

Theoretically speaking, the tab page should be the first-level navigation, and the sidebar is the second-level, indicating the folder where the tab page is located. Therefore, the tab bar runs through the top of the entire window, which makes logical sense.

![image](https://user-images.githubusercontent.com/6630660/105139410-3d020600-5b31-11eb-8681-5ef6100a0c75.png)
OS X 10.0+-macOS 10.14, Finder Screenshot

**Describe the solution you'd like**
But this is not the case. With the continuous development of the times, the design has been evolving.

First of all, Windows 10 uses the left and right structure to organize the interface for the first time. The left side is a sidebar with acrylic effect, and the right side is a black background.

![image](https://user-images.githubusercontent.com/6630660/105139512-5efb8880-5b31-11eb-8a92-e7437b305dad.png)
Windows 10 Setting

Next, macOS 11 Big Sur also adopted a similar structure

![image](https://user-images.githubusercontent.com/6630660/105139628-789cd000-5b31-11eb-9848-f509cd23d268.png)
macOS 11 Finder

So how to deal with the tab bar?

The answer given by macOS is
![image](https://user-images.githubusercontent.com/6630660/105139733-9bc77f80-5b31-11eb-9a12-e7894b8c02c6.png)
![image](https://user-images.githubusercontent.com/6630660/105139680-83576500-5b31-11eb-8a99-4a7351265429.png)

I think this solution is excellent, it brings a more overall and more beautiful design, it maintains the left and right structure of "transparent on the left and opaque on the right".

I think Files should also adopt this structure, or at least as an option.

![未标题-222](https://user-images.githubusercontent.com/6630660/105139801-bdc10200-5b31-11eb-997a-f7703b437c58.png)
Files conceptual design

**Describe alternatives you've considered**
There is no other solution, Files is the only file explorer with Fluent Design available on Windows

**Additional context**
Add any other context or screenshots about the feature request here. Ex: What device type/UI layout and app version?

gave92:
[...]

PR #3705, opened 2021-02-24
NavBar improvements step 1

yair100:
- Moved the navigation buttons to the left of the path bar
- Fixed spacing of the toolbar buttons

### DIFF
--- a/Files/UserControls/NavigationToolbar.xaml
+++ b/Files/UserControls/NavigationToolbar.xaml
@@ -15,7 +15,7 @@
     xmlns:local="using:Files.UserControls.MultitaskingControl"
     xmlns:local1="using:Files"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-    d:DesignHeight="80"
+    d:DesignHeight="72"
     d:DesignWidth="800"
     mc:Ignorable="d">
     <UserControl.Resources>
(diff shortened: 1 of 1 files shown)


d2dyno1:
Could you share a screenshot?

yaichenbaum:
@d2dyno1 This PR is part of a larger effort to redesign the navbar area, but this is what it looks like so far.
image

gave92:
LGTM (actually it looks horrible but it's part of the plan right? 🤣)

gave92 (approved):

d2dyno1:
> LGTM (actually it looks horrible but it's part of the plan right? )

tbh I like this one more

yaichenbaum:
> LGTM (actually it looks horrible but it's part of the plan right? 🤣)

Exactly 😆

d2dyno1:
> > LGTM (actually it looks horrible but it's part of the plan right? rofl)
> 
> Exactly

Hmm. How will the final version of this redesign look alike?

yaichenbaum:
@d2dyno1 There are several parts to this design change including #3184, there are also plans to add some shadows around the app. It's hard to describe how it will look, but this concept should give you the basic idea.
image

gave92:
Ooof every time I see one of this mockups I cry for the lost sidebar extending all the way to the top 😂. Aware of the reason behind the change my gut says otherwise.

yaichenbaum:
@gave92 What don't you like about the sidebar extending to top?

gave92:
I meant I don't like too much the current design where the sidebar is *not* extented to the top. When the sidebar is made acrylic I miss the "old" design.

yaichenbaum:
@gave92 Oh good, then I am sure you will like the new design when #3184 is merged 😄

R3voA3:
I see that little progress circle in the bottom right. Will this replace this progress bar which indicates that the folder is being loaded? I'd appreciate that.

yaichenbaum:
@R3voA3 The progress circle is just a concept, don't read much into it.

R3voA3:
> @R3voA3 The progress ring is just a concept, there isn't anything planned to change there.

Too bad =(

SOI7:
> @d2dyno1 There are several parts to this design change including #3184, there are also plans to add some shadows around the app. It's hard to describe how it will look, but this concept should give you the basic idea.
> image

That has... everything I wanted 😲Even the W10X quick actions below the navbar

gave92:
@yair100 one thing I noticed is that with the navigation buttons on the same row as the address bar on my screen the address bar becomes very cramped (in dual pane even more so and almost unusable). Just saying that this change should probably adapt with the window width.

yaichenbaum:
@gave92 It's something to look into, I am curious if this is really an issue or not because the dual screen feature is mainly used on larger screens. Regardless, I plan on reducing the size of the text in the nav bar, so that should help with this issue.

gave92:
Thanks, it's simply that currently I can use the dual pane feature with no issue, but with this change it gets a little awkward. Thinking that the search button and the others are going to also be moved into the same row.. :)

yaichenbaum:
@gave92 The only other button that's going to be moved is the search button, it's something we should start discussing to find possible solutions. Thank you for bringing it up.

h38d7v:
Wondering if you have considered adding the title of the app (and logo) to the top left of the NavBar. Probably good from a branding point of view, but also would be consistent with a lot of the other 'modern' apps. At the moment it sort of looks like the app is called 'favourites'. :) Screenshot below for example.

image

R3voA3:
Not really sure where the title and logo would fit since the Tabs are taking the space and soon the nav bar.
Also, they should not try to be consistent with Microsoft apps since that's impossible :D

SOI7:
The Win32 Explorer doesn't use a logo or a title anyway, so as it is now Files is consistent with the legacy file browser

R3voA3:
@yair100 Any chance to see a screenshot? I need to see where

`          <target state="new">Browse</target>
        </trans-unit>
        <trans-unit id="RecentLocations" translate="yes" xml:space="preserve">
          <source>Recent locations</source>
          <target state="new">Recent locations</target>`

Is used so I can properly translate it.

h38d7v:
Here is two good examples I think. Adding a title to the top of the navbar doesn't get in the way of tabs, or the navbar items.

image

Also here: https://www.behance.net/gallery/74300939/Windows-10-File-Explorer

tsvietOK:
@R3voA3 
image

yaichenbaum:
> Wondering if you have considered adding the title of the app (and logo) to the top left of the NavBar. Probably good from a branding point of view, but also would be consistent with a lot of the other 'modern' apps. At the moment it sort of looks like the app is called 'favourites'. :) Screenshot below for example.
> 
> image

Thanks for bringing this up, it seems like Microsoft is moving back towards this trend, but as mentioned above, it wouldn't fit into our design right now.

PR #3743, opened 2021-02-26
NavBar updates step 2

yaichenbaum:
This pull request continues to improve the nav bar design by updating the corner radius to match the new guidelines.

### DIFF
--- a/Files/UserControls/NavigationToolbar.xaml
+++ b/Files/UserControls/NavigationToolbar.xaml
@@ -22,18 +22,6 @@
         <ResourceDictionary>
             <converters1:BoolNegationConverter x:Key="BoolNegationConverter" />
 
-            <ResourceDictionary.ThemeDictionaries>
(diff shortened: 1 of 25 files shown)


gave92:
@yaichenbaum irrelevant to this PR: one thing I noticed is that with the navigation buttons on the same row as the address bar on my screen the address bar becomes very cramped(in dual pane even more so and almost unusable). Just saying that this change would adapt with the window width.

Files/UserControls/NavigationToolbar.xaml
```diff
@@ -33,9 +33,24 @@
                     <SolidColorBrush x:Key="PathBoxHoverBrush" Color="{ThemeResource SystemColorWindowColor}" />
```
gave92, line 33: Remove this unused resource?

gave92 (approved):

PR #3765, opened 2021-02-28
Focus text input when dialog opened.

PMExtra:
### Steps:
1. Open a folder in Files.
2. Right click and select `New -> anything` in context menu.
3. The dialog that request the item name is pop up.

### Actual behavior (before this patch):
We have to click the text box or press `Tab` key to move the focus.

### Excepted behavior (after this patch):
The text input is auto focused.

### DIFF
--- a/Files/Helpers/DynamicDialogFactory.cs
+++ b/Files/Helpers/DynamicDialogFactory.cs
@@ -5,6 +5,7 @@ using Files.ViewModels.Dialogs;
 using Microsoft.Toolkit.Uwp.Extensions;
 using System;
 using Windows.System;
+using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 
 namespace Files.Helpers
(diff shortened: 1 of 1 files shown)


Files/Helpers/DynamicDialogFactory.cs
```diff
@@ -109,6 +110,11 @@ public static DynamicDialog GetFor_RenameDialog()
                 DynamicButtons = DynamicDialogButtons.Primary | DynamicDialogButtons.Cancel
             });
 
+            dialog.Opened += (sender, args) =>
+            {
+                inputText.Focus(FocusState.Keyboard);
```
yaichenbaum, line 115:
```suggestion
                inputText.Focus(FocusState.Programmatic);
```

yaichenbaum (changes requested):

yaichenbaum (approved):

PR #3773, opened 2021-02-28
Added folder previews to the preview pane

winston-de:
image


d2dyno1:
@winston-de 
https://github.com/files-community/Files/pull/3639#issuecomment-782439024

+ Could you also add an option in settings to disable folder previews

winston-de:
@d2dyno1 Yeah, I could add an option to disable them. 

> Item tiles with names for folder preview would be perfect!

I'm not sure about item names, as my end goal is to achieve something like this:
image
The current GridView display is more of a rough prototype/proof of concept.

However, this is by no means set in stone, please let me know if you have any feedback and/or suggestions :)

d2dyno1:
I kinda want filenames to be displayed, but what if we were to put them in tooltip? @winston-de

yaichenbaum:
> Could you also add an option in settings to disable folder previews :)

At the end of the day, I don't an option is necessary here. As long as the preview pane is open, it doesn't hurt to put information there. If the user doesn't want to see a preview, they can already hide the preview pane.

winston-de:
@d2dyno1 
Personally, I would rather have it show just thumbnails, because I generally find thumbnails to be a better representation of what's in the file than it's name.
However, I could easily do both.
Perhaps instead of simply a toggle setting for folder previews, there was a drop-down with:
- Tiles view
- thumbnail view
- none

Would that be a good solution?

Also, did you want to be able to disable folder details, too, or just the preview?

---edit---
@yair100 I just saw your comment now. 
What are your thoughts on having an option to change the style of folder previews?

winston-de:
I don't think having a tile view will work well, as it's slow to load.
As it turns out, StorageFolder has built in thumbnails, so I will just use those for the preview for the time being.

image

Files/ViewModels/PreviewPaneViewModel.cs
```diff
@@ -98,11 +98,7 @@ private async Task LoadPreviewControlAsync()
 
             if (SelectedItem.PrimaryItemAttribute == StorageItemTypes.Folder)
             {
-                // TODO: Finish folder previews and reimplement later
-                //PreviewPaneContent = new FolderPreview(new FolderPreviewViewModel(SelectedItem));
-                DetailsErrorText = "PreviewPaneDetailsNotAvailableText".GetLocalized();
-                PreviewErrorText = "DetailsPanePreviewNotAvaliableText".GetLocalized();
```
yaichenbaum, line 104: Can these strings be removed?

winston-de: @yair100 No, they're used elsewhere

Those are the "No preview available" and "No details available" strings.

yair100 (approved):

PR #3778, opened 2021-03-01
UI Tweaks

yair100:
- Fixed padding on home page
- Hid preview pane button on home page
- Fixed grid splitter style by the preview pane
- Tweaked shadow style on tabs


tsvietOK (approved):

winston-de:
@yair100 I think the shadows for the preview pane got a bit messed up with this PR since the offset was set to account for the grid splitter.

image

It's a quick fix, I can add it to my other PR if you'd like.

yaichenbaum:
@winston-de That would be great, thank you!

PR #3779, opened 2021-03-01
Changed selected tab color to make it easier to see

winston-de:
Before it was the same as the tab strip background, which made it harder to see.

image


yaichenbaum:
@winston-de What does it look like in light theme?

winston-de:
> @winston-de What does it look like in light theme?

@yair100 

image

winston-de:
Also, @yair100, currently the pointer over background for unselected tabs is currently the same as the tab strip background. Maybe it should be changed to something like this?

image

image

yaichenbaum:
@winston-de That might be a little better than what I did in my PR, lets wait until that's merged before finishing this one.

winston-de:
@yair100 Oh, great! I didn't realize your pr already added a pointer over effect to tabs.

yair100 (approved):

PR #3793, opened 2021-03-03
Updated brushes in the settings pages

yaichenbaum:


duke7553 (approved):
These are from a set of changes we discussed and will share with the community once we're sure of the direction were headed towards

PR #3854, opened 2021-03-05
Change windowing/tab behavior when launching Files though different methods

### LINKED ISSUE #3852: [Feature request] Change launch handling to allow opening in new tabs instead of new instances for different launch types

Currently, when Files is launched using command line arguments, it opens in a new instance instead of a new tab. When Files is set as the default explorer, folders opened through Search, PowerToys run, and VSCode all open in new windows instead of tabs.

Also, when Files is launched with no arguments from the start menu, it opens a new instance. This is annoying as opening a new instance is slower, and pressing other apps on the start menu such as edge open the current instance, not a new one.

- Adjust the launch event handing in Program.cs to open in a new tab when launched with CommandLineArgs. 

- Remove requirement for arguments to open in a new tab instead of instance so that launching from start opens the current instance

- ~~I also think the "when launching from the jump list -> open new tab / new instance" option should be changed to a universal setting for opening new windows or tabs when launched.~~ It appears this is already the case, it's just that it's labeled oddly. I think the label should be changed to say "When opening a folder in Files".

**Additional context**
I have already done the work for it and will open a PR soon. :)

winston-de:
Closes #3852.

- When Files is launched from the command line, the folder will open in a new tab if Files is already open, instead of a new window
- Changed "when opening from jumplist" text to "When opening folder in Files" 
- This pr also fixes a bug I noticed where "open in new instance" setting was ignored because it was being read from the roaming settings in program.cs, while the setting was saved to local settings.

- When Files is launched from the start menu with no event args and "open in new instance" is disabled, it will open the current instance (like Edge does)
NOTE: this will prevent shift+click from opening a new window in the taskbar, so if thats an issue I can undo those changes.


tsvietOK:
Hmmm, i guess last change should be reverted, because i pretty often use middle mouse button click to open new instance of any program

winston-de:
@tsvietOK Ok, changes made.

tsvietOK (approved):

PR #3863, opened 2021-03-07
Fixed drop shadow on sidebar

### LINKED ISSUE #3850: Make drop shadow between the sidebar and the main content area more prominent
In light theme, the sidebar often blends with the main content area because their colors end up being nearly the same.

image

Add a prominent drop shadow between the sidebar and the main content area.

image

yaichenbaum:
Closes #3850

### DIFF
--- a/Files/Views/ModernShellPage.xaml
+++ b/Files/Views/ModernShellPage.xaml
@@ -195,9 +195,9 @@
             Grid.Row="2"
             HorizontalAlignment="Stretch"
             HorizontalContentAlignment="Stretch"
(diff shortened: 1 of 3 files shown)


gave92:
When the second pane is open the two status bars are not aligned.
I think it's better to always show the 2px border and set its color to the status bar color when the pane is not active
image

yaichenbaum:
@gave92 We can try showing it above the status bar, however, let's do that in another PR.

gave92 (approved):

PR #3861, opened 2021-03-07
Remove ContentFrame member from IShellPage

duke7553:
Cleanups to IShellPage, and remove IShellPage constructor arg from FolderSettingsViewModel (@d2dyno1 for reviewing this)

PR #3902, opened 2021-03-09
Improvements to codebase

d2dyno1:
Just a couple of suggestions of mine \:)

future proofing


gave92:
This is the change that was discussed in #3861, right? For the moment I don't really see the point of this since we have the FolderSettingsViewModel as the sender arg. We'd only need to do this if we ever move that event out of FolderSettings.

duke7553 (commented):
@d2dyno1 Could you give a reason for this change? What is your rationale because I thought we agreed on the current method with the last PR?

d2dyno1:
@duke7553 I've made this PR if we ever wanted in the future to move the `FolderSettings_LayoutPreferencesUpdateRequired` outside of `ModernShellPage`

gave92:
Thanks @d2dyno1 I'm closing this for now since with the current code there is no particular reason to change. If we end up changing things we'll totally consider this.

PR #3915, opened 2021-03-09
Update Files.de-DE.xlf

R3voA3:
To finalize the German translation for the upcoming version I need to know where the following translations are used

```xml
<trans-unit id="PropertyProviderStyle" translate="yes" xml:space="preserve">
 <source>Provider Style</source>
 <target state="new">Provider Style</target>
</trans-unit>
```
I couldn't find it in the details selection UI.
image

---
Also this translation; What is it used for?
```xml
<trans-unit id="PropertyItemArguments" translate="yes" xml:space="preserve">
 <source>Arguments</source>
 <target state="translated">Argumente</target> <!---might not be correct in this context-->
</trans-unit>
```


yaichenbaum:
@winston-de Do you have any idea where these are being used?

winston-de:
@yair100 Yes, it's one of the details used for shortcut previews.

R3voA3:
I will leave it as it is for now I will translate them once I stumble over them in the released version.

yair100 (approved):

PR #3946, opened 2021-03-10
Add section property to LocationItem

oleitao:
Right clicking on a library item shows the option to unpin them and that's not the correct behaviour. With that PR i want to correct that issue.


yaichenbaum:
@gave92 Do you think this scales better than the solution we currently use for this?

gave92:
I like the idea of adding a `Section` property to sidebar items so you don't have to write ugly code such as:
```cs
var isLibrary = (MainPage.SideBarItems.FirstOrDefault(x => x.Text == "SidebarLibraries".GetLocalized()) as LocationItem)?.ChildItems?.Contains(item) ?? false;
```
Adding the possibility to unpin single libraries from the section might be less important,

oleitao:
> I like the idea of adding a `Section` property to sidebar items so you don't have to write ugly code such as:
> 
> ```cs
> var isLibrary = (MainPage.SideBarItems.FirstOrDefault(x => x.Text == "SidebarLibraries".GetLocalized()) as LocationItem)?.ChildItems?.Contains(item) ?? false;
> ```
> 
> Adding the possibility to unpin single libraries from the section might be less important (if we do we could just have left all items under quick access).

Can i continue that development?

gave92:
> Can i continue that development?

@yair100 what do you think about allowing users to unpin single library folders? Personal opinion: not needed.

yaichenbaum:
@gave92 Perhaps we can add an option to delete libraries when we add proper detection for libraries created in File Explorer.

gave92:
@oleitao For now we'd limit thie changes to the following:
- Adding the `Section` property to LocationItem
- Use the new property to simplify [this code](https://github.com/files-community/Files/blob/0f1800e007694373712517adf24506a4e6485f5c/Files/UserControls/SidebarControl.xaml.cs#L241)

oleitao:
> @oleitao For now we'd limit the changes to the following:
> 
> * Adding the `Section` property to LocationItem (~only to LocationItem, INavigationControlItem does not probably need this~edit: current solution is good)
> * Using the new property to simplify [this code](https://github.com/files-community/Files/blob/0f1800e007694373712517adf24506a4e6485f5c/Files/UserControls/SidebarControl.xaml.cs#L241)

@gave92 please check my last commit and tell me if it's everything fine to continue this development?

Files/App.xaml.cs
```diff
@@ -140,7 +140,14 @@ public static void UnpinItem_Click(object sender, RoutedEventArgs e)
...
             else
             {
-                SidebarPinnedController.Model.RemoveItem(RightClickedItem.Path.ToString());
+                if (RightClickedItem.Section == SectionType.Favorites)
+                {
+                    SidebarPinnedController.Model.RemoveItem(RightClickedItem.Path.ToString());
+                }
+                else if (RightClickedItem.Section == SectionType.Library)
+                {
+                    SidebarPinnedController.Model.RemoveLibraryItem(RightClickedItem.Path.ToString());
```
gave92, line 149: For now we have decided to not allow to unpin single libraries.

Files/DataModels/SidebarPinnedModel.cs
```diff
@@ -141,6 +143,18 @@ public void RemoveItem(string item)
             }
         }
 
+        /// <summary>
+        /// Removes the library item.
+        /// </summary>
+        /// <param name="item">The item.</param>
+        public void RemoveLibraryItem(string path)
```
gave92, line 150: For now we have decided to not allow to unpin single libraries.

gave92 (changes requested):

gave92:
@oleitao I've pushed here a commit to update the PR to latest main branch. This looks ready to be merged, or is there still something you're planning to add?

gave92 (approved):

oleitao:
> @oleitao I've pushed here a commit to update the PR to latest main branch. This looks ready to be merged, or is there still something you're planning to add?

To me it's fine, you can merge.

yair100 (approved):

PR #3956, opened 2021-03-11
Added try/catch to prevent crash with reload item icons

yaichenbaum:
Fixes [2656712569u](https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/2656712569u)

### DIFF
--- a/Files/Views/LayoutModes/GridViewBrowser.xaml.cs
+++ b/Files/Views/LayoutModes/GridViewBrowser.xaml.cs
@@ -412,15 +412,23 @@ namespace Files.Views.LayoutModes
         private async void ReloadItemIcons()
         {
             ParentShellPageInstance.FilesystemViewModel.CancelExtendedPropertiesLoading();
-            foreach (ListedItem listedItem in ParentShellPageInstance.FilesystemViewModel.FilesAndFolders.ToList())
+
+            try
             {
-                listedItem.ItemPropertiesInitialized = false;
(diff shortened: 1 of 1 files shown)


Files/Views/LayoutModes/GridViewBrowser.xaml.cs
```diff
@@ -412,15 +412,23 @@ private void FolderSettings_GridViewSizeChangeRequested(object sender, EventArgs
         private async void ReloadItemIcons()
         {
             ParentShellPageInstance.FilesystemViewModel.CancelExtendedPropertiesLoading();
-            foreach (ListedItem listedItem in ParentShellPageInstance.FilesystemViewModel.FilesAndFolders.ToList())
+
+            try
             {
-                listedItem.ItemPropertiesInitialized = false;
-                if (FileList.ContainerFromItem(listedItem) != null)
+                foreach (ListedItem listedItem in ParentShellPageInstance.FilesystemViewModel.FilesAndFolders.ToList())
```
gave92, line 418: I added `ToList()` here in order to solve the same issue. Have you experienced this bug despite it?

yaichenbaum: @gave92 I haven't, the bug report was from v1.3 so it could be your fix was enough.

gave92: Ok I don't have the confirmation my fix works either. But perhaps we can try with it first.

PR #3967, opened 2021-03-11
Cleanups to Interaction

d2dyno1:
This PR aims to reduce `IShellPage` references in `Interaction` and also moves some stuff around

Motivation:
Since `IShellPage` contains a reference to `Interaction` and `Interaction` contains `IShellPage` field it creates circular references which causes memory leaks. To solve such problem we must get rid of that reference in `Interaction` completely. Ideally, we'd have `Interaction` class completely removed and all handlers moved to respective View Models.
In my next PR I'll try to move as many functions out of `Interaction` as possible.

### DIFF
--- a/Files/Helpers/XamlHelpers/DependencyObjectHelpers.cs
+++ b/Files/Helpers/XamlHelpers/DependencyObjectHelpers.cs
@@ -1,0 +1,69 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Media;
+
+namespace Files.Helpers.XamlHelpers
+{
+    public static class DependencyObjectHelpers
+    {
+        public static T FindChild<T>(DependencyObject startNode) where T : DependencyObject
+        {
+            int count = VisualTreeHelper.GetChildrenCount(startNode);
+            for (int i = 0; i < count; i++)
+            {
+                DependencyObject current = VisualTreeHelper.GetChild(startNode, i);
+                if (current.GetType().Equals(typeof(T)) || current.GetType().GetTypeInfo().IsSubclassOf(typeof(T)))
+                {
+                    T asType = (T)current;
+                    return asType;
+                }
+                var retVal = FindChild<T>(current);
+                if (retVal != null)
+                {
+                    return retVal;
+                }
+            }
+            return null;
+        }
+
+        public static void FindChildren<T>(IList<T> results, DependencyObject startNode) where T : DependencyObject
+        {
+            int count = VisualTreeHelper.GetChildrenCount(startNode);
+            for (int i = 0; i < count; i++)
+            {
+                DependencyObject current = VisualTreeHelper.GetChild(startNode, i);
+                if (current.GetType().Equals(typeof(T)) || (current.GetType().GetTypeInfo().IsSubclassOf(typeof(T))))
+                {
+                    T asType = (T)current;
+                    results.Add(asType);
+                }
+                FindChildren<T>(results, current);
+            }
+        }
+
+        public static T FindParent<T>(DependencyObject child) where T : DependencyObject
+        {
+            T parent = null;
+            if (child == null)
+            {
+                return parent;
+            }
+            DependencyObject CurrentParent = VisualTreeHelper.GetParent(child);
+            while (CurrentParent != null)
+            {
+                if (CurrentParent is T)
+                {
+                    parent = (T)CurrentParent;
+                    break;
+                }
+                CurrentParent = VisualTreeHelper.GetParent(CurrentParent);
+            }
+            return parent;
+        }
+    }
+}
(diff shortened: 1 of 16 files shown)


yaichenbaum:
@d2dyno1 Are there any noticeable improvements to memory usage with these changes?

d2dyno1:
> @d2dyno1 Are there any noticeable improvements to memory usage with these changes?

Not really, but by going step by step this will make moving out code out of Interaction easier, and eventually fix this memory leak. @yair100

d2dyno1:
In my next PRs I'll focus on removing as much code from `Interaction` as possible, moving functions out of this class and to ultimately remove `Interaction`. The reason I want Interaction gone, is to remove circular dependency between it and `IShellPage`. Moving out functions will also help us in the future make Files MVVM.

What are your thoughts? Help is also welcome ;)
@duke7553 @gave92 @yair100

Files/Interacts/Interaction.cs
```diff
@@ -894,13 +830,17 @@ private async void Manager_DataRequested(DataTransferManager sender, DataRequest
                 }
                 else if (item.PrimaryItemAttribute == StorageItemTypes.Folder)
                 {
-                    await AssociatedInstance.FilesystemViewModel.GetFolderFromPathAsync(item.ItemPath)
-                        .OnSuccess(folderAsItem => items.Add(folderAsItem));
+                    if (await StorageItemHelpers.ToStorageItem<StorageFolder>(item.ItemPath) is StorageFolder folder)
```
gave92, line 833: I'd also pass `AssociatedInstance` to the method like this:
```cs
await StorageItemHelpers.ToStorageItem<StorageFolder>(item.ItemPath, AssociatedInstance)
```
Difference being that this should also work inside e.g phone storage. Same thing below.

Might still be posssible to get rid of the AssociatedInstance by changing ToStorageItem() somehow..

Files/Interacts/Interaction.cs
```diff
@@ -1386,13 +1323,13 @@ private void JumpTimer_Tick(object sender, object e)
         public async Task<string> GetHashForFileAsync(ListedItem fileItem, string nameOfAlg, CancellationToken token, Microsoft.UI.Xaml.Controls.ProgressBar progress)
         {
             HashAlgorithmProvider algorithmProvider = HashAlgorithmProvider.OpenAlgorithm(nameOfAlg);
-            StorageFile itemFromPath = await AssociatedInstance.FilesystemViewModel.GetFileFromPathAsync((fileItem as ShortcutItem)?.TargetPath ?? fileItem.ItemPath);
-            if (itemFromPath == null)
+            StorageFile file = await StorageItemHelpers.ToStorageItem<StorageFile>((fileItem as ShortcutItem)?.TargetPath ?? fileItem.ItemPath);
```
gave92, line 1326: Same comment as above

gave92:
@duke7553 do you have any new insight about the memory leak issue? Do you think this "down-with-the-circular-references" movement can actually help when only "normal c# objects/classes" are involved?

duke7553:
@gave92 My position is the same as yours. While there is no evidence that these changes can improve the memory usage situation, it is still preferrable to have them for code quality purposes.

Files/Helpers/XamlHelpers/DependencyObjectHelpers.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Media;
+
+namespace Files.Helpers.XamlHelpers
+{
+    public static class DependencyObjectHelpers
+    {
```
duke7553, line 13: Excellent organization here! We want contributors to find these helpers with ease.

Files/Interacts/Interaction.cs
```diff
@@ -1329,17 +1266,17 @@ private void NewFile(ShellNewEntry itemType)
             CreateFileFromDialogResultType(AddItemType.File, itemType);
         }
 
-        public RelayCommand SelectAllContentPageItems => new RelayCommand(() => SelectAllItems());
+        public RelayCommand SelectAllContentPageItems => new RelayCommand(() => SelectAllItems(AssociatedInstance.ContentPage));
```
lukeblevins, line 1269: I'd suspect @yair100 has some good opinions about where we should place the RelayCommands for code quality.

lukeblevins (commented):
Looks good to me once @gave92 's feedback is addressed

Files/Interacts/Interaction.cs
yaichenbaum: @duke7553 I would make separate classes to group relay commands depending on the kind of actions. For example, clipboard actions could go be grouped together, selection commands could be placed in a separate class etc.

d2dyno1:
Comments addressed! @gave92 @duke7553

gave92 (approved):
Looks good. Thanks for this @d2dyno1

d2dyno1:
> Looks good. Thanks for this @d2dyno1

This is just the beginning :)

lukeblevins (approved):

PR #3977, opened 2021-03-11
Fixed json property name for pinned items

yaichenbaum:


gave92 (approved):
Is this why it was loosing pinned folders?

yaichenbaum:
@gave92 It seems like it, that's the only thing that changed regarding how pinned items are saved.

Issue #17, opened 2019-02-12
Files Instalation Failed


vallerydelexy:
![image](https://user-images.githubusercontent.com/17786532/52607428-c2da9880-2ea8-11e9-9812-570b849a36d6.png)
App Installer failed to install package dependencies. Ask the developer for Microsoft.VCLibs.140.00.Debug package. 

~\Files_0.4.5.0\Files_0.4.5.0_x64_Debug_Test\Files_0.4.5.0_x64_Debug.appx

vallerydelexy:
![image](https://user-images.githubusercontent.com/17786532/52607628-8ce9e400-2ea9-11e9-9cf4-e06dcee2c7d3.png)
Either a new certificate must be installed for this app package or you need a new app package with trusted certificates. Your system administrator or the app developer can help. A certificate chain was processed but terminated in a root certificate that isn't trusted (0x800B0109)
~\Files_0.4.5.0\Files_0.4.5.0_x64_Debug_Test\Files_0.4.5.0_x64_Debug.appx
after failing to install once

duke7553:
Yeah, there should be a powershell script which automates the whole side load process. Try running that first, and we’ll reopen if needed.

PR #3979, opened 2021-03-11
Possible fix NullReferenceException (AppCenter)

d2dyno1:
Possible fix for AppCenter report 

Report: #1254168959u

### DIFF
--- a/Files/Helpers/AdaptiveLayoutHelpers.cs
+++ b/Files/Helpers/AdaptiveLayoutHelpers.cs
@@ -2,19 +2,34 @@
 using Files.ViewModels.Previews;
 using System;
 using System.Linq;
+using System.Threading.Tasks;
 using Windows.Storage;
 
 namespace Files.Helpers
 {
     public static class AdaptiveLayoutHelpers
     {
-        public static bool PredictLayoutMode(FolderSettingsViewModel folderSettings, ItemViewModel filesystemViewModel)
+        public static async Task<bool> PredictLayoutMode(FolderSettingsViewModel folderSettings, ItemViewModel filesystemViewModel)
         {
             if (App.AppSettings.AreLayoutPreferencesPerFolder && App.AppSettings.AdaptiveLayoutEnabled && !folderSettings.LayoutPreference.IsAdaptiveLayoutOverridden)
             {
                 bool desktopIniFound = false;
 
-                var iniPath = System.IO.Path.Combine(filesystemViewModel.WorkingDirectory, "desktop.ini");
+                string path = filesystemViewModel?.WorkingDirectory;
+
+                if (string.IsNullOrWhiteSpace(path))
+                {
+                    return false;
+                }    
+                else
+                {
+                    if (!(await StorageItemHelpers.ToStorageItem<StorageFolder>(path) is StorageFolder folder) || folder == null)
+                    {
+                        return false;
+                    }
+                }
+
+                var iniPath = System.IO.Path.Combine(path, "desktop.ini");
                 var iniContents = NativeFileOperationsHelper.ReadStringFromFile(iniPath)?.Trim();
                 if (!string.IsNullOrEmpty(iniContents))
                 {
--- a/Files/Helpers/AdaptiveLayoutHelpers.cs
+++ b/Files/Helpers/AdaptiveLayoutHelpers.cs
@@ -14,7 +14,7 @@ namespace Files.Helpers
             {
                 bool desktopIniFound = false;
 
-                var iniPath = System.IO.Path.Combine(filesystemViewModel.CurrentFolder.ItemPath, "desktop.ini");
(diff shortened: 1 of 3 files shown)


yaichenbaum (approved):

winston-de:
@d2dyno1 FYI this crash came from me when I was debugging unrelated code I wrote in my own branch. I don't know how or why it made it to appcenter.

yaichenbaum:
@winston-de AppCenter is only setup for the store builds so this probably isn't from you.

winston-de:
I checked it, App Center shows a Surface Pro 6 on the latest windows dev build with English as the primary language, which is what I have. I also remember running into this error at the time it was reported in the same place in the code.
It was caused by me changing the startup code to allow for a selection switch.
Maybe my debug build got messed up somehow?

Files/Helpers/AdaptiveLayoutHelpers.cs
```diff
@@ -2,19 +2,34 @@
...
-                var iniPath = System.IO.Path.Combine(filesystemViewModel.CurrentFolder.ItemPath, "desktop.ini");
+                string path = filesystemViewModel?.WorkingDirectory;
+
+                if (string.IsNullOrWhiteSpace(path))
+                {
+                    return false;
+                }    
+                else
+                {
+                    if (!(await StorageItemHelpers.ToStorageItem<StorageFolder>(path) is StorageFolder folder) || folder == null)
```
gave92, line 26: Should not be needed. `ReadStringFromFile` does not crash if the file does not exist. Appcenter report the crash on line #17 so the change above is enough.

gave92 (changes requested):

Files/Helpers/AdaptiveLayoutHelpers.cs
d2dyno1: But we are then checking for files and folders in this directory. Having this check allows us to skip the rest if the folder is invalid

@gave92

gave92: This check with storagefiles is probably slower than doing the rest of the method :) Plus I'm not sure if it works for hidden folders

d2dyno1: Ok :)

Files/Helpers/AdaptiveLayoutHelpers.cs
```diff
@@ -2,19 +2,27 @@
...
 using System.Linq;
+using System.Threading.Tasks;
 using Windows.Storage;
 
 namespace Files.Helpers
 {
     public static class AdaptiveLayoutHelpers
     {
-        public static bool PredictLayoutMode(FolderSettingsViewModel folderSettings, ItemViewModel filesystemViewModel)
+        public static async Task<bool> PredictLayoutMode(FolderSettingsViewModel folderSettings, ItemViewModel filesystemViewModel)
```
gave92, line 12: Since there are no more `await` in the function this can go back to be a plain `public static bool PredictLayoutMode()`

d2dyno1: https://github.com/files-community/Files/pull/3979/commits/bc1b46a1842ab9d058811a7caa2a6f153998db88 :+1:

Files/Helpers/AdaptiveLayoutHelpers.cs
```diff
@@ -2,6 +2,7 @@
 using Files.ViewModels.Previews;
 using System;
 using System.Linq;
+using System.Threading.Tasks;
```
gave92, line 5: Aaand this can go away too... this is the last thing, promise.

Files/Helpers/AdaptiveLayoutHelpers.cs
```diff
@@ -14,7 +15,14 @@ public static bool PredictLayoutMode(FolderSettingsViewModel folderSettings, Ite
             {
                 bool desktopIniFound = false;
 
-                var iniPath = System.IO.Path.Combine(filesystemViewModel.CurrentFolder.ItemPath, "desktop.ini");
+                string path = filesystemViewModel?.WorkingDirectory;
+
```
gave92, line 19: And there is an extra new line?

Files/Helpers/AdaptiveLayoutHelpers.cs
d2dyno1: You wouldn't be able to sleep well, eh? 😆

d2dyno1: Same 😅

d2dyno1: Fixed in https://github.com/files-community/Files/pull/3979/commits/cd49e36b2b2eb6ee7f5c152c7085ff0a29033bc5 :+1: @gave92

gave92 (approved):

PR #4014, opened 2021-03-12
Use single instance of AppServiceConnection for preview pane thumbnails

winston-de:
Before it was loading a new one each time, which I think may have caused other issues after continued use.

### DIFF
--- a/Files/Helpers/FileThumbnailHelper.cs
+++ b/Files/Helpers/FileThumbnailHelper.cs
@@ -8,10 +8,11 @@ namespace Files.Helpers
 {
     public static class FileThumbnailHelper
     {
+        private static AppServiceConnection AppServiceConnection { get; set; }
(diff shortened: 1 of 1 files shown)


yaichenbaum (approved):

gave92:
Careful! Does this work after minimizing and restoring the app? I think it may end up using a non-null but closed connection. Also before it wasn't creating a new connection every time.

winston-de:
@gave92 Oh? I thought that Instance returned a function to create a new connection.
In that case, this PR is pointless and should be reverted.  
My mistake, sorry for this.

gave92:
No problem. Previous code was:
```cs
public static Task<AppServiceConnection> Instance = BuildConnection();
 ....
var connection = await Instance;
```
This should be causing a new connection to be created only the first time "await Instance" is called, after that it should be returning an already completed Task with the existing connection as result. If you find that this does not hold true please let me know xD

PR #4016, opened 2021-03-12
[Farewell, Interaction]: Step 1 of removing the Interaction class

d2dyno1:
This is the very first part of removing the `Interaction` class. Since the overall scope is very large, I've split it into multiple parts. For #4180 

Many UI event handlers have been moved to `GenericFileBrowser` and `GridViewBrowser`. This is not ideal because a lot of the code is duplicated between these classes. I propose the best MVVM solution I could think of. The way it works is, we have a `BaseLayoutCommandsViewModel` class in `BaseLayout` which contains all UI `ICommand`s and those commands call functions in `IBaseLayoutCommandImplementationModel` interface that contains implementation for these commands.
Since this is a big change that'll last for long, we must choose right and pick a sufficient solution.
Side note: The work hasn't been done using this pattern yet but classes and [example usage](https://github.com/files-community/Files/pull/4016/files#diff-bb62dc9582be4cc1b4eb63839013d0c6d45be5dab7ad79e67a425be666e0b5caR7) have been added.

**Update**
This pattern has been approved and implemented. 

Let me know what do you think of this solution! @yair100 @duke7553 @gave92


d2dyno1:
> Many UI event handlers have been moved to `GenericFileBrowser` and `GridViewBrowser`. This is not ideal because a lot of the code is duplicated between these classes. I propose the best MVVM solution I could think of. The way it works is, we have a `BaseLayoutCommandsViewModel` class in `BaseLayout` which contains all UI `ICommand`s and those commands call functions in `IBaseLayoutCommandImplementationModel` interface that contains implementation for these commands.
> Since this is a big change that'll last for long, we must choose right and pick a sufficient solution.
> Side note: The work hasn't been done using this pattern yet but classes and [example usage](https://github.com/files-community/Files/pull/4016/files#diff-bb62dc9582be4cc1b4eb63839013d0c6d45be5dab7ad79e67a425be666e0b5caR7) have been added.

@duke7553 @gave92 @yaichenbaum I need to have your feedback before further work is done on this PR

d2dyno1:
Ready for review! @yair100

d2dyno1:
@duke7553 fyi. This PR also replaces `BaseLayout.ContentPage` with `IBaseLayout.ContentPage`

Files/Views/LayoutModes/GridViewBrowser.xaml.cs
```diff
@@ -471,5 +483,75 @@ private void FileList_DoubleTapped(object sender, DoubleTappedRoutedEventArgs e)
                 ParentShellPageInstance.InteractionOperations.OpenSelectedItems(false);
             }
         }
+
+        private void QuickLook_Click(object sender, RoutedEventArgs e)
```
duke7553, line 487: @d2dyno1 To reduce duplicate code, couldn't we move these click event handlers into BaseLayout?

d2dyno1: @duke7553 I removed the handlers and bound the commands in https://github.com/files-community/Files/pull/4016/commits/b406f2507ef27055662eb65091c43227bd8e15dd

d2dyno1: The originally inteded way

Files/ViewModels/Properties/FileProperties.cs
```diff
@@ -337,5 +338,64 @@ private async void ViewModel_PropertyChanged(object sender, System.ComponentMode
                     break;
             }
         }
+
+        #region Private Helpers
```
yaichenbaum, line 342:
```suggestion
```

Files/ViewModels/Properties/FileProperties.cs
```diff
@@ -337,5 +338,64 @@ private async void ViewModel_PropertyChanged(object sender, System.ComponentMode
...
+            inputStream.Dispose();
+            stream.Dispose();
+            if (token.IsCancellationRequested)
+            {
+                return "";
+            }
+            return CryptographicBuffer.EncodeToHexString(hash.GetValueAndReset()).ToLower();
+        }
+
+        #endregion
```
yaichenbaum, line 399:
```suggestion
```

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -102,6 +107,11 @@ public GenericFileBrowser()
             AllView.PointerCaptureLost += AllView_ItemPress;
         }
 
+        protected override void InitializeCommandsViewModel()
+        {
+            CommandsViewModel = new BaseLayoutCommandsViewModel(new BaseLayoutCommandImplementationModel(ParentShellPageInstance));
```
yaichenbaum, line 112: Do you need to spin up a new view model for every instance?

Files/Views/PaneHolderPage.xaml.cs
```diff
@@ -36,6 +38,8 @@ public sealed partial class PaneHolderPage : Page, IPaneHolder, ITabItemContent,
 
         private readonly GridLength CompactSidebarWidth;
 
+        public ICommand EmptyRecycleBinCommand { get; private set; }
```
yaichenbaum, line 41: Shouldn't there should be a reference to `BaseLayoutCommandsViewModel`?

yair100 (changes requested):

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
d2dyno1: @yair100 that's what was also done with Interaction case. IShellPage with Interaction was populated for every instance. It's just that it's more clear now.

d2dyno1: We could experiment with new solutions to this after the `Interaction` is completely removed. It's some pretty exciting performance benefits stuff that wasn't possible with `Interaction` being in IShellPage :eyes:

Files/Views/PaneHolderPage.xaml.cs
d2dyno1: Hmm.. That command appeared after resolving merge issues 🤔

yaichenbaum: It's something we can look at later on, it shouldn't hold us back from merging this PR.

yair100 (approved):

yaichenbaum:
@tsvietOK Do you have any thoughts on this change?

lukeblevins (approved):

PR #4017, opened 2021-03-12
Implementation of elevated full trust process

yaichenbaum:


duke7553 (commented):
Before we add any actual user-facing functionality, we should start designing a UI which transparently informs the user what will happen. In particular, taking ownership of folders.

yaichenbaum:
@duke7553 Absolutely!

gave92:
Probably already thought of, but while the FTP is running as admin we could show a "shield" icon in Files to let the user know this.

duke7553 (approved):

PR #4018, opened 2021-03-12
Prevent FullTitle string from being null in StatusCenter.cs

winston-de:
Fix for https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/3768192674u/overview:
```
StatusCenter_obj17_Bindings.Update_FullTitle (String obj, Int32 phase)
D:\a\1\s\Files\obj\x64\Release\UserControls\StatusCenter.g.cs at 566, line 29
System.ArgumentNullException: Null strings may not be marshaled in Windows Runtime arguments. Parameter name: sourceString
```

I believe the issue was caused by a null value being passed to AutomationProperties.Name, which binds to FullTitle.

### DIFF
--- a/Files/UserControls/StatusCenter.xaml.cs
+++ b/Files/UserControls/StatusCenter.xaml.cs
@@ -201,7 +201,7 @@ namespace Files.UserControls
         public string FullTitle
         {
             get => fullTitle;
-            set => SetProperty(ref fullTitle, value);
+            set => SetProperty(ref fullTitle, value ?? string.Empty);
         }
 
         #endregion Public Properties


yaichenbaum (approved):

gave92:
Thanks this issue has been already fixed (FullTitle was not always assigned in the constructor). While this adds a welcome additional check, this would be the only property to be handled like this.. still wondering what's so special about it. Have you experienced this crash with main?

winston-de:
@gave92 Sorry I didn't realize it was already fixed xD.

> While this adds a welcome additional check, FullTitle would be the only property to be handled like this.. still wondering what's so special about it. 

```AutomationProperties.Name``` for list view items doesn't seem to accept null values, and it binds to FullTitle but not other properties.

winston-de:
Should I close this?

gave92 (approved):

gave92:
Current fix works "by luck" since currently the status banner is always created with an empty (but not null) title. So this is good to have, thanks @winston-de

ekrem95 (approved):

PR #4019, opened 2021-03-12
Moved some hard coded strings for the status center to string resources

winston-de:


R3voA3:
Are the default texts equal to the ones used in Windows 10 right now?

winston-de:
@R3voA3 I don't quite understand what you mean. Could you please elaborate?

R3voA3:
I just wanted to make sure that for example

<value>Close tabs to the right</value> is also called like this in Windows for consistency (In this case it is called like this in Edge as well), but I guess we can always handle that through localisation later on.

yaichenbaum (approved):

PR #4023, opened 2021-03-12
Revert "Use single instance of AppServiceConnection for preview pane thumbnails"

winston-de:
Reverts files-community/Files#4014
As @gave92 pointed out, a new connection was not being created each time, which was the motive for the original pr, and also that the new method would cause more issues when the app is restored from suspension.
I apologize for this mistake.

### DIFF
--- a/Files/Helpers/FileThumbnailHelper.cs
+++ b/Files/Helpers/FileThumbnailHelper.cs
@@ -8,11 +8,10 @@ namespace Files.Helpers
 {
     public static class FileThumbnailHelper
     {
-        private static AppServiceConnection AppServiceConnection { get; set; }
         public static async Task<(byte[] IconData, byte[] OverlayData, bool IsCustom)> LoadIconOverlayAsync(string filePath, uint thumbnailSize)
         {
-            AppServiceConnection ??= await AppServiceConnectionHelper.Instance;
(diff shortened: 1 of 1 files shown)


Files/Helpers/FileThumbnailHelper.cs
```diff
@@ -8,19 +8,18 @@ namespace Files.Helpers
...
+            if (connection != null)
             {
                 var value = new ValueSet
                 {
                     { "Arguments", "GetIconOverlay" },
                     { "filePath", filePath },
                     { "thumbnailSize", (int)thumbnailSize }
                 };
-                var response = await AppServiceConnection.SendMessageAsync(value);
+                var response = await connection.SendMessageAsync(value);
```
gave92, line 22: This PR makes me notice that the above should be changed to `connection.SendMessageSafeAsync()` which handles exceptions and is used everywhere else. Probably best done in a new PR.

gave92 (approved):

Files/Helpers/FileThumbnailHelper.cs
winston-de: @gave92 Got it, I'll make it once this is merged to avoid conflicts.

PR #4030, opened 2021-03-13
Fixed issue where ScrollIntoView could throw argument exception

winston-de:
I noticed that this would, on occasion, throw something like ```Argument Exception: Row index could not be found```, which would cause a crash.

This PR adds a catch for that.

### DIFF
--- a/Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
+++ b/Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
@@ -208,7 +208,14 @@ namespace Files.Views.LayoutModes
 
         public override void ScrollIntoView(ListedItem item)
         {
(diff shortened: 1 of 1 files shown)


gave92 (approved):

gave92 (changes requested):
Thanks for this change. For coherence I'd add the same to GridViewBrowser

gave92 (approved):

winston-de:
> Thanks for this change.
> 
> ~For coherence I'd add the same to GridViewBrowser (even if the crash has been noticed only in ListView).~ nevermind xD
> 
> Edit: Mm, now I wonder if `FocusSelectedItems` could cause the same issue.

@gave92 I havent seen this occur with the grid view, but I'll test some more to be certain.

Issue #3855, opened 2021-03-06
Files freezes when right-clicking files/folders


kimlimjustin:
**Describe the bug**
Files freeze when right-clicking files/folders and sometimes it crashes. And if this happens, closing and reopen Files will showing the Home tab instead of continuing where you left off.

**To Reproduce**
1. Open Files
2. Right click on a folder/files
3. Files freeze/crashes (it's more frequent to freeze rather than crashes)

**Expected behavior**
Files doesn't freeze or crash

**Screenshots**
https://user-images.githubusercontent.com/61906684/110193845-05dd7f00-7e69-11eb-9fff-0a1271564aad.mp4

**Desktop (please complete the following information):**
 - OS Version: Microsoft Windows 10 Home Single Language 10.0.18363 N/A Build 18363
 - App version: 1.3.0.0

**Additional context**
Add any other context about the problem here. Does this problem occur again after restarting the app? Yes

**Log file**
[debug.txt](https://github.com/files-community/Files/files/6094620/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/6094621/debug_fulltrust.txt)

gave92:
Is there any difference if you wait 1 sseconds before right clicking?

kimlimjustin:
> Is there any difference if you wait 10 seconds before right clicking? (until a few more items appear on the sidebar)

No, no difference at all.


https://user-images.githubusercontent.com/61906684/110195919-77222f80-7e73-11eb-8243-546aec18e2f6.mp4

Abdennournemr213:
same, first time using the app

gave92:
@kimlimjustin could you try the following?
- create an empty folder
- set Files to open this folder on startup
- check if right-click is working

kimlimjustin:
> check if right-click is working (in the folder empty space)

No, it's not working and resulting Files freeze.

> do shortcut files (.lnk) appear for you in Files?

Sometimes I need to wait a long time to let it appears in Files, but right-clicking also not working too after it appears. But sometimes although I waited for a long time, it still doesn't appear, IDK why tho.

> have you had this issue with previous versions of Files?

I remember version 1.1.0 had no this bug, but I'm not really sure (I know it's 1.1.0 because I opened a bug soon after trying Files and I check back that version). The thing I can sure of is the bug occurs recently.

Abdennournemr213:
> @kimlimjustin could you try the following?
> 
> * create an empty folder
> * open Files
[...]

PR #4033, opened 2021-03-13
Use named pipe instead of appservice

gave92:
@yair100, @winston-de if you've got time, could you test if this PR would fix the issues people are experiencing on newer Windows Insider builds? (E.g. only basic items in context menu, no shortcut files, etc..)

[Even if it works, it may still better to _not_ merge this and wait for Microsoft to fix the issue, unless this method proves beneficial for stability/speed]


winston-de:
@gave92 It works!

duke7553:
@gave92 Great minds think alike. I was just researching alternative IPC methods for the FullTrustProcess and was going to suggest the idea of a named pipe to you.

gave92:
Thanks @duke7553, actually got the idea of using named pipes from Rx-Explorer 😅. Anyway I haven't tested this much yet but it looks to be working, let me know if you like the changes and if you think it'd make sense to merge this.

gave92:
Has anyone experienced #3855? I'd be curious to know if this PR could fix it.

winston-de:
> Has anyone experienced #3855? I'd be curious to know if this PR could fix it.

@gave92 yeah, I had that issue quite a bit. And it does seem to be fixed with this pr.

yaichenbaum:
@gave92 This seems to not only solve the issue on insider builds, but it seems to also reduce the latency when right clicking on items.

yair100 (approved):

duke7553:
@gave92 I'm noticing that changing the layout mode while thumbnails are being loaded results in a hang with output being flooded with a System.OperationCanceledException exception: 
image

I'm also trying to reproduce a bug where thumbnails load in blurry and the FullTrustProcess completely dies 😀

gave92:
@duke7553 I'm pretty sure both issue existed before this PR :) The first is due to the use of cancellationtoken to stop LoadExtendedItemProperties, the second due to an AccessViolationException that sometimes is thrown in the FTP when loading thumbnails.

PR #4053, opened 2021-03-14
Updated default value for the acrylic setting

yair100:


Files/ViewModels/SettingsViewModel.cs
```diff
@@ -515,7 +515,7 @@ public bool PinRecycleBinToSideBar
         /// </summary>
         public bool IsAcrylicDisabled
```
d2dyno1, line 516: That's a weird setting name - `IsAcrylicDisabled`. Wouldn't it be better if it was called `IsAcrylicEnabled`?

gave92 (approved):

Files/ViewModels/SettingsViewModel.cs
yaichenbaum: @d2dyno1 This was added before we started using converters, we can definitely change around the name. Do you want to work on this?

gave92:
Sort of unrelated: perhaps the setting to show the library section should also be enable by default? Users who update will likely think they lost those pinned folders and go to add those back manually ;)

Files/ViewModels/SettingsViewModel.cs
d2dyno1: @yair100 sure!

yaichenbaum:
@gave92 We can show the library section by default if that's what users are looking for, but users who updated from previous versions will still have the library items pinned to the favorites section.

gave92:
@yair100 it's all right then. I haven't noticed because when I tested that PR I lost all my pinned folders 😂

yaichenbaum:
@gave92 That's been resolved, I changed the property name we stored in the json file by mistake.

PR #4060, opened 2021-03-14
Added a badge to the status bar to indicate the status of the full trust process

### LINKED ISSUE #4036: Add UI to indicate when the full trust process is elevated

yair100:
Closes #4036


gave92:
Button is bigger and has a different background. Let me know if you wish to fix this here.

image

Plus the background of the teaching tip is too dark (like sometimes the context menu is but with this the issue seems consistent)

image

yaichenbaum:
@gave92 Nice! I put set the transparent background on the wrong control 😆.

gave92:
@yair100 the foreground color might be too much though, should I remove it? I wanted to distinguish it a bit from the others

image

yaichenbaum:
@gave92 I think we should remove the color for now.

gave92 (approved):

Issue #3869, opened 2021-03-07
Enhancement: Improving the library section on the sidebar


yaichenbaum:
- [x] Remove desktop item from library section.
- [x] The app requires a restart after hiding or showing the library section.
- [x] Right clicking on a library item shows the option to unpin the item, the unpin option should only show when right clicking on a pinned item.
- [x] Pinned folders are not saved
- [x] You can add the same folder to the sidebar multiple times
- [x] If you delete the PinnedItems.json file, on the next startup libraries folders will be added to the first sidebar section (not to the library section) -> this is apparently by design? By default the documents folder will be shown in both sections.
- [x] Libraries are displayed in English instead of the system language
- [ ] Code in LibraryManager has unrelated method names (e.g. `EnumerateDrivesAsync`)

yaichenbaum:
@gave92 It seems like right clicking on library items will still show the unpin from sidebar option.

yaichenbaum:
@d2dyno1 Do you want to work on this?

lukeblevins:
@yair100 This seems to be completely resolved now.

yaichenbaum:
@duke7553 The last item in the list doesn't seem to have been implemented yet. Let's reopen this for now.

PR #4086, opened 2021-03-15
Removed desktop path from the library section

yair100:
For #3869


Files/Filesystem/LibraryManager.cs
```diff
@@ -110,7 +110,6 @@ private async Task SyncLibrarySideBarItemsUI()
                             MainPage.SideBarItems.Insert(1, librarySection);
 
                             libraryItems.Clear();
-                            libraryItems.Add(AppSettings.DesktopPath);
```
d2dyno1, line 113: Why? It was useful to me!

yaichenbaum: @d2dyno1 The desktop path isn't a default library, until we add proper support for libraries, I think we should try to follow the default behavior.

gave92: Then should we move desktop folder to the first section with the recyclebin?

XPoppyX:
"Downloads" is not part of the library either,.....I think

tsvietOK:
I think both Downloads and Desktop are parts of library.
image

XPoppyX:
Captura de pantalla 2021-03-15 185206

tsvietOK:
@XPoppyX Agreed. Desktop and Downloads are still usable in Libraries section.

lukeblevins:
@yair100 We currently don't fully support libraries in Files. While this can change, we need to rename the Libraries section to Locations so users don't get false expectations regarding the grouping functionality available in Explorer.

d2dyno1:
> @yair100 We currently don't fully support libraries in Files. While this can change, we need to rename the Libraries section to Locations so users don't get false expectations regarding the grouping functionality available in Explorer.

How about naming it "Quick Access"?

Files/Filesystem/LibraryManager.cs
yaichenbaum: @gave92 The desktop location is already one of the default locations pinned to the favorite section.

yair100:
> @yair100 We currently don't fully support libraries in Files. While this can change, we need to rename the Libraries section to Locations so users don't get false expectations regarding the grouping functionality available in Explorer.

@duke7553 @nvi9 is working on adding proper support for libraries so that shouldn't make a difference for long.

Files/Filesystem/LibraryManager.cs
gave92: Keep forgetting about that, for me it's a yes

gave92 (approved):

PR #4120, opened 2021-03-16
Hide network section in sidebar when no drives are detected

### LINKED ISSUE #3924: Enhancement: Hide network section in the sidebar when no network drives are detected

Daninator1:
Closes #3924 

I just added an additional check before creating the network drive LocationItem.

This is my first PR, feedback is appreciated 🙂

### DIFF
--- a/Files/Filesystem/NetworkDrivesManager.cs
+++ b/Files/Filesystem/NetworkDrivesManager.cs
@@ -35,6 +35,7 @@ namespace Files.Filesystem
         {
             var networkItem = new DriveItem()
             {
+                DeviceID = "network-folder",
                 Text = "Network".GetLocalized(),
                 Path = App.AppSettings.NetworkFolderPath,
                 Type = DriveType.Network,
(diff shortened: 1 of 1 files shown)


XPoppyX:
Still visible to me
Captura de pantalla 2021-03-16 190358

Daninator1:
Oh, seems like I missed that, will alter the check tomorrow probably, thanks!

Files/Filesystem/NetworkDrivesManager.cs
```diff
@@ -112,7 +112,7 @@ private async Task SyncSideBarItemsUI()
                     MainPage.SideBarItems.BeginBulkOperation();
 
                     var section = MainPage.SideBarItems.FirstOrDefault(x => x.Text == "SidebarNetworkDrives".GetLocalized()) as LocationItem;
-                    if (section == null)
+                    if (section == null && this.drivesList.Count > 0)
```
yaichenbaum, line 115: I don't think this will work, not all mapped drives will have a letter.

Daninator1:
Alright, so I adapted the check to include the network folder which gets added to the list before. As I had no way to tell if the DriveItem is in fact just the network folder, I added a custom device id to perform the check. The other solution would have been to go strictly by the count of the items (Count > 1), although that wouldn't have been as flexible for future changes. Any thoughts on that?

Files/Filesystem/NetworkDrivesManager.cs
Daninator1: I'm not sure if I get your point about drive letters, any detected network drives should be in that list (+ network folder), so my thought was that as long as there are no drives in that list (apart from the network folder itself) we do not create the section in the sidebar. Or did I get something wrong?

gave92 (approved):

Issue #4115, opened 2021-03-16
Add support for third party/optional widgets


d2dyno1:
**Describe the solution you'd like**
Add support for third party widgets to the widgets tab via app extensions -> add third-party Weather widget integration in Files to match the upcoming Sun Valley Windows update.

**Additional Context**
Add more widgets to Files.

gave92:
Having the possibility to add new widgets using app extensions it's very interesting. I'd probably not include a weather widget by default though :)
> Mainly we get to use more weather widgets around Windows

There are going to be too many weather widgets already, no need to add ours xD

d2dyno1:
> Having the possibility to add new widgets using app extensions it's very interesting. I'd probably not include a weather widget by default though :)

:/ This was discussed on the [UWP Community](https://discord.gg/X4kZhfw8) server though. @gave92

gave92:
Oooh a weather widget in a file manager.. ok why not.

gave92:
Still makes 0 sense to me to have that included by default but if it's third party / option totally agreed 😂

FireCubeStudios:
We need this. Weather widgets will provide useful insights in file management operations for example if there is a weather widget a user will be able to know the current temperature of the weather outside and use that information to process how many files to delete in a folder for a peak efficient file deletion temperature ratio to prevent overheating of CPU and increase the life span of the CPU which delays the need to replace the CPU or computer in the long run and saves money for the user.

winston-de:
I got optional packages set up and working, I can merge it so that the code can be written to utilize them if its wanted.
![image](https://user-images.githubusercontent.com/59544401/111382928-cda61f80-8664-11eb-9079-2dc6f884ec74.png)

d2dyno1:
Quick weather widget concept I did in Paint 3D:

![20210316_221230](https://user-images.githubusercontent.com/53011783/111384978-6dd26780-86aa-11eb-8c0f-87316d5ebbc2.png)

yaichenbaum:
> I got optional packages set up and working (as was discussed in discord), I can merge it so that the code can be written to utilize them if its wanted.
> ![image](https://user-images.githubusercontent.com/59544401/111382928-cda61f80-8664-11eb-9079-2dc6f884ec74.png)

That would be great! Do you think we can do something similar for #3133?

winston-de:
@yair100 Totally!

winston-de:
Interestingly, you have to use a workaround to get optional packages to work with xaml controls.
[...]

Issue #4116, opened 2021-03-16
Enhancement: Move the options to customize widgets to the home screen


yaichenbaum:
- [x] Remove the widgets settings page.
- [x] Add a flyout to the home page with a list of widgets that users can add to the home screen.
- [ ] Add a button that will show the option to unpin widgets and reorder them #2263.

winston-de:
Maybe something similar to the edge ntp page?
![image](https://user-images.githubusercontent.com/59544401/111372757-4b176300-8658-11eb-9bf5-1b2489fe044c.png)

yaichenbaum:
@winston-de I was thinking of having it work closer to the way you can add widgets to the home screen on Android.

yaichenbaum:
@winston-de Since there are only a couple widgets right now, I ended up using that design, thanks for the idea!

mangoepic:
I really like this feature idea because there are like 10 buttons on Home page, one of which i occasionally use. I would probably like to just get it clean
![image](https://user-images.githubusercontent.com/71931749/118393321-de601b00-b63e-11eb-871c-78cb065e7ef6.png)

MalteTasler:
close?

PR #4125, opened 2021-03-17
Added sample weather widget as optional package

winston-de:
#4115 
This widget is provided entirely through an optional package. 

image
image

To test it:
1. set config to release
2. Deploy the solution
3. Manually deploy SampleWidget project
4. Launch files

The sample widget can only be deployed after Files has been deployed, or else it will throw an error.


yaichenbaum:
@winston-de A couple notes, how does this work before #4116 is done? Also, this widget should be kept separate from Files. We can create a separate repo for it since it's a good sample.

winston-de:
@yair100 I dont think we can keep the optional packages completely separate from this repo, since the winmd files outputted by the sample project have to be referenced by the Files project in order to compile.

Also, this doesn't have to be merged immediately, I think it will be useful for testing. I'll mark it as draft.

d2dyno1:
I could work on the weather's appearance and backend.

FireCubeStudios:
> I could work on the weather's appearance and backend.

make it not webview

d2dyno1:
@FireCubeStudios I have something similar in mind:

des3

winston-de:
@yair100 My idea worked. It is possible to completely separate the optional packages from the main Files repo. Even better, despite what I previously thought, it doesn't even require any modifications to the build pipeline. All that has to be done is add a reference to a nuget package which contains the metadata outputted by the widget build.

winston-de:
Here's the link to the repo.
https://github.com/winston-de/WidgetSample/
The nuget package but published and referenced (also unlisted) so it should work.
If you'd like, I can transfer ownership over to you.

Jaiganeshkumaran:
Weather widgets are everywhere these days. Why does Files now need a weather widget? What's the use?

d2dyno1:
@Jaiganeshkumaran See #4115

winston-de:
For the time being, I will close this because there isn't much use for this. However, I do think optional packages warrant further investigation in the future as a way to allow for components of Files to be developed, installed, and published separately, reducing build times, install size, and making the project easier to maintain.

PR #4161, opened 2021-03-18
MainPage MVVM

d2dyno1:
- For #4180

**Details of Changes**
MainPage now follows MVVM pattern.

**Validation**

- [x] Built and ran the app


Files/ViewModels/MainPageViewModel.cs
```diff
@@ -0,0 +1,255 @@
...
+using Windows.System;
+using Windows.UI.Xaml.Input;
+using Windows.UI.Xaml.Navigation;
+using static Files.Views.MainPage;
+
+namespace Files.ViewModels
+{
+    public class MainPageViewModel : ObservableObject
+    {
+        #region Private Members
```
duke7553, line 19: @d2dyno1 FYI: I'd prefer we don't use regions in the code.

d2dyno1: Why? 😧
They really help a lot with grouping code, making it look clean. Moreover, you can collapse a #region section in VS leaving the focus on the code that's important to you.

d2dyno1: @yair100 It doesn't affect the performance. It only makes code look more clean and doesn't make it look like one big mess which is very apparent with `ItemViewModel` and `Interaction` classes.

d2dyno1: >  we shouldn't be using regions.

Why is that? @yair100

d2dyno1: :shipit:

tsvietOK: 1. Regions are painful for other developers inspecting your the first time. They first have to expand all the collapsed sections to get the full picture of the code. And because of all the #regionand #endregion lines the code gets even larger.
2. Regions often lead to bad code because developers easily put a region around larger code blocks to make them hide instead of thinking how to structure their code well.
[Source](https://marcduerst.com/2016/10/03/c-regions-are-evil/)

d2dyno1: These are not viable points for me. @tsvietOK 
Addressing the first point, regions can be easily expanded and collapsed by the following shortcuts: `Ctrl + M, P` and `Ctrl + M, O`.
> And because of all the #regionand #endregion lines the code gets even larger.

Assuming you'll have an average maximum of 10 region directives per class, it doesn't make the code larger whatsoever.

re 2.
It really all depends on the person who wrote the code. For instance the `ItemViewModel` class has some massive functions of which the complexity is "Very High" (CodeFactor) and there aren't any `#region`s enclosements used at all. In other words, the code doesn't degrade with `region` directives use.

That being said the arguments brought by me **should** be taken into consideration. @tsvietOK @yair100

d2dyno1: (@tsvietOK the opposite of article referenced by you say [C# Corner]("Using C# Regions To Improve Code Readability" https://www.c-sharpcorner.com/article/using-regions-to-improve-code-readability/))

d2dyno1: I'll remove regions in this PR because I don't think my opinion matters in the end, because a "quick" update to the codestyle guidelines will probably soon roll-out as it did before from my past experience

duke7553: @d2dyno1 Your opinion matters. We make changes to code style guidelines based on the consensus of contributors and to prevent mistakes like this from happening after PRs have been opened initially.

Regarding regions, I think there are better ways we can simplify the code besides hiding entire blocks completely. Clean code should be more than aesthetics and you'll find that most other languages don't rely on the concept of regions. With that said, I think we should work together to compromise on a new code style guideline for regions that maybe allows them for very specific situations, but not everywhere in every file. Perhaps we can think of them as a way to group _some_ functionality instead of assign names to already self-explanatory constructs.

d2dyno1:
@yair100 @duke7553 Ready!

Files/DataModels/SidebarPinnedModel.cs
```diff
@@ -96,7 +97,7 @@ public async void AddItem(string item)
 
         public async Task ShowHideRecycleBinItemAsync(bool show)
         {
-            await MainPage.SideBarItemsSemaphore.WaitAsync();
+            await SidebarControl.SideBarItemsSemaphore.WaitAsync();
```
yaichenbaum, line 100: We shouldn't be referencing the sidebar control directly, technically, we should be able to spin up several sidebars and give each one their own list of items.

Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml
```diff
@@ -25,15 +25,15 @@
                         Opening="TabItemContextMenu_Opening">
                         <MenuFlyoutItem
                             x:Uid="HorizontalMultitaskingControlNewTab"
-                            Click="{x:Bind views:MainPage.AddNewTabAtIndex}"
+                            Click="{x:Bind vm:MainPageViewModel.AddNewTabAtIndex}"
```
yaichenbaum, line 28: It doesn't need to be included in this PR, but we should remove the reference to MainPageViewModel and implement proper dependency properties so that the control is completely decoupled.

Files/DataModels/SidebarPinnedModel.cs
d2dyno1: @yair100 But the `SideBarItemsSemaphore` was static and is static anyway. I just moved it there so it wasn't ugly placed in MainPage.

Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml
d2dyno1: @yair100 :+1: The function I'm referencing there is static anyway.

Files/ViewModels/MainPageViewModel.cs
```diff
@@ -0,0 +1,546 @@
...
+                            tabLocationHeader = currentFolder.DisplayName;
+                        }
+                    }
+                }
+            }
+
+            return (tabLocationHeader, fontIconSource);
+        }
+
+        public static string GetDriveTypeIcon(System.IO.DriveInfo drive)
```
yaichenbaum, line 339: Should this be moved to its own helper? I think it will be useful beyond MainPageViewModel.

d2dyno1: Will do!

Files/DataModels/SidebarPinnedModel.cs
yaichenbaum: Let's keep it in MainPageViewModel until we set it up properly since it will help keep the control decoupled.

Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml
yaichenbaum: @d2dyno1 It's fine for now since it was already like this, I opened an issue to track this. #4221

Files/DataModels/SidebarPinnedModel.cs
d2dyno1: @yair100 Can we just keep it where it is now? A quite a lot of files have been changed for this change and moving it to MainPageViewModel changes nothing.

yaichenbaum: It means that one more thing would break if the sidebar control was removed, however, we can keep it where it is for now. Can you open an issue to track this?

Files/Filesystem/LibraryManager.cs
```diff
@@ -107,8 +107,8 @@ private async Task SyncLibrarySideBarItemsUI()
                                 SelectsOnInvoked = false,
                                 ChildItems = new ObservableCollection<INavigationControlItem>()
                             };
-
-                            MainPage.SideBarItems.Insert(1, librarySection);
+
```
yaichenbaum, line 110: Can you remove the extra space?

yaichenbaum, line 110:
```suggestion
```

yair100 (approved):
Great job!

Issue #3966, opened 2021-03-11
Allow customising context menus


mbeps:
### Is your feature request related to a problem?

Being able to rearrange or completely remove certain entries in the context menus. This is useful as some apps add unnecessary entries to the context menus which clutter the menu. 

### Describe the solution you'd like

Include a new tab in the settings called _Context Menu_. Include a graphical solution for folders and different file types where the user can drag options or remove them. The removed items are stored in another place which can then be dragged and dropped if needed in the future. 
![image](https://user-images.githubusercontent.com/58662575/110772217-48310280-8253-11eb-8817-1961eaaa26b9.png)

### Requirements

-

### Files version

2.3.12.0

### Windows version

10.0.19043.2006

gave92:
@mbeps I know the following does not fully cover your suggestion but you can customize the menu context items in Files by using any software that customizes the windows explorer context menu (a google search will suggest a bunch of those)

mbeps:
@gave92 I don't really know of any good software. I use multiple utilities and registry tweaks.

gave92:
>  I use multiple utilities and registry tweaks.
Yeah changes done by any of those should be affecting Files context menu as well as Explorer's

josh-65:
I would like this feature too. It's kind of need at this point with all the options currently.

mdtauk:
I would really like to be able to arrange, add and remove context menu entries.

You could include the standard options, and leave those fixed in position and order, but allow arranging of items in the rest of the menu.

As a thought experiment, I look at the File Explorer context menus, and see so much cruft.  Where a modern app like Files should give a cleaner and better thought out experience.

![image](https://user-images.githubusercontent.com/7389110/121271728-3398f080-c8bc-11eb-8842-7919288f5279.png)

Tropix126:
It would also be nice to control whether the context menu was a CommandBarFlyout or not, and also animation timing/speed.

winston-de:
> It would also be nice to control whether the context menu was a CommandBarFlyout or not, and also animation timing/speed.

Worth noting for this, if a ``CommandBarFlyout`` has no primary items, only secondary items, the CommandBar part will be hidden making it look and act like a normal menu flyout.

mdtauk:
> It would also be nice to control whether the context menu was a CommandBarFlyout or not, and also animation timing/speed.
>
[...]

Issue #2780, opened 2020-12-30
Feature: Add command palette


winston-de:
**Is your feature request related to a problem? Please describe.**

One of my favorite features in Windows Terminal and Visual Studio Code is the command palette. It makes it very easy to quickly access commands from the keyboard without having to memorize a keyboard shortcut or grab the mouse and locate a button. 

This would also be very helpful when trying to find a shell command, because there can be quite a few of them to sift through in a context menu:

![image](https://user-images.githubusercontent.com/59544401/103386263-e4e85e80-4ab2-11eb-8fdb-3a4a98df9652.png)

I think this would also give Files more of an appeal over Windows Explorer, where finding the button you need is like finding a needle in a haystack.

Since Files is quickly gaining more features (which is great!), I think it may be worth considering this.

**Describe the solution you'd like**

A command pallet like the one in Windows Terminal would be nice.

**Describe alternatives you've considered**

**Additional context**
Windows terminal command pallet:
![image](https://user-images.githubusercontent.com/59544401/103385508-a3a27f80-4aaf-11eb-92a5-41910263a538.png)

**Possible commands**
- shell context menu items for selection
- cut, copy, paste selection
- enable dual pane
- switch to tab
- new tab
- close other tabs
- go up directory
- navigate back, forward
- search

heistema:
Would be awesome if this could get integrated into the search bar... So you have on bar for everything.

yaichenbaum:
@winston-de I think we can go ahead on this feature without a spec sheet since we can pretty much follow the design in Windows Terminal (obviously with our own tweaks to the design).

winston-de:
@yair100 Great!

Code wise, I think the best way to go about this is to have all the commands/actions implement a single interface, and then adding them to a master list that a ListView (the commands) can bind to.

What are your thoughts?

yaichenbaum:
@winston-de That's the goal, in fact, it will make implementing #1660 easier if all the code relating to keyboard shortcuts is in one place. We already are part of the way there with `Interaction.cs`, but there is a lot of room for improvement.

yaichenbaum:
@cinqmilleans we haven't created a design for this yet but it'll be a good idea to keep this in mind when planning the shortcut work.

QuaintMako:
This feature might be a breeze to implement once we're done with Rich Commands?

yaira2:
Yes

Issue #2342, opened 2020-11-15
Switch the right click context menu to CommandBarFlyout


yaichenbaum:
Here is a concept from https://twitter.com/zeealeid/status/1327710601691455488?s=20

![image](https://user-images.githubusercontent.com/39923744/99173711-596c6700-26d8-11eb-991e-16af32ebb164.png)

d2dyno1:
> Here is a concept from https://twitter.com/zeealeid/status/1327710601691455488?s=20
Does the author of this tweet know about Files? Maybe if we dm them, he could bring some creative ideas to Files... 🤔

PR #4190, opened 2021-03-21
Rebuild and merge context menu code

### LINKED ISSUE #2747: Code base: Merge context menus so they can be reused in different layouts
We currently reimplement the right click context menu in every layout. This leads to duplicate code and can make the context menu a little harder to work with.

The context menus that are reused can be merged so that changes will only need to be made once.

### LINKED ISSUE #4331: Create new file of type doesn't add specified extension
**Explain the bug**
I noticed the create new file no longer works with the type parameter.

**To reproduce**
1. Go to Folder
2. Click new -> Word doc or something
3. Enter name
4. File is not a word doc

**Expected behavior**
The file extension should be added.

**Logfile**
Nothing related.

**Screenshots**

https://user-images.githubusercontent.com/59544401/112729121-cf36d980-8ee7-11eb-9fde-78b453e24da5.mp4

**System Information**
 - OS Version: 21343.1000
 - App version: 1.5.2.0

**Additional comment**
Add any other comment about the problem here. Does this problem happen again after restarting the app?

winston-de:
**Resolved / Related Issues**

- Closes #2747
 
**Details of Changes**
This is still a work in progress.

#### Goals:
- Separate the context menu items from the MenuFlyout control (and UI in general) to make it easier to implement #2342 (and possibly #2780) in the future.
- Add code to allow for customization of items to make it easier to implement #3966

## ToDo:
- [x] Remove the ".Text" from the string resource keys, and use the localized strings
- [x] Add recycle bin items
- [x] Add the layout flyout
- [x] Clean up/remove old context menu code in BaseLayout
- [ ] Add pin to start commands (depends on #4218)
- [x] Implement IsChecked for the model
- [x] Move items that are only shown on shift to the more menu if shift isn't held 
- [x] Closes #4331 

**Validation**

- [x] Built and ran the app

**Screenshots (optional)**

Add screenshots here.


winston-de:
Alright, I got most of the basic menu items implemented.

Without & with shift held.
image

yaichenbaum:
@winston-de What still needs to be done?

winston-de:
@yair100 I still have to add support custom font families for the glyphs (which is why the shortcut icon is wrong).
Also
- [ ] Remove the ".Text" from the string resource keys
- [ ] Add recycle bin items
- [ ] Add the base layout flyout
- [ ] Clean up/remove old context menu code in BaseLayout

yaichenbaum:
@winston-de Sounds like good progress, this will make maintaining the context menu so much easier!

winston-de:
@yair100 I've updated all the localization keys (removed the ".Text"). Should I run the multilingual app toolkit?

winston-de:
I don't know what do do about the build error. It worked fine before I merged with main.
All it tells me is ```GENERATEPROJECTPRIFILE(0,0): Error : PRI175: 0x80073b0f - Processing Resources failed with error: Duplicate Entry. ```. 
I've tried turning up verbose logging to the "diagnostic" level, but it still doesn't show anything about where this is happening.
I've tried manually looking through each of the resource files changed in the merge for duplicates.
If anyone can help me debug this, it would be greatly appreciated.

winston-de:
It looks like the build blows up (the error in the previous comment) when I remove the ".Text" from the string resources, though I'm not really sure why. The only option I see is to create new ones, but that would mean loosing all the current translations.

winston-de:
Yay it worked!

duke7553:
@winston-de Great! Once this is finished, we can finally start on the CommandbarFlyout update for the newly unified flyout.

winston-de:
> @winston-de Great! Once this is finished, we can finally start on the CommandbarFlyout update for the newly unified flyout.

I did a little bit of testing, it should be a relatively simple change as well.

winston-de:
I've been noticing that there's been some issues with users not being able to find items that are hidden unless you shift+right click. What do you think of moving those items to the "more" menu when shift isn't held?

It would look something like this:

winston-de:
Inadvertently, this also mostly fixed #4331. I'll fix the nav toolbar one as well and just bundle the full fix in this pr.

Files/BaseLayout.cs
```diff
@@ -475,6 +424,11 @@ protected override async void OnNavigatedTo(NavigationEventArgs eventArgs)
             catch (Exception e)
             {
             }
+
+            ItemContextMenuFlyout.Opening += ItemContextFlyout_Opening;
+            ItemContextMenuFlyout.AreOpenCloseAnimationsEnabled = AppSettings.AreRightClickContentMenuAnimationsEnabled;
+            BaseContextMenuFlyout.Opening += BaseContextFlyout_Opening;
+            BaseContextMenuFlyout.AreOpenCloseAnimationsEnabled = AppSettings.AreRightClickContentMenuAnimationsEnabled;
```
duke7553, line 431: Should these events be unsubscribed in the OnNavigatingFrom() override?

lukeblevins (dismissed):
LGTM

winston-de:
I think we should wait for #4218 to be merged before merging this one, as it adds a command to pin and unpin items to the Start menu. Currently pin to start doesn't do anything since there isn't a command.

yair100:
> @yair100 I've updated all the localization keys (removed the ".Text"). Should I run the multilingual app toolkit?

@winston-de Sure! By the way, why do a lot of the keys have the number 2 attached to them?

> I've been noticing that there's been some issues with users not being able to find items that are hidden unless you shift+right click. What do you think of moving those items to the "more" menu when shift isn't held?

This sounds like a good idea!

winston-de:
> > @yair100 I've updated all the localization keys (removed the ".Text"). Should I run the multilingual app toolkit?
> 
> @winston-de Sure! By the way, why do a lot of the keys have the number 2 attached to them?

@yair100 I don't know why, but it wouldn't build unless I added them.

yaichenbaum:
@winston-de Have you checked to make sure there are no duplicate keys?

winston-de:
@yair100 Yes, there weren't any. I think the problem initially was that I was not changing all the keys, just then ones in the en-us file, so the compiler saw that as duplicate keys. Now that I'm changing them all, I can probably remove the 2s without any issue.

winston-de:
Ok, so I tried removing the 2 for one set of resources, and it built just fine. Then I removed them all, and the build failed with the duplicate error. I checked each resource key for duplicates, and didn't find any. I'm guessing one of them is causing the problem for whatever, but I have no clue which one (except that it's not BaseLayoutContextFlyoutLayoutMode).

yaichenbaum:
@winston-de Another possibility is that one of the resources are being used where they shouldn't and removing the `.Text` is throwing the error. The best solution I can think of for now is to go through each string one at a time.

winston-de:
@yair100 That would take me hours to do, are the 2s really that bad? It works fine as it is. And if so, is it possible that I could come back and fix them in their own PR?

winston-de:
Also, @TyJOrtiz is it alright with you that I make the necessary changes to implement the new context menus in Column View?

TyJOrtiz:
@winston-de Yes that's fine

yair100:
> @yair100 That would take me hours to do, are the 2s really that bad? It works fine as it is. And if so, is it possible that I could come back and fix them in their own PR?

I don't want to cause all that extra work, how about keeping the new keys, but can you change the names of the new keys to something like `RegularContextFlyout...`? This will make it easier to search for the older keys and remove them in a future PR.

winston-de:
@yair100 I can do that, but I don't see why you would to search through and remove the old keys, as they're gone. I changed the keys with find/replace throughout the entire solution so that the current translations wouldn't be lost.

winston-de:
The new context menus in Column View are up and running. 
image

Btw, @TyJOrtiz the column view is awesome! :)

yair100:
> @yair100 I can do that, but I don't see why you would to search through and remove the old keys, as they're gone. I changed the keys with find/replace throughout the entire solution so that the current translations wouldn't be lost.

There is a duplicate key somewhere, I will try to find it.

yaichenbaum:
@winston-de It seems like the sort by "Date deleted" option is showing in all directories, additionally, the layout menu items seem to be disabled.

winston-de:
> > @yair100 I can do that, but I don't see why you would to search through and remove the old keys, as they're gone. I changed the keys with find/replace throughout the entire solution so that the current translations wouldn't be lost.
> 
> There is a duplicate key somewhere, I will try to find it.

@yair100 Thanks for doing this.
I searched every key for anything with the same name throughout the entire solution, and each one had somewhere between 37-54 references (which corresponds to the number of languages it was translated to). If there is a duplicate, I think it exists in only one resource file. I don't know if that's of any help to you.

yair100 (approved):

winston-de:
@yair100 It looks like I may have never had to remove the ".Text" from everything.
I didn't realize it was possible to access them by using a slash, like this ```"resourcename/Text".GetLocalized()```.
I can change them all back if you'd prefer.

winston-de:
@yair100 I apologize for this, I just learned that the resources could be accessed like that.

Issue #4158, opened 2021-03-18
Bug: Double click to rename is crashing Files


yaichenbaum:
AppCenter log [3232660495u](https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/3232660495u/overview).

PR #4194, opened 2021-03-21
Fix crash due to wrong thread exception

gave92:
**Resolved / Related Issues**

Dragging and hovering on a folder / sidebar item is currently broken since timers actions are not invoked on the UI thread.
This caused, among others, #4158.

**Details of Changes**

Properly create timers so they invoke the action on the UI thread.

**Validation**

- [x] Built and ran the app


yaichenbaum (approved):
Nice catch!

Issue #4180, opened 2021-03-19
Feature: Improving Performance, optimizations, reliability and codebase 


d2dyno1:
**Is your feature request related to a problem? Please describe.**
Full details on the issues https://github.com/0x5bfa/Website/blob/5bfa/codebase-docs/src/routes/blog/posts/outlook-on-the-codebase-toward-vnext/%2Bpage.md

A large number of users have been recently complaining over the performance, optimization and reliablity with Files. These requests haven't been really addressed before (at least taken with enough seriousness).

**Describe the solution you'd like**
These problems really should be taken care of. We should start with upgrading the codebase and its structure. This will help us later to work with it and solve performance issues.

**Additional context**
Since a new update just dropped we have more time now to work on some of the backend stuff.

Please reference this issue in every PR that focuses on optimizations to keep track of the progress.

- [ ] Startup performance (in progress)
- [ ] Home page, favorites and drives
- [ ] Thumbnails
- [ ] Deleting large batches of files
- [ ] Less than 1% crash rate (this was achieved in Files v4, however, there are some issues that don't appear in Sentry, and the crash rate might be higher).
- [ ] Improve performance when resuming multiple tabs from the background (restoring multiple tabs is visibly slow)
- [ ] Improve ram usage
- [ ] Search causes CPU usage to increase

yaira2:
I updated the description with the latest feedback from users.

yaira2:
We're excited to share that startup performance will be significantly improved in the upcoming release of version 3. You can download the preview version from our website https://files.community/download.

yaira2:
The performance issues in Git repos has been resolved in our dev branch and will be included in the next release.

yaira2:
Thumbnail performance is a little better in our dev branch, but we're still trying to fix an issue with loading a lot of thumbnails simultaneously in the Grid View layout.

yaira2:
> Thumbnail performance is better in the dev branch, but we're still trying to fix the issue with loading a lot of thumbnails simultaneously in the Grid View layout. In the meantime, turning off the option to show thumbnails helps as a workaround for this issue.

This issue has been fixed in the latest preview build and will be included in the next stable release.

yaira2:
Performance improvements for opening file pickers are included in #15386.

yaira2:
Thanks to @Tptogiar, improvements to loading large folders have been made in #15831.

yaira2:
A couple of noticeable improvements have been made to the address bar in #16665
[...]

Issue #3163, opened 2021-01-19
FilesFullTrust process crashes for Access Violation Exception


hez2010:
**Describe the bug**
FilesFullTrust crashes for Access Violation Exception while loading large icons for some files.

**To Reproduce**
1. Switch to Grid layout
2. Open `C:\Windows\System32` and scroll down
3. FilesFullTrust process crashes

**Expected behavior**
No crashes in FilesFullTrust

**Screenshots**
![image](https://user-images.githubusercontent.com/14960345/104977924-adc6f680-5a3b-11eb-9b89-c132845de1ef.png)

**Desktop (please complete the following information):**
 - OS Version: Windows 10 21292
 - App version: main

**Log file**
Unfortunately, this exception wasn't logged into log file.

**Additional Information**
The crashes on FilesFullTrust won't impact Files, so there's has minimal impact on user experience.

PR #4199, opened 2021-03-21
Improve folder enumeration speed

### LINKED ISSUE #4028: Files shows temporary files created when opening Word or AutoCad even if they're hidden files
Hello,

I think it would be great if after opening the actual file additional file(s) would not appear in the folder. For instance, when I open dwg file, additional two files open at the same time, see attached.

Kind regards,
[Screen Shot 2021-13-3 at 11-48-32.pdf](https://github.com/files-community/Files/files/6134721/Screen.Shot.2021-13-3.at.11-48-32.pdf)

Jan [Krulak[](url)](url)

### LINKED ISSUE #4037: Navigate to another folder very slow: 1-4 seconds
**Describe the bug**
Performed a fresh install of version 1.3

**To Reproduce**
1. Navigate to any folder (local disk). This can be down a level or up a level.
2. The blue progress indicator starts "swiping". Sometimes less than once. Often twice. Sometimes more than twice.
3. Then the folder is displayed
4. A little after, the progress indicator stops swiping

**Expected behavior**
Almost immediate display of folder, similar to Windows explorer. And the swipe indicator stops as soon as the folder is displayed. Because the progress indicator keeps moving even after the folder display, it seems even slower then it actually is.

 - OS Version: Windows 10 Version 2004 19041.746
 - App version: v1.3.0.0

**Additional context**
Same behavior after restarting the app

[debug.txt](https://github.com/files-community/Files/files/6136224/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/6136225/debug_fulltrust.txt)

gave92:
**Resolved / Related Issues**

Fixes #4028

**Details of Changes**

This PR:
- ~Removes the caching feature which appears to slow _down_ loading for small, local folders~ -> restored
- Temporary hidden files are not shown unless "show hidden items" is turned on (#4028)
- Loads items faster when navigating back/forward
- Hides the loading indicator as soon as the items from cache have been loaded -> @yair100 could you test this and see if you like it?


yaichenbaum:
@gave92 I'll test the changes.

gave92:
Thanks @yair100 I'm curious to know if it's just my mind playing tricks xD

yaichenbaum:
@gave92 I wouldn't say it's as fast as File Explorer, but it does seem to be faster than before.

gave92:
@yair100 thanks, I feared the "maybe a bit faster" answer, probably going to close this xD

Edit: did you have any caching feature turned on before testing this PR?

yaichenbaum:
@gave92 I thought this PR removes the cache option. Either way, it's a nice improvement and I think it's worth it to continue.

gave92:
@yair100 Yep, funniest thing is that removing the caching feature makes things faster :) I'm going to leave this open to gather more feedback. If confirmed it would be a nice simplification of the code.

yaichenbaum:
@gave92 It definitely makes things faster, and I can see items mush faster than before.

yaichenbaum:
@jakoss Can you test these changess?

yair100 (dismissed):

XPoppyX:
Does the context menu work for you in empty space?

gave92:
@XPoppyX could you try again with the latest "Small fixes" commit?

XPoppyX:
it works now 👍

gave92:
@XPoppyX thanks, I'll be interested in knowing how's loading speed for you (small/large/network folders) with this PR.

d2dyno1:
For #4180

XPoppyX:
I notice better loading speed.

A couple of problems that I noticed is that after opening different folders, eventually the shortcuts and the extra options of the contextual menu are no longer shown

gave92:
@XPoppyX I would not expect the issue to be tied to this PR, is this something that does not happen on main? Could you post the `debug_fulltrust.txt`?

XPoppyX:
I am trying to reproduce the issue in `main` but it works fine. With this PR it takes me only a few seconds to reproduce it

There is nothing useful in `debug_fulltrust.txt`, no problems, but there is something in `debug.txt`
[debug.txt](https://github.com/files-community/Files/files/6177894/debug.txt)

XPoppyX:
Ok apparently this happens after opening a folder on my external drive, which contains 22 folders, all with custom icons. I will test it in `main`

gave92:
@XPoppyX my best guess is that you're getting this annoying issue -> #3163. If you find this happening on main and you manage to reproduce it consistently, I'm not above asking you to send me those folders/icons so I can reproduce the issue (and hopefully fix it) xD

XPoppyX:
After testing in `main`, it is producible (with different steps) only after disabling cache.

I can't reproduce the issue with 82 empty folders with custom icons, maybe it depends on the number of items they contain, 
in total the folder in which the problem occurs, contains +28,000 files (From 100 to 8000 items per subfolder)

gave92:
> After testing in main, it is producible (with different steps) only after disabling cache.

Since this PR effectively disables the cache, I'm convinced the issue exists on main as well.

jakoss:
It does seems to make the loading faster, the question here is at what point the cache became the choking point? Because all cache-related code shouldn't make that of a difference, even the database operations were at 50ms tops

gave92:
@jakoss I'm not sure if I'm right about this but I'm suspecting that the real issue here is the usage of async/await. Basically we have a lot of context switches between UI and background thread that add variable delays.

jakoss:
If a lot of local context switches happens (for example, i remember the "ExtendedProperties" are implemented to work independent to each other) then this might be the case. I think some usage of `ConfigureAwait(false)` could help avoid some of those, but it seems like another hack. The real solution should be to limit synchronization with UI to the minimum

gave92:
Keeping this draft as removing the cache feature feels like killing a fly with a sledgehammer. Let me know if you think this should be merged as-is or if it makes more sense to look for other ways to improve the current code e.g reduce context-switches, disable cache for local folders or anything.

jakoss:
Well, I see no value in holding this. If the overall experience is better without caching - remove it. Adjustments in context synchronization might be a huge architectural change, so caching (if necessary) would have likely to be changed too

Jaiganeshkumaran:
Maybe instead of removing the feature, why not enable it only for folders with a specific number of items or more?

yaichenbaum:
> Maybe instead of removing the feature, why not enable it only for folders with a specific number of items or more?

I think that's what we should do, the cache feature seems to make a difference for directories with around 6,000 items.

gave92:
Unfortunately you have to access the cache in order to know how many items a folder has :)

yaichenbaum:
@gave92 Let's continue to work on the code.

gave92:
@yair100 good for me. I'm gonna keep this to separate the other changes I ended up putting in here.

yaichenbaum:
@gave92 It looks like scrolling is a little choppy when items are still loading.

gave92:
> Overall, though, there seems to be a massive increase in performance

_Perceived_ increase, thing is currently for folders with less than 32 elements, it wastes time loading from the cache but never adds the items to the UI.

> It looks like scrolling is a little choppy when items are still loading.

Yup, it shows cached items immediately but then it's busy loading the actual folder.

> It also looks like it doesn't always work to change the layout mode

Could you clarify this?

yaichenbaum:
> Could you clarify this?

Clicking on the different layout options in the display flyout don't always switch the layout mode.

gave92:
@yair100 mostly tried in details view only (they use a ListView underneath I think, not DataGrid which may explain some of the difference?)

yair100:
> @yair100 mostly tried in details view only (they use a ListView underneath I think, not DataGrid which may explain some of the difference?)

That's why I asked, I figured the DataGrid might be part of the slowness. Regarding the cache, if we can improve the speed, we can definitely remove the cache altogether.

yair100 (approved):

winston-de:
I noticed this significantly decreases the loading speed of shortcuts, I'm assuming because now they have to be obtained through the FTP every time. This may not be feasible, but perhaps just shortcuts could be cached?

gave92:
@winston-de at the end we _did not_ remove the caching feature so it shouldn't take longer to load folder contents.

winston-de:
@gave92 It looks like the json deserializer was breaking when trying to deserialize shortcut items, which slowed down the loading speed. Oddly enough, it seems to have stopped, and the loading speed is just as fast as normal folders. 

Here was the error that was causing it:
```
2021-03-25 21:17:46.7340|WARN|.ctor|Unable to find a constructor to use for type Files.Filesystem.ShortcutItem. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute. Path 'FileList[0].TargetPath', line 1, position 74.|Newtonsoft.Json.JsonSerializationException: Unable to find a constructor to use for type Files.Filesystem.ShortcutItem. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute. Path 'FileList[0].TargetPath', line 1, position 74.
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateNewObject(JsonReader, JsonObjectContract, JsonProperty, JsonProperty, String, Boolean&) + 0x17d
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader, Type, JsonContract, JsonProperty, JsonContainerContract, JsonProperty, Object) + 0x2f8
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader, Type, JsonContract, JsonProperty, JsonContainerContract, JsonProperty, Object) + 0xb6
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateList(IList, JsonReader, JsonArrayContract, JsonProperty, String) + 0x3d5
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateList(JsonReader, Type, JsonContract, JsonProperty, Object, String) + 0x110
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader, Type, JsonContract, JsonProperty, JsonContainerContract, JsonProperty, Object) + 0xef
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.SetPropertyValue(JsonProperty, JsonConverter, JsonContainerContract, JsonProperty, JsonReader, Object) + 0x16d
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object, JsonReader, JsonObjectContract, JsonProperty, String) + 0x6b2
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader, Type, JsonContract, JsonProperty, JsonContainerContract, JsonProperty, Object) + 0x321
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader, Type, JsonContract, JsonProperty, JsonContainerContract, JsonProperty, Object) + 0xb6
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader, Type, Boolean) + 0x22d
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader, Type) + 0xf3
   at Newtonsoft.Json.JsonConvert.DeserializeObject(String, Type, JsonSerializerSettings) + 0xd6
   at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String, JsonSerializerSettings) + 0x36
   at Files.Helpers.FileListCache.PersistentSQLiteCacheAdapter.<ReadFileListFromCache>d__3.MoveNext() + 0x2e6
```
However, this does not seem to be an issue with this pr, as my log reports the same error going back to at least February. I'll file a new issue for this.

gave92:
#4283 is supposed to have fixed that issue.

winston-de:
Hmmmm I'm still getting the error. I'll try clearing the cache, maybe that'll fix it.

PR #4209, opened 2021-03-21
Move horizontal TabView outside of PaneHolderPage

### LINKED ISSUE #4144: Enhancement: Enable dragging Files from top of sidebar
Currently, the drag area is limited to the right of the tabs, however, users can't drag the window from the top of the sidebar. This can be confusing, since users expect to be able to drag from any blank spot on the title bar.

Extend the drag over the sidebar.

lukeblevins:
**Resolved / Related Issues**

- e.g. closes https://github.com/files-community/Files/issues/4144

**Details of Changes**

Prevent duplicate TabView instances on each page by moving the TabView to MainPage. This also includes an expanded drag region allowing users to drag the app window from the sidebar while keeping the right inset drag region which is now set a more reliable way. Finally, this PR adds back the removed TabViewItem transitions.

**Validation**

- [x] Built and ran the app
- [x] Tested for regressions in right inset region.

**Screenshots (optional)**
image
Design remains the same and still allows for sidebar resizing


duke7553:
I can confirm a *slight* reduction in memory usage after this PR. For example, 10 open tabs went from 170 MB to 150 MB. The important part is that this PR fixes weird, user-facing behavior associated with having different instances of the TabView at once.

d2dyno1:
For #4180

yaichenbaum:
image
The tab strip is shifted when the multitasking setting is set to horizontal and the window is sized small.

lukeblevins:
@yair100  This should be fixed now. I suggest merging this one despite the occasional weird behavior with the sidebar because it potentially fixes a crash during resize.

yaichenbaum:
@duke7553 What's the occasional weird behavior with the sidebar that you're talking about?

lukeblevins:
@yair100 The sidebar may enter a state where there is a gap between it and the content when it is in compact mode. This behavior is relatively rare and only occurs because the property change can't be communicated to the control on every open tab when their sidebar instances are unloaded due to the tab being unselected. This only occurs when multiple tabs are open and is usually easy to get back to normal by resizing the app window/toggling the sidebar open and closed.

yaichenbaum:
@duke7553 I'll check that out when I review this PR. We can probably resolve that and reduce memory usage by moving the sidebar out of PaneHolderPage as well.

yair100 (approved):
Great work! LGTM

gave92:
Uff this crashes the app when toggling the "Enable dual pane mode" option in settings (because `Window.Current.Content` is not a MainPage when in Settings). Is there a better way to get a reference to the sidebar view model?

yaichenbaum:
> Uff this crashes the app when toggling the "Enable dual pane mode" option in settings (because `Window.Current.Content` is not a MainPage when in Settings). Is there a better way to get a reference to the sidebar view model? If not we need to add some null checks.

@gave92 Is that related to this issue https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/2706044090u/overview?

gave92:
@yair100 yeah but @duke7553 has a fix in his other PR.

PR #4227, opened 2021-03-22
Fix selection options

### LINKED ISSUE #4211: Some menu items and options in the navigation toolbar don't work
**Explain the bug**
Write a clear description of what the bug is.

**To reproduce**
1. Go to any folder
2. Open selection flyout, click on select all
3. Nothing happens
4. Repeat for all other options in flyout with same result

**Expected behavior**
The flyout items should perform their tasks.

**Logfile**
Nothing relevant. 

**Screenshots**

https://user-images.githubusercontent.com/59544401/111933877-8ca37600-8a7d-11eb-8e93-198d3d23c2e5.mp4

**System Information**
 - OS Version: 21337.1010
 - App version: 1.4.47.0

**Additional comment**
Restart does not fix.
The keyboard shortcuts, however, do still work.

d2dyno1:
**Resolved / Related Issues**

- Closes #4211

**Details of Changes**

Fixes selection commands binding. Can you confirm that everything is working correctly after this change? @winston-de

**Validation**

- [x] Built and ran the app

### DIFF
--- a/Files/Views/ModernShellPage.xaml
+++ b/Files/Views/ModernShellPage.xaml
@@ -248,9 +248,9 @@
                 x:FieldModifier="public"
                 
                 MultiselectEnabled="{x:Bind InteractionViewModel.MultiselectEnabled, Mode=TwoWay}"
-                SelectAllInvokedCommand="{x:Bind SelectAllContentPageItemsCommand}"
-                InvertSelectionInvokedCommand="{x:Bind InvertContentPageSelctionCommand}"
-                ClearSelectionInvokedCommand="{x:Bind ClearContentPageSelectionCommand}"
+                SelectAllInvokedCommand="{x:Bind SelectAllContentPageItemsCommand, Mode=OneWay}"
+                InvertSelectionInvokedCommand="{x:Bind InvertContentPageSelctionCommand, Mode=OneWay}"
+                ClearSelectionInvokedCommand="{x:Bind ClearContentPageSelectionCommand, Mode=OneWay}"
                 
                 LayoutModeInformation="{x:Bind FolderSettings.LayoutModeInformation, Mode=OneWay}"
                 ToggleLayoutModeDetailsView="{x:Bind FolderSettings.ToggleLayoutModeDetailsView}"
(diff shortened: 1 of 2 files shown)


winston-de:
I'm testing it now. You may have seen it mentioned earlier, but new tab, new window, and new pane in the ... menu are also broken.

yaichenbaum (approved):

Files/Views/ModernShellPage.xaml
```diff
@@ -248,9 +248,9 @@
                 x:FieldModifier="public"
                 
                 MultiselectEnabled="{x:Bind InteractionViewModel.MultiselectEnabled, Mode=TwoWay}"
-                SelectAllInvokedCommand="{x:Bind SelectAllContentPageItemsCommand}"
-                InvertSelectionInvokedCommand="{x:Bind InvertContentPageSelctionCommand}"
-                ClearSelectionInvokedCommand="{x:Bind ClearContentPageSelectionCommand}"
+                SelectAllInvokedCommand="{x:Bind SelectAllContentPageItemsCommand, Mode=OneWay}"
```
gave92, line 251: Curious: is this change necessary given the other changes e.g. `new RelayCommand(() => SlimContentPage?.SelectAllItems());`?

yaichenbaum: I think the default is set to `OneTime`.

d2dyno1: I've also noticed these were set somewhere else, I've added `Mode=OneWay` because it was like that before Interaction removal and to unify with rest of those. @gave92

gave92: Thanks @d2dyno1 that's fine by me.
@yair100 I asked because given the changes to ModernShellPage.xaml.cs I think OneTime should be ok here.

d2dyno1: @gave92 Not so sure about `OneTime` here because it'd be set to null on home page and then it wouldn't update upon navigation

gave92: I'd bet that when you do this, the "WorkingDirectory" value is evaluted each time the command is called, so it would update on navigation, but I'm actually not sure :)
```cs
public ICommand Cmd = new RelayCommand(() => NavigationHelpers.OpenDirectoryInTerminal(this.FilesystemViewModel.WorkingDirectory, this));
```

PR #4313, opened 2021-03-26
Move SidebarControl to MainPage

duke7553:
**Resolved / Related Issues**
Itemize resolved / related issues by this PR.
- Closes #4299
- Closes #4288 
- Closes #4336

**Details of Changes**
Moved sidebar to MainPage. Further rationale is being discussed in the ticket linked above, but I'd appreciate code review to find any MVVM-incompliance.

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [x] Tested the changes for accessibility
- [x] Tested many scenarios including inset region, vertical tabs, panes, and selection across tabs

Design remains the same.

PR #4325, opened 2021-03-27
Fix crash when toggling dual pane setting

gave92:
**Resolved / Related Issues**
- Fixes [2706044090u](https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/2706044090u/overview)

**Details of Changes**
Add details of changes here.
- Fixes an issue that causes app crash when toggling the "Enable dual-pane mode" option

**Validation**
- [x] Built and ran the app


duke7553:
@gave92 Good catch. I also fixed this a different way in #4313 . Not sure which is better.

gave92:
@duke7553 sorry I haven't noticed, happy to see the `Window.Current.Content as MainPage` is gone from PaneHolderPage in #4313 :)

PR #4365, opened 2021-03-29
Added a setting to always open folders in a new tab

### LINKED ISSUE #2129: Add a setting to always open folders in a new tab
Add an option to open folders in a new tab instead of opening in the existing tab.

oleitao:
**Details of Changes**
Add details of changes here.
- Added on settings Preferencies panel an option to set folder click open on new tab.

Closes #2129

**Screenshots (optional)**
Capture


Files/Views/SettingsPages/Preferences.xaml
```diff
@@ -114,6 +114,13 @@
                     Header="Show Library section on the sidebar"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{Binding ShowLibrarySection, Mode=TwoWay}" />
+
+                <ToggleSwitch
+                    x:Name="OpenFoldersNewTab"
+                    x:Uid="SettingsPreferencesOpenFoldersNewTab"
+                    Header="Open Folders on new tab"
```
yaichenbaum, line 121:
```suggestion
                    x:Uid="SettingsPreferencesOpenFoldersNewTab"
                    Header="Open folders in new tab"
```

Can you also add this string to the en-US resource file?

yair100 (changes requested):

Files/Views/SettingsPages/Preferences.xaml
d2dyno1: `Always open folders in new tab`

yaichenbaum: @d2dyno1 Is that accurate? There is still the option to open folders in a new window and I don't want it to sound like that option will be ignored.

d2dyno1: I just wanted to suggest different name for it

```suggestion
                    x:Uid="SettingsPreferencesOpenFoldersNewTab"
                    Header="Always open folders in new tab"
```

yaichenbaum: @d2dyno1 Using a different name is fine, I just don't think the one you suggested fits well in this case.

yaichenbaum:
@oleitao To prevent these merge conflicts in the future, a good trick is to revert changes to all the resource files (with the exception of the en-US resource file). Doing this will make it much easier to resolve conflicts since you only need to fix one file.

yaichenbaum:
@oleitao can you solve the merge conflicts?

oleitao:
> @oleitao can you solve the merge conflicts?

@yair100  I can, but i don't have access to do that

yaichenbaum:
@oleitao Do you have the GitHub desktop app? It is fairly simple to solve the conflicts using the UI when you merge the `main` branch into your dev branch.

d2dyno1:
@oleitao in Files repo, open cmd:

```
git remote add upstream https://github.com/files-community/Files.git
```

```
git fetch upstream
```

```
git merge upstream/main
```

winston-de (changes requested):
@oleitao You seem to have not created the OpenFoldersNewTab property in ```PreferencesViewModel.cs```, just the field, so the bindings on the toggle switch aren't working.

image


You need to add something like this:
```cs
  public bool OpenFoldersNewTab
  {
      get => openFoldersNewTab;

      set
      {
          if (SetProperty(ref openFoldersNewTab, value))
          {
              App.AppSettings.OpenFoldersNewTab = value;
          }
      }
  }
```

winston-de:
@oleitao It looks like clicking folders when in grid or tiles view still opens them in the same tab.

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -642,7 +642,15 @@ private void RadioMenuSortDirection_Click(object sender, RoutedEventArgs e)
...
-            NavigationHelpers.OpenSelectedItems(ParentShellPageInstance, false);
+            
+            if (AppSettings.OpenFoldersNewTab)
+            {
+                NavigationHelpers.OpenPathInNewTab(((e.OriginalSource as FrameworkElement)?.DataContext as ListedItem).ItemPath);
+            }
+            else
+            {
+                NavigationHelpers.OpenSelectedItems(ParentShellPageInstance, false);
+            }
```
yaichenbaum, line 653: This is only going to help when double clicking on an item, it's not going to help when pressing the enter key on an item.

Files/Strings/en-US/Resources.resw
```diff
@@ -2067,6 +2067,12 @@
   <data name="PropertyISO" xml:space="preserve">
     <value>ISO speed</value>
   </data>
+  <data name="SettingsPreferencesOpenFoldersNewTab.Header" xml:space="preserve">
+    <value>Open folders in new tab</value>
+  </data>
+  <data name="NavToolbarShowBundlesWidget" xml:space="preserve">
```
winston-de, line 2073: This resource is a duplicate, and I believe it is what is causing the build error.

Files/Views/LayoutModes/GenericFileBrowser.xaml.cs
```diff
@@ -543,7 +543,14 @@ private void AllView_PreviewKeyDown(object sender, KeyRoutedEventArgs e)
                 }
                 else
                 {
-                    NavigationHelpers.OpenSelectedItems(ParentShellPageInstance, false);
+                    if (AppSettings.OpenFoldersNewTab)
```
winston-de, line 546: This feature still only works in the details view browser. It might be easier to move this check to ```NavigationHelpers.OpenSelectedItems``` so that it is opened in a new tab regardless of the layout.

winston-de (approved):
LGTM, nice work!

Issue #4402, opened 2021-03-31
Problem to switch the language


LuanRoger:
**Explain the bug**
In the release version (the one that is on the Microsoft Store) the Portuguese (Brazil) language is not being applied, being applied Portuguese (Portugual). Which makes me confused because in the development version, Portuguese (Portugal) is neither listed, nor does it have a language file

**To reproduce**
In the release version:
1. Go to Settings
2. Go to Preferences
3. Switch between Portuguese (Portugal) and Portuguese (Brazil)

**Expected behavior**
Strings update according to the selected language

## Screenshots
**Release Version:**
![ListLanguagesLaunch](https://user-images.githubusercontent.com/46284513/113214848-6dac9d00-9250-11eb-96c9-b3979e083b65.png)
**Dev Version:**
![ListLanguagesDev](https://user-images.githubusercontent.com/46284513/113214933-8ae16b80-9250-11eb-8a63-aac0f5a9ce84.png)
**Wrong (Release):**
![WrongString](https://user-images.githubusercontent.com/46284513/113215906-f37d1800-9251-11eb-9973-4f2f500fc65c.png)
**Right (Dev):**
![CorrectString](https://user-images.githubusercontent.com/46284513/113215956-07c11500-9252-11eb-92b6-7cd828af4883.png)

**System Information**
 - OS Version: Windows 10 20H2
 - App version: Dev: v1.3.1.0. Release: v1.5.0.0

LuanRoger:
Fixed in #4416

PR #4416, opened 2021-04-01
Update translation: pt-BR, pt-PT

LuanRoger:
**Details of Changes**
- Updated pt-BR, pt-PT language

**Validation**
- [x] Built and ran the app.


tsvietOK (changes requested):
Looks like you modified `Resources.resw` file manually. All changes should be done only in `Files.pt-BR.xlf`. In PR include only .xlf file. For more details see our [documentation](https://files-community.github.io/docs/#/articles/translating-files).

LuanRoger:
@tsvietOK I didn't modify the file manually, but I think I know why it is as modified:
I created an issue (#4402) explaining that the language was not changing from PT-PT to PT-BR, but it is really changing the language but some strings are the same, which was not meant to be.
When I did the merge of the upstream for my fork I chose not to bring these two files because I had already worked with them, so the only thing that should have been modified would be Files.pt-BR.xlf, but as I explained earlier this does not happen.

tsvietOK:
I recommend you to exclude Resources file from this PR, since it will be regenerated later.

tsvietOK:
I guess the problem is here. (target-language)
https://github.com/files-community/Files/blob/cd2f0355bf8bfabe18c31647ca49a0d5cff79763/Files/MultilingualResources/Files.pt-PT.xlf#L3

Files/MultilingualResources/Files.pt-BR.xlf
```diff
@@ -1184,12 +1191,12 @@
           <target state="translated">Ver perfis de amostra</target>
         </trans-unit>
         <trans-unit id="SettingsPreferencesAppLanguage.Header" translate="yes" xml:space="preserve">
-          <source>App language</source>
-          <target state="translated">Idioma do aplicativo</target>
+          <source>App Language</source>
```
tsvietOK, line 1194: By the way, don't change string in `source` tag, the value is taken from Strings/en-US/Resources.resw file. Just change translation in `target` tag.

tsvietOK (approved):

Issue #4369, opened 2021-03-30
Text overlapping on address bar.


Tropix126:
**Explain the bug**
When the search bar is focused, opening the "More options" menu will cause a strange visual bug where the search text overlaps the address bar.

**To reproduce**
1. Click the search button to trigger search.
2. With the bar still focused, click on the 3 dots to open the "More options" menu.

**Expected behavior**
Search menu should close upon the menu being opened.

**Logfile**
not relevant

**Screenshots**

https://user-images.githubusercontent.com/42101043/112923626-37e2a980-90d4-11eb-834b-55b6096d2977.mp4

**System Information**
 - OS Version: Windows 10 Dev (21332.1000)
 - App version: 1.5.0.0

**Additional comment**
None.

cinqmilleans:
Hello, I can fix it if the team assigns me this bug.

yaichenbaum:
@Tropix126 This issue should be resolved with #4649.

PR #4424, opened 2021-04-02
Fixed Text overlapping on address bar (#4369)

cinqmilleans:
**Explain the bug**
When the search bar is focused, opening the "More options" menu will cause a strange visual bug where the search text overlaps the address bar.

**To reproduce**
- Click the search button to trigger search.
- With the bar still focused, click on the 3 dots to open the "More options" menu.

#4369

### DIFF
--- a/Files/UserControls/NavigationToolbar.xaml
+++ b/Files/UserControls/NavigationToolbar.xaml
@@ -841,6 +841,8 @@
                 PlaceholderText="Search"
                 QuerySubmitted="SearchRegion_QuerySubmitted"
                 SuggestionChosen="SearchRegion_SuggestionChosen"
+                Background="{ThemeResource SolidBackgroundFillColorQuarternary}"
+                BorderBrush="{ThemeResource SystemBaseMediumLowColor}"
                 TextBoxStyle="{StaticResource AutoSuggestBoxTextBoxStyleFixed}"
                 TextChanged="SearchRegion_TextChanged"
                 UpdateTextOnSelect="False"


Files/UserControls/NavigationToolbar.xaml
```diff
@@ -841,6 +841,8 @@
                 PlaceholderText="Search"
                 QuerySubmitted="SearchRegion_QuerySubmitted"
                 SuggestionChosen="SearchRegion_SuggestionChosen"
+                Background="{ThemeResource SolidBackgroundFillColorQuarternary}"
+                BorderBrush="{ThemeResource SystemBaseMediumLowColor}"
```
yaichenbaum, line 845: Should this use the accent color?

cinqmilleans: The background appears transparent and a text is visible below. With this change, the search textbox has the same opaque background as the path textbox. The textbox below uses this background.

yaichenbaum: I was talking about the border color.

yaichenbaum: @cinqmilleans Instead of changing the color, I think we should just close the search box when it loses focus.

cinqmilleans: I will test this idea.

cinqmilleans: In event SearchRegion_LostFocus, there is a rule to only do when the focused element is a popup. Removing the rule fixes the problem. I don't know what this rule is for, but it must be useful. The event is not triggered by the search results popup.

yaichenbaum: @duke7553 Do you have any idea what the rule is being used for?

cinqmilleans: Maybe this was done for the results popup, but the event is not triggered by this popup.

yaichenbaum: It could be it's to prevent losing focus when right clicking on the search box.

cinqmilleans: This is it, I have tested. I will modify the test to only filter this popup.

yaichenbaum: Thanks!

cinqmilleans: It doesn't seem to be a problem to remove the test on the popup. No popup benefits. The search bar menu is filtered by another test.

There is a similar problem in dark mode, but it doesn't seem to have the same cause. The TextBoxStyle does not seem to work and therefore the box has no background. I searched for a long time without finding why.

yaichenbaum: It turns out that the styles for dark mode were recently updated in WinUI to be semi transparent, the best solution would probably be to hardcode the background.

cinqmilleans:
I will open a new PR with the correction. I close.

PR #4431, opened 2021-04-02
Preview pane improvements & fixes

### LINKED ISSUE #4444: Image resolution in preview pane has too many decimal places
When selecting an image and opening the Preview pane, the image resolution (both horizontal and vertical) shows a many many decimal places that overflow the pane and get cut off.

**Describe the solution**
These numbers should be rounded after a certain number of significant figures.

You could make it scrollable, but I don't see why you would do that when the simple solution is to round to a certain number of digits.

**Additional comment**
App version 1.5.0.0
Screenshot 2021-04-03 222901

winston-de:
**Resolved / Related Issues**
- Fixes an issue I noticed where empty files would be loaded as text files, even if they were not text files. 
- Moves the date taken property down with the date modified and date created property.
- Displays the exposure property as a fraction
- Fix alignment of gripper icon in horizontal layout
- Show exposure time as fraction instead of as a decimal 
- Closes #4444

**Details of Changes**
- Don't run ```TryLoadAsText``` on empty files, as it will always think the file is a text file, instead just show the thumbnail.
- Move "date taken" below date modified and date created
- Show exposure time as a fraction instead of as a decimal, and add helper extensions to convert doubles to a fraction (the fractions are inaccurate, I think the storage file properties API is giving inaccurate values, for example I keep seeing 59 fps when it should be 60)
- Added hours to duration prop format
- Fix gripper alignment being off when in horizontal layout
- Round image resolution

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Empty pptx files before and after: 
image
image

Duration format:
image

### DIFF
--- a/Files/Strings/en-US/Resources.resw
+++ b/Files/Strings/en-US/Resources.resw
@@ -2035,6 +2035,6 @@
     <value>Open Storage Sense</value>
   </data>
   <data name="PropertyISO" xml:space="preserve">
-    <value>ISO</value>
+    <value>ISO speed</value>
   </data>
 </root>
--- a/Files/Strings/en-US/Resources.resw
+++ b/Files/Strings/en-US/Resources.resw
@@ -2034,4 +2034,7 @@
   <data name="DrivesWidgetOpenStorageSenseButton.AutomationProperties.Name" xml:space="preserve">
     <value>Open Storage Sense</value>
   </data>
+  <data name="PropertyISO" xml:space="preserve">
+    <value>ISO</value>
+  </data>
 </root>
--- a/Files/ViewModels/Properties/FileProperty.cs
+++ b/Files/ViewModels/Properties/FileProperty.cs
@@ -288,6 +288,7 @@ namespace Files.ViewModels.Properties
             { "Fraction" , input => ((double)input).ToFractions(2000)},
             { "AddF" , input => $"f/{(double)input}"},
             { "AddISO" , input => $"ISO-{(UInt16)input}"},
+            { "RoundDouble" , input => $"{Math.Round((double)input)}"},
         };
     }
 }
--- a/Files/ViewModels/Properties/FileProperty.cs
+++ b/Files/ViewModels/Properties/FileProperty.cs
@@ -1,4 +1,5 @@
 ﻿using Files.Converters;
+using Files.Helpers;
 using Microsoft.Toolkit.Mvvm.ComponentModel;
 using Microsoft.Toolkit.Uwp;
 using Newtonsoft.Json;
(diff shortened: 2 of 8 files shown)


tsvietOK:
Seems like it is not working with 0 byte bmp files.

tsvietOK:
image
Aperture should not be rounded. By the way, it should start with f/

winston-de:
> Seems like it is not working with 0 byte bmp files. Should it work?

@tsvietOK 
It should not. My changes for empty files only effect ```TryLoadAsTextAsync```, which isn't called unless there isn't a preview control available for that extension (which there is for bmp files).

winston-de:
> Aperture should not be rounded. By the way, it should start with f/

@tsvietOK Fixed
image

R3voA3:
Was the exposure time fixed as well?

R3voA3:
> image
> Aperture should not be rounded. By the way, it should start with f/

Adding to that: Focal lenght should be:
`55 mm`

R3voA3:
How does Files know what properties to display? Currently Files is for example missing a lot of them for .raw files 

image

winston-de:
@R3voA3 There's a list. If there are any you specifically would like added, please tell me.

tsvietOK:
It would be good if you add ISO value. Format `ISO-XXX`, like ISO-100, ISO-1600.

winston-de:
image

Files/Strings/en-US/Resources.resw
```diff
@@ -2034,4 +2034,7 @@
   <data name="DrivesWidgetOpenStorageSenseButton.AutomationProperties.Name" xml:space="preserve">
     <value>Open Storage Sense</value>
   </data>
+  <data name="PropertyISO" xml:space="preserve">
+    <value>ISO</value>
```
tsvietOK, line 2038:
```suggestion
    <value>ISO speed</value>
```

tsvietOK (approved):

tsvietOK (approved):

Files/ViewModels/Properties/FileProperty.cs
```diff
@@ -283,7 +284,11 @@ public async static Task<List<FileProperty>> RetrieveAndInitializePropertiesAsyn
         private static readonly Dictionary<string, Func<object, string>> DisplayFuncs = new Dictionary<string, Func<object, string>>()
         {
             { "DivideBy1000", input => (((uint) input)/1000).ToString() },
-            { "FormatDuration", input => new TimeSpan(Convert.ToInt64(input)).ToString("mm':'ss")},
+            { "FormatDuration", input => new TimeSpan(Convert.ToInt64(input)).ToString("hh':'mm':'ss")},
+            { "Fraction" , input => ((double)input).ToFractions(2000)},
+            { "AddF" , input => $"f/{(double)input}"},
+            { "AddISO" , input => $"ISO-{(UInt16)input}"},
+            { "RoundDouble" , input => $"{Math.Round((double)input)}"},
```
R3voA3, line 291: Focal Length property needs "mm" added.

Example:

Focal Length:
300 mm

R3voA3 (approved):

PR #4436, opened 2021-04-03
Added empty homepage information

d2dyno1:
**Details of Changes**
Adds a nice easter egg to the Homepage and hyperlink to settings when no widgets are enabled.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

### DIFF
--- a/Files/Views/YourHome.xaml.cs
+++ b/Files/Views/YourHome.xaml.cs
@@ -1,6 +1,7 @@
 ﻿using Files.Dialogs;
 using Files.Filesystem;
 using Files.Helpers;
+using Files.SettingsPages;
 using Files.UserControls.Widgets;
 using Files.ViewModels;
 using Files.ViewModels.Bundles;
@@ -10,6 +11,7 @@ using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
 using Windows.ApplicationModel.AppService;
+using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Navigation;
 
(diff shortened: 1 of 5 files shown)


d2dyno1:
Ready!

Files/Views/YourHome.xaml.cs
```diff
@@ -1,35 +1,63 @@
...
+        }
+        
+        public string HomePageEmptyPageSubtitleP1Text
+        {
+            get => "HomePageEmptyPageSubtitleP1".GetLocalized().Replace(' ', (char)160) + "\ufeff";
+        }
+        
+        public string HomePageEmptyPageSubtitleP2Text
+        {
+            get => "HomePageEmptyPageSubtitleP2".GetLocalized().Replace(' ', (char)160) + "\ufeff";
```
d2dyno1, line 44: This had to be done, because xaml textblocks cut off last character(s) if it's space ` `

d2dyno1: No, `xml:preserve="space"` doesn't work.

tsvietOK: What is the purpose of space at the end of line?

d2dyno1: @tsvietOK the ` ` is for the space between text and hyperlink.

HomePageEmptyPageSubtitleP1Text + HyperlinkButton + HomePageEmptyPageSubtitleP2Text

yaichenbaum:
These changes haven't been approved for development, these changes are ok, but they shouldn't be done until users can adjust the widgets from the homepage themselves #4116.

yaichenbaum:
@zeealeid Any ideas to improve the design?

zeealeid:
I see, this screen is supposed to appear when there are no widgets in the home page right?

So maybe we could spice it up a bit by maybe adding an illustration and making the settings button a bit more pronounced, if that makes sense :) 

I'm not on my PC right now so I'll share a quick mock-up soon to explain it better.

yaichenbaum:
> I see, this screen is supposed to appear when there are no widgets in the home page right?
> 
> So maybe we could spice it up a bit by maybe adding an illustration and making the settings button a bit more pronounced, if that makes sense :)
> 
> I'm not on my PC right now so I'll share a quick mock-up soon to explain it better.

Since this PR, the option to choose which widgets are invisible has been move to the toolbar.

zeealeid:
Here's a rough mockup of what I meant :)

Illustration is from https://undraw.co/ (Empty Street) but I slightly edited the colors, maybe use a version of this illustration with brighter colors for light mode?

Frame 503

yaichenbaum:
@zeealeid It will need to be adjusted a little to work with the new flyout for managing widgets, but it looks really nice!

R3voA3:
Could also use a teaching tip control to point users to the widget settings in the toolbar.

image

yaichenbaum:
@d2dyno1 Can we get an update on this PR?

d2dyno1:
@yair100 I've stopped my work on it because I don't have any instructions on it. Moreover I don't have the image that Zee used.

yaichenbaum:
@d2dyno1 The design from @zeealeid is what we should aim for, the only difference is, that instead of opening the settings page, the button should open the widgets flyout.

d2dyno1:
:shipit:

d2dyno1:
I'll close it for now

Issue #2263, opened 2020-11-07
Ability to rearrange widgets


d2dyno1:
**Is your feature request related to a problem? Please describe.**
As more widgets come in the future it'd be nice if users had the ability to rearrange them

**Describe the solution you'd like**
Ability to rearrange widgets in the Widgets settings

yaichenbaum:
@d2dyno1 considering the fact that no additional widgets are on the roadmap, I think we can close this issue.

PR #4448, opened 2021-04-04
Overhaul for widgets codebase

d2dyno1:
**Resolved / Related Issues**
- For #2263, #4180

**Details of Changes**
- Widgets can now be "cached" meaning you don't have to load new ones when existing ones are loaded (It's not implemented though).
- Widgets will no longer be re-loaded when opening settings (from homepage) and going back.
- (It looks like) widgets loading performance also has improved :thinking: 
- It is now possible to implement reordering.
- Widgets are no longer limited to hardcoded controls. Any control that implements `IWidgetItemModel` or `ICustomWidgetItemModel` can be a widget now.
- IDisposable control - Dispose() will be called for all widgets on their removal
- @winston-de @yair100 This PR makes displaying custom widgets possible :tada:

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Views/YourHome.xaml.cs
+++ b/Files/Views/YourHome.xaml.cs
@@ -62,11 +62,14 @@ namespace Files.Views
                 libraryCards.LibraryCardPropertiesInvoked += LibraryWidget_LibraryCardPropertiesInvoked;
                 libraryCards.LibraryCardDeleteInvoked -= LibraryWidget_LibraryCardDeleteInvoked;
                 libraryCards.LibraryCardDeleteInvoked += LibraryWidget_LibraryCardDeleteInvoked;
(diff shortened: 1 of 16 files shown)


Files/Views/YourHome.xaml.cs
```diff
@@ -188,7 +201,7 @@ protected override async void OnNavigatedTo(NavigationEventArgs eventArgs)
         //       This IDisposable.Dispose() needs to be called to unhook events in BundlesWidget to avoid memory leaks.
         public void Dispose()
```
d2dyno1, line 202: Actually this Dispose() will never be called as the comment says.
I'll see what could be done in my next PR :)

d2dyno1:
That's... weird..

image

**Edit:** This has been fixed (Thanks @yair100)

yair100 (dismissed):
Great work!

yair100 (approved):

PR #4472, opened 2021-04-06
Dispose Homepage on navigation

d2dyno1:
**Resolved / Related Issues**
- For #4180 

**Details of Changes**
YourHome.Dispose() is now called on navigation. It is possible to modify the code for any page to be disposed on navigation however that's not the case here. The problem with disposing every page but YourHome is that any other page is being reused and in YourHome's case, the widgets are created on navigation.
This PR fixes the issue where widgets weren't being disposed on navigaion and in Bundles case, it fixes a memory leak.

**Validation**
- [x] Built and ran the app


lukeblevins (commented):
This is a bit overkill, considering we can just call Dispose() in YourHome's OnNavigatingFrom override. The code looks good, but I'm not sure if we need the FrameExtensions at all unless there's another reason for it.

d2dyno1:
@duke7553 we could call a function similar to `ICleanup.Clanup()` on other views.

yaichenbaum:
@duke7553 Should we merge this?

d2dyno1:
@yair100 could you review this one? It's fixing a memory leak :eyes:

cinqmilleans (changes requested):
@yaichenbaum @d2dyno1 "using File.Extensions" is unused and useless in 2 files. "WidgetPage.xaml.cs" is better.

lukeblevins (approved):
Clever enough solution to call the Dispose() method, so I suppose

PR #4474, opened 2021-04-06
Added the option to manage widgets from the homepage

d2dyno1:
**Resolved / Related Issues**
- For #4116

**Details of Changes**
It is now possible to manage widgets from the homepage + hybrid YourHome MVVM (Not full MVVM because widgets except for Bundles are not MVVM themselves).

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [x] Tested the changes for accessibility

**Screenshots (optional)**

https://user-images.githubusercontent.com/53011783/113751108-db067500-970b-11eb-836a-eb05aa42db4d.mp4

PR #4483, opened 2021-04-07
Remove IShellPage reference from Bundles

d2dyno1:
**Details of Changes**
This PR removes IShellPage instance from Bundles. There's a known bug where fulltrust process wouldn't be called to get shortcut file icons. However it looks like it was not caused by this PR nor #4474.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Files.csproj
+++ b/Files/Files.csproj
@@ -190,8 +190,8 @@
     <Compile Include="Enums\FilesystemOperationType.cs" />
     <Compile Include="Enums\FolderLayout.cs" />
     <Compile Include="Enums\FolderLayoutModes.cs" />
-    <Compile Include="EventArguments\BundlesLoadIconOverlayEventArgs.cs" />
-    <Compile Include="EventArguments\BundlesOpenPathEventArgs.cs" />
+    <Compile Include="EventArguments\Bundles\BundlesLoadIconOverlayEventArgs.cs" />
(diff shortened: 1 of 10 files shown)


Files/ViewModels/Widgets/Bundles/BundlesViewModel.cs
```diff
@@ -36,14 +38,24 @@ public class BundlesViewModel : ObservableObject, IDisposable
...
-        private IShellPage associatedInstance;
+        private bool isInitialized;
 
         private bool itemAddedInternally;
 
         private int internalCollectionCount;
 
         #endregion Private Members
 
+        #region Events
```
yaichenbaum, line 49: Please remove the regions in this class.

yaichenbaum (changes requested):

Files/Files.csproj
```diff
@@ -190,6 +190,8 @@
     <Compile Include="Enums\FilesystemOperationType.cs" />
     <Compile Include="Enums\FolderLayout.cs" />
     <Compile Include="Enums\FolderLayoutModes.cs" />
+    <Compile Include="EventArguments\BundlesLoadIconOverlayEventArgs.cs" />
+    <Compile Include="EventArguments\BundlesOpenPathEventArgs.cs" />
```
yaichenbaum, line 194: Should these be put in a sub folder for bundles?

d2dyno1: @yair100 all done!

Issue #4506, opened 2021-04-08
Opening new tab should keep you on current tab


meowish:
**Describe the solution**
Opening a new tab should not automatically move you to the new tab, and should keep you on the current one.

cinqmilleans:
I agree. I can take care of the community accept this change.

yaichenbaum:
@cinqmilleans That would be great! Let's continue tracking this in #2512.

cinqmilleans:
yaichenbaum, meowish and myself wished it in issue # 4506. Most software with tabs works this way, so it's more intuitive for the user. If we need the folder right away, we open it in the current tab. If it's for later, you open it in a pending tab. You can open multiple subfolders in tabs easily.

Issue #2512, opened 2020-12-02
Opening directories in a new tab should open them in the background


tgrelka:
**Is your feature request related to a problem? Please describe.**
I’m used to browsers opening a new tab with Middle Click in the background. I find it quite annoying to use the feature in Files UWP, since it opens a new tab *and* focuses it.

At first, I was confused and thought I had found a bug where it opens the folder in a new tab, but also in the current one until I noticed I had been focused on the new tab.
This is especially annoying when you want to open multiple folders (from a single parent) in new tabs for later use, like you may e.g. do with search results when you filter through them.

I’m confident I’m not alone since many of us spend a lot of time in a browser every day and this seems to be the default behaviour.

**Describe the solution you'd like**
It’s common in browsers to have Middle Click open a new tab (with the clicked link) in the background, while Shift+Middle Click opens a new tab and puts the focus on it. Files UWP should replicate this behaviour as it is more intuitive and in-line with other applications.

**Describe alternatives you've considered**
I hope I made it clear enough that I’ve thought about how it should work and that behaviour already known to the user seems better to me than change up what’s already established elsewhere. (This isn’t always the case but holds true this time.)

Olcay91:
I'd make more sense if you could decide if it should open in the background or switch to the new tab. As far as I know the standard is that it switches to the new tab in multiple Internet Browsers. A 'Tab' entry in the settings could offer this feature, asking you what should happenj if a new tab is opened.

yaichenbaum:
The correct behavior when opening a directory in a new tab is for it to open in the background. However, opening a regular tab should open it in the foreground.

tgrelka:
> The correct behavior when opening a directory in a new tab is for it to open in the background.
If that is the case, should I open a bug report? Using the ‚open in new tab‘ option from the context menu opens the new tab in the foreground as well.

yaichenbaum:
@tgrelka This issue should be enough.

tgrelka:
I suppose this should be considered a bug report instead of enhancement suggestion then.

yaichenbaum:
@tgrelka Considering that the current behavior is working correctly, I would consider this proposal an enhancement rather than a bug report.

tgrelka:
@yaichenbaum Oh sorry, I interpreted your "the correct behaviour should be" as "it's currently not working as intended (aka a bug)", not as "we should change the behaviour".

yaichenbaum:
@tgrelka This issue should be resolved with #4549.

PR #4536, opened 2021-04-10
New tab in background

cinqmilleans:
**Resolved / Related Issues**
https://github.com/files-community/Files/issues/4506 and https://github.com/files-community/Files/issues/2512

**Details of Changes**
Disable open new tabitem in events HorizontalTabView_TabItemsChanged and VerticalTabView_TabItemsChanged.
The switch has become useless and is replaced by an if. 


lukeblevins:
@yair100 Is this PR's behavior change desirable?

cinqmilleans:
yair100, meowish and myself wished it in issue # 4506. Most software with tabs works this way, so it's more intuitive for the user. If we need the folder right away, we open it in the current tab. If it's for later, you open it in a pending tab. You can open multiple subfolders in tabs easily.

yaichenbaum:
@cinqmilleans Opening a standard new tab looks like it's opening in the background, can you resolve this issue?

cinqmilleans:
Ok I do it.

PR #4549, opened 2021-04-11
Improved behavior when opening new tabs

### LINKED ISSUE #2512: Opening directories in a new tab should open them in the background

cinqmilleans:
**Resolved / Related Issues**
Opening a standard new tab (button add new tab, ctrl+t) active the new tab immediately.
#4506 and closes #2512

**Details of Changes**

### DIFF
--- a/Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
+++ b/Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
@@ -27,15 +27,9 @@ namespace Files.UserControls.MultitaskingControl
 
         private void HorizontalTabView_TabItemsChanged(TabView sender, Windows.Foundation.Collections.IVectorChangedEventArgs args)
         {
-            switch (args.CollectionChange)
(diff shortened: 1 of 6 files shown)


yaichenbaum (approved):
Great work!

Issue #4511, opened 2021-04-08
Use CTRL + Left click for new tab


cinqmilleans:
**Is your feature request related to a problem?**
In most software, you can open a new tab with middle click or ctrl + left click.
The second option is not available in Files.
This is however useful when the mouse does not have a middle button.

**Describe the solution**
Implement ctrl + left click to open new tab.

**Additional comment**
I can implement it.

yaichenbaum:
@cinqmilleans That would be great!

kiikoh:
I agree this needs to be there, as well as middle clicking

yaichenbaum:
@cinqmilleans We don't have plans to go through with this for regular folders. However, we still plan to do this for the sidebar.

cinqmilleans:
@yair100 I hadn't done it because there were bugs with the libraries. These bugs seem to have been fixed. I will create a pr today after further testing.

PR #4552, opened 2021-04-11
ctrl + left click opens in new tab

cinqmilleans:
**Resolved / Related Issues**
ctrl + left click opens items in new tab.
#4511

Only when setting OpenItemsWithOneclick = false.
Otherwise, there would be a conflict with the selection.


yaichenbaum:
@cinqmilleans Being that this will cause conflicts with selecting items, perhaps it should only work for the sidebar and widgets. What do you think?

cinqmilleans:
I will analyze the problem to find the best functional solution to propose.

cinqmilleans:
@yair100 I propose to pause this feature. It is not the most essential. We will make a complete list of keyboard shortcuts and manage them in one place. It will be easier to do after setting up mvvm.

PR #4553, opened 2021-04-11
New tab in background

cinqmilleans:
**Resolved / Related Issues**
Opening a standard new tab (button add new tab, ctrl+t) active the new tab immediately.
#4506 and #2512

**Details of Changes**


yaichenbaum:
@cinqmilleans Didn't we already merge these changes?

cinqmilleans:
Yes, you changed the title and it was committed very quickly. As I couldn't see it I thought I forgot to create the PR. I close.

yaichenbaum:
@cinqmilleans Aa, that makes sense, thanks for your great work!

Issue #4551, opened 2021-04-11
App suddenly close when you are navigating between folders


lettucemarcvel:
**Explain the bug**
Open de app go to Documents and start to navigate between my folders and the app suddenly closes. No specific action or folder.

**To reproduce**
1. Go to your documents that have saved in the C: and start to navigate. 

**Expected behavior**
Do not close the app

**Logfile**

[debug_fulltrust.txt](https://github.com/files-community/Files/files/6292073/debug_fulltrust.txt)
[debug.txt](https://github.com/files-community/Files/files/6292074/debug.txt)

**Screenshots**
NA

**System Information**
 - OS Version: Windows 10.0.19042
 - App version: 1.5.2.0

**Additional comment**
Add any other comment about the problem here. Does this problem happen again after restarting the app?

gave92:
Probably same issue as #4304, happens more or less frequently to most users. Less likely in GridView, less likely with cache turned off.

gave92:
This should be resolved in main

Issue #4510, opened 2021-04-08
Action New Tab always available


cinqmilleans:
**Is your feature request related to a problem?**
The middle click opens files and folders in a new tab. It does not work for drives.
In the SideBar, the middle click opens in the current tab and not in a new tab.

**Describe the solution**
Always open in new tab with middle click.

**Additional comment**
I can implement it.

PR #4566, opened 2021-04-12
In sidebar, middle click opens item in new tab

cinqmilleans:
**Resolved / Related Issues**
In sidebar, middle click opens item in new tab.
#4510

**Details of Changes**
Ctrl+Left is not active because this was causing unexplained crashes. 
It may be related to #4551, because the crashes always concerned the documents.


yaichenbaum (approved):
Nice work, LGTM!

PR #4588, opened 2021-04-13
Move status center logic from code behind to a view model

winston-de:
**Resolved / Related Issues**
The main purpose of this PR is to fix [this](https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/183616358u/overview) crash, which would occur because ```StatusCenterViewModel``` would call the status center control before it was loaded.

**Details of Changes**
Add details of changes here.
- Moved the logic for the status center from code behind to it's own view model
- Added the StatusCenterViewModel to ModernShellPage and ColumnViewShellPage
 
**Validation**
- [x] Built and ran the app

I haven't noticed any issues in any of my testing, but if someone else could also test it would be appreciated.


yaichenbaum (approved):
Looks good, LGTM!

PR #4602, opened 2021-04-14
Improved header background in details layout

yair100:


winston-de:
I was thinking we could make a style for the buttons, so that it would be easier to make changes in the future.

yaichenbaum:
> I was thinking we could make a style for the buttons, so that it would be easier to make changes in the future.

Being that this is the only change for now, wouldn't a style be a little overkill?

winston-de:
@yair100 Yeah, it can probably be done in another PR. It will be a good to have eventually though since there may be a lot more header items in the future.

winston-de (approved):

yair100:
> @yair100 Yeah, it can probably be done in another PR. It will be a good to have eventually though since there may be a lot more header items in the future.

The WinUI DataGrid will hopefully be ready by then, if not, then we will probably want to make this more future proof and flexible.

PR #4603, opened 2021-04-14
best close in debug

cinqmilleans:
In Visual Studio, the application may stop compiling after being closed, because of a lock.
This fix, only in Debug mode, cleanly closes the application instead of suspending it, so that it can be recompiled after a modification.

I don't know if this fix is useful or if I am the only one having the problem.


cinqmilleans:
@yair100 Am I the only one having this problem? How do other developers do it?

yair100 (approved):
LGTM

PR #4604, opened 2021-04-14
More reliable Continue where you left off

### LINKED ISSUE #3801: [Enhancement] Offer to re-open tabs after crash
Crashes are inevitable, but I believe that there is room for improvement for the experience when the user re-opens Files after a crash. Currently, it is hard for continue where you left off after a crash.

I think a good solution would be to show an in-app notification when the user restarts Files, with some options, including one to reopen tabs.
Some other options could be:
- show the error stack trace
- file a bug report

This would make it easy to for users to continue where they left off after a crash, and for devs to view the stack trace.

Toast notification that has these options instead. I believe an in-app notification is better, however, because it's less intrusive.

**Additional context**
I think the best way to do this would be to save a list of open directories in the app settings when a crash is detected.

Rough mockup I made.
image

d2dyno1:
**Resolved / Related Issues**
- Closes #3801 (not exactly it but it does the job)

**Details of Changes**
Tabs are now more reliably saved by Continue where you left off. Now tabs are recovered even after a crash! 

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -442,6 +442,7 @@ namespace Files
 
         private static void AppUnhandledException(Exception ex)
         {
+            SaveSessionTabs();
             Logger.Error(ex, ex.Message);
             if (ShowErrorNotification)
             {


yaichenbaum:
I think this is a good way to solve #3801, LGTM!

yaichenbaum (approved):

PR #4626, opened 2021-04-16
Allow dragging items without selecting them first

### LINKED ISSUE #673: Allow dragging files and folders without selecting them first

winston-de:
**Resolved / Related Issues**
- Closes #673

**Details of Changes**
- For all list view based layout modes (aka all but DataGrid), use built in drag and drop instead of the custom one currently used

This does not work with the DataGrid, but if we plan to switch to the ListView one anyway I don't think it's an issue.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

**Note**
Most of the column view file changes are the xaml auto formatter (that's also why this PR is so large xD).

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -445,24 +445,12 @@ namespace Files
         {
             List<IStorageItem> selectedStorageItems = new List<IStorageItem>();
 
-            object senderContext = null;
-
-            // Select the sender
-            if (sender is GridViewItem gridViewItem)
+            if (sender is DataGridRow dataGridRow)
             {
-                senderContext = gridViewItem.DataContext;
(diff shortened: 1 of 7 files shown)


yaichenbaum (approved):
Great work! Since we are switching to the ListView control, I think it's fine that it doesn't work in the DataGrid control.

PR #4650, opened 2021-04-17
Update all .xlf with new texts

cinqmilleans:
Update all .xlf with new texts. This prevents these files from being changed automatically in Visual Studio.


yaichenbaum (approved):
Thank you! In the future, the project maintainers will take care of updating the strings at the appropriate times before releasing updates.

Issue #4678, opened 2021-04-20
Escape key to exit settings


voiiden:
I often try to exit the settings page using the escape key, as is convention in many other apps, such as discord and steam. I suggest the implementation of this feature, as I don't see any other use for the escape key in this situation.

cinqmilleans:
It's a good idea.
I am always confused because the button to close is not in the same place as to open it. We could put a Close button at the bottom of the sidebar.

If the community accepts these changes, I can implement it.

voiiden:
I'm not sure about that last idea. Back buttons are typically on the top, usually to the left.

yaichenbaum:
@voiiden Another option is to use the back button, it doesn't make much of a difference to me though.

voiiden:
Escape seems the most logical to me. It's literally called escape, as in escape from a menu, and backspace to go back is getting dated, usually getting replaced by Alt + Left arrow.

yaichenbaum:
The reason I suggested back is because you are going back to the last page, esc is usually used for a dialog.

Jaiganeshkumaran:
Settings page is full screen and not a modal. Esc is used for modal and transient surfaces, not for page navigation

yaichenbaum:
@Jaiganeshkumaran Agreed, that's why I think we should use the back button.

yaichenbaum:
This issue should be resolved with #5077.

PR #4688, opened 2021-04-21
close settings with escape (#4678)

cinqmilleans:
**Resolved / Related Issues**

### DIFF
--- a/Files/Views/Settings.xaml.cs
+++ b/Files/Views/Settings.xaml.cs
@@ -2,6 +2,7 @@
 using Files.ViewModels;
 using Windows.ApplicationModel.Core;
 using Windows.ApplicationModel.Resources.Core;
+using Windows.System;
 using Windows.UI.Core;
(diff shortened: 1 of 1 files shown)


Files/Views/Settings.xaml.cs
```diff
@@ -29,6 +30,8 @@ protected override void OnNavigatedTo(NavigationEventArgs eventArgs)
             var currentView = SystemNavigationManager.GetForCurrentView();
             currentView.BackRequested += OnBackRequested;
 
+            Window.Current.CoreWindow.Dispatcher.AcceleratorKeyActivated += CoreDispatcher_AcceleratorKeyActivated;
```
yaichenbaum, line 33: Don't forget to unsubscribe from this event when the user leaves the page. Another option would be to register this from the xaml.

yaichenbaum (changes requested):

cinqmilleans:
I do it.

PR #4690, opened 2021-04-21
Fix texts of multiselect menu

cinqmilleans:
**Resolved / Related Issues**
The translations of the MultiSelect menu do not use the correct id.
These translation texts are not used and this menu always appears in English. 

**Details of Changes**
Change id in all files xlf.


yaichenbaum (approved):
Nice work, LGTM!

PR #4691, opened 2021-04-21
Close settings with escape (#4678)

cinqmilleans:
**Resolved / Related Issues**

### DIFF
--- a/Files/Views/Settings.xaml.cs
+++ b/Files/Views/Settings.xaml.cs
@@ -2,6 +2,7 @@
 using Files.ViewModels;
 using Windows.ApplicationModel.Core;
 using Windows.ApplicationModel.Resources.Core;
+using Windows.System;
 using Windows.UI.Core;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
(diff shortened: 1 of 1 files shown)


Files/Views/Settings.xaml.cs
```diff
@@ -43,6 +46,8 @@ protected override void OnNavigatedFrom(NavigationEventArgs eventArgs)
         {
             var currentView = SystemNavigationManager.GetForCurrentView();
             currentView.BackRequested -= OnBackRequested;
+
+            Window.Current.CoreWindow.Dispatcher.AcceleratorKeyActivated -= CoreDispatcher_AcceleratorKeyActivated;
```
yaichenbaum, line 50: Shouldn't this be in `CoreDispatcher_AcceleratorKeyActivated`?

cinqmilleans: @yair100 I used the same logic as for OnBackRequested. It unsubscribes even if we quit using the button. In your proposal, you can only unsubscribe when you exit using the keyboard shortcut.

yaichenbaum: @duke7553 will this work?

Files/Views/Settings.xaml.cs
```diff
@@ -69,6 +74,15 @@ private void GoBack()
             }
         }
 
+        private void CoreDispatcher_AcceleratorKeyActivated(CoreDispatcher sender, AcceleratorKeyEventArgs e)
+        {
+            if (!e.Handled && e.VirtualKey == VirtualKey.Escape)
```
yaichenbaum, line 79: Can you change it to recognize the back key instead?

cinqmilleans: Ok I do it tomorrow.

PR #4720, opened 2021-04-23
Close settings with key back

### LINKED ISSUE #4678: Escape key to exit settings

cinqmilleans:
**Resolved / Related Issues**
Closes #4678

### DIFF
--- a/Files/Views/Settings.xaml.cs
+++ b/Files/Views/Settings.xaml.cs
@@ -2,6 +2,7 @@
 using Files.ViewModels;
 using Windows.ApplicationModel.Core;
 using Windows.ApplicationModel.Resources.Core;
+using Windows.System;
 using Windows.UI.Core;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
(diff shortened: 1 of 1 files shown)


Files/Views/Settings.xaml.cs
```diff
@@ -29,6 +30,8 @@ protected override void OnNavigatedTo(NavigationEventArgs eventArgs)
             var currentView = SystemNavigationManager.GetForCurrentView();
             currentView.BackRequested += OnBackRequested;
 
+            Window.Current.CoreWindow.Dispatcher.AcceleratorKeyActivated += CoreDispatcher_AcceleratorKeyActivated;
```
d2dyno1, line 33: Could that be done in MainPage.xaml?

cinqmilleans: It’s technically possible. But why ?
This is a behavior of the Settings page, It should be in there.
In addition, we can manage other shortcuts later, to use the settings entirely from the keyboard.

Jaiganeshkumaran:
Please don't use escape for back

Jaiganeshkumaran:
VirtualKey.Back seems the backspace key which will cause issues with text boxes inside the page. Maybe use VirtualKey.GoBack?

yaichenbaum:
> VirtualKey.Back seems the backspace key which will cause issues with text boxes inside the page. Maybe use VirtualKey.GoBack?

This shouldn't be an issue if it's properly handled.

Issue #4682, opened 2021-04-20
Add keyboard shortcuts to change layout mode


pklion:
**Is your feature request related to a problem?**
I'm sometimes frustrating because I have to click 2 times when changing layout modes.

**Describe the solution**
Add keyboard shortcuts to change layout mode.

**Describe alternatives you've considered**
N/A

**Additional comment**
N/A

Rasmagoric:
yeah maybe Alt+ 1 to 5? Or Ctrl + Shift instead of alt?

pklion:
I think Ctrl+Alt+1 to 5 or Alt+1 to 5 is good, isn't it?

cinqmilleans:
@Rasmagoric @pklion @yair100
I have started to implement this issue. I have questions.

1) The order of layouts is not the same in the navigation bar menu and in the context menu. It is annoying to assign a number. It is therefore necessary to use the same order. I propose the order, from the smallest to the biggest.
1. Column
2. Details
3. Tiles
4. Grid (small)
5. Grid (medium)
6. Grid (big)

2) You have to choose shortcuts. I propose:
Alt + 1/6: Layout 1/6
Alt + Subtract -> previous layout
Alt + Wheel Down -> previous layout
Alt + Add -> next layout
Alt + Wheel Up -> next layout

yaichenbaum:
Btw, we already have `ctrl` + wheel up/down and `ctrl` + `+`\`-`

cinqmilleans:
Yes, I just saw it. We could still use ctrl.
ctrl + 1, ctrl + add, ...

yaichenbaum:
Ctrl + 1 - 9 is used to switch tabs.

cinqmilleans:
ok, we will stay on alt + 1/6.

yaichenbaum:
What about `ctrl` + `shift` + 1/6?

cinqmilleans:
ok for me.

pklion:
> What about `ctrl` + `shift` + 1/6?

Looks good to me, too.

Rasmagoric:
I prefer the alt + 1/6.

yaichenbaum:
`ctrl` + 1/6 would be consistent with the existing shortcut `ctrl` + `+`, is `alt` + 1/6 used in other applications for this type of thing?

pklion:
[...]

PR #4722, opened 2021-04-23
Add shortcuts for layouts

cinqmilleans:
Add keyboard shortcuts for layouts: ctrl+shift+1/6.
Reorder layouts in menu, from the smallest to the biggest.


pklion:
#4682

PR #4737, opened 2021-04-24
Added support for executables as drop targets

### LINKED ISSUE #3013: Support drop targets
**Enhance usability and compatibility**
When dragging items in the file or folder view (any view style) applicable drop targets are not recognized and thus an item can not be dropped onto it.

Compare here Files and Explorer actions
Files - drop target Explorer
Files - drop target Files

winston-de:
**Resolved / Related Issues**
- Closes #3013

**Details of Changes**
Add details of changes here.
- Added an "arguments" arg to OpenPath to pass arguments
- Added navigation helper to open paths with items as the arguments
- When an item is dropped on an executable, or a shortcut that has an executable as it's target, invoke the executable with the paths as arguments

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**

https://user-images.githubusercontent.com/59544401/115973101-7454ca00-a507-11eb-8276-4135119a880e.mp4


yaichenbaum:
@winston-de I can't get this to work, I tried dragging a text file over a shortcut to Notepad, and it doesn't look like anything is happening.

winston-de:
@yair100 I can't get it to work with Notepad either, but I think it's because of another issue. For some reason, the "target path" property is null, so it doesn't think the shortcut leads to a program and disables drop support. Does it work with other shortcuts, like VSCode?

yair100:
> @yair100 I can't get it to work with Notepad either, but I think it's because of another issue. For some reason, the "target path" property is null, so it doesn't think the shortcut leads to a program and disables drop support. Does it work with other shortcuts, like VSCode?

That seems to work.

yair100 (approved):
Good work, LGTM.

PR #4744, opened 2021-04-25
Load icons directly from imageres for Libraries, Drives, Sidebar

### LINKED ISSUE #4047: Implement color icons
I feel like the file explorer lacks in color for the Disks, Clouds or any other storage icons (Documents, Downloads, Pictures, etc...) The design is really cool, I love it. Simple, modern but still lacks in color...

I would edit the icons I circled in red with more colorful ones (with the same shape)

Unfortunately, I don't think I've considered any alternative, but I would suggest to let the user decide whether to use the basic ones or the colored ones. There should even be an option where you can add your owns ?

**Additional context**
pic

duke7553:
This PR starts the initial batch of work for loading custom icons for various app surfaces from imageres.dll. These surfaces include two popular widgets and the sidebar. To keep things easy for review, the scope of this PR is limited to this work, but custom folder icons is a work in progress alongside smart logic to detect platform versions which would lack Fluent icon resources. 

Subsequent PRs will also include work to show colorful icons for your cloud storage providers as this is not supported yet.

closes #4047


winston-de:
You may already be aware of this, but FYI the icons currently don't show in the properties dialog. You can fix this by setting the ```SelectedItemsPropertiesViewModel.IconData``` property with a byte array of the image data.

duke7553:
We're going back to the drawing board on some of these changes. No need for review at this point in time.

PR #4750, opened 2021-04-26
Improve debug exception logging

d2dyno1:
**Details of Changes**
Improves exception logging

**Validation**
- [x] Visual Studio showed no errors

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -446,44 +446,46 @@ namespace Files
 
         private static void AppUnhandledException(Exception ex)
         {
-#if DEBUG
-            Debug.WriteLine("--------- UNHANDLED EXCEPTION ---------");
+            string formattedException = string.Empty;
+
+            formattedException += "--------- UNHANDLED EXCEPTION ---------";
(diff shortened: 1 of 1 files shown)


Files/App.xaml.cs
```diff
@@ -446,6 +446,42 @@ private void OnSuspending(object sender, SuspendingEventArgs e)
...
+            }
+            else
+            {
+                Debug.WriteLine("\nException is null!\n");
+            }
+
+            Debug.WriteLine("---------------------------------------");
+
+            Debugger.Break(); // Please check "Output Window" for exception details (View -> Output Window) (CTRL + ALT + O)
+#endif
```
yaichenbaum, line 483: Are we able to save this information to the log file in release mode, or is it only available in debug mode?

d2dyno1: This is only for debug mode.

d2dyno1: @yair100 Do you want me to also include it in log file?

yaichenbaum (approved):

PR #4759, opened 2021-04-27
Possible fix for crash when moving folders

winston-de:
**Resolved / Related Issues**
- Related: [2833388839u](https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/2833388839u/overview)

**Details of Changes**
Add details of changes here.
- The crash seems to have resulted from using ```NativeFileOperationsHelper.MoveFileFromApp``` native function. This pr changes the behavior of ```MoveAsync``` to always use ```MoveDirectoryAsync``` instead of ```MoveFileFromApp```.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**

```MoveDirectoryAsync``` seems to be a bit slow, so I would appreciate any feedback or alternative solutions.

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -446,8 +446,46 @@ namespace Files
 
         private static void AppUnhandledException(Exception ex)
         {
+            string formattedException = string.Empty;
+
+            formattedException += "--------- UNHANDLED EXCEPTION ---------";
+            if (ex != null)
+            {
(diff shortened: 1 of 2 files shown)


winston-de:
It looks like the crash when moving items only occurs under certain conditions. I'm going to see if I can figure out what those conditions are, and add a check for them instead of not using ```NativeFileOperationsHelper.MoveFileFromApp``` at all.

PR #4767, opened 2021-04-27
Allow user to right click text in item name box without ending rename

winston-de:
**Resolved / Related Issues**
- There was an issue for this at some point, however I can't seem to find it. 

**Details of Changes**
Add details of changes here.
- Changed behavior to not end rename if the focused element is an ```AppBarButton``` (the type used for the text box context flyout). Before, rename would end if the user right clicked the text inside the name text box.

**Validation**
- [x] Built and ran the app

I have tested this by accessing the menu via mouse, keyboard, and touchscreen.

**Screenshots (optional)**
image

Note: a more elegant solution would be to check if the element has a parent of type ```FlyoutBase```, however I couldn't get this to work. ```DependencyObjectHelpers.FindParent<FlyoutBase>``` would always return null.

### DIFF
--- a/Files/Views/LayoutModes/ColumnViewBase.xaml.cs
+++ b/Files/Views/LayoutModes/ColumnViewBase.xaml.cs
@@ -275,8 +275,12 @@ namespace Files.Views.LayoutModes
 
         private void RenameTextBox_LostFocus(object sender, RoutedEventArgs e)
         {
-            TextBox textBox = e.OriginalSource as TextBox;
-            CommitRename(textBox);
+            // This check allows the user to use the text box context menu without ending the rename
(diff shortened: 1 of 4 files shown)


yair100 (approved):
LGTM

yaichenbaum:
It seems like the issue was fixed a while back, either way, it doesn't seem like a bad idea to keep this change for edge cases.

winston-de:
@yair100 I just tested on 5.0, it looks like it worked properly with the DataGrid, but not with the grid view browsers.

yaichenbaum:
@winston-de That could be why I didn't see it, I was testing it in the Details layout.

PR #4810, opened 2021-05-01
Add key GoBack for close Settings

### LINKED ISSUE #4678: Escape key to exit settings

cinqmilleans:
**Resolved / Related Issues**
Use key GoBack for close settings
Closes #4678

### DIFF
--- a/Files/Views/Settings.xaml.cs
+++ b/Files/Views/Settings.xaml.cs
@@ -2,6 +2,7 @@
 using Files.ViewModels;
 using Windows.ApplicationModel.Core;
 using Windows.ApplicationModel.Resources.Core;
+using Windows.System;
 using Windows.UI.Core;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
@@ -29,6 +30,8 @@ namespace Files.Views
             var currentView = SystemNavigationManager.GetForCurrentView();
             currentView.BackRequested += OnBackRequested;
 
+            Window.Current.CoreWindow.Dispatcher.AcceleratorKeyActivated += CoreDispatcher_AcceleratorKeyActivated;
+
             var flowDirectionSetting = ResourceContext.GetForCurrentView().QualifierValues["LayoutDirection"];
 
             if (flowDirectionSetting == "RTL")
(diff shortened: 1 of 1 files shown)


Files/Views/Settings.xaml.cs
```diff
@@ -43,6 +46,7 @@ protected override void OnNavigatedFrom(NavigationEventArgs eventArgs)
         {
             var currentView = SystemNavigationManager.GetForCurrentView();
             currentView.BackRequested -= OnBackRequested;
+            Window.Current.CoreWindow.Dispatcher.AcceleratorKeyActivated -= CoreDispatcher_AcceleratorKeyActivated;
```
yaichenbaum, line 49: Can this be handled in the GoBack method?

yaichenbaum: Better yet, can we just declare the keyboard accelerator in the xaml so we don't have to register the event from code behind?

cinqmilleans: I tried this method first. There is a problem, especially visible on the widgets page .
The selection remains in keyboard mode and selects a block.
I tried to move the KeyboardAccelerator tag, use handle, nothing works.
Microsoft recommends this method to add a GoBack shortcut ([link](https://docs.microsoft.com/en-us/windows/uwp/design/basics/navigation-history-and-backwards-navigation#support-access-keys)).

Settings

yaichenbaum: The selection rectangle is the correct behavior when you navigate using the keyboard. In fact, this another reason to implement it this way.

cinqmilleans: You are right. I do it.

cinqmilleans: @yair100 This method generates an error with GoBack. It compiles, but at runtime the enum cannot be parsed. I do not know why. However, it works for the other keys.

cinqmilleans: @yair100 Using KeyboardAccelerator and Goback key is not possible. Maybe after pr #4858. This point is pending.

cinqmilleans: The bug seem be in the framework. Change the .net version could fix it.

yaichenbaum: @cinqmilleans UWP doesn't have official support for .NET 5, the open PR only changes the framework for the FullTrustProcess.

yaichenbaum: @cinqmilleans Instead of detecting the keyboard accelerator, you can use `SystemNavigationManager.GetForCurrentView().BackRequested` which also works for the back button in tablet mode.

cinqmilleans: @yair100 Thanks I will try.

cinqmilleans: @yair100 This is the event that is currently used to detect button presses. My problem is detecting GoBack key. This cannot work with KeyboardAccelerator, but works for other keys. The method with CoreDispatcher works. I still say there is a bug with VirtualKey.GoBack in uwp.

yaichenbaum: Were you able to get `GetForCurrentView().BackRequested` to work?

cinqmilleans: @yair100 I do not understand the question. I should not detect this event but trigger it. My problem is with VirtualKey.GoBack, which causes an error that other keys do not, although VirtualKey.GoBack exists in the enum, in the documentation, and intellisense. The event BackRequested works.

yaichenbaum: Never mind, I thought we were implementing the back button shortcut here.

yaichenbaum:
This pull request is no longer relevant with #5077.

PR #4814, opened 2021-05-02
Fixes an issue for which opening a folder in ColumnView did not add a new column

gave92:
**Resolved / Related Issues**
Fixes an issue for which double clicking on a folder in columnview does not add a new column.

**Details of Changes**
Changes the check to test if an item is a folder to use the PrimaryItemAttribute property instead of ItemType which only works if windows language is set as English.

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

Issue #4643, opened 2021-04-16
Improving the new details view


winston-de:
This issue is to track bugs and enhancements for the new details view. If you have any, please post them here and I'll add them.

- [x] Collapse status column when not in a cloud drive
- [x] Add recycle bin columns, and hide them when outside of the recycle bin
- [x] Columns aren't always collapsed after switching layout modes
- [x] Add empty space to the right of the last column instead of stretching
- [x] #2570 (wasn't possible before with the DataGrid)
- [x] Add button visuals for ascending and descending sort
- [x] Implement horizontal scrolling
- [x] Remember column sizes (feature request)
- [x] Fix double click to rename
- [x] File icons are too small
- [x] Folder icons are cut off slightly
- [x] Better handling of overlapping columns
- [x] Reduce grid splitter width, remove element
- [x] Context menu sometimes fails to open because selected items is empty
- [x] Reimplement #4460
- [x] Fix columns getting squished
- [x] Show status column in a library with a cloud folder
- [x] Fix collapsed column headers not being updated when switching layout modes
- [x] Right click menu to collapse columns
- [x] Don't end rename when name text box content is right clicked
- [x] Add a size column
- [x] Add a date created column (collapsed by default?)
- [x] Fix crash when navigating while renaming
- [x] Fix sticky headers
- [x] Focus file list upon navigation instead of header
- [x] reduce width so that it doesn't horizontal scroll when the window is already large enough
- [x] Double clicking the header divider should fit the column to width
- [x]  Allow tab selecting columns
- [x] Fix issue where columns can't be resized with keyboard only
- [x] Can't invoke headers with keyboard alone
- [ ] #8384
- [ ] #8385

yaichenbaum:
>  Add padding to sides to make it easier to activate rectangle selection and clear selection

@winston-de Instead of stretching the last column, can you just have empty space like we have with the current DataGrid?

winston-de:
@yair100 Yeah, that's a better idea.

josh-65:
#4460 still works for me in 1.5.16.0

complex:
I'd like to be able to reorder columns, for example moving size before date.

An option to auto-size the name column (to be as wide as possible) would be nice too.

winston-de:
> I'd like to be able to reorder columns, for example moving size before date.

@complex This is out of scope for this issue, as the main purpose of this issue to get get to feature parity with the DataGrid based details view. However, please feel free to create a new issue for this.

[...]

PR #4821, opened 2021-05-02
Small improvements for new details view 

gave92:
**Resolved / Related Issues**
For #4643

- [x] Fix collapsed column headers not being updated when switching layout modes
- [x] Fix crash when navigating while renaming
- [x] Fix keyboard shortcuts not working

**Details of Changes**
In addition to the above I've removed the `ColumnsViewModel` property from ListedItem and used Binding+ElementName instead. I haven't found particular performance issues with this change and I personally prefer the code this way (actually I dislike both ways, I wish there was a better way to do this xD). Let me know what you think, reversing this would fine by me.
PS: @winston-de amazing work btw!

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Filesystem/ListedItem.cs
+++ b/Files/Filesystem/ListedItem.cs
@@ -108,9 +108,9 @@ namespace Files.Filesystem
             set => SetProperty(ref customIconSource, value);
         }
 
-        [JsonIgnore]
         private byte[] customIconData;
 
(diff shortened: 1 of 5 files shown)


winston-de (approved):
Looks good to me, thanks for this!

yaichenbaum (approved):
LGTM

PR #4827, opened 2021-05-03
Get smaller size icons (48 vs 128), never look for overlays

duke7553:
Loads library icons from the shell faster by getting them in a smaller size without much difference in quality. Also, never look for overlays for the library widget.

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
Looks good! In fact, this improves the quality of the icons on older versions of Windows.

PR #4833, opened 2021-05-04
Use CTRL + Left click for new tab #4511

oleitao:
**Details of Changes**
Add details of changes here.
- Added option on settings to enable to open folders on new tab using ctrl+left mouse click

**Validation**
- [x] Enable that feature on settings
- [x] Navigate into folder explorer
- [x] Click on Ctrl key and simultaneously perform an left mouse click on folder


Files/Views/SettingsPages/Preferences.xaml
```diff
@@ -119,6 +119,13 @@
                     Header="Open folders in new tab"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{Binding OpenFoldersNewTab, Mode=TwoWay}" />
+
+                <ToggleSwitch
+                    x:Name="OpenFoldersNewTabCtrlLeftClick"
+                    x:Uid="SettingsPreferencesOpenFoldersNewTabCtrlLeftClick"
+                    Header="Ctrl + Left mouse click to open folders in new tab"
+                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
+                    IsOn="{Binding OpenFoldersNewTabCtrlLeftClick, Mode=TwoWay}" />
```
yaichenbaum, line 128: Can you remove the setting for this since it should be the default behavior?

yaichenbaum (changes requested):

gave92:
@oleitao I've extended Ctrl+Left click to the remaining layout modes & added the new string resource.

Files/Views/SettingsPages/Preferences.xaml
gave92: Are you sure about this being default? So you won't be able to Ctrl+Click to select multiple items?

yaichenbaum: Oh, this is for folders? I got mixed up with the sidebar.

yaichenbaum: In that case, it might be best not to add this behavior add all. There is already support for middle clicking on the mouse to open in a new tab.

gave92: Yeah this PR is for folders but I'm not sure if this was what #4511 was supposted to be about

gave92: > In that case, it might be best not to add this behavior add all
Exactly

yaichenbaum: Let's close this for now.

PR #4843, opened 2021-05-05
Load DriveItem icons from filesystem

duke7553:
**Details of Changes**
This changed behavior will fallback to getting the old icons when necessary.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

### DIFF
--- a/Files/Filesystem/DriveItem.cs
+++ b/Files/Filesystem/DriveItem.cs
@@ -7,6 +7,7 @@ using System;
 using System.Threading.Tasks;
 using Windows.ApplicationModel.Core;
 using Windows.Storage;
+using Windows.Storage.Streams;
 using Windows.UI.Core;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Media.Imaging;
@@ -15,7 +16,7 @@ namespace Files.Filesystem
 {
     public class DriveItem : ObservableObject, INavigationControlItem
     {
-        public SvgImageSource Icon { get; set; }
+        public BitmapImage Icon { get; set; }
         public Uri IconSource { get; set; }
 
         private string path;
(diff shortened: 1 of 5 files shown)


yaichenbaum:
> This changed behavior will fallback to getting the old icons when necessary.

@duke7553 Which old icons? I assume the Segoe assets, can you confirm?

Files/Filesystem/DriveItem.cs
```diff
@@ -74,7 +75,10 @@ public DriveType Type
             set
             {
                 type = value;
-                SetGlyph(type);
+                if (Icon == null)
+                {
+                    SetGlyph(type);
+                }
```
lukeblevins, line 81: @yair100

duke7553: For now, it should fallback to the previous SetGlyph behavior

yaichenbaum: It doesn't look like the fallback is doing anything on 20H1.

lukeblevins: @yair100 Can you elaborate because that could mean a number of things?

yaichenbaum: Shouldn't I see the Segoe glyphs instead of the icons from File Explorer on 20H1?

lukeblevins: @yair100 That's not included in this PR.

yaichenbaum: Wasn't the previous SetGlyph behavior to use the segoe assets?

yaichenbaum: @duke7553 When is it supposed to fallback to that?

PR #4856, opened 2021-05-07
Close the search bar with the search button

cinqmilleans:
**Resolved / Related Issues**
When the search bar is open, the search button close and reopen the search bar.
The expected behavior is to close the search bar. 

**Details of Changes**
When the search bar is open, the search button close close the search bar.

### DIFF
--- a/Files/UserControls/NavigationToolbar.xaml.cs
+++ b/Files/UserControls/NavigationToolbar.xaml.cs
@@ -1241,6 +1241,7 @@ namespace Files.UserControls
             {
                 SearchRegion.Text = "";
                 IsSearchRegionVisible = false;
+                SearchButtonIcon.Glyph = "\uE721";
             }
             else
             {
@@ -1251,6 +1252,7 @@ namespace Files.UserControls
                 SearchRegion.UpdateLayout();
 
                 SearchRegion.Focus(FocusState.Programmatic);
+                SearchButtonIcon.Glyph = "\uE711";
             }
         }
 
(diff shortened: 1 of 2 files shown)


yaichenbaum:
@cinqmilleans Should the icon of the search button change to an "x" when the search bar is open?

cinqmilleans:
@yair100 The use of this cross would not be understandable. Keeping the same icon in switch seems better to me. If the community prefers the cross, I can change.

cinqmilleans:
@yair100 I've tested it, it's finally more readable and avoids duplicating the same icon side by side. I do it.

Files/UserControls/NavigationToolbar.xaml.cs
```diff
@@ -1241,6 +1241,7 @@ private void SearchButton_Click(object sender, RoutedEventArgs e)
             {
                 SearchRegion.Text = "";
                 IsSearchRegionVisible = false;
+                SearchButtonIcon.Glyph = "\uE721";
```
yaichenbaum, line 1244: Instead of changing the icon from code behind, can you use x:Bind?

cinqmilleans: @yair100 What's the advantage? Instead of changing the value, we are going to change the property, we have to manage an additional property.

yaichenbaum: The advantage is that button isn't as tightly coupled to the code and therefor it will be easier to work with both the xaml and code.

cinqmilleans: Ok. I do it tomorrow.

cinqmilleans: It is done.

yair100 (approved):
LGTM

PR #4861, opened 2021-05-08
Display correctly (Un)Pin To Start in the context menu

### LINKED ISSUE #4850: Pin to start in overflow submenu when settings off
**Explain the bug**
Pin to start menu is in a sub menu even when "move overflow items to sub menu" is off.

**Expected behaviour**
Should be out with the rest of them.

**Screenshots**
image

**System Information**
 - OS Version: Windows 10 20H2 19042.928
 - App version: v1.5,26

cinqmilleans:
**Resolved / Related Issues**
Pin to start menu is in a sub menu even when "move overflow items to sub menu" is off.
Should be out with the rest of them.
Fixes #4850

**Details of Changes**
This command uses ShowOnShift. This property is bugged.
The expected behavior of ShowOnShift is to show the command only if shift is pressed.
The real behavior of ShowOnShift is to show the command in More when Shift is not pressed.
ShowOnShift is only used for PinItemToStart and UnpinItemFromStart.

However, this command should not use it to always be visible (PinItemToStart and UnpinItemFromStart).
This PR disables ShowOnShift for these commands.

The bug with ShowOnShift is not fixed but it would no longer be used and would not have any impact.
We can decide to fix the bug or remove the unused functionality.

**Validation**
- [ ] Built and ran the app

### DIFF
--- a/Files/Helpers/ContextFlyoutItemHelper.cs
+++ b/Files/Helpers/ContextFlyoutItemHelper.cs
@@ -308,7 +308,6 @@ namespace Files.Helpers
                     Text = "PinItemToStart/Text".GetLocalized(),
                     Glyph = "\uE840",
                     Command = commandsViewModel.PinItemToStartCommand,
-                    ShowOnShift = true,
                     ShowItem = !itemViewModel.CurrentFolder.IsItemPinnedToStart,
                 },
                 new ContextMenuFlyoutItemViewModel()
@@ -316,7 +315,6 @@ namespace Files.Helpers
                     Text = "UnpinItemFromStart/Text".GetLocalized(),
                     Glyph = "\uE77A",
                     Command = commandsViewModel.UnpinItemFromStartCommand,
-                    ShowOnShift = true,
                     ShowItem = itemViewModel.CurrentFolder.IsItemPinnedToStart,
                 },
                 new ContextMenuFlyoutItemViewModel()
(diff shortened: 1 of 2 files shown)


winston-de:
I don't think removing the show on shift functionality entirely is a good solution to this problem. It's purpose is to move less-used/important items to the overflow menu to make the menu less cluttered. I think a better solution to this problem would be to check if "move items to overflow menu" is enabled before moving them.

cinqmilleans:
@winston-de Your analysis is interesting. I propose to apply it.
However, the name ShowOnShift is confusing. I suggest DisplayInMoreMenu. Show on shift is a bonus but not the main one.

winston-de:
> @winston-de Your analysis is interesting. I propose to apply it.
> However, the name ShowOnShift is confusing. I suggest ShowInMoreMenu. Show on shift is a bonus but not the main one.

@cinqmilleans I think a description (```///<summary></summary>```) for the variable would probably help clarify it's purpose better than changing the name.

cinqmilleans:
@winston-de I do not agree. A name must not mislead and a description does not correct the problem, otherwise the reading of the code is misleading. There are only 6 occurrences to change.

winston-de:
> @winston-de I do not agree. A name must not mislead and a description does not correct the problem, otherwise the reading of the code is misleading. There are only 6 occurrences to change.

I guess renaming it is probably fine.

cinqmilleans:
@winston-de It is committed on the branch.

winston-de (approved):

yaichenbaum:
@cinqmilleans The options to pin/unpin to the start menu should only be visible when the shift key is pressed. The issue is that it should only show in the overflow menu when the setting to move items to the overflow menu is turned on.

cinqmilleans:
@yair100 @winston-de So the problem is not directly related to PinItemToStart. This is a bug with ShowInShift. ShowInShift should not show the option if the user does not want a More menu. All the code in this pr is unnecessary. I'm going to close this pr and open a new one for ShowInShift.

PR #4863, opened 2021-05-08
Fix async methods

cinqmilleans:
**Resolved / Related Issues**
Some async methods don't use await.
Some sync methods could be async for better performance.
This causes compiler warnings. 

**Details of Changes**
Modify async/sync state if necessary.
Disable warning if necessary (implementation of an interface)


Files/Filesystem/Cloud/Providers/AmazonDriveProvider.cs
```diff
@@ -35,4 +36,5 @@ public async Task<IList<CloudProvider>> DetectAsync()
             }
         }
     }
+    #pragma warning restore 1998
```
gave92, line 39: This should be one line up (on line 38)

gave92 (changes requested):

cinqmilleans:
@gave92 You are right. It is fixed.

gave92 (dismissed):

yaichenbaum:
@cinqmilleans Can you solve the merge conflicts?

yaichenbaum (approved):

PR #4877, opened 2021-05-11
Fix ShowOnShift

cinqmilleans:
**Resolved / Related Issues**
The functionnality ShowOnShift is bugged.
The expected behavior of ShowOnShift is to show the command only if shift is pressed.
The real behavior of ShowOnShift is to show the command in More when Shift is not pressed.
This is the cause of #4850.

**Details of Changes**
When Shift is not held, the items with ShowOnShift are not displayed.
Else the items are displayed, in the main or more menu, depending on the settings.

**Validation**
- [ ] Built and ran the app


winston-de:
@yair100 @cinqmilleans I believe we still want to be showing ```ShowOnShift``` items in the overflow menu when shift is not held and "and move items to overflow menu" is enabled, no? Also, ```ShowOnShift``` items are now hidden when shift is held "and move items to overflow menu" is enabled.

yaichenbaum:
@winston-de That's not the behavior I envisioned, I expected the overflow to work the same was as it does with other items.

winston-de:
@yair100 I'm not sure if I'm fully understanding, is this the behavior you expect?

| | Move items to overflow enabled | Move items to overflow disabled |
| :------- | :------- | :----------|
| Shift pressed | shown | shown |
| Shift not pressed | shown in overflow | hidden |

This is how it currently behaves in this pr:
| | Move items to overflow enabled | Move items to overflow disabled |
| :------- | :------- | :----------|
| Shift pressed | shown in overflow | shown |
| Shift not pressed | hidden | hidden |

I would like to note that the main reason for moving ```ShowOnShift``` items to the overflow menu in the first place was to make them accessible when only using a touch screen (aka no shift key accessible).

winston-de:
@cinqmilleans Yes I realize my mistake, I apologize for that.

cinqmilleans:
@winston-de Thanks you. Now it is good.

winston-de:
> @winston-de the second chart is closer to the expected behavior, the only difference is that it shouldn't automatically be moved to the overflow menu when it's turned on. It should only move to the overflow menu if there are more than three items before it.

Hiding ```ShowOnShift``` items entirely when shift isn't held makes them inaccessible to touch users. There were also issues where users were unable to find the "pin to start" menu item because they did not realize they had to hold shift to be able to see it. That's why ```ShowOnShift``` items were moved to the overflow menu instead of being hidden in the first place (see [here](https://github.com/files-community/Files/pull/4190#issuecomment-808986414)). 

I would suggest something more like this to help avoid confusion about when items are shown and to keep features accessible when using a touchscreen. 
|| Move items to overflow enabled | Move items to overflow disabled
-- | --| -- |
Shift pressed | shown | shown
Shift not pressed | shown in overflow | shown

yaichenbaum:
@winston-de That looks like a satisfactory solution.

cinqmilleans:
@yair100 @winston-de Ok, I do it tomorrow.

cinqmilleans:
@yair100 @winston-de The branch is commited.

winston-de (approved):
Sorry for the delay, LGTM

PR #4907, opened 2021-05-14
Assorted sidebar icons work

duke7553:
**Details of Changes**
- Fixed issue with NavigationView having a large margin set for its ContentPresenter
- Get filesystem thumbnails for pinned items
- Increases icon size to 20px for navigation surfaces like the sidebar
- Get custom icons for SyncFolder of cloud-based storage providers

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

### DIFF
--- a/Files/App.xaml
+++ b/Files/App.xaml
@@ -138,7 +138,9 @@
                         </LinearGradientBrush.GradientStops>
                     </LinearGradientBrush>
 
-                    <!--  WinUI style is breaking us, overriding with a modified version solves crashes in release  -->
+					<Thickness x:Key="NavigationViewContentPresenterMargin">0,0,0,0</Thickness>
+
+					<!--  WinUI style is breaking us, overriding with a modified version solves crashes in release  -->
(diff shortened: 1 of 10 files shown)


Files/App.xaml
```diff
@@ -138,7 +138,10 @@
                         </LinearGradientBrush.GradientStops>
                     </LinearGradientBrush>
 
-                    <!--  WinUI style is breaking us, overriding with a modified version solves crashes in release  -->
+					<x:Double x:Key="SidebarIconSize">20</x:Double>
+					<Thickness x:Key="NavigationViewContentPresenterMargin">0,0,0,0</Thickness>
```
yaichenbaum, line 142: Can you add a comment that this is temporary?

lukeblevins: @yair100 This likely isn't temporary because they've added that resource with default values. Either way, it doesn't matter as this will always be our desired value.

yaichenbaum (approved):
Great work! LGTM

Issue #4304, opened 2021-03-25
Crash when accessing AppData/LocalLow and sub-directories


emilnymann:
**Explain the bug**
Write a clear description of what the bug is.

**To reproduce**
1. Replace Explorer with Files using the registry files from the documentation.
2. Navigate to C:/Users/username/AppData/ (sometimes it crashes already here.)
3. Navigate to LocalLow and keep navigating through sub-directories.
4. At some point the loading bar will appear for a few seconds and then the app crashes with no error.

**Expected behavior**
Being able to navigate to AppData and subdirectories

**Logfile**
[debug.txt](https://github.com/files-community/Files/files/6206873/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/6206874/debug_fulltrust.txt)

**System Information**
 - OS Version: 
Udgave	Windows 10 Home
Version	20H2
Installeret d.	‎18-‎03-‎2021
Operativsystemets build	19042.868
Visning	Windows Feature Experience Pack 120.2212.551.0

 - App version:
Version 1.5.0.0

**Additional comment**
It seems random whether it just crashes in AppData or a sub-directory, and other sub-directories of AppData than LocalLow don't seem to cause any issues.

emilnymann:
I realised there might be some irrelevant errors from earlier in those logs, so I deleted them and reproduced the crash again. Here are the new logs, which don't seem to reflect the crash at all.

[debug.txt](https://github.com/files-community/Files/files/6206923/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/6206925/debug_fulltrust.txt)

gave92:
Thanks for the report:

- Do you get a notification saying "something went wrong" or does the app just close?
- Could you try if it works is you disable these two settings (turn off caching) under Experimental? (warning: probably won't change a thing but it's worth trying)

![image](https://user-images.githubusercontent.com/9673091/110851026-ce2f6680-82b0-11eb-8974-599e144ff399.png)

emilnymann:
The app just closes with notification.

I had already disabled caching, since I'm not even able to access my user files with it enabled (#4282)

gave92:
@emilnymann What about the following? (I'd like to know if this only happens in details view)
1. Turn Off "Enable individual preferences for individual directories" in Settings > Files and Folders
2. Set Tiles or GridView layout mode
3. Try to trigger the crash

emilnymann:
[...]

PR #4922, opened 2021-05-15
Improve folder enumeration speed

gave92:
**Resolved / Related Issues**
For #4180

**Details of Changes**
- Removes caching feature
- Perf tweaks to folder loading code: removed calls to StorageFolder methods when not strictly needed
- Fixes an issue for which icons are not loaded when browsing mtp devices

**Warning**
These changes creates issues with the "current" details view, I find the app crashing every 5 folders.
This only happens with the WCT datagrid layout mode, the "new" details view works good.
This actually looks like a bug with WCT and has been known for a while (#4304) but these changes seem to make the crash more common.
Depending on how we plan to ship the next version (new or old details view) it may make sense to wait before doing this change, but I'd love to get some feedbacks on how well this PR work for you.

**Validation**
- [x] Built and ran the app
- [x] Tested loading folders from disk, network, mtp (phone)


Files/ViewModels/ItemViewModel.cs
```diff
@@ -740,17 +744,16 @@ public async Task LoadExtendedItemProperties(ListedItem item, uint thumbnailSize
...
                                     using var Thumbnail = await matchingStorageFile.GetThumbnailAsync(ThumbnailMode.SingleItem, thumbnailSize, ThumbnailOptions.UseCurrentScale);
                                     using var headerThumbnail = loadGroupHeaderInfo && isFileTypeGroupMode ? await matchingStorageFile.GetThumbnailAsync(ThumbnailMode.DocumentsView, 36, ThumbnailOptions.UseCurrentScale) : null;
                                     if (Thumbnail != null)
                                     {
                                         await CoreApplication.MainView.DispatcherQueue.EnqueueAsync(async () =>
                                         {
-                                            item.FileImage = new BitmapImage();
                                             item.CustomIconData = await Thumbnail.ToByteArrayAsync();
-                                            await item.FileImage.SetSourceAsync(Thumbnail);
+                                            item.FileImage = await item.CustomIconData.ToBitmapAsync();
```
duke7553, line 756: If there isn't any IconData, then why reference it here?

yaichenbaum:
> Depending on how we plan to ship the next version (new or old details view) it may make sense to wait before doing this change, but I'd love to get some feedbacks on how well this PR work for you.

@gave92 The new details layout will replace the old one in v2.

Files/ViewModels/ItemViewModel.cs
gave92: Here `fileIconInfo.IconData` is null (the icon could not be loaded from FTP) but `item.CustomIconData` (icon from StorageFile's thumbnail) is not.

The previous code did not work:
```cs
item.CustomIconData = await Thumbnail.ToByteArrayAsync(); --> this reads the Thumbnail stream to the end
await item.FileImage.SetSourceAsync(Thumbnail);  --> this is left with a stream already read to the end (empty)
```

duke7553: 👍

lukeblevins (dismissed):
This looks like a good improvement as long as we remove the old details layout before v2

yaichenbaum:
@gave92 It seems like the status bar isn't updating anymore. Another issue is that everything seems very choppy when loading large directories.

gave92:
@yair100 is status bar working in `main` for you?

yair100:
> @yair100 is status bar working in `main` for you?

It looks like it's an issue in `main` as well.

gave92:
>  Another issue is that everything seems very choppy when loading large directories.

@yair100 is this still an issue? If so could you clarify? I would not expect it to be "more choppy" than the current.

yair100:
> > Another issue is that everything seems very choppy when loading large directories.
> 
> @yair100 is this still an issue? If so could you clarify? I would not expect it to be "more choppy" than the current.

It appears to be working better now, I noticed a different issue where the app seems to scroll down after loading the system 32 directory in the details layout.

gave92:
> It appears to be working better now, I noticed a different issue where the app seems to scroll down after loading the system 32 directory in the details layout.

Could you get me a screen recording of this?

yaichenbaum:
> > It appears to be working better now, I noticed a different issue where the app seems to scroll down after loading the system 32 directory in the details layout.
> 
> Could you get me a screen recording of this? I have trouble reproducing

I reset Files and I can't reproduce it anymore 😂

yair100 (approved):
LGTM

PR #4925, opened 2021-05-16
Fixed header alignment on the new details layout

yaichenbaum:


Files/Views/LayoutModes/GenericFileBrowser2.xaml
```diff
@@ -310,14 +311,14 @@
 
                             <Button
                                 Grid.Column="2"
+                                HorizontalContentAlignment="Left"
```
winston-de, line 314: It might be simpler to make this change in the ```HeaderButtonStyle``` instead of doing it in each button individually.

winston-de (approved):

PR #4928, opened 2021-05-16
Fix icon size in new details view

gave92:
**Resolved / Related Issues**
- For #4643

**Details of Changes**
Add details of changes here.
- Fixed too small file icons and overlays (now same size as "old" details view)
- Removed content animation when the list is reset to avoid flickering when loading directories larger than 32 elements
- Left aligned the headers except for the cloud drive status column

**Validation**
- [x] Built and ran the app


gave92:
Wait I think I broke the sticky header behaviour  O.o

winston-de (approved):

winston-de:
@gave92 yeah sticky headers are broken due to [this issue](https://github.com/windows-toolkit/WindowsCommunityToolkit/issues/3952). I think there's a workaround though.

PR #4945, opened 2021-05-18
Reserve space for status glyph in cloud drive directories

### LINKED ISSUE #3837: Reserve space for sync status icons on cloud drives in Tile and Grid views
Currently, when opening a cloud drive or folder, the app doesn't reserve space for sync status icons in Tile and Grid views. As a result of this, the names and icons of files/folders jump to the right once their status actually loads and the sync icons appear.

Reserve space for the sync icons to avoid the jumpy behavior.

Provide a placeholder "loading sync status" icon.

winston-de:
**Resolved / Related Issues**
- Closes #3837

**Details of Changes**
Add details of changes here.
- Hard code width for status glyph and bind visibility to ```InstanceViewModel.IsPageTypeCloudDrive``` so that the space is reserved until it is loaded (if that occurs)

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

### DIFF
--- a/Files/Views/LayoutModes/GridViewBrowser.xaml
+++ b/Files/Views/LayoutModes/GridViewBrowser.xaml
@@ -201,11 +201,12 @@
                 <FontIcon
                     x:Name="CloudDriveSyncStatusGlyph"
                     Grid.Column="0"
+                    Width="20"
                     Margin="10,0,0,0"
-                    x:Load="{x:Bind ((local3:CloudDriveSyncStatusUI)SyncStatusUI).LoadSyncStatus, Mode=OneWay}"
                     x:Phase="2"
                     Foreground="{x:Bind ((local3:CloudDriveSyncStatusUI)SyncStatusUI).Foreground, Mode=OneWay}"
-                    Glyph="{x:Bind ((local3:CloudDriveSyncStatusUI)SyncStatusUI).Glyph, Mode=OneWay}" />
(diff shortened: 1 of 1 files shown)


duke7553 (approved):

Files/Views/LayoutModes/GridViewBrowser.xaml
```diff
@@ -201,11 +201,12 @@
                 <FontIcon
                     x:Name="CloudDriveSyncStatusGlyph"
                     Grid.Column="0"
+                    Width="20"
                     Margin="10,0,0,0"
-                    x:Load="{x:Bind ((local3:CloudDriveSyncStatusUI)SyncStatusUI).LoadSyncStatus, Mode=OneWay}"
```
yaichenbaum, line 205: Does this have a performance tradeoff?

winston-de: Not that I could tell, if there is any it's minimal.
I tried to use ``x:Load``, but the compiler threw an ``Invalid parameter`` exception with the binding.

yaichenbaum: Btw, doesn't this lead to the reverse issue where there is extra space when there is no status?

winston-de: Yes, but only in cloud drive directories (which I don't think is a bad thing because it also keeps items lined up nicely for cloud folders with shortcuts).

yaichenbaum (approved):
LGTM

PR #4968, opened 2021-05-20
Fix null exception in AddItemToSidebarAsync when item StorageFolder is null

winston-de:
**Resolved / Related Issues**
- Fix issue discussed in Discord server caused by awaiting a null task

**Details of Changes**
Add details of changes here.
- Added a null check for the thumbnail task before awaiting it

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**


duke7553 (approved):
This looks good. I think we have a placeholder icon in case the StorageFolder isn't available, but we should investigate that.

yaichenbaum:
> This looks good. I think we have a placeholder icon in case the StorageFolder isn't available, but we should investigate that.

It doesn't seem like it, but it's something to add.

PR #4969, opened 2021-05-21
Fixed an issue where Files would crash when using high contrast themes

### LINKED ISSUE #4319: Files Crashes when Use High Contrast [System] is enabled
**Explain the bug**
When trying to use 'Files' while High Contrast mode is enabled results in a force quit/crash

**To reproduce**
1. Open Windows' Settings
2. Navigate 'Ease of Access' > 'High Contrast' > Enable 'Use high contrast'
3. Open 'Files' app
4. 'Files' crashes

**Expected behavior**
Conforms to Windows' High Contrast standards, doesn't crash

**Logfile**
[Attached]

**System Information**
 - OS Version: Windows 10 20H2 19042.870
 - App version: 1.5.0.0

**Additional comment**
No way to open app, Logfile attached was reset and then crash reproduced before uploading... just to save you some looking through for the cause

yaichenbaum:
- Added missing brush to high contrast resources
- Removed extra resources from app.xaml

Fixes #4319

### DIFF
--- a/Files/App.xaml
+++ b/Files/App.xaml
@@ -37,7 +37,6 @@
                             <SolidColorBrush x:Key="CloudDriveSyncStatusOnlineColor" Color="#0078D7" />
                             <SolidColorBrush x:Key="CloudDriveSyncStatusOfflineColor" Color="#30BB03" />
                             <SolidColorBrush x:Key="CloudDriveSyncStatusExcludedColor" Color="#AAAAAA" />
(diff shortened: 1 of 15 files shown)


winston-de (approved):

winston-de:
This may be unrelated, but it seems like there's no difference between the selected and unselected tabs in hc mode. 
image

yaichenbaum:
> This may be unrelated, but it seems like there's no difference between the selected and unselected tabs in hc mode.
> image
> The tab on the right is the selected one.

@winston-de The brushes need to be updated but that's unrelated.

PR #4970, opened 2021-05-21
Added support for switching the app theme without a restart

yaichenbaum:


Files/Helpers/ExternalResourcesHelper.cs
```diff
@@ -117,12 +118,39 @@ public async Task<bool> TryLoadThemeAsync(AppTheme theme)
...
+            {
+                await TryLoadThemeAsync(NewTheme);
+            }
+        }
+
+        public bool RemoveTheme(AppTheme theme)
+        {
+            try
+            {
+                App.Current.Resources.MergedDictionaries.Remove(App.Current.Resources.MergedDictionaries.Where(x => x.TryGetValue("CustomThemeID", out var key) && (key as string) == theme.Key).FirstOrDefault());
```
duke7553, line 139:
```suggestion
                App.Current.Resources.MergedDictionaries.Remove(App.Current.Resources.MergedDictionaries.FirstOrDefault(x => x.TryGetValue("CustomThemeID", out var key) && (key as string) == theme.Key));
```
Does something like this work in place of two LINQ calls?

Files/ViewModels/SettingsViewModels/AppearanceViewModel.cs
```diff
@@ -107,16 +108,35 @@ public AppTheme SelectedTheme
...
-        public bool ShowRestartControl
+        /// <summary>
+        /// Forces the application to use the correct resource styles
+        /// </summary>
+        private async void UpdateTheme()
         {
-            get => showRestartControl;
-            set => SetProperty(ref showRestartControl, value);
+            // Allow time to remove the old theme
+            await Task.Delay(250);
```
duke7553, line 128: Is there any way we can verify the old theme is removed without hardcoding in a delay?

Files/Helpers/ExternalResourcesHelper.cs
yaichenbaum: @winston-de

winston-de: Yes it would, I didn't know that was possible thanks for pointing it out.

winston-de (approved):

duke7553 (approved):

PR #4978, opened 2021-05-21
Fix a few crashes caused by grouping PR

winston-de:
**Resolved / Related Issues**
N/A

**Details of Changes**
Add details of changes here.
- Fix crash that could occur when ``ItemManipulationModel.RefreshItemsOpacity`` was invoked
- Fix crash that could occur when calling ``GetAllItems`` by updating it to work with grouped collections
- Consolidate some repetitive layout mode code into ``BaseLayout`` (specifically ``GetAllItems`` and the ``RefreshItemsOpacity`` event handler)

**Validation**
- [x] Built and ran the app
- [x] Tested invert selection (which calls ``GetAllItems``) with and without grouping
- [x] Tested cut and paste (which calls ``ItemManipulationModel.RefreshItemsOpacity``) with and without grouping

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -70,8 +70,9 @@ namespace Files
         public CollectionViewSource CollectionViewSource
         {
             get => collectionViewSource;
-            set  {
-                if(collectionViewSource != value)
+            set
+            {
+                if (collectionViewSource != value)
                 {
                     collectionViewSource = value;
                     NotifyPropertyChanged(nameof(CollectionViewSource));
@@ -248,7 +249,7 @@ namespace Files
         {
             ItemManipulationModel.RefreshItemsOpacityInvoked += ItemManipulationModel_RefreshItemsOpacityInvoked;
         }
-        
+
         private void UnhookBaseEvents()
         {
             ItemManipulationModel.RefreshItemsOpacityInvoked -= ItemManipulationModel_RefreshItemsOpacityInvoked;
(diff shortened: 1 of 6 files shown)


Files/BaseLayout.cs
```diff
@@ -253,7 +265,16 @@ private void JumpTimer_Tick(object sender, object e)
 
         protected abstract void InitializeCommandsViewModel();
 
-        protected abstract IEnumerable GetAllItems();
+        protected IEnumerable<ListedItem> GetAllItems()
+        {
+            if (CollectionViewSource.IsSourceGrouped)
+            {
+                // add all items from each group to the new list
+                return (CollectionViewSource.Source as BulkConcurrentObservableCollection<GroupedCollection<ListedItem>>)?.SelectMany(g => g);
```
gave92, line 273: @winston-de are you ok with this change?

Files/BaseLayout.cs
```diff
@@ -438,7 +438,6 @@ private async void FolderSettings_GroupOptionPreferenceUpdated(object sender, Ev
         protected override void OnNavigatingFrom(NavigatingCancelEventArgs e)
         {
             base.OnNavigatingFrom(e);
-            UnhookBaseEvents();
```
gave92: Removed otherwise the method is not called anymore after navigating to another folder.
Alternative is to add call to `HookBaseEvents` in `OnNavigatedTo`, but `HookEvent` is handled differently.

Files/BaseLayout.cs
winston-de: Looks much cleaner, fine by me.

gave92 (approved):

PR #4999, opened 2021-05-23
Auto refresh network folders on change

### LINKED ISSUE #2042: Pasting to shared folder/mapped drive doesn't trigger refresh
Hi,

when pasting to a shared folder or a mapped drive is not causing refresh of the windows as it is done on a local folder, and there is no progress windows shown. It is a problem when pasting large files (WIM file 10Gb ie).

v 0.15.4.0

gave92:
**Resolved / Related Issues**
- Fixes #2042

**Details of Changes**
Add details of changes here.
- Adds support for automatically refresh a network folder when a file is created/deleted/renamed..
Done using the "old" method of watching for folder changes (code taken from 0.9.2) only in case of network folders
Note: triggers a full refresh of the folder on any change <- @duke7553 there's no way to get what's changed is there?
- Fixes some issues with recycle bin updating:
    - Crash when the recycle bin is open in one tab and multiple files are deleted from another tab
    - "Folder empty" text sometimes not appearing/disappearing
    - Contents not updated when emptying the recycle bin

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/ViewModels/ItemViewModel.cs
+++ b/Files/ViewModels/ItemViewModel.cs
@@ -423,8 +423,10 @@ namespace Files.ViewModels
                 FilesAndFolders.Remove(item);
                 if (newIndex != -1)
                 {
-                    FilesAndFolders.Insert(newIndex, item);
+                    FilesAndFolders.Insert(Math.Min(newIndex, FilesAndFolders.Count), item);
                 }
+                IsFolderEmptyTextDisplayed = FilesAndFolders.Count == 0;
+                DirectoryInfoUpdated?.Invoke(this, EventArgs.Empty);
             });
         }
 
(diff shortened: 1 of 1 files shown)


duke7553 (changes requested):
These changes are fine, but I think it would be better if we leveraged the reletiavely new [TryGetChangeTracker](https://docs.microsoft.com/en-us/uwp/api/windows.storage.storagefolder.trygetchangetracker?view=winrt-19041#Windows_Storage_StorageFolder_TryGetChangeTracker) method on StorageFolder.

It would provide more rich change descriptions and also works on any StorageFolder as of 1803. 

This works a bit differently since we aren't subscribing to an event, but saves us from having to get a query result.

duke7553 (approved):
Approving this because StorageLibraryChangeTracker doesn't work with NAS drives

PR #5018, opened 2021-05-24
Removed the old details layout

yaichenbaum:


winston-de (approved):
LGTM. Also I think there's some other code around that exists only for the old details view (eg a lot of the drag and drop code in ``BaseLayout``). We should work on cleaning some of that up as well.

yaichenbaum:
@winston-de We still have a lot of work to do with the new details layout, removing the old one should help us focus on the new one.

PR #5057, opened 2021-05-29
Add experimental option to disable getting shell context menu items

### LINKED ISSUE #5032: Taking 15 seconds to open menu when rightclicking a file or folder
**Explain the bug**
When right-clicking a file or folder to get the menu, where you can delete and rename, its taking about 15 seconds to open, sometimes the app will freeze and then I need to reopen it in order to use it again.
My app is in the latest version from the Microsoft Store so the latest update didnt fixed it.

**To reproduce**
Right click filr ot folder

**Expected behavior**
It should just open the menu immidiately like the normal file explorer

**Logfile**

[debug.txt](https://github.com/files-community/Files/files/6540816/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/6540817/debug_fulltrust.txt)

**System Information**
 - OS Version: Windows 10 21H1 19043.985
 - App version: v1.5.0.0

**Additional comment**
The entire app is very laggy and almost unusable because of it, my pc is very strong (Ryzen 9 5900x nvme and all of the good stuff) so its not the problem

gave92:
**Resolved / Related Issues**
- Closes #5032 (sort of)

**Details of Changes**
Add details of changes here.
- Added experimental option to disable getting shell context menu items

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Strings/en-US/Resources.resw
+++ b/Files/Strings/en-US/Resources.resw
@@ -2412,4 +2412,7 @@
   <data name="NavToolbarWidgets.Label" xml:space="preserve">
     <value>Widgets</value>
   </data>
+  <data name="SettingsLoadShellContextMenuItems.Header" xml:space="preserve">
+    <value>Show explorer's context menu items in Files</value>
+  </data>
 </root>
--- a/Files/Strings/en-US/Resources.resw
+++ b/Files/Strings/en-US/Resources.resw
@@ -2413,6 +2413,6 @@
     <value>Widgets</value>
   </data>
   <data name="SettingsLoadShellContextMenuItems.Header" xml:space="preserve">
-    <value>Show explorer's context menu items in Files</value>
+    <value>Show shortcut menu handlers from Windows Explorer</value>
   </data>
 </root>
--- a/Files/Views/SettingsPages/Experimental.xaml
+++ b/Files/Views/SettingsPages/Experimental.xaml
@@ -43,7 +43,7 @@
 
                 <ToggleSwitch
                     x:Uid="SettingsLoadShellContextMenuItems"
-                    Header="Show explorer's context menu items in Files"
+                    Header="Show shortcut menu handlers from Windows Explorer"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{Binding LoadShellContextMenuItems, Mode=TwoWay}" />
             </StackPanel>
--- a/Files/Views/SettingsPages/Experimental.xaml
+++ b/Files/Views/SettingsPages/Experimental.xaml
@@ -40,6 +40,12 @@
                     Header="Show item owner in properties"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{Binding ShowFileOwner, Mode=TwoWay}" />
+
+                <ToggleSwitch
+                    x:Uid="SettingsLoadShellContextMenuItems"
+                    Header="Show explorer's context menu items in Files"
+                    HeaderTemplate="{StaticResource CustomHeaderStyle}"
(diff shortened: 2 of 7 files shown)


Files/Views/SettingsPages/Experimental.xaml
```diff
@@ -40,6 +40,12 @@
                     Header="Show item owner in properties"
                     HeaderTemplate="{StaticResource CustomHeaderStyle}"
                     IsOn="{Binding ShowFileOwner, Mode=TwoWay}" />
+
+                <ToggleSwitch
+                    x:Uid="SettingsLoadShellContextMenuItems"
+                    Header="Show explorer's context menu items in Files"
```
duke7553, line 46:
```suggestion
                    Header="Show shortcut menu handlers from Windows Explorer"
```

duke7553: Thought it would be better to have some clarity about which items we support displaying alongside standard operations.

Files/Strings/en-US/Resources.resw
```diff
@@ -2412,4 +2412,7 @@
   <data name="NavToolbarWidgets.Label" xml:space="preserve">
     <value>Widgets</value>
   </data>
+  <data name="SettingsLoadShellContextMenuItems.Header" xml:space="preserve">
+    <value>Show explorer's context menu items in Files</value>
```
duke7553, line 2416:
```suggestion
    <value>Show shortcut menu handlers from Windows Explorer</value>
```

lukeblevins (approved):

winston-de:
I think this is worth mentioning here, it turns out we can actually add items to the command bar flyout when it is opened, so we *could* show the flyout with the default options, then add the shell ones once they've been loaded. That would also kind of solve the issue of the flyout taking a long time to open, but it may not look good to have a bunch of items suddenly added.

https://user-images.githubusercontent.com/59544401/120083663-16278380-c07f-11eb-816b-0c179a60d009.mp4

duke7553:
Is there a root cause of why the context menu is occasionally delayed? Is the delay caused by getting the items from the shell, or their icons? I'd be interested in any profiling that's been done in this area, if it's possible.

I haven't looked at the shell shortcut code in almost a year, but we could investigate how to improve perceived responsiveness here.

@winston-de I agree we shouldn't hold up the entire menu for some items, but maybe we can detect scenarios where the menu needs to be displayed but items aren't finished loading yet (after a reasonable duration) and show those particular items in a sub menu item so they don't appear "tacked on" after the primary menu is visible.

gave92:
This is the profiling I did long ago:

| QueryContextMenu [ms] | LoadContextMenu [ms] | SetShellContextmenu (total) [ms] |
|:-----------------:|:-----------------: |:--------------------: |
|57                       |70                       |93                            |
|334                     |368                     |420                          |
|1576                   |1609                   |1681                        |
|2208                   |2232                   |2285                        |

More than 80% of the time is taken by the single line QueryContextMenu and the time taken by it varies greatly from a few milliseconds to more that two seconds.
QueryContextMenu is pretty much a single Win32 API call so I'm not really sure on how to improve this.
_______________________________
But:
- At the time we did not had the possibility to add items to the menu after it was opened
- For some users the delay looks related to the fact that the fulltrustprocess communication is not working for them so the uwp app just freezes for 10s and still shows only Files context menu items

yaichenbaum:
> At the time we did not had the possibility to add items to the menu after it was opened

@gave92 The main reason to disable the shell menu items would be because of the delay, now that we have the option to show the context menu items after the menu is loaded, should work on this instead of a setting to turn off the shell menu items?

gave92:
@yair100 agreed, this can be a backup plan if the smart thing is not ready for v2.

PR #5061, opened 2021-05-30
Details view improvements

gave92:
**Resolved / Related Issues**
- For #4643

**Details of Changes**
Add details of changes here.
- Removed unused RectangleSelection_DataGrid, CustomDataGridStyle classes
- Added sort direction indicators to column headers
- Added size and date created columns
- Fixed an issue with details view icon column width

**Validation**
- [x] Built and ran the app


yaichenbaum:
@gave92 It looks like the icon column is too wide now.

gave92:
Should be same as before: 44px

image

yaichenbaum:
@gave92 I think there is some extra padding already there because so it's not needed.

gave92:
@yair100 ~could you check the new icon column width? I've changed it back to 24px, but looks kinda cramped.~ Nevermind I think I've messed up some paddings.
image

yair100:
> @yair100 ~could you check the new icon column width? I've changed it back to 24px, but looks kinda cramped.~ Nevermind I think I've messed up some paddings.
> image

You can add a fixed width to the icon content so that it doesn't shift after the content is loaded.

gave92:
I hate UIs, could you figure out how to make it stop moving when the folder is empty?

yaichenbaum:
> I hate UIs, could you figure out how to make it stop moving when the folder is empty?

I can try. By the way, can we allow sorting by sync status?

gave92:
Probably but it was never implemented.. next PR :)

yaichenbaum:
> Probably but it was never implemented.. next PR :) Disabled button in sync status looks bad, but if we're going to allow sorting by status it's not an issue.

Sounds good to me, File Explorer allows sorting by status so we might as well.

yaichenbaum:
@gave92 Can you unload the sort icon when it's not being used? It's causing the text to be trimmed when it shouldn't be.

yaichenbaum:
@gave92 I found the cause of the header shifting. It shifts because the header width depends on the size of all the column combined. When the user changes the directory to a directory with a different number of columns, the header needs to recalculate the correct width. I don't have any ideas on how to solve this right now.

gave92:
Mmm not sure that's the same issue: what I see is that all columns shift to the left 12px if the folder is empty. Anyway can be fixed next.

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -319,33 +307,30 @@
                                 <ColumnDefinition x:Name="Column8" Width="0" />
                             </Grid.ColumnDefinitions>
 
-                            <Grid Grid.Column="0" Width="44" />
```
yaichenbaum, line 322: Can you revert this change?

gave92: Nope if I do the columns headers and contents are not aligned

yaichenbaum: I'll take a look again. It seems like some folders have a different width for that column.

gave92: Maybe because it has been saved in preferences for some reason? You coud try disabling "Enable individual preferences for individual folders"

yaichenbaum: That seems to cause other issues.
image

yaichenbaum: > Maybe because it has been saved in preferences for those folders? You could try disabling "Enable individual preferences for individual folders" and see if they are all the same width.

Good catch, I'll give that a try.

yaichenbaum: Can we prevent the user setting from being changed for the icon column?

yair100 (approved):
Great work! LGTM

PR #5108, opened 2021-06-02
Move navigation toolbar to main page

### LINKED ISSUE #4959: Codebase: Move NavigationToolbar to MainPage
| Capability | Priority |
| :---------- | :------- |
| Navigation toolbar will control focused pane | Must |

winston-de:
**Resolved / Related Issues**
- Closes #4959

**Details of Changes**
Add details of changes here.
- [x] move properties to a view model, and add a shell page property for it, so that the model can be updated when the selected tab/pane changes
- [x] update column shell page with the new changes
- [x] move nav toolbar to main page

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**


Files/Files.csproj
```diff
@@ -328,7 +328,9 @@
     <Compile Include="UserControls\Widgets\WidgetsListControl.xaml.cs">
       <DependentUpon>WidgetsListControl.xaml</DependentUpon>
     </Compile>
+    <Compile Include="ViewModels\NavToolbarViewModel.cs" />
```
yaichenbaum, line 331: What value does the view model offer over having the events in the code behind of the control?

duke7553:
@winston-de Looking forward to this improvement. I think it makes sense to separate the processing from the methods which should really be exposed by the control itself. In our case, we would probably want to investigate in future PRs how to accomplish this with NavigationToolbar, which currently has quite a bit of logic unnecessarily in ModernShellPage's code behind.

Files/Files.csproj
winston-de: This way each shell page can only effect it's own nav toolbar view model, so if multiple pages are loading at the same time, they won't both be making conflicting changes to the toolbar ui at once. It also tabs/panes from responding to events when they aren't selected without unsubscribing and subscribing to the events every time the selected page is changed.

yaichenbaum: What is the performance implications of doing this? Are the ViewModels completely destroyed when a tab is closed?

winston-de: Atm, they should be destroyed when ``ModernShellPage`` gets gced, however since there's a bit of an issue with that I'll destroy them manually in the ``Dispose`` function.

winston-de:
https://user-images.githubusercontent.com/59544401/121625459-5cd68f80-ca28-11eb-8445-900f541b2e02.mp4

yair100 (changes requested):

yaichenbaum:
@winston-de it looks like the tabs aren't updating correctly to indicate which tab is selected when two tabs are open.

winston-de:
> @winston-de it looks like the tabs aren't updating correctly to indicate which tab is selected when two tabs are open.

@yair100 I think this issue existed before this pr, and it affects the status bar as well. I'm pretty sure it's an issue with the way events are handled when having both the vertical and horizontal tabs active at the same time. I thought I fixed it, but it would appear not.

yair100:
> > @winston-de it looks like the tabs aren't updating correctly to indicate which tab is selected when two tabs are open.
> 
> @yair100 I think this issue existed before this pr, and it affects the status bar as well. I'm pretty sure it's an issue with the way events are handled when having both the vertical and horizontal tabs active at the same time. I thought I fixed it, but it would appear not.

I can't reproduce this is main so I think it's something to look into.

winston-de:
Hmmm. Does it work properly when you use the vertical multitasking flyout?

yaichenbaum:
> Hmmm. Does it work properly when you use the vertical multitasking flyout?

It mush have been a one time issue, I can't reproduce it anymore.

yair100 (dismissed):

Files/Views/MainPage.xaml.cs
```diff
@@ -98,18 +93,20 @@ public void MultitaskingControl_CurrentInstanceChanged(object sender, CurrentIns
...
         private void PaneHolder_ActivePaneChanged(object sender, EventArgs e)
         {
             SidebarAdaptiveViewModel.NotifyInstanceRelatedPropertiesChanged(SidebarAdaptiveViewModel.PaneHolder.ActivePane.TabItemArguments.NavigationArg.ToString());
-            UpdateStatusBarProperties();
+            UpdateSidebarProperties();
+            UpdateNavToolbarProperties();
         }
 
-        private void UpdateStatusBarProperties()
+        private void UpdateSidebarProperties()
```
duke7553, line 109: Was there rationale behind this name change?

winston-de: Thanks for catching that, I must've mixed up status bar with sidebar xD

Files/Views/ModernShellPage.xaml.cs
```diff
@@ -1206,25 +1205,31 @@ public void Dispose()
...
+            //NavigationToolbar.PathBoxQuerySubmitted -= NavigationToolbar_QuerySubmitted;
+            //NavigationToolbar.BackRequested -= ModernShellPage_BackNavRequested;
+            //NavigationToolbar.ForwardRequested -= ModernShellPage_ForwardNavRequested;
+            //NavigationToolbar.UpRequested -= ModernShellPage_UpNavRequested;
+            //NavigationToolbar.RefreshRequested -= ModernShellPage_RefreshRequested;
+            //NavigationToolbar.RefreshWidgetsRequested += ModernShellPage_RefreshWidgetsRequested;
+            //NavigationToolbar.ItemDraggedOverPathItem -= ModernShellPage_NavigationRequested;
+            //NavigationToolbar.SearchBox.QueryChanged -= ModernShellPage_QueryChanged;
+            //NavigationToolbar.SearchBox.QuerySubmitted -= ModernShellPage_QuerySubmitted;
+            //NavigationToolbar.SearchBox.SuggestionChosen -= ModernShellPage_SuggestionChosen;
```
duke7553, line 1218:
```suggestion
```

We won't need these anymore, will we?

Files/UserControls/NavigationToolbar.xaml.cs
```diff
@@ -28,902 +28,120 @@
...
-                PathText = value;
-                NotifyPropertyChanged(nameof(PathText));
-            }
-        }
 
         private void NotifyPropertyChanged([CallerMemberName] string propertyName = "")
         {
-            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+            //PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
         }
```
duke7553, line 131: We can probably completely remove this since it won't be implemented inside the control's code behind itself.

winston-de:
Marking as draft so I can work on cleaning up a lot of old code.

lukeblevins (approved):
Excellent work!

PR #5149, opened 2021-06-06
Fix for navigationbar: click anywhere on the path components to navigate

### LINKED ISSUE #4879: Path bar only allow clicking on the text to navigate
**Explain the bug**
When I try to go somewhere in the path the text is the only clickable part even though the is a button around it that highlights. it should trigger it with the whole button.

gave92:
**Resolved / Related Issues**
- Closes #4879

**Details of Changes**
Add details of changes here.
- Allows to trigger navigation by clicking anywhere on the path components and not just on the text

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

PR #5151, opened 2021-06-06
Upgraded BaseJsonSettingsModel to v2

d2dyno1:
**Details of Changes**
Upgraded `BaseJsonSettingsViewModel` to Clipboard Canvas' `BaseJsonSettingsModel` v2. Benefits:
- Faster settings access thanks to v2 settings cache
- Now multiple settings can be saved. (With previous version that wasn't possible because of a bug)
- More stability

Bundles also should load even faster now.

**Validation**
- [x] Built and ran the app


yaichenbaum:
@d2dyno1 Can you run benchmarks on this format? If the results are good, we can merge the settings together.

d2dyno1:
> Ok, so I've decided not to make those benchmarks
it doesn't seem like it's worth it and the performance benefit wouldn't be that much to change to new solution
*user won't probably notice
but as I said before, there are some other benefits to using BaseJsonSettingsModel

*06/07/2021*

yaichenbaum (approved):
LGTM

PR #5155, opened 2021-06-07
Removed acrylic from properties dialog to match other dialogs

yaichenbaum:


tsvietOK:
I am not sure this is good idea. Is there a design concept for how this should look like?

tsvietOK (approved):

mdtauk:
I assume the properties dialog will be a ContentDialog now, and not a model window?

If it is still a window, then some Acrylic should be kept.

Properties in Dialog
Properties in Window

yaichenbaum:
@mdtauk It'll continue to be a separate window so that users can open more than one dialog at a time, however, users will be able to add acrylic back with custom themes.

PR #5160, opened 2021-06-07
📓 Updated Hungarian translation

KristofKekesi:


gave92 (approved):

gave92:
Acc there's an issue in this xlf file, it won't build -> lines 2963, 2983

yaichenbaum:
@gave92 These files are very tricky, we should probably build the app before approving these PRs. For some reason the build pipeline doesn't find any issues even if they are there.

gave92:
> For some reason the build pipeline doesn't find issues when it comes to these files.

Yeah that was what got me xD

PR #5174, opened 2021-06-10
Added teaching tip for custom skins

yaichenbaum:
image

### DIFF
--- a/Files/Helpers/ExternalResourcesHelper.cs
+++ b/Files/Helpers/ExternalResourcesHelper.cs
@@ -16,7 +16,7 @@ namespace Files.Helpers
         {
             new AppTheme
             {
-                Name = "DefaultScheme".GetLocalized(),
+                Name = "DefaultSkin".GetLocalized(),
             },
         };
 
--- a/Files/Helpers/ExternalResourcesHelper.cs
+++ b/Files/Helpers/ExternalResourcesHelper.cs
@@ -12,72 +12,72 @@ namespace Files.Helpers
 {
     public class ExternalResourcesHelper
     {
-        public List<AppTheme> Themes = new List<AppTheme>()
+        public List<AppSkin> Skins = new List<AppSkin>()
         {
-            new AppTheme
+            new AppSkin
             {
                 Name = "DefaultSkin".GetLocalized(),
             },
         };
 
-        public StorageFolder ThemeFolder { get; set; }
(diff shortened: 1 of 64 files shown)


Files/Helpers/ExternalResourcesHelper.cs
```diff
@@ -12,72 +12,72 @@ namespace Files.Helpers
...
-            if (App.OptionalPackageManager.TryGetOptionalPackage(Constants.OptionalPackages.ThemesOptionalPackagesName, out var package))
+            if (App.OptionalPackageManager.TryGetOptionalPackage(Constants.OptionalPackages.SkinsOptionalPackagesName, out var package))
             {
                 Debug.WriteLine(package.InstalledLocation.Path);
-                OptionalPackageThemeFolder = package.InstalledLocation;
+                OptionalPackageSkinFolder = package.InstalledLocation;
             }
 
-            ThemeFolder = await ApplicationData.Current.LocalFolder.CreateFolderAsync("Themes", CreationCollisionOption.OpenIfExists);
+            SkinFolder = await ApplicationData.Current.LocalFolder.CreateFolderAsync("Skins", CreationCollisionOption.OpenIfExists);
```
winston-de, line 36: This will mean that any skins the user already has won't be in the new folder. Should we maybe rename the old folder if it exists?

yaichenbaum: I don't believe we shipped support for custom themes yet so I think it's fine to skip renaming the old folder.

winston-de: Support for themes was initially shipped in 1.5. 
image

Files/Helpers/ExternalResourcesHelper.cs
```diff
@@ -33,6 +33,17 @@ public async Task LoadSelectedSkin()
                 OptionalPackageSkinFolder = package.InstalledLocation;
             }
 
+            // ToDo this is for backwards compatability, remove after a couple updates
+            var themeFolder = await StorageFolder.GetFolderFromPathAsync(ApplicationData.Current.LocalFolder.Path + "\\Themes");
```
winston-de, line 37: This could throw a ``FileNotFound`` exception if the user just installed the app and doesn't have the Themes folder, I would suggest moving it to the try catch as well.

winston-de (approved):
Lgtm

PR #5180, opened 2021-06-11
Fix binding error in details layout

winston-de:
**Resolved / Related Issues**
- Fixes invalid cast ``GridLength`` to ``double`` binding error in details layout

**Details of Changes**
Add details of changes here.
- Bind to ``StatusColumn.Length.Value`` instead of ``StatusColumn.Length``

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**

### DIFF
--- a/Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
+++ b/Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
@@ -598,7 +598,7 @@
                                 <Grid
                                     x:Name="ItemStatusGrid"
                                     Grid.Column="4"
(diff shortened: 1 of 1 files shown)


yaichenbaum (approved):
Will this solve the issue where it sometimes has the wrong length?

winston-de:
> Will this solve the issue where it sometimes has the wrong length?

I don't think so, no. This error just affected the width of the status glyph, not the column widths.

yaichenbaum:
> > Will this solve the issue where it sometimes has the wrong length?
> 
> I don't think so, no. This error just affected the width of the status glyph, not the column widths.

I was referring to the width of the status column, it appears that this PR has resolved that issue.

PR #5194, opened 2021-06-13
feat: Switch GridViewBrowser to use StyleSelector

lukeblevins:
I think we are somewhat misusing the ChoosingItemContainer event

We almost always do a `GridViewItem item = new GridViewItem();` inside that handler, when it should be done by the control itself and it defeats the purpose of virtualization unless we stored the containers needed like the [example](https://docs.microsoft.com/en-us/windows/uwp/debug-test-perf/optimize-gridview-and-listview#:~:text=%20Container%20caching%20is%20something%20that%20can%20be,number%20of%20elements%20cached%20and%20available%20for%20recycling.) suggests. The problem is we rely on that event to setup drag and drop and other events subscribed when an item is scrolled into view.

Instead of doing this, (because the benefits are nonexistent for our scenario) I changed GridViewBrowser only [for now] to only subscribe to the events without having to manually create a new ItemContainer each time. I basically use another type of feature called StyleSelector so we don't have to intercept the process of creating the item container. 

**I am still investigating whether this change actually reduces memory usage and if it allows events for item drag and drop to continue working. Code is not final, but your testing would be great here.**

### DIFF
--- a/Files/BaseLayout.cs
+++ b/Files/BaseLayout.cs
@@ -717,9 +717,13 @@ namespace Files
             deferral.Complete();
         }
 
-        protected void InitializeDrag(UIElement element)
+        protected void InitializeDrag(UIElement element, ListedItem item = null)
         {
-            ListedItem item = GetItemFromElement(element);
+            if (item == null)
+            {
+                item = GetItemFromElement(element);
+            }
+
             if (item != null)
             {
                 element.AllowDrop = false;
--- a/Files/Views/LayoutModes/GridViewBrowser.xaml
+++ b/Files/Views/LayoutModes/GridViewBrowser.xaml
@@ -15,8 +15,8 @@
     xmlns:local3="using:Files.Filesystem.Cloud"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
-    xmlns:tui="using:Microsoft.Toolkit.Uwp.UI"
-    x:Name="PageRoot"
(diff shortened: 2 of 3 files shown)


winston-de:
I'm pretty sure those drag and drop event subscriptions aren't necessary anymore since we moved to the GridView built-in drag and drop implementation, so that should work fine.

winston-de:
I'm noticing a memory leak when scrolling through system32, can you reproduce this? Seems to go up as thumbnails load.
image

duke7553:
> I'm pretty sure those drag and drop event subscriptions aren't necessary anymore since we moved to the GridView built-in drag and drop implementation, so that should work fine.

Does that mean we can get rid of them?

winston-de:
> > I'm pretty sure those drag and drop event subscriptions aren't necessary anymore since we moved to the GridView built-in drag and drop implementation, so that should work fine.
> 
> @winston-de Does that mean we can get rid of them?

@duke7553 Yes.

lukeblevins:
> I'm noticing a memory spike when scrolling through system32, can you reproduce this? Seems to go up as thumbnails load.
> image

@winston-de This is likely because we fallback to using Windows.Storage for file thumbnails when the FullTrust process is unavailable. This is useful information because it shows we shouldn't switch to that API full time.

The issue faced shouldn't be related to the changes from this PR, but since @yair100 has a working FullTrust process, he can test the changes here.

gave92:
There's also the possibility of wiring the Drag events using the `ContainerContentChanging` event like it's done [here](https://github.com/zhuxb711/RX-Explorer/blob/8105ec3565461fede35273ca8f632a4ded7c7634/RX_Explorer/View/FilePresenter.xaml.cs#L3305). Don't know about prins and cons of this solution though.

PR #5198, opened 2021-06-13
Fix issue where hamburger button could overlay tabs

winston-de:
**Resolved / Related Issues**
- Fix issue where hamburger button would overlay tabs when window was opened with the sidebar in minimal mode

**Details of Changes**
Add details of changes here.
- Check pane display mode when sidebar is loaded and update tab margin accordingly

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before:
image

After:
image

### DIFF
--- a/Files/Views/MainPage.xaml.cs
+++ b/Files/Views/MainPage.xaml.cs
@@ -278,5 +278,10 @@ namespace Files.Views
 
             e.Handled = true;
         }
+
+        private void SidebarControl_Loaded(object sender, RoutedEventArgs e)
+        {
+            SidebarAdaptiveViewModel.UpdateTabControlMargin(); // Set the correct tab margin on startup
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


Files/Views/MainPage.xaml.cs
```diff
@@ -278,5 +278,10 @@ private void ToggleFullScreenAccelerator(KeyboardAcceleratorInvokedEventArgs e)
 
             e.Handled = true;
         }
+
+        private void SidebarControl_Loaded(object sender, RoutedEventArgs e)
+        {
+            SidebarAdaptiveViewModel.UpdateTabControlMargin(); // Set the correct tab margin on startup
+        }
```
yaichenbaum, line 285: Is it possible detect the loaded event from the sidebar control itself?

winston-de: No, the sidebar doesn't have access to ``SidebarAdaptiveViewModel``, it is only accessible in the main page.

yaichenbaum (approved):

PR #5219, opened 2021-06-14
Update pt-PT translation

esibruti:
Added new translations for pt-PT language.

I add an observation that the word skins in Portuguese is "pele", that is, it is human skin. Therefore, I left the word skin as a foreign word, which makes it easier for end users to understand.


yaichenbaum:
> I add an observation that the word skins in Portuguese is "pele", that is, it is human skin. Therefore, I left the word skin as a foreign word, which makes it easier for end users to understand.

Thanks! This is probably going to be a common issue 😂

gave92 (approved):

PR #5239, opened 2021-06-16
Add sideload release pipeline

yaichenbaum:


gave92 (approved):

gave92:
Is this working? I thought I would be able to download the side load package from [Azure](https://dev.azure.com/lukeblevins150823/Files/_build/results?buildId=7038&view=results) but there's only the store build.

yaichenbaum:
@gave92 It's only setup for release builds right now for security reasons.

PR #5245, opened 2021-06-17
Add Ctrl + Shift + Tab to select previous tab

### LINKED ISSUE #5235: Feature Request: Switch to previous tab with Ctrl + Shift + Tab
I don't think so.

**Describe the solution**
Hello, I saw that you can use Ctrl + Tab to go to the next tab, but you can't use Ctrl + Shift + Tab to go to the previous tab, as you can do in navigators. I would like to have this option so you can switch easily among tabs.

It would be to have another keybind to Ctrl + Shift + Tab to go to the previous tab.

I don't know any alternative to this.

**Scope**
To be honest I'm not sure how I have to fill this since it's a very simple request. If this is mandatory for the request I can edit if you let me know how can I do it for this request.

**System Information**
 - OS Version: Windows 10 Pro 21H1 compilation: 19043.1055 Windows Feature Experience Pack: 120.2212.2020.0

 - App version: 1.5.0.0

**Additional comment**
Thanks so much for the app, this is what windows explorer should be.

BanCrash:
Now you can select previous tab as easy as select next tab.

**Resolved / Related Issues**
- Closes #5235

**Details of Changes**
Added the Ctrl + Shift + Tab keybinding to select previous tab, similar to Ctrl + Tab to select next tab.

I copied and adapted the code used to handle new tabs (Ctrl + T and Ctrl + Shift + T).

**Validation**
- [X] Built and ran the app
- [ ] Tested the changes for accessibility -> I don't know how I can do this.

**Screenshots (optional)**
GIF 17-06-2021 15-44-51

### DIFF
--- a/Files/ViewModels/MainPageViewModel.cs
+++ b/Files/ViewModels/MainPageViewModel.cs
@@ -103,15 +103,31 @@ namespace Files.ViewModels
                     break;
 
                 case VirtualKey.Tab:
-                    // Select the next tab
-                    if ((App.MainViewModel.TabStripSelectedIndex + 1) < AppInstances.Count)
+                    bool shift = e.KeyboardAccelerator.Modifiers.HasFlag(VirtualKeyModifiers.Shift);
+
+                    if (!shift) // ctrl + tab, select next tab
                     {
-                        indexToSelect = App.MainViewModel.TabStripSelectedIndex + 1;
+                        if ((App.MainViewModel.TabStripSelectedIndex + 1) < AppInstances.Count)
+                        {
+                            indexToSelect = App.MainViewModel.TabStripSelectedIndex + 1;
+                        }
+                        else
+                        {
+                            indexToSelect = 0;
+                        }
                     }
-                    else
+                    else // ctrl + shift + tab, select previous tab
                     {
-                        indexToSelect = 0;
+                        if ((App.MainViewModel.TabStripSelectedIndex - 1) >= 0)
+                        {
+                            indexToSelect = App.MainViewModel.TabStripSelectedIndex - 1;
+                        }
+                        else
+                        {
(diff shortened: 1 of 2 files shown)


Files/ViewModels/MainPageViewModel.cs
```diff
@@ -103,15 +103,31 @@ private void NavigateToNumberedTabKeyboardAccelerator(KeyboardAcceleratorInvoked
...
+                        {
+                            indexToSelect = App.MainViewModel.TabStripSelectedIndex + 1;
+                        }
+                        else
+                        {
+                            indexToSelect = 0;
+                        }
                     }
-                    else
+                    else // ctrl + shift + tab, select previous tab
```
yaichenbaum, line 119: Instead of using an if statement, can you add a separate case to the switch statement?

yaichenbaum (changes requested):

Files/ViewModels/MainPageViewModel.cs
BanCrash: Do you mean to create a new switch statement inside the case of Tab?

Like:

case VirtualKey.Tab:
     switch (e.KeyboardAccelerator.Modifiers)

Because the key is the same on both cases, only change modificator, as happen on Ctrl + T scenario:

https://github.com/files-community/Files/blob/f503abcaea35f8c121e1aa3cf90406850c80401b/Files/Views/MainPage.xaml#L176
https://github.com/files-community/Files/blob/f503abcaea35f8c121e1aa3cf90406850c80401b/Files/Views/MainPage.xaml#L183

yaichenbaum: You know what, the way you did it should be fine.

yaichenbaum (approved):
Great work! LGTM

PR #5272, opened 2021-06-19
DrivesWidget: Remove hardcoded row size values

### LINKED ISSUE #5216: Drive Widget: Parts of the content are cut off
Screenshot 2021-06-14 090243

This seems to be causes by having an increased text size set for windows. I'd recommend to move the text a bit closer to the progress bar.

duke7553:
**Resolved / Related Issues**
- Closes #5216

**Details of Changes**
Remove hardcoded row and column size values to properly support text scaling.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
At 130% text scaling
image

### DIFF
--- a/Files/UserControls/Widgets/DrivesWidget.xaml
+++ b/Files/UserControls/Widgets/DrivesWidget.xaml
@@ -167,18 +167,19 @@
                                 VerticalAlignment="Stretch"
                                 ColumnSpacing="12">
                                 <Grid.ColumnDefinitions>
-                                    <ColumnDefinition Width="44" />
+                                    <ColumnDefinition Width="Auto" />
                                     <ColumnDefinition Width="*" />
                                 </Grid.ColumnDefinitions>
                                 <Grid.RowDefinitions>
+                                    <RowDefinition Height="*" />
                                     <RowDefinition Height="Auto" />
-                                    <RowDefinition Height="28" />
-                                    <RowDefinition Height="16" />
+                                    <RowDefinition Height="*" />
                                 </Grid.RowDefinitions>
                                 <Image
                                     Grid.RowSpan="3"
                                     Grid.Column="0"
                                     Height="40"
+									Width="40"
                                     HorizontalAlignment="Center"
                                     x:Phase="1"
                                     Source="{x:Bind Icon, Mode=OneWay}" />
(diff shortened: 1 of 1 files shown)


Files/UserControls/Widgets/DrivesWidget.xaml
```diff
@@ -167,18 +167,19 @@
                                 VerticalAlignment="Stretch"
                                 ColumnSpacing="12">
                                 <Grid.ColumnDefinitions>
-                                    <ColumnDefinition Width="44" />
+                                    <ColumnDefinition Width="Auto" />
                                     <ColumnDefinition Width="*" />
                                 </Grid.ColumnDefinitions>
                                 <Grid.RowDefinitions>
+                                    <RowDefinition Height="*" />
                                     <RowDefinition Height="Auto" />
```
yaichenbaum, line 175: Can you hard code the height for the progress bar row? This ensures there is a enough padding so it doesn't look too crowded.

yaichenbaum (changes requested):

Files/UserControls/Widgets/DrivesWidget.xaml
lukeblevins: @yair100 There's no need for this. I tested it.

yaichenbaum: If you look at the comparison, the progress bar has the wrong padding.

lukeblevins: @yair100 The padding difference is normal because the text above and below is larger (being at 130% scaling). Further, if we wanted to increase the padding, we wouldn't want to hardcode it into the row's height.

yaichenbaum: @duke7553 In that case, can you adjust the padding so that it matches the previous design? Additionally, I just noticed you set the others rows to stretch, is there any reason you choose that over `auto`?

lukeblevins: @yair100 Users will not notice the difference pertaining to the padding because it looks essentially the same when text is at 100 percent scaling. Remember, this PR fixes a critical issue with text being clipped when scaled for accessibility. 

**It is never a good idea to hardcode values.** In fact, Microsoft's guidelines for accessibility advise against this. Further, there is no visual or functional difference in choosing stretch for the row height instead of what existed previous except that the margin is now applied neatly to everything because it is stretched.

yaichenbaum (approved):
Nice work! LGTM

Issue #2963, opened 2021-01-13
Bug rename file/folder on Details View


ariadevl:
**Describe the bug**
There is bug on renaming file on details view with mouse/touchpad. No problem with other views

**To Reproduce**
1. Change view to Details View
2. Rename file/folder
3. **Click/select/move the cursor to the text position with MOUSE/TOUCHPAD**
4. And then back to selection

**Expected behavior**
When click/select/move cursor to text position, the cursor position changed

**Screenshots**
![QSnr3S9C4D](https://user-images.githubusercontent.com/61529802/104408624-5fc77400-559f-11eb-8c02-ad1c08007b59.gif)

**Desktop (please complete the following information):**
 - OS Version: [Windows 10 20H2 19042.685]
 - App version: [v1.0.0.0]

winston-de:
This should be resolved with the new details view.

https://user-images.githubusercontent.com/59544401/116586451-56ec7b00-a8ce-11eb-9ad1-7eae7424f1b7.mp4

yaichenbaum:
@winston-de This is still an issue if you click in a blank area and then click on the file again.

winston-de:
@yair100 I'm having trouble reproducing this. Is it possible you could make a screen recording?

esibruti:
@yair100 I think we can close this

yaichenbaum:
@esibruti this is still an issue in `main`.

yaichenbaum:
This issue should be resolved in `main`.

PR #5289, opened 2021-06-20
Bug rename file/folder on Details View

oleitao:
**Details of Changes**
Based on some tests after change a file/folder name that folder disappear for no reason (image below).
unsolved

- To solve that issue i perform an list refresh in order to update the content with the renamed file names (image below)
solved

**Validation**
- [x] Renamed files with a new name and renamed again back to it's original name
- [x] Renamed folders with a new name and renamed again back to it's original name

### DIFF
--- a/Files/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/Files/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -365,9 +365,11 @@ namespace Files.Views.LayoutModes
             else if (e.Key == VirtualKey.Enter)
             {
                 TextBox textBox = sender as TextBox;
-                textBox.LostFocus -= RenameTextBox_LostFocus;
+                textBox.LostFocus -= RenameTextBox_LostFocus;                
                 CommitRename(textBox);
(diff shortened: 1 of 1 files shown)


gave92:
This is not related to #2963 (which is already solved on main for me). As for the issue shown in the first clip (which I can't reproduce) I'd like to find a better solution than refreshing the whole list..

yaichenbaum:
@gave92 This is still an issue in `main`, the latest steps to repro is to select an item, then move the focus to a different item and then back to the original item.

gave92:
But that's not what it's being solved here, also #2963 states that renaming is stopped when you move the text selection/cursor, which is not an issue on main. The only issue I see on main is that sometimes rename is triggered on first click instead of second click (when you do as you say).

gave92:
HI @oleitao there's no issue if you keep working here on the issue you show in the first post (file disappearing after rename).
But refreshing the whole list is something I'd really like to avoid, if you can you could take a look at ItemsViewModel.cs and try to see why the new reamed files is not added to the list.

yaichenbaum:
Closing due to inactivity.

PR #5300, opened 2021-06-22
Now preview dates uses getfriendlydate format / Fixed a bug with files and folders without creation or modification date

### LINKED ISSUE #5269: Bug report: folders without creation date show "01/01/1601" as creation date.
**Explain the bug**
With folders that on windows explorer doesn't have creation date (only an empty space instead of the date), Files show "01/01/1601". My assumption is that the epoch is 0, which Windows thinks is that date.
I only see this issue with SMB root folder of my Synology NAS. 

**To reproduce**
1. Go to a SMB top folder. I'm using a Synology NAS.
2. Check modification date or right click on properties.
3. It will show "01/01/1601" instead an empty space as windows explorer does.

**Expected behavior**
I think it should show an empty space, as windows does. I already created a simple implementation on GetFriendlyDateFromFormat adding a new first if scenario when the date year is 1601, it will show an empty space, as windows explorer does:

` if (d.Year == 1601)
            {
                return " ";
            }`

**Logfile**
Not pertinent IMO, but here they are:
[debug.txt](https://github.com/files-community/Files/files/6680597/debug.txt)
[debug_fulltrust.txt](https://github.com/files-community/Files/files/6680598/debug_fulltrust.txt)

**Screenshots**
Bug:
image

My solution:
image

Windows explorer solution:
image

**System Information**
- OS Version: Windows 10 Pro 21H1 compilation: 19043.1055 Windows Feature Experience Pack: 120.2212.2020.0

- App version: (1.5.0.0) and main branch.

EDIT: I edited the condition code to make it as easy as before but more specific to the issue at hand, since I found another issue with files that show "0001" as the year.

BanCrash:
**Resolved / Related Issues**
- Closes #5269

**Details of Changes**

- Fixed a bug that caused files or folders without creation or modification date to show "01/01/1601". For this, I added a new if case to check if the year is 1601 return an empty string, as Windows explorer does:
image

- Now preview files and folder uses getfriendlydate format, to fix that bug and to keep consistency while showing dates also.

- Also files and folders properties and details uses getfriendlydate format.

- Fixed a bug that causes that "1 hour" was showed as "1 hours" and "1 minute" as "1 minutes".

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
- Before:
- - File properties of #5269 :
image

- -  Preview pane:

- - - Bug #5269:
image

- - - Friendly date on preview:
image

- After:

- - File properties of #5269 :
image

- - Preview pane:

- - - Bug #5269:
image

- - - Friendly date on preview:
image

### DIFF
--- a/Files/Converters/DateTimeOffsetToString.cs
+++ b/Files/Converters/DateTimeOffsetToString.cs
@@ -13,7 +13,7 @@ namespace Files.Converters
             {
                 ApplicationDataContainer localSettings = ApplicationData.Current.LocalSettings;
                 string returnformat = Enum.Parse<TimeStyle>(localSettings.Values[Constants.LocalSettings.DateTimeFormat].ToString()) == TimeStyle.Application ? "D" : "g";
-                return (Extensions.DateTimeExtensions.GetFriendlyDateFromFormat((DateTimeOffset)value, returnformat));
+                return (Extensions.DateTimeExtensions.GetFriendlyDateFromFormat((DateTimeOffset)value, returnformat, true));
             }
 
             return "";
--- a/Files/Converters/DateTimeOffsetToString.cs
+++ b/Files/Converters/DateTimeOffsetToString.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿using Files.Enums;
+using System;
+using Windows.Storage;
 using Windows.UI.Xaml.Data;
 
 namespace Files.Converters
@@ -9,7 +11,9 @@ namespace Files.Converters
         {
             if (value != null)
             {
-                return (Extensions.DateTimeExtensions.GetFriendlyDateFromFormat((DateTimeOffset)value, "D"));
+                ApplicationDataContainer localSettings = ApplicationData.Current.LocalSettings;
+                string returnformat = Enum.Parse<TimeStyle>(localSettings.Values[Constants.LocalSettings.DateTimeFormat].ToString()) == TimeStyle.Application ? "D" : "g";
+                return (Extensions.DateTimeExtensions.GetFriendlyDateFromFormat((DateTimeOffset)value, returnformat));
             }
 
             return "";
--- a/Files/Converters/DateTimeOffsetToString.cs
+++ b/Files/Converters/DateTimeOffsetToString.cs
@@ -9,7 +9,7 @@ namespace Files.Converters
         {
             if (value != null)
             {
-                return ((DateTimeOffset)value).ToLocalTime().ToString("D");
+                return (Extensions.DateTimeExtensions.GetFriendlyDateFromFormat((DateTimeOffset)value, "D"));
             }
 
             return "";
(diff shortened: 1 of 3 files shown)


yaichenbaum:
> Now preview files and folder uses getfriendlydate format, to fix that bug and to keep consistency while showing dates also.

@BanCrash Can you make sure the date format follows the setting the user choose in the settings dialog?

BanCrash:
> > Now preview files and folder uses getfriendlydate format, to fix that bug and to keep consistency while showing dates also.
> 
> @BanCrash Can you make sure the date format follows the setting the user choose in the settings dialog?

Yep, good point. Added on the last commit.

BanCrash:
I just realize this also changes the functionality with properties and details, so I think this PR makes Files to use friendlydate format in every place:

Properties with app date setting:
image

Properties with system date setting:
image

Details on properties with app date setting:
image

Details on properties with system date setting:
image



I think this is the proper way, so it keeps consistency showing the dates on every place, but I let you know since I didn't mentioned this on the first post because I didn't knew it.

winston-de:
> I think this is the proper way, so it keeps consistency showing the dates on every place, but I let you know since I didn't mentioned this on the first post because I didn't knew it.

I don't think we should show the friendly date format in the preview pane (or details page either). The details in the preview pane are meant to show more detailed information than what's shown in the file list. Same goes for the details page, it is meant to show more precise information.

BanCrash:
> > I think this is the proper way, so it keeps consistency showing the dates on every place, but I let you know since I didn't mentioned this on the first post because I didn't knew it.
> 
> I don't think we should show the friendly date format in the preview pane (or details page either). The details in the preview pane are meant to show more detailed information than what's shown in the file list. Same goes for the details page, it is meant to show more precise information.

What about keeping user format choice (system or app), but if they enable app date format then give them the data without relative timespan in those scenarios?

I mean, if creation date is 22/06/2021 23:00, normally it will say "3 minutes ago", but in details or preview it will show "tuesday, 22 june 2021" if the user has choosen app date format, or 22/06/2021 in any part of Files including details or preview if he has choosen the system date format.

winston-de:
How about showing both (when the user has the date format set to "application"). For example 06/21/2021 would show as "Wednesday, June 21, 2021 (1 day ago)". When set to system date format, it would just show system date.

BanCrash:
> How about showing both (when the user has the date format set to "application"). For example 06/21/2021 would show as "Wednesday, June 21, 2021 (1 day ago)". When set to system date format, it would just show system date.

Sounds good to me, though maybe too much text in that space. In any case I will implement that change and show you the result so you can provide more feedback.

winston-de:
Here's a rough mockup. 
image

BanCrash:
@winston-de done! Here it's the result in spanish:

image

I think it's good.

By the way, in the last commit I fixed a bug that caused "1 hour" to be show as "1 hours" and "1 minute" to be show as "1 minutes". Also added the equal operator in the strange case that the days, hours or minutes are exactly the unit (now comparison is >= instead only >).

winston-de (approved):
LGTM

yaichenbaum:
@BanCrash Great work, thank you!

BanCrash:
> @BanCrash Great work, thank you!

Perfect, thanks you!

PR #5334, opened 2021-06-27
Change grid view browser to use toolkit AdaptiveGridView

### LINKED ISSUE #5305: Grid View Incorrect Item Widths
**Explain the bug**
Grid view item widths are incorrect if you have less files than the number required to wrap the files around to the next row.
For example, if your grid view layout sizing is such that your files should wrap to the next row if there are more than 6 items and you only have two files, the entire row's width is split between those two files instead of the width being calculated as if there were 6 items.

**To reproduce**
1. Use any of the Grid View layouts.
2. Open any folder with one file.
3. Instead of the Grid View 'cell' being the correct width it takes up the entire row.

**Expected behavior**
The width of a 'cell' for a file/item should always be the same regardless of the number of items.

**Screenshots**
Single file with the correct height, but the width uses the full width instead of being restricted to e.g. 1/6 of the width:
image

Multiple files with each 'cell' having the correct width, this happens as long as there are the enough files to complete the first row (meaning that adding one more file would create a new row) which in this case is 6:
image

**System Information**
 - OS Version: Windows 10 21H1 19043.1052
 - App version: 1.5.46

**Additional comment**
I still get this issue after restarting.
Note, if you have enough files so that the width is correct e.g. 6 in this case and delete files the width remains correct until you refresh or change directory.

winston-de:
**Resolved / Related Issues**
- Closes #5305
- Related #issue...

**Details of Changes**
Add details of changes here.
- Removed custom behavior for stretched grid view items and use the Toolkit's AdaptiveGridView control, which does the same thing. This fixes issues like item being stretched if it's the only item. 

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**


yaichenbaum:
@winston-de We switched away from this control due to performance issues. Perhaps @duke7553 can add more to this.

lukeblevins:
@winston-de @yair100 I think the main concern was virtualization, but AdaptiveGridView is based on GridView so I wonder if it did in fact have that feature after all.

yair100 (approved):

yaichenbaum:
@winston-de thank you!

PR #5328, opened 2021-06-27
Added options to hide/show sections on the sidebar

oleitao:
**Resolved / Related Issues**
- Closes #5285

**Details of Changes**
- Added right-click on sidebar sections option to remove sections from sidebar
- Added option on settings panel what sections should be hide/show on sidebar

**Validation**
- [x] Right-click on each sidebar sections and select option 'Unpin from sidebar' to remove from sidebar
- [x] Select on settings panel sections to be hide/show

Edition: Windows 10 Pro
OS build: 19042.1052

PR #5339, opened 2021-06-28
Fix crash when the network drives section on the sidebar is shown

CyberAndrii:
Just cloned the repo for the first time and got the ArgumentOutOfRangeException on startup.

Caused by yesterday's merged PR: https://github.com/files-community/Files/pull/5328

To reproduce just need to enable the "_Show network drives section on the sidebar_" option.

image
image


gave92:
Thanks a lot for bringing this up. This issue happens I believe if no cloud drives are detected so the cloud drives section is not added to the sidebar. I've changed your solution a bit so it works for all sidebar sections, could you verify if it fixes the issue on your end?

CyberAndrii:
Works fine! You're right, I don't have cloud drives.

gave92 (approved):

yaichenbaum:
@CyberAndrii Thank you!

PR #5384, opened 2021-07-08
Phase 1 settings update

yaichenbaum:
image


winston-de:
Looks good. One thing I noticed, there isn't any padding for the expanded content, which looks a bit off.
image

Files/Views/SettingsPages/Multitasking.xaml
```diff
@@ -3,16 +3,612 @@
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:local="using:Files.UserControls.Settings"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     mc:Ignorable="d">
+    <Page.Resources>
+        <Style x:Key="RightAlignedToggleSwitchStyle" TargetType="ToggleSwitch">
```
winston-de, line 10: What do you think about moving this style to it's own file so that it isn't defined in all the settings pages.

yaichenbaum: Good idea, I'll fix this.

yaichenbaum:
> Looks good. One thing I noticed, there isn't any padding for the expanded content, which looks a bit off.
> image

This is a temporary solution that will be changed when we start showing rich previews of custom skins when the section is expanded.

winston-de (approved):

PR #5419, opened 2021-07-13
Fix crash due to missing resource key

gave92:
App on main currently crashes on startup due to missing resource key.

**Details of Changes**
Add details of changes here.
- Changed `RootBackgroundBrush` from ThemeResource to StaticResource

**Validation**
- [x] Built and ran the app


winston-de (approved):
Lgtm, thank you!

PR #5424, opened 2021-07-13
Fix null reference exception crash on startup when the window is small

winston-de:
**Resolved / Related Issues**
- Closes #issue...
- Related #issue...

**Details of Changes**
Add details of changes here.
- Add a null check to ActivePane.TabItemArguments

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/Files/Views/MainPage.xaml.cs
+++ b/Files/Views/MainPage.xaml.cs
@@ -148,7 +148,7 @@ namespace Files.Views
 
         private void PaneHolder_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
         {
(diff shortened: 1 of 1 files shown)


gave92 (approved):

gave92:
Do you know why the crash happens when the window is small? I can't reproduce that crash bug I have nothing against null checks =)

winston-de:
> Do you know why the crash happens when the window is small? I can't reproduce that crash but I have nothing against null checks =)

I'm not really certain why. The exception would occur when the app was opened in a small enough size to change the sidebar to minimal or compact mode, so my best guess is somehow that transition would somehow trigger ``PaneHolder_PropertyChanged``.

PR #5504, opened 2021-07-23
Details view improvements

gave92:
**Resolved / Related Issues**
- Fir #4643

**Details of Changes**
Add details of changes here.
- Fix columns squished to the left on navigation
- Focus file list upon navigation instead of header
- Reduce width so that it doesn't horizontal scroll when the window is already large enough

**Validation**
- [x] Built and ran the app


winston-de (approved):
lgtm

PR #5506, opened 2021-07-23
Double click gridsplitter to resize columns to fit content

gave92:
**Resolved / Related Issues**
- For #4643

**Details of Changes**
Add details of changes here.
- Double clicking the header divider should fit the column to width

@winston-de I was looking into implementing the above change and this is what I came up with.
This seems to work fine when the items are not too many (until list virtualization kicks in). What do you think?

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Helpers/XamlHelpers/DependencyObjectHelpers.cs
+++ b/Files/Helpers/XamlHelpers/DependencyObjectHelpers.cs
@@ -69,27 +69,6 @@ namespace Files.Helpers.XamlHelpers
             }
         }
 
-        public static IEnumerable<T> FindChildren<T>(DependencyObject startNode, Func<T, bool> predicate) where T : DependencyObject
-        {
-            int count = VisualTreeHelper.GetChildrenCount(startNode);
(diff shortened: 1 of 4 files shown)


winston-de:
What about approximately calculating the width based on the length of the string in listed item and font size? It may not be as accurate, but I'd imagine it would be more performant then measuring the width of each text block.

gave92:
I thought that wouldn't work as it's dependent of font scaling, but I haven't tried.

gave92:
3e6c283 -> done! Now only the first few textboxes are measured, the rest is calculated using the length of the string in listed item.

winston-de (approved):
lgtm

gave92:
Sorry to ask to review again @winston-de I've noticed a few issues which I fixed:
- Fix crash when folder is empty
- Resized column size was not saved to folder preferences
- Column max width was not respected
- Added ability resize-to-fit all the columns while I was at it

winston-de (approved):

PR #5517, opened 2021-07-25
Changed the height of the breadcrumb bar to match the search box

yaichenbaum:


Files/UserControls/NavigationToolbar.xaml
```diff
@@ -463,7 +463,7 @@
             <Grid
                 x:Name="ClickablePath"
                 Grid.Column="1"
-                MinHeight="32"
+                MinHeight="33"
```
winston-de, line 466: Nit, should the height of the path box be divisible by 4?

yaichenbaum: Normally yes, but this appears to be an exception. I believe it's because of the border but it may be something else.

winston-de (approved):

PR #5532, opened 2021-07-25
Fix icons in Properties for network & cloud drives

gave92:
**Details of Changes**
Add details of changes here.
- Fixes an issue for which "Hide this section" item would appear when right-clicking on non-header sidebar items
- Fixes an issue for which Properties showed an empty icon for cloud drives and for the network folder
- Remove code for detecting MEGA (newer MEGA client versions use a completely different method)
- Remove code to set drive glyph (those assets have been removed)

**Validation**
- [x] Built and ran the app


winston-de (approved):

yaichenbaum:
Thank you!

Issue #5502, opened 2021-07-23
Shortcut to open the settings dialog.


Tropix126:
**Is your feature request related to a problem?**
At the moment, getting to settings with keyboard navigation can be tedious. Most applications offer a shortcut to it's settings page.

**Describe the solution**
Add a shortcut for opening settings. A general standard for a fair few applications i've seen is <kbd>Ctrl</kbd> + <kbd>,</kbd>

**Describe alternatives you've considered**
N/A

**Scope**

| Capability | Priority |
| :---------- | :------- |
| Keyboard shortcut to open settings. | Must |

**System Information**
 - OS Version: Windows 11 22000.71
 - App version: v1.9.0.0

**Additional comment**
None

winston-de:
Might be good to add an access key as well.

yaichenbaum:
Settings can be accessed by the `alt` + `i` access key.

PR #5534, opened 2021-07-25
Added Access key to open settings

BanCrash:
**Resolved / Related Issues**
- Related #5502 (@winston-de comment)

**Details of Changes**
- Added an access key for accessing settings. I went with "O" since "S" is already used by selection menu. Also, I figured out that it should be an access key that is also used on settings, to avoid a possible exception when you are already on settings and click the access key for opening them again.

I tried to also add a shortcut for entirely closing #5502, but it's throwing an odd exception about the impossibility of opening more than 1 Content Dialog, so until someone implemented it well or I figure out what it's going on at least you will be able to open settings using keyboard via access keys.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before:
image

After:
image

### DIFF
--- a/Files/UserControls/NavigationToolbar.xaml
+++ b/Files/UserControls/NavigationToolbar.xaml
@@ -572,6 +572,7 @@
                     Width="36"
                     Height="32"
                     AutomationProperties.Name="Settings"
+                    AccessKey="O"
                     Background="Transparent"
                     Command="{x:Bind SettingsButtonCommand, Mode=OneWay}"
                     Style="{StaticResource ToolBarButtonStyle}"
--- a/Files/UserControls/NavigationToolbar.xaml
+++ b/Files/UserControls/NavigationToolbar.xaml
@@ -572,7 +572,7 @@
                     Width="36"
                     Height="32"
                     AutomationProperties.Name="Settings"
-                    AccessKey="O"
+                    AccessKey="I"
                     Background="Transparent"
                     Command="{x:Bind SettingsButtonCommand, Mode=OneWay}"
                     Style="{StaticResource ToolBarButtonStyle}"


Files/UserControls/NavigationToolbar.xaml
```diff
@@ -572,6 +572,7 @@
                     Width="36"
                     Height="32"
                     AutomationProperties.Name="Settings"
+                    AccessKey="O"
```
yaichenbaum, line 575:
```suggestion
                    AccessKey="I"
```

It's probably better to use a letter that's in the word "Settings".

yaichenbaum (changes requested):

winston-de:
> 
I tried to also add a shortcut for entirely closing #5502, but it's throwing an odd exception about the impossibility of opening more than 1 Content Dialog, so until someone implemented it well or I figure out what it's going on at least you will be able to open settings using keyboard via access keys.

A simple way to add a shortcut to close the dialog is to add an access key (and/or keyboard accelerator) to ``ButtonClose`` in ``Dialogs\SettingsDialog.xaml``.

Files/UserControls/NavigationToolbar.xaml
BanCrash: image

Could be one of those letters that¡ is not already taken on the main menu? I went with that letter because while testing I pressed again the key and it throws an exception because the settings were already open, so to bypass this I used a key that is already present on settings, so when you open the settings it got overwritten.

yaichenbaum: I don't see where else it's being used 🤔

BanCrash:
> > I tried to also add a shortcut for entirely closing #5502, but it's throwing an odd exception about the impossibility of opening more than 1 Content Dialog, so until someone implemented it well or I figure out what it's going on at least you will be able to open settings using keyboard via access keys.
> 
> A simple way to add a shortcut to close the dialog is to add an access key (and/or keyboard accelerator) to `ButtonClose` in `Dialogs\SettingsDialog.xaml`.

I'm not sure If I don't understand you or it was you that didn't understand me haha.

I want to add a keyboard shortcut for openning settings from main menu. And I achieve it, but when the settings opens also throws an exception that says something like "You can't open more than 1 DialogContent at the same time".

Files/UserControls/NavigationToolbar.xaml
BanCrash: > I don't see where else it's being used 🤔

No, no. My point is, the access key for opening the settings should be one access key used on settings, so when you open settings, the access key for opening settings is disabled, avoiding the possibility of getting an exception.

If you use for example "E", since it's not used on settings, if you open settings and after that you used again the access key "E", it will throw an exception. If the access key is, for example, "O", when you open settings and click again the access key "O" it will not try to open settings, but to open "At startup" page.

yaichenbaum: I'll work on adding proper handling to prevent the exception so that you can use the suggested access key.

BanCrash: Ok then, replaced.

yaichenbaum (approved):
LGTM

PR #5535, opened 2021-07-25
Fixed an issue where opening the settings dialog would sometimes crash the app

yaichenbaum:


gave92 (dismissed):

gave92:
Looks good even if I'm not sure how can that happen, do you have steps to reproduce this issue?

yaichenbaum:
> Looks good. I'm not sure how that can happen though, do you have steps to reproduce this issue?

It happens when you quickly click on the setting button multiple times before the dialog opens.

Files/ViewModels/MainViewModel.cs
```diff
@@ -146,11 +146,11 @@ private void DetectFontName()
 
         public static async void OpenSettings()
         {
-            if (!UIHelpers.IsAnyContentDialogOpen())
+            if (UIHelpers.IsAnyContentDialogOpen())
```
gave92, line 149: Mmmm it looked correct before.

yaichenbaum: It was the opposite of what I wanted 😂
We only want to open the settings dialog when there aren't any other dialogs open.

yaichenbaum: It doesn't seem to have worked to fix the issue 🤔

gave92 (approved):
Right the first time ;)

PR #5602, opened 2021-07-28
Fixes an issue for which non-conflicting files are not copied if there's any conflict

gave92:
**Details of Changes**
Add details of changes here.
- Fixes an issue for which non-conflicting files are not copied if there's any conflict

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

PR #5608, opened 2021-07-29
Added colored preview pane icon

yaichenbaum:


winston-de (approved):

mdtauk:
How are you colourising these glyphs, IIRC Coloured Fonts have an embedded colour palette, but this can't be changed by the user?  Have you found a way to do it?

winston-de:
> How are you colourising these glyphs, IIRC Coloured Fonts have an embedded colour palette, but this can't be changed by the user? Have you found a way to do it?

Each icon is split into two glyphs and displayed stacked, the top one colored with the user's accent color.
For example, this is the paste icon:
image 

which becomes
image

mdtauk:
> > How are you colourising these glyphs, IIRC Coloured Fonts have an embedded colour palette, but this can't be changed by the user? Have you found a way to do it?
> 
> Each icon is split into two glyphs and displayed stacked, the top one colored with the user's accent color.
> For example, these are the glyphs for the paste icon:
> image
> 
> which become
> image

Ah so its not a coloured font, but two glyphs which overlap - that is a neat solution.

Issue #5555, opened 2021-07-26
Right click menu has a visual glitch when shown


Neoran01:
**Explain the bug**
Right click menu has a visual glitch when shown. The glitch can occur multiple times and it happens regardless if the app is maximized or not. A restart does not fix the issue, both temporarily and permanently.

**To reproduce**
1. Open the Files app
2. Go to any directory (ex. the C: directory)
3. Right click on an item or empty space 
4. A glitched right click menu will be shown.

**Expected behavior**
The right click menu should be displayed correctly with no visual glitch.

**Logfile**
[debug.log](https://github.com/files-community/Files/files/6878471/debug.log)
[debug_fulltrust.log](https://github.com/files-community/Files/files/6878475/debug_fulltrust.log)

**Screenshots**
Unmaximized:
![Screenshot 2021-07-26 210810](https://user-images.githubusercontent.com/87983083/126994762-7793aec0-04a6-4364-af3e-03b3fe0370df.png) 

Maximized:
![Screenshot 2021-07-26 210847](https://user-images.githubusercontent.com/87983083/126995004-4e905bd7-08a8-46f6-9c19-fc37dcf2bc9e.png)

**System Information**
 - OS Version: Windows 11 Home 21H2 (10.22000.100) 
 - App version: Version: 1.9.14.0

**Additional comment**
Does this problem happen again after restarting the app? _Yes_

gave92:
Thanks for the report. Closing this as caused by a WinUI issue.

gave92:
@winston-de currently the Commandbarflyout does not seem to grow in width when items are added dynamically. I'm pretty sure this was asked on the WinUI repo but as feature request not bug so I'm not sure if/when they're going to fix it.
Hopefully we can find a workaround, e.g trying setting a MinWidth or something?

winston-de:
> @winston-de currently the Commandbarflyout does not seem to grow in width when items are added dynamically. I'm pretty sure this was asked on the WinUI repo but as feature request not bug so I'm not sure if/when they're going to fix it.
> Hopefully we can find a workaround, e.g trying setting a MinWidth or something?

I think settings min width would probably fix it.

yaichenbaum:
This issue should be resolved with #5622.

PR #5622, opened 2021-07-29
Minor ui fixes and enhancements

### LINKED ISSUE #5069: New tabs created via app activation aren't switched to automatically
**Explain the bug**
When a new tab is created by activating the app (eg. jumplist item, Start tile, or files.exe), it is not automatically switched to.

**To reproduce**
1. Make sure "open folders in new tab" is selected in startup settings
2. Open Files jumplist
3. Click on any folder
4. New tab is opened, but not switched to

**Expected behavior**
Tab should be automatically switched to when created via app activation.

**Logfile**
N/A

**Screenshots**

**System Information**
 - OS Version: 19043.1023
 - App version: 1.5.27.0

**Additional comment**
A possible fix for this is to add an optional parameter ``switchToTab`` (false by default) to ``MainPageViewModel.AddNewTabByPathAsync``, and then add a check that if true, switches to that tab after adding it.

winston-de:
**Resolved / Related Issues**
- Closes #issue...
- Related #issue...

**Details of Changes**
Add details of changes here.
- added privacy policy expander to about
- added link to fluent ui system icons license in about -> third party licenses (since many toolbar icons are based off of icons in that repo)
- semi transparent for list in security properties instead of opaque one so that mica is shown
- fix issue where the searchbox and path autosuggestbox had different heights then the breadcrumb causing a small layout shift
- Remove excess border around tab content (spanned left and bottom)
- Fix issue where accent color for colored icons wouldn't update when the user's accent color changed
- fix issue where tab would not be switched to when added via launch event args (closes #5069)
- fix issue where tab visual state would be unselected when opened with path
- fix issue where dv header would be focused when using filter type
- disable share when all selected items are hidden
- remove copy path option from toolbar
- add cut option to toolbar
- double click resizer to expand or collapse sidebar
- fix issue where markdown preview background was different the parent background
- workaround for visual glitch with layout right click menu (#5555)
- change accent color foreground for colored icons when parent toggle is checked
- fix issue where sort option had wrong icon in the layout context flyout
- lighten icon accents in dark mode to improve contrast
- center status glyph in dv
- fix inconsistent MenuFlyoutItemWithImage styling
- improve breadcrumb item styling 
- use extended tooltip instead of just item name in details view
- account for grid splitter width in details view items to fix some alignment issues towards the end of the rows
- reduce status column width
- use ControlFillColorDefault for status center card bg to match other surfaces
- Hide operation status when there are no ongoing operations

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

image
image
image


Files/Views/SettingsPages/About.xaml
```diff
@@ -221,6 +241,11 @@
                     AutomationProperties.Name="QuickLook"
                     Content="QuickLook"
                     NavigateUri="https://github.com/QL-Win/QuickLook/blob/master/LICENSE-GPL.txt" />
+                
+                <HyperlinkButton
+                    AutomationProperties.Name="Fluent UI System Icons"
```
yaichenbaum, line 246: Should this get strings in the en-US resource file?

Files/Views/SettingsPages/About.xaml
```diff
@@ -176,6 +181,21 @@
                     </local:SettingsBlockControl.ExpandableContent>
                 </local:SettingsBlockControl>
 
+                <local:SettingsBlockControl
+                    x:Uid="SettingsPrivacy"
+                    Title="Privacy policy"
+                    Width="{StaticResource SettingsBlockControlDefaultWidth}"
+                    HorizontalAlignment="Left"
+                    Loaded="SettingsBlockControl_Loaded">
```
yaichenbaum, line 189: Should we include the actual text in the app so that it works offline and can also be trasnlated?

Files/Strings/en-US/Resources.resw
```diff
@@ -2691,4 +2691,24 @@
...
+  </data>
+  <data name="SettingsFluentUISystemIconsLicenseButton.Content" xml:space="preserve">
+    <value>Fluent UI System Icons</value>
+  </data>
+  <data name="SettingsFluentUISystemIconsLicenseButton.AutomationProperties.Name" xml:space="preserve">
+    <value>Fluent UI System Icons</value>
+  </data>
+  <data name="SettingsPrivacyPolicyMarkdownBlock.Text" xml:space="preserve">
+    <value>
+**Privacy Policy**
```
d2dyno1, line 2705: Should that really be here? What if privacy policy is updated and some languages didn't reflect the changes? Or maybe meaning and phrasing is altered in a translation? Wouldn't it be better if privacy policy was just pulled from github through an API interface?

winston-de: That's fair, but it seems unlikely the privacy policy will really change. And if it does, the existing translations can just be reset or marked as needs review so that translators know to update it. Having it in the app also allows it to work offline as well.

Also cc @yair100

Files/BaseLayout.cs
```diff
@@ -546,8 +546,14 @@ public async void BaseContextFlyout_Opening(object sender, object e)
                     (i as AppBarButton).Click += new RoutedEventHandler((s, e) => BaseContextMenuFlyout.Hide());  // Workaround for WinUI (#5508)
                 });
                 primaryElements.ForEach(i => BaseContextMenuFlyout.PrimaryCommands.Add(i));
-                secondaryElements.ForEach(i => BaseContextMenuFlyout.SecondaryCommands.Add(i));
-
+                secondaryElements.ForEach(i =>
+                {
+                    if(i is AppBarButton appBarButton)
+                    {
+                        appBarButton.MinWidth = 350; // setting the minwidth to a large number is a workaround for #5555
```
gave92, line 553: Same should be done for ItemContextMenuFlyout I think (has the same issue)

Files/Strings/en-US/Resources.resw
yaichenbaum: The privacy policy isn't really something that will be updated very often and like Winston said, we can update the strings so translators have to translate it again.

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -596,13 +596,12 @@
...
                                     Grid.Column="4"
                                     Width="{Binding ColumnsViewModel.StatusColumn.Length.Value, ElementName=PageRoot, Mode=OneWay}"
-                                    Padding="12,0,0,0"
                                     HorizontalAlignment="Stretch"
                                     VerticalAlignment="Center"
                                     Visibility="{Binding ColumnsViewModel.StatusColumn.Visibility, ElementName=PageRoot, Mode=OneWay}">
                                     <FontIcon
                                         x:Name="CloudDriveSyncStatusGlyph"
-                                        HorizontalAlignment="Left"
+                                        HorizontalAlignment="Center"
```
gave92, line 604: The Status column header should be centered as well

Files/Views/Pages/PropertiesSecurityAdvanced.xaml
```diff
@@ -155,7 +155,7 @@
                 x:Name="ListView"
                 Grid.Row="2"
                 x:Load="{x:Bind ViewModel.FilePermissions.CanReadFilePermissions, Mode=OneWay}"
-                Background="{ThemeResource SolidBackgroundAcrylic}"
+                Background="{ThemeResource ControlFillColorDefaultBrush}"
```
gave92, line 158: Maybe it was there before but is it possible to fix the background color of the list header? Or remove the margins.

image

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
winston-de: image

Files/Views/Pages/PropertiesSecurityAdvanced.xaml
winston-de: I updated it to be tinted with the control color instead of a hard coded color, and also remove the margin.

image

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -596,13 +602,12 @@
...
```
gave92, line 610: There's still something strange with the alignment, doesn't look centered. Sidenote: we might reduce the default status column width, it's pretty large for showing an icon..

image

Files/Views/Pages/PropertiesSecurityAdvanced.xaml
```diff
@@ -175,48 +183,48 @@
                 </ListView.ItemContainerStyle>
                 <ListView.Header>
                     <Grid
```
gave92, line 185: This can be probably be turned into a Border now.

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
winston-de: thanks, it should be properly lined up now
image

gave92:
You know what I'm going to say right? xD I won't cry if it's left left-aligned.
image

winston-de:
ok *now* I think I fixed it 😆
image

Files/ViewModels/ColumnsViewModel.cs
```diff
@@ -6,6 +6,7 @@ namespace Files.ViewModels
 {
     public class ColumnsViewModel : ObservableObject
     {
+        [JsonIgnore]
```
gave92, line 9: Does the [JsonIgnore] attribute work if you put it on a private field? Isn't it is supposed to be put on the `public ColumnViewModel IconColumn` property below?

winston-de: Yeah, thanks for catching that

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
yaichenbaum: I think we should keep it left aligned to match the other columns.

winston-de: > I think we should keep it left aligned to match the other columns.

Just the header or both?

yaichenbaum: I think both should be left aligned, what we can do to improve the design is make the icons a little smaller and closer in line with the icons in File Explorer.

Files/UserControls/NavigationToolbar.xaml
```diff
@@ -521,7 +522,7 @@
                     </Button.KeyboardAccelerators>
                 </Button>
 
-                <Grid x:Name="StatusCenterGrid" Visibility="{x:Bind ShowStatusCenter, Mode=OneWay}">
+                <Grid x:Name="StatusCenterGrid" Visibility="{x:Bind StatusCenterViewModel.AnyOperationsOngoing, Mode=OneWay}">
```
gave92, line 525: This means that the user can't see if an operation completed successfully or not. This probably needs to be hidden if no operations were started OR if the user cleared all messages from the flyout

Files/ViewModels/ColumnsViewModel.cs
```diff
@@ -120,14 +119,15 @@ private void SetColumnSizeProportionally(double factor)
 
     public class ColumnViewModel : ObservableObject
     {
-
+        [JsonIgnore]
```
gave92, line 122: This too should be moved on the public property (or if you know that it works regardless I'd learn something).

Files/ViewModels/ColumnsViewModel.cs
```diff
@@ -179,6 +181,16 @@ public GridLength Length
...
+
+        public GridLength LengthIncludingGridSplitter
+        {
+            get => IsHidden || UserCollapsed ? new GridLength(0) : new GridLength(UserLength.Value + (IsResizeable ? gridSplitterWidth : 0));
+        }
+
+        [JsonIgnore]
+        public bool IsResizeable { get; set; } = true;
+
         [JsonIgnore]
```
gave92, line 194: This too should be moved on the public property.

Files/UserControls/NavigationToolbar.xaml
yaichenbaum: That's a good point, do we have any properties that track if there are any uncleared messages in the flyout?

winston-de: What about checking if the status banner list count is 0?

yaichenbaum: We would need to add a property for that.

gave92 (approved):

yaichenbaum:
@winston-de great work!

winston-de:
> @winston-de great work!

Thank you :)

Issue #5600, opened 2021-07-28
Rename using mouse click is easily triggered and is unstable


xezrunner:
**Explain the bug**
Clicking to select files/folders at their names unintentionally triggers the rename functionality. During multi-selection, the rename functionality is still invoked and can crash the app.

**To reproduce**
1. Open Files and navigate to a folder.
2. Click on different items' names. _Timing doesn't matter, you just have to click one item after the other._
3. Observe how the rename functionality is invoked unintentionally.
_Additionally:_
_4. Try selecting multiple files using Ctrl / Shift + clicking_
_5. Observe how the rename functionality is invoked on the first selection item only._
_6. Now try selecting and unselecting the first selected item._
_7. The app should crash._

**Expected behavior**
- Rename using single-click should not be triggered this easily.
- Rename using single-click should not work while multiple items are being selected.

**Logfile**
Since the app crashed during the recording of this bug, I'm including a log file here.
I was holding Control and selecting folders. Check recording below.
[debug.log](https://github.com/files-community/Files/files/6892362/debug.log)
[debug_fulltrust.log](https://github.com/files-community/Files/files/6892363/debug_fulltrust.log)

**Screenshots / screen recordings**
https://user-images.githubusercontent.com/8061077/127310052-60aca982-4cfe-4d32-8290-97ce85b6d87a.mp4

**System Information**
 - OS Version: Windows 11 21H2 build 22000.100 (co_release)
 - App version: 1.9.15.0

**Additional comment**
- Perhaps there could be a toggle for whether renaming using single-click is enabled.
- Click to rename should only be invoked when the user has selected the item **and** clicked on the name label. It should **not** trigger when the user selects another item _(Is it checking against a global 'is item selected' variable?)_

yaichenbaum:
@xezrunner is this the same issue as #2963?

xezrunner:
@yair100 No, that issue is in regards to the textbox and its cursor/focus behaviors, **once rename has been invoked**. This one is about the app invoking the rename function unintentionally during simple selection.

xezrunner:
With version 1.9.18.0 and commit https://github.com/files-community/Files/pull/5635, click-renaming is no longer erratic when selecting items and cannot be invoked when selecting multiple items.

Issue #4153, opened 2021-03-18
Legacy Renaming


mtratchet16:
**Is your feature request related to a problem? Please describe.**
Not really a problem as much a loss of productivity (though I admit, not by much).

**Describe the solution you'd like**
Being given the option to rename files in the manner by which the legacy Explorer has an option to: By clicking the file name (twice) and typing the name. As of now, it demands a right-click, finding "rename" and then typing the name.

**Describe alternatives you've considered**
Maybe having an option in the settings to turn on/off some legacy features such as this one. I'm sure some people despise this manner of renaming so having it optional would be best.

**Additional context**
N/A

josh-65:
you can do this in version 1.4.1 not sure when the update will be out and not sure if it works with 1.3.

mtratchet16:
> you can do this in version 1.4.1 not sure when the update will be out and not sure if it works with 1.3.

Awesome, I'm still waiting on that update but glad it's been introduced!

yaichenbaum:
@mtratchet16 This feature has been included in Files since v1.0, what version of Files do you have?

mtratchet16:
> @mtratchet16 This feature has been included in Files since v1.0, what version of Files do you have?

I have 1.3, but clicking on the label (or double clicking) does not allow me to rename; I'm only able to via right-clicking.

yaichenbaum:
@mtratchet16 Do you have this issue in all the layout modes?

generalguy41:
#2976 👀

yaichenbaum:
Merging with #2976

Issue #2967, opened 2021-01-13
Double clicking on name of selected items always enters "rename" mode


Hefaistos68:
**Describe the bug**
When I select an item in the files list (details view), file or folder, and then double click on the name, instead of opening the file or opening the folder, it goes into rename mode where I can change the name. This does not happen when clicking anywhere else in the area of the selected item, only in the area of the "name" column.

**To Reproduce**
Open any folder location, switch to "Detail view", select a file or folder by single clicking it, then double click in the "Name" column area of selected item. 
The "Rename" mode will be initiated where you can alter the items name, instead of opening the item.

**Expected behavior**
I expect an item to open when double clicking it, selected or not.

**Desktop (please complete the following information):**
 - OS Version: Windows 10 Home, Version 20H2, OS Build 19042.746, Windows Feature Experience Pack 120.2212.551.0
 - App version: 1.0.0.0
 - Hardware:
Processor	Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz   1.99 GHz
Installed RAM	12.0 GB (11.9 GB usable)
System type	64-bit operating system, x64-based processor
Pen and touch	No pen or touch input is available for this display

**Additional context**
No change in context, misbehaviour is constant.

**Log file**
[debug_fulltrust.txt](https://github.com/files-community/Files/files/5808293/debug_fulltrust.txt)
[debug.txt](https://github.com/files-community/Files/files/5808294/debug.txt)

gave92:
@Don-Vito we might need to reintroduce the debounce before starting rename

Don-Vito:
> 
> 
> @Don-Vito we might need to reintroduce the debounce before starting rename

@gave92 - I hope we not :blush: Need to debug this. We have tested this scenario when applying the fix. Guess there is some sort of race that we need to understand.

gave92:
@Don-Vito not sure we tested this. The change was to fix the issue when you click outside the datagrid.
In this case you click once to select an item and then double click on the name. Since renaming starts at first tap and files open on double tap, renaming wins

Don-Vito:
> not sure we tested this. The change was to fix the issue when you click outside the datagrid.
> In this case you click once to select an item and then double click on the name. Since renaming starts at first tap and files open on double tap, renaming wins I guess :)

@gave92 -  Yeah.. it looks so.. I would expect the DataGrid to solve this for us. I am taking a look if we can somehow avoid it.. If not I will return the debouncer  😢

[...]

PR #5635, opened 2021-07-31
rename items in all layout views

xwnb:
**Resolved / Related Issues**
- Closes...
  - https://github.com/files-community/Files/issues/4153
- Related...
  - https://github.com/files-community/Files/issues/2976
  - https://github.com/files-community/Files/issues/2963
  - https://github.com/files-community/Files/issues/5600

**Details of Changes**
Add details of changes here.
- Added...
- Fixed...
- Fix
  - Fix item-rename in all layout reviews (detail, grid views, and etc.)

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

https://user-images.githubusercontent.com/45145954/127733002-0c9761f8-b47c-49a4-8c63-d30a10bb732e.mp4


gave92:
Thanks for this contribution, it works great! Only think it does not allow to open a file by double clicking on the filename if the file is already selected (that was the reason for those timers), but I think the same is currently broken on main too.

gave92 (approved):

yaichenbaum:
@xwnb Thank you!

yaichenbaum:
> Thanks for this contribution, it works great! Only think it does not allow to open a file by double clicking on the filename if the file is already selected (that was the reason for those timers), but I think the same is currently broken on main too.

@gave92 This is a pretty significant issue, can you open an issue for tracking?

xwnb:
> Thanks for this contribution, it works great! Only think it does not allow to open a file by double clicking on the filename if the file is already selected (that was the reason for those timers), but I think the same is currently broken on main too.

@gave92 @yair100 thanks for reviews. I missed this situation. Maybe we should delay `CheckRenameDoubleClick` execution. It would be easier and more stable to avoid this issue. I'm willing to solve the issue if tracked.

gave92:
Thanks @xwnb, issue is tracked in #2967.

PR #5655, opened 2021-08-02
Fix Path visibility

cinqmilleans:
**Resolved / Related Issues**
The control VisiblePath and ClickablePath are not visible when SearchBox is focused by CTRL+F.
This should only be done in compact mode, since the search bar is always displayed in normal mode. 

**Details of Changes**
The control VisiblePath and ClickablePath are visible when ShowSearchBox or not IsSearchBoxVisible.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before/After
PathBefore
PathAfter


gave92 (dismissed):

yaichenbaum (approved):
LGTM

PR #5672, opened 2021-08-03
await to check the double-click on selected item when renaming

### LINKED ISSUE #2967: Double clicking on name of selected items always enters "rename" mode

xwnb:
**Resolved / Related Issues**
- Closes #2967

**Details of Changes**
Add details of changes here.
- [x] Prevent rename when double-click on selected item (further commit based on #5635)

**Validation**
- [x] Built and ran the app


xwnb:
please help review it at your convenience. Thanks @gave92 .

gave92:
@xwnb I've pushed a couple of changes, I hope you don't mind. Changes:
- Fix an issue where double-clicking on an already selected file would open it but also start renaming
- Consider keyboard selection as a valid "first click" for triggering rename (like explorer does)

gave92 (approved):

xwnb:
It still starts renaming when double-click on a selected item.

xwnb:
Is the timer necessary? @gave92

xwnb:
I make a change for `IntervalSampler`.  `DateTime` will be influenced by the system date setting. Maybe `Stopwatch` is more reliable.

gave92:
I'd revert the last two commits because:
- 0010c86: causes double clicking on a selected item to also trigger rename
- 96a2393: I don't think it's worth it, user is not likely to change datetime settings every 500ms. Also changing `DateTime.Now` to `DateTime.UtcNow` would probably accomplish the same thing without having to run a timer.

xwnb:
https://github.com/files-community/Files/commit/0010c8650b56309b0fa0b55f1a100e574dfcd38a : Based on the previous commit, `preRenamingItem` will be reset when double-click to open a folder. when `Debounce()` delay invoked, there is a double-check `preRenamingItem`. I think it can work well on this case and be more stable and elegant than the Timer.

https://github.com/files-community/Files/commit/96a2393f8934e6de1baa3772c216b01c517b020e: Consider the situation: click item and then change the date time, not the time zone. I think the time measurement won't work well. That's why I perfer using Stopwatch to measure time if I need Timer or Samper.

gave92:
> 0010c86: I've tested it and I can confirm that the issue is present: double clicking a selected item will also start rename. That's because the double tapped event is fired _before_ the first item click event and resetting preRenaming item has no effect.

> 96a2393: could be a valid point but where IntervalSampler is used we don't need any particular precision

xwnb:
0010c86: It is a little different on my machine. I saw the double-tapped fired and then call `NavigationHelpers.OpenSelectedItems`, which then invoke ResetRenameDoubleClick to set preRenamingItem==null and the tapDebounceTimer.Stop(). Even though the action in `tapDebounceTimer.Debounce` is invoked after 500ms, the double-check will avoid the call of renaming.

https://user-images.githubusercontent.com/45145954/128384337-f914b370-d9cd-4084-8ba0-127bf9b858d2.mp4

96a2393: it can be reverted.

gave92:
This is what happens if I double click on a selected file: file opens but rename is also triggered.

ezgif com-gif-maker

xwnb:
👍

gave92:
b5ff4c2: reverted my change about setting preRenamingItem in SelectionChanged as it creates issues in ColumnView.

gave92 (approved):
Looks like it's working pretty well, thank you!

PR #5717, opened 2021-08-08
Fix file thumbnail cropping and rounded thumbnail corners

winston-de:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #5682
- Closes #5631 

**Details of Changes**
Add details of changes here.
- Switch to single item thumbnail mode and change resize mode
- Round corners for grid view thumbnails
- Show proper file preview images in preview pane and grid view
- Fix issue where file name would sometimes be cut off in tiles view

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [ ] Tested the changes for accessibility

**Screenshots (optional)**
image

PR #5734, opened 2021-08-09
Update font style 

### LINKED ISSUE #5730: Disc Number is not displayed in properties of Music file
### Description

The Disc Number property is not being displayed in Music file properties.

### Steps To Reproduce

1. Right click on a Music file 
2. Click on "Properties" in the context menu
3. Check file details, Disc Number does not show up

### Expected behavior

I expect to see the disc number in the file properties.

### Files Version

v1.9.20.0

### Windows Version

Windows 11 21H2 22000.120

### Relevant Assets

image

This is what shows up in the file properties

winston-de:
**Resolved / Related Issues**
- Closes #5730
- https://github.com/files-community/Files/pull/5717#discussion_r684726948

**Details of Changes**
Add details of changes here.
- Updated font style in details layout, grid view, and preview pane to use caption text style
- added file tag to preview pane
- Fix issue where right clicking text in rename box would commit rename
- add part of set property
- fix issue where tab item would not be selected on load
- fix alignment issue with tags in tiles view

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

### DIFF
--- a/Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
+++ b/Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
@@ -4,6 +4,7 @@ using Microsoft.Toolkit.Uwp;
 using Microsoft.Toolkit.Uwp.UI;
 using Microsoft.UI.Xaml.Controls;
 using System;
+using System.Diagnostics;
 using System.Linq;
 using Windows.ApplicationModel.DataTransfer;
 using Windows.Storage;
@@ -224,6 +225,16 @@ namespace Files.UserControls.MultitaskingControl
         public static readonly DependencyProperty TabStripVisibilityProperty =
             DependencyProperty.Register("TabStripVisibility", typeof(Visibility), typeof(HorizontalMultitaskingControl), new PropertyMetadata(Visibility.Visible));
 
+        private bool firstLoad = true;
 
+        private void TabViewItem_Loaded(object sender, RoutedEventArgs e)
+        {
+            // fixes issue where tab would not show as selected when opened with path argument
+            if(firstLoad)
+            {
+                firstLoad = false;
+                (sender as TabViewItem).IsSelected = true;
+            }
+        }
     }
 }
--- a/Files/ViewModels/ItemViewModel.cs
+++ b/Files/ViewModels/ItemViewModel.cs
@@ -802,7 +802,9 @@ namespace Files.ViewModels
                         var matchingStorageFolder = (StorageFolder)matchingStorageItem ?? await GetFolderFromPathAsync(item.ItemPath);
                         if (matchingStorageFolder != null)
                         {
-                            using var Thumbnail = await matchingStorageFolder.GetThumbnailAsync(ThumbnailMode.ListView, thumbnailSize, ThumbnailOptions.ReturnOnlyIfCached);
+                            var mode = thumbnailSize < 80 ? ThumbnailMode.ListView : ThumbnailMode.SingleItem;
+
+                            using var Thumbnail = await matchingStorageFolder.GetThumbnailAsync(mode, thumbnailSize, ThumbnailOptions.UseCurrentScale);
                             if (!(Thumbnail == null || Thumbnail.Size == 0 || Thumbnail.OriginalHeight == 0 || Thumbnail.OriginalWidth == 0))
                             {
                                 await CoreApplication.MainView.DispatcherQueue.EnqueueAsync(async () =>
--- a/Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
+++ b/Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
@@ -150,6 +150,16 @@
             <Style TargetType="uc:DataGridHeader">
                 <Setter Property="HorizontalContentAlignment" Value="Stretch" />
             </Style>
+
+            <Style
+                x:Key="ColumnContentTextBlock"
+                BasedOn="{StaticResource CaptionTextBlockStyle}"
+                TargetType="TextBlock">
+                <Setter Property="TextWrapping" Value="NoWrap" />
+                <Setter Property="TextTrimming" Value="CharacterEllipsis" />
+                <Setter Property="VerticalAlignment" Value="Center" />
+                <Setter Property="HorizontalAlignment" Value="Stretch" />
+            </Style>
         </ResourceDictionary>
     </local:BaseLayout.Resources>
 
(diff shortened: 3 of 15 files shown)


Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
```diff
@@ -224,6 +225,16 @@ public Visibility TabStripVisibility
...
 
+        private bool firstLoad = true;
 
+        private void TabViewItem_Loaded(object sender, RoutedEventArgs e)
+        {
+            // fixes issue where tab would not show as selected when opened with path argument
+            if(firstLoad)
+            {
+                firstLoad = false;
+                (sender as TabViewItem).IsSelected = true;
```
gave92, line 236: Would `(sender as TabViewItem).Loaded -= TabViewItem_Loaded` work instead of using the `firstLoad` thing?

Files/ViewModels/ItemViewModel.cs
```diff
@@ -816,7 +816,9 @@ private async Task LoadItemThumbnail(ListedItem item, uint thumbnailSize = 20, I
                         var matchingStorageFolder = (StorageFolder)matchingStorageItem ?? await GetFolderFromPathAsync(item.ItemPath);
                         if (matchingStorageFolder != null)
                         {
-                            using var Thumbnail = await matchingStorageFolder.GetThumbnailAsync(ThumbnailMode.ListView, thumbnailSize, ThumbnailOptions.ReturnOnlyIfCached);
+                            var mode = thumbnailSize < 80 ? ThumbnailMode.ListView : ThumbnailMode.SingleItem;
+
+                            using var Thumbnail = await matchingStorageFolder.GetThumbnailAsync(mode, thumbnailSize, ThumbnailOptions.UseCurrentScale);
```
gave92, line 821: Completely unrelated: could someone explain me the difference between `UseCurrentScale` and `ResizeThumbail` options? When should I use one or the other?

Files/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
winston-de: I don't think so, the event would have to be removed from all TabViewItems since the event is in the template.

gave92: Sorry I had misunderstood your intentions here.

Files/ViewModels/ItemViewModel.cs
winston-de: My understanding is that resize thumbnail resizes the thumbnail (up or down) to match the requested size, while increase size just scales up the size to match the scaling of the display.

gave92: So `UseCurrentScale` should be used all the times?

winston-de: Yeah I don't think it would make much of a difference.

gave92: Thanks. I believe we do the opposite in most other places xD

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -815,18 +801,26 @@
...
                                                                 <ItemsPresenter
                                                                     Padding="{TemplateBinding Padding}"
                                                                     HorizontalAlignment="Stretch"
-                                                                    Footer="{TemplateBinding Footer}"
                                                                     FooterTemplate="{TemplateBinding FooterTemplate}"
                                                                     FooterTransitions="{TemplateBinding FooterTransitions}"
                                                                     HeaderTemplate="{TemplateBinding HeaderTemplate}"
                                                                     HeaderTransitions="{TemplateBinding HeaderTransitions}" />
+
+                                                                <ContentPresenter
```
gave92, line 820: Why the change? The footer now appears left aligned instead of center-aligned, was it intentional?

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -150,6 +150,16 @@
...
             </Style>
+
+            <Style
+                x:Key="ColumnContentTextBlock"
+                BasedOn="{StaticResource CaptionTextBlockStyle}"
+                TargetType="TextBlock">
+                <Setter Property="TextWrapping" Value="NoWrap" />
+                <Setter Property="TextTrimming" Value="CharacterEllipsis" />
+                <Setter Property="VerticalAlignment" Value="Center" />
+                <Setter Property="HorizontalAlignment" Value="Stretch" />
```
gave92, line 161: What about keeping the `FontWeight="Light"` or add `Opacity="0.6"` like in TilesView?

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -150,6 +150,16 @@
             <Style TargetType="uc:DataGridHeader">
                 <Setter Property="HorizontalContentAlignment" Value="Stretch" />
             </Style>
+
+            <Style
+                x:Key="ColumnContentTextBlock"
+                BasedOn="{StaticResource CaptionTextBlockStyle}"
```
gave92, line 156: I'm not sold to this change in DetailsView: CaptionTextBlockStyle looks _really_ small and seeing two different font sizes on the same row does not look good to me, but I will bow to the designers opinion.. @yair100? xD

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
winston-de: Should be fixed in 649e004

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
winston-de: Windows 11 introduced Segoe UI variable font which is much more legible at small sizes then old segoe ui. This change is intended to leverage that. Perhaps the old style should be used on windows 10? 

Variable vs standard at size 12:
image

yaichenbaum: @gave92 I don't have any preference here so I think we should probably use what the design specs suggest.

Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
winston-de: Good idea

winston-de: image

gave92 (approved):
LGTM

PR #5769, opened 2021-08-11
Do not read item path in narrator

### LINKED ISSUE #5079: a11y: extraneous information is read in file/folder list
Hello.

When using screenreader, and navigating files and follders list, information is announced in the following way.

fsdn, C:\Users\beqap\Downloads\fsdn, Folder

if folder has some sort of needed value, it cannot be told the same about path.
because of that, navigation becomes extremely slow.

It would be great if you rethink display behaviour of such information for screenreaders and do it for example in this way.
fsdn, Folder
or maybe better, as it is in total commander for example
fsdn, <DIR>
And remove File from type of files, as we already know that folders have identification sign.

Thanks.

### LINKED ISSUE #5766: Adresses of folders to be opened at startup are cut off
### Description

image

The adresses are cut off due to the limited width of the UI and the lack of a tooltip showing the full path

Addtionally, there is no way to remove an adress from that list

### Steps To Reproduce

_No response_

### Expected behavior

- A button to remove an entry
- Increased UI width and/or a tooltip that shows the full path

### Files Version

Version: 1.9.21.0 

### Windows Version

OS Version: 10.0.19043.1151

### Relevant Assets

_No response_

gave92:
**Resolved / Related Issues**
- Closes #5079
- Closes #5766

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Filesystem/ListedItem.cs
+++ b/Files/Filesystem/ListedItem.cs
@@ -323,7 +323,7 @@ namespace Files.Filesystem
             {
                 suffix = PrimaryItemAttribute == StorageItemTypes.File ? "FileItemAutomation".GetLocalized() : "FolderItemAutomation".GetLocalized();
             }
-            return $"{ItemName}, {ItemPath}, {suffix}";
+            return $"{ItemName}, {suffix}";
         }
 
         public bool IsRecycleBinItem => this is RecycleBinItem;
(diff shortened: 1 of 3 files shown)


yaichenbaum (approved):
Great work! LGTM

PR #5782, opened 2021-08-11
Dialog improvements

### LINKED ISSUE #5729: [UX] Autofocus TextBoxes in action dialogs where they are the primary content.
### What's the Problem?

One slight annoyance i've had as of recent is that the new folder/file dialogs don't provide a way to fill in a text field immediately. Rather, they require mouse input or additional tabbing to focus the textbox which is required to complete the action. I wish there was a way to just immediately start typing, then complete the action with no additional input.

### Solution/Idea

Autofocus the TextBox in dialogs where they are the primary focus of the action (eg. providing a name for a file or folder). Additionally, (this might be better suited for the WinUI repo) it might also improve UX to perform the primary action of the dialog when the user hits enter on a single-line textbox to avoid another additional tab input or mouse click (similar to how discord handles their modal components).

### Alternatives

I suppose that a possible alternative would be getting rid of some modals entirely in place for a similar UI to what file explorer uses now, but I don't endorse that personally.

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow users to immediately enter a file name in the "New <x>" dialogs. | Must |

### Files Version

1.9.20.0

### Windows Version

Windows 11 22000.120

### Additional comment

https://user-images.githubusercontent.com/42101043/128645578-3c6d9e9c-6839-4560-8105-145f69625e28.mp4

winston-de:
**Resolved / Related Issues**
- Closes #5729

**Details of Changes**
Add details of changes here.
- Autofocus text box for new item dialog
- Add apply to all right click option in filesystem dialog to apply the selected option to all items
- improve filesystem dialog right click reliability and keyboard accessibility by moving context flyout to the item container
- allow pressing continue even if not all conflicts have been resolved, and apply the default option for all items that have not had an option applied to reduce amount of clicks needed

image

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Dialogs/FilesystemOperationDialog.xaml.cs
+++ b/Files/Dialogs/FilesystemOperationDialog.xaml.cs
@@ -65,7 +65,7 @@ namespace Files.Dialogs
                 }
             }
 
-            if (DetailsGrid.SelectedItems.Count == 1 && DetailsGrid.SelectedItems.Any(x => (x as FilesystemOperationItemViewModel).ActionTaken))
+            if (DetailsGrid.Items.Count > 1 && DetailsGrid.SelectedItems.Count == 1 && DetailsGrid.SelectedItems.Any(x => (x as FilesystemOperationItemViewModel).ActionTaken))
             {
                 ApplyToAllOption.Visibility = Windows.UI.Xaml.Visibility.Visible;
                 ApplyToAllSeparator.Visibility = Windows.UI.Xaml.Visibility.Visible;
--- a/Files/Dialogs/FilesystemOperationDialog.xaml.cs
+++ b/Files/Dialogs/FilesystemOperationDialog.xaml.cs
@@ -1,6 +1,7 @@
 ﻿using Files.Enums;
 using Files.ViewModels.Dialogs;
 using System.Collections.Generic;
+using System.Linq;
 using Windows.UI.Xaml.Controls;
 
 // The Content Dialog item template is documented at https://go.microsoft.com/fwlink/?LinkId=234238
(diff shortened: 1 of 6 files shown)


yaichenbaum:
@winston-de what action does "apply to all" do?

Files/Strings/en-US/Resources.resw
```diff
@@ -2740,6 +2740,8 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
   <data name="SettingsShowFavoritesSection.Title" xml:space="preserve">
     <value>Show Favorites section on the sidebar</value>
   </data>
+  <data name="ConfictingItemsDialogOptionApplyToAll.Text" xml:space="preserve">
+    <value>Apply to all</value>
```
BanCrash, line 2744:
```suggestion
    <value>Apply to all</value>
   </data>
```

winston-de: Thanks!

Files/Dialogs/FilesystemOperationDialog.xaml.cs
```diff
@@ -23,6 +26,65 @@ public FilesystemOperationDialog(FilesystemOperationDialogViewModel viewModel)
             ViewModel = viewModel;
             ViewModel.View = this;
         }
+
+        private void MenuFlyoutItem_Click(object sender, Windows.UI.Xaml.RoutedEventArgs e)
+        {
```
d2dyno1, line 31: Why isn't it done in View Model 🤨 😄

winston-de: Because the code in the faction has to work directly with the ui, mainly ``DetailsGrid``.

BanCrash:
@winston-de should the "apply to all" be used only for items that doesn't have yet a selected option, or should be applied for all items overriding their already selected option?

Currently it works as the first way, which personally I think is good, but I'm not sure if that is the expected behaviour of the "apply to all" button.

GIF: When I selected "apply to all" when all items have already an option selected, their option doesn't change to the new.

GIF 13-08-2021 12-27-59

Files/Dialogs/FilesystemOperationDialog.xaml.cs
```diff
@@ -23,6 +26,65 @@ public FilesystemOperationDialog(FilesystemOperationDialogViewModel viewModel)
...
+
+            if (((sender as MenuFlyout)?.Target as ListViewItem)?.Content is FilesystemOperationItemViewModel li)
+            {
+                if(!DetailsGrid.SelectedItems.Contains(li))
+                {
+                    DetailsGrid.SelectedItems.Add(li);
+                }
+            }
+
+            if (DetailsGrid.SelectedItems.Count == 1 && DetailsGrid.SelectedItems.Any(x => (x as FilesystemOperationItemViewModel).ActionTaken))
```
BanCrash, line 68: @winston-de what do you think about adding this to not giving the "Apply to all" option when only copying one item?:

```suggestion
            if (DetailsGrid.Items.Count > 1 && DetailsGrid.SelectedItems.Count == 1 && DetailsGrid.SelectedItems.Any(x => (x as FilesystemOperationItemViewModel).ActionTaken))
```

BanCrash (approved):
LGTM. This will be useful, thanks!

yaichenbaum:
Thank you!

PR #5789, opened 2021-08-12
Hebrew translation updates

zvizvi:
- Hebrew translation updates

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/MultilingualResources/Files.he-IL.xlf
+++ b/Files/MultilingualResources/Files.he-IL.xlf
@@ -348,7 +348,7 @@
         </trans-unit>
         <trans-unit id="FileFolderListItem" translate="yes" xml:space="preserve">
           <source>File folder</source>
-          <target state="translated" state-qualifier="tm-suggestion">‏‏תיקיית קבצים</target>
+          <target state="translated" state-qualifier="tm-suggestion">תיקיית קבצים</target>
         </trans-unit>
         <trans-unit id="ItemTypeFile" translate="yes" xml:space="preserve">
           <source>File</source>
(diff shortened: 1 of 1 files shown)


Files/MultilingualResources/Files.he-IL.xlf
```diff
@@ -1679,7 +1679,7 @@
         </trans-unit>
         <trans-unit id="NavToolbarDisplayOptions.Text" translate="yes" xml:space="preserve">
           <source>Display Options</source>
-          <target state="needs-review-translation" state-qualifier="tm-suggestion">אפשרויות &amp;תצוגה</target>
+          <target state="needs-review-translation" state-qualifier="tm-suggestion">אפשרויות תצוגה</target>
```
BanCrash, line 1682: It's fine to use needs review state when you want to leave a string so others (or you in the future) could check if it's okay if you are not totally confident about the result, but if you are confident about it the state should be translated. 

So, the question is, are the use of needs review intentional or should be changed to translated? This applies to all the needs review strings.

zvizvi: It is the right translation, so i will change the state to "translated".

BanCrash: > It is the right translation, so i will change the state to "translated".

Fine then! Don't forget to do the same on the rest if they are correct too:

Lines 1714, 1718, 1902, 1910, 1946, 2030 and 2042.

BanCrash (approved):
LGTM. Thanks!

PR #5844, opened 2021-08-16
Added Windows 10 style

### LINKED ISSUE #5293: Seletected Items Should Show More Distinctions
A clear description of what the problem is.

> Selecting folder or/and files should be more distinctive, in light mode, grey shades almost doesn't have contrast background ​

> image

**Describe the solution**
A clear description of what you want.

> Using Windows accent color would be a good solution similar to native file manager
> image

A clear description of any alternative solutions or features you've considered.

**Scope**

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow end users to accomplish better UX | Could |

**System Information**
​- OS Version: [Windows 10 21H1 19043.1052]
​- App version: [1.5.46]

BanCrash:
**Resolved / Related Issues**
- Closes #5293

**Details of Changes**
- Added Windows 10 style without overriding Windows 11 style.

Why adding to the code instead just creating a theme? -> Because with a theme, this will only apply to the theme at hand, while with this change there is no need to have for example "Black pixel - Windows 10 version" because it will be inherited by all themes.

This shouldn't have any impact on Windows 11 and any impact should be treated as a bug.

Added as awaited because this need to be loaded before the theme itself is loaded.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Selected item on list view:
- Before:
image

- After:
image

Selected item on grid view:
- Before:
image

- After:
image


Files/Helpers/ExternalResourcesHelper.cs
```diff
@@ -25,6 +25,20 @@ public class ExternalResourcesHelper
...
+            int majorVersion = osVersion.Version.Major;
+
+            if (majorVersion == 10) // Windows 10 style
+            {
+                var windows10Style = new ResourceDictionary() { Source = new Uri("ms-appx:///ResourceDictionaries/Windows10Style.xaml") };
+                Application.Current.Resources.MergedDictionaries.Add(windows10Style);
+            }
+
+            return Task.CompletedTask;
+        }
```
yaichenbaum, line 40: Should we load this theme no matter the Windows version?

BanCrash: If you mean about Windows 10 version, I think there is no Windows 10 version that have the "after" screenshot instead the "before".

If you mean about Windows 11, this shouldn't be loaded to Windows 11, and I think that this way it won't.

yaichenbaum (approved):
Great work!

PR #5861, opened 2021-08-18
Open only directories in new tab when OpenFoldersNewTab is enabled

gave92:
**Resolved / Related Issues**
- Currently when the OpenFoldersNewTab option is enabled, double-clicking a file results in a new tab opening. A new tab should be opened only if the selected item is a folder (or a shortcut to a folder).

**Details of Changes**
Add details of changes here.
- Fixes an issue with the OpenFoldersNewTab setting
- Refactor OpenPath function into more manageable OpenLibrary, OpenDirectory, OpenFile

**Validation**
- [x] Built and ran the app


BanCrash (approved):
LGTM

Issue #5376, opened 2021-07-05
Some translated strings won't fit on their place


BanCrash:
I have translated the remaining lines of ES-ES (Spain spanish) and I found that in some places those strings got cut.

Security properties:
![image](https://user-images.githubusercontent.com/11770760/129962301-bcced470-0214-41c7-942c-cccc2bb9e27f.png)
Should show "Mostrar el contenido de la carpeta". Since this string is copied literally from Windows explorer security properties, I think this string should stay untouched.

Easy solution is to add more space to make it fit.

EDIT: Solved issues:

<del>Another on about:</del>

![image](https://user-images.githubusercontent.com/11770760/129962249-9d98a6fe-6def-48cc-bb9a-81a2f33bb39e.png)

<del>Should show "Política de privacidad".</del>

<del>Besides, I found another incomplete string on settings, on preferences:</del>

![image](https://user-images.githubusercontent.com/11770760/125179065-fd1c0180-e1ea-11eb-99d9-70d1f02771d6.png)

<del>Settings:
![image](https://user-images.githubusercontent.com/11770760/124503315-98fdd580-ddc5-11eb-9197-e08d4f6f2935.png)
Should show "Archivos y carpetas". Not much more space needed, but a little.
</del>

yaichenbaum:
@BanCrash is this still an issue?

BanCrash:
> @BanCrash is this still an issue?

Yes, but only partially. I've updated the original post to clarify what is solved and what is still an issue.

BanCrash:
Closing this since it's fixed on last release.

PR #5874, opened 2021-08-20
Improved width of settings in the setting dialog

yaichenbaum:


Files/Strings/en-US/Resources.resw
```diff
@@ -2794,4 +2791,10 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
   <data name="NoSearchResultsFound" xml:space="preserve">
     <value>No items found</value>
   </data>
-</root>
+  <data name="OpenLogLocation.Title" xml:space="preserve">
+    <value>Open log location</value>
+  </data>
+  <data name="OpenLogLocation.ToolTipService.ToolTip" xml:space="preserve">
+    <value>Open log location</value>
+  </data>
```
BanCrash, line 2799:
```suggestion
  </data>
  <data name="SettingsAboutOpenLogLocationButton.ToolTipService.ToolTip" xml:space="preserve">
    <value>Open log location</value>
  </data>
```

For this: (on my testing version I added the 3 to check it)
image

BanCrash (changes requested):
This change is not a big deal, but since it's easy to add this string as translatable I think could be added within this PR.

BanCrash (approved):
LGTM. I've just updated #5376 to remove the resolved issue by this PR, thanks!

Issue #5833, opened 2021-08-15
Improving FTP support


gave92:
### Description

This issue is to track bugs and enhancements for the new FTP support.

- [ ] Progress support for file transfers
- [ ] Remember credentials
- [ ] Double click downloads to TEMP and opens file
- [ ] Move operation
- [ ] Copy and create folders
- [ ] Handle copy conflicts
- [ ] Autorefresh folder after changes

### Expected behavior

aa

### Files Version

Main (1.9.17)

### Windows Version

Versione 1.9.0.0 OS Version: 10.0.19043.1110

josh-65:
Maybe also add a startup first time tip for how to connect to an FTP server. Isn't very clear as it is in 1.9.24

josh-65:
Should also allow saving ftp servers to the network drives section and pin folders with in the ftp server to the favorites section

gave92:
Thanks for the feedback.

> saving FTP addresses to the network drives

You should be already able to do that from the "map Network drive" dialog

> pin folders with in the FTP server to the favorites section

You should be already able to do that, don't you have the option in the context menu?

> Better icons support like folder icons

It should be already showing the windows icons, do you only see the monochrome icons? Sounds like a bug. Does the ftp server you're using require credentials?

josh-65:
No it in anonymous mode

josh-65:
When adding, removing or renaming an FTP connection the side bar doesn't update until files reloaded.

hez2010:
Path bar doesn't recognize `ftps://`

hez2010:
Some FTP servers are providing different contents depending on credentials, we should add the ability to login without using anonymous.

hez2010:
Also, we should make it possible to login to the same server but with different credentials in different tabs. The current credential saving logic is not in good fit IMO. (we should cache credentials for per-tab, not per server host)

yaira2:
@gave92 is this list up to date?

gave92:
Yeah should be

lukeblevins:
@yair100 Is this a bug or feature issue?

yaira2:
@lukeblevins it's a feature but we should probably split it into several issues.

Issue #45, opened 2019-02-17
Feature: Improve support for compressing and decompressing archives


yaichenbaum:
- [ ] #9947
- [ ] Keep timestamps from the .zip file
- [x] Extract files
- [x] Menu options to extract items quickly without the use of the dialog
- [x] #9948
- [x] When pasting a file into an archive the file is fully currently loaded into memory first (bad)
- [x] "Extract" command on toolbar when an archive is selected or being browsed #7639
- [x] 7zip & other formats support (requires switching to SharpCompress or something)
- [x] Specific icon to show when zip files are associated with Files?
- [ ] Add shell extensions to File Explorer
- [ ] Add dedicated progress view when using shell extensions in File Explorer
- [ ] #9949
- [x] 7z
- [ ] gz
- [x] tar
- [x] rar
- [x] zip
- [ ] lzma
- [ ] lzh
- [ ] arj
- [ ] bz2
- [ ] cab
- [ ] chm
- [ ] deb
- [ ] iso (known to have issues with files displaying)
- [ ] rpm
- [ ] wim
- [ ] udf
- [ ] mub
- [ ] xar
- [ ] hfs
- [ ] dmg
- [ ] Z
- [ ] xz
- [ ] flv
- [ ] swf
- [ ] exe
- [ ] dll
- [ ] vhd
- [x] msix
- [x] appx

yaichenbaum:
Take a look at https://github.com/adamhathcock/sharpcompress

szaimen:
Maybe also looking inside zip files and opening files from there should be supported?
(it is working like that currently in the file explorer)

ChihweiLHBird:
Why not just use https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.zipfile?view=netcore-3.1 ?

chuchanpr:
This is a recommend to be added to Files UWP, when possible

Jaiganeshkumaran:
@ChihweiLHBird Using System.IO will make it slower because UWP apps have limited access and you need to use StorageItems to get them. Instead we can use the native API: https://docs.microsoft.com/en-us/uwp/api/Windows.Storage.Compression?view=winrt-19041

duke7553:
@Jaiganeshkumaran Does the native API support folder items? In my analysis, it appears support for that is lacking, but I'm still unsure.

Jaiganeshkumaran:
@duke7553 It seems to not support by default but System.IO should be sufficient

[...]

PR #5885, opened 2021-08-21
Browse zip archives & other things

gave92:
**Resolved / Related Issues**
- For #45
- For #5833
- For #4180

**Details of Changes**
Add details of changes here.
Replaced most of references to StorageFile and StorageFolder with new base classes that implements IStorageItem (BaseStorageFile[Folder]). The standard StorageFile[Folder], the FtpStorageFile[Folder] and the new ZipStorageFile[Folder] all derive from these base classes. This enables the following features "for free".

**Features enabled by this PR**
- Browse zip archives as folders
- Add files to zip by drag & drop
- Delete from zip
- Open FTP files with double click
- Search and navigation bar suggestions in FTP folders
- Copy folders to/from FTP
- Preview files in zip and ftp folders

Comments, testing and contributions are welcome xD

**Validation**
- [x] Built and ran the app


winston-de:
@gave92 do you mind if I implement making files a handler for zip files to the pr (so that it can be set as the default app for archives)?

Files/ViewModels/SettingsViewModel.cs
```diff
@@ -81,6 +81,7 @@ private void TrackAnalytics()
             Analytics.TrackEvent($"{nameof(ShowLibrarySection)} {ShowLibrarySection}");
             Analytics.TrackEvent($"{nameof(ShowBundlesWidget)} {ShowBundlesWidget}");
             Analytics.TrackEvent($"{nameof(ListAndSortDirectoriesAlongsideFiles)} {ListAndSortDirectoriesAlongsideFiles}");
+            Analytics.TrackEvent($"{nameof(AreFileTagsEnabled)} {AreFileTagsEnabled}");
```
yaichenbaum, line 84: Should this be changed to `AreTagsEnabled` since tags work for folders as well?

gave92:
> @gave92 do you mind if I implement making files a handler for zip files to the pr (so that it can be set as the default app for archives)?

That's a great idea, please do 👍

Files/ViewModels/SettingsViewModel.cs
gave92: I'd leave the variables name as they are, if any we could change the user-facing string resources to something different than "file tags", but just "tags" doesn't sound clear 🤔

yaichenbaum: I'm not sure file tags is much clearer 🤔. Perhaps we should add information to the docs or a teaching tip to the app.

winston-de:
Love it, just a couple of minor things I noticed:

- dragging a file from a zip to another directory moves it, should it copy instead?
- changes to zip files don't seem to be detected, so operations like delete don't show until refreshing

gave92:
> * dragging a file from a zip to another directory moves it, should it copy instead?

True, the default action should be copy. There's also another issue for which dragging a file _into_ a zip archive defaults to move (which fails), default should be copy here as well.

> * changes to zip files don't seem to be detected, so operations like delete don't show until refreshing

True, it's the same for FTP, I don't think I'm adding either here

d2dyno1:
Wow! Excellent job @gave92!

winston-de:
I implemented making copy the default operation when dragging a file from zip over a tab by using the same function for dropping into the layout blank space (also reduces code duplication), is that good with you?

gave92:
I hate CodeMaid lol

winston-de (approved):

winston-de:
Amazing

Files/Filesystem/StorageItems/FtpStorageFile.cs
```diff
@@ -8,47 +8,59 @@
...
 
         private async void FtpDataStreamingHandler(StreamedFileDataRequest request)
         {
             try
             {
-                var ftpClient = _viewModel.GetFtpInstance();
+                using var ftpClient = new FtpClient();
+                ftpClient.Host = FtpHelpers.GetFtpHost(Path);
+                ftpClient.Port = FtpHelpers.GetFtpPort(Path);
+                ftpClient.Credentials = FtpManager.Credentials.Get(ftpClient.Host, FtpManager.Anonymous);
```
hez2010, line 51: I think this should also be distinguished by port.

Files/ViewModels/Previews/TextPreviewViewModel.cs
```diff
@@ -44,8 +44,8 @@ public static async Task<TextPreview> TryLoadAsTextAsync(ListedItem item)
 
             try
             {
-                item.ItemFile = await StorageFile.GetFileFromPathAsync(item.ItemPath);
-                var text = await FileIO.ReadTextAsync(item.ItemFile);
+                item.ItemFile = await StorageFileExtensions.DangerousGetFileFromPathAsync(item.ItemPath);
+                var text = await ReadFileAsText(item.ItemFile); // await FileIO.ReadTextAsync(item.ItemFile);
```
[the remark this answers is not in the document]
d2dyno1: Also while at it, `StorageFileExtensions.DangerousGetFileFromPathAsync` could be replaced with safer `StorageItemHelpers.ToStorageItem<StorageFile>(item.Path);`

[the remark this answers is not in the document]
gave92: @winston-de `FileIO.ReadTextAsync` did not work for some reason on FTP but if it causes issues with normal previews and can't be easily fixed we can live without previews on FTP paths.
@d2dyno1 good point. (Note: it's `StorageItemHelpers.ToStorageItem<BaseStorageFile>(...)` now)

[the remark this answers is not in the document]
gave92: Ah well just seen why `ReadFileAsText` always contains "\0\0\0\0", uops.

PR #5889, opened 2021-08-22
Preview pane design update and fixes

### LINKED ISSUE #5883: Switching from Column View to any other type of view unexpectedly changes the File Path
### Description

When you have a Column View tab with more than one column displayed, switching from Column View to any other type of view causes the displayed folder to be the first column in the previous view, NOT the last directory in the Column View in the hierarchy and the one the File Path Bar points to.

### Steps To Reproduce

1) Open a tab with a Column View.
2) Open folders using the Column View so at least one additional column is revealed.
3) Switch from Column View to any other view.

Screenshot 2021-08-21 123607
Screenshot 2021-08-21 123638

(pictures display a Column View tab, and what happens when the view is switched)

### Expected behavior

The expected behavior is that the new view displays the directory shown on the File Path Bar before the view is switched.

### Files Version

Version: 1.9.25.0

### Windows Version

Windows 11 Pro 21H2 22000.160

### Relevant Assets

_No response_

### LINKED ISSUE #5886: Column View doesn't show the "active" selected item in the Preview Pane
### Description

This issue also relates to a previous design issue I've raised (https://github.com/files-community/Files/issues/5880) as it's not clear in Column View what the active selection is.

---

While it's clear Files has an internal indication on what the "active" and selected items are as QuickLook and Delete functions work off this, but that isn't reflected in the Preview Pane which will only show:

1) Items selected in the first column.
2) The open folder of the first column.

### Steps To Reproduce

Select any item not in the first column with the Preview Pane open.

Screenshot 2021-08-21 145016

### Expected behavior

The Preview Pane should show the most recently selected item no matter what column it is in.

### Files Version

Version: 1.9.25.0

### Windows Version

Windows 11 Pro 21H2 22000.160

### Relevant Assets

_No response_

winston-de:
**Resolved / Related Issues**
- Closes #5886
- closes #5883

**Details of Changes**
Add details of changes here.
- preview pane design updates
- add option to only show preview 
- fix issue where some colored icons wouldn't display in some languages
- fix issue where unpin from start option could be shown unexpectedly
- fix issue where preview pane and status bar wouldn't work properly in column view
- fix issue where BaseLayout would be navigated to twice when opening a column
- fix some navtoolbar stuff with column view

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image
image


Files/Helpers/ContextFlyoutItemHelper.cs
```diff
@@ -454,12 +454,12 @@ public static List<ContextMenuFlyoutItemViewModel> GetBaseLayoutMenuItems(Curren
                 new ContextMenuFlyoutItemViewModel()
                 {
                     Text = "BaseLayoutContextFlyoutPaste/Text".GetLocalized(),
-                    //Glyph = "\uE16D",
+                    //Glyph = "\uF16D",
```
yaichenbaum, line 457:
```suggestion
                    // Glyph = "\uF16D",
```

Files/ViewModels/SettingsViewModel.cs
```diff
@@ -140,6 +140,15 @@ public bool PreviewPaneEnabled
             set => Set(value);
         }
 
+        /// <summary>
+        /// True if only the file preview in the preview pane should be shown
```
yaichenbaum, line 144:
```suggestion
        /// Gets or sets a value indicating if the preview pane should only show the item preview without the details section
```

yair100 (dismissed):

Files/ViewModels/PreviewPaneViewModel.cs
```diff
@@ -65,6 +66,7 @@ public UIElement PreviewPaneContent
 
         public PreviewPaneViewModel()
         {
+            App.AppSettings.PropertyChanged += AppSettings_PropertyChanged;
```
gave92, line 69: Don't think in makes much of a difference but I added a Dispose method to unsubscribe this event.

gave92:
@winston-de  help! It looks like I can't get out of ColumnView anymore 🤣.. 8eb2444c83cf805b27024285a3cefe9c5433e3a0?

Files/Views/LayoutModes/ColumnViewBrowser.xaml.cs
```diff
@@ -768,10 +774,33 @@ private void FileList_ContainerContentChanging(ListViewBase sender, ContainerCon
             }
         }
 
+        public void SetSelectedPathOrNavigate(PathNavigationEventArgs e)
+        {
+            var p = e.ItemPath;
```
gave92, line 766: var p = e.ItemPath.TrimEnd("\\") ;)

winston-de: Thanks!

gave92:
Lol looks like you got sucked in the ColumnView black hole ;)

gave92:
Nice job, we are still far from having ColumnView working properly though:
- Pathbox suggestions don't work
- Clicking on pathbox items doesn't work
- Getting out of columnview requires changing layoutmode twice
- ....

Do you wish to address something else in this PR? If not it looks good.

I'm sorry to see the "show preview only" option go away, for PDFs for e.g the preview is really small. Would you consider to implement the feature as follows (notice the collapse button)?
Screenshot 2021-08-24 005807

winston-de:
> Nice job, we are still far from having ColumnView working properly though:
> 
> * Pathbox suggestions don't work
> * Clicking on pathbox items doesn't work
> * Getting out of columnview requires changing layoutmode twice
> * Those `public static event`s (@5879)
> * ....

At least some of these should be somewhat simple, I'll try implementing them.

> Would you consider to implement the feature as follows (notice the collapse button)?

yeah sure, I didn't think there was demand for this. Also an expand button is a really good idea.

yaichenbaum:
Instead of a collapse details button, what about clicking on the grid splitter to automatically collapse the details section (after it's collapsed, an expand button can be shown in case the user want's to view the details again)?

mdtauk:
What about making the preview image clickable, to show and hide the details?

yaichenbaum:
> What about making the preview image clickable, to show and hide the details?

That's also an option, and we could probably use the expand button with that design as well to show the details section.

winston-de:
I though the grid splitter was supposed to be removed?

also how about something like this?

https://user-images.githubusercontent.com/59544401/130556806-351bf569-bb61-4f91-a842-b5a2e5e5f377.mp4

mdtauk:
The expand button is not the most elegant of options IMO.

winston-de:
> What about making the preview image clickable, to show and hide the details?

The issue with this is that some previews, particularly text and html previews are interactive, so expanding/collapsing on click would interfere with normal usage.

yaichenbaum:
> I though the grid splitter was supposed to be removed?

@winston-de I guess you're right. I think we should hold off on this for now so we can get the PR merged.

winston-de:
Can the right click menu option be kept until the design is decided?

Jaiganeshkumaran:
Please don't keep the title in the center while other details left aligned.

yaichenbaum:
> Can the right click menu option be kept until the design is decided?

@winston-de sure.

winston-de:
> Please don't keep the title in the center while other details left aligned.

@Jaiganeshkumaran like this? I think it looks a bit better, @yair100 what do you think?

yair100 (approved):

jaigak:
> > Please don't keep the title in the center while other details left aligned.
> 
> @Jaiganeshkumaran like this? I think it looks a bit better, @yair100 what do you think?
> 

This looks better.

yair100:
> > Please don't keep the title in the center while other details left aligned.
> 
> @Jaiganeshkumaran like this? I think it looks a bit better, @yair100 what do you think?
> 

I think the centered design will look better once the rest of the items in the preview pane are centered as well.

PR #5891, opened 2021-08-22
Fix home page not loading on app start

gave92:
**Resolved / Related Issues**
- Fix home page not loading on app start

I don't know if it's just me but I'm getting this issue where on app startup the homepage doesn't load. This happens if I select "continue where you left off" and I close the app with only the homepage being open.

**Details of Changes**
Add details of changes here.
Added `App.MainViewModel.TabStripSelectedIndex = index;` in `AddNewTabByParam` to match `AddNewTabByPathAsync`.

Since I don't understand why the solution works, can you confirm that it makes sense?

@BanCrash perhaps you've seen this too?

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/ViewModels/MainPageViewModel.cs
+++ b/Files/ViewModels/MainPageViewModel.cs
@@ -441,7 +441,7 @@ namespace Files.ViewModels
             TabItem tabItem = new TabItem()
             {
                 Header = null,
-                IconSource = fontIconSource,
+                IconSource = null,
                 Description = null
             };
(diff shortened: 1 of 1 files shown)


yaichenbaum:
@gave92 I was seeing this issue as well, thanks for the fix!

BanCrash:
@gave92 
> @BanCrash perhaps you've seen this too?

Yep, thanks for fixing it!

winston-de (approved):
lgtm

PR #5894, opened 2021-08-22
Fixed an issue where app center wasn't logging events

yaichenbaum:


Files/ViewModels/SettingsViewModel.cs
```diff
@@ -50,6 +53,7 @@ private async Task<SettingsViewModel> Initialize()
             FileTagsSettings = new FileTagsSettings();
 
             // Send analytics to AppCenter
+            StartAppCenter();
```
gave92, line 56: If the purpose of this change is to make sure that `AppCenter.Start()` is called before `TrackAnalytics()` note that the `async void` method will return before reaching it.

yaichenbaum: Would `await` solve this issue?

gave92: Yup

winston-de (approved):

gave92:
Uff perhaps I was wrong, adding await means that the app will wait for AppCenter before continuing startup. Probably the difference is not noticeable though and it could even be the right thing to do 🤔

yaichenbaum:
> Uff perhaps I was wrong, adding await means that the app will wait for AppCenter before continuing startup. Probably the difference is not noticeable though and it could even be the right thing to do 🤔

Hmm, I have a different idea.

PR #5952, opened 2021-08-28
Focus readme files only when preview pane is open

d2dyno1:
**Resolved / Related Issues**
\-

**Details of Changes**
A readme file will be selected only when preview pane is opened.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Views/MainPage.xaml.cs
+++ b/Files/Views/MainPage.xaml.cs
@@ -28,8 +28,6 @@ namespace Files.Views
     /// </summary>
     public sealed partial class MainPage : Page, INotifyPropertyChanged
     {
-        public static MainPage Instance => (MainPage)((Frame)Window.Current.Content).Content;
-
         public SettingsViewModel AppSettings => App.AppSettings;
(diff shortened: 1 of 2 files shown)


winston-de:
May not be caused by this PR, but if another item contains the term readme, it will be selected instead of readme.md on navigated. Eg, a folder named readme_assets will be selected instead of readme.md.

winston-de (approved):

Files/Views/MainPage.xaml.cs
```diff
@@ -28,6 +28,8 @@ namespace Files.Views
     /// </summary>
     public sealed partial class MainPage : Page, INotifyPropertyChanged
     {
+        public static MainPage Instance => (MainPage)((Frame)Window.Current.Content).Content;
```
yaichenbaum, line 31: Is this change related to the auto preview of readme files?

d2dyno1: Can be removed

yaichenbaum (approved):

Issue #1069, opened 2020-06-15
Add the ability to rearrange pinned items 


Jaiganeshkumaran:
**Is your feature request related to a problem? Please describe.**
No. Currently, to change the order of pinned items shown on the sidebar, I need to manually edit the file or unpin all items and pin them again in the order I want. This is not convenient. 

**Describe the solution you'd like**
The right-click context menu could have a move up and a move down option.

Br0ckout:
Describe the solution you'd like
Click and drag would be a better solution to order them IMO, like how Finder works on MacOS.

ldsands:
If it isn't too much to add in addition to the draging up and down to rearrange, access to a json file similar to what we have access to for the terminal options would be very useful as well.

jaigak:
@yair100 @duke7553 Can I create a custom control for this? Rearranging with drag is not easy with the navigation view control. It was designed for static navigation items and not meant for pinning stuff.

yaichenbaum:
@Jaiganeshkumaran We plan on implementing some other changes to the sidebar #692. It's ok to use a different control such as a ListView as long as it can support the UI we are aiming for.

Jaiganeshkumaran:
List view has built-in rearrange feature so we only need to save the state.

yaichenbaum:
@Jaiganeshkumaran Will it support a hierarchical menu?

jaigak:
@yair100 List view doesn't support that so we can use tree view instead which also supports reordering.

yaichenbaum:
@Jaiganeshkumaran TreeView seems like it can do everything we need.

DeveloperWOW64:
#5609

yaichenbaum:
This issue should be resolved with #5979.

PR #5979, opened 2021-08-30
Added context menu options to rearrange Favorites item

BanCrash:
Opening this as a draft since there is still a visual bug that I didn't figure out how to fix it.

I will try to solve this, but since the timeline for the V2 is almost done I suggest to set a timeline also for this, and if I can't solve this issue since it's only visual merge this and open a new issue to fix it in the future.

~~The issue is that when you move the actual selected item, it will stay as selected even if you select another one:~~ partially fixed, now it will work as expected except that moved item won't have the selection pill.

image

**Resolved / Related Issues**
- Related #1069

**Details of Changes**
- Added context menu options to rearrange Favorites item (right click on an item to have to option to move it up/down/top/bottom).

**Current limitations**
- Not available for Home nor Recycle bin. Maybe in a future PR.
- ~~There is a bug when you move the actual selected item, that it will keep it selected even if you switch to another selected item until you select it back again.~~ There is a bug that doesn't show the selection pill when you change the actual selected item.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

GIF 30-08-2021 15-49-29


BanCrash:
I think the bug I've found it's not from Files but from WinUi. I've found this similar issue, though not exactly the same: 
https://github.com/microsoft/microsoft-ui-xaml/issues/2845

And also this that I'm not sure neither if it's the same: https://github.com/microsoft/microsoft-ui-xaml/issues/3015

Files/Strings/en-US/Resources.resw
```diff
@@ -2848,4 +2848,16 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
...
+  </data>
+  <data name="SideBarFavoritesMoveOneUp.Text" xml:space="preserve">
+    <value>Move one up</value>
+  </data>
+  <data name="SideBarFavoritesMoveToBottom.Text" xml:space="preserve">
+    <value>Move to bottom</value>
+  </data>
+  <data name="SideBarFavoritesMoveToTop.Text" xml:space="preserve">
+    <value>Move to top</value>
+  </data>
```
yaichenbaum, line 2862:
```suggestion
  <data name="SideBarFavoritesMoveOneDown" xml:space="preserve">
    <value>Move one down</value>
  </data>
  <data name="SideBarFavoritesMoveOneUp" xml:space="preserve">
    <value>Move one up</value>
  </data>
  <data name="SideBarFavoritesMoveToBottom" xml:space="preserve">
    <value>Move to bottom</value>
  </data>
  <data name="SideBarFavoritesMoveToTop" xml:space="preserve">
    <value>Move to top</value>
  </data>
```

Files/UserControls/SidebarControl.xaml.cs
```diff
@@ -930,6 +1026,34 @@ public List<ContextMenuFlyoutItemViewModel> GetLocationItemMenuItems()
...
+                    Glyph = "\uE70D",
+                    Command = new RelayCommand(() => MoveItemDown_Click(null, null)),
+                    ShowItem = ShowMoveItemDown
+                },
+                new ContextMenuFlyoutItemViewModel()
+                {
+                    Text = "SideBarFavoritesMoveToBottom/Text".GetLocalized(),
+                    Glyph = "\uE118",
+                    Command = new RelayCommand(() => MoveItemToBottom_Click(null, null)),
+                    ShowItem = ShowMoveItemDown
```
yaichenbaum, line 1054:
```suggestion
                    Text = "SideBarFavoritesMoveToTop".GetLocalized(),
                    Glyph = "\uE11C",
                    Command = new RelayCommand(() => MoveItemToTop_Click(null, null)),
                    ShowItem = ShowMoveItemUp
                },
                new ContextMenuFlyoutItemViewModel()
                {
                    Text = "SideBarFavoritesMoveOneUp".GetLocalized(),
                    Glyph = "\uE70E",
                    Command = new RelayCommand(() => MoveItemUp_Click(null, null)),
                    ShowItem = ShowMoveItemUp
                },
                new ContextMenuFlyoutItemViewModel()
                {
                    Text = "SideBarFavoritesMoveOneDown".GetLocalized(),
                    Glyph = "\uE70D",
                    Command = new RelayCommand(() => MoveItemDown_Click(null, null)),
                    ShowItem = ShowMoveItemDown
                },
                new ContextMenuFlyoutItemViewModel()
                {
                    Text = "SideBarFavoritesMoveToBottom".GetLocalized(),
                    Glyph = "\uE118",
                    Command = new RelayCommand(() => MoveItemToBottom_Click(null, null)),
                    ShowItem = ShowMoveItemDown
```

yaichenbaum (changes requested):

R3voA3:
While I like the idea of being able to change the order of appearance, wouldn't it be more intuitive to just be able to drag and drop the items?

BanCrash:
> While I like the idea of being able to change the order of appearance, wouldn't it be more intuitive to just be able to drag and drop the items? With the context menu, rearranging many items can become tedious.

My idea would be to have both ways and everyone choose whatever option they want (or even sometimes one and sometimes the other). But currently there is a but on WinUI that doesn't allow to have the drag and drop functionality. When that bug is fixed we can have also that way.

R3voA3:
> > While I like the idea of being able to change the order of appearance, wouldn't it be more intuitive to just be able to drag and drop the items? With the context menu, rearranging many items can become tedious.
> 
> My idea would be to have both ways and everyone choose whatever option they want (or even sometimes one and sometimes the other). But currently there is a bug on WinUI that doesn't allow to have the drag and drop functionality. When that bug is fixed we can have also that way.

Sounds great!

yaichenbaum:
> While I like the idea of being able to change the order of appearance, wouldn't it be more intuitive to just be able to drag and drop the items? With the context menu, rearranging many items can become tedious.

As @BanCrash mentioned, drag and drop would be ideal, but until the WinUI issue is resolved, we're implementing this as a workaround.

gave92 (approved):
Nice!

yaichenbaum (approved):

PR #5984, opened 2021-08-30
Updated links to point to new website

yaichenbaum:


Files/ViewModels/SettingsViewModels/AboutViewModel.cs
```diff
@@ -57,7 +57,7 @@ private async void ClickAboutFeedbackItem(ItemClickEventArgs e)
                     break;
 
                 case "Documentation":
-                    await Launcher.LaunchUriAsync(new Uri(@"https://files-community.github.io/docs"));
+                    await Launcher.LaunchUriAsync(new Uri(@"https://files.community/docs"));
```
winston-de, line 60: Should this be a constant to make it easier to update in the future?

winston-de:
Also the links don't seem to work, is the site just not up yet?

Files/ViewModels/SettingsViewModels/AboutViewModel.cs
yaichenbaum: Sounds like a good idea, where would you store the constant?

yaichenbaum:
> Also the links don't seem to work, is the site just not up yet?

The site will be going live with v2.

winston-de (approved):

Issue #5980, opened 2021-08-30
Moving files from a WinRAR archive into a folder in Files does not work


bucklingspring:
### Description

With a Files window and a WinRAR window open side-by-side, dragging a file listed in WinRAR over to Files immediately causes both applications to become unresponsive for a couple seconds. Following this, Files will push a "Something went wrong!" notification, and the app will become responsive once again. However, WinRAR will continue to not respond to user input and must be killed via Task Manager.

### Steps To Reproduce

1. Open Files and navigate to any directory
2. Open a WinRAR archive
3. Drag an item inside a WinRAR archive over to the active directory in Files
4. See the error

### Expected behavior

The expected behaviour is for the dragged item to be extracted from the archive by WinRAR and copied over to the active folder in Files, without freezes or errors.

### Files Version

1.9.30.0 

### Windows Version

Windows 11 (10.0.22000.168)

### Relevant Assets

[debug.log](https://github.com/files-community/Files/files/7077372/debug.log)

gave92:
Could you upload also the other log file (debug_fullrust.txt)? 🙏

gave92:
Reproducible, but only in Azure build (or Release?)

yaichenbaum:
@bucklingspring is this still an issue in the preview version? http://files.community/download/preview

bucklingspring:
@yair100 Archived files/folders from WinRAR can be dragged and dropped seamlessly into Files on 2.2.9. Thank you!

PR #5999, opened 2021-08-31
Avoid crash when dragging files from WinRAR

gave92:
**Resolved / Related Issues**
- Related #5980

**Details of Changes**
Add details of changes here.
- Added null check to avoid crash when dragging files from WinRAR

The issue is reproducible only in Azure/Release mode. This PR _only avoids the crash_, dragging Files from WinRAR is currently not working in Release.

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

PR #6018, opened 2021-09-02
Updated preview pane background and border

yaichenbaum:
image


winston-de:
It feels a bit off to have the status bar and preview pane connected like this.

image

yaichenbaum:
> It feels a bit off to have the status bar and preview pane connected like this.
> 
> image

That's an issue with the way I implemented the border on the status bar, it's something to work on, but I don't think it should hold back this PR.

winston-de (approved):

PR #6077, opened 2021-09-07
Fixed an exception while writing some paths on the address bar

### LINKED ISSUE #5459: Error on VS while writing "Home" in address bar
**Explain the bug**
While testing my implementation of translation of "Home" in the addess bar (#5441) I've found this bug that in the beginning I thought it was caused by my solution, but after that I check that is happening also on main.

The easy way to check this bug is withing VS, but you can check on Files too because in the address bar it will still show "Home" even if you are on another directory.

**To reproduce**
1. Go to a directory, like Documents.
2. Click on Address bar and write Home
3. If you are debugging with VS, it will show you the error. If not, you can easily check doing those extra steps:
4. After forcing the error, go again to the same directory of the step 1. This is important, if you go to another it will not work the next step.
5. Click on Address bar and it will show "Home" instead the actual directory.

**Expected behavior**
It should not throw an error on VS, and it should show the actual directory instead "Home" again.

**Logfile**
[debug_fulltrust.log](https://github.com/files-community/Files/files/6849580/debug_fulltrust.log)
[debug.log](https://github.com/files-community/Files/files/6849581/debug.log)

**Screenshots**
image

image

**System Information**
 - OS Version: Windows 10 Pro 21H1 19043.1081 / Windows Feature Experience Pack 120.2212.3530.0
- App version: Last main commit.

**Additional comment**
This problem still happen after restarting the app.

BanCrash:
**Resolved / Related Issues**
- Closes #5459

**Details of Changes**
- Added a return when the parentFolder is null. This fixes the exception of #5459, though the other issue is not solved, so it seems that after all it was caused for some other reason. 
This exception has been happening to me when writing any path on the address bar from some time ago while debugging. 

**Validation**
- [x] Built and ran the app


gave92:
@BanCrash could you check if the issue #5459 is fixed by 2a2086f?

Files/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
```diff
@@ -217,6 +217,11 @@ public async static Task<IList<StorageFileWithPath>> GetFilesWithPathAsync(this
 
         public async static Task<IList<StorageFolderWithPath>> GetFoldersWithPathAsync(this StorageFolderWithPath parentFolder, string nameFilter, uint maxNumberOfItems = uint.MaxValue)
         {
+            if (parentFolder == null)
```
gave92, line 220: I'd revert this change. There is nothing particularly evil about that exception. The exception in related to the address bar suggestions, not really to the addressbar text and is used to show the "No suggestions" message below the box. There are nicer ways to do this but the exception does its job.

BanCrash: I can revert it if you want, but I didn't see any difference on the adress bar suggestions after this change. The "no suggestions" is still working as expected. The only change I saw is that every time I write a letter on the address bar debugging won't get stopped by that exception.

I will test now your added commit.

BanCrash:
> @BanCrash could you check if the issue #5459 is fixed by [2a2086f](https://github.com/files-community/Files/commit/2a2086fef7e95d81334e27b5ed72ee254a555b40)?

@gave92 yes, that commit fixes it, thanks!

Files/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
gave92: It works because a different exception (ArgumentNullException) is thrown on the next line :)  Anyway on second thought I this is a fine change 👍

BanCrash: Aaah, okay!

Thanks, I appreciate it, it was very annoying for me that exception 😢

gave92 (approved):

PR #6096, opened 2021-09-10
Fixed Delay to Closing Open Themes Dialog/Tooltip

### LINKED ISSUE #5901: Delay in Closing Open Themes Dialog/Tooltip
### Description

While trying to access the themes folder via Settings > Appearances, I noticed that the dialog/tooltip for `Open themes folder` has delayed for around 2-3 seconds after clicking it.

### Steps To Reproduce

1. Go to Settings
2. Click on Appearances Tab and Surf to the Custom Themes Section
3. Click on the Info Button and a Tooltip/Dialog will Appear
4. Click on the `Open themes folder` button
5. See the error

### Expected behavior

No delay in closing modal/tooltip

### Files Version

v1.9.26.0

### Windows Version

Windows 10 21H1 19043.1165

### Relevant Assets

https://user-images.githubusercontent.com/69080584/130393559-66969ed3-58a0-4bff-8bf2-5e37b4bcdaba.mp4

esibruti:
**Resolved / Related Issues**
- Closes #5901 

**Details of Changes**
Add details of changes here.
- Fixed a small delay when the user presses the "Open Themes Folder" button and the dialog/tooltip remains for a few seconds.

**Validation**
- [X] Built and ran the app


Files/ViewModels/SettingsViewModel.cs
```diff
@@ -105,8 +105,12 @@ public static async void OpenLogLocation()
             await Launcher.LaunchFolderAsync(ApplicationData.Current.LocalFolder);
         }
 
-        public static async void OpenThemesFolder() => await NavigationHelpers.OpenPathInNewTab(App.ExternalResourcesHelper.ThemeFolder.Path);
-
+        public static async void OpenThemesFolder()
+        {
+            await Task.Delay(500);
```
yaichenbaum, line 110: I'm not so sure this behavior is better than the alternative.
@gave92 what do you think?

esibruti: @yair100  was the only way I found for him to close the dialog completely and then go to the themes folder.
And the best: it works :)

hez2010: Can you try using `await CoreApplication.MainView.Dispatcher.YieldAsync()` instead?

esibruti: image
@hez2010 With the delay, he solved this...

yair100: > @yair100 was the only way I found for him to close the dialog completely and then go to the themes folder.
> And the best: it works :)

We have some other changes planned that will resolve this issue differently, but it the meantime, we don't want to trade one problem for another 😂

esibruti: So, but does it mean that we are going to use my solution or will it be replaced by your solution?👀

yaichenbaum: I think we should try your solution in the meantime, but we should try to do it without the delay.

esibruti: I will look for alternatives to delay

esibruti: @yair100 I've been looking for alternatives but I can't find another way to do as the Delay currently does....

yaichenbaum: @esibruti can you try the solution @hez2010 suggested?

esibruti: https://user-images.githubusercontent.com/78214358/133153576-fd5b6fd8-5391-4e31-a1c5-e0aebe4e1314.mp4
Method 1 - with delay

https://user-images.githubusercontent.com/78214358/133153573-1428cfee-1254-42be-b514-67eac852a6b2.mp4
Method 2 - @hez2010 sugestion

@yair100 So far I have not been able to find alternatives. Other suggestions?

yaichenbaum: Neither of those actually look too bad.

hez2010: I'll prefer using `Dispatcher.YieldAsync()` than `Task.Delay`(). They both achieve a same thing: yield the UI to respond before doing navigation. `Task.Delay` is too tricky IMO.

esibruti: @yair100 So which option do we use? Or are we going to look for more alternatives?

yaichenbaum: Lets go with `Dispatcher.YieldAsync()`.

esibruti: [https://github.com/files-community/Files/pull/6096/commits/a3dce330c83a01eadf26cf6802f766f798d795d1](url)

Added Dispatcher.YieldAsync()

esibruti: Problem Resolved

BanCrash:
@esibruti did those changes solves the issue to you? I can't see any difference:

GIF 18-09-2021 16-22-01

esibruti:
> @esibruti did those changes solves the issue to you? I can't see any difference (gif made with your version):
> 
> GIF 18-09-2021 16-22-01
> 
> 
> 

This is really weird 👀
In my tests, he didn't do that... I'll see what goes on

esibruti:
@BanCrash I was all weekend trying and trying to see what was going on and it never did to me... It's very strange.

BanCrash:
> @BanCrash I was all weekend trying and trying to see what was going on and it never did to me... It's very strange.

I had visual studio issues, when I solve them I will give it this a new try to see if I can still replicate them.

BanCrash:
I've been able to test this with a virtual machine since my Visual Studio is still not working, and this is the result:

image

image

Before disappearing it appears small. Well, it seems that this doesn't totally fix the issue but improves it.

Does this happens to you too @esibruti ?

esibruti:
> I've been able to test this with a virtual machine since my Visual Studio is still not working, and this is the result:
> 
> image
> 
> image
> 
> Before disappearing it appears small. Well, it seems that this doesn't totally fix the issue but improves it.
> 
> Does this happens to you too @esibruti ?

@BanCrash It does just that in my Visual Studio. I tried to use the solution to put `Task.Delay()`, which in this way solves it, but @yair100 said it would be better to use `Dispatcher.YieldAsync()`.

yaichenbaum:
The issue with the delay is that every device is different, I might be wrong but I think `Dispatcher.YieldAsync()` will be more consistent.

esibruti:
> The issue with the delay is that every device is different, I might be wrong but I think `Dispatcher.YieldAsync()` will be more consistent.

@yair100 I agree. It's better to have the code consistent and stay that way, a really tiny thing than the code to be inconsistent and can cause problems.

BanCrash (approved):
Since this improves (when not directly fixes) the issue, it's fine for me.

LGTM.

yaichenbaum:
@esibruti thank you!

PR #6104, opened 2021-09-11
feat: Don't convert primary thumbnail to byte array

### LINKED ISSUE #6089: Discussion: Investigate the fetching of unneeded byte array per storage item
### Description

Previously, we used the FullTrustProcess for all item thumbnails and an optimization was put in place to associate the underlying byte array with the ListedItem. I'm proposing that we remove this trait from ListedItem and ItemViewModel because we could see a small performance gain for the items affected through simply fetching the StorageItemThumbnail and setting it as the BitmapImage source directly. 

The only area leveraging this byte array is the Properties dialog. In my opinion, this makes little sense and we should load thumbnails from the filesystem for the Properties experience on-demand.

### Steps To Reproduce

Load a large directory.

### Expected behavior

As we know, users expect item thumbnails to appear quickly. I think we should do this without the overhead of converting the thumbnail into a byte array only to subsequently form a bitmap.

### Files Version

v2.0

### Windows Version

Windows 11, 10

### Open questions:

Is a converted byte array via the extension method still somehow preferrable over  direct use of the thumbnail stream with `BitmapSource.SetSourceAsync(fileThumbnail)` ?

lukeblevins:
**Resolved / Related Issues**
- Related #6089 

**Details of Changes**
- Removes unneeded reliance on thumbnail byte array for remaining app surfaces
- Switches away from glyphs as icons of items with unloaded extended properties

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
N/A


Files/ViewModels/ItemViewModel.cs
```diff
@@ -753,48 +753,39 @@ private async Task LoadItemThumbnail(ListedItem item, uint thumbnailSize = 20, I
...
-                                    item.LoadUnknownTypeGlyph = false;
-                                    item.LoadWebShortcutGlyph = false;
-                                    item.LoadFileIcon = true;
-                                }, Windows.System.DispatcherQueuePriority.Low);
-                                wasIconLoaded = true;
-                            }
+                                item.LoadUnknownTypeGlyph = false;
+                                item.LoadWebShortcutGlyph = false;
+                                item.LoadFolderGlyph = false;
+                                item.LoadFileIcon = true;
```
duke7553, line 771: Ensures Image is added to visual tree before we try to set a source as per [this](https://docs.microsoft.com/en-us/windows/uwp/debug-test-perf/app-analysis#image-is-being-called-when-the-imagesource-is-not-in-the-live-tree) recommendation

lukeblevins:
@yair100 and I did some testing and it's a mixed bag:

1.) This change didn't yield a very large improvement in thumbnail load times. In fact, they were mostly comparable across the two versions
2.) It did, however, reduce memory usage by 35% during the particularly synthetic scenario where one loads a large directory and scrolls every item into the viewport (That amounted to 200MB for 4620 items in System32)
3.) CPU usage was lower (-5%), but not consistently less enough to be significant


I think we should go ahead with this change after I fix the Properties dialog thumbnails, but I'll let this sit for some more input

yaichenbaum:
@duke7553 I think it makes sense to continue with this and hopefully we can include it in one of the next preview builds. For the meantime, do you think you can add back the wireframe placeholder icon until we get a change to come up with a proper alternative?

d2dyno1:
Is this still relevant?

duke7553:
@d2dyno1 Why wouldn't it be? It solved an issue where we used an unnecessary amount of memory for thumbnails. I just have to get around to fixing the remaining bug where surfaces like Properties dialog relied on the byte array.

d2dyno1:
Ok! I was just checking on if any work is still being made.

yair100 (approved):
Great work!

PR #6107, opened 2021-09-12
Set Files as Open dialog & set Files as default from Settings

gave92:
**Description**
A funny experiment to see if it's possible to replace the shell file open dialog with Files (spoiler: it's possible!)
This is by no means complete, work is necessary on the Files side to adapt the interface for the purpose but it shows what we can achieve.

**How to install/uninstall**
Option to set Files as open file dialog can be found under Settings -> Experimental
May need to uninstall Store version if installed

**How to use**
- Trigger the open dialog and enjoy your favorite Files manager open
- Select a file
- Close the app

**Notes**
- Does not work for older apps that use a different way to show the files open dialog (e.g Notepad)
- Not used for apps running as administrator
- Setting Files as open file dialog adds a couple of registry keys in HKEY_CURRENT_USER, I don't think it has the capability to blow up your PC but keep in mind the warnings about registry edits ;)

**Validation**
- [x] Tested with: Notepad++, Microsoft edge, Visual Studio, Office, Quarrel (UWP).

**Screenshots**


yaichenbaum:
@gave92 which apps does this work with?

gave92:
@yaichenbaum at least notepad++, Microsoft edge, visual studio.

gave92:
> @gave92 which apps does this work with?

Added tested apps to the top post.

yaichenbaum:
@gave92 looks like a good start! How do you envision users setting this up?

gave92:
> @gave92 looks like a good start! How do you envision users setting this up?

We can probably offer an option in Files settings to set/unset Files as default explorer and as open file dialog.

sylveon:
The way this is currently implemented, it results in the .NET Framework Runtime being loaded in every process that opens a file browse dialog. While it's not a shell extension directly, the same [concerns that MS raises against writing C# shell extensions](https://docs.microsoft.com/en-us/windows/win32/shell/shell-and-managed-code) applies. This will lead to a higher likelyhood of crashes/bugs, and puts the burden of these issues on third party application devs (faulty shell extensions is already why many apps implement their own file browser instead of using the system one).

I would ***strongly*** recommend rewriting this portion in C++, as it seems to be mostly trivial anyways (and it would allow you to use the correct way to register COM classes, `regsvr32`, instead of registry files).

gave92:
Yeah a rewrite in C++ is coming. But I suck at C++ and I wanted to see if this thing was working before getting my hands dirty ;)

>  allow you to use the correct way to register COM classes, regsvr32

I'd still use the reg file: regsvr32 install the dll in HKEY_LOCAL_MACHINE. I prefer the current that writes stuff to HKCU instead (less risky and you don't need to be administrator).

sylveon:
regsvr32 just calls `DllRegisterServer`. The DLL can do whatever it wants in that, including registering in HKCU. The default does use HKLM, but you can change it.

gave92:
> regsvr32 just calls `DllRegisterServer`. The DLL can do whatever it wants in that, including registering in HKCU. The default does use HKLM, but you can change it.

That's good, thanks!

sumeshir26:
Would this work with the 'Save As' Dialog?

d2dyno1:
> Would this work with the 'Save As' Dialog?

I found no instances of `IFileSaveDialog` so I presume the answer is no

gave92:
> Would this work with the 'Save As' Dialog?

Can be done for Save Dialog too, but currently not implemented

gave92:
Converting to draft till the porting to C++ is done.

gave92:
Converted dll to C++ 💪. Curious if someone had the guts to test this xD

sylveon:
looks okay to me. The files use italian comments but that's because of VS.

Can also get rid of the two reg files and call `regsvr32.exe /n /i:user thing.dll` (add `/u` to uninstall)

gave92:
Added ability to set Files as default file manager and optionally as file open dialog from Files Experimental settings.
The option to use Files as open file dialog should be hidden before merging the PR as it's really early stage.

winston-de:
Just tested it, works great in apps like Edge. It seems to be causing uwp apps like your phone or quick pad to crash or just not do anything when attempting to invoke the file picker.

gave92:
@winston-de for UWP apps it fails miserably with an InvalidCastExpeption. The custom dialog constructor gets called but that's it, no clue, sigh,

sylveon:
Check the calls to QueryInterface, you're probably missing an interface.

gave92:
Thanks @sylveon it looks like I'm missing the following:
```
QueryInterface: {FC4801A3-2BA9-11CF-A229-00AA003D7352} = IObjectWithSite
QueryInterface: {79EAC9EE-BAF9-11CE-8C82-00AA004BA90B} = IInternetSecurityManager
```

gave92:
@winston-de thanks to @sylveon pointers and @Cthulhu intervention it looks like UWP apps now work correctly, could you confirm?

winston-de:
Works now :)

One other issue, if you try and open more then one dialog at a time (click open in two different apps), this error will be shown (if you click ignore it crashes).
image

gave92:
@winston-de thanks should be just because of the debug logs to the same file. Disabled logging now & hidden the "Set as open file dialog" UI option.

winston-de (dismissed):
lgtm

Files/Views/SettingsPages/Experimental.xaml
```diff
@@ -70,6 +72,49 @@
...
+                    <i:Interaction.Behaviors>
+                        <icore:EventTriggerBehavior EventName="Toggled">
+                            <icore:InvokeCommandAction Command="{Binding SetAsDefaultExplorerCommand, Mode=OneWay}" />
+                        </icore:EventTriggerBehavior>
+                    </i:Interaction.Behaviors>
+                </ToggleSwitch>
+            </local:SettingsBlockControl>
+
+            <local:SettingsBlockControl
+                x:Uid="SettingsSetAsOpenDialog"
```
yaichenbaum, line 100: I've been working on moving over to a new method of supporting translations via `x:Bind`, do you think you can use this method as well? https://devblogs.microsoft.com/ifdef-windows/use-a-custom-resource-markup-extension-to-succeed-at-ui-string-globalization/

yair100 (changes requested):

yaichenbaum:
@gave92 is the option to set Files as the open file dialog supposed to be hidden?

gave92:
Yeah I would enable it once some work has been done to adapt Files UI for the purpose.

yaichenbaum:
> Yeah I would enable it once some work has been done to adapt Files UI for the purpose.

Should double clicking a file be disabled when Files is acting as the open file dialog?

gave92:
@yair100 many thing should change when Files is used as open file dialog (no tabs? an ok button? support for file types filters?), this is "just" ground work, no UI changes done or planned.
As for double clicking guess we should follow explorer and use it to confirm selection.

winston-de:
Can you resolve the merge conflicts please?

winston-de:
Also side note, maybe we should add a link to app execution alias settings, in case the files.exe execution alias has been disabled (which would break the default explorer setting).

yaichenbaum:
> Also side note, maybe we should add a link to app execution alias settings, in case the files.exe execution alias has been disabled (which would break the default explorer setting).

Once the user toggles this switch, should we adjust the execution alias in the event is was disabled?

gave92:
Merged main.

winston-de (approved):

yair100 (approved):
Great work! LGTM

neoOpus:
Yes Please! this would be the best option so far and from there make it more robust, also, please integrate some of Listary features for the Open/Save Dialog... also make it easy to switch to Windows Explorer just in case something isn't working properly in some scenarios (on fly without the need to change the default explorer)

vivadavid:
Hi, I've enabled "Set files as default explorer", but now I can't disable it. I've tried uninstalling the app, but then I can't even use the Windows file explorer.

I'm on version Versión: 2.0.33.0, OS Version: 10.0.19043.1288.

I'd appreciate some help! Thanks!

BanCrash:
@vivadavid Hi, follow the steps here to unset it: https://files.community/docs/configuring/replace-file-explorer

You need to use the UnsetFilesAsDefault.reg file.

vivadavid:
Thank you, @BanCrash ! I tried to do it manually, but I must have done something wrong. However running the reg file did the trick and things are back to normal.

Issue #5195, opened 2021-06-13
[Codebase] Use IDialogService for dialogs


d2dyno1:
**Is your feature request related to a problem?**
We currently don't have any solution for showing `ContentDialog`s in ViewModels. Creating an instance of a `ContentDialog` doesn't comply with MVVM since the ViewModel should now nothing about the UI and it's components.

**Describe the solution**
To solve that we could use a `IDialogService` to show dialogs instead.

**Describe alternatives you've considered**
no.

**Scope**

| Capability | Priority |
| :---------- | :------- |
| Use IDialogService to replace `ContentDialog` creation in ViewModels | Must |

**Additional comment**
I already have the code for that in Clipboard Canvas so it could be reused here, if assigned. For #2341

d2dyno1:
@yair100 Can you assign me?

yaichenbaum:
This issue hasn't been approved yet but I can assign you if/when it is approved.

duke7553:
@d2dyno1 Can you provide a quick run-down of what IDialogService would consist of and the pattern it will use to enable showing dialogs from ViewModels?

d2dyno1:
@duke7553 Take a look at the implementation. It explains pretty much itself what it does: [IDialogService](https://github.com/d2dyno1/ClipboardCanvas/blob/master/ClipboardCanvas/Services/IDialogService.cs), [DefaultDialogService](https://github.com/d2dyno1/ClipboardCanvas/blob/master/ClipboardCanvas/Services/DefaultDialogService.cs)

Example use:
```cs
DialogVM viewModel = new();
IDialog<DialogVM> dialog = App.DialogService.GetDialog<DialogVM>(viewModel); // Get IDialog<DialogVM>

DialogResult option = await dialog.ShowAsync(); // Show dialog
// You can also call IDialogService.ShowDialog() directly, see above reference links to CC's code
```

d2dyno1:
@duke7553 @yair100 Any thoughts about the above logic and how we can use it in Files?

gave92:
It looks pretty similar to what were already doing, eg from FilesystemHelpers.cs:

```cs
FilesystemOperationDialog dialog = await FilesystemOperationDialogViewModel.GetDialog(new FilesystemItemsOperationDataModel(...));
ContentDialogResult result = await dialog.ShowAsync();
```

I'm not sure I'm understanding what benefits this change would bring.

d2dyno1:
[...]

PR #6113, opened 2021-09-13
Transitioned to Json Settings

### LINKED ISSUE #6033: Transition settings to json format
### What's the Problem?

The current method of saving settings works fine, but there is fragmentation between tags, themes and preferences.

### Solution/Idea

Moving all settings to json will all settings to be stored in one files with the added benefit of allowing users to export and import settings manually.

### Alternatives

NA

### Priorities

_No response_

### Files Version

v2.0

### Windows Version

Windows 11

### Additional comment

_No response_

d2dyno1:
**Resolved / Related Issues**
- Closes #6033

**Details of Changes**
This PR transitions registry based `SettingsViewModel` to json settings. Now importing and exporting your settings is possible!
This PR also introduces BaseJsonSettingsModel v3 with refactored codebase and `ISettingsSharingContext` for sharing one settings file across multiple `BaseJsonSettingsModel` instances. This allows for settings organization while keeping everything in one place.

**Benefits**
- More performant settings access thanks to caching
- Flexible settings store allows serialization and deserialization of various object types
- All-new BaseJsonSettingsModel v3 with reworked codebase and support of sharing settings instances
- Flexible settings access stores all settings in one file which allows easy importing/exporting of your configuration.

**Regressions**
- Potential regressions and bugs may occur

**Note for Contributors**
- This PR also includes `IServiceProvider` for injecting dependencies into classes (currently only used for injecting setting services). The IoC container was added in mind of further expansion.
- `SettingsViewModel` is deprecated and should be no longer used for storing settings. It is only used for merging the settings database.
- With these changes also comes a new way to access settings:

BEFORE:
```cs
bool val = App.AppSettings.ShowBundlesWidget;
```
AFTER:
```cs
IUserSettingsService userSettingsService = Ioc.Default.GetService<IUserSettingsService>();
bool val = userSettingsService.WidgetsSettingsService.ShowBundlesWidget;
```

**Validation**
- [x] Built the app

For #4180, #5195


Files/Models/Settings/BundlesSettingsModel.cs
```diff
@@ -11,7 +12,7 @@ public class BundlesSettingsModel : BaseJsonSettingsModel, IBundlesSettings
 
         public BundlesSettingsModel()
             : base(System.IO.Path.Combine(ApplicationData.Current.LocalFolder.Path, Constants.LocalSettings.SettingsFolderName, Constants.LocalSettings.BundlesSettingsFileName),
-                  isCachingEnabled: true)
+                  isCachingEnabled: false)
```
d2dyno1, line 15: Set `isCachingEnabled` to prevent from breaking Bundles serialization

d2dyno1:
@yair100 Ready for review!

winston-de:
I'm getting a null reference exception at startup on line 338 of MainPageViewModel.cs

image

```
System.NullReferenceException: 'Object reference not set to an instance of an object.'

InterfaceDispatchCell00ADAFE8_$0_Files.Services.IUserSettingsService_slot_9(...) returned null.
```

Files/Models/Settings/BundlesSettingsModel.cs
d2dyno1: Re-added the caching feature in https://github.com/d2dyno1/Files/commit/fb0fa9da31a2f6942150c4d09180dc0a1afcdeb2

gave92:
@d2dyno1 I've pushed a few changes. Feel free to revert them if you disagree. Basically:
- Added registration for StartupSettingsService in App.xaml.cs, should fix issue reported by @winston-de
- Removed old settings definitions from SettingsViewModel (they were only used for porting)
- Track settings changes through AppCenter

d2dyno1:
I've just noticed them, thank you :heart: From the initial look, these look fine so I'm going to keep them in. I'll merge some other settings soon and then it's finished!

Files/Models/JsonSettings/Implementation/CachingJsonSettingsDatabase.cs
```diff
@@ -13,6 +13,26 @@ public CachingJsonSettingsDatabase(IJsonSettingsSerializer jsonSettingsSerialize
...
 
+        public override TValue GetValue<TValue>(string key, object defaultValue = null)
+        {
+            if (settingsCache.ContainsKey(key))
+            {
+                var value = settingsCache[key];
+                if (value is Newtonsoft.Json.Linq.JToken jTokenValue)
+                {
+                    var objValue = jTokenValue.ToObject<TValue>();
+                    settingsCache[key] = objValue;
```
gave92, line 24: For file tags I need `GetValue` to return the same instance of an object on subsequent calls, so I cache the actual value instead of the JToken.

gave92:
@d2dyno you probably know this already but since the new settings are not "observable" some things do not work, e.g. toggling the preview pane.

d2dyno1:
> @d2dyno1 you probably know this already but since the new settings are not "observable" some things do not work, e.g. toggling the preview pane or hiding/showing hidden files.

We did hook up `PropertyChanged` event (found in SettingsViewModel) to get notified when a setting is updated, I've changed it since and now we hook up the `OnSettingChangedEvent`. Is that not working? Could you please test for me if [this](https://discord.com/channels/725513575971684472/732087708754903100/887795120643113018) works (since VS is preventing me from running the app) @gave92

gave92:
The problem is that settings do not implement INotifyPropertyChanged so this kind of thinkgs don't work : `IsChecked="{x:Bind UserSettingsService.PreviewPaneSettingsService.PreviewPaneEnabled, Mode=TwoWay}"`

d2dyno1:
@gave92 I've fixed the issue. The problem actually was nothing to do with `INotifyPropertyChanged`. It was simply due to events firing separately -> now they're redirected to base `BaseJsonSettingsModel` (`IUserSettings`) through `ISettingsSharingContext`

gave92:
Looks mostly fixed 👍 This setting does not work though: IsVerticalTabFlyoutEnabled

d2dyno1:
> @d2dyno1 could you check if a fresh install is working?
> 
> If I make a fresh install of this branch, it will give me a lot of errors and won't work. If I install first last main and then install this version so it gets updated, it works as expected.

I've just ran a fresh build of `iusersettings`. Haven't had any problems with it

gave92 (approved):
LRGTM (R is for "really")

yair100 (approved):

PR #6129, opened 2021-09-15
Added drop to Pin folder/s to Favorites

### LINKED ISSUE #2626: Pin folders by dropping them on left pane
I'm frustrated to see that I cannot pin any folder to left pane. When i tried to do so, my folder is copied under a folder present on the pane.

Users should be able to pin folders under a Quick Access tab.

Existing File Explorer solution is just fine.

**Additional context**
feature request

BanCrash:
**Resolved / Related Issues**
- Closes #2626

**Details of Changes**
- Added the ability to drop folder/s to favorites.

You can drop several folders to Favorites at the same time. You can include files on the selection, but they will be ignored. You can include on the selection folders that are already pinned, but they will be ignored too.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
With two files and folders selected:
GIF 15-09-2021 14-30-01

When the operation is allowed:
image

With a folder that is already pinned selected the operation is not allowed:
image

Just adding files will not work neither:
image

### DIFF
--- a/Files/UserControls/SidebarControl.xaml.cs
+++ b/Files/UserControls/SidebarControl.xaml.cs
@@ -499,6 +499,8 @@ namespace Files.UserControls
 
         private object dragOverSection, dragOverItem = null;
 
+        private bool isDropOnProcess = false;
+
         private void NavigationViewItem_DragEnter(object sender, DragEventArgs e)
         {
             VisualStateManager.GoToState(sender as Microsoft.UI.Xaml.Controls.NavigationViewItem, "DragEnter", false);
@@ -543,6 +545,8 @@ namespace Files.UserControls
         {
             VisualStateManager.GoToState(sender as Microsoft.UI.Xaml.Controls.NavigationViewItem, "DragLeave", false);
 
+            isDropOnProcess = false;
+
             if ((sender as Microsoft.UI.Xaml.Controls.NavigationViewItem).DataContext is INavigationControlItem)
             {
                 if (sender == dragOverItem)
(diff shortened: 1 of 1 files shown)


BanCrash:
I added a variable to fix a bug that caused to pin to favorites the folders that you copied / movied to a favorite folder using the sidebar.

gave92:
Perhaps with this we can consider removing the "Pin to favorites" context menu option, or move it to the More submenu like "Pin to start".

yaichenbaum:
> Perhaps with this we can consider removing the "Pin to favorites" context menu option, or move it to the More submenu like "Pin to start".

I would opt for the second option for now since users already expect the option to be in the menu.

yaichenbaum:
@BanCrash should there be a tooltip that says "Pin to favorites"?

BanCrash:
> @BanCrash should there be a tooltip that says "Pin to favorites"?

Do you mean this?

image

If so, it's already on this PR (Anclar a Favoritos = Pin to favorites).

In the screenshots of the first post it doesn't show it because it's not allowed, but when the operation is allowed it will be show as expected.

Or do you mean somewhere else?

yaichenbaum:
> Do you mean this?

@BanCrash yes, that's what I meant. Looks good!

yair100 (approved):

winston-de (approved):

BanCrash:
Btw, about your comments @gave92 and @yair100 , I agree about moving to the more submenu, but there are folders that you can only pin when you are inside, so removing this would cause that they couldn't be pinned to favorites.

When my VS works again, I will work on it.

PR #6134, opened 2021-09-15
Fix for grouping by date issues #5217

### LINKED ISSUE #5217: Date grouping mixed with last year files
**Explain the bug**
When enabling Date grouping (Date Modified and Date Created), grouped files and folder is mixed with last year date modified.

**To reproduce**
1. Go to a folder with older files
2. Click on Sort and Group By Date Modified
3. All the grouped items mixed with last year date items

**Expected behavior**
Items grouped normally with today and yesterday date 

**Logfile**
[debug.log](https://github.com/files-community/Files/files/6646397/debug.log)

**Screenshots**
image

**System Information**
 - OS Version: Windows 10 21H1 19043.1055
 - App version: Build from `main` branch from 0f53a4241323f64ccaf66d6c0e281a09349e411c commit

SHADOW-XIII:
**Resolved / Related Issues**
- Closes #5217

**Details of Changes**
Add details of changes here.
- Fixed showing wrong files in "Today" and "Yesterday" section when grouping by date

**Validation**
- [ ] Built and ran the app

### DIFF
--- a/Files/Extensions/DateTimeExtensions.cs
+++ b/Files/Extensions/DateTimeExtensions.cs
@@ -66,6 +66,7 @@ namespace Files.Extensions
             Windows.Globalization.Calendar cal = new Windows.Globalization.Calendar();
             var t = DateTimeOffset.Now;
             var t2 = dt.ToLocalTime();
+            var y = t.AddDays(-1);
             var today = DateTime.Today;
 
             var diff = t - dt;
@@ -74,7 +75,7 @@ namespace Files.Extensions
                 return ("ItemTimeText_Today".GetLocalized(), today.ToUserDateString(), "\ue184", 0);
             }
 
-            if (t.Month == t2.Month && t.Day - t2.Day < 2 && t.Year == t2.Year)
+            if (y.Month == t2.Month && y.Day == t2.Day && y.Year == t2.Year)
             {
                 return ("ItemTimeText_Yesterday".GetLocalized(), today.Subtract(TimeSpan.FromDays(1)).ToUserDateString(), "\ue161", 1);
             }
--- a/Files/Extensions/DateTimeExtensions.cs
+++ b/Files/Extensions/DateTimeExtensions.cs
@@ -69,12 +69,12 @@ namespace Files.Extensions
             var today = DateTime.Today;
 
             var diff = t - dt;
-            if (t.Month == t2.Month && t.Day == t2.Day)
+            if (t.Month == t2.Month && t.Day == t2.Day && t.Year == t2.Year)
             {
                 return ("ItemTimeText_Today".GetLocalized(), today.ToUserDateString(), "\ue184", 0);
             }
 
-            if (t.Month == t2.Month && t.Day - t2.Day < 2)
+            if (t.Month == t2.Month && t.Day - t2.Day < 2 && t.Year == t2.Year)
             {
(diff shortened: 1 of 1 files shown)


Files/Extensions/DateTimeExtensions.cs
```diff
@@ -66,15 +66,16 @@ public static (string text, string range, string glyph, int index) GetFriendlyTi
             Windows.Globalization.Calendar cal = new Windows.Globalization.Calendar();
             var t = DateTimeOffset.Now;
             var t2 = dt.ToLocalTime();
+            var y = t.AddDays(-1);
             var today = DateTime.Today;
 
             var diff = t - dt;
-            if (t.Month == t2.Month && t.Day == t2.Day)
+            if (t.Month == t2.Month && t.Day == t2.Day && t.Year == t2.Year)
```
winston-de, line 73: looks good. can you please add this fix to the other two checks below this one (line 83 and 88)?

SHADOW-XIII: thanks, I will, 
I will also fix line 98 which I think will break with year change (Jan - 1 => -11)

SHADOW-XIII: @winston-de  I just realised, on line 83, "This Week" can be partially in two years (ie. this year Mon-Fri is in 2021 and Sat-Sun is 2022) so maybe the year check shouldn't be there. diff.Days should ensure it won't be showing files older than 7 days anyway, right?). Need to run some checks what GetWeekOfYear returns depending if it's 31st or 1st (but same week) etc

Similarly line 88: "Last Week" could be in a previous year, however this one needs another fix regardless for year change, ie. when 't2' week is in new year (week 1) and previous one is week 52.

winston-de: Good catch, a solution could be to calculate the number of weeks since the default date, and use that to check if the weeks are the same.

Files/Extensions/DateTimeExtensions.cs
```diff
@@ -66,15 +66,16 @@ public static (string text, string range, string glyph, int index) GetFriendlyTi
...
             {
                 return ("ItemTimeText_Today".GetLocalized(), today.ToUserDateString(), "\ue184", 0);
             }
 
-            if (t.Month == t2.Month && t.Day - t2.Day < 2)
+            if (y.Month == t2.Month && y.Day == t2.Day && y.Year == t2.Year)
```
winston-de, line 78: also what's the reason for replacing the `t.Day - t2.Day < 2` with `y.Day == t2.Day`? This section is meant to show items from the previous day, hence why it checks that it's not more then two days apart.

SHADOW-XIII: it doesn't go well when month changes, for that I created "y" to be "Now-1 day" and compare that instead
`
var t  = new DateTime(2021,12,01,01,02,03);
var t2 = new DateTime(2021,11,29,23,02,03);
Console.WriteLine(t.Day - t2.Day);
`
(output is -28)

Files/Extensions/DateTimeExtensions.cs
SHADOW-XIII: OK, just pushed some more changes to the date grouping, using variable "w" for week comparisons X days ago (where X is the days difference). This should correctly handle different culture/calendar setups

And small fix for LastMonth check as well.

I hope that fixes it all :)

winston-de (approved):

yaichenbaum:
@SHADOW-XIII thank you!

PR #6144, opened 2021-09-16
Fix issue where image previews in preview pane wouldn't work

winston-de:
**Resolved / Related Issues**
- Closes #issue...
- Related #issue...

**Details of Changes**
Add details of changes here.
- An invalid cast to `FileRandomAccessStream` was being made, causing image previews to not be shown. The cast was unnecessary. This pr removes the cast and instead just uses the `IRandomAccessStream` return type, fixing the issue. 
- also removes some old code for svg previews that's no longer being used
- fix issue where header images would show item preview thumbnails instead of type thumbnails in zoomed out view when grouped by type. This issue was due to needing to get the thumbnail via the ftp in order to show just the app thumbnail. 

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/Files/ViewModels/ItemViewModel.cs
+++ b/Files/ViewModels/ItemViewModel.cs
@@ -1074,7 +1074,8 @@ namespace Files.ViewModels
             ImageSource groupImage = null;
             if (item.PrimaryItemAttribute != StorageItemTypes.Folder)
             {
-                var headerIconInfo = await FileThumbnailHelper.LoadIconFromPathAsync(item.ItemPath, 76, ThumbnailMode.ListView);
+                var headerIconInfo = await FileThumbnailHelper.LoadIconWithoutOverlayAsync(item.ItemPath, 64u);
+
                 if (headerIconInfo != null && !item.IsShortcutItem)
                 {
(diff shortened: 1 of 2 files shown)


gave92 (approved):

Files/ViewModels/ItemViewModel.cs
```diff
@@ -1072,9 +1072,10 @@ public async Task LoadExtendedItemProperties(ListedItem item, uint thumbnailSize
         private async Task<ImageSource> GetItemTypeGroupIcon(ListedItem item, BaseStorageFile matchingStorageItem = null)
         {
             ImageSource groupImage = null;
-            if (item.PrimaryItemAttribute != StorageItemTypes.Folder)
+            if (item.PrimaryItemAttribute != StorageItemTypes.Folder || item.IsZipItem)
             {
-                var headerIconInfo = await FileThumbnailHelper.LoadIconFromPathAsync(item.ItemPath, 76, ThumbnailMode.ListView);
+                var headerIconInfo = await FileThumbnailHelper.LoadIconWithoutOverlayAsync(item.ItemPath, 64u);
```
gave92, line 1077: I understand there's no way to force a StorageFile to return the icon instead of the thumbnail, correct? Does `ThumbnailMode.DocumentsView` make any difference?

winston-de: doesn't seem like it

PR #6188, opened 2021-09-20
Removed negative margin from settings dialog

yaichenbaum:


d2dyno1:
The content had padding so that's why negative margin was used

yaichenbaum:
> The content had padding so that's why negative margin was used

I added the original negative margin but it's no longer needed.

duke7553 (approved):
LGTM

Issue #2568, opened 2020-12-09
Improve App Startup Performance


dreaddr:
**Describe the bug**
Slow to start, almost sluggish. I have an SSD and this hasn't happened before, overall the app feels laggy to use.

**To Reproduce**
1. Start the app and use it.

**Expected behavior**
Be fast and not be laggy.

**Screenshots**
How to add screenshots of the lag? 

**Desktop (please complete the following information):**
 - OS Version: 
Edition	Windows 10 Pro
Version	20H2
Installed on	‎11/‎18/‎2020
OS build	19042.630
Experience	Windows Feature Experience Pack 120.2212.31.0

 - App version: [v0.21.1.0]

**Additional context**
Add any other context about the problem here. Does this problem occur again after restarting the app? **Yes.**

**Log file**
Please post the log file here so that we can understand your problem better. You can access it from Settings->About->Open log location.
[debug.txt](https://github.com/files-community/Files/files/5666643/debug.txt)

fisforfaheem:
takes some time to load..

dreaddr:
> takes some time to load..

It usually does, yes, but these wait times can be the difference between switching or sticking with native file explorer.

yaichenbaum:
@dreaddr Do you have the setting to continue where you left off turned on? This can cause the app to load a little slower.

dreaddr:
@yair100 nope, its turned off but I did toggle and try experimental features a bit, though it's the same performance even if I use experimental or not (its not noticeable) (Version 0.23.1.0 seems a bit smoother but maybe that has more to do with the animations or me having confirmation bias)

dreaddr:
As of version 1.5.2.0, the app has become significantly more fluid and fast in its animations and transitions. It still needs refinement to get on par with explorer. One app I've seen that matches or even exceeds expectations of loading times is Notepads (also found in the MS Store) although its definitely not the same category of apps, maybe the devs can look into methods they've used to make sure the app loads fast.

fisforfaheem:
Thanks! hoping to see this app improve over time and rival MSFt version

gulshanmudgal:
The load time of files and folders is quite slow, After loading the application it takes at least 2-3 seconds to load all the drives and all the folders.

Another issue is, if any folder is opened then the folder icons don't appear at once, instead appear one by one.

fisforfaheem:
i don't understand but version 2 seems so bland to me too ??? version 1 in
2020 was so much more than the current version??? why this happened?
[...]

PR #6241, opened 2021-09-26
feat: Prelaunch app content & properly redirect to pre-launched instance if available

lukeblevins:
**Resolved / Related Issues**
- Should help in many cases with #2568 -- since we now properly preload the initial app UI, rather than just an empty window
- Redirect properly to pre-launched AppInstance if available

**Validation**
- [x] Built and ran the app

In a simulated prelaunch scenario, the app resumes from suspend almost instantly and displays the app UI with minimal/no splash screen shown.

**Please note:** The changes to Program.cs are tentative due to the high risk of regressions in functionality. ~~Additionally, an issue is present where trying to open more app instances after being redirected to the initial pre-launched instance won't work. Your feedback and suggestions here are very welcome.~~ **(Fixed)**

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -166,6 +166,23 @@ namespace Files
                 Window.Current.Activate();
                 Window.Current.CoreWindow.Activated += CoreWindow_Activated;
             }
+            else
+            {
+                if (rootFrame.Content == null)
+                {
+                    // When the navigation stack isn't restored navigate to the first page,
+                    // configuring the new page by passing required information as a navigation
+                    // parameter
+                    rootFrame.Navigate(typeof(MainPage), e.Arguments, new SuppressNavigationTransitionInfo());
+                }
+                else
+                {
+                    if (!(string.IsNullOrEmpty(e.Arguments) && MainPageViewModel.AppInstances.Count > 0))
+                    {
+                        await MainPageViewModel.AddNewTabByPathAsync(typeof(PaneHolderPage), e.Arguments);
+                    }
+                }
+            }
         }
 
         protected override async void OnFileActivated(FileActivatedEventArgs e)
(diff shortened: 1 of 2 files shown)


duke7553:
Steps to simulate load from prelaunch state with VS:
0.) Clone this PR
1.) Debug > Other Debug targets > Debug UWA Prelaunch
2.) Wait for app to fully load into memory and appear in task manager under "background apps"
3.) "Detach All" from VS Debug menu
4.)  Launch "Files - Dev" from the Start menu, taskbar, or other shell surfaces (Take notice of drastically improved speed)

yaichenbaum:
@duke7553 this seems to be failing in release mode.

lukeblevins:
@yair100 Not sure why it would fail given I didn't modify the sln file. Can you test again and provide the debug output? Is the build failing or does it fail to deploy and run?

yaichenbaum:
@duke7553 it builds fine but it crashes when running it.

yair100:
> @yair100 Not sure why it would fail given I didn't modify the sln file. Can you test again and provide the debug output? Is the build failing or does it fail to deploy and run?

It appears to be working after clearing and rebuilding the solution.

yaichenbaum:
@zhuxb711 we noticed you used to use pre launch in RX Explorer, did you run into any issues with support multiple instances?

zhuxb711:
@yair100 RX-Explorer removed it because the startup accelerating is not obvious.

Prelaunch is called by Windows. And whether it calls it is unpredictable. On the user side, they might not be able to aware it.

Besides, I'm not sure when my app will be suspended during prelaunch. So I have to make everything executed sync until MainPage is loaded.

I know VS could simulate prelaunch. But if you could not check this function in published app then you just introduced an unpredictable bug and it might lead to crash for unknown reason (After all, there are lots of bugs in UWP framework)

zhuxb711:
RX-Explorer removed this feature years ago. I'm not sure if this feat capable with multi-instance. RX might not support multi-instance at that time.

yair100:
> @yair100 RX-Explorer removed it because the startup accelerating is not obvious.
> 
> Prelaunch is called by Windows. And whether it calls it is unpredictable. On the user side, they might not be able to aware it.
> 
> Besides, I'm not sure when my app will be suspended during prelaunch. So I have to make everything executed sync until MainPage is loaded.
> 
> I know VS could simulate prelaunch. But if you could not check this function in published app then you just introduced an unpredictable bug and it might lead to crash for unknown reason (After all, there are lots of bugs in UWP framework)

Hopefully there won't be any surprises after publishing this in the store, I guess we'll find out how this works in the long run.

yair100 (approved):
Great work! LGTM

PR #5985, opened 2021-08-30
allow restoring tabs from previous instance using ctrl+shift+t

winston-de:
this will make it possible to reopen tabs after update, crash, etc, without enabling 

**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #issue...
- Related #issue...

**Details of Changes**
Add details of changes here.
- allow setting groups of previously opened tabs
- remove unneeded references to closed ITabItems and instead store the arguments (since those are whats used anyway)
- add previous session tabs to closed tabs list if not re-opening those tabs
- fix issue where share would be disabled for zip files

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [x] tested w/ and w/o continue where you left off setting enabled
- [ ] Tested the changes for accessibility

**Screenshots (optional)**
Add screenshots here.

PR #6255, opened 2021-09-29
Fixed an exception on fresh installing

BanCrash:
**Resolved / Related Issues**
- Related to PR #5985

**Details of Changes**
- Added a null checking to avoid an exception that were happening on Visual Studio on Files fresh installations since LastSessionPages is null on that scenario.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image


duke7553 (approved):
This looks acceptable. I ran into the same bug in my earlier testing

PR #6262, opened 2021-10-01
Made slight adjustments to data grid header to match design spec

yaichenbaum:
image

### DIFF
--- a/Files/UserControls/DataGridHeader.xaml
+++ b/Files/UserControls/DataGridHeader.xaml
@@ -27,11 +27,13 @@
             </Setter>
         </Style>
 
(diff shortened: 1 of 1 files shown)


winston-de (approved):

gave92 (approved):

gave92:
Having the header font smaller than the content's looks a bit strange tom my eyes / the header seem now too tall.. but it's the design spec.

yaichenbaum:
> Having the header font smaller than the content's looks a bit strange to my eyes / the header bar seems now too tall.. but if it's the design spec :)

The extra space helps there be a better seperation between the data grid and toolbar above it. Otherwise it just looks like section after section with only a thin line dividing them.

Issue #6116, opened 2021-09-14
Files does not correctly show year 1601 or year 9999 in the column


gitbbkload:
### Description

Hello,

I have found a problem in Files, it does not correctly show in the column (Date modified and Date created) the year 1601 or year 9999. I use NTFS as file system.

![Screenshot 2021-09-14 094228](https://user-images.githubusercontent.com/90634153/133175339-4e3f3955-f314-457c-ad9c-8f82aee901c8.png)

7-Zip shows the year 1601 and year 9999 inside the .zip file.

![Screenshot 2021-09-14 005235](https://user-images.githubusercontent.com/90634153/133116940-b6ca3c47-aff1-4319-ab5e-816b1b9bf935.png)

### Steps To Reproduce

1. [1601+9999.zip](https://github.com/microsoft/winfile/files/7155695/1601%2B9999.zip)
2. The .zip file must be extracted with [7-Zip](https://7-zip.org/download.html) because Files internal zip function cannot set the timestamps (year 1601 and year 9999) correctly.

### Expected behavior

- It correctly shows year 1601 or year 9999 in the column

- Files internal zip function can set year 1601 and year 9999 correctly during extraction

- Files internal zip function can correctly display year 1601 and year 9999 inside a .zip file for Date modified, Date created, Date accessed

- Date accessed can be added to the column so that I can display all 3 timestamps by default: Date modified, Date created, Date accessed

### Files Version

v2.0.12.0

### Windows Version

Windows 11 21H2 22000.184

BanCrash:
Is there any special reason that you want to see the year 1601? The not showing is intentional since 1601 is showed when there is no date set. On Windows File Explorer those files show 1601 too?

josh-65:
The 1601 might not be needed but if a date is set to 9999 it should show that rather than -44 seconds.

BanCrash:
> The 1601 might not be needed but if a date is set to 9999 it should show that rather than negative seconds.

I agree about the not showing negative seconds, but I would prefer the empty space also. Those dates are not real and I can't think on a scenario that they could be useful to be showed. Let's see what @gitbbkload thinks though.

yaichenbaum:
Closing as the underlying issue was resolved.

PR #6270, opened 2021-10-03
Fixes a bug that showed negative dates when the year was 9999

BanCrash:
**Resolved / Related Issues**
- Related #6116

**Details of Changes**
- Added 9999 year to the not showing list. This fixes a bug that showed negative seconds on the dates. As I stated on #6116 I think we should hide those dates (1601 and 9999 years) since they are unreal and there isn't (or at least I can't imagine one) a scenario where having those dates are useful.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before:
image

After:
image

### DIFF
--- a/Files/Extensions/DateTimeExtensions.cs
+++ b/Files/Extensions/DateTimeExtensions.cs
@@ -10,7 +10,7 @@ namespace Files.Extensions
         {
             var elapsed = DateTimeOffset.Now - d;
 
-            if (d.Year == 1601)
+            if (d.Year == 1601 || d.Year == 9999)
             {
                 return " ";
             }


yaichenbaum (approved):
LGTM

Files/Extensions/DateTimeExtensions.cs
```diff
@@ -10,7 +10,7 @@ public static string GetFriendlyDateFromFormat(this DateTimeOffset d, string ret
         {
             var elapsed = DateTimeOffset.Now - d;
 
-            if (d.Year == 1601)
+            if (d.Year == 1601 || d.Year == 9999)
```
yaichenbaum, line 13: Should we manually display the years? I can't think of a useful use case, but it might be preferable to show the year instead of an empty space.

BanCrash: > Should we manually display the years? I can't think of a useful use case, but it might be preferable to show the year instead of an empty space.

@yair100 I've followed the 1601 case (#5269) since Windows doesn't show a creation / modification date on files or folders that doesn't have a creation date:

Files before my PR that changes this on 1601 case:
image

Windows explorer on the same case:
image

I personally prefer an empty space than an unreal date as 1601 or 9999.

BanCrash:
> This PR should not be added, according to the report it is a bug, this should be fixed properly.
> 
> As I understand it, the OP has tested the date range of NTFS and this cannot display it correctly under "Files".
> 
> Here is a screenshot of Files 2.0.13 and File Explorer on Windows 11, File Explorer supports this date range (1601 and 9999 years).
> 
> Files 2 0 13 - Properties
> 
> Windows 11 File Explorer - Properties

Is there any reason that makes you be able to have those dates? 

Btw, don't know W11, but on windows 10 on the details view those files doesn't have a date neither:
image

icecoffeeabell1835:
> Is there any reason that makes you be able to have those dates?

On Windows it is generally possible to set higher or lower years with SetFile API https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfiletime.

In principle this is a bug in Files, because NTFS or ReFS support a very large date range, more than the old file systems, like FAT or EXFAT.

> Btw, don't know W11, but on windows 10 on the details view those files doesn't have a date neither:

Here also the same result under W11, in the column it is not indicated, but under "Properties" it is indicated.

My suggestion is that Files should behave like File Explorer from Windows. That is, in the column nothing is shown, but under "Properties" it is shown.

BanCrash:
> > Is there any reason that makes you be able to have those dates?
> 
> On Windows it is generally possible to set higher or lower years with SetFile API https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfiletime.
> 
> In principle this is a bug in Files, because NTFS or ReFS support a very large date range, more than the old file systems, like FAT or EXFAT.
> 

Sorry, I meant "is there any reason that makes you **want to** be able to have those dates? 

> > Btw, don't know W11, but on windows 10 on the details view those files doesn't have a date neither:
> 
> Here also the same result under W11, in the column it is not indicated, but under "Properties" it is indicated.
> 
> My suggestion is that Files should behave like File Explorer from Windows. That is, in the column nothing is shown, but under "Properties" it is shown.

We can check it, but could you explain me why do you want to be able to have those unrealistic dates? Like I said I personally prefer an empty space rather than a date that is totally random.

icecoffeeabell1835:
> 
> 
> > > Is there any reason that makes you be able to have those dates?
> > 
> > 
> > On Windows it is generally possible to set higher or lower years with SetFile API https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfiletime.
> > In principle this is a bug in Files, because NTFS or ReFS support a very large date range, more than the old file systems, like FAT or EXFAT.
> 
> Sorry, I meant "is there any reason that makes you **want to** be able to have those dates?

The lowest year I personally do not need, but some photos that have no exif data or other files I have already set to the year 1870 or 1930. The highest year I personally will not experience myself, but the next generation will experience it.

> > > Btw, don't know W11, but on windows 10 on the details view those files doesn't have a date neither:
> > 
> > 
> > Here also the same result under W11, in the column it is not indicated, but under "Properties" it is indicated.
> > My suggestion is that Files should behave like File Explorer from Windows. That is, in the column nothing is shown, but under "Properties" it is shown.
> 
> We can check it, but could you explain me why do you want to be able to have those unrealistic dates? Like I said I personally prefer an empty space rather than a date that is totally random.

Personally, I prefer that it is completely solved now, so not the same behavior as the File Explorer (only under Properties), but just show everything as the file system provides. If it's fixed now, you don't have to fix it later or the successor developer, for example when we use Windows 4000 in the year 4000.

PR #6312, opened 2021-10-05
Updated strings

yair100:


d2dyno1:
@yair100 I think building and running the app to check for missing/corrupt strings would be beneficial here which I cannot do due to Microsoft pushing updates to VS indiscriminately without running proper unit tests nor checking if everything works correctly... Could you assign someone that can actually build the app and run it who's not on v16.11.3?

winston-de (approved):

PR #6427, opened 2021-10-10
feat: Load thumbnails asap, don't check for child items with each folder

lukeblevins:
**Resolved / Related Issues**
- Reverts an optimization mentioned over discord where we defer loading visible thumbnails until the full list of items is loaded
- Only call GetLocalized() extension method on file folder type text once
- Assume all folders have child items because we don't use the value anymore (for thumbnails)

**Details**
Checking if each folder that is about to be constructed as a ListedItem has child items is currently useless, so remove this check from the item enumeration code, defaulting the property value to true, and leave the work in case it's needed down the road.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
N/A


duke7553:
Specifically, the non thumbnail related changes reduced the normal item enumeration time of C:\Windows\WinSxS from 1400ms to 192ms. That's an 86 percent reduction in the typical delay users face before all ListedItems are UI-ready!

Of course, this is more pronounced in large directories, but more than makes up for the potential slowness added with the thumbnail changes I also made above.

PR #6457, opened 2021-10-12
Respect default app for zip files

### LINKED ISSUE #6110: Only open .zip files if file explorer is set to open them by default
### What's the Problem?

I have 7-zip set to open all archives and right now files doesn't use that. It tries using itself

### Solution/Idea

It should only attempt to open an archive if file explorer is set to open that type by default.

### Alternatives

An alternative is allowing to set files as the default to open .zips in windows default app protocols. (Settings > Apps > default apps > Chose default applications by file type)

### Priorities

_No response_

### Files Version

2.0.12.0

### Windows Version

10.19043.1202

### Additional comment

_No response_

### LINKED ISSUE #6381: Doesn't respect default zip app
### Description

Opening a zip file should open it in the default app or at least have an option to not use files integrated zip opener

### Steps To Reproduce

1. Open zip file
2. Opens in files zip preview 

### Expected behavior

1. Open zip file
2. Opens in default zip app (Winrar, 7-zip) 

### Files Version

Version: 2.0.13.0 

### Windows Version

OS Version: 10.0.22000.194

### Relevant Assets

_No response_

gave92:
**Resolved / Related Issues**
- Closes #6381
- Closes #6110 

**Details of Changes**
Add details of changes here.
This PR adds the option to open zip archives in an external application e.g 7zip instead of Files.
There are 2 possible implementations to choose from:
1) a59b29b: queries the default application associated with zip files, if it's Files or explorer the internal zip viewer is used
2) 2565ef9: adds a setting under "Preferences" to use internal zip viewer or not

I'm unsure what's the better option.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files.Launcher/MessageHandlers/Win32MessageHandler.cs
+++ b/Files.Launcher/MessageHandlers/Win32MessageHandler.cs
@@ -165,6 +165,13 @@ namespace FilesFullTrust.MessageHandlers
                         }
                     }
                     break;
+
+                case "GetFileAssociation":
+                    {
+                        var filePath = (string)message["filepath"];
+                        await Win32API.SendMessageAsync(connection, new ValueSet() { { "FileAssociation", await Win32API.GetFileAssociationAsync(filePath, true) } }, message.Get("RequestID", (string)null));
+                    }
+                    break;
             }
         }
 
--- a/Files.Launcher/MessageHandlers/Win32MessageHandler.cs
+++ b/Files.Launcher/MessageHandlers/Win32MessageHandler.cs
@@ -165,13 +165,6 @@ namespace FilesFullTrust.MessageHandlers
                         }
                     }
                     break;
-
-                case "GetFileAssociation":
-                    {
-                        var filePath = (string)message["filepath"];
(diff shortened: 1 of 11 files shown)


yaichenbaum:
Is it possible to allow the user to set the default application for the zip protocol?

gave92:
@yair100 In any case I'd like to keep the setting to view zip files in the embedded viewer separated from the default zip program setting. I use both Files and Explorer: when I click a zip file in Files it's ok to open it in Files, but when I click a zip file in explorer I want to open it in explorer. Both options 1 and 2 allow this, your suggestion of changing the default zip program makes me think you have different ideas :)

Could you clarify your expectation on this? Do you expect a switch in Files settings that controls whether to use the embedded zip viewer when browsing with Files _and also_ sets Files as default for zip files?

yair100:
> @yair100 In any case I'd like to keep the setting to view zip files in the embedded viewer separated from the default zip program setting. I use both Files and Explorer: when I click a zip file in Files it's ok to open it in Files, but when I click a zip file in explorer I want to open it in explorer. Both options 1 and 2 allow this, your suggestion of changing the default zip program makes me think you have different ideas :)
> 
> Could you clarify your expectation on this? Do you expect a switch in Files settings that controls whether to use the embedded zip viewer when browsing with Files _and also_ sets Files as default for zip files?

@gave92 option 1 looks good, I was just trying to better understand the different options 👍

yair100 (approved):
LGTM

PR #6465, opened 2021-10-12
Notify when terminal config changed

### LINKED ISSUE #1140: Modification to terminal.json won't take effect immediately
**Describe the bug**
Modification to terminal.json won't take effect immediately, and changing default terminal in settings or clicking `open in terminal` will overwrite changes to terminal.json

**Solution**
Needs to use a FileWatcher to monitor terminal.json and reload it if changes were detected.

**Version used**
Files: v0.10.2

hez2010:
**Resolved / Related Issues**
- Closes #1140

**Details of Changes**
Add details of changes here.
- Notify when terminal config changed

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Controllers/TerminalController.cs
+++ b/Files/Controllers/TerminalController.cs
@@ -4,9 +4,12 @@ using Files.Filesystem;
 using Files.Helpers;
 using Newtonsoft.Json;
 using System;
+using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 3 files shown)


yaichenbaum:
> The terminals changed event triggers any time you change any of the other settings too, but that should not be a big issue.

That will actually be useful once we allow users to modify settings from the json file (#6467).

PR #6558, opened 2021-10-19
Open directory with a selected item from command line

### LINKED ISSUE #6263: Opening the folder containing the file from command line.
### What's the Problem?

Can't open folder containing the file from command line.

### Solution/Idea

Add support to open the folder containing the specified file from the command line.

Something like it is in Explorer "explorer.exe /select,"C:\Folder\subfolder\file.txt".

### Alternatives

What's with these alternatives?
It just doesn't work yet, and it would be nice if it would work.

### Priorities

| Capability | Priority |
| :---------- | :------- |

### Files Version

Wersja: 2.0.0.0 OS Version: 10.0.19043.1237

### Windows Version

Windows 10 21H1 19043.1237

### Additional comment

_No response_

abdonkov:
**Resolved / Related Issues**
- Closes #6263 

**Details of Changes**

- Added the ability to open a directory with a selected item from the command line
    - Introduced "Select" command argument, through which the name of the file or folder to select on startup can be provided (selected item is part of the opened directory which has to be provided with the "Directory" command argument)

**Known issues**
- Details Layout View doesn't scroll into view the selected item (part of a different bug - first keyboard letter press in this layout, doesn't scroll into view either)

**Validation**
- [x] Built and ran the app

Example usage:
files.exe -Direcroty "D:\Downloads" -Select "image.png"
files.exe -Directory "D:\Downloads" -Select "Some Folder"

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -406,6 +406,10 @@ namespace Files
                                     break;
 
                                 case ParsedCommandType.SelectItem:
+                                    if (Path.IsPathRooted(command.Payload))
+                                    {
+                                        await PerformNavigation(Path.GetDirectoryName(command.Payload), Path.GetFileName(command.Payload));
+                                    }
                                     break;
 
                                 case ParsedCommandType.Unknown:


neoOpus:
This works but only on small folder with few files, on my Download folder which contains hundreds of files it just open the folder but doesn't select the file... (please fix)

gave92:
@neoOpus could you check if the file gets selected, and the issue actually being that the list does not scroll down to it?

neoOpus:
> @neoOpus could you confirm whether the file gets selected, the issue actually being that the list does not scroll down to it?

you are right, it get selected but doesn't scroll to it.

Issue #6182, opened 2021-09-19
Feature: Auto refresh drive widget & tags widget when changes are detected


m1karii:
1, Enable reload button on The Home page as a way to update displayed capacity and used space of drives (so i don't have to close and reopen whole Files app or go to drive properties every time i want it updated)
2. displayed capacity and used space of drives should be automatically updated when The Home page is entered (on new tab for e.g.)

yaichenbaum:
@gave92 shouldn't we use the file watcher to automatically update the home page without requiring a refresh?

gave92:
@yaichenbaum sounds good, we can try.

mon-jai:
Coming from #14334.

Closing all windows and reopening files doesn't fix the problem.

![image](https://github.com/files-community/Files/assets/91261297/12757109-ed32-43f4-9a7f-e4a159e3e20e)

### Log

```
2024-01-06 00:13:22.1463|Information|Drive added: C:\, Fixed
2024-01-06 00:13:22.1467|Information|Drive added: C:\, Fixed
2024-01-06 00:13:22.2894|Information|Drive added: D:\, Fixed
2024-01-06 00:13:22.3175|Information|Drive added: D:\, Fixed
2024-01-06 00:13:28.2555|Information|SIDELOAD: Checking for updates...
2024-01-06 00:13:28.2596|Information|SIDELOAD: Checking for updates...
2024-01-06 00:13:28.2760|Error|System.Net.Http.HttpRequestException: No such host is known. (cdn.files.community:443)
 ---> System.Net.Sockets.SocketException (11001): No such host is known.
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
   at System.Net.Sockets.Socket.<ConnectAsync>g__WaitForConnectWithCancellation|281_0(AwaitableSocketAsyncEventArgs saea, ValueTask connectTask, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectToTcpHostAsync(String host, Int32 port, HttpRequestMessage initialRequest, Boolean async, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
[...]

PR #6542, opened 2021-10-17
Added export/import options for settings

d2dyno1:
**Resolved / Related Issues**
- Closes #1260

**Details of Changes**
- Adds options for exporting and importing settings file + fixes an issue where settings weren't respecting AppCenter.

**Known issues**
- Sidebar won't update shown sections after importing configuration. (Even though OnSettingChangedEvent is raised)

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #6577, opened 2021-10-20
Added a setting to enable or disable saving recent items opened

### LINKED ISSUE #5882: Recent items shouldn't register elements when you have it disabled
### What's the Problem?

If you have disabled recent items, when you enable it back it should show an empty list

### Solution/Idea

When you disable recent items, it should automatically clear the list and not register new recent items until you enable it back.

### Alternatives

Adding a simple setting about disabling the items history, so if you have hidden recent items but you have disabled that setting, it will still work as it does now.

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish clear the list of recent items when disabling it | Must |
| This proposal will allow developers to accomplish not register new recent items when it's disabled | Must |

### Files Version

1.9.25.0 

### Windows Version

### Additional comment

_No response_

BanCrash:
**Resolved / Related Issues**
- Closes #5882

**Details of Changes**
- Added a setting to enable or disable saving recent items opened (enabled by default)
Disabling this setting will disable also Recent files widget, and won't let you enable it back until the setting is enabled again. Will remove recent items too.
You can have the setting enable and not have the recent files widget enabled, because this setting will also affect the jump list (that's why I added a new setting instead just making this when you disable recent files widget).

TO DO:
- [ ] ~Figure out why it needs a Files restart to refresh the value of IsSavingRecentItemsEnabled in the Widgets settings.~
Unrelated to this PR since File Tags are not getting their value refreshed on the InnerNavigationToolbar neither.

- [ ] ~Figure out how to refresh all home tabs opened when this changes.~
Unrelated to this PR since this also present on Recent Files, if you have several tabs with Home opened, if you disable Recent Files it will only be update on the tab where you change it, but on the others will not be refreshed. I think this should be include in #6182 , and automatically refreshing the home page when there are widgets or drives changes. So, it's unrelated to this PR.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

Disable Recent Files widget and not letting you enable it back (currently it needs a restart of Files app):
image


BanCrash:
> Instead adding a new setting, I guess it shouldn't register recent items when the widget is disabled.

I did this way because of this:

> You can have the setting enable and not have the recent files widget enabled, because this setting will also affect the jump list (that's why I added a new setting instead just making this when you disable recent files widget).

BanCrash:
> @duke7553 brought up the idea of keeping the list in sync with the list of recent items in File Explorer, just wondering if that changes anything in regards to this PR.

@yair100 I've been thinking but I think there is nothing to change here, maybe when that is added it will modify NavigationHelper but I don't think that would require a lot of changes regarding the additions on this PR (probably just move the if condition added on this PR to the new place, but without knowing the future changes there is nothing sure).

d2dyno1 (dismissed):

yair100 (changes requested):

yair100 (commented):
Code wise this looks good, but I'm concerned about the UX of this setting. Would it be better to just add recent items to the same list File Explorer uses so you can control everything in the same place?

d2dyno1:
> About the other issue:
> 
> @d2dyno1 InnerNavigationToolbar is not getting Tags state update neither until you restart Files, If you disable Tags they will still be showed as an option of rearrange until you restart Files

It's been fixed in [`641e1d9` (#6542)](https://github.com/files-community/Files/pull/6542/commits/641e1d9b7f6d6a5da07b4b607fc19e336bba30ab)

BanCrash:
> Code wise this looks good, but I'm concerned about the UX of this setting. Would it be better to just add recent items to the same list File Explorer uses so you can control everything in the same place?

@yair100 sorry about the delay to answer, I was waiting to made some test and research but since I have exams I doesn't have much time these weeks. When I made that research and tests I will answer your question with the results that I'll get.

GabrielGarriock:
Please merge this I would really like this feature! :)

PR #6601, opened 2021-10-23
Fix StartAppCenter not called

gave92:
**Details of Changes**
Add details of changes here.
- Fixes AppCenter not working due to missing call to StartAppCenter()

@yair100 How to test if the fix actually works?

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/App.xaml.cs
+++ b/Files/App.xaml.cs
@@ -140,7 +140,7 @@ namespace Files
             TerminalController ??= new TerminalController();
         }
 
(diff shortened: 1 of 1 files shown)


yaichenbaum:
@gave92 being that that the version number in `main` is higher than the latest build in the store, just running the package from the pipeline should be enough to tell via AppCenter if the change worked.

yair100 (approved):

PR #6639, opened 2021-10-28
Custom folder icons

### LINKED ISSUE #1015: Allow setting custom folder icons
Allow setting customer folder icons

Current folder options does not allow setting customer icons

Set customer icon using windows explorer. but same settings do not apply

yaichenbaum:
Closes #1015


BanCrash:
@gave92 Is necessary to add this file: Files.Launcher/Assets/FilesOpenDialog/CustomOpenDialog64.dll ?

It seems to me that it's from the Open dialog when Files is default file explorer, is necessary to be added for some reason? (I don't mean to be added on this PR, but added to the repository)

gave92:
No it can be added to .gitignore

.gitignore
```diff
@@ -333,3 +333,4 @@ ASALocalRun/
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
 Files.Extensions/nul
+Files.Launcher/Assets/FilesOpenDialog/CustomOpenDialog64.dll
```
BanCrash, line 336:
```suggestion
```

This was already added by @gave92 in #6639

BanCrash:
The default dll is working well, but if you manually select another dll, even if it's the shell32.dll, it will throw an exception:

image

Files/Helpers/UIHelpers.cs
```diff
@@ -1,7 +1,9 @@
 ﻿using Files.Common;
+using Files.Filesystem.StorageItems;
```
BanCrash, line 2: Are this and line 6 necessary? Visual studio says no.

gave92:
@BanCrash it should not throw an exception any longer, but I was not able to reproduce the .dll failing to load the icons, is it still the case?

BanCrash:
> @BanCrash it should not throw an exception any longer, but I was not able to reproduce the .dll failing to load the icons, is it still the case?

Nope, I've tried yesterday with a custom dll downloaded from github and the windows standard dll and they worked as expected.

BanCrash (approved):
LGTM.

There are 2 things that IMO should be issued but not prevent this from be merged on its actual state.

- Customization string translation on properties will make troubles on some languages for sure. (I tested with Customization22 and this is the result:)
image
image

- On some special folders (I've tested with SMB, but I guess MTP and network folders will be affected too) it's showed the option to change the icon, but it won't work (for future reference, I think for fixing this should be added a new case to Files/Views/Pages/Properties.xaml.cs to the scenario at hand and probably a new bool like IsNetworkFolder / IsSpecialFolder to Files/Filesystem/ListedItem.cs)

yaichenbaum:
@duke7553 @gave92 thank you for your work on this.

PR #6667, opened 2021-10-31
Fix jump string not working in DetailsView

### LINKED ISSUE #6322: Jump to found folder not working
If you type a character to search in the detail view, the view does not jump to the search result. The first finding is marked, but the view does not jumpt to it.

gave92:
**Resolved / Related Issues**
- Closes #6322

**Details of Changes**
Add details of changes here.
- Fixes an issue for which typing a letter does not scroll the file list to the first matching item

Explanation: in a custom ListView template the ItemsPresenter must be the first child of the ScrollViewer or ScrollIntoView stops working

@yair100 could you check if the listview footer (shown in search page) padding is ok?

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
+++ b/Files/Views/LayoutModes/DetailsLayoutBrowser.xaml
@@ -779,6 +779,7 @@
                                                     <ScrollViewer
                                                         x:Name="ScrollViewer"
                                                         Grid.Row="1"
+                                                        Padding="0,0,8,0"
                                                         HorizontalContentAlignment="Stretch"
                                                         AutomationProperties.AccessibilityView="Raw"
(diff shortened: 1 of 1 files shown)


yaichenbaum:
@gave92 the padding is good, unfortunately this seems to have removed the blank space after the last column.

gave92:
Perhaps we can put the padding on the ScrollViewer

gave92:
Oh fuffit, it's one or the other..

yaichenbaum:
@winston-de do you have any ideas?

gave92:
Well give that the items height is constant we could just manually scroll to "item index x item height".

yaichenbaum:
> Given that the items height is constant we could just manually scroll to "item index * item height".

That should work, I would keep in mind to make it flexible since the plan is to add an option for a compact mode.

Issue #2588, opened 2020-12-12
Add the option to filter items by from the column headers in the DataGrid layout


Vixb1122:
**Is your feature request related to a problem? Please describe.**
Nope

**Describe the solution you'd like**
Nope

**Describe alternatives you've considered**
Nope

**Additional context**
![image](https://user-images.githubusercontent.com/17810492/101999644-1541a600-3d1a-11eb-9515-8c27b36f28fe.png)

In the file explorer, you can filter the entries by clicking on the arrow beside of button.

yaichenbaum:
@Vixb1122 We usually expect users to fill out the issue template when reporting a bug, please reopen this issue when you have filled out the issue template.

Vixb1122:
@yair100 reopen please.

yaichenbaum:
This issue will be resolved with #6679

Issue #4813, opened 2021-05-02
Add search filters


cinqmilleans:
**Is your feature request related to a problem?**
It is not possible to filter the search results. 

**Describe the solution**
Use a menu to filter results.
This menu, visible after the searchbar, show the filters (date, size, extension, picture, video, ...)
This menu will require research to be complete but simple, practical and beautiful. 

| Capability | Priority | 
| - | - | 
|  Show when a search is open | Must |
| Apply changes with a refresh | Must | 

**Additional comment**
I created a pull request that implements a minimal menu to better show the concept #4812.
![Filter](https://i.ibb.co/N6fS2qQ/filter.png)

yaichenbaum:
@cinqmilleans Will the option always be visible, or only when the user searches something?

cinqmilleans:
Only when the user searches something, to keep the ui as simple as possible.

yaichenbaum:
@cinqmilleans I noticed you mentioned that the user will need to do the search again, is there a way we can apply the filter without the user doing the search again?

cinqmilleans:
Not yet, but it's a good idea to add. We should keep the search bar open on the results page.

yaichenbaum:
@cinqmilleans This issue has been approved, would you like me to assign it to you?

cinqmilleans:
@yair100 yes, thanks you.

yaichenbaum:
@cinqmilleans Btw, instead of showing the filter option next to the path box, it should be in the toolbar underneath it.

cinqmilleans:
@yair100  Here is a presentation project for the search menu.

When
- A day/week/month/year ago
- Custom (display inputs from/to)

Where
- In the current folder
- In the current drive
- In all drives
- Custom (display multi-folder selector)

What
- file/document/picture/audio/video/custom
- This choice add implicit filter and display others options
	
File
- Includes folders ?
- Size (<1Kb, <1Kb, custom, ...)

Document
- Only document (txt, pdf, word, excel, libreoffice, ...)
- type with icons (pdf, ...)

Picture
- Only picture
- Multiselector type with icons (jpg, gif, bmp, ...)
- Resolution (small/medium/big/custom)
- Shooting date

Audio
- Only audio ...
- Artist

Video
- Only video

Custom
- All filters in a list (size, modification, ...)
	
I am waiting for a return of this principle before implementation.

![file](https://i.ibb.co/sCzx8qq/file.png)
[...]

PR #6679, opened 2021-11-01
Search extension

cinqmilleans:
Added a filter menu for the search box (#4813).
This pr creates the menu and the structure to manage the filters.

- Search in subfolders
- Size of file
- Date (created, modified, accessed)
- Group (and, or, not)
	
I disabled closing the search box when it loses focus, as it prevents us from using the menu search button. 

I put the menu in a button next to the SearchBox, because it didn't require modifying all of the SearchBox code.
I can move this menu (directly in the searchbox, in the toolbar, in a sidebar, ...). This will be the subject of another pr after decision. 

Improvements to be added in other pr:
- It will be necessary to change the filter icons, ideally a different one for each one.
- Filter Filename (start with, ....). The searchbox use the filename, but not the complex queries.
- Filter Kind/Extensions (audio, document, pdf, folder, ...)
- Filter Metadata (image resolution, ...)
- Others filters
- Add filters in SearchBox, by parsing. This has been removed temporarily because this pr is big. A pr with the overlay will be created after the commit.
- Add options of location (search in compressed files, in a folder list, ...)

I thank all the people who gave their vision of this menu, often with magnificent mockups. I hope I have been able to synthesize them as well as possible.


gave92:
This is looking very nice! Pushed a small change to avoid showing incorrect results when filters exclude all items and to fix a crash I was getting when hitting "Save" (SearchPicker.xaml -> <Style TargetType="ContentPresenter"> instead of `<Style TargetType="ListBoxItem">`.

yaichenbaum:
> I disabled closing the search box when it loses focus, as it prevents us from using the menu search button.

What's the menu search button?

cinqmilleans:
@gave92 Thank you for this correction.

@yair100 It's the "Search" button in the filter menu. This button starts the search without having to click on the icon in the SearchBox. Disabling the closing of the SearchBox when it loses focus allows you to use filters with a text to search, otherwise you have to type your text, you lose your text as soon as you do something.

Files/Filesystem/Search/SearchSettings.cs
```diff
@@ -0,0 +1,213 @@
...
+    public interface IHeader
+    {
+        string Glyph { get; }
+        string Title { get; }
+        string Description { get; }
+    }
+
+    public class SearchSettings : ObservableObject, ISearchSettings
+    {
+        public static SearchSettings Instance { get; } = new();
```
d2dyno1, line 50: Can you use `IServiceProvider`? (Declare in App.xaml.cs)

cinqmilleans: It is committed.

cinqmilleans:
@gave92 I have modify to use IServiceProvider. I also added a tooltip on the menu button, it was missing.

yair100:
> @gave92 Thank you for this correction.
> 
> @yair100 It's the "Search" button in the filter menu. This button starts the search without having to click on the icon in the SearchBox. Disabling the closing of the SearchBox when it loses focus allows you to use filters with a text to search, otherwise you have to type your text, you lose your text as soon as you do something.

Would it make sense to apply the filter after the flyout is closed without the need of an additional button?

cinqmilleans:
@yair100 It's an interesting idea. It makes me think that often only one filter will be used. Allowing the search to be triggered without having to click on "Save" then "Search" would be easier. We can add a "Search" button on all pages. I'll think about it.

yaichenbaum:
Personally, aside from the long list in file explorer, I think it's a pretty good UX. I think a future PR can cover an advanced option that lets users add filters quicker directly from the search box by typing something like `date-created:11/3/2021` etc.

gave92:
> Add filters quicker directly from the search box by typing

@yair100 that you can do already, anything that these filters do, you can do the same by typing.

cinqmilleans:
@gave92 I have code to parse the query and automatically create the filters. It also displays syntax explanations in the suggestions. This will allow the 2 methods to work together (manual + assistant). This will make a new pr when this one is integrated.

yaichenbaum:
@cinqmilleans can you upload screenshots?

cinqmilleans:
@yair100 You just have to ask
menu_dark
menu_date
menu_empty
menu_light
menu_size
menu_window

yaichenbaum:
@cinqmilleans looking good! I know you made some changes for usability purposes and those look good. In terms of the UI, can you try to match the design spec a little closer?
image

cinqmilleans:
@yair100 Ok, I'm doing this. Can you give me the codes used by your icons?

yaichenbaum:
Tag: `&#xE1CB;`
Grid: `&#xE152;`
Size: `&#xE2B2;`
Date: `&#xEC92;`

cinqmilleans:
@yair100 Here are some changes.

Regarding the visual:
- The SizePicker is thicker and rounded.
- The picker header has been adapted, the icon removed, and the back button is at the top.
- Styles have been improved.
- The active filter buttons have been changed, and the close button is included inside, but visible on mouse over.
- I modified the "SubFolders" checkbox. I didn't used double buttons because 1) You need 2 short texts, and it depends on the language 2) other checkboxes are planned but not yet functional.

Regarding behavior:
- It is no longer necessary to save in the pickers, it suffices to modify the value.
- Search/Clear buttons are on all pages. We can therefore do a search on a single filter without having to return to the main page.

menu2_date
menu2_light
menu2_size

yaichenbaum:
@cinqmilleans that's getting very close 🚀 
Can you try to match the font sizes and usage of the accent color?

cinqmilleans:
@yair100 Yes, I'll try.

cinqmilleans:
@yair100 Could you explain the requested change in more detail? I do not understand.

d2dyno1:
> @yair100 Could you explain the requested change in more detail? I do not understand.

Yair asked to try to match `FontSize` and `FontWeight` and where accent color is used so it looks like in the [picture](https://github.com/files-community/Files/pull/6679#issuecomment-963303359). @cinqmilleans

cinqmilleans:
@yair100 @gave92 I applied the modification.

style_light
style_dark

yaichenbaum:
@cinqmilleans is there a way to make the filter button only show when there is an active search?

yaichenbaum:
Can you also switch the arrow for a chevron?
image

cinqmilleans:
> @yair100 Can you also switch the arrow for a chevron?

Ok, I do that.

> @cinqmilleans is there a way to make the filter button only show when there is an active search?

Like @BanCrash, I think the user should be able to search on filters only (and without having to use *, which is too technical). 
I can put the filter button in the search box, like in thread #4813. It will be more integrated.

Another solution is to use a panel, as the preview pane, which contains the field and the filters. The toolbar only displays a toggle button.

yaichenbaum:
@cinqmilleans I was originally thinking it would only be used when doing searches, but I see how it can be useful to filter within a directory as well. Considering the use case, I think the way you originally had it inside the search box makes more sense.

cinqmilleans:
@yair100 I incorporated the chevron. I'll put the button back in the search box.

chevron2

yaichenbaum:
@cinqmilleans it's looking amazing! I know we requested a lot of changes but I want to let you know that I'm very impressed with all the work you've done on this.

Files/Filesystem/Search/Date.cs
```diff
@@ -0,0 +1,265 @@
+﻿using Microsoft.Toolkit.Uwp;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+
+namespace Files.Filesystem.Search
+{
+    public struct Date : IEquatable<Date>, IComparable<Date>, IFormattable
```
d2dyno1, line 9: Hmmm, what's the point of this struct?

cinqmilleans: @d2dyno1 The interest is to manage the dates, without the hours. A native Date structure has been around for a short time in .net, but not yet in uwp. This class is scheduled to disappear when the native class becomes available. It greatly simplifies the DateRange structure, which manages ranges of dates, named or not. Of specific interest are MinValue and MaxValue which correspond to days and not hours. The exact comparison is therefore possible.

cinqmilleans:
@yair100 Thank you so much. I too am impressed with how many different things you are doing for this project.
I committed the modification.

inner_close
inner_open

yaichenbaum:
@cinqmilleans thanks! Can you see if you can add back the rounded corners?

cinqmilleans:
@yair100 It is done.

inner_corner

yair100:
> @yair100 It is done.

Is it possible to do the same to the search box?

cinqmilleans:
@yair100 Do what exactly?

d2dyno1:
image

yaichenbaum:
@cinqmilleans add round corners to match how it currently looks in `main`.
image

cinqmilleans:
@yair100 It is done.

box_close
box_open

Files/Strings/en-US/Resources.resw
```diff
@@ -2755,6 +2755,123 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
...
+  </data>
+  <data name="SearchMenu_Search" xml:space="preserve">
+    <value>Search</value>
+  </data>
+  <data name="SearchFilterPage_Cancel.Content" xml:space="preserve">
+    <value>Cancel</value>
+  </data>
+  <data name="SearchMultiFilterPage_Cancel.Content" xml:space="preserve">
+    <value>Cancel</value>
+  </data>
```
yaichenbaum, line 2871: I think there is already a string for cancel, can you reuse it?

Files/UserControls/SearchBox.xaml
```diff
@@ -5,80 +5,340 @@
...
-        x:Name="SearchRegion"
-        x:Uid="NavigationToolbarSearchReigon"
-        ItemTemplate="{StaticResource SuggestionTemplate}"
-        ItemsSource="{x:Bind SearchBoxViewModel.Suggestions, Mode=OneWay}"
-        PlaceholderText="Search"
-        QuerySubmitted="SearchRegion_QuerySubmitted"
-        SuggestionChosen="SearchRegion_SuggestionChosen"
+    <AutoSuggestBox x:Name="SearchRegion" x:Uid="NavigationToolbarSearchReigon"
         Text="{x:Bind SearchBoxViewModel.Query, Mode=TwoWay}"
+        PlaceholderText="Search" UpdateTextOnSelect="False"
```
yaichenbaum, line 329:
```suggestion
        PlaceholderText="{helpers:ResourceString Name=Search}" UpdateTextOnSelect="False"
```

yair100 (changes requested):

Files/Strings/en-US/Resources.resw
```diff
@@ -2755,6 +2752,117 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
...
+  <data name="SearchSizeRange_Empty" xml:space="preserve">
+    <value>Empty</value>
+  </data>
+  <data name="SearchSize_NoLimit" xml:space="preserve">
+    <value>No limit</value>
+  </data>
+  <data name="SearchAndFilterCollection_Title" xml:space="preserve">
+    <value>And</value>
+  </data>
```
yaichenbaum, line 2781:
```suggestion
  <data name="And" xml:space="preserve">
    <value>And</value>
  </data>
```
Making this change will allow us to reuse this string in other areas.

yair100 (changes requested):

cinqmilleans:
@yair100 It's done.

yair100 (dismissed):

Files/UserControls/SearchBox.xaml.cs
```diff
@@ -7,27 +12,52 @@ namespace Files.UserControls
...
-        private void SearchRegion_Escaped(KeyboardAccelerator sender, KeyboardAcceleratorInvokedEventArgs e) => SearchBoxViewModel.SearchRegion_Escaped(sender, e);
+        private void MenuFrame_Loaded(object sender, RoutedEventArgs e)
+        {
+            navigator.SearchBox = SearchBoxViewModel;
+            navigator.Frame = sender as Frame;
+            GoRootPage();
+        }
+        private void MenuButton_Loaded(object sender, RoutedEventArgs e)
+        {
+            var allowFocusOnInteractionAvailable = ApiInformation.IsPropertyPresent("Windows.UI.Xaml.FrameworkElement", "AllowFocusOnInteraction");
```
d2dyno1, line 46: Should this be static so we don't check it every time?

Files/UserControls/Search/SizeRangeSlider.xaml.cs
```diff
@@ -0,0 +1,132 @@
...
+                    int index = int.Parse(value);
+                    var size = (index % 3) switch
+                    {
+                        1 => Range.MinSize,
+                        2 => Range.MaxSize,
+                        _ => Steps[index / 3],
+                    };
+                    ToolTipBlock.Text = size.ToString("N");
+
+                    // show fake tooltip inside popup.
```
d2dyno1, line 95: What's this for?

d2dyno1 (commented):
I must admit that this is very well done! Here are some nitpicks

Files/UserControls/Search/SizeRangeSlider.xaml.cs
cinqmilleans: Size implements IFormattable to provide many formatting options. It seems simpler and more scalable to me than having many properties or a function with an enum parameter. The downside is that it is not very readable. 

This returns the named label: the size with unit or "No Limit" for the maximum value.

Files/UserControls/SearchBox.xaml.cs
cinqmilleans: @d2dyno1 This is what I would have done for another control. But this is always displayed, and therefore this event is only triggered once. It is useless to be encumbered with a static field which will not be reused.

Files/UserControls/Search/SizeRangeSlider.xaml.cs
d2dyno1: But what's the fake tooltip for?

cinqmilleans: It is about a tooltip which is displayed when one manipulates one of the 2 cursors (min or max value). Using a real tooltip is not possible because it is displayed too late and it is not practical while moving a cursor. RangeSelector developers have therefore made their own tooltip. I took again their template, which I modified to display a modified value in the tooltip which only allowed to display a number, and to modify the appearance.

One problem with the fake tooltip is that it does not display out of control, and therefore does not display fully when on edge. So I added placement code to correct this. I had planned to propose this modification directly in the source project, so as not to have to manage it ourselves.

d2dyno1: Thanks for the explanation :+1:

d2dyno1:
Could you resolve conflicts?

cinqmilleans:
It is done.

d2dyno1:
@cinqmilleans I've found a bug. Groups are not saved on the root page:

https://user-images.githubusercontent.com/53011783/143764079-e4b81e09-28dc-4311-b9ac-32b836e88ca5.mp4

cinqmilleans:
@d2dyno1 Indeed it does not work any more. I look at it. Thank you.

cinqmilleans:
It is fix.

d2dyno1:
@cinqmilleans Still getting some issues. This time different ones 😅 
1. Adding three items to "And" group **doesn't update** the collection - i.e. shows only one item (did you forget `ObservableCollection<T>` or `INotifyPropertyChanged`?) (- the items are added but not shown)
2. Going back **correctly** displays "3 items" in the group
3. Entering the group again, UI doesn't get updated when trying to remove the first and second item only the last item is removed (did you forget `INotifyPropertyChanged`?) (- when going back (to the main page) the items are updated as they should)

https://user-images.githubusercontent.com/53011783/143770778-eb05831c-f17c-4cda-8378-4cc26e82bc12.mp4

cinqmilleans:
@d2dyno1 Thank you. The problem was in the morning's fix, which only registered a new instance instead of updating it. This implies that the new filters were added on the old instance of the group, but the new one was displayed. It's fix.

d2dyno1 (dismissed):
:tada:

yaichenbaum:
@mdtauk can you look over the UI in this PR?

yaichenbaum:
@cinqmilleans can you change the mouse cursor when hovering over the filter button to match the cursor used on other buttons?

d2dyno1:
> @cinqmilleans can you change the mouse cursor when hovering over the filter button to match the cursor used on other buttons? As a sidenote, I was comparing the options between Files and File Explorer and I feel that the UI in File Explorer is cleaner and easier to use, curious to hear your feedback on that. image

Are there options for "And", "Or", "Not"?

yaichenbaum:
@d2dyno1 there aren't.

d2dyno1:
> @d2dyno1 there aren't.

The existing design is pretty intuitive and user-friendly already. Why oversimplify it at a cost of features? There is only so much stuff you can fit inside a dropdown menu, and if we are trying to match Everything's high standards, taking away features isn't going to do good.

@yair100

yair100:
> > @d2dyno1 there aren't.
> 
> The existing design is pretty intuitive and user-friendly already. Why oversimplify it at a cost of features? There is only so much stuff you can fit inside a dropdown menu, and if we are trying to match Everything's high standards, taking away features isn't going to do good.
> 
> @yair100

@d2dyno1 anytime you start having flyouts with menus and then submenus with submenus in turn, it starts to become confusing. A simple menu with a submenu or two is a little easier to use.

d2dyno1:
> @d2dyno1 anytime you start having flyouts with menus and then submenus with submenus in turn, it starts to become confusing. A simple menu with a submenu or two is a little easier to use.

But a menu limits you and doesn't allow you to add new features without expanding beyond it. Perhaps something better could be found that is equal between these two and that doesn't sacrifice on features. @yair100

cinqmilleans:
@yair100 @d2dyno1 Great debate.
Here are some elements to help in the decision. This concerns the main page. The details pages seem to me a good compromise between ease and possibilities (Microsoft is too limited, Apple is not easy enough).

Apple offers this: https://www.youtube.com/watch?v=J09jrt9FNO8&t=5s
This system constructs sentences. It is very complete but very long to implement and visually heavy.
The number of filters is impressive and it is possible to assemble them at will.

Here is the mockup of @mdtauk
https://user-images.githubusercontent.com/7389110/124290630-93955680-db4b-11eb-9f0f-82a1fb3f9e32.png

The current implementation has flaws.
- The type of filter is not displayed in the summary. We cannot differentiate the type of date, for example.
- Important and frequent filters will be drowned in the menu among the filters of lesser importance.
- The rounded button layout is nice for short settings, but less so when the text is long. You can't always prevent a long text.

An easy to implement solution.
- We replace the rounded buttons by a title and a value (as in the image).
- The user can pin a type of filter, which will always be displayed in the menu even if empty by default.
- We initialize the list of pinned filters with the most common ones (date, size, kind).
- The additional filter menu will only be used infrequently.

cinqmilleans:
@yair100 Regarding the cursor bug, the fix is not straightforward because the cursor is changed by the AutoSuggestBox control. My attempts to force the cursor are unsuccessful on this control. I will read the controller code to find a solution. It will take a few days.

d2dyno1:
@yair100 Can this pull request be reviewed and merged without design decisions in it. This PR is too big to be constantly put off.

yair100:
> @yair100 Can this pull request be reviewed and merged without design decisions in it. This PR is too big to be constantly put off.

@d2dyno1 cinqmilleans is still working on this and it's best to wait for everything to be figured out before merging it. Unlike most features, this is a little more complicated and we're still figuring out the details on the best UX so I don't want to merge it before it's ready.

cinqmilleans:
@yair100 @d2dyno1 The improvements on this pr are
- Fix the bug with the cursor. It is progressing.
- UI for the main page. I await your feedback on my comment before implementing an improvement project.

The rest is planned for the following phases #7124.

yaichenbaum:
@d2dyno1 what kinds of features are limited by the menu?
@cinqmilleans have you looked at the search filters in the Microsoft Store? Maybe we can borrow some design principles from that.

d2dyno1:
> @d2dyno1 what kinds of features are limited by the menu?

The addition of logic gates

yaichenbaum:
> > @d2dyno1 what kinds of features are limited by the menu?
> 
> The addition of logic gates

Those are a little hard to do in a menu, have you seen how it works in Discord?

cinqmilleans:
@yair100 @d2dyno1 There are many others. The File Explorer menu does not allow you to search for a specific date, metadata, ... 
Filters in Microsoft Store are very limited. The presentation is pretty, but we chose something else. What are the advantages that you see?

d2dyno1:
> > 
> > The addition of logic gates
> 
> Those are a little hard to do in a menu, have you seen how it works in Discord?

Discord doesn't have logic gates, that's why dropdown menu should be ditched.

yaichenbaum:
@d2dyno1 what do you mean by logic gates?

yaichenbaum:
> @d2dyno1 what do you mean by logic gates? And/or?

cinqmilleans:
@yair100 @d2dyno1 Adding pinned filters results in a layout that looks a lot like a menu where you can easily choose values for frequent filters. This does not prevent having a button to add rarer filters. I had planned to offer this feature later, but it might gather your views. Parsing is planned for phase 2, and is complementary to the menu (for basic/advanced user)

d2dyno1:
> @d2dyno1 what do you mean by logic gates? And/or? Those can easily be added to a design like the one in Discord. image

I don't recall logic gates for search working on discord. Either way, having them in search box and managing them there will result in one big mess.

yaichenbaum:
@d2dyno1 it's not supported in Discord, that was just a demonstration about how it can easily be adapted. It can be a bit messy but it's cleaner than the current solution. I'm not sure the best way to move forward but I'm not a fan of putting everything in a flyout.
@cinqmilleans my apologies for dragging this PR out for so long, I just want to make sure we reach a good design before merging it.

d2dyno1:
> @d2dyno1 it's not supported in Discord, that was just a demonstration about how it can easily be adapted. It can be a bit messy but it's cleaner than the current solution. I'm not sure the best way to move forward but I'm not a fan of putting everything in a flyout.

How is it cleaner than the current solution?

cinqmilleans:
The And/Or/Not filters will be used very little. If I put them first, it's to test the solidity of the model, and because the main menu is a hidden And filter. It should not be highlighted, just available for the advanced user. It will be available in basic (flyout) and advanced (parsing) mode.

yaichenbaum:
> > @d2dyno1 it's not supported in Discord, that was just a demonstration about how it can easily be adapted. It can be a bit messy but it's cleaner than the current solution. I'm not sure the best way to move forward but I'm not a fan of putting everything in a flyout.
> 
> How is it cleaner than the current solution?

@d2dyno1 because you don't need to go through a couple different menus to find what you're looking for. Typing the simple keyword will open the exact flyout you need as opposed to going back and forth in a separate flyout. I agree with the point @cinqmilleans made about the advanced filters not being uses as often. Putting them inline in the text box like in Discord seems like a balanced solution of having the options for advanced users but not over complicating the feature for standard users.

d2dyno1:
> > How is it cleaner than the current solution?
> 
> @d2dyno1 because you don't need to go through a couple different menus to find what you're looking for. Typing the simple keyword will open the exact flyout you need as opposed to going back and forth in a separate flyout. I agree with the point @cinqmilleans made about the advanced filters not being used as often. Putting them inline in the text box like in Discord seems like a balanced solution of having the options for advanced users but not over complicating the feature for standard users.

How about we made both solutions possible - that is using words like "before" or "after" in the search box to set a filter, and also using a flyout for more advanced options (and, or, size range etc.) @yair100

yaichenbaum:
@d2dyno1 I was thinking more the other way around of the simple options in the flyout (range can be one of them) and the more advanced filters such as and/or in the text box.

d2dyno1:
How about 

> @d2dyno1 I was thinking more the other way around of the simple options in the flyout (range can be one of them) and the more advanced filters such as and/or in the text box.

That wouldn't make sense because we end up with the mess in the searchbox

yaichenbaum:
@d2dyno1 keep in mind, this would be for the advanced options that aren't used as often.

d2dyno1:
> @d2dyno1 keep in mind, this would be for the advanced options that aren't used as often.

So advanced options + logic gates in the flyout, and more common ones in the searchbox, right?

yaichenbaum:
@d2dyno1 no, the other way around 😂 
Simple logic will be handled in the flyout. Advanced logic will be handled in the search box. Perhaps adding an `and` or `or` can insert an additional filter button in the search box which in turn will open a new flyout for the user to define the new logic. This will keep the search box clean but will have support for the advanced logic while keeping everything nice and simple for everyone else.

cinqmilleans:
@yair100 @d2dyno1 
Here are more details on the implementation, including phase 2 which is already well advanced.
- The filters model is very generic. The menu pickers must be able to display all the possible cases, but not necessarily to create them.
For example, the file size filter can contain any minimum and maximum size, but the picker only allows you to select specific sizes. The size parsing does not have this limit. The picker is designed to correctly display the size received by the filter, even if it cannot be selected. If the user opens the picker on this size and modifies it, he can no longer put it back with the picker.
- When a filter is parsed in the search bar, it is deleted from the search bar and added to filters in the menu. This makes it possible not to clutter up the search bar, the small size of which does not allow you to accumulate filters. This displays the filters in a nicer, more structured and readable way.
- A badge is added to the filter menu to indicate the number of active filters (at the first level). This bagde is hidden when there is no filter. This gives feedback to the user. When his filter is removed from the search bar, the bagde appears to let him know that the filter is in the menu.
- It is clearly useful to offer more direct access to frequent filters in the menu.

We have several points of view on the interface.
- For some, the menu should be very simple, and not allow everything.
- For others, the menu should do it all but highlight frequent options.

I am of the opinion that the menu should do it all, but currently it does not sufficiently emphasize the frequent options.
If that's okay with you, I can suggest (in video) a new layout for the main page that will make the frequent options much more accessible, while making the others available.

yaichenbaum:
> I am of the opinion that the menu should do it all, but currently it does not sufficiently emphasize the frequent options.
If that's okay with you, I can suggest (in video) a new layout for the main page that will make the frequent options much more accessible, while making the others available.

I think everything can be placed inside the flyout, but we need a better design for the advanced options. Do you think you can hide the advanced operations for now so we can merge the rest since we all agree on the UX for the basic options?

d2dyno1:
Hi @cinqmilleans! Do you mind joining Files discord? https://www.discord.gg/Files. We would like to discuss the future of the search PR

d2dyno1 (dismissed):

cinqmilleans:
@d2dyno1 I added to Discord.

d2dyno1 (dismissed):

cinqmilleans:
@yair100 @d2dyno1 Here is a demo of the new main research page. Everything does not work perfectly, but it is to validate the concept.

- Some filters are pinned. The pinned filters appear first, even if they do not contain any values.
- The list of pinned filters can become a user parameter to allow him to add or remove some (only at the first level, not in the and/or/not filters). To be discussed whether this functionality will be retained.
- I changed the date filter names to match the rest of the app.
- In the add menu, I no longer display a description when it is not necessary.
- It is no longer possible to add 2x the same filter in the same parent. In the video, active or pinned filters are disabled. After some thought, we should leave them active but reach the picker of the active filter to modify it.
- I added the name of the filter in the main page. It was missing a lot. This is even more important when you need to display the filter even without a value. Just displaying the name of the filter instead is unattractive. 
- I think values ​​in a rounded button are not the best solution. It is difficult to display a long value, such as an aggregate of several values (especially for kinds). I made an example of what we can easily have by displaying more structured values. I am attaching this in pictures. 


https://user-images.githubusercontent.com/46631671/145720771-50155774-160b-4db3-ba00-1f4b8f53a7fb.mp4

black
light

yaichenbaum:
@cinqmilleans the UX pattern looks much better now.

> I think values ​​in a rounded button are not the best solution. It is difficult to display a long value, such as an aggregate of several values (especially for kinds). I made an example of what we can easily have by displaying more structured values. I am attaching this in pictures.

What if the font was made smaller?

cinqmilleans:
@yair100 I have already tried, it did not convince me. Importantly, this is inconsistent with the rest of the app, which never displays this kind of button. I will post different images to show the problem with some values too long (especially in other languages).

yair100:
> @yair100 I have already tried, it did not convince me. Importantly, this is inconsistent with the rest of the app, which never displays this kind of button. I will post different images to show the problem with some values too long (especially in other languages).

I don't prefer one way or the other, but it would be good to see what @itsmichaelwest and @mdtauk say since they worked on this design. One other important change to make is to show the options for filters on the main flyout right away so users know they can add those filters. It would look the same as if a filter was already added but without any filter applied.

src/Files/Strings/en-US/Resources.resw
```diff
@@ -2752,6 +2749,114 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
...
+  <data name="SearchPickerDateRange_NoSelected.Text" xml:space="preserve">
+    <value>No date range selected.</value>
+  </data>
+  <data name="SearchPickerDateRange_Choose.Text" xml:space="preserve">
+    <value>Or choose a date range</value>
+  </data>
+  <data name="SearchPickerSizeRange_NoSelected.Text" xml:space="preserve">
+    <value>No size range selected.</value>
+  </data>
+  <data name="SearchPickerSizeRange_Choose.Text" xml:space="preserve">
```
yaichenbaum, line 2854: Can you remove the type specific extensions so strings can be reused?

cinqmilleans: Ok, I'll do this for all the texts in this pr.

cinqmilleans:
@yair100 @d2dyno1 @itsmichaelwest @mdtauk @gave92 Here is the news of the ticket.
- I am invited to discuss it orally. My level of English does not allow me an oral discussion, but we can organize a written discussion.
- I won't have much time in the next 2 weeks to work on code. I don't think I'm the only one in this case. However, the chat can take place in December.
- I am attaching some elements of reflection to improve the interface. We can always improve the interface. All this is to be discussed.

Suggestions:
- Do not allow to have several instances of the same filter in a group has an inconvenience. It is not possible to choose separate ranges. For example: May 2018 or May 2019 or May 2020 (If we know the month but not the year). I therefore propose to authorize several instances of the same filter (via the add menu). 
- The big question is how to display the chosen parameters? I think that offering a structured display (as in the image) has advantages: more readable, more complete, allows certain sub-elements to be removed directly from the main page, displays the details of the groups (at least on one level). Consistency with the parsing mode which will offer the same possibilities. 

black

Here are some examples for dates. The dates will probably be the more complex case to manage.

2021-12-16..2021-12-16
	Today
2021-12-13..2021-12-15
	from:Monday
	to:Yesterday
2021-10-01..2021-10-31
	October
2020-10-01..2020-10-31
	October 2020
2021-04-01..2021-06-30
	from:April
	to:June
2021-01-01..2021-12-16
	from:2021
2021-04-02..2021-04-07
	from:4/2/2021
	to:4/7/2021
..2019-06-30
	to:June 2019
2021-12-10..
	from:This week
2021-01-01..2021-01-15
	from:1/1/2021
	to:1/15/2021

src/Files/Strings/en-US/Resources.resw
cinqmilleans: @yair100 It is done.

yaichenbaum:
> The big question is how to display the chosen parameters? I think that offering a structured display (as in the image) has advantages: more readable, more complete, allows certain sub-elements to be removed directly from the main page, displays the details of the groups (at least on one level). Consistency with the parsing mode which will offer the same possibilities.

I thought the tag like design from one of your earlier iterations looked fairly good. Your latest iteration in that screenshot is almost there. I would recommend hiding the filter name once a filter is added in order to take up less space and because more text can make it harder to know what you're supposed to be looking at. Adding a background color to the chosen parameters is also a common pattern to indicate that it's an element that can be removed in case the user wants to change their search query.

cinqmilleans:
@yair100 
- I don't think hiding the label when there is value is effective. The value does not indicate the filter (eg for different date filters). This is a problem in the current commited version.
- I do not understand your comment about colored backgrounds. Could you give an example in another software? 
- Being able to apply filters to unfiltered content is a suggestion to investigate. I will test. You have to look at the behavior of the subfolders. Should we hide a subfolder that does not match? You also have to pay attention to performance. I am excited about the idea if it is feasible.

yaichenbaum:
> I don't think hiding the label when there is value is effective. The value does not indicate the filter (eg for different date filters). This is a problem in the current commited version.

I meant the part that says "Size" or "Modified, not the "before" and "after".

> I do not understand your comment about colored backgrounds. Could you give an example in another software?

image

> Being able to apply filters to unfiltered content is a suggestion to investigate. I will test. You have to look at the behavior of the subfolders. Should we hide a subfolder that does not match? You also have to pay attention to performance. I am excited about the idea if it is feasible.

I guess it's up for discussion, perhaps we should finish up with this PR and come back to this question later on.

cinqmilleans:
@yair100 I understand better the idea of ​​"colored backgrounds". This control is interesting but I think it is not suitable in our case. Here are some arguments.
- Tags are suitable when there are mainly yes / no options (contains attachments, ...). This is not our case. If we use tags we will also have to replace the checkboxes (include the subfolders) to make it consistent.
- Tags are made to be displayed in lines. This implies not to display them after an icon. The icon must be part of the tag.
- Tags must be short but complete. In our case, if the user chooses "Modification date between this week and yesterday", how to display it?
"after: This week" + "before: Yesterday" (but we don't know if this is the modification or creation date)
"modified: This week - Yesterday" (but can be very long in some languages, see too long)
"{icon} This week - Yesterday" (but you need a different icon for each filter class)
- Tags are also useful when they are pre-filled. For example, in github, you can easily access a page containing the tags "is: open is: issue author: @me". We do not have any pre-filled cases.

Regarding the fact of hiding the label, I was talking about the difference between "Modification date" and "Creation date" for example. This difference should appear.

yaichenbaum:
> Tags are suitable when there are mainly yes / no options (contains attachments, ...). This is not our case. If we use tags we will also have to replace the checkboxes (include the subfolders) to make it consistent.

There are many cases where tags are used without there being an action required, in fact we already use this for the tag feature in Files for labeling Files. Another example where it's used otherwise is on GitHub, and besides, we already have an action on the flyout to apply the arguments. Can you explain what you mean by the checkboxs?

> Tags are made to be displayed in lines. This implies not to display them after an icon. The icon must be part of the tag.
Tags must be short but complete. In our case, if the user chooses "Modification date between this week and yesterday", how to display it?

The tag feature in Files and GitHub both display them in line with other text and icons. If we do move the flyout back out of the search box so that it can be used as a regular filter feature separately from the search box, we can make the flyout widget and the text in the labels a little smaller so that more can fit in the flyout at once.

> Tags are also useful when they are pre-filled. For example, in github, you can easily access a page containing the tags "is: open is: issue author: @me". We do not have any pre-filled cases.

Wouldn't the date/size/type be the prefilled data?

cinqmilleans:
@yair100 I started a demo which uses tags. I redefined the style to show more. It looks better than I expected, but still struggles to display long values. It will also be necessary to manage the groups. I will publish 2 videos, one with a structured menu, one with tags. I am using predefined tags with no values for Size and Date.

yaichenbaum:
@cinqmilleans did you try moving the flyout button out of the search box again and making the flyout wider?

cinqmilleans:
@yair100 This is not the problem. The problem is that too much tag is ugly. In French, we can have as text "Modifié: La semaine dernière - Plus tôt cette semaine".

yair100:
> @yair100 This is not the problem. The problem is that too much tag is ugly. In French, we can have as text "Modifié: La semaine dernière - Plus tôt cette semaine".

`Modifié:` doesn't have to be in the tag as that's the header and if the flyout is another 60px wide, I think we can fit a label per line in French and possibly two or three labels in English.

R3voA3:
Can we have a checkbox to automatically wrap the string that is being searched for into * ?

cinqmilleans:
@R3voA3 I do not understand your request. Could you concretely explain your suggestion?

R3voA3:
> @R3voA3 I do not understand your request. Could you concretely explain your suggestion?

Nevermind. Seems like the search does already what I want by default.

On another note, will the new UI allow users to save search patterns?

cinqmilleans:
@R3voA3 It is envisaged but not yet decided. We could add backups of search parameters, and include test cases (all empty folders, this week's documents, ...). It is not the priority, but a lot of things are planned to improve research.

cinqmilleans:
@yair100 Here are some tag mockups. I did not delete the titles because it is very detrimental to understanding. There are several versions: tags, labels, with or without delete button, with button visible on mouse over, with collapsible or pinnable filters. 

Empty
Demo1
Demo2
Demo3
Demo4
Demo5
Demo6
Demo7

yaichenbaum:
@cinqmilleans those look much closer to my vision. Have you tried making the flyout a bit wider so there is more room for the filters?

cinqmilleans:
@yair100 With this presentation, the width is not a problem. However, it can be widened a bit if needed.

I will add the modifications this weekend (tags, pinned filters, ...). The list of pinned filters will be fixed. We can add the edition later, it is desirable.

yair100:
> @yair100 With this presentation, the width is not a problem. However, it can be widened a bit if needed.
> 
> I will add the modifications this weekend (tags, pinned filters, ...). The list of pinned filters will be fixed. We can add the edition later, it is desirable.

What are pinned filters and how can you set them?

yaichenbaum:
@cinqmilleans the UI takes up a lot more vertical space than initially anticipated. I think we should consider moving it to a pane like @mdtauk originally suggested.

cinqmilleans:
@yair100
1) Regarding the panel vs flyout choice. The example images are intentionally very full but the user will rarely select so many elements. By default, there are only 2 items with no values.
If we opt for a panel, we must decide what to do with the search field:
- Leave it in the navigation bar and only use the panel for filters. It's practical, but the elements are scattered.
- Display it only in the panel. It's unified but less practical, especially with phase 2 which will provide parsing.
- Only display it in the navigation bar if the panel is closed. It is effective but can be confusing.

I can do demo screenshoots with a panel when I'm done implementing the new UI for the page.

2) Regarding the pinned filters. In the current implementation, there are 2 filters that are always visible (Size and Date Modified). The list is fixed. However, an easy evolution will be to allow pinning / unpinning certain elements on the main page (as seen on some screenshoots.

3) Regarding the implementation. I have reviewed the code completely.
- The model is more complete to provide for each filter the possibilities directly in the model. It's easier.
- For the developer, it is easier to create a new type of filter, without having to edit many switches as before. You just have to create the model and the view.
- The dates and sizes have been revised to have a more maintainable and efficient code.
- The model manages labels and tags better.
- Date tags are richer to display information in a concise way (only the month, year or day of the week when relevant).
- The ViewModel has been greatly reduced. It was too complex and redundant with the model and the view.
- The view has been clipped into more controls when it was relevant. This saves the viewmodel when it made more sense to have a dedicated controller.
- The page navigation system has been moved to one centralized service instead of being in many viewmodels.

I still need to complete the group filters UI. Sorry in advance to the reviewers who will have to reread everything.

cinqmilleans:
@yair100 Here is a video of the new version (fully implemented) and pictures of the panel version (demo). We can not put the searchbox there. I've included the edits from the previous post and the badge. To decide which version to use.

https://user-images.githubusercontent.com/46631671/149631129-1a55fbc4-f038-4eb0-9bd4-256118521e22.mp4

home
size
date
full

yaichenbaum:
@cinqmilleans I'd opt for leaving the search box in the navigation toolbar. Other than that, I think those changes are satisfactory.

cinqmilleans:
@yair100 For information, here is how I will proceed.
1) I will refactor the pane display to make it generic. This will make it easy to add new panels but only one can be visible.
2) Add the search panel and all the model layer and associated controls.
I would do 2 prs because it's easier for reviewers.

YoseeF27 (approved):

yair100:
> @yair100 For information, here is how I will proceed.
> 
> 1. I will refactor the pane display to make it generic. This will make it easy to add new panels but only one can be visible.
> 2. Add the search panel and all the model layer and associated controls.
>    I would do 2 prs because it's easier for reviewers.

@cinqmilleans great! Will you add a seperate button next to the preview pane button for filters?

cinqmilleans:
@yair100 Yes it is planned.

yaichenbaum:
@cinqmilleans can you confirm that this will resolve #2588?

cinqmilleans:
@yair100 No, this is not planned. I didn't know this feature existed in File Explorer.

This feature is not related to search (which allows searching in a set of folders). It is a simple display filter in the Detail view. You can filter any folder without doing a search. It's a nice feature, but unrelated to the SearchBar.

However, we could implement it in the SearchPanel. This would make it possible to have it in all views. This requires some thought and will not be part of phase 1.

yair100:
> @yair100 No, this is not planned. I didn't know this feature existed in File Explorer.
> 
> This feature is not related to search (which allows searching in a set of folders). It is a simple display filter in the Detail view. You can filter any folder without doing a search. It's a nice feature, but unrelated to the SearchBar.
> 
> However, we could implement it in the SearchPanel. This would make it possible to have it in all views. This requires some thought and will not be part of phase 1.

That's more or less what I meant, I think having consistency between the search and filter feature makes sense.

d2dyno1:
@cinqmilleans Could you resolve conflicting files?

Gitfjdkfl:
Un subscribe

cinqmilleans:
@d2dyno1 No, I wouldn't because this pr is no longer meant to be committed.

As mentioned above, I've completely reworked the code for more efficiency, performance and maintainability but haven't released it yet. I do not publish in this pr because the code has changed too much. It was also decided not to place the filters in the search bar but in a new right panel (like the previews). For this I created a pr, waiting for acceptance, to be able to incorporate new right panels.

When this pr is integrated, I will publish a new pr containing the search options panel and the associated model. I did not close this pr to leave the discussion on the content of the options.

YoseeF27:
> @d2dyno1 No, I wouldn't because this pr is no longer meant to be committed.
> 
> 
> 
> As mentioned above, I've completely reworked the code for more efficiency, performance and maintainability but haven't released it yet. I do not publish in this pr because the code has changed too much. It was also decided not to place the filters in the search bar but in a new right panel (like the previews). For this I created a pr, waiting for acceptance, to be able to incorporate new right panels.
> 
> 
> 
> When this pr is integrated, I will publish a new pr containing the search options panel and the associated model. I did not close this pr to leave the discussion on the content of the options.

Issue #6689, opened 2021-11-03
Improvements for custom folder icons


BanCrash:
### Description

Taken from #6639
- Customization string translation on properties will make troubles on some languages for sure. (I tested with Customization22 and this is the result:)
![image](https://user-images.githubusercontent.com/11770760/139963216-b98647c5-e9b9-4e1b-a1dd-0e346b04cf29.png)
![image](https://user-images.githubusercontent.com/11770760/139963476-cc26da00-0bcf-4b6c-983f-36fa564fb84d.png)

- On some special folders (I've tested with SMB, but I guess MTP and network folders will be affected too) it's showed the option to change the icon, but it won't work (for future reference, I think for fixing this should be added a new case to Files/Views/Pages/Properties.xaml.cs to the scenario at hand and probably a new bool like IsNetworkFolder / IsSpecialFolder to Files/Filesystem/ListedItem.cs)

### Expected behavior

- Customization string should be showed regardless the characters it has
- Customization page shouldn't be showed on folders that can't have a customized icon

### Files Version

Last main

### Windows Version

Windows 10 21H1

gave92:
Pehaps we can also merge CustomFolderIcons.xaml with PropertiesCustomization.xaml. Unless there's a plan to reuse the CustomFolderIcons control somewhere else?

yaichenbaum:
> Pehaps we can also merge CustomFolderIcons.xaml with PropertiesCustomization.xaml. Unless there's a plan to reuse the CustomFolderIcons control somewhere else? (Not much of an issue anyway)

Perhaps we can move it to the main properties page and put it in and expander control like we have in the settings dialog. The properties dialog is about due for a redesign anyway.

yaira2:
> Pehaps we can also merge CustomFolderIcons.xaml with PropertiesCustomization.xaml. Unless there's a plan to reuse the CustomFolderIcons control somewhere else? (Not much of an issue anyway)

This would be a good idea

yaira2:
@BanCrash do you want to work on this when you have some time?

BanCrash:
> @BanCrash do you want to work on this when you have some time?

Since I'm currently studying and working and past experiences with Files code, I don't think that would happen any time soon. I already tried a couple of times to fix #5993 and #6111 but didn't have any luck. So better leave to someone with more free time and knowledge.

yair100:
@BanCrash a fix will be included in the next update.

PR #6691, opened 2021-11-03
Added compatibility section to properties

### LINKED ISSUE #5663: Add compatibility section to the properties dialog
### What's the Problem?

Properties window lacks the Compatibility section.

### Solution/Idea

```markdown
I have an example below that show's the amount of sections that Files UWP currently lacks compared to the standard explorer.

example: https://i.imgur.com/BaZrmHG.png
```

### Alternatives

```markdown
⠀
```

### Priorities

_No response_

### Files Version

v1.9.17.0

### Windows Version

Windows 11 21H2 22000.100

### Additional comment

```markdown
Submitting this issue as a feature request and not a bug as the properties window is working as intended.

My only issue with the lack of the compatibility section or any other missing section is how tedious it is to open the file explorer property window especially seeing as I've gone through with 'Replacing File Explorer with Files' so in file explorer every time I navigate a folder Files opens up instead.
```

gave92:
**Resolved / Related Issues**
- Closes #5663

**Details of Changes**
Add details of changes here.
- Added Compatibility tab to properties

UI's not great but I understand there's a big redesign of Properties coming?

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.


yaichenbaum:
@gave92 do you think you can move it to the main properties page and add the options in an expander like we have in the settings dialog?

gave92:
Not sure if you meant like this :)

image

yaichenbaum:
@gave92 well I meant just the compatibility tab but I guess this will do until we have time to fully redesign the properties dialog.

yaichenbaum:
@gave92 can you remove the strings that aren't being used anymore?

yaichenbaum:
@gave92 It probably doesn't make sense to do it in this pr but we can reorganize the properties dialog to improve the ux and lay out the options a little better.

gave92:
@yair100 agreed, just let me know if you think it makes more sense to merge it like this or if you prefer to revert the last changes (going back tabs instead of expanders). 
What does the final UX look like? Expanders or tabs?

yair100:
> @yair100 agreed, just let me know if you think it makes more sense to merge it like this or if you prefer to revert the last changes (going back tabs instead of expanders).

We can go ahead with this, still got a lot of work to do myself on the settings dialog.

gave92:
Ok, then I believe this is ready.

yair100 (dismissed):

Common/CompatibilityOptions.cs
```diff
@@ -0,0 +1,97 @@
...
+        public bool DisableMaximized { get; set; } // DISABLEDXMAXIMIZEDWINDOWEDMODE
+        public bool RunAsAdministrator { get; set; } // RUNASADMIN
+        public bool RegisterForRestart { get; set; } // REGISTERAPPRESTART
+        public HighDpiOption HighDpiOption { get; set; }
+        public HighDpiOverride HighDpiOverride { get; set; }
+
+        public override string ToString()
+        {
+            var value = $"~ {OSCompatibility.GetDescription()} {ReducedColorMode.GetDescription()} {HighDpiOption.GetDescription()} {HighDpiOverride.GetDescription()} " +
+                $"{(ExecuteAt640X480 ? "640X480" : "")} {(DisableMaximized ? "DISABLEDXMAXIMIZEDWINDOWEDMODE" : "")}" +
```
d2dyno1, line 20: Could you move it to `Constants`? It's being used more than once.

Common/CompatibilityOptions.cs
```diff
@@ -0,0 +1,97 @@
...
+                $"{(RunAsAdministrator ? "RUNASADMIN" : "")} {(RegisterForRestart ? "REGISTERAPPRESTART" : "")}";
```
d2dyno1, line 21: Could you move it to `Constants`? It's being used more than once.

d2dyno1 (changes requested):

Common/CompatibilityOptions.cs
gave92: In the global "Files.Constant" class or in a Constants class private to CompatibilityOptions? Don't think this values are gonna be used anywhere else.

gave92: What am I saying, CompatibilityOptions does not belong to the Files project lol

BanCrash:
@gave92 did those 2 settings works for you? I enable them but after closing and open the properties again, they are disabled, and on Windows explorer they are disabled too:

image
image

BanCrash:
Also, old versions of Windows that are IMO the most important on compatibility issues are not showed, at least for me:

image
image

Files.Launcher/MessageHandlers/FileOperationsHandler.cs
```diff
@@ -660,6 +661,42 @@ await Win32API.SendMessageAsync(connection, new ValueSet()
                         { "PickedObject", pickedObject }
                     }, message.Get("RequestID", (string)null));
                     break;
+
+                case "ReadCompatOptions":
+                    {
+                        var filePath = (string)message["filepath"];
+                        var compatOptions = Extensions.IgnoreExceptions(() =>
+                        {
+                            using var compatKey = Registry.CurrentUser.OpenSubKey(@"SOFTWARE\Microsoft\Windows NT\CurrentVersion\AppCompatFlags\Layers");
```
gave92, line 670: @BanCrash regarding the "run as admin" option not working could you check in regedit HKCU what value is set in the above key for an .exe where you have set "run as admin" through Files or explorer?

BanCrash: I have to check later with debugger, because now without debugger both options are working. In any case, here you have:

Files: 
image

Windows explorer:
image

I'm using the Debug version of Files - Dev, but executing directly from Start menu instead using Visual Studio.

BanCrash:
@gave92 the issue happens (at least) when you enable all the settings:

image

Probably not necessary to enable all settings though. Windows registry in this case:
image

It seems that both options are missing, when I enable them individually them it works: 
image

gave92:
@BanCrash thanks indeed there was a missing space, corrected.

BanCrash (approved):
LGTM.

BTW, this also should solve the first issue of #6689

d2dyno1 (approved):

yaichenbaum:
@gave92 great work!

Issue #6686, opened 2021-11-02
Empty the recycle bin in direct access on the Recycle Bin page


cinqmilleans:
### What's the Problem?

On the Recycle Bin page, there is no button to empty the Recycle Bin.
However, there is a New button which is never used. 

### Solution/Idea

On the Recycle Bin page, replace the button "New" by "Empty".
This button is only active if the Recycle Bin is not empty. 

### Alternatives

Also add a button to restore everything.

### Files Version

v2

### Windows Version

Windows 11

### Additional comment

I made a prototype. This functionality requires little change. 
![RecycleBin](https://user-images.githubusercontent.com/46631671/139879325-ec4446a0-7ca3-4d75-b3ac-ab2cca14aad4.png)

cinqmilleans:
Ok, I'm doing this.

yaichenbaum:
Let's sync with @winston-de to figure out the best way to do it as we also want to add buttons depending on the context for other files/folders such as extracting items when an archive is selected/open.

esibruti:
@yair100 We can close this issue.

PR #6693, opened 2021-11-03
Added a toolbar button to empty recycle bin 

cinqmilleans:
**Resolved / Related Issues**
On the Recycle Bin page, there is no button to empty the Recycle Bin.
However, there is a New button which is never used. #6686

**Details of Changes**
On the Recycle Bin page, replace the button "New" by "Empty".
This button is only active if the Recycle Bin is not empty.

We can easily do the same for other pages.
@winston-de, can you give your opinion on my way of doing things? 

**Validation**
- [X] Built and ran the app


Files/Strings/en-US/Resources.resw
```diff
@@ -2791,6 +2791,9 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
   <data name="SettingsImportErrorTitle" xml:space="preserve">
     <value>Error importing settings</value>
   </data>
+  <data name="InnerEmptyRecycleBinButton.Label" xml:space="preserve">
```
yaichenbaum, line 2794:
```suggestion
  <data name="Empty" xml:space="preserve">
```
For context https://devblogs.microsoft.com/ifdef-windows/use-a-custom-resource-markup-extension-to-succeed-at-ui-string-globalization/

Files/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -54,10 +54,22 @@
                 </ResourceDictionary>
             </CommandBar.Resources>
             <CommandBar.PrimaryCommands>
+                <AppBarButton
+                    x:Name="EmptyRecycleBinButton"
+                    x:Uid="InnerEmptyRecycleBinButton"
+                    Label="Empty"
+                    IsEnabled="{x:Bind ViewModel.CanEmptyRecycleBin, Mode=OneWay}"
+                    Command="{x:Bind ViewModel.EmptyRecycleBinCommand, Mode=OneWay}"
+                    Visibility="{x:Bind ViewModel.InstanceViewModel.IsPageTypeRecycleBin, Mode=OneWay}">
```
yaichenbaum, line 63:
```suggestion
                <AppBarButton
                    x:Name="EmptyRecycleBinButton"
                    Label="{helpers:ResourceString Name=Empty}"
                    IsEnabled="{x:Bind ViewModel.CanEmptyRecycleBin, Mode=OneWay}"
                    Command="{x:Bind ViewModel.EmptyRecycleBinCommand, Mode=OneWay}"
                    x:Load="{x:Bind ViewModel.InstanceViewModel.IsPageTypeRecycleBin, Mode=OneWay}">
```
Can you add a tooltip as well?

Files/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -54,10 +54,22 @@
...
+                    <AppBarButton.Icon>
+                        <FontIcon Glyph="&#xEF88;" FontFamily="{StaticResource RecycleBinIcons}" />
+                    </AppBarButton.Icon>
+                </AppBarButton>
                 <AppBarButton
                     x:Name="NewEmptySpaceAppBarButton"
                     x:Uid="BaseLayoutContextFlyoutNew"
                     IsEnabled="{x:Bind ViewModel.InstanceViewModel.IsCreateButtonEnabledInPage, Mode=OneWay}"
+                    Visibility="{x:Bind ViewModel.InstanceViewModel.IsPageTypeRecycleBin, Mode=OneWay, Converter={StaticResource NegatedBoolToVisibilityConverter}}"
```
yaichenbaum, line 72:
```suggestion
```
I think it's best to keep this visible to ensure consistency with other directories.

yair100 (changes requested):
Looking good! I added a couple comments on things I think should be adjusted.

Files/UserControls/InnerNavigationToolbar.xaml
cinqmilleans: @yair100 The other changes look ok. But why display the New button if it is not in use? It is neither useful nor aesthetic. I suggest creating a CanCreateNew property, temporarily true if the page is not Recycle Bin. This will allow you to modify its definition for other pages later.

cinqmilleans:
@yair100 I applied the changes except showing the New button. I don't think showing it is a good idea. What does the community think?

yair100:
> @yair100 I applied the changes except showing the New button. I don't think showing it is a good idea. What does the community think?

I think it's best to avoid shifting the toolbar buttons.

cinqmilleans:
@yair100 What are we doing concretely?

yair100:
> @yair100 What are we doing concretely?

I think it's best to add the icons to the right of the toolbar for now and to add back the "new" button.

cinqmilleans:
@yair100 Ok for me. I would do that tomorrow.

cinqmilleans:
@yair100 It is commited.
I added a separator, hidden when there is no additional button. This code should make it easy to add more.
I put Visibility back instead of x: load. This caused an unpleasant visual effect, the button being active for a second before going gray when we reached an already empty Recycle Bin.

Files/Strings/en-US/Resources.resw
```diff
@@ -2779,7 +2779,13 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
   <data name="SettingsImportErrorTitle" xml:space="preserve">
     <value>Error importing settings</value>
   </data>
-   <data name="Sidebar" xml:space="preserve">
+  <data name="Empty" xml:space="preserve">
+    <value>Empty</value>
+  </data>
+  <data name="EmptyRecycleBin" xml:space="preserve">
+    <value>Empty Recycle Bin</value>
+  </data>
```
yaichenbaum, line 2787: Should we merge these and use just one or the other?

cinqmilleans: We can use "Empty Recycle Bin" for the 2. But in this case why put a tooltip?

yaichenbaum: Let's keep it how it is, good work!

cinqmilleans: After thinking it over, it's better to put the long label everywhere. This avoids having to know the context. It is modified shortly before your message.

yair100 (approved):
LGTM

PR #6657, opened 2021-10-30
Move registry usage to FTP

gave92:
**Resolved / Related Issues**
Items resolved / related issues by this PR.

As it was proven that references to Microsoft.Win32.Registry in UWP causes the app to hang on start (Store version), this PR moves functionalities that use Registry to fulltrust process (mainly "new" context menu parsing, cloud drives detection)

**Details of Changes**
Add details of changes here.
- Moved registry reading to FTP
- Removed CustomOpenDialog32.dll, CustomOpenDialog64.dll from repo and added them to .gitignore
- Fixed an issue where the fulltrust process keeps running when the main app crashes (not that it ever happens)

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #6703, opened 2021-11-04
Fix copying shortcuts, small changes to ftp

gave92:
**Resolved / Related Issues**
- Fixes an issue where files are not copied if any is a shortcut
- Fixes an issue where folder shortcuts don't open in ColumnView

**Details of Changes**
Add details of changes here.
- Fixed an issue where files are not copied if any is a shortcut
- Fixed an issue where folder shortcuts did not open in ColumnView
- Process commands in parallel in the fulltrust process
- Delay loading the "new item" context menu after the rest of the stuff is loaded (and stop using "Task.Run().Result"). Rationale: after #6657 the "new item" menu is fetched from FTP but it does not make sense to load it before the sidebar etc.)

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files.Launcher/Program.cs
+++ b/Files.Launcher/Program.cs
@@ -157,7 +157,7 @@ namespace FilesFullTrust
                 {
                     var message = info.Message.ToString().TrimEnd('\0');
 
-                    _ = Task.Run(() => Connection_RequestReceived(connection, JsonConvert.DeserializeObject<Dictionary<string, object>>(message)));
+                    OnConnectionRequestReceived(connection, JsonConvert.DeserializeObject<Dictionary<string, object>>(message));
 
                     // Begin a new reading operation
                     var nextInfo = (Buffer: new byte[connection.InBufferSize], Message: new StringBuilder());
@@ -170,7 +170,7 @@ namespace FilesFullTrust
             }
         }
 
-        private static async Task Connection_RequestReceived(PipeStream conn, Dictionary<string, object> message)
+        private static async void OnConnectionRequestReceived(PipeStream conn, Dictionary<string, object> message)
         {
             // Get a deferral because we use an awaitable API below to respond to the message
             // and we don't want this call to get cancelled while we are waiting.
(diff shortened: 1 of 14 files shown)


Files.Launcher/Program.cs
```diff
@@ -170,7 +170,7 @@ private static void EndReadCallBack(IAsyncResult result)
             }
         }
 
-        private static async void Connection_RequestReceived(PipeStream conn, Dictionary<string, object> message)
+        private static async Task Connection_RequestReceived(PipeStream conn, Dictionary<string, object> message)
```
d2dyno1, line 173: `async void`?

gave92: This method I have to call with Task.Run() so needs to return a Task.

d2dyno1: Ah right, could you break down this function into `RequestReceivedFromConnection()` and call that async Task function in Connectio_RequestReceived()

Files.Launcher/Win32API.cs
```diff
@@ -197,24 +199,31 @@ public static (string icon, string overlay) GetFileIconAndOverlay(string path, i
...
-                        overlayStr = Convert.ToBase64String(bitmapData, 0, bitmapData.Length);
+                        return (iconStr, null);
                     }
+
+                    var overlayIdx = shfi.iIcon >> 24;
+                    if (overlayIdx != 0)
+                    {
+                        var overlayImage = imageList.GetOverlayImage(overlayIdx);
+                        using var hOverlay = imageList.GetIcon(overlayImage, ComCtl32.IMAGELISTDRAWFLAGS.ILD_TRANSPARENT);
+                        if (!hOverlay.IsNull && !hOverlay.IsInvalid)
```
d2dyno1, line 215: Shouldn't there be `||` ?

gave92: Nope, hOverlay is only good if not null and not invalid

d2dyno1 (approved):
lgtm

PR #6741, opened 2021-11-05
Fixed an issue where changing the sort option would create a hidden file

### LINKED ISSUE #6739: File created when changing sorting settings of folder
Why does Files create hidden file when changing sorting settings of folder, older version was also able to change sorting settings for each folder but to remember them it didn't need to create file in that folder. This is very annoying when you have enabled show hidden files in Windows Explorer.

Files version: 2.0.33
OS version: Windows 10 21H1 19043.1320

gave92:
**Resolved / Related Issues**
- Closes #6739

**Details of Changes**
Add details of changes here.
This PR reverts a change mistakenly introduced in #6648 which causes layout mode preferences to be saved in a hidden file called "files.desktop.json" instead as an alternate data stream of the folder.

{ I was testing such a change because the current solution causes some issues for me: I have a few folders "stuck" in a layout mode I can't change anymore and the only solution is to create a new folder and copy the contents over. The file-based solution would fix this and give the users a little more control. But I didn't mean to include it in that PR and it was not discussed. }

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/ViewModels/FolderSettingsViewModel.cs
+++ b/Files/ViewModels/FolderSettingsViewModel.cs
@@ -8,7 +8,6 @@ using Microsoft.Toolkit.Mvvm.DependencyInjection;
 using Microsoft.Toolkit.Mvvm.Input;
 using Newtonsoft.Json;
 using System;
-using System.IO;
 using System.Linq;
 using Windows.Storage;
 using static Files.ViewModels.FolderLayoutInformation;
@@ -539,7 +538,7 @@ namespace Files.ViewModels
 
         private static LayoutPreferences ReadLayoutPreferencesFromAds(string folderPath)
         {
-            var str = NativeFileOperationsHelper.ReadStringFromFile(Path.Combine(folderPath, "desktop.files.json"));
+            var str = NativeFileOperationsHelper.ReadStringFromFile($"{folderPath}:files_layoutmode");
             try
             {
                 return string.IsNullOrEmpty(str) ? null : JsonConvert.DeserializeObject<LayoutPreferences>(str);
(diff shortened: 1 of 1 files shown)


d2dyno1 (approved):

d2dyno1:
@yair100 This is imperative! Otherwise users will end up with dozens of hidden `desktop.files.json` files which is happening now.

BanCrash:
@gave92 what do you think about creating that kind of files inside Files data folder? I thought about creating a json that contains all the data, but maybe in perfomance impact would be better to have each file separately.

Also since they are small files I don't think space would be an issue neither.

gave92:
@BanCrash both ADS and desktop.json inside the folder have the big advantage of not loosing layout customizations when the folder is moved or rename.

BanCrash:
> @BanCrash both ADS and desktop.json inside the folder have the big advantage of not loosing layout customizations when the folder is moved or renamed. (Cons is that they won't get cleaned up when Files is uninstalled, the only difference between ADS and desktop.json files is that users won't know about the first lol)

And what about using ADS to have the reference to the file? That will have the same issue that you were trying to solve with this change?

I mean, in ADS add that C/Documents have associated Files\Info\documents.json, if you move Documents to C/Users/User/Documents it will still use the same file since ADS won't get modified, and Files will check that ADS information already has a json file associated.

gave92:
That's a neat idea, which made me remember that files and folder on NTFS have a unique ID that does not change on move and rename. We use it for the file tags feature but could be used for what you propose as well. Would not be necessary to write to ADS at all.

BanCrash (approved):
LGTM

> That's a neat idea, which made me remember that files and folder on NTFS have a unique ID that does not change on move and rename. We use it for the file tags feature but could be used for what you propose as well. Would not be necessary to write to ADS at all. [Or we could just leave everything as is (i.e. json embedded as ADS), it may be an issue just on my machine]

@gave92 about this, unless @yaichenbaum thinks other thing IMO both actual method and the one that you say are fine so feel free to do whatever you want and / or think it's better ;) But I imagine that more people could have the same issue that you have.

gave92:
@BanCrash could be. Also it feels a little dirty to leave files (even invisible ones) splattered on a user's PC after uninstall.. this is a UWP app after all, we have some standards 😂

PR #6742, opened 2021-11-05
Updated Croatian translation

streger:
* Fixed one typo, optimized some expressions, chose "better" plural translation for "items" (there's more than one plural in Croatian for same words) and added translation for missing text


gave92 (approved):

yaichenbaum:
@streger thank you!

PR #6762, opened 2021-11-06
Set window title to be the same as the selected tab name

### LINKED ISSUE #6214: Set window name to be the same as the selected tab name
### What's the Problem?

I can't work fine if I don't want the tabs feature.. because every window name have the same name "Files"

### Solution/Idea

Hello, I am using this program with a combination of the feature Anchors in WindowTop (A new concept that I think is better than window tabs)
I don't like tabs. I prefer to have multiple windows.

The issue is that every window has the same name so this is what I am getting
image

We can see the same issue in the taskbar
image

So the problem is very simple.. the window names are always the same.

This is how it should be
image

image

I really want that every window will have the name of the folder.
Otherwise, the program can't be used nice without tabs.

Please fix the issue.
Thanks

### Alternatives

Using the default file explorer :( 

### Priorities

_No response_

### Files Version

2.0

### Windows Version

Windows 11 22000.194

### Additional comment

_No response_

### LINKED ISSUE #6749: Every Taskbar item is called Files!
If the program was limited to a single instance then it wouldn't matter but as it is it's unusable (or at least I won't be using it).

I looked for an option for this but didn't see one.

Other than the fact that it doesn't support existing Quick Access links it seems pretty good in the short amount of time I've looked at it.

Every Taskbar Item Called Files

gave92:
**Resolved / Related Issues**
- Closes #6214

**Details of Changes**
Add details of changes here.
- Set current tab header as window title so it appears on the taskbar

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/Files/ViewModels/MainPageViewModel.cs
+++ b/Files/ViewModels/MainPageViewModel.cs
@@ -17,6 +17,7 @@ using System.Threading.Tasks;
 using System.Windows.Input;
 using Windows.Storage;
 using Windows.System;
+using Windows.UI.ViewManagement;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Input;
 using Windows.UI.Xaml.Navigation;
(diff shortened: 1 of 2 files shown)


duke7553 (approved):

d2dyno1 (approved):

yaichenbaum:
@gave92 thank you!

Issue #6729, opened 2021-11-05
Can't remove FILES as the default file manager after enabling it


goalie000:
### Description

I installed FILES yesterday and tried the experimental option to make it my default file manager.  I use Box a lot and I can open the base Box folder but it just keeps refreshing over and over again and all of the subfolders are empty.  

I said "fine, that's that I guess" and I uninstalled FILES but it broke Windows File Explorer since the OS still thinks FILES is the default.  I reinstalled FILES and tried to turn off the Default option but it doesn't do anything.  

So, now i have an OS without a file explorer that works.

### Expected behavior

I would have expected that the default file manger would have reverted when I uninstalled.

### Files Version

Version: 2.0.33.0 OS Version: 10.0.22000.282

### Windows Version

windows 11 home 21h2 22000.282

gave92:
> I would have expected that the default file manger would have reverted when I uninstalled.

Unfortunately that is simply not possible (UWP apps don't have "uninstallers"), and it's one of the main reason that feature is marked as "Experimental".
To unset Files you'll need to run the "UnsetFilesAsDefaut.reg" from [this archive](https://raw.githubusercontent.com/files-community/files-community.github.io/main/data/SetFilesAsDefault.zip).
We'll see if it's at least possible to make "unset" works after a reinstall, or at least add some warning/reminder when enabling that setting.

For Box merging with #6409.

gave92:
@yair100 proposal: add a warning in the Settings page to unset Files as default before uninstalling. Not that anybody will remember it, but at least it's there ;) Also we should check if it's possible to make "unset" work after reinstall.

goalie000:
Thanks for the quick response - - the reg file fixed the default issue.  Any thoughts on why Box folders just continually refreshed at the root and the subfolders never showed their contents?

yair100:
> @yair100 proposal: add a warning in the Settings page to unset Files as default before uninstalling. Not that anybody will remember it, but at least it's there ;) Also we should check if it's possible to make "unset" work after reinstall.

@gave92 Can you remind me to do this in a couple days?

gave92:
Lol unsetting after reinstall does not work because we don't copy the reg file to the appdata folder.

koddajr:
can we have a directy download link to this .reg on tooltip on experimental (if possible), will be more easy to undo instead searching on google for example or asking here on how to.

yaichenbaum:
[...]

PR #6769, opened 2021-11-07
Fixed an issue where the option to reset the default file explorer wouldn't work after reinstalling Files

gave92:
**Resolved / Related Issues**
- Helps with #6729

**Details of Changes**
Add details of changes here.
- Fixed an issue where after reinstalling Files it was not possible to turn off the "Set Files As Default" option. The install/uninstall reg files were only copied when enabling the setting.

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

PR #6779, opened 2021-11-07
Moved startup settings to the preferences settings page

yaichenbaum:
**Validation**
- [x] Built and ran the app

**Known issues**
Startup settings combo box doesn't select the correct item when the app is opened


gave92:
In my locale the combobox appears one row below:

image

yaichenbaum:
> In my locale the combobox appears one row below:
> 
> image

@gave92 I think there is an issue with the setting control template we're using, I'll work on a fix for this in a future PR.

gave92 (approved):

vodafone999 (approved):
Approved

PR #6852, opened 2021-11-10
Increased Native Read > 4096 limit

d2dyno1:
**Resolved / Related Issues**
Fixes issues with deserialization where data exceeds buffer length.

**Details of Changes**
This PR fixes native `ReadStringFromFile()` where it could only read a limited number of characters. With this PR 4095 bytes is no longer a limit. fyi @gave92

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files/Helpers/NativeFileOperationsHelper.cs
+++ b/Files/Helpers/NativeFileOperationsHelper.cs
@@ -242,29 +242,58 @@ namespace Files.Helpers
 
         public static string ReadStringFromFile(string filePath)
         {
-            IntPtr hStream = CreateFileFromApp(filePath,
-                GENERIC_READ, 0, IntPtr.Zero, OPEN_EXISTING, (uint)File_Attributes.BackupSemantics, IntPtr.Zero);
-            if (hStream.ToInt64() == -1)
+            IntPtr hFile = CreateFileFromApp(filePath,
+                GENERIC_READ,
+                0,
+                IntPtr.Zero,
+                OPEN_EXISTING,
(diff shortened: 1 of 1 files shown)


gave92:
Nice, this this cause issues when reading settings json files?

d2dyno1:
> Nice, this this cause issues when reading settings json files?

I had the same logic in Clipboard Canvas and it did cause issues. Bringing fix for Files too since it's got the same issue.

Files/Helpers/NativeFileOperationsHelper.cs
```diff
@@ -242,29 +242,58 @@ public static bool UnsetFileAttribute(string lpFileName, FileAttributes dwAttrs)
...
                 {
-                    ReadFile(hStream, pBuff, 4096 - 1, &dwBytesRead, IntPtr.Zero);
-                    //str = Encoding.UTF8.GetString(pBuff, dwBytesRead);
+                    using (StreamReader reader = new StreamReader(msBuffer, true))
+                    {
+                        do
+                        {
+                            fixed (byte* pBuffer = buffer)
+                            {
+                                Array.Clear(buffer, 0, buffer.Length);
```
gave92, line 276: I'm fairly sure we can skip this, we only ever read up to "dwBytesRead" so we ain't gonna read garbage, right?

gave92 (approved):

Files/Helpers/NativeFileOperationsHelper.cs
d2dyno1: I just left it there to indicate that the array is reused.

PR #6858, opened 2021-11-11
Updated the nav view style on the setting dialog

yaichenbaum:
image

### DIFF
--- a/Files/Dialogs/SettingsDialog.xaml
+++ b/Files/Dialogs/SettingsDialog.xaml
@@ -15,19 +15,57 @@
         <ResourceDictionary>
             <x:Double x:Key="ContentDialogMaxWidth">800</x:Double>
             <Thickness x:Key="ContentDialogPadding">0</Thickness>
+            <Thickness x:Key="NavigationViewTopPaneHeight">40</Thickness>
+            <Thickness x:Key="TopNavigationViewTopNavGridMargin">10</Thickness>
             <SolidColorBrush x:Key="ContentDialogTopOverlay" Color="Transparent" />
             <ResourceDictionary.ThemeDictionaries>
                 <ResourceDictionary x:Key="Light">
                     <SolidColorBrush x:Key="NavigationViewExpandedPaneBackground" Color="Transparent" />
                     <SolidColorBrush x:Key="NavigationViewDefaultPaneBackground" Color="Transparent" />
+                    <SolidColorBrush x:Key="NavigationViewItemBackground" Color="{ThemeResource SolidBackgroundFillColorTertiary}" />
+                    <SolidColorBrush x:Key="TopNavigationViewItemBackgroundPointerOver"  Color="{ThemeResource SolidBackgroundFillColorTertiary}" />
+                    <SolidColorBrush x:Key="TopNavigationViewItemBackgroundPressed" Color="{ThemeResource SolidBackgroundFillColorTertiary}" />
+                    <SolidColorBrush x:Key="TopNavigationViewItemBackgroundSelected" Color="{ThemeResource SolidBackgroundFillColorTertiary}"/>
+                    <SolidColorBrush x:Key="TopNavigationViewItemBackgroundSelectedPointerOver" Color="{ThemeResource SolidBackgroundFillColorTertiary}"/>
+                    <SolidColorBrush x:Key="TopNavigationViewItemBackgroundSelectedPressed" Color="{ThemeResource SolidBackgroundFillColorTertiary}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrush" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushPointerOver" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushPressed" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushSelected" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushSelectedPointerOver" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushSelectedPressed" Color="{ThemeResource ControlStrokeColorDefault}"/>
                 </ResourceDictionary>
                 <ResourceDictionary x:Key="Dark">
                     <SolidColorBrush x:Key="NavigationViewExpandedPaneBackground" Color="Transparent" />
                     <SolidColorBrush x:Key="NavigationViewDefaultPaneBackground" Color="Transparent" />
+                    <SolidColorBrush x:Key="NavigationViewItemBackground" Color="{ThemeResource SolidBackgroundFillColorTertiary}" />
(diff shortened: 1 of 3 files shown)


Files/Dialogs/SettingsDialog.xaml
```diff
@@ -15,19 +15,57 @@
...
+                    <SolidColorBrush x:Key="TopNavigationViewItemBackgroundSelectedPointerOver" Color="{ThemeResource SolidBackgroundFillColorTertiary}"/>
+                    <SolidColorBrush x:Key="TopNavigationViewItemBackgroundSelectedPressed" Color="{ThemeResource SolidBackgroundFillColorTertiary}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrush" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushPointerOver" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushPressed" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushSelected" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushSelectedPointerOver" Color="{ThemeResource ControlStrokeColorDefault}"/>
+                    <SolidColorBrush x:Key="NavigationViewItemBorderBrushSelectedPressed" Color="{ThemeResource ControlStrokeColorDefault}"/>
```
yaichenbaum, line 36: @chingucoding the border brush doesn't seem to be working, is there a different resource I need to change as well?

chingucoding: The V2 NavigationViewItem presenter style doesn't have border setters like the V1 version has (and neither does the NavigationViewItem itsself). I think this is worth opening an issue on the WinUI repo since you cannot effectively deal with a border than. Would you like to do that or should I?

yaichenbaum: @chingucoding https://github.com/microsoft/microsoft-ui-xaml/issues/6283
Is there a way to use v1 in the meantime?

chingucoding: Not really without having to use v1 styles with larger parts of the app. You could use a custom NavigationViewItemPresenter style to add the border back though. But again, retemplating is not that great as an option.

Files/Files.csproj
```diff
@@ -1465,7 +1465,7 @@
       <Version>7.1.1</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.UI.Xaml">
-      <Version>2.7.0</Version>
+      <Version>2.8.0-prerelease.210927001</Version>
```
BanCrash, line 1468: The update to a prerelease version is intentional, right?

yaichenbaum: Yes

Files/Dialogs/SettingsDialog.xaml
```diff
@@ -52,7 +90,7 @@
 
             <muxc:NavigationView.MenuItems>
                 <muxc:NavigationViewItem
-                    AccessKey="A"
+                    AccessKey="A" CornerRadius="4,0,0,4"
```
BanCrash, line 93: Shouldn't be this on another line, like this?

```suggestion
                    AccessKey="A"
                    CornerRadius="4,0,0,4"
```

Files/Dialogs/SettingsDialog.xaml
```diff
@@ -62,7 +100,7 @@
                 </muxc:NavigationViewItem>
                 <muxc:NavigationViewItem
                     x:Uid="SettingsNavPreferences"
-                    AccessKey="P"
+                    AccessKey="P" CornerRadius="0"
```
BanCrash, line 103: If previous comment is right, then:

```suggestion
                    AccessKey="P"
                    CornerRadius="0"
```

Files/Dialogs/SettingsDialog.xaml
```diff
@@ -72,7 +110,7 @@
                 <muxc:NavigationViewItem
                     x:Uid="SettingsNavMultitasking"
                     AccessKey="M"
-                    Content="Multitasking"
+                    Content="Multitasking" CornerRadius="0"
```
BanCrash, line 113: Same as before
```suggestion
                    Content="Multitasking" 
                    CornerRadius="0"
```

Files/Dialogs/SettingsDialog.xaml
```diff
@@ -81,24 +119,21 @@
                 <muxc:NavigationViewItem
                     x:Uid="SettingsNavExperimental"
                     AccessKey="E"
-                    Content="Experimental"
+                    Content="Experimental" CornerRadius="0"
```
BanCrash, line 122: +1
```suggestion
                    Content="Experimental" 
                    CornerRadius="0"
```

Files/Dialogs/SettingsDialog.xaml
```diff
@@ -81,24 +119,21 @@
...
                     <muxc:NavigationViewItem.Icon>
                         <FontIcon Glyph="&#xF1AD;" />
                     </muxc:NavigationViewItem.Icon>
                 </muxc:NavigationViewItem>
-            </muxc:NavigationView.MenuItems>
-
-            <muxc:NavigationView.FooterMenuItems>
                 <muxc:NavigationViewItem
-                    AccessKey="B"
+                    AccessKey="B" CornerRadius="0,4,4,0"
```
BanCrash, line 129: +1
```suggestion
                    AccessKey="B" 
                    CornerRadius="0,4,4,0"
```

Files/Dialogs/SettingsDialog.xaml.cs
```diff
@@ -29,13 +29,22 @@ private void Current_SizeChanged(object sender, WindowSizeChangedEventArgs e)
 
         private void UpdateDialogLayout()
         {
-            if (Window.Current.Bounds.Height <= 600)
+            if (Window.Current.Bounds.Height <= 640)
```
BanCrash, line 32: I would change this part of the code (lines 32-39) with this:


> if (Window.Current.Bounds.Height <= 710)
            {
                ContainerGrid.Height = Window.Current.Bounds.Height - 70;
            }
            else
            {
                ContainerGrid.Height = 640;
            }


The reason is that when the window height is low it will get overlapped the x of the program itself with the settings Window, and the last field will got cut:

Before:
image
before

After:
image
After

BanCrash, line 32: Implementation as suggestion:

```suggestion
            if (Window.Current.Bounds.Height <= 710)
```

Files/Dialogs/SettingsDialog.xaml.cs
```diff
@@ -29,13 +29,22 @@ private void Current_SizeChanged(object sender, WindowSizeChangedEventArgs e)
...
             {
                 SettingsScrollViewer.Height = Window.Current.Bounds.Height - 40;
```
BanCrash, line 34:
```suggestion
                ContainerGrid.Height = Window.Current.Bounds.Height - 70;
```

Files/Dialogs/SettingsDialog.xaml.cs
```diff
@@ -29,13 +29,22 @@ private void Current_SizeChanged(object sender, WindowSizeChangedEventArgs e)
...
             }
             else
             {
-                SettingsScrollViewer.Height = 560;
+                SettingsScrollViewer.Height = 600;
```
BanCrash, line 38:
```suggestion
                ContainerGrid.Height = 640;
```

BanCrash (approved):
LGTM

PR #6931, opened 2021-11-15
Fix build pipeline for x86 and x64 msix packages

gus33000:
**Details of Changes**
Due to some mistake done by the Powershell Team in the ```Microsoft.Management.Infrastructure``` nuget package, x64 and x86 compiles of the application when performed by the CI task do not produce expected binaries, leading to the app being non functional on these platforms.

image

The ```Microsoft.Management.Infrastructure.Runtime.Win``` package uses a mismatch of RIDs (Runtime Identifiers) with ```win10-x64```/```win10-x86``` being used versus ```win-arm``` and ```win-arm64```. Due to another issue with Windows Application Packaging projects, the files only get copied if the following RID scheme is used: ```win-XXX```.

This PR therefore brings a workaround to this problem by bundling a local fixed version of the ```Microsoft.Management.Infrastructure.Runtime.Win``` nuget package, and adjusts the CI to restore said Nuget package.

Given MSBuild ability to restore nuget packages both from nuget.org and from local sources, the Nuget installer/Restore tasks were removed from the CI pipelines, and instead msbuild was instructed to restore packages by itself. This allows us to bring in the fixed package.

**Validation**
- [x] Built and ran the app
- [ ] Tested the changes for accessibility (does not apply here)
- [x] Test CI pipeline (non release) and sideloaded/ensured the built package was working on x64 Windows. 

**Screenshots**
image


duke7553 (approved):
We're going to go ahead with this temporary solution to get some packages built while I investigate not using MI or WMI in future PRs.

yaichenbaum (approved):

SuperJMN:
My Microsoft Store application is failing because if this problem. Please, fix this.

SuperJMN:
OK; I'm impacted with the MMI issue. Is there any way to get this sorted out in the official packages? Thanks!

yaichenbaum:
@SuperJMN v2.1 will be released when it's ready, but it depends on a couple things that will take some time to implement.

PR #6985, opened 2021-11-18
Fixed an issue where restoring items from recycle bin would remove the file extension

### LINKED ISSUE #6709: Restore from Recycle Bin loses filetype
### Description

When attempting to restore a file from the recycle bin, the recovered file loses its extension and is simply listed as "File". Therefore you may not be able to tell what the application association is, especially if the file was downloaded or emailed to you.

### Steps To Reproduce

1. Navigate/Create a temp folder such C:\Temp.
2. Create a new file such as MS Word Rich Text Format, i.e. "Delete Test.rtf"
3. Place a string inside "Hello World" or similar.
4. Delete the file so it goes to the bin.
5. Restore the file and notice how the file is restored only as "Delete Test" without the file extension.

### Expected behavior

Restore as normal File Explorer would be full filename intact.

### Files Version

Version: 2.0.33.0

### Windows Version

Windows 11 21H2 10.0.22000.282

### Relevant Assets

_No response_

d2dyno1:
**Resolved / Related Issues**
- Closes #6709

**Details of Changes**
- File extensions would not show in recycle bin even when the option is turned on in Files. Disabling Show file extensions in File Explorer causes this issue, this PR fixes it. 

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files.Launcher/DeviceWatcher.cs
+++ b/Files.Launcher/DeviceWatcher.cs
@@ -41,9 +41,9 @@ namespace FilesFullTrust
         private async void DeviceModifiedEvent(object sender, EventArrivedEventArgs e)
         {
             CimInstance obj = e.NewEvent.Instance;
-            var deviceName = (string)obj.CimInstanceProperties["Name"].Value;
-            var deviceId = (string)obj.CimInstanceProperties["DeviceID"].Value;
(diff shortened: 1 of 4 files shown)


gave92:
Wow thanks, i would have never found that. Enabling file extensions in file explorer is the first thing I do upon windows install xD

gave92 (approved):

PR #6994, opened 2021-11-19
Rolled back to WinUI 2.7

yaichenbaum:
The WinUI 2.8 preview doesn't currently work in the store with apps that support ARM64 so this is a temporary workaround.


d2dyno1 (approved):
Is the winui/ms team aware of the store issue?

yaichenbaum:
@d2dyno1 they already resolved the issue on their end.

gus33000:
Just a note: WinUI 2.7 also does not have the fix, the fix is only currently present on WinUI 2.7.1 pre-release, so this PR will not address the issue.

image

PR #7098, opened 2021-11-27
Remove async when not needed

meziantou:
**Details of Changes**
Using `async` keyword generate a state machine which has perf impact. I removed it from methods where it's trivial to remove them and doesn't impact the readability of the code


duke7553 (approved):
This pattern looks superior to what we had. Thanks!

gave92 (approved):

yaichenbaum:
@meziantou thank you! LGTM

PR #7103, opened 2021-11-27
Move OrderBy after Where

meziantou:
**Details of Changes**
Performance improvements as `OrderBy` will operate on a smaller collection.


yaichenbaum:
@meziantou thank you!

PR #7107, opened 2021-11-28
Replace regex with string operations

meziantou:
**Details of Changes**
Quick perf optimization :)


duke7553 (approved):
Definitely would be more efficient

winston-de (approved):
lgtm

PR #7110, opened 2021-11-28
Use async/await to read from the named pipe

meziantou:
**Details of Changes**
- Simplify the code by using `ReadAsync`
- Fix a logic issue: Using `UTF8.GetString` on an incomplete buffer may generate invalid strings ([example](https://www.meziantou.net/how-not-to-read-a-string-from-an-utf-8-stream.htm)). The new code is using a `MemoryStream` to get the message content until the message is complete.

### DIFF
--- a/Files.Launcher/Program.cs
+++ b/Files.Launcher/Program.cs
@@ -109,69 +109,34 @@ namespace FilesFullTrust
                 Logger.Warn(ex, "Could not initialize pipe!");
             }
 
-            if (connection.IsConnected)
-            {
-                var info = (Buffer: new byte[connection.InBufferSize], Message: new StringBuilder());
-                BeginRead(info);
-            }
-            else
-            {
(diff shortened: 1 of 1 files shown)


gave92:
Can the same be done on the ("other side")[https://github.com/files-community/Files/blob/main/Files/Helpers/AppServiceConnectionHelper.cs] of the pipe connection?

meziantou:
> Can the same be done on the [other side](https://github.com/files-community/Files/blob/main/Files/Helpers/AppServiceConnectionHelper.cs) of the pipe connection?

Yes, I can do it. I prefer to keep MR small. So, I can do it in another MR as the two changes can be merged independently.

yaichenbaum (approved):

meziantou:
The follow-up PR: https://github.com/files-community/Files/pull/7130

gave92 (approved):

PR #7112, opened 2021-11-28
Do not release the Semaphore if it was not acquired first

meziantou:
You should release the semaphore only if you have acquired it first. In case an exception occurs when acquiring it, the previous code was releasing it anyway.


yaichenbaum (approved):
LGTM

PR #7113, opened 2021-11-28
Dispose the BlockingCollection in the Dispose method

meziantou:


duke7553 (approved):
Thank you for catching this 👍🏻

PR #7115, opened 2021-11-28
Fix ToByteArrayAsync to read the result of ReadAsync

meziantou:
`ReadAsync` doesn't guarantee to read the _requested_ bytes. The contract is to read at least 1 byte. So, you need to iterate until the stream is fully read.

### DIFF
--- a/Files/Extensions/ImageSourceExtensions.cs
+++ b/Files/Extensions/ImageSourceExtensions.cs
@@ -20,25 +20,21 @@ namespace Files.Extensions
 
         private static async Task<byte[]> ToByteArrayAsync(this Stream stream, CancellationToken cancellationToken = default)
         {
-            MemoryStream memoryStream = null;
-            try
+            MemoryStream memoryStream;
+            if (stream.CanSeek)
             {
-                if (stream.CanSeek)
-                {
-                    var length = stream.Length - stream.Position;
-                    memoryStream = new MemoryStream((int)length);
-                }
-                else
-                {
-                    memoryStream = new MemoryStream();
-                }
+                var length = stream.Length - stream.Position;
+                memoryStream = new MemoryStream((int)length);
+            }
+            else
+            {
+                memoryStream = new MemoryStream();
+            }
 
+            using (memoryStream)
+            {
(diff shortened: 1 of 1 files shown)


Files/Extensions/ImageSourceExtensions.cs
```diff
@@ -12,10 +14,43 @@ internal static async Task<byte[]> ToByteArrayAsync(this IInputStream stream)
...
+                    return Array.Empty<byte>();
+
+                var buffer = new byte[length];
+                var actualLength = await TryReadAllAsync(stream, buffer, cancellationToken).ConfigureAwait(false);
+                Array.Resize(ref buffer, actualLength);
+                return buffer;
+            }
+
+            using var memoryStream = new MemoryStream();
+            await stream.CopyToAsync(memoryStream, bufferSize: 81920, cancellationToken).ConfigureAwait(false);
```
gave92, line 37: Curious: why do you distinguish between a seekable and non-seekable stream? Isn't the second part ok for both cases?

meziantou: It's for perf reasons. The MemoryStream may need to grow its local array multiple times (and copy data from the old array to the new one) when copying the original stream. If you know the length of the stream in advance, you can allocate the array once and fill it. => So, you reduce the number of allocations

gave92: What about using `memoryStream = new MemoryStream(LENGTH)`?

meziantou: I can do it using the MemoryStream if you think it's easier to maintain. In terms of perf, it will be a little bit slower (1 more allocation and a more complex logic in the `MemoryStream.Write` method), but the difference should be negligeable in this case

gave92: Thanks for the explanation, I'd go with the simple method (new MemoryStream(stream.Length)) here.

meziantou: The latest commit should be ok 🙂

gave92 (approved):
Thank you!

gave92 (approved):

yaichenbaum (approved):
LGTM

PR #7123, opened 2021-11-28
Avoid reading the whole file into memory when adding a file to a .zip archive

gave92:
**Resolved / Related Issues**
- For #45

**Details of Changes**
Add details of changes here.
- Avoid reading the whole file into memory when adding a file to a .zip archive

Unfortunately this requires some trivial changes to SharpZipLib which currently does not support opening a ZipOutputStream to an existing archive. Modified nuget package added to the project, like it was done for `microsoft.management.infrastructure.runtime.win`

To test: copy a 1GB and paste it into a .zip archive. See Files.exe memory not-growing anymore to 1GB :)

**Validation**
- [x] Built and ran the app


d2dyno1:
Does this fix zip bombs?

gave92:
My understanding is that zip bombs are about extracting archives, this reduces memory usage when _compressing_ files.

gave92:
But now I'm curious to test what happens extracting the famous `42.zip` in Files ;)

gave92:
Ah nevermind: those kinds of zip bombs only work if you try recursively unzip nested archives (which we don't). This one (non recursive) works though: https://www.bamsoftware.com/hacks/zipbomb/. It's a 42Kb file that gets expanded to 5Gb, but the only consequence is that you occupy some disk space and you can stop extraction at any moment.

yaichenbaum:
@gave92 would it make sense to switch to using 7zip?

gave92:
@yair100 we can look at integrating 7zip through e.g [SevenZipExtractor](https://github.com/adoconnection/SevenZipExtractor). I suggest we still merge this as it's beneficial for FTP too. Bundling a custom SharpZipLib is also not too bad if it's temporary (till we switch to 7zip).

d2dyno1:
@yair100 I'm going to review it next weekend. If you want to, you can review it now.

Files/Filesystem/StorageItems/StreamWithContentType.cs
```diff
@@ -10,74 +10,98 @@ namespace Files.Filesystem.StorageItems
...
         {
-            this.stream = stream.AsInputStream();
+            this.stream = stream;
+            this.iStream = stream.AsInputStream();
         }
 
         public IAsyncOperationWithProgress<IBuffer, uint> ReadAsync(IBuffer buffer, uint count, InputStreamOptions options)
         {
-            return stream.ReadAsync(buffer, count, options);
+            return iStream.ReadAsync(buffer, count, options);
```
d2dyno1, line 25: `stream.AsInputStream().ReadAsync(...)`?

d2dyno1 (approved):
From initial look, it looks ok!

Files/Filesystem/StorageItems/StreamWithContentType.cs
gave92: In the constructor I dot this `this.iStream = stream.AsInputStream();` to avoid calling AsInputStream() at every read operation.

yair100 (approved):

PR #7126, opened 2021-11-28
Mousewheel click opens new tab at selected path

### LINKED ISSUE #3667: Click with Wheel on Subpath on PathBar -> Open in new Tab
When I click with the mouse wheel on a subpath, the field changes to edit mode. It would be greate if a new tab would be open.
Same behaviour could be added to the navigation buttons "back" and "up".

Change the behaviour. 

?

**Additional context**
Example: At the moment files shows the folder "C:/Users/max/Documents/work/importan" . Now I want to open "C:/Users/max/Documents/work" in a new tab. I have to go up and click with the mouse wheel on the subfolder to open it again. It would be faster if i could open an parent folder (or the parent of the parent ...) directly in a new tab.

Raithlin:
**Resolved / Related Issues**
- Closes #3667

**Details of Changes**
Add details of changes here.
- Added method to capture mouse clicks. Middle clicks are filtered and a new tab opened.

**Validation**
- [X ] Built and ran the app
- [X ] Tested the changes for accessibility

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/Files.Launcher/Win32API.cs
+++ b/Files.Launcher/Win32API.cs
@@ -21,553 +21,553 @@ using Windows.System;
 
 namespace FilesFullTrust
 {
-	internal class Win32API
-	{
-		public static Task<T> StartSTATask<T>(Func<T> func)
-		{
-			var tcs = new TaskCompletionSource<T>();
-			Thread thread = new Thread(() =>
-			{
-				Ole32.OleInitialize();
-				try
-				{
-					tcs.SetResult(func());
-				}
-				catch (Exception ex)
-				{
-					tcs.SetResult(default);
-					Program.Logger.Warn(ex, ex.Message);
-					//tcs.SetException(e);
-				}
-				finally
-				{
-					Ole32.OleUninitialize();
-				}
-			})
-			{
-				IsBackground = true,
-				Priority = ThreadPriority.Normal
-			};
-			thread.SetApartmentState(ApartmentState.STA);
-			thread.Start();
-			return tcs.Task;
-		}
+    internal class Win32API
+    {
+        public static Task<T> StartSTATask<T>(Func<T> func)
+        {
+            var tcs = new TaskCompletionSource<T>();
+            Thread thread = new Thread(() =>
+            {
+                Ole32.OleInitialize();
+                try
+                {
+                    tcs.SetResult(func());
+                }
+                catch (Exception ex)
+                {
+                    tcs.SetResult(default);
+                    Program.Logger.Warn(ex, ex.Message);
+                    //tcs.SetException(e);
+                }
+                finally
+                {
+                    Ole32.OleUninitialize();
+                }
+            })
+            {
+                IsBackground = true,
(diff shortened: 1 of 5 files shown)


Files.Launcher/Win32API.cs
```diff
@@ -21,553 +21,553 @@
...
+				return new Win32Window() { Handle = new IntPtr(hwnd) };
+			}
+		}
+
+		// Get information from recycle bin.
+		[DllImport(Lib.Shell32, SetLastError = false, CharSet = CharSet.Auto)]
+		public static extern int SHQueryRecycleBin(string pszRootPath,
+			ref SHQUERYRBINFO pSHQueryRBInfo);
+	}
 }
```
yaichenbaum, line 573: Can you fix the spacing?

yaichenbaum, line 573: Can you fix the tab spacing?

Raithlin: I reverted the change to Win32Api entirely. It is not a part of this commit.

yaichenbaum (approved):

BanCrash:
@Raithlin do you know why sometimes it opens the same folder in both new tab and actual tab? I can't replicate consistently, but it happens some times in my tests without any apparently reason.

GIF 02-12-2021 14-41-20

Raithlin:
I had that happen once, thought I had clicked on something else. I can take a look - most likely the click method is being called as well. I can't check the mouse status in that method either, so it makes it a little more difficult to prevent. Any ideas?

Raithlin:
I added logging (temporarily) to see which methods were being called. Both were. 
I changed the process to use a private pointerroutereventargs to pass to the tapped event as per https://stackoverflow.com/questions/44111224/pointerpressed-vs-tapped#44125921. No click through happening any more, and the middle click still works.

Files/ViewModels/NavToolbarViewModel.cs
```diff
@@ -604,9 +606,36 @@ public void VisiblePath_QuerySubmitted(AutoSuggestBox sender, AutoSuggestBoxQuer
             (this as INavigationToolbar).IsEditModeEnabled = false;
         }
 
-        public void PathBoxItem_Tapped(object sender, TappedRoutedEventArgs e)
+        public async Task PathBoxItem_PointerPressed(object sender, PointerRoutedEventArgs e)
```
BanCrash, line 609: Any reason for this to be an async Task method? IMO a public void would be enough.

Raithlin: I was following best practice (and Roslyn analyser recommendations) - but you are right. It isn't being called by an async method, so either way it doesn't make any difference.

BanCrash: Ok, then:

```suggestion
        public void PathBoxItem_PointerPressed(object sender, PointerRoutedEventArgs e)
```

BanCrash (approved):
LGTM

PR #7130, opened 2021-11-29
Use ReadAsync in NamedPipeServerStream

meziantou:
Follow-up of https://github.com/files-community/Files/pull/7110


meziantou:
The `CodeFactor` check failed, but the file already contains two classes before the change. Do you want me to move one of the classes to its own file?

gave92 (approved):

PR #7142, opened 2021-11-29
Fix select search suggestion

cinqmilleans:
**Resolved / Related Issues**
The selection in search suggestions has a problem. The file is opened by selecting the item. If there are several suggestions and you want to open the second, moving with the keyboard will open the first by passing over it, and you cannot open the second.

**Details of Changes**
The event SuggestionChosen is removed. The event QuerySubmitted opens the file when he receives a suggestion.

**Validation**
- [ ] Built and ran the app

**Screenshots (optional)**
before/after for select the second item on the keyboard.

https://user-images.githubusercontent.com/46631671/144091046-0cac7c4f-97a6-401d-a758-d0de99f4bad1.mp4

https://user-images.githubusercontent.com/46631671/144091114-1535b239-fd5b-4c3d-bd19-8656671f2e9b.mp4

### DIFF
--- a/Files/Views/ColumnShellPage.xaml.cs
+++ b/Files/Views/ColumnShellPage.xaml.cs
@@ -174,9 +174,7 @@ namespace Files.Views
             NavToolbarViewModel.EditModeEnabled += NavigationToolbar_EditModeEnabled;
             NavToolbarViewModel.ItemDraggedOverPathItem += ColumnShellPage_NavigationRequested;
             NavToolbarViewModel.PathBoxQuerySubmitted += NavigationToolbar_QuerySubmitted;
-
             NavToolbarViewModel.SearchBox.TextChanged += ColumnShellPage_TextChanged;
-            NavToolbarViewModel.SearchBox.SuggestionChosen += ColumnShellPage_SuggestionChosen;
             NavToolbarViewModel.SearchBox.QuerySubmitted += ColumnShellPage_QuerySubmitted;
 
             NavToolbarViewModel.InstanceViewModel = InstanceViewModel;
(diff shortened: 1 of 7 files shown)


d2dyno1:
Video previews aren't working

cinqmilleans:
@d2dyno1 It's reuploaded.

Files/Views/ColumnShellPage.xaml.cs
```diff
@@ -280,26 +278,25 @@ private async void ColumnShellPage_TextChanged(ISearchBox sender, SearchBoxTextC
...
                 {
-                    NavPathParam = e.SelectedSuggestion.ItemPath,
-                    AssociatedTabInstance = this
-                });
-            }
-            else
-            {
-                // TODO: Add fancy file launch options similar to Interactions.cs OpenSelectedItems()
-                await Win32Helpers.InvokeWin32ComponentAsync(e.SelectedSuggestion.ItemPath, this);
+                    ItemDisplayFrame.Navigate(typeof(ColumnViewBase), new NavigationArguments()
```
d2dyno1, line 287: Huh? There already exists a helper method that opens a folder or a file [here](https://github.com/files-community/Files/blob/818565547cf19d675c7daebbe284ec89a729ed6a/Files/Helpers/NavigationHelpers.cs#L126)

Files/Views/ModernShellPage.xaml.cs
```diff
@@ -279,26 +276,25 @@ private async void ModernShellPage_TextChanged(ISearchBox sender, SearchBoxTextC
...
                 {
-                    NavPathParam = e.SelectedSuggestion.ItemPath,
-                    AssociatedTabInstance = this
-                });
-            }
-            else
-            {
-                // TODO: Add fancy file launch options similar to Interactions.cs OpenSelectedItems()
-                await Win32Helpers.InvokeWin32ComponentAsync(e.SelectedSuggestion.ItemPath, this);
+                    ItemDisplayFrame.Navigate(InstanceViewModel.FolderSettings.GetLayoutType(item.ItemPath), new NavigationArguments()
```
d2dyno1, line 285: Same here

d2dyno1 (changes requested):

Files/Views/ModernShellPage.xaml.cs
cinqmilleans: @d2dyno1  This comment is not mine. This is the code for the SuggestionChosen event that was moved to the QuerySubmitted event when the suggestion is not null. Replace the call or remove the comment is another problem than the purpose of this pr. This code seems operational. Should this comment be deleted?

cinqmilleans: @d2dyno1 Can you detail your remark. Is the problem the code or the comment?

d2dyno1: I meant that a convenient function `OpenPath()` found in `NavigationHelpers` could be used here (instead of `ItemDisplayFrame.Navigate(...)`. Though, I'm not 100% certain that it should be replaced, could you check if `NavigationHelpers.OpenPath()` works here?

cinqmilleans: @d2dyno1 I made the modification, it works perfectly. This simplifies the code because you no longer need to distinguish between files and folders, this is already handled by OpenPath.

d2dyno1: Thank you!

d2dyno1 (approved):
Looks good!

PR #7162, opened 2021-11-30
Change return type of AddIfNotPresent

cinqmilleans:
**Resolved / Related Issues**
EnumerableExtensions.AddIfNotPresent returns a Boolean. This result is never used.
This makes sense, because the purpose of this function is to avoid testing.

**Details of Changes**
This pr change the return type by the input (list or dictionary).
This will allow this function to be placed in a modification chain, like LINQ

**Validation**
- [ ] Built and ran the app

### DIFF
--- a/Files/Extensions/EnumerableExtensions.cs
+++ b/Files/Extensions/EnumerableExtensions.cs
@@ -43,30 +43,22 @@ namespace Files.Extensions
             return block.Completion;
         }
 
-        public static bool AddIfNotPresent<T>(this IList<T> list, T element)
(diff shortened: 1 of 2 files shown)


d2dyno1:
I haven't thought of that. Really nice! Are there any more places where this could ne used?

cinqmilleans:
Je ne sais pas, mais probablement.

d2dyno1:
While at it, could you update other places too?

cinqmilleans:
It's done. I only found one interesting case in the Extension folder (AddSorted).

d2dyno1 (approved):

PR #7166, opened 2021-11-30
Fix inner toolbar state on startup

cinqmilleans:
**Resolved / Related Issues**
Before, on startup, the icons are actives in the inner toolbar and the EmptyRecycleBin button is visible.
They are disabled when the viewmodel is loaded. It is not beautiful and it will become even less so when we add buttons only visible on certain pages. 

**Details of Changes**
Add Fallback in the bindings.

**Validation**
- [ ] Built and ran the app


yaichenbaum (approved):
Looks good, lgtm

PR #7191, opened 2021-12-02
Add action on label for change layout

cinqmilleans:
**Resolved / Related Issues**
In UI design, it is recommended to be able to activate a radio button by clicking on its label. It is not possible to change the layout. I always have the reflex to click on the label instead of the button.

**Details of Changes**
Add a click event on the labels to activate the change layout command. 

**Validation**
- [ ] Built and ran the app

**Screenshots (optional)**

https://user-images.githubusercontent.com/46631671/144438854-6deef908-9178-40d2-ab33-f3115b5a1e05.mp4

### DIFF
--- a/Files/UserControls/InnerNavigationToolbar.xaml
+++ b/Files/UserControls/InnerNavigationToolbar.xaml
@@ -486,10 +486,12 @@
                                 </RadioButton>
 
                                 <TextBlock
+                                    x:Name="NavToolbarDetailsHeader"
                                     x:Uid="NavToolbarDetailsHeader"
                                     Grid.Row="0"
                                     Grid.Column="1"
                                     VerticalAlignment="Center"
+                                    PointerReleased="NavToolbarDetailsHeader_PointerReleased"
                                     Text="Details" />
 
                                 <RadioButton
@@ -513,10 +515,12 @@
                                 </RadioButton>
 
                                 <TextBlock
+                                    x:Name="NavToolbarTilesHeader"
                                     x:Uid="NavToolbarTilesHeader"
                                     Grid.Row="1"
                                     Grid.Column="1"
                                     VerticalAlignment="Center"
+                                    PointerReleased="NavToolbarTilesHeader_PointerReleased"
                                     Text="Tiles" />
 
                                 <RadioButton
@@ -540,10 +544,12 @@
                                 </RadioButton>
 
                                 <TextBlock
+                                    x:Name="NavToolbarSmallIconsHeader"
                                     x:Uid="NavToolbarSmallIconsHeader"
                                     Grid.Row="2"
                                     Grid.Column="1"
                                     VerticalAlignment="Center"
+                                    PointerReleased="NavToolbarSmallIconsHeader_PointerReleased"
                                     Text="Small Icons" />
 
                                 <Border
@@ -576,10 +582,12 @@
                                 </RadioButton>
 
                                 <TextBlock
+                                    x:Name="NavToolbarMediumIconsHeader"
                                     x:Uid="NavToolbarMediumIconsHeader"
                                     Grid.Row="0"
                                     Grid.Column="4"
                                     VerticalAlignment="Center"
+                                    PointerReleased="NavToolbarMediumIconsHeader_PointerReleased"
                                     Text="Medium Icons" />
 
                                 <RadioButton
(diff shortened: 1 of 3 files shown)


yaichenbaum:
@cinqmilleans looking good! Would it make sense to toggle the hover style to indicate the label can be clicked?

cinqmilleans:
@yair100 I don't think this is necessary. In html forms, labels are often clickable, but without visual effect.

yair100:
> @yair100 I don't think this is necessary. In html forms, labels are often clickable, but without visual effect.

@cinqmilleans I'm not familiar with that but it seems intuitive to add one so the user knows they can click on the button.

cinqmilleans:
@yair100 You have to keep consistency. If we change the cursor, it must be done everywhere, including on the buttons, which currently use the arrow, which does not distinguish the click. I don't know if this is the default in uwp, but if we can change it globally, that would probably be a good thing.

yair100:
> @yair100 You have to keep consistency. If we change the cursor, it must be done everywhere, including on the buttons, which currently use the arrow, which does not distinguish the click. I don't know if this is the default in uwp, but if we can change it globally, that would probably be a good thing.

@cinqmilleans I agree, I played around with a couple options when first building this flyout, but I wasn't very happy with them. The cleanest solution might be to move the text into the button.

cinqmilleans:
@yair100 I compared with other software. Windows applications do not change the cursor on buttons. They use the arrow. The hand is used for links, and rarely in software, only in web pages.

yair100:
> @yair100 I compared with other software. Windows applications do not change the cursor on buttons. They use the arrow. The hand is used for links, and rarely in software, only in web pages.

@cinqmilleans that's correct but there is usually some visual feedback such as the color changing. What do you think about moving the text into the button?

cinqmilleans:
@yair100 I think it is not necessary. In the same flyout, there are on/off buttons. It is possible to interact by clicking on on, but there is no visual. This is the default behavior of this button in uwp. I wanted to reproduce the same logic. If we add a visual, we will lose consistency.

yaichenbaum:
@cinqmilleans if you look closely, the toggle switch has a slight animation, perhaps we can do something similar with the icon in the layout buttons.

cinqmilleans:
@yair100 Yes it's a good idea. But that's another feature for another pr. One step at a time.

Files/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -486,10 +486,12 @@
                                 </RadioButton>
 
                                 <TextBlock
+                                    x:Name="NavToolbarDetailsHeader"
                                     x:Uid="NavToolbarDetailsHeader"
                                     Grid.Row="0"
                                     Grid.Column="1"
                                     VerticalAlignment="Center"
+                                    PointerReleased="NavToolbarDetailsHeader_PointerReleased"
```
yaichenbaum, line 494: Is there a click event?

cinqmilleans: No, not on TextBlock.

yaichenbaum: Will this work on touch screens?

cinqmilleans: @yair100 Probably not. I don't think this is necessary. This pr is just a practical addition but not essential. If we want it to also work on a tablet, we will have to make xaml more complex (hyperlink with a style, button template, ...)

yaichenbaum: It's probably not too hard to restyle a button for this, I think it's important to allow support touch screens as the main button is a little small. I can work on this change if you want.

cinqmilleans: You can do it if you want to. Here is my current schedule:
- Experimental ShowFolderSize: I have to test the clouds and ftp disks. The pr should arrive within 2 days.
- Search filter: Right after.
- This pr then, if it is not finished.

yaichenbaum: I'll work on this 👍

yaichenbaum:
@cinqmilleans do you have a touch screen to see if the `tapped` event does the trick?

cinqmilleans:
@yair100 No, I don't have this hardware. It should work because Tapped is triggered by touchscreen or mouse.

yair100 (approved):
LGTM

PR #7006, opened 2021-11-20
Added support for symbolic file links

gave92:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #6995

**Details of Changes**
Add details of changes here.
- Adds basic support for symlinks (the ones generated by e.g. `> mklink dest_file.txt source_file.txt`. Currently treated as Shortcut files.

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #7201, opened 2021-12-03
Fix properties window not closing on OK

### LINKED ISSUE #7209: Potential race condition in FileOperationsHandler
### Description

The file contains the class `ProgressHandler` whose methods `AddOperation`, `RemoveOperation`, and `UpdateOperation` may be called in parallel judging from the use of `ConcurrentDictionary`. All three methods invoke `UpdateTaskbarProgress` which ultimately uses the property `Progress`: 
https://github.com/files-community/Files/blob/3a8da90d8009010d699a1c94df9b2420d97c06fa/Files.Launcher/MessageHandlers/FileOperationsHandler.cs#L809-L816

The property makes use of the field `operations` of type `ConcurrentDictionary`. Now, to compute the actual progress, it uses a combination of LINQ-Statements. I assume the intent of `ToList` here is to create a snapshot/copy of the `operations` dictionary to run the `Where` method on it safely. Unfortunately, the `ToList` method uses an optimized copy-mechanism for collections that implement the `ICollection` interface, which `ConcurrentDictionary` does. It first retrieves the length of the source collection, initializes the internal data structure accordingly, and copies the values after that; thus, there’s a race condition between the initialization and the copy-mechanism. You can see this behavior within the reference source:

- [Enumerable.ToList]( https://github.com/dotnet/runtime/blob/e80476680121b400676f4c244afeeb98f3b08a66/src/libraries/System.Linq/src/System/Linq/ToCollection.cs#L29)
- [List]( https://github.com/dotnet/runtime/blob/e80476680121b400676f4c244afeeb98f3b08a66/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L66)

### Steps To Reproduce

Due to the nature of race conditions, it's hard to reproduce the error in the actual code (and probably unlikely to occur). However, you can reproduce the race condition with the following example:

```cs
public class Program {
  private static ConcurrentDictionary<int, int> dictionary = new();

  static async Task Main() {
    Console.WriteLine("starting");
    var producer = Task.Run(Produce);
    CreateToListCopies();
    await producer;
    Console.WriteLine("completed");
  }

  private static void Produce() {
    for(int i = 0; i < 100_000_000; i++) {
      dictionary.TryAdd(i, i);
    }
    Console.WriteLine("finished producing");
  }

  private static void CreateToListCopies() {
    for(int i = 0; i < 100_000_000; i++) {
      dictionary.ToList();  // race condition that leads to System.ArgumentException
      //dictionary.ToArray(); // no race condition, ToArray() correctly creates a snapshot
    }
    Console.WriteLine("finished copying");
  }
}
```

### Expected behavior

The straightforward fix is replacing the invocation of LINQ’s `ToList` with `ToArray` implementation of `ConcurrentDictionary`. Alternatively, you can apply `ToList` after the `Where` clause to not have a race condition between `Any` and `Average` (`GetEnumerator` does not create a snapshot but is thread-safe).

### Files Version

latest commit to the main branch:  fe684bc; latest commit to the document: 3a8da90

### Windows Version

any

### Relevant Assets/Logs

none

gave92:
**Resolved / Related Issues**
- Follow up of #7006
- Closes #7209

**Details of Changes**
Add details of changes here.
- Set symlink target path in Properties as readonly and hide arguments, working directory
- Fixes an issue where Properties window does not close if Shortcut section is expanded
- Fixes #3524514594u (crash if fulltrust process fails to fetch Compatibility properties)

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files.Launcher/MessageHandlers/FileOperationsHandler.cs
+++ b/Files.Launcher/MessageHandlers/FileOperationsHandler.cs
@@ -810,7 +810,7 @@ namespace FilesFullTrust.MessageHandlers
             {
                 get
                 {
-                    var ongoing = operations.ToList().Where(x => !x.Value.Canceled);
+                    var ongoing = operations.ToArray().Where(x => !x.Value.Canceled);
                     return ongoing.Any() ? (int)ongoing.Average(x => x.Value.Progress) : 0;
                 }
             }
--- a/Files/ViewModels/Properties/FileProperties.cs
+++ b/Files/ViewModels/Properties/FileProperties.cs
@@ -73,10 +73,11 @@ namespace Files.ViewModels.Properties
                     ViewModel.ShortcutItemType = isApplication ? "PropertiesShortcutTypeApplication".GetLocalized() :
                         Item.IsLinkItem ? "PropertiesShortcutTypeLink".GetLocalized() : "PropertiesShortcutTypeFile".GetLocalized();
                     ViewModel.ShortcutItemPath = shortcutItem.TargetPath;
+                    ViewModel.IsShortcutItemPathReadOnly = shortcutItem.IsSymLink;
                     ViewModel.ShortcutItemWorkingDir = shortcutItem.WorkingDirectory;
-                    ViewModel.ShortcutItemWorkingDirVisibility = Item.IsLinkItem ? Visibility.Collapsed : Visibility.Visible;
+                    ViewModel.ShortcutItemWorkingDirVisibility = Item.IsLinkItem || shortcutItem.IsSymLink ? Visibility.Collapsed : Visibility.Visible;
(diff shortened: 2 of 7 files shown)


BanCrash (approved):
LGTM.

An issue for a future PR (I'll open an issue about this):
When you edit the info of a shortcut, you'll need to refresh the folder to get the updated shortcut information (the icon is updated though):

GIF 06-12-2021 16-06-10

On the gif I've tried two times before refreshing to make sure that it's not a question about opening one time after editing it to get the refreshed info.

I opened the Eclipse lua shortcut and edited its info with the GifCam info, and I needed to refresh the folder to have the new shortcut information.

EDIT: I've forgot to capture the mouse , I can add a new gif if it's not very clear what the issue is.

Issue #6880, opened 2021-11-12
Display size of folder 


cinqmilleans:
### What's the Problem?

It is not practical to know the total size occupied by a folder. You have to select one, open its property panel, and wait for the calculation. It is therefore not easy to spot a folder whose size is larger than desired. 

### Solution/Idea

Display the size of the folders in the size column of the detail view, in the preview pane and/or in the bottom description, with an asynchronous calculation as in the properties panel. As this can affect performance and disk access, this should be an option disabled by default. This option can be in the settings (with a warning description) or in the layout menu of the toolbar.

### Alternatives

In the size column, display a button for show the size of folder.

### Files Version

v2.0.33.0

### Windows Version

Windows 11

gave92:
Merging with #3944

PR #7244, opened 2021-12-08
Added experimental option to show folder sizes

### LINKED ISSUE #3944: Showing folder size in details view
There are so many third-party file explorer's that offer showing folder size, e.g. Explorer++, Xinorbis, WizTree, TreeSize, still nothing that allows you to view folder size in Explorer. Files, as far as the UI is concerned, is much better than all those alternatives. It would be nice to see that functionality implemented.

cinqmilleans:
**Resolved / Related Issues**
It is not practical to know the total size occupied by a folder. You have to select one, open its property panel, and wait for the calculation. It is therefore not easy to spot a folder whose size is larger than desired. Closes #3944 #6880.

**Details of Changes**
Add experimental setting for show folder size. The size is calculated asynchronously with no visible loss in performance. The size is visible in details view and status bar.

Works for local folders & library & OneDrive.
Should work on other clouds and ftp but I haven't tested because I don't have one.
Does not work inside a zip file. It can come later.

**Validation**
- [ ] Built and ran the app

### DIFF
--- a/Files/Strings/en-US/Resources.resw
+++ b/Files/Strings/en-US/Resources.resw
@@ -2875,7 +2875,7 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
   <data name="FolderWidgetCreateNewLibraryInputPlaceholderText" xml:space="preserve">
     <value>Enter Library name</value>
   </data>
-  <data name="SettingsShowFolderSize.Title" xml:space="preserve">
+  <data name="ShowFolderSize" xml:space="preserve">
     <value>Show folder size</value>
   </data>
 </root>
--- a/Files/Strings/en-US/Resources.resw
+++ b/Files/Strings/en-US/Resources.resw
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
+  <!--
+    Microsoft ResX Schema
+
     Version 2.0
-    
-    The primary goals of this format is to allow a simple XML format 
-    that is mostly human readable. The generation and parsing of the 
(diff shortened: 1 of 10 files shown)


Files/Strings/en-US/Resources.resw
```diff
@@ -2875,4 +2875,7 @@ We use App Center to keep track of app usage, find bugs, and fix crashes. All in
   <data name="FolderWidgetCreateNewLibraryInputPlaceholderText" xml:space="preserve">
     <value>Enter Library name</value>
   </data>
+  <data name="SettingsShowFolderSize.Title" xml:space="preserve">
```
yaichenbaum, line 2878:
```suggestion
  <data name="ShowFolderSize" xml:space="preserve">
```

Files/Views/SettingsPages/Experimental.xaml
```diff
@@ -73,6 +73,16 @@
                 </local:SettingsBlockControl.ExpandableContent>
             </local:SettingsBlockControl>
 
+            <local:SettingsBlockControl
+                x:Uid="SettingsShowFolderSize"
```
yaichenbaum, line 77:
```suggestion
```

yaichenbaum:
@cinqmilleans I found an issue that if you select a folder when the size is still calculating, the size in the status bar doesn't get updated.

cinqmilleans:
@yair100 Indeed. I take care of it.

cinqmilleans:
@yair100 I was able to correct the problem. However I notice that the cumulative size is not displayed when there are folders in the multiselection. I will try to add this functionality.

cinqmilleans:
@yair100 The size is now displayed for a selection containing folders.

yair100 (approved):
Great work!

PR #7255, opened 2021-12-09
Try close disc tray

### LINKED ISSUE #4071: Close disc tray
A nice accessibility feature that's not present in Files is to close the disc tray. We can already open the tray but not close it.

Add an ability to close the disc tray from within Files in context menu option.

**Additional context**
A piece of code that might help:
```csharp
private string returnString = "";

public void OpenTray(char cdDriveLetter)
{
    mciSendStringA("open " + cdDriveLetter + ": type CDaudio alias drive" + cdDriveLetter, returnString, 0, IntPtr.Zero);
    mciSendStringA("set drive" + cdDriveLetter + " door open", returnString, 0u, IntPtr.Zero);
}

public void CloseTray(char cdDriveLetter)
{
    mciSendStringA("open " + cdDriveLetter + ": type CDaudio alias drive" + cdDriveLetter, returnString, 0, IntPtr.Zero);
    mciSendStringA("set drive" + cdDriveLetter + " door closed", returnString, 0u, IntPtr.Zero);
}

[DllImport("winmm.dll", CallingConvention = CallingConvention.StdCall, CharSet = CharSet.Auto)]
[return: MarshalAs(UnmanagedType.I8)]
internal static extern long mciSendStringA([In][MarshalAs(UnmanagedType.LPStr)] string lpstrCommand, [In][Out][MarshalAs(UnmanagedType.LPStr)] string lpstrReturnString, [In][MarshalAs(UnmanagedType.U4)] uint uReturnLength, [In] IntPtr hwndCallback);
```

d2dyno1:
**Resolved / Related Issues**
- Closes the disc tray and #4071 .

**Details of Changes**
This PR adds an option to the context menu to try to close the disc tray. This PR is a draft.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/Files.Launcher/DeviceWatcher.cs
+++ b/Files.Launcher/DeviceWatcher.cs
@@ -10,7 +10,7 @@ namespace FilesFullTrust
 {
     public class DeviceWatcher : IDisposable
     {
-        private ManagementEventWatcher insertWatcher, removeWatcher, modifyWatcher;
+        private ManagementEventWatcher insertWatcher, removeWatcher, modifyWatcher, cdWatcher;
         private PipeStream connection;
 
         private const string WpdGuid = "{6ac27878-a6fa-4155-ba85-f98f491d4f33}";
@@ -31,6 +31,11 @@ namespace FilesFullTrust
             modifyWatcher = new ManagementEventWatcher(modifyQuery);
             modifyWatcher.EventArrived += new EventArrivedEventHandler(DeviceModifiedEvent);
             modifyWatcher.Start();
+            
+            WqlEventQuery cdQuery = new WqlEventQuery("SELECT * FROM __InstanceOperationEvent WITHIN 1 WHERE TargetInstance ISA 'Win32_CDROMDrive'");
+            cdWatcher = new ManagementEventWatcher(cdQuery);
+            cdWatcher.EventArrived += new EventArrivedEventHandler(CdModifiedEvent);
+            cdWatcher.Start();
 
             WqlEventQuery removeQuery = new WqlEventQuery("SELECT * FROM __InstanceDeletionEvent WITHIN 1 WHERE TargetInstance ISA 'Win32_LogicalDisk'");
             removeWatcher = new ManagementEventWatcher(removeQuery);
@@ -47,6 +52,19 @@ namespace FilesFullTrust
             var eventType = volumeName != null ? DeviceEvent.Inserted : DeviceEvent.Ejected;
             System.Diagnostics.Debug.WriteLine($"Drive modify event: {deviceName}, {deviceId}, {eventType}");
             await SendEvent(deviceName, deviceId, eventType);
+        }
+
+        private async void CdModifiedEvent(object sender, EventArrivedEventArgs e)
+        {
+            CimInstance obj = (CimInstance)e.NewEvent.Instance.CimInstanceProperties["TargetInstance"].Value;
+            var deviceName = (string)obj.CimInstanceProperties["Name"]?.Value;
+            var deviceId = (string)obj.CimInstanceProperties["DeviceID"]?.Value;
+            bool? mediaLoaded = (bool?)obj.CimInstanceProperties["MediaLoaded"]?.Value;
+
+            if (mediaLoaded is bool notNullMediaLoaded)
+            {
+                await SendEvent(deviceName, deviceId, notNullMediaLoaded ? DeviceEvent.MediaLoaded : DeviceEvent.MediaNotLoaded);
+            }
         }
 
         private async void DeviceRemovedEvent(object sender, EventArrivedEventArgs e)
@@ -84,9 +102,11 @@ namespace FilesFullTrust
             insertWatcher?.Dispose();
             removeWatcher?.Dispose();
             modifyWatcher?.Dispose();
+            cdWatcher?.Dispose();
             insertWatcher = null;
             removeWatcher = null;
             modifyWatcher = null;
+            cdWatcher = null;
         }
     }
 }
--- a/Common/DeviceEvent.cs
+++ b/Common/DeviceEvent.cs
@@ -5,6 +5,8 @@
         Added,
         Removed,
         Inserted,
-        Ejected
+        Ejected,
+        MediaLoaded,
+        MediaNotLoaded
     }
 }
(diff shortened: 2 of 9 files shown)


Files.Launcher/DeviceWatcher.cs
```diff
@@ -31,6 +31,11 @@ public void Start()
             modifyWatcher = new ManagementEventWatcher(modifyQuery);
             modifyWatcher.EventArrived += new EventArrivedEventHandler(DeviceModifiedEvent);
             modifyWatcher.Start();
+            
+            WqlEventQuery cdQuery = new WqlEventQuery("SELECT * FROM __InstanceOperationEvent WITHIN 1 WHERE TargetInstance ISA 'Win32_CDROMDrive'");
```
d2dyno1, line 35: @gave92 I need some help here 😅 Will the above even work?

duke7553: @d2dyno1 Just took a look at this. AFAIK we're trying to avoid future use of the Microsoft.Management.Infrastructure library in the FullTrustProcess. The hope is to remove it all completely where we don't need it. 

With that said, there are some relatively robust WinRT APIs to send IOControlCodes to arbitrary devices. Let's leave this open for a bit while I experiment with them. We should theoretically be able to send a standard IOCTL_CDROM_LOAD_MEDIA code without the full trust process using [these](https://docs.microsoft.com/en-us/uwp/api/windows.devices.custom?view=winrt-22000) APIs and the optical drive capability.

duke7553: @d2dyno1 Actually, I was a little confused with what we need to use the FullTrustProcess to check in this PR. As long as we're not sending the IO control code to the device from the FullTrust component, something like this I wrote really quick: 

``` cs
private async Task<bool> CloseOpticalDiskTray()
        {
            Guid GUID_DEVINTERFACE_CDROM = new Guid("{53F56308-B6BF-11D0-94F2-00A0C91EFB8B}");
            string opticalDriveSelector = CustomDevice.GetDeviceSelector(GUID_DEVINTERFACE_CDROM);
            DeviceInformationCollection opticalDrives = await DeviceInformation.FindAllAsync(opticalDriveSelector, new string[] { "System.ItemPathDisplay" });
            foreach (DeviceInformation deviceInfo in opticalDrives)
            {
                string enumeratedPath = (string)deviceInfo.Properties["System.ItemPathDisplay"];
                string filename = @"\\.\" + driveLetter + ":";

                if (Path.GetPathRoot(enumeratedPath).Equals(Path.GetPathRoot(filename), StringComparison.OrdinalIgnoreCase))
                {
                    ushort FILE_DEVICE_CD_ROM = 0x00000002;

                    CustomDevice opticalDrive = await CustomDevice.FromIdAsync(deviceInfo.Id, DeviceAccessMode.ReadWrite, DeviceSharingMode.Shared);
                    IOControlCode IOCTL_CDROM_LOAD_MEDIA = new IOControlCode(FILE_DEVICE_CD_ROM, 0x0203, IOControlAccessMode.Read, IOControlBufferingMethod.Buffered);
                    return await opticalDrive.TrySendIOControlAsync(IOCTL_CDROM_LOAD_MEDIA, null, null);
                }
            }

            return false;
        }
```

while nicer to have, is likely unnecessary. If we have functionality in the FullTrustProcess that can't be replicated with WinRT APIs, then we can go ahead with this PR. 👍🏻

d2dyno1: This looks interesting, however, rewriting the entirety of `RemovableDevice` section would be out of the scope of this PR. That is not to say we shouldn't make a switch. If you'd like to you could open a PR with these changes, you seem more knowledgeable of these WinRT APIs than I am 😄

duke7553: @d2dyno1 Fair enough. I also understand your work better now, but I'll defer to @gave92's expertise on this one.

Files.Launcher/DeviceWatcher.cs
```diff
@@ -49,6 +54,19 @@ private async void DeviceModifiedEvent(object sender, EventArrivedEventArgs e)
             await SendEvent(deviceName, deviceId, eventType);
         }
 
+        private async void CdModifiedEvent(object sender, EventArrivedEventArgs e)
+        {
+            CimInstance obj = (CimInstance)e.NewEvent.Instance.CimInstanceProperties["TargetInstance"].Value;
+            var deviceName = (string)obj.CimInstanceProperties["Name"]?.Value;
+            var deviceId = (string)obj.CimInstanceProperties["DeviceID"]?.Value;
```
gave92, line 61: To be able to match a drive this should be: `var deviceId = (string)obj.CimInstanceProperties["Drive"]?.Value;`

Files.Launcher/DeviceWatcher.cs
gave92: I don't have a physical drive to test with (only daemontools virtual drive) but this should work.
I can think of two other ways to check is a drive is inserted from UWP:
1) Call `StorageHeldersExists("F:")` (F: being the cdrom drive letter)
2) Use DeviceIoControl with IOCTL_STORAGE_CHECK_VERIFY2: https://stackoverflow.com/questions/3158054/how-to-detect-if-media-is-inserted-into-a-removable-drive-card-reader

Files.Launcher/DeviceWatcher.cs
gave92, line 61: This should be `var deviceId = (string)obj.CimInstanceProperties["Drive"]?.Value;` or you won't find a matching drive I think.

Files.Launcher/DeviceWatcher.cs
gave92: The UWP alternative would be to use DeviceIoControl to check tray open/close status -> https://ws0.org/windows-how-to-get-the-tray-status-of-the-optical-drive/

yaichenbaum:
@d2dyno1 is there a way to detect if there is an automatic option to close the tray?

d2dyno1:
> @d2dyno1 is there a way to detect if there is an automatic option to close the tray?

What do you mean by "automatic option"? Check whether it has the capability to close the tray? (In this case, I assume it's not possible though @gave92 might prove me wrong here)

@yair100

yaichenbaum:
@d2dyno1 I'd rather not have an option that doesn't always work as I mentioned in the original issue.

d2dyno1:
> @d2dyno1 I'd rather not have an option that doesn't always work as I mentioned in the original issue.

That's why it says "**Try** close tray"

gave92:
> Check whether it has the capability to close the tray?

If there's a way I don't know it. Currently the option will show for virtual/iso cdrom drives too for example.

d2dyno1:
@gave92 @duke7553 Sorry for taking your time here but I've decided to close this PR. As @yair100 suggested, this feature would benefit not big enough group of users to be considered worth adding. I'll also be closing the original feature request associated with this PR.

PR #7311, opened 2021-12-17
Added subtle listview animation

yair100:


d2dyno1:
Could you share a video? :eyes:

BanCrash:
@yair100 when and where is this animation executed? I've tried it but I didn't see any difference on the places that I've tried.

yair100:
> @yair100 when and where is this animation executed? I've tried it but I didn't see any difference on the places that I've tried.

@BanCrash in the details and column layouts.

BanCrash:
> > @yair100 when and where is this animation executed? I've tried it but I didn't see any difference on the places that I've tried.
> 
> @BanCrash in the details and column layouts.

When loading / entering a folder, right?

I've should searched what meant subtle before, for some reason I thought it means something else :sweat_smile:

yair100:
> > > @yair100 when and where is this animation executed? I've tried it but I didn't see any difference on the places that I've tried.
> > 
> > 
> > @BanCrash in the details and column layouts.
> 
> When loading / entering a folder, right?
> 
> I've should searched what meant subtle before, for some reason I thought it means something else 😅

@BanCrash yeah, unfortunately I wasn't able to get it working when refreshing a folder but maybe I'll get that another time. The idea is that the animation is there, but barely noticeable.

BanCrash (approved):
LGTM

PR #7366, opened 2021-12-27
fix: Prevent repetitive viewport analysis in PointerMoved

### LINKED ISSUE #7300: Improve rectangle selection performance
The performance of rectangle selection is poor in large folders. 
I created this issue for tracking the optimization work of rectangle selection.

duke7553:
**Resolved / Related Issues**
- Closes #7300 

**Details of Changes**
Analyze the ListView viewport for item rectangles a single time, upfront, rather than every time the pointer is moved. Further, do this rather intensive work on selection rectangle y-axis dilation only when the scrolling gesture becomes determinate for 1 second. 

**English**: Drag-to-select will no longer lag in very large folders (e.g. System32), and similar safeguards are put in place when attempting to select more items outside the bottom (or top) of the ListView.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
N/A

### DIFF
--- a/src/Files/UserControls/Selection/RectangleSelection_ListViewBase.cs
+++ b/src/Files/UserControls/Selection/RectangleSelection_ListViewBase.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.Toolkit.Uwp.UI;
 using Windows.Foundation;
 using Windows.System;
 using Windows.UI.Xaml;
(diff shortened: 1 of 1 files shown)


yaichenbaum (approved):
Nice work!

gave92:
If you select a lot of items by scrolling without stopping, items in the middle won't get selected (by the time you stop some items will be virtualized again).

duke7553:
@gave92 This was part of the safeguard to ensure we don't lag when dilating the selection rectangle over the y axis. Otherwise, we would be adding every item in the viewport for each tiny vertical offset change. Waiting until the user stops scrolling seemed like a decent temporary solution until someone can figure out a better method, but considering how bad the behavior was before in large directories, we should go ahead with this imo

gave92:
Made a small change that should fix the above issue. It should not affect performance, could you check/confirm if it's ok?

duke7553:
@gave92 Not really sure how you did it, but the issue mentioned above is indeed fixed. Thanks!

gave92 (approved):

yaichenbaum (approved):

PR #7388, opened 2021-12-30
Improve styles in property dialog

### LINKED ISSUE #7386: Bad column sizes in the new property popup
### Description

In the disk size part of the properties, the column sizes are suitable for English, but not for other languages. The texts can be too long.

### Steps To Reproduce

1. Use Files after commit #7378.
2. In settings, choose the language French.
3. In Home (Accueil in French), open the properties panel of a drive.
4. Look Général/more details

### Expected behavior

The texts should be readable, or at least cut cleanly. 

### Files Version

Commit #7378

### Windows Version

Windows 11

### Relevant Assets/Logs

properties

yaichenbaum:
**Resolved / Related Issues**
- Closes #7386

**Details of Changes**
Add details of changes here.
- Fixed padding
- Added card style to details page
- Fixed text clipping
- Added a button to open disk cleanup from the properties window
- Switched to x:Load to prevent extra space between rows in the stackpanel controls

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.Package/Package.appxmanifest
+++ b/src/Files.Package/Package.appxmanifest
@@ -9,7 +9,7 @@
          xmlns:uap4="http://schemas.microsoft.com/appx/manifest/uap/windows10/4"
          xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5"
          IgnorableNamespaces="uap uap5 mp rescap desktop4 desktop">
-    <Identity Name="ssFilesDev" Publisher="CN=53EC4384-7F5B-4CF6-8C23-513FFE9D1AB7" Version="2.0.42.0" />
+    <Identity Name="FilesDev" Publisher="CN=53EC4384-7F5B-4CF6-8C23-513FFE9D1AB7" Version="2.0.42.0" />
     <Properties>
         <DisplayName>Files - Dev</DisplayName>
         <PublisherDisplayName>Yair A</PublisherDisplayName>
--- a/src/Files.Package/Package.appxmanifest
+++ b/src/Files.Package/Package.appxmanifest
@@ -9,7 +9,7 @@
          xmlns:uap4="http://schemas.microsoft.com/appx/manifest/uap/windows10/4"
          xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5"
          IgnorableNamespaces="uap uap5 mp rescap desktop4 desktop">
-    <Identity Name="FilesDev" Publisher="CN=53EC4384-7F5B-4CF6-8C23-513FFE9D1AB7" Version="2.0.42.0" />
+    <Identity Name="ssFilesDev" Publisher="CN=53EC4384-7F5B-4CF6-8C23-513FFE9D1AB7" Version="2.0.42.0" />
     <Properties>
         <DisplayName>Files - Dev</DisplayName>
         <PublisherDisplayName>Yair A</PublisherDisplayName>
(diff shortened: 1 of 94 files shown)


src/Files/Views/Pages/PropertiesGeneral.xaml.cs
```diff
@@ -108,5 +109,10 @@ public override async Task<bool> SaveChangesAsync(ListedItem item)
         public override void Dispose()
         {
         }
+
+        private async void DiskCleanupButton_Click(object sender, Windows.UI.Xaml.RoutedEventArgs e)
+        {
+            await Launcher.LaunchUriAsync(new Uri("ms-settings:storagesense"));
```
d2dyno1, line 115: Could you include the same logic as found in [DrivesWidget.xaml.cs](https://github.com/files-community/Files/blob/main/src/Files/UserControls/Widgets/DrivesWidget.xaml.cs#L202)?

yaichenbaum: Is there a way to put the code one location to make it easier to reuse?

d2dyno1:
Nit
----
Align these items horizontally (also, that section wasn't expanded)
image

Bug
----
I don't know why, but when clicking "Advanced Permissions", closing the popup window, and clicking the "Advanced Permissions" again, a new window won't open, instead, they'll be displayed in the Properties window.
image

Question
----
Will the Customization section be refreshed with the new design?

Nit
----
While the Properties are being redesigned, I think that this ProgressRing could be replaced with something more fitting. (e.g. ProgressBar)
image
Moreover, if we follow the same idea, another progress should also be added for the files and folders count (when they are being traversed and counted)

Nit
----
The file icon isn't properly aligned and sized
image

Bug
----
Renaming the folder won't let you close the window when it is still being traversed. The window will only close itself when it's  deactivated (lost focus)

Nit
----
AAAAAAAAAAAAAAAAAAAAA
image

Nit
----
Add more paddings in the Drive Properties
image
Disk cleanup button is too close to the right. (The bug with the security window is also present here)

Bug
----
Clicking on the empty network folder causes the security tab to freak out
image

yaichenbaum:
> ## Nit
> Align these items horizontally (also, that section wasn't expanded) image

They seem aligned to me, can you elaborate? The design spec called for this section to be collapsed by default.

> ## Bug
> I don't know why, but when clicking "Advanced Permissions", closing the popup window, and clicking the "Advanced Permissions" again, a new window won't open, instead, they'll be displayed in the Properties window. image

Is that specific to this PR?

> ## Question
> Will the Customization section be refreshed with the new design?

Yes 🙂 

> ## Nit
> While the Properties are being redesigned, I think that this ProgressRing could be replaced with something more fitting. (e.g. ProgressBar) image Moreover, if we follow the same idea, another progress should also be added for the files and folders count (when they are being traversed and counted)

I'll give that a try.

> ## Nit
> The file icon isn't properly aligned and sized image

It's center aligned in the row and left aligned to match the text boxes and other items underneath it.

> ## Bug
> Renaming the folder won't let you close the window when it is still being traversed. The window will only close itself when it's deactivated (lost focus)

I can't reproduce this, can you see if you can figure out the cause?

src/Files.Package/Package.appxmanifest
```diff
@@ -9,7 +9,7 @@
          xmlns:uap4="http://schemas.microsoft.com/appx/manifest/uap/windows10/4"
          xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5"
          IgnorableNamespaces="uap uap5 mp rescap desktop4 desktop">
-    <Identity Name="FilesDev" Publisher="CN=53EC4384-7F5B-4CF6-8C23-513FFE9D1AB7" Version="2.0.42.0" />
+    <Identity Name="FilesDev" Publisher="CN=Files" Version="2.0.42.0" />
```
d2dyno1, line 12: What's the reason of this change?

yaichenbaum: The real reason is that the installation of Files on my device somehow got into a bad state and while debugging I found that changing this fixed the issue. I can revert it if you want.

d2dyno1 (approved):

PR #7394, opened 2021-12-31
Install fonts through powershell

### LINKED ISSUE #7093: Fonts not installed correctly using shell context menu
### Description

I noticed that when I install **fonts** using the shell context menu in Files, the process continues with no apparent errors, however the fonts are apparently not installed, and are not detected by the OS, as they couldn't be found in either the fonts settings or the control panel. Also, **install for all users** gives an error.

### Steps To Reproduce

Install a font in Files using the shell menu, and it won't be detected by the system (not installed correctly?)

### Expected behavior

Fonts should be installed as if installed using Explorer.

### Files Version

Version: 2.0.34.0 

### Windows Version

Windows 11 22504.1010

### Relevant Assets

_No response_

gave92:
**Resolved / Related Issues**
- Closes #7093

**Details of Changes**
Add details of changes here.
- Given that "Install" and "Install for all users" don't correctly install fonts, "manually" install fonts through powershell instead

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Launcher/MessageHandlers/ContextMenuHandler.cs
+++ b/src/Files.Launcher/MessageHandlers/ContextMenuHandler.cs
@@ -3,6 +3,7 @@ using FilesFullTrust.Helpers;
 using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.IO.Pipes;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


BanCrash (approved):
LGTM

Fun fact, with the previous version installing for just my user was working but they can't be uninstalled lol. With this commit those can be removed as expected.

gave92:
@BanCrash yeah the font was copied to the fonts folder but the it was not added to the registry (so uninstall wasn't possible)

PR #7402, opened 2022-01-02
Migrate from FilesystemResult to SafeWrapper

d2dyno1:
**Resolved / Related Issues**
\-

**Details of Changes**
- Coming from Clipboard Canvas, this PR replaces FilesystemTasks with a more advanced solution called SafeWrapper. The goal of SafeWrapper is to ensure software stability and make it less vulnerable to bugs and crashes, it also provides the ability to customize error messages. The long-term plan is to replace try-catch blocks with SafeWrapperRoutines and replace return values of methods with SafeWrapper<> and SafeWrapperResult.

I'll leave this PR for now for other contributors to investigate SafeWrapper code, once it's approved, I'll start converting the codebase to use it.

**Validation**
- [ ] Built and ran the app


gave92:
Could you add an example of how some code that uses FilesystemTasks would become with SafeWrapper?

d2dyno1:
Well, the main benefit of SafeWrapper is that it helps unify the codebase. A lot of places utilize it to return and notify of the state of the method while providing useful information:
image
The method will return different errors in case of failure.

It also comes with SafeWrapperRoutines which is an expansion of FilesystemTasks that supports both synchronous and asynchronous callbacks. Callbacks witch chaining (OnFailure, OnSuccess) (though these are a bit broken at this stage 😅)
image

Another use of SafeWrapper is that it can provide flexible error reporting thanks to ISafeWrapperExceptionReporter so you can provide in-depth information about the operation state to the user.
image
(Accepts custom ISafeWrapperExceptionReporter, if not provided, the default one is used)

Unlike FilesystemTasks, SafeWrapper comes with SafeWrapperResultDetails which in turn, provide details about the result of an operation:
image
It was engineered from the inception, to provide as many details as possible.

**Important note**
SafeWrapper isn't something new, but rather it's an upgrade to existing FilesystemTasks. An upgrade flexible enough and designed to adapt to the whole codebase.

gave92:
@d2dyno1 still have to check this out, but could the SafeWrapper be put in the Common project? I wouldn't mind using it in the fulltrust process as well instead of the "Extensions.IgnoreExceptions()" method.

d2dyno1:
> @d2dyno1 still have to check this out, but could the SafeWrapper be put in the Common project? I wouldn't mind using it in the fulltrust process as well instead of the "Extensions.IgnoreExceptions()" method.

That could work but SafeWrapperExceptionReporter has some references that make it impossible to port to the Common project entirely. We could define the implementation in the Files project while having the definition (interface) in Common, though we'd need to pass it every time to SafeWrapperRoutines if we wanted to have error reporting. @gave92

src/Files/SafetyHelpers/ExceptionReporters/IStorageFolder.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Common.SafetyHelpers.ExceptionReporters
+{
+    internal interface IStorageFolder
```
gave92, line 3: Are these placeholders for something? What for?

d2dyno1: Oh no! I must have accidentally selected "Create in new file" while resolving namespaces in VS 😅 That's obviously not to stay and I didn't notice it

d2dyno1:
@duke7553 Ready to merge

PR #7494, opened 2022-01-07
Improve keyboard nav in column layout

yaichenbaum:


src/Files/Views/LayoutModes/ColumnViewBase.xaml.cs
```diff
@@ -385,8 +385,18 @@ private async void FileList_PreviewKeyDown(object sender, KeyRoutedEventArgs e)
...
                     FileList.SelectedIndex = 0;
-                    e.Handled = true;
                 }
+
+                // Open slected directory
+                if (!IsRenamingItem && !ParentShellPageInstance.NavToolbarViewModel.IsEditModeEnabled)
+                {
+                    if (IsItemSelected && SelectedItem.PrimaryItemAttribute == StorageItemTypes.Folder)
+                    {
+                        ItemInvoked?.Invoke(new ColumnParam { NavPathParam = (SelectedItem is ShortcutItem sht ? sht.TargetPath : SelectedItem.ItemPath), ListView = FileList }, EventArgs.Empty);
```
yaichenbaum, line 395: Is there a way to keep the focus in the current column?

gave92: I'll check but it's done in baselayout I think.

gave92: Also I'd add some delay before opening the folder, what if you're trying to go down many items with the keyboard?

src/Files/Views/LayoutModes/ColumnViewBase.xaml.cs
```diff
@@ -382,11 +382,21 @@ private async void FileList_PreviewKeyDown(object sender, KeyRoutedEventArgs e)
...
                     FileList.SelectedIndex = 0;
                     e.Handled = true;
                 }
+
+                // Open slected directory
+                if (!IsRenamingItem && !ParentShellPageInstance.NavToolbarViewModel.IsEditModeEnabled)
+                {
+                    if (IsItemSelected && SelectedItem.PrimaryItemAttribute == StorageItemTypes.Folder)
+                    {
+                        ItemInvoked?.Invoke(new ColumnParam { NavPathParam = (SelectedItem is ShortcutItem sht ? sht.TargetPath : SelectedItem.ItemPath), ListView = FileList }, EventArgs.Empty);
```
gave92, line 396: This does not work as it prevents moving selection, should be done in FileList_SelectionChanged

gave92:
My suggestion is to merge the PR like this. Opening a folder on selection is nice to have but can wait after 2.1 and would not work anyway if next column in selected on navigation.

gave92 (dismissed):

src/Files/Views/LayoutModes/ColumnViewBase.xaml.cs
yaichenbaum: Focus is expected to stay in the current column even when opening an item with the mouse.

yaichenbaum:
> My suggestion is to merge the PR like this. Opening a folder on arrow selection is nice to have but can wait after 2.1 and would not work anyway if next column in selected on navigation (no trivial change).

Would adding the delay allow us to implement it for v2.1? The next update won't be for another couple months 🤔

gave92:
Latest change should not move focus to next column, hopefully without destroying keyboard navigation in the other layouts.

gave92:
Added back opening selected folder on arrow up/down. Pressing an arrow key repeatedly causes some stuttering, some tweaking of the delay might be necessary.

yaichenbaum:
> Added back opening selected folder on arrow up/down. Pressing an arrow key repeatedly causes some stuttering, some tweaking of the delay (200ms) might be necessary.

100ms worked fine on my device but 200 is probably better

gave92:
Done! Do you see any other issues?

gave92 (approved):

yaichenbaum:
> Done! Do you see any other issues?

Seems to work well 🙂

Vanillabacke:
needs a bit more tweaks.
- column view
- navigate with arrow keys back and forth
- content is not fully loaded of an folder
- focus is now on the first tab
- have to click back into the column view

it should stay in this focus section without escaping to the main section

https://user-images.githubusercontent.com/56526856/153396242-29a44419-a17b-4b36-9740-ec99d7577a11.mp4

PR #7552, opened 2022-01-12
Created navigation test

yaichenbaum:
Created a test to navigate to a directory


tests/Files.InteractionTests/Tests/NavigationTests.cs
```diff
@@ -0,0 +1,24 @@
...
+		[TestCleanup]
+		public void Cleanup()
+		{
+			TestHelper.InvokeButtonByName("Home");
+		}
+
+		[TestMethod]
+		public void VerifyNavigationWorks()
+		{
+			TestHelper.InvokeButtonByName("Desktop");
```
chingucoding, line 20: Was there a specific reason why you chose to invoke by name here?

yaichenbaum: It's the first thing I tried and it worked.

chingucoding: Probably should have said this, but you would also need to set the AutomationProperties.AutomationID property on the XAML controls.

chingucoding: Now you have updated the AutomationID to the same value as the text again which again will fail with localization. What I have done for the settings test is set the AutomationID to a fixed value regardless of locale. Sorry about the confusion.

yaichenbaum: The items aren't fixed as they're from an item source, how would I set the fixed value?

chingucoding: Oh shoot, totally missed that. That's a very difficult thing. There are a few hacky workarounds we could use but I think the best would be to either accept this issue with localization or set the automation ID of the parent control and grab the first child (assuming we don't care which folder to open here).

On a side note, it seems that it should be `AutomationProperties.AutomationId` with a lowercase d, sorry forgot that.

yaichenbaum: One way to work around the localization issue would be to create and pin a custom folder as part of the test.

yaichenbaum: > or set the automation ID of the parent control and grab the first child (assuming we don't care which folder to open here).

This sounds like a better idea, is there a way to do that?

chingucoding: Yes that is possible, you can query the sub children of an element by calling `FindElementsBy...` on the selenium object (in our case the element with the AutomationID).

chingucoding (approved):

PR #7558, opened 2022-01-12
Ability to set tags on read-only folders

### LINKED ISSUE #7534: Folder tags won't save on folders with custom icons
### Description

If I change the icon of a folder, it will not save the assignment of a tag. If the folder had a tag before changing the icon, it will not save the removing of it.
If the icon has been changed once, even "Restore default" in the Customization tab will not fix it. I did not find any other way to fix it, other than creating a new folder.

### Steps To Reproduce

1. Go to the properties of any folder
2. Assign a custom folder icon
2. Give the folder a tag or remove the tag, if it had one before assigning a custom icon
3. Refresh
4. The Tag is gone or back - reset to how it was before the custom icon

### Expected behavior

Assigning and removing tags should be saved, even if the folder icon has been changed.

### Files Version

2.1.0.0

### Windows Version

Windows 11 Pro 21H2 22000.376

### Relevant Assets/Logs

[debug_fulltrust.log](https://github.com/files-community/Files/files/7847144/debug_fulltrust.log)
[debug.log](https://github.com/files-community/Files/files/7847145/debug.log)

gave92:
**Resolved / Related Issues**
- Closes #7534

**Details of Changes**
Add details of changes here.
- Folders with a custom icon are set as "read-only". To save the tag remove the RO attribute, write the tag and restore the attribute.

### DIFF
--- a/src/Files/Filesystem/FileTagsHelper.cs
+++ b/src/Files/Filesystem/FileTagsHelper.cs
@@ -35,6 +35,11 @@ namespace Files.Filesystem
 
         public static void WriteFileTag(string filePath, string tag)
         {
+            var isReadOnly = NativeFileOperationsHelper.HasFileAttribute(filePath, System.IO.FileAttributes.ReadOnly);
+            if (isReadOnly) // Unset read-only attribute (#7534)
+            {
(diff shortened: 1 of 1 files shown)


yaichenbaum (approved):
LGTM

PR #7607, opened 2022-01-13
Added App Center tracking on startup

d2dyno1:


gave92:
Does it works calling Analitics methods before the StartAppCenter call?

yaichenbaum:
> Does it works calling Analytics methods before the StartAppCenter call?

No

PR #7608, opened 2022-01-13
Added privacy section to the readme

yaichenbaum:


d2dyno1:
Translations will have to be updated. That's why Clipboard Canvas always pulls the privacy policy directly from github. Why won't you make Files do it too?

d2dyno1 (approved):

PR #7616, opened 2022-01-13
Bundled themes inside of the app

yair100:
The goal of this PR is to replace the themes addon by building them into the app.


gave92:
Looks good. My only doubt is if it still makes sense to show the "open them folder" button given that install folder is protected so users can't add/remove/edit themes.

yaichenbaum:
@gave92 would it be possible to provide an option to import themes or will the user still need access?

gave92:
That folder is read only. If we wish users to have access we'd need to put themes in appdata folder like before (themes bundled in the app are ok inside InstalledLocation but app would need to fetch themes from appdata as well)

yaichenbaum:
@gave92 made a couple changes, what do you think?

gave92 (approved):

thierryntoh24:
I just redownloaded the app after a format and noticed this. Cool.
Thanks a ton, @yair100 . 🙂

PR #7717, opened 2022-01-16
Update Package.appxmanifest v2.1.2

yaichenbaum:


gave92:
Is v2.1.2 going to Store stable after insider testing? Or will it be just for sideload?

yaichenbaum:
> Is v2.1.2 going to public Store after insider testing? Or will it be just for sideload?

It's going to the store for testing too.

gave92:
Nice, as we somewhat botched v2.1 xD

PR #7792, opened 2022-01-18
Is it possible to bring flash memory/drive formatting feature?

legalMori:
Hi,

I use the application by default, but when I insert a flash memory, while formatting easily with the wibdows file manager, I could not see the formatting option in this application. Will it come in the future?


Josh-65:
Use the issues page not pull requests

Issue #7769, opened 2022-01-18
Add set as background button to toolbar when image files are selected


yaichenbaum:
### Solution/Idea

Adding a toolbar button will increase ease of use when setting an image as the desktop background.

### Alternatives

NA

rr-da-cat:
Yea plz do it

puppetsw:
@yair100 Is this another one I can take on?

generalguy41:
Would it be possible to make it a dropdown with the option to Set as Lockscreen as well? (Or alternatively add that as a button too).

yaichenbaum:
> Would it be possible to make it a dropdown with the option to Set as Lockscreen as well? (Or alternatively add that as a button too).

At the moment, that's not part of the design spec.

generalguy41:
> > Would it be possible to make it a dropdown with the option to Set as Lockscreen as well? (Or alternatively add that as a button too).
> 
> At the moment, that's not part of the design spec.

Right, but... could it be? Should I raise a separate feature request? Figured it'd be better to ask now before this gets merged in.

yaichenbaum:
@generalguy41 can you open a separate feature request?

esibruti:
@yair100 Can I open the feature request myself?

esibruti:
@yair100 We can close this issue.

PR #7795, opened 2022-01-18
Added set as background button on toolbar when selecting image

puppetsw:
**Resolved / Related Issues**
- Related #7769

**Details of Changes**
Add details of changes here.
- Added "Set as background" button on toolbar when selecting an image file.
- ~~SetAsBackground resource string to be added.~~

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Example22


src/Files/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -225,6 +225,18 @@
                         <FontIcon Glyph="&#xE756;" />
                     </AppBarButton.Icon>
                 </AppBarButton>
+                <AppBarButton
+                    x:Name="SetAsBackgroundButton"
+                    x:Load="{x:Bind ViewModel.IsImage, Mode=OneWay, FallbackValue=False}"
+                    AutomationProperties.Name="SetAsBackground"
```
yaichenbaum, line 231: The automation name will automatically be picked up in this case since we already set the label. I usually set the name if there isn't any text or if the text alone wouldn't provide enough context.

yaichenbaum:
@puppetsw nice work! Can you add the resource string to the en-us file?

src/Files/UserControls/InnerNavigationToolbar.xaml
puppetsw: Ah okay, would you like to remove it?

yaichenbaum: Please

src/Files/Filesystem/ListedItem.cs
```diff
@@ -414,7 +414,7 @@ public override string ToString()
         public bool IsLinkItem => IsShortcutItem && ((ShortcutItem)this).IsUrl;
         public bool IsFtpItem => this is FtpItem;
         public bool IsZipItem => this is ZipItem;
-
+        public virtual bool IsImage => new[] { ".jpg", ".jpeg", ".bmp", ".png" }.Contains(Path.GetExtension(ItemPath), StringComparer.OrdinalIgnoreCase);
```
yaichenbaum, line 417: I think we already have a property for this `selectedItemsPropertiesViewModel.IsSelectedItemImage`, is it possible to reuse it?

puppetsw: I’ll have a look and check.

puppetsw: I removed the IsImage property from this file, and made use of the SelectedItemsViewModel. I couldn't see anywhere I could access it without instantiating it, and using the CheckFileExtension method.

yaichenbaum: It might make sense to create a helper for this

puppetsw: I'll just double check the existing helpers and see if there was something already there, if not I'll find somewhere to add it.

src/Files/Helpers/ListedItemHelpers.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                return false;
+            }
+
+            if (string.IsNullOrEmpty(listedItem.FileExtension))
+            {
+                return false;
+            }
+
+            return listedItem.FileExtension.Equals(".png", StringComparison.OrdinalIgnoreCase) ||
```
yaichenbaum, line 27: Instead of passing in a listed item, can this be simplified to passing in a file extension? This will make the helper more flexible and let us check the extension even when the file isn't a listed item.

puppetsw: Sure, can do. That makes sense.

puppetsw: I've added the helper. Should I go an update the other properties in the viewmodel that uses the extension checks? or leave that for another PR.

yaichenbaum: I'm going to push a small change first and then you can apply it to the other places so that it's consistent

yaichenbaum (approved):
LGTM

Issue #5584, opened 2021-07-27
Allow pressing Enter on a Delete Confirmation Dialog to proceed with the deletion.


rlaphoenix:
**Is your feature request related to a problem?**
When I press SHIFT+DEL to delete some files, I'm used to Windows asking if I'm sure with a Yes/No prompt, with Yes auto-focused. This allowed me to simply press SHIFT+DEL+Enter to proceed.

**Describe the solution**
Like Windows File Explorer's own Dialog, it would be ideal to autofocus on the "Delete" button so I can simply press enter.

**Describe alternatives you've considered**
N/A

**Scope**
| Capability | Priority |
| :---------- | :------- |
| Press Enter on a Delete Confirmation Dialog to proceed with Deletion | Must |
| Mouse should not need to be used | Should |

**System Information**
 - OS Version: Windows 11 22000.100
 - App version: v1.9.15.0

**Additional comment**
N/A

yaichenbaum:
@rlaphoenix this should already be implemented in v1.9.15, can you check if it ever works?

rlaphoenix:
> 
> 
> @rlaphoenix this should already be implemented in v1.9.15, can you check if it ever works?

I've tried already. It seems to focus on the first file in the list of items to be deleted, so pressing enter tries to click on the file rather than the button.

rlaphoenix:
Ok strangely I've retried again, and it now works fine. That's quite odd. I was definitely testing this on v1.19.15 too.

rlaphoenix:
**I've found out why!** The bug in question only occurs if the "Permanently delete" is disabled (cannot be unticked). E.g. while deleting from the Recycle Bin.

d2dyno1:
This has been resolved with #7882

PR #7882, opened 2022-01-22
Fixed delete dialog primary button focus issue

### LINKED ISSUE #5788: Wrong selection for deleting permanently with "del" key
### Description

image

Bug 1:
"Delete permanently" gets selected instead of "Delete" on the bottom.

Bug 2:
Pressing enter does not select "Delete permanently" but triggers the "Delete" button

### Steps To Reproduce

_No response_

### Expected behavior

/

### Files Version

Version: 1.9.21.0 OS Version: 10.0.19043.1110

### Windows Version

Win10Home21H1

### Relevant Assets

/

puppetsw:
**Resolved / Related Issues**
- Closes #5788
- Related #5584

**Details of Changes**
Add details of changes here.
- Fixed default focus of Primary button in ContentDialog by moving x:Bind tags to the dialog loaded event. 
- Deleting a file in the recycle bin and pressing enter now confirms the deletion.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
+++ b/src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
@@ -33,6 +33,7 @@ namespace Files.Dialogs
 
         private void OnLoaded(object sender, RoutedEventArgs e)
         {
+            Loaded -= OnLoaded;
             DetailsGrid.ItemsSource = ViewModel.Items;
             DetailsGrid.SelectionMode = ViewModel.ItemsSelectionMode;
         }
--- a/src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
+++ b/src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
@@ -3,8 +3,6 @@ using Files.ViewModels.Dialogs;
 using Microsoft.Toolkit.Uwp.UI;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading.Tasks;
-using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 
 // The Content Dialog item template is documented at https://go.microsoft.com/fwlink/?LinkId=234238
@@ -28,14 +26,6 @@ namespace Files.Dialogs
             ViewModel = viewModel;
             ViewModel.View = this;
             ViewModel.LoadedCommand.Execute(null);
-
-            Loaded += OnLoaded;
-        }
-
-        private void OnLoaded(object sender, RoutedEventArgs e)
-        {
-            DetailsGrid.SelectionMode = ViewModel.ItemsSelectionMode;
-            DetailsGrid.ItemsSource = ViewModel.Items;
         }
 
         private void MenuFlyoutItem_Click(object sender, Windows.UI.Xaml.RoutedEventArgs e)
@@ -95,17 +85,6 @@ namespace Files.Dialogs
             {
                 (sender as Grid).FindAscendant<ListViewItem>().ContextFlyout = ItemContextFlyout;
             }
-        }
-
-        private async void RootDialog_Loaded(object sender, Windows.UI.Xaml.RoutedEventArgs e)
-        {
-            ViewModel.LoadedCommand.Execute(null);
-            await Task.Delay(50);
-            if (chkPermanentlyDelete != null)
-            {
-                chkPermanentlyDelete.IsEnabled = ViewModel.PermanentlyDeleteEnabled;
-            }
-            DetailsGrid.IsEnabled = true;
         }
     }
 
--- a/src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
+++ b/src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
@@ -91,7 +91,10 @@ namespace Files.Dialogs
         {
             ViewModel.LoadedCommand.Execute(null);
             await Task.Delay(50);
-            chkPermanentlyDelete.IsEnabled = ViewModel.PermanentlyDeleteEnabled;
+            if (chkPermanentlyDelete != null)
+            {
+                chkPermanentlyDelete.IsEnabled = ViewModel.PermanentlyDeleteEnabled;
+            }
             DetailsGrid.IsEnabled = true;
         }
     }
--- a/src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
+++ b/src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
@@ -3,6 +3,7 @@ using Files.ViewModels.Dialogs;
 using Microsoft.Toolkit.Uwp.UI;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using Windows.UI.Xaml.Controls;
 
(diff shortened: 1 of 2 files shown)


src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
```diff
@@ -25,6 +25,15 @@ public FilesystemOperationDialog(FilesystemOperationDialogViewModel viewModel)
 
             ViewModel = viewModel;
             ViewModel.View = this;
+
+            //HACK: Moved setting controls to loaded event seems to fix DefaultButton focus issue.
+            Loaded += (_, _) =>
```
d2dyno1, line 30: This will cause a memory leak, consider registering the event in the UI and having the eventhandler method in code behind :)

puppetsw: Ah of course. I’ll get it fixed up.

src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
```diff
@@ -85,6 +86,14 @@ private void Grid_Loaded(object sender, Windows.UI.Xaml.RoutedEventArgs e)
                 (sender as Grid).FindAscendant<ListViewItem>().ContextFlyout = ItemContextFlyout;
             }
         }
+
+        private void FilesystemOperationDialog_OnLoaded(object sender, RoutedEventArgs e)
+        {
+            chkPermanentlyDelete.IsEnabled = ViewModel.PermanentlyDeleteEnabled;
```
gave92, line 92: PermanentlyDelete was a two-way binding, you may need to register to the Checked event and update the viewmodel.

puppetsw: I’ll make that change shortly. Thanks.

gave92: (I haven't tested this but) do all "x:Bind"s cause the focus issue? Or could it be that it's only the "ItemsSource" binding for example that causes the issue.

puppetsw: It definitely seemed to be all the x:Binds from my testing. I might just double check that though when I make these changes. 

I don’t really understand why it’s causing an issue. I tried setting the default button after the dialog had loaded but even that didn’t work.

puppetsw: Okay, apparently my previous commit "broke" the fix somehow. I think it has something to do with the order that the PrimaryButton is enabled and the ListView. The Checkbox with the x:Binds seems to not have any affect. I have just double checked that it is working as intended with how it is now. 

I may come back to this later try work through it again when I'm not so tired.

gave92:
If you don't find anything better this works but it's kinda horrid:
```cs
private async void RootDialog_Loaded(object sender, Windows.UI.Xaml.RoutedEventArgs e)
{
    ViewModel.LoadedCommand.Execute(null);
    await Task.Delay(500);
    FocusManager.TryMoveFocus(FocusNavigationDirection.Previous);
    FocusManager.TryMoveFocus(FocusNavigationDirection.Previous);
}
```

gave92:
I've pushed a possible solution that works for me, please let me know what you think / if it works for you.

puppetsw:
> I've pushed a possible solution that works for me, please let me know what you think / if it works for you.

@gave92 That doesn't seem to work for me. When deleting a file in a normal directory, the button seems to flash and then lose focus. In the recycle bin it doesn't have focus at all.

gave92:
Funny delay may be too short. Feel free to revert the changes.

puppetsw:
> Funny, delay may be too short. Feel free to revert the changes.

It's definitely a strange one.

puppetsw:
I've reverted the changes back. It seems to work again. 
Example22222

gave92:
Mm I'm seeing yo have _not_ reverted my changes, the added `RootDialog_Loaded` is still there as well as adding `IsEnabled="False"` to the ListView. In addition the `OnLoaded` method does not seem to do anything given that `ItemsSource="{x:Bind ViewModel.Items}"` was not removed from xaml.. are we merging/testing each other changes correctly? :)

puppetsw:
> Mm I'm seeing you have _not_ reverted my changes, the added `RootDialog_Loaded` is still there as well as adding `IsEnabled="False"` to the ListView. In addition the `OnLoaded` method does not seem to do anything given that `ItemsSource="{x:Bind ViewModel.Items}"` was not removed from xaml.. are we merging/testing each other changes correctly? 🤨
> In any case current code works fine for me (and it still does if I remove the `OnLoaded` method).

Hmm I’m not sure now, I thought I had reverted correctly. I’ll check it out in a bit.

puppetsw:
@gave92 It should be fixed now. It seems that doing the LoadedCommand in the Loaded event for the page was causing the issue. Just moving it to the constructor has fixed it for me.

puppetsw:
Okay, this is in a working state now and I believe the files are cleaned up.

src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
```diff
@@ -25,6 +26,15 @@ public FilesystemOperationDialog(FilesystemOperationDialogViewModel viewModel)
 
             ViewModel = viewModel;
             ViewModel.View = this;
+            ViewModel.LoadedCommand.Execute(null);
+
+            Loaded += OnLoaded;
+        }
+
+        private void OnLoaded(object sender, RoutedEventArgs e)
+        {
```
d2dyno1, line 35: Add `Loaded -= OnLoaded` here :)

d2dyno1 (dismissed):

gave92:
Current solution only works in recycle bin, outside it still focus the checkbox instead of the button. Trying a couple of options.

gave92:
@puppetsw ok pushed a change that works for me both inside and outside recycle bin. I'm hoping you can confirm it still works for you =)

puppetsw:
> @puppetsw ok pushed a change that works for me both inside and outside recycle bin. I'm hoping you can confirm it still works for you =)

@gave92 Just tested. It appears to work for me, both in and outside the recycle bin.

gave92 (approved):

yaichenbaum:
@gave92 is it possible this PR affect the status center?
image

src/Files/Dialogs/FilesystemOperationDialog.xaml.cs
```diff
@@ -25,6 +26,24 @@ public FilesystemOperationDialog(FilesystemOperationDialogViewModel viewModel)
...
+            if (primaryButton != null)
+            {
+                primaryButton.GotFocus += PrimaryButton_GotFocus;
+            }
+        }
+
+        private void PrimaryButton_GotFocus(object sender, RoutedEventArgs e)
+        {
+            (sender as Button).GotFocus -= PrimaryButton_GotFocus;
+            chkPermanentlyDelete.IsEnabled = ViewModel.PermanentlyDeleteEnabled;
```
gave92, line 45: Need to add a null check around line 45: "if (chkPermanentDelete != null)..".

gave92 (commented):
@yaichenbaum

PR #7891, opened 2022-01-23
Fixed repeatedly clicking Restore Default button causing crash

### LINKED ISSUE #6737: Clicking restore to default repeatedly crashes files
### Description

Click on the Restore to Default button in the customize icon UI will crash files.

### Steps To Reproduce
https://user-images.githubusercontent.com/17484252/140539999-90ab8bfc-ccae-4cd3-b517-5194eb19c149.mp4

### Expected behavior

It should not crash.

### Files Version

Version: 2.0.33

### Windows Version

Win 10 latest

### Relevant Assets

[debug_fulltrust.log](https://github.com/files-community/Files/files/7487297/debug_fulltrust.log)
[debug.log](https://github.com/files-community/Files/files/7487298/debug.log)

puppetsw:
**Resolved / Related Issues**
- Closes #6737

**Details of Changes**
Add details of changes here.
- Disables the button with a short delay before re-enabling.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

**Comment**
I was thinking maybe we could use events (as opposed to the delay) to restore the button once the RefreshItems() method had completed. But as it is running on a different thread I don't believe that is possible?

### DIFF
--- a/src/Files/Views/CustomFolderIcons.xaml.cs
+++ b/src/Files/Views/CustomFolderIcons.xaml.cs
@@ -30,7 +30,6 @@ namespace Files.Views
         {
             this.InitializeComponent();
             RestoreDefaultIconCommand = new AsyncRelayCommand(RestoreDefaultIcon);
-
         }
 
         protected override void OnNavigatedTo(NavigationEventArgs e)
@@ -124,9 +123,7 @@ namespace Files.Views
                     appInstance?.FilesystemViewModel?.RefreshItems(null);
                 });
             }
-
             await Task.Delay(1500);
-
             RestoreDefaultButton.IsEnabled = true;
         }
 
--- a/src/Files/Views/CustomFolderIcons.xaml.cs
+++ b/src/Files/Views/CustomFolderIcons.xaml.cs
@@ -30,6 +30,7 @@ namespace Files.Views
         {
             this.InitializeComponent();
             RestoreDefaultIconCommand = new AsyncRelayCommand(RestoreDefaultIcon);
+
         }
 
         protected override void OnNavigatedTo(NavigationEventArgs e)
@@ -111,6 +112,8 @@ namespace Files.Views
 
         private async Task RestoreDefaultIcon()
         {
+            RestoreDefaultButton.IsEnabled = false;
+
             var setIconTask = IsShortcutItem ?
                 SetCustomFileIcon(selectedItemPath, null) :
                 SetCustomFolderIcon(selectedItemPath, null);
@@ -121,6 +124,10 @@ namespace Files.Views
                     appInstance?.FilesystemViewModel?.RefreshItems(null);
                 });
             }
+
+            await Task.Delay(1500);
+
+            RestoreDefaultButton.IsEnabled = true;
         }
 
         private async Task<bool> SetCustomFolderIcon(string folderPath, string iconFile, int iconIndex = 0)
--- a/src/Files/Views/CustomFolderIcons.xaml.cs
+++ b/src/Files/Views/CustomFolderIcons.xaml.cs
@@ -10,6 +10,7 @@ using System.Windows.Input;
 using Windows.ApplicationModel.AppService;
 using Windows.ApplicationModel.Core;
 using Windows.Foundation.Collections;
+using Windows.UI.Core;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Navigation;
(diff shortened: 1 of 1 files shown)


gave92:
> I was thinking maybe we could use events (as opposed to the delay)

There are a few option:
1) Use the second parameter of RefreshItems, it's a callback that is executed after folder has finished reloading. Use Dispatcher.RunAsync to change IsEnabled property on the correct thread.
```cs
appInstance?.FilesystemViewModel?.RefreshItems(null, async () =>
{
    await Dispatcher.RunAsync(Windows.UI.Core.CoreDispatcherPriority.Normal, () => RestoreDefaultButton.IsEnabled = true);
});
```

2) Add an overload of the RefreshItems that returns a Task that you can `await`
```cs
public async Task RefreshItems(string previousDir)
{
    await RapidAddItemsToCollectionAsync(WorkingDirectory, previousDir, null);
}
```
So the RestoreDefaultIcon method becomes (no need for delay):
```cs
await CoreApplication.MainView.DispatcherQueue.EnqueueAsync(async () =>
{
    await appInstance?.FilesystemViewModel?.RefreshItems(null);
});
```
With option 2  there's also the possibility to set enabling of the button directly from XAML by doing domething like:
`IsEnabled="{x:Bind RestoreDefaultIconCommand.IsRunning, Converter={StaticResource BoolNegationConverter}, Mode=OneWay}"`

puppetsw:
> > I was thinking maybe we could use events (as opposed to the delay)
> 
> There are a couple of options:
> 
> 1. Use the second parameter of RefreshItems, it's a callback that is executed after folder has finished reloading. Use Dispatcher.RunAsync to change IsEnabled property on the correct thread.
> 
> ```cs
> appInstance?.FilesystemViewModel?.RefreshItems(null, async () =>
> {
>     await Dispatcher.RunAsync(Windows.UI.Core.CoreDispatcherPriority.Normal, () => RestoreDefaultButton.IsEnabled = true);
> });
> ```
> 
>  2. Add an overload of the RefreshItems that returns a Task that you can `await`
> 
> ```cs
> public async Task RefreshItems(string previousDir)
> {
>     await RapidAddItemsToCollectionAsync(WorkingDirectory, previousDir, null);
> }
> ```
> 
> So the RestoreDefaultIcon method becomes (no need for delay):
> 
> ```cs
> await CoreApplication.MainView.DispatcherQueue.EnqueueAsync(async () =>
> {
>     await appInstance?.FilesystemViewModel?.RefreshItems(null);
> });
> ```
> 
> With option 2 there's also the possibility to set enabling of the button directly from XAML by doing domething like: `IsEnabled="{x:Bind RestoreDefaultIconCommand.IsRunning, Converter={StaticResource BoolNegationConverter}, Mode=OneWay}"`

I think I like option 2 best. I’ll work on implementing that. Thanks for your advice and help on these PRs.

src/Files/Views/CustomFolderIcons.xaml.cs
```diff
@@ -111,14 +112,19 @@ private async void IconSelectionGrid_SelectionChanged(object sender, SelectionCh
 
         private async Task RestoreDefaultIcon()
         {
+            RestoreDefaultButton.IsEnabled = false;
```
d2dyno1, line 115: Where's RestoreDefaultButton.IsEnabled = true?

src/Files/Views/CustomFolderIcons.xaml.cs
```diff
@@ -111,14 +112,19 @@ private async void IconSelectionGrid_SelectionChanged(object sender, SelectionCh
...
                 SetCustomFileIcon(selectedItemPath, null) :
                 SetCustomFolderIcon(selectedItemPath, null);
             if (await setIconTask)
             {
                 await CoreApplication.MainView.DispatcherQueue.EnqueueAsync(() =>
                 {
-                    appInstance?.FilesystemViewModel?.RefreshItems(null);
+                    appInstance?.FilesystemViewModel?.RefreshItems(null, async () =>
+                    {
+                        await Dispatcher.RunAsync(CoreDispatcherPriority.Normal, () => RestoreDefaultButton.IsEnabled = true);
```
gave92, line 126: Here :)

gave92:
@puppetsw I see you went with option 1 at the end? Curious: where there issues with option 2?

puppetsw:
> @puppetsw I see you went with option 1 at the end? Curious: where there issues with option 2?

Ahh if I remember correctly it was something with the overloaded method and being async. Just ended up being easier to use the callback method.

gave92 (approved):

Issue #7897, opened 2022-01-23
[CRITICAL] Files Crashes When Renaming Files in Documents Folder


Gryzle:
### Description

When attempting to rename a file specifically in the Documents Folder, Files crashes. There are no special permissions within the Documents library, or on any of the folders I have attempted to rename. I renamed a file in Downloads, and it worked just fine.

### Steps To Reproduce

1. Open Files
2. Navigate to Documents
3. Rename an existing folder

### Expected behavior

Files should not crash when renaming a file in Documents.

### Files Version

Version: 2.1.3.0 OS Version: 10.0.22000.438

### Windows Version

Windows 11 22000.438

### Relevant Assets/Logs

There are no relevant issues/logs. This issue persists throughout complete restarts of the Files application.

Gryzle:
This bug appears to only effect the Documents folder, I tested it in every other library.

yaichenbaum:
Are you using the tiles layout?

gave92:
Is the library grouped? If so, could you try without grouping?

d2dyno1:
@Gryzle #7900 May fix this issue

Gryzle:
> Are you using the tiles layout?

I am, and since it was the only directory that I was using it in at the time, I misidentified the issue. The title should be "Renaming with the "Tiles" View causes Files to Crash."

StLa1:
> > Are you using the tiles layout?
> 
> I am, and since it was the only directory that I was using it in at the time, I misidentified the issue. The title should be "Renaming with the "Tiles" View causes Files to Crash."

You can rename your issue

gave92:
@yair100 why did you think of tiles layout? Is there a known issue with it?

yair100:
> @yair100 why did you think of tiles layout? Is there a known issue with it?

I noticed the issue in the tiles layout myself while working on the visual tree pull request.

yaichenbaum:
Merging with #7996

PR #7900, opened 2022-01-23
Correctly cancel all directory watchers

gave92:
**Resolved / Related Issues**
- Fixes (maybe) #7897

**Details of Changes**
Add details of changes here.
- Fixed an issue for which renaming an item in a library would sometimes result in the new item appearing multiple times (or in a crash).

Problem was that in a library WatchForDirectoryChanges is called multiple times (one for every folder the library is pointing to) and each time the `aWatcherAction` variable used to cancel the watcher gets overwritten, so only one of the watcher threads can be canceled.
Now use CancellationToken.Register() method so that each watcher instance can register it's own cleanup.

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

PR #7902, opened 2022-01-23
Removed duplicate strings

yair100:


d2dyno1:
@yair100 Could you resolve conflicts?

BanCrash:
@yair100 instead of removing duplicated english strings is there any way with MAT to translate automatically all duplicated strings when one is translated? So you only have to check that it's the same translation on all the strings.

If this is merged for example in spanish we'll have a wrong translation, since here "Unknown" is translated as "Desconocido":

```
<trans-unit id="DriveCapacityUnknown" translate="yes" xml:space="preserve">
          <source>Unknown</source>
          <target state="translated">Desconocido</target>
        </trans-unit>
```
and here as "Desconocida":
```
<trans-unit id="FileTagUnknown" translate="yes" xml:space="preserve">
          <source>Unknown</source>
          <target state="translated">Desconocida</target>
        </trans-unit>
```

yair100:
> @yair100 instead of removing duplicated english strings is there any way with MAT to translate automatically all duplicated strings when one is translated? So you only have to check that it's the same translation on all the strings.
> 
> If this is merged for example in spanish we'll have a wrong translation, since here "Unknown" is translated as "Desconocido":
> 
> ```
> <trans-unit id="DriveCapacityUnknown" translate="yes" xml:space="preserve">
>           <source>Unknown</source>
>           <target state="translated">Desconocido</target>
>         </trans-unit>
> ```
> 
> and here as "Desconocida":
> 
> ```
> <trans-unit id="FileTagUnknown" translate="yes" xml:space="preserve">
>           <source>Unknown</source>
>           <target state="translated">Desconocida</target>
>         </trans-unit>
> ```

@BanCrash not that I know of.

BanCrash:
> > @yair100 instead of removing duplicated english strings is there any way with MAT to translate automatically all duplicated strings when one is translated? So you only have to check that it's the same translation on all the strings.
> > If this is merged for example in spanish we'll have a wrong translation, since here "Unknown" is translated as "Desconocido":
> > ```
> > <trans-unit id="DriveCapacityUnknown" translate="yes" xml:space="preserve">
> >           <source>Unknown</source>
> >           <target state="translated">Desconocido</target>
> >         </trans-unit>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and here as "Desconocida":
> > ```
> > <trans-unit id="FileTagUnknown" translate="yes" xml:space="preserve">
> >           <source>Unknown</source>
> >           <target state="translated">Desconocida</target>
> >         </trans-unit>
> > ```
> 
> @BanCrash not that I know of.

:( 

In any case, I suggest to revert [a1fae4d5ece30026b51bf6bf03d95073f9e0bada commit](https://github.com/files-community/Files/pull/7902/commits/a1fae4d5ece30026b51bf6bf03d95073f9e0bada). The others seems to refer to the same context so I think there shouldn't be any issue with translation differences.

yaichenbaum:
@BanCrash the idea is to reduce the workload for new translators.

BanCrash:
> @BanCrash the idea is to reduce the workload for new translators. Considering the context is the same, I'd expect the translation to be the same across the board and this shouldn't have an impact.

But like I've said the a1fae4d5ece30026b51bf6bf03d95073f9e0bada commit introduces a changes that makes a difference in some languages.

Checking the changes not only Files.es-ES.xlf is affected, but also Files.bg.xlf, Files.pl-PL.xlf, Files.pt-BR.xlf, Files.sk-SK.xlf and Files.tr-TR.xlf. On those files `Unknown` is not always translated with the same word. I know there should be less strings but IMO that shouldn't sacrifice the language gramatical rules.

yaichenbaum:
@BanCrash the strings are used in the same context in all places, if you want, I can reset those strings but we shouldn't let that prevent us from improving the workflow for new translators.

BanCrash:
> @BanCrash the strings are used in the same context in all places, if you want, I can reset those strings but we shouldn't let that prevent us from improving the workflow for new translators.

But the point is that those strings are not in the same context. Some of them are refering to male nouns, and others to female nouns, and so the translation will be affected in languages that have differences between male and female nouns, like Spanish.

In any case, since I've saw that @d2dyno1 's language is also affected, let's see what he thinks about this.

yaichenbaum:
@BanCrash those translations need updating regardless of this change since the strings were already being used in the same context. Removing duplicate strings will actually improve the translation since the translations will be consistent throughout the app.

PR #7940, opened 2022-01-25
Added shortcut to open terminal in current directory

### LINKED ISSUE #5607: Keyboard shortcut to open current/selected directory in terminal.
Many programs such as VSCode have a global shortcut to open a terminal in the selected project or folder (typically <kbd>Ctrl</kbd> + <kbd>`</kbd>). While the native file explorer doesn't support this, I believe that it would be a convenient thing for Files to include.

**Describe the solution**
Add a keyboard shortcut to open the selected folder in the user's terminal of choice.

None.

**Scope**

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow the use of the keyboard to open the user's terminal. | Must |

**System Information**
 - OS Version: 11 22000.100
 - App version: 1.9.15.0

**Additional comment**
None

puppetsw:
**Resolved / Related Issues**
- Closes #5607

**Details of Changes**
Add details of changes here.
- Added shortcut key Ctrl+` to open default terminal in current working directory.

**Validation**
- [x] Built and ran the app


d2dyno1:
@yair100 Right keyboard shortcut?

d2dyno1 (dismissed):

yaichenbaum:
@d2dyno1 the correct shortcut would be <kbd>Ctrl</kbd> + <kbd>`</kbd>).

yaichenbaum:
@puppetsw you can use the keyboard code instead virtual key, I believe we do this for a couple of other shortcuts throughout the app.

puppetsw:
> @puppetsw you can use the keyboard code instead virtual key, I believe we do this for a couple of other shortcuts throughout the app.

@yair100 I might be missing something simple here, but I don't think this is possible to do through the KeyboardAccelerator method. The keycode for ` is 96, which is NumPad0 in the VirtualKey enum. Would we need to use the KeyDown event or something similar for this one?

yair100:
> > @puppetsw you can use the keyboard code instead virtual key, I believe we do this for a couple of other shortcuts throughout the app.
> 
> @yair100 I might be missing something simple here, but I don't think this is possible to do through the KeyboardAccelerator method. The keycode for ` is 96, which is NumPad0 in the VirtualKey enum. Would we need to use the KeyDown event or something similar for this one?

It looks look it

puppetsw:
Probably not the most elegant solution. But it seems to work. Maybe a shortcut manager or something would help and eliminate duplicate code between different pages.

yaichenbaum:
@puppetsw can you try reverting the last commit and using this code instead? `case (true, true, false, true, (VirtualKey)192):`

puppetsw:
> @puppetsw can you try reverting the last commit and using this code instead? `case (true, false, false, true, (VirtualKey)192):`

I think that would work, but the xaml doesn't seem to trigger the KeyAccelerator event with

```xml
<KeyboardAccelerator
    Key="{x:Bind AccentKey}"
    Invoked="KeyboardAccelerator_Invoked"
    IsEnabled="{x:Bind IsCurrentInstance, Mode=OneWay}"
    Modifiers="Control,Shift" />
```

With AccentKey being ```public readonly VirtualKey AccentKey = (VirtualKey)192;```

yaichenbaum:
@puppetsw I don't think we need the KeyboardAccelerator to be declared in xaml

yaichenbaum:
@puppetsw you might need to remove the shift modifier.

puppetsw:
> @puppetsw you might need to remove the shift modifier.

I'll get back to this PR shortly. The problem I was having with it, is that the ` (accent key) isn't picked up by the KeyboardAccelerator events. So it's going to have to be handled in the PreviewKeyDown event. In my original commit I've put it in the CoreWindow keydown event, but I'll change this.

yaichenbaum:
> > @puppetsw you might need to remove the shift modifier.
> 
> I'll get back to this PR shortly. The problem I was having with it, is that the ` (accent key) isn't picked up by the KeyboardAccelerator events. So it's going to have to be handled in the PreviewKeyDown event. In my original commit I've put it in the CoreWindow keydown event, but I'll change this.

Would it work to handle it with the rest of the shortcuts?

puppetsw:
> > > @puppetsw you might need to remove the shift modifier.
> > 
> > 
> > I'll get back to this PR shortly. The problem I was having with it, is that the ` (accent key) isn't picked up by the KeyboardAccelerator events. So it's going to have to be handled in the PreviewKeyDown event. In my original commit I've put it in the CoreWindow keydown event, but I'll change this.
> 
> Would it work to handle it with the rest of the shortcuts?

Unfortunately it doesn't work with the existing KeyboardAccelerator_Invoked method (in the ModernShellPage, and ColumnShellPage)

puppetsw:
Hmm the merge seems to have broken some things

puppetsw:
I think I'm just going to close this PR and start a new branch. There are only 2 files that need to be changed.

Issue #1940, opened 2020-09-18
Add option for compact layout


soraphis:
**Is your feature request related to a problem? Please describe.**
Its annoying that so much space is wasted in comparison to the default explorer

**Describe the solution you'd like**
An option to reduce the font size

**Describe alternatives you've considered**
Changing the windows scale seems not to work on a per-app basis on UWP apps

**Additional context**
comparison screenshot:
left the old explorer showing 27 items, right the files-uwp showing 16 on the same window size. 
![image](https://user-images.githubusercontent.com/3066582/93582486-d7471700-f9a2-11ea-816e-4b9a13ce8001.png)

| Capability | Priority |
| :---------- | :------- |
| This proposal will add a setting to turn on the compact layout | Must |
| This option will use the compact styles for the sidebar and details layout | Must |

duke7553:
Some of the work that is currently in the proposal drafting stage should pave the way for density adjustments to the List control. We adjusted the default font size several months ago, but recognize the padding isn't as suitable as it could be for some users.

Jaiganeshkumaran:
Not needed. Already the size is so small and reducing it will make it unusable with touch. The File Explorer is an old application designed when you only used your PC on a desk with a mouse so that's why it's too small. This application should also work with touch and having it too small will be bad.

generalguy41:
@Jaiganeshkumaran I disagree that it's not needed. As a primarily desktop user, I think that there should be an option to have a "compact" mode. By default, it could use the more spaced out mode, but I'd like to see the option available at least.

Jaiganeshkumaran:
@generalguy41 I'm a desktop user too and don't have a touch screen but it shouldn't be that small

generalguy41:
@Jaiganeshkumaran Agree to disagree, I guess. I would much rather have an option to show more items in the window and have a more compact view. Ultimately the decision isn't up to me, but I'd seriously hope it would be taken it into consideration for more "power users" I guess.

perplexityjeff:
Searched as I am having this issue as well. I find that from a productivity standpoint that the old explorer shows you so much more information at a glance. For me at least an option to tweak this to my liking would be fantastic. Either font settings and/or the padding.

trama:
[...]

Issue #7625, opened 2022-01-14
📢📢Switching the license to MIT


yaichenbaum:
I'm pleased to announce our plans to relicense the Files repository under the MIT license. The reason for this change is to allow other projects to be able to consume the innovation from Files more broadly, even if they happen to be closed source. The Files Community bolstered by your contributions isn't going anywhere, and you can trust that org members will preserve that.

Thank you for your contributions to Files. Please let us know if you're okay with us switching from GPL to MIT by commenting on this issue. If you don't agree to the change, we'll have to remove your contributions from Files.

ChihweiLHBird:
It's okay for me. Good luck!

chingucoding:
Good with me!

yaichenbaum:
@marciozomb13 @mafra99 @petrjancar @SimoneRaspagni @MelvinSnijders @LeonHoog @mate131909 @zvizvi @scls19fr @onein528 @nik2143 @alessandrocaseti @BuraChuhadar @femboy-tech @dileepsam @SHADOW-XIII @Raithlin @h1dden-da3m0n @MichaelSorensen @avknaidu @bl-ue @heistema @tsakalerisk @PabloDaniel0 @Gitoffthelawn @JordyEGNL @Coding-Young @jiejasonliu @BlairCurrey @TomaszSteblik @timwenger @typeling1578 @vikashraghavan @jovanzers @heckelson @Athey81 @AJLeuer @guillaume-nizet @maikkunrk @abdonkov @VOOM108 @Johndoe0153 @ricardosousaIB @coldu @MalteTasler

TheOnlyTails:
Ok by me!

esibruti:
By me, everything ok🚀🚀

meziantou:
It's okay for me

onein528:
okey for me!!

gave92:
Fine for me!

heckelson:
Fine for me!

xwnb:
Good for me.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

linsui:
I'm OK.

SHADOW-XIII:
Fine with me

Coding-Young:
Okay for me!

marciozomb13:
No problem.

samuelrafo:
It's okay for me

Em qui., 13 de jan. de 2022 às 21:13, Yair Aichenbaum <
***@***.***> escreveu:

[...]

PR #7941, opened 2022-01-25
Added Spacing and Padding options for compact layouts

Gabryxx7:
**Resolved / Related Issues**
- Related #1940 

**Details of Changes**
Add details of changes here.
- Added a list of `ThemeResource` settings to `App.xaml` to make it easily modifiable (instead of being `StaticResource`s)
- Added two sliders Spacing and Padding to the Navbar setting dialog
- The sliders are linked to two settings `SpacingSizePx` and `PaddingSizePx` which I already reigstered to JSON and added them both to `UserSettingsService.AppearanceSettingsService`
-- The changes are updated LIVE through `ModernShellPage` and `ColumnShellPage` by calling a new abstract method in `BaseLayout` called `UpdateAppearanceSettings()`.
-- `UpdateAppearanceSettings()` then calls the method `ContainerFromItem()` implemented by the children classes such as `DetailsBrowserLayout` 
-- The new method `ContainerFromItem()` is abstract and located in `BaseLayout` so the inheriting layouts can implement it however they want to return whatever item they want from a `listedItem`.
-- All of this was needed to update Margin and Padding live since they are set in the item style template and wouldn't be updated otherwise
- Had to change some `Grid` elements' `Height` to `Auto` to automatically scale icons and avoid them getting clipped due to the `Padding` changes.

**Validation**
- [ x] Built and ran the app
- [ x] Tested the changes for accessibility

**Screenshots (optional)**
image
image

### DIFF
--- a/src/Files/App.xaml
+++ b/src/Files/App.xaml
@@ -65,6 +65,19 @@
                             <StaticResource x:Key="SidebarContentBackgroundBrush" ResourceKey="LayerFillColorDefaultBrush" />
                             <SolidColorBrush x:Key="TitlebarContentBackgroundBrush" Color="Transparent" />
                             <SolidColorBrush x:Key="NavigationViewContentBackground" Color="Transparent" />
+                            <Thickness x:Key="ExpanderContentPadding">20,0,0,0</Thickness>
+                            <Thickness x:Key="ExpanderMinHeight">1</Thickness>
+                            <Thickness x:Key="GridViewHeaderPadding">12,8,12,0</Thickness>
+                            <Thickness x:Key="GridViewHeaderMinHeight">0</Thickness>
+                            <Thickness x:Key="GridViewHeaderItemMinHeight">0</Thickness>
+                            <Thickness x:Key="ListViewRowPadding">8,0,8,8</Thickness>
+                            <Thickness x:Key="ListViewHeaderGridHeight">38</Thickness>
+                            <Thickness x:Key="ListedItemPadding">10,0</Thickness>
+                            <Thickness x:Key="ListedItemHeight">32</Thickness>
+                            <Thickness x:Key="ListViewMinHeight">1</Thickness>
+                            <Thickness x:Key="ListViewHeight">36</Thickness>
+                            <Thickness x:Key="ListViewItemMinHeight">1</Thickness>
+                            <Thickness x:Key="ListViewHeaderItemPadding">12,8,12,0</Thickness>
                         </ResourceDictionary>
                         <ResourceDictionary x:Key="Dark">
                             <Color x:Key="SolidBackgroundAcrylic">#2C2C2C</Color>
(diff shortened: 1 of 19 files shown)


R3voA3:
Fabulous!

Gabryxx7:
@yair100 No, `ColumnViewBase` inherits from `BaseLayout`, it only needs to override `ContainerFromItem()` from `BaseLayout` which I have already done. Any future layout that inherits from `BaseLayout` will only need to implement `ContainerFromItem()`  to return any `SelectorItem` (such as `ListViewItem` or `GridViewItem`) so that `BaseLayout`'s `UpdateAppearanceSettings()` can update Padding and Margin for those items.

I thought about generalizing it so that `ContainerFromItem()` only needs to return a `Control` (since `SelectorItem` inherits from `Control` and we only change Margin and Padding which are `Control`'s variables. But I thought it could have been too generic and potentially dangereous if the wrong object gets returned!

yaichenbaum:
I would change a couple things before merging this
- Move the resources to the specific layouts and controls where they will be needed so that we don't run into issues where the UI isn't supposed to have the compact mode.
- Move the settings to the settings dialog
- Limit the options to spacious, medium, and compact, it might sound trivial, but we try to keep everything limited to 3 or less options because once you start adding more than that, users will constantly request "just one more option". We've found that enforcing limits like this is the best balance between customizability and keeping the options simple for users to play around with.

Gabryxx7:
> I would change a couple things before merging this
> 
> * Move the resources to the specific layouts and controls where they will be needed so that we don't run into issues where the UI isn't supposed to have the compact mode.
> * Move the options to the settings dialog
> * Limit the options to spacious, medium, and compact, it might sound trivial, but we try to keep everything limited to 3 or less options because once you start adding more than that, users will constantly request "just one more option". We've found that enforcing limits like this is the best balance between customizability and keeping the options simple for users to play around with.

Sounds good! And yeah I thought about just having three options rather than a slider, but I guess that's still useful to find the best balance for the three options.

As discussed in the issue's thread, I'll try to scale the context menu flyout and the sidebar as well and move the option into the settings page so it's more obvious it relates to the whole UI.

As for the font scaling, I am not sure that scaling it with padding and spacing helps. What happens if a user want more "breath" in the UI but then the font and icon scale up too, making the UI too cluttered? I think maybe a separate option for small, medium, large font might be a good alternative?

yaichenbaum:
> As for the font scaling, I am not sure that scaling it with padding and spacing helps. What happens if a user want more "breath" in the UI but then the font and icon scale up too, making the UI too cluttered? I think maybe a separate option for small, medium, large font might be a good alternative?

I think we should hold off on it and wait for some feedback on the compact layout before discussing the font and icon scaling.

Gabryxx7:
> > As for the font scaling, I am not sure that scaling it with padding and spacing helps. What happens if a user want more "breath" in the UI but then the font and icon scale up too, making the UI too cluttered? I think maybe a separate option for small, medium, large font might be a good alternative?
> 
> I think we should hold off on it and wait for some feedback on the compact layout before discussing the font and icon scaling.

Sounds good! BTW should I add a shortcut to the navbar flyout from the settings, similar to "Show Hidden Items" being both in the Settings page and in the navbar flyout?

yaichenbaum:
> BTW should I add a shortcut to the navbar flyout from the settings, similar to "Show Hidden Items" being both in the Settings page and in the navbar flyout?

Probably better not to since it affects (or at least it will) other areas of the app beyond the file area.

yaichenbaum:
@Gabryxx7 before we merge this, can you add your agreement to #7625?

Gabryxx7:
Hey! Unfortunately I'll be a bit busy for another week or two with my PhD thesis. I'll try to finish it up on the weekend. Sorry about that!

yaichenbaum:
> Hey! Unfortunately I'll be a bit busy for another week or two with my PhD thesis. I'll try to finish it up on the weekend. Sorry about that!

@Gabryxx7 thank you!

Gabryxx7:
At the moment I'm adding the spacing settings to the Appearance tab with a dropdown and a few options such as Tighter/Tight/Default/Spacious/MoreSpacious
I'm just wondering where it's best to add these options? Spacing is made up of margins, padding and sometimes height, so I added a few classes:
- `Helpers.SpacingHelper` with a subclass `Helpers.SpacingHelper.Spacing` with padding and margin
- `Enums.SpacingOptions` with the options enum as per my previous message
Now the question is: Shall I make a hard-coded list of Spacing objects with the padding and margin, or is it better to define them in a config file such as `Files/App.xaml]` ?

d2dyno1:
Could you resolve conflicts? #8464 messed up a number of PRs, fortunately you can just copy-paste your changes to the files 🙂

PR #7974, opened 2022-01-27
Translated ko-kr

Johndoe0153:
**Details of Changes**
Add details of changes here.

In this work, I literally 'add and fix' the korean translation.

- Added: I added new translation string which are about added new.
- Fixed... And also I fixed some string for more naturally and also tried to provide better UX than before.

**Validation**

I couldn't test it. Because I don't have way to how can I test translated stuff 😓  But I almost sure - and hope - it will work well.

I hope this PR would be helpful for Files! Thanks for your work :)

### DIFF
--- a/src/Files/MultilingualResources/Files.ko-KR.xlf
+++ b/src/Files/MultilingualResources/Files.ko-KR.xlf
@@ -1392,7 +1392,7 @@
           <target state="translated">자세히</target>
         </trans-unit>
         <trans-unit id="No" translate="yes" xml:space="preserve">
-          <source>No</source>/
+          <source>No</source>
           <target state="translated">아니오</target>
         </trans-unit>
         <trans-unit id="UpdateConsentDialogContent" translate="yes" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


src/Files/MultilingualResources/Files.ko-KR.xlf
```diff
@@ -1384,15 +1384,15 @@
...
+          <target state="translated">Github에서 도움주기</target>
           <note from="MultilingualUpdate" annotates="source" priority="2">Please verify the translation’s accuracy as the source string was updated after it was translated.</note>
         </trans-unit>
         <trans-unit id="Details" translate="yes" xml:space="preserve">
           <source>Details</source>
           <target state="translated">자세히</target>
         </trans-unit>
         <trans-unit id="No" translate="yes" xml:space="preserve">
-          <source>No</source>
+          <source>No</source>/
```
esibruti, line 1395: @Johndoe0153 You have to remove the "/" at the end of the line

esibruti (changes requested):

src/Files/MultilingualResources/Files.ko-KR.xlf
d2dyno1: Good catch!

d2dyno1 (approved):

PR #8041, opened 2022-01-30
Improved performance of the delete confirmation dialog

gave92:
**Resolved / Related Issues**
- Fixes an issue for which deleting a single folder takes a long time to show the confirmation dialog

**Details of Changes**
Add details of changes here.
- When deleting a folder, only test the first file inside the folder to check if files can be sent to the recycle bin
- Small refactor of "Clipboard" FTP method

To test: try request deletion of the "Program Files" folder. Confirmation dialog should appear much quicker.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Launcher/MessageHandlers/FileOperationsHandler.cs
+++ b/src/Files.Launcher/MessageHandlers/FileOperationsHandler.cs
@@ -77,20 +77,12 @@ namespace FilesFullTrust.MessageHandlers
                         var operation = (DataPackageOperation)(long)message["operation"];
                         var fileList = new System.Collections.Specialized.StringCollection();
                         fileList.AddRange(fileToCopy.Split('|'));
-                        if (operation == DataPackageOperation.Copy)
-                        {
-                            System.Windows.Forms.Clipboard.SetFileDropList(fileList);
-                        }
-                        else if (operation == DataPackageOperation.Move)
-                        {
-                            byte[] moveEffect = new byte[] { 2, 0, 0, 0 };
-                            MemoryStream dropEffect = new MemoryStream();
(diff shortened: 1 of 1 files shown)


src/Files.Launcher/MessageHandlers/FileOperationsHandler.cs
```diff
@@ -821,6 +814,28 @@ await Win32API.SendMessageAsync(connection, new ValueSet()
...
+            }
+            using var shf = new ShellFolder(shi);
+            if (shf.FirstOrDefault(x => !x.IsFolder || x.Attributes.HasFlag(ShellItemAttribute.Stream)) is ShellItem item)
+            {
+                return item;
+            }
+            foreach (var shsfi in shf.Where(x => x.IsFolder && !x.Attributes.HasFlag(ShellItemAttribute.Stream)))
+            {
+                using var shsf = new ShellFolder(shsfi);
+                if (GetFirstFile(shsf) is ShellItem item2)
```
d2dyno1, line 831: Looks cool! Would it be possible to do this using a loop instead of recursion?

d2dyno1 (approved):

src/Files.Launcher/MessageHandlers/FileOperationsHandler.cs
gave92: Probably, I understand that there's a programming theorem that says "every recursion be converted into iteration" :) In this case I find recursion more intuitive, but can be changed if you think it's necessary.

d2dyno1: Ah, can be left as is :)

PR #8042, opened 2022-01-30
Stop loading icons when the conflict dialog is closed

gave92:
**Resolved / Related Issues**
- Fixes an issue for which UI can be unresponsive after canceling the conflicts dialog due to loading of icons not being stopped

**Details of Changes**
Add details of changes here.
- Stop icon loading when conflicts dialog is being closed
- Use ParallelForEach to throttle the number of icons being loaded concurrently

**Validation**
- [x] Built and ran the app


d2dyno1:
nvm, sorry

yaichenbaum:
@gave92 can you resolved the merge conflicts?

src/Files/ViewModels/Dialogs/FilesystemOperationDialogViewModel.cs
```diff
@@ -284,29 +298,30 @@ public static FilesystemOperationDialog GetDialog(FilesystemItemsOperationDataMo
...
         {
-            await Task.Run(() => Task.WhenAll(items.ToList().Select(async (item) =>
+            await items.ParallelForEach(async (item) =>
             {
                 try
                 {
                     var iconData = await FileThumbnailHelper.LoadIconFromPathAsync(item.SourcePath, 64u, Windows.Storage.FileProperties.ThumbnailMode.ListView);
                     if (iconData != null)
                     {
+                        if (token.IsCancellationRequested) return;
```
d2dyno1, line 316: Could you move it to the top?

src/Files/Dialogs/FilesystemOperationDialog.xaml
```diff
@@ -21,6 +21,11 @@
     SecondaryButtonText="{x:Bind ViewModel.SecondaryButtonText, Mode=OneWay}"
     Style="{StaticResource DefaultContentDialogStyle}"
     mc:Ignorable="d">
+    <i:Interaction.Behaviors>
+        <icore:EventTriggerBehavior EventName="Closing">
+            <icore:InvokeCommandAction Command="{x:Bind ViewModel.ClosingCommand, Mode=OneWay}" />
```
d2dyno1, line 26: Won't trigger the command because command args don't match (?)
Either way, I prefer to have the event handler in the codebehind calling a command in the ViewModel

src/Files/ViewModels/Dialogs/FilesystemOperationDialogViewModel.cs
```diff
@@ -284,29 +298,30 @@ public static FilesystemOperationDialog GetDialog(FilesystemItemsOperationDataMo
...
                         await Windows.ApplicationModel.Core.CoreApplication.MainView.Dispatcher.RunAsync(Windows.UI.Core.CoreDispatcherPriority.Low, async () =>
                         {
                             item.ItemIcon = await iconData.ToBitmapAsync();
                         });
                     }
                 }
                 catch { }
-            })));
+            }, 10, token);
```
d2dyno1, line 324: Why `10`?

src/Files/Dialogs/FilesystemOperationDialog.xaml
gave92: Like in fc5f172190fd557c3541eb0477186bb4aa96d9b9?

src/Files/ViewModels/Dialogs/FilesystemOperationDialogViewModel.cs
```diff
@@ -106,16 +108,28 @@ public bool MustResolveConflicts
 
         public ICommand ApplyToAllCommand { get; private set; }
 
+        public ICommand ClosingCommand { get; private set; }
+
+        private CancellationTokenSource ClosingCts { get; }
+
         public FilesystemOperationDialogViewModel()
         {
+            ClosingCts = new CancellationTokenSource();
```
d2dyno1, line 117: One more thing. I think the cts needs to be disposed

gave92: Thankfully no, my understanding is that calling Dispose is only required if:
- you use CancelAfter which sets up a timer
- you access the WaitHandle member
As you can see in ItemsViewModel we never Dispose of cancellation tokens as it tends to create issues when you try to cancel a bunch of threads and then the token is Disposed before all of them have been stopped.

d2dyno1: Nice! Didn't know that

src/Files/Dialogs/FilesystemOperationDialog.xaml
d2dyno1: The reason is to prevent calling a method when it can't be executed.
image
(CommunityToolkit.Mvvm.Input.RelayCommand<T>)

d2dyno1 (approved):

PR #8099, opened 2022-02-01
Update Files.fr-FR.xlf

yair100:
Reverted changes from @scls19fr


cinqmilleans:
@yair100 Why this pr? The translations were correct but this pr adds empty texts.

yair100:
> @yair100 Why this pr? The translations were correct but this pr adds empty texts.

This is a necessary step as part of the move the MIT. Would you be interested in translating these strings?

cinqmilleans:
Yes, but it will be the same translation.  It's the only suitable one.

yaichenbaum:
> Yes, but it will be the same translation. It's the only suitable one.

I think that's fine @scls19fr, can you confirm?

d2dyno1 (approved):

PR #8115, opened 2022-02-02
Cleaned commands code

cinqmilleans:
This pr fixes some properties of commands that return a new instance on each call. Now they are initialized only once in the constructor. I take this opportunity to make other small corrections to names and styles.

Commands are sometimes fields and sometimes properties, inconsistently and sometimes in the same file. This pr does not change that. 

**Validation**
- [ ] Built and ran the app


d2dyno1 (changes requested):

cinqmilleans:
@d2dyno1 It's done, well seen.

d2dyno1 (approved):
lgtm

PR #8162, opened 2022-02-05
WIP: linger

duke7553:
TBA

### DIFF
--- a/src/Files.Launcher/Program.cs
+++ b/src/Files.Launcher/Program.cs
@@ -14,6 +14,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Windows.Foundation.Collections;
 using Windows.Storage;
+using static FilesFullTrust.Win32API;
 
 namespace FilesFullTrust
 {
@@ -21,6 +22,8 @@ namespace FilesFullTrust
     {
         public static Logger Logger { get; private set; }
         private static readonly LogWriter logWriter = new LogWriter();
+
+        private static bool shouldPrelaunchBeforeTerminate;
 
         [STAThread]
         private static async Task Main(string[] args)
@@ -173,8 +176,17 @@ namespace FilesFullTrust
             switch (arguments)
             {
                 case "Terminate":
+                    if (shouldPrelaunchBeforeTerminate)
+                    {
+                        ApplicationActivationManager appActiveManager = new ApplicationActivationManager();
+                        appActiveManager.ActivateApplication((string)message["PrelaunchAppId"], null, ActivateOptions.Prelaunch, out uint pid);
+                    }
                     // Exit fulltrust process (UWP is closed or suspended)
                     appServiceExit?.Set();
+                    break;
+
+                case "PrepareForPrelaunch":
+                    shouldPrelaunchBeforeTerminate = true;
                     break;
 
                 case "Elevate":
--- a/src/Files/App.xaml.cs
+++ b/src/Files/App.xaml.cs
@@ -27,9 +27,11 @@ using Windows.ApplicationModel;
 using Windows.ApplicationModel.Activation;
 using Windows.ApplicationModel.Core;
 using Windows.ApplicationModel.DataTransfer;
+using Windows.Foundation.Collections;
 using Windows.Foundation.Metadata;
 using Windows.Storage;
 using Windows.UI.Core;
+using Windows.UI.Core.Preview;
 using Windows.UI.Notifications;
 using Windows.UI.ViewManagement;
 using Windows.UI.Xaml;
@@ -215,12 +217,19 @@ namespace Files
 
             var rootFrame = EnsureWindowIsInitialized();
 
-            if (e.PrelaunchActivated == false)
+            if (AppInstance.GetInstances().Any() && !AppInstance.GetInstances().FirstOrDefault(x => x.IsCurrentInstance).Key.Equals(Program.PrelaunchInstanceKey))
+            {
+                AppInstance.Unregister();
+            }
+
+            if (!e.PrelaunchActivated)
             {
                 if (canEnablePrelaunch)
                 {
                     TryEnablePrelaunch();
                 }
+
+                AppInstance.Unregister();
 
                 if (rootFrame.Content == null)
                 {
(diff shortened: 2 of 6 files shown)


src/Files/App.xaml.cs
```diff
@@ -274,12 +273,25 @@ private async void App_CloseRequested(object sender, SystemNavigationCloseReques
...
+                if (plInstance.IsCurrentInstance)
+                {
+                    AppInstance.Unregister();
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
```
duke7553, line 294: unused

duke7553:
This works flawlessly until after the user launches a prelaunched instance. Basically, the problem is that only instances that were NOT originally started with prelaunch are able to prelaunch another at app startup. 

Ex: User opens the app for the first time, closes it, and reopens the app. The copy prelaunched by the first app instance launches very quickly like we want. During this copy's activation redirection, it goes through the OnLaunched method in App.cs again--where it invokes the FullTrust component to initiate another prelaunch. Unfortunantely, the app never is launched in the background and closing it leaves the user with no prelaunched app instance to redirect to (causing the slower startup)

* This could be because there still might be an instance with a `Key == "PrelaunchedInstance"` that wasn't unregistered properly and our multi-instancing code [as a result] wouldn't try to prelaunch another copy.

I'm honestly stumped for now.

Any ideas from org members are welcome. Feel free to reach out through the necessary internal channel if needed.

src/Files.Launcher/Program.cs
```diff
@@ -22,6 +23,8 @@ internal class Program
         public static Logger Logger { get; private set; }
         private static readonly LogWriter logWriter = new LogWriter();
 
+        private static bool shouldPrelaunchBeforeTerminate;
```
duke7553, line 26: unused

gave92:
Pushed some changes I have yet to test: idea is to use CoreWindow_Activated to launch a new Prelaunch instance when the old one has been opened.

src/Files.Launcher/Files.Launcher.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net6.0-windows10.0.22000.0</TargetFramework>
+    <TargetFramework>net6.0-windows10.0.19041.0</TargetFramework>
```
duke7553, line 3: @gave92 Was this intentional?

gave92: No, I need it to compile (I'm on VS 2019) but was not supposed to be committed.

src/Files/App.xaml.cs
```diff
@@ -315,17 +320,30 @@ private Frame EnsureWindowIsInitialized()
...
                 ShowErrorNotification = true;
-                ApplicationData.Current.LocalSettings.Values["INSTANCE_ACTIVE"] = Process.GetCurrentProcess().Id;
+
                 if (MainViewModel != null)
                 {
                     MainViewModel.Clipboard_ContentChanged(null, null);
                 }
+
+                var proc = Process.GetCurrentProcess();
+                ApplicationData.Current.LocalSettings.Values["INSTANCE_ACTIVE"] = proc.Id;
```
gave92, line 336: TODO: docs suggest using `VisibilityChanged` instead of `Activated` to check if the prelaunch instance has been opened.

gave92, line 336: TODO: docs suggest using `VisibilityChanged` instead of `Activated` to check if the prelaunch instance has been opened.

gave92, line 336: TODO: docs suggest using `VisibilityChanged` instead of `Activated` to check if the prelaunch instance has been opened.

yaichenbaum: Would that prevent the sidebar items from loading?

gave92: It would (at least the icons and some of the sections), until the app gets actually open. I don't really know if leaving stuff like this would be a problem, it's just that the [docs](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/handle-app-prelaunch) state: `Apps should not perform long running operations during prelaunch because the app will terminate if it can't be suspended quickly.`
But how long is "long"? Does it hold only for prelaunch by the system or also for our "manual" prelaunch?

gave92: I have yet to test this in Release (Azure won't compile, @yair100 do you know why?) but if it works I'd rather keep stuff as is (load everything).

yaichenbaum: It takes a couple of seconds until it suspends the process in release mode.

yaichenbaum: The pipeline is failing because it's trying to test the suspended process. @chingucoding is it possible to force the ci to test the main process and not the addiitonal process that's pre launched?

yaichenbaum:
It looks like "continue where you left off" is broken.

gave92:
I think continue where you left off shows "old" open tabs (tabs that were open when the prelaunch instance was set up). We might need to delay loading tabs until instance is actually open?

yaichenbaum:
> I think continue where you left off shows "old" open tabs (tabs that were open when the prelaunch instance was set up). We might need to delay loading tabs until instance is actually opened?

Agreed 👍

gave92:
Uops sorry for the force-push, I forgot this is not one of my branches 😅.
In any case latest commit removes the calls to "TerminateUwp" as @duke7553 did in the beginning as they appear to cause issues with prelaunch in Release mode.
Unfortunately this brings back an old issue were multiple suspended instances remain in Task Manager if you open Files from the taskbar or command line..

src/Files/App.xaml.cs
chingucoding: Sorry for the late reply. While this is marked as resolved, are we sure that we succeeded the test on purpose or rather on accident? And if there is a way to find the main process, sure we can force testing the main process.

gave92: 100% solved by accident 😂 so perhaps there's a more robust way to handle this.

chingucoding: The question would be how you find that window then. The TLDR of the test is that we need to find the process ID for the window so that Axe can scan the app correctly. Currently, we are just finding the first process with the name "Files".

PR #8187, opened 2022-02-06
Correctly cancel ProcessOperationQueue

gave92:
**Details of Changes**
Add details of changes here.
- Fixes a bug introduced in #7900 where the ProcessOperationQueue would not be canceled when switching directory or refreshing.

This should fix the warning/error seen in recent debug logs:
`2022-02-05 09:53:16.3558|WARN|ApplyFilesAndFoldersChangesAsync|The application called an interface that was marshalled for a different thread. (Exception from HRESULT: 0x8001010E)`

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

PR #8190, opened 2022-02-06
Changed default value for showing files starting with '.'

d2dyno1:
@gave92

### DIFF
--- a/src/Files/Services/Implementation/PreferencesSettingsService.cs
+++ b/src/Files/Services/Implementation/PreferencesSettingsService.cs
@@ -44,7 +44,7 @@ namespace Files.Services.Implementation
         
         public bool ShowDotFiles
         {
-            get => Get(false);
+            get => Get(true);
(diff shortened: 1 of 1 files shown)


gave92 (approved):

d2dyno1:
Oops
image

gave92 (approved):

src/Files/Services/Implementation/PreferencesSettingsService.cs
```diff
@@ -44,7 +44,7 @@ public bool AreSystemItemsHidden
         
         public bool ShowDotFiles
         {
-            get => Get(false);
+            get => Get(true);
```
yaichenbaum, line 47: Is this the same as file explorer?

gave92: Yep, explorer does not hide files starting with dot by default

PR #8210, opened 2022-02-07
Fix warnings (continuation)

cinqmilleans:
Here is a new batch of warning corrections. We are going forward.

**Validation**
- [ ] Built and ran the app


gave92 (commented):
Nice. Just be careful when renaming elements in xaml as stuff like this is common in the code: `listViewItem.FindDescendant("ItemName") as TextBlock`

gave92 (approved):

PR #8245, opened 2022-02-08
Removed the minimize and maximize buttons from the Properties window when using AppWindow

ahmed605:
**Resolved / Related Issues**
- None

**Details of Changes**
- Added `AppWindow` support for the file/folder Properties dialog
- Removed minimize and maximize buttons from the file/folder properties window when using `AppWindow`

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**

https://user-images.githubusercontent.com/34550324/153066237-340f0730-57e6-499e-bd05-6fedb3869a9f.mp4


src/Files/Views/Pages/Properties.xaml
```diff
@@ -38,6 +38,36 @@
...
+            VerticalAlignment="Top"
+            Background="Transparent">
+            <Rectangle
+                x:Name="CloseRect"
+                Width="46"
+                VerticalAlignment="Stretch"
+                Fill="Transparent"
+                PointerEntered="CloseRect_PointerEntered"
+                PointerExited="CloseRect_PointerExited"
+                PointerPressed="CloseRect_PointerPressed"
```
yaichenbaum, line 55: Can you add an automation name for screenreaders?

src/Files/Views/Pages/Properties.xaml.cs
```diff
@@ -70,16 +83,84 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+
+                    crossIcon.Foreground = ThemeHelper.RootTheme switch
+                    {
+                        ElementTheme.Default => new SolidColorBrush((Color)Application.Current.Resources["SystemBaseHighColor"]),
+                        ElementTheme.Light => new SolidColorBrush(Colors.Black),
+                        ElementTheme.Dark => new SolidColorBrush(Colors.White),
+                        _ => new SolidColorBrush((Color)Application.Current.Resources["SystemBaseHighColor"])
+                    };
+
+                    AppWindowTitleBarCaptionButtonsGrid.Visibility = Visibility.Visible;
```
yaichenbaum, line 105: Can you use conditional xaml in the xaml file instead of this?

ahmed605: nope, we can't use custom functions/conditions in conditional XAML
We can bind the property but it doesn't worth it, that will also increase the complexity
If we really need to use binding I would suggest binding x:Load instead
What do you think?

src/Files/Views/Pages/Properties.xaml
ahmed605: Sure, do we have a localization for the description of a close button?
Or we should just use "Close" (localized ofc)

yaichenbaum: You can reuse the localization that you used for the tooltip.

ahmed605: Done

yair100 (dismissed):

src/Files/App.xaml.cs
```diff
@@ -259,6 +259,8 @@ protected override async void OnLaunched(LaunchActivatedEventArgs e)
                     }
                 }
             }
+
+            WindowDecorationsHelper.RequestWindowDecorationsAccess();
```
gave92, line 263: To be put also in OnActivated and OnFileActivated?

yaichenbaum: I think it only needs to happen once in the app lifecycle.

ahmed605:
> Windows 10, fully black in dark mode (if possible I'd keep the current color where mica is not suported):
> 
> 
> 
> Windows 10, with system in dark mode and app in light mode
> 
> 
> 

Hmmm this seems to be caused from @yair100 commits because it looked fine (normal Dark/Light colors) before these commits on my end (on Windows 10)

src/Files/App.xaml.cs
ahmed605: Oh, so is OnActivated always called?
If so we can move that call to be only there

gave92: No I think we must put it in all 3 places. The method called depends how the app is launched.

yair100 (dismissed):

gave92 (approved):

yaichenbaum:
@ahmed605 thank you!

PR #8405, opened 2022-02-17
dev/lubl/storage-enum

lukeblevins:
TBA

For #7540


src/Files/ServicesImplementation/Win32StorageEnumerator.cs
```diff
@@ -1,13 +1,59 @@
...
 
 namespace Files.ServicesImplementation
 {
     internal sealed class Win32StorageEnumerator : IStorageEnumeratorService
     {
-        public bool IsAvailable { get; }
+        private IntPtr hFile;
+        private WIN32_FIND_DATA findData;
+
+        public string WorkingDirectory
```
d2dyno1, line 17: I would put it in BaseListedLayoutViewModel

duke7553: This would complement what's in that view model. We just needed some way to have the path be tied to the service instance in order to enable checking if that service is supported for the file system location.

gave92: Alternative would be to pass a path to Enumerate(string folderPath) / IsSupported(string folderPath)

d2dyno1: Yes! That'd make it much more flexible

duke7553: @gave92 @d2dyno1 I'm not a fan of that pattern. It seems more readable if we don't have duplicate opportunities to pass parameters for what will be the same object. 

Moreover, the way I committed means we only have to validate the path once rather than for each method that takes it for an argument.

d2dyno1: There's no real reason why we should keep the service in know of the current directory. The sole purpose of services is to process the provided data and return appropriate results. It shouldn't really know of the state of the navigated directory. There are some services which may know particular states, but this doesn't apply here. My intent was to create an abstract way directories are traversed, and so these enumerating services should only be supplied with desired path and should only supply back the result. There's no need to incorporate the current directory path nor any contextual data. That would be done in `BaseListedLayout` (or `ListingModel`) and shall be enough.

@duke7553

d2dyno1: On the other hand, I'm still thinking whether to make it transient or singleton. **If** we used the transient we'd obviously be able to use the contextual data, but that's not my intent. The question is, how will the service behave with multithreading - will singleton affect it?

d2dyno1:
I'll work on it next week 🙂

PR #8421, opened 2022-02-19
Add c# functionality "init" and "record"

cinqmilleans:
This pr enables the "init" and "record" functionality in C#.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/init
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record

**Validation**
- [ ] Built and ran the app

### DIFF
--- a/src/Files/Files.csproj
+++ b/src/Files/Files.csproj
@@ -241,6 +241,7 @@
     <Compile Include="Helpers\ZipHelpers.cs" />
     <Compile Include="Interacts\ItemManipulationModel.cs" />
     <Compile Include="ISearchBox.cs" />
+    <Compile Include="IsExternalInit.cs" />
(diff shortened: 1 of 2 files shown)


d2dyno1:
Can't we use C# 9?

cinqmilleans:
I propose this while waiting to upgrade to c#9 but it would be better to upgrade to c#9 completely. If there are no obstacles, I agree to go.

yaichenbaum:
> Can't we use C# 9?

We should definitely upgrade the c# version.

yaichenbaum (approved):

d2dyno1:
Ok, so we're already using c#9

d2dyno1 (approved):

PR #8437, opened 2022-02-20
Remove RegistryToJsonSettingsMerger class

gave92:
**Details of Changes**
Add details of changes here.
- Removes RegistryToJsonSettingsMerger class

I think we can safely remove this class now. Appcenter shows 99.2% of users are on File v2.0.34 or later.

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

PR #8457, opened 2022-02-21
Fix wrong foreground text color in zoomed-out Group headers

gave92:
**Details of Changes**
Add details of changes here.
- Fixed an issue where  switching theme would make the group headers unreadable until app restart

**Validation**
- [x] Built and ran the app

**Screenshots**


yaichenbaum (approved):
LGTM

PR #8458, opened 2022-02-21
[infrastructure] Introduce IDialogService

d2dyno1:
Related #5195


yaichenbaum:
@d2dyno1 can we target `main`?

PR #8473, opened 2022-02-23
Fix main

cinqmilleans:
The main branch is broken due to a namespace change. Here is the fix.

**Validation**
- [ ] Built and ran the app


cinqmilleans:
The problem came from me. I close this pr.

PR #8495, opened 2022-02-24
SearchPane - Phase 1

cinqmilleans:
Here is the new version of the search options panel.
This version is phase 1 which creates the panel and main options.
Many additions will follow (parsing, type, content, metadata, location, other).
A part will be moved to the backend when the ListedItem refactoring is more advanced.

https://user-images.githubusercontent.com/46631671/155551566-c6ba613b-5882-49cf-9fa1-95ea551f330f.mp4


cinqmilleans:
@yair100 Yes, the filters button is still visible because the search bar is still visible too and still usable. The panel must obviously be opened with the toggle button but it is not seen in the video.

yair100:
> @yair100 Yes, the filters button is still visible because the search bar is still visible too and still usable. The panel must obviously be opened with the toggle button but it is not seen in the video.

Does it only work on search result pages?

cinqmilleans:
@yair100 Currently, these are filters that are added to the search (it is not necessary to write anything in the search field, you can search without). The longer term objective is to have a filter bar independent of the content displayed in the layout (folder or search) without having to click on search and keep the filters from page to page. It will be very easy to set up with the new items backend while keeping the same UI.

yaichenbaum:
@cinqmilleans can you rename this feature to "Filters"?

cinqmilleans:
ok

cinqmilleans:
@yair100 It's renamed.

d2dyno1:
Part of the infrastructure?

cinqmilleans:
@d2dyno1  Not yet, but some parts will have to be migrated later when ListedItem and its factory have been refactored.

d2dyno1:
> @d2dyno1  Not yet, but some parts will have to be migrated later when ListedItem and its factory have been refactored.

@cinqmilleans I'd rather already have the BaseListedItemViewModel redesign ready so we don't have to struggle and rewrite more code (this PR) 🙂

cinqmilleans:
As you wish.

d2dyno1:
Thank you! I believe that change is really going to push things forward! 🚀

duke7553:
@cinqmilleans Really like how innovative this feature is. I think it'll be a game changer for highlighting the filtering functionality when users need it the most.

d2dyno1:
A question, how modular the new pane system really is? How easy would it be to implement a bookmarks sections of sorts that'd be able to hold references to files and folders?

src/Files.Shared/Enums/PaneContents.cs
```diff
@@ -5,5 +5,6 @@ public enum PaneContents : long
     {
         None,
         Preview,
+        Search,
```
yaichenbaum, line 8: Should this be changed to "Filters" as well?

cinqmilleans:
1) The PaneControl contains a template selector to display another UserControl based on an enum (IPaneSettingsService.Content). We can therefore easily add new panels. If the UserControl implements IPane, it can manage its own display according to the size of the layout (min/max width/height, vertical/horizontal, ...)
2) Bookmarks will be easy with the new item interface. There will be a specific implementation for bookmarks that reference another item.

yair100 (approved):

cinqmilleans:
@yair100 @d2dyno1 The button is hidden to embed the code without the functionality. The UX will be used later.

src/Files.Uwp/Converters/HasTextConverter.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Data;
+
+namespace Files.Converters
```
d2dyno1, line 5: Files.Uwp.Converters

and please change other namespaces too :)

d2dyno1:
something is broken

https://user-images.githubusercontent.com/53011783/155809033-8bc5cdd1-c321-4cad-b838-705a04bdaa8e.mp4

d2dyno1:
Also, I'd be nice if there was right click that'd remove a filter

d2dyno1:
image
UI issue?

cinqmilleans:
@d2dyno1
> When you do it in reverse, "Empty" option unselects all. Same does happen with date range

It's a feature. Initially, nothing is selected, and that means don't filter so take everything. I don't activate everything initially because it's not beautiful and not practical to select a range. When you select all, I deselect all to indicate that the selection has no effect.

> UI issue?

I don't understand your comment. What is the problem? If it's the "1 item" button, it's because the display is limited to 2 nesting levels. Beyond that, it just displays the element count. You can click on the button, or its parent, to see the details. This had been put in place because of the limited width. Now we can have more width (the width is adjustable because it is a panel).

src/Files.Uwp/Converters/HasTextConverter.cs
cinqmilleans: Ok, I'll do that.

cinqmilleans:
Ok. I did this to keep the button elegant when the cross is not displayed. We must be able to do better, I will try.

PR #8503, opened 2022-02-25
Select next item after deletion and reduce code complexity

### LINKED ISSUE #8418: Select next file in list after delete
### What's the Problem?

_No response_

### Solution/Idea

After deleting a file, the next file in the list should be selected automatically. Currently the functionality either focuses on the 'name' header if you are already at the top of the list, or the top most visible file in the list, but an arrow press or mouse click is required to actively select a file.

### Alternatives

None

### Priorities

| Capability | Priority |
| :---------- | :------- |

### Files Version

Version: 2.1.13.0 OS Version: 10.0.19043.1526

### Windows Version

_No response_

### Comments

_No response_

jiejasonliu:
**Resolved / Related Issues**
- Closes #8418 

**Details of Changes**
- Select the item following the matched item to be deleted 
   - If the item was the last item, choose the item right before it
- Reduce code complexity with selecting newly created items by reusing `OnSelectionRequestedEvent` 
   - (I did this PR awhile back)

**Validation**
- [X] Built and ran the app

**Screenshots (optional)**
Files-8418

### DIFF
--- a/src/Files.Uwp/ViewModels/ItemViewModel.cs
+++ b/src/Files.Uwp/ViewModels/ItemViewModel.cs
@@ -1891,6 +1891,7 @@ namespace Files.ViewModels
 
             bool anyEdits = false;
             ListedItem lastItemAdded = null;
+            ListedItem nextOfLastItemRemoved = null;
             var rand = Guid.NewGuid();
 
             try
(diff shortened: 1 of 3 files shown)


jiejasonliu:
Quick note. This doesn't work when deleting items in the Recycle Bin. 
We handle things slightly differently there.

Did we want this feature there as well?

yaichenbaum:
> Quick note. This doesn't work when deleting items in the Recycle Bin. We handle things slightly differently there.
> 
> Did we want this feature there as well?

It would be best to have the same behavior in recycle bin

gave92:
Creashed for me when deleting a bunch of files at once:
> System.ArgumentOutOfRangeException: 'This collection cannot work with indices larger than Int32.MaxValue - 1 (0x7FFFFFFF - 1).
Arg_ParamName_Name'

Issue could be on `main` as well, I'll check.

jiejasonliu:
> Does not happen always and I don't see how it's related with the changes in this PR, but appears ok on `main`. Can you reproduce?

I cannot seem to reproduce this issue in my branch nor in `main`. 
* Does this occur in the Recycle Bin or any ordinary directory?
* How much is a "bunch" of files -- do they include folders?

gave92 (approved):

PR #8580, opened 2022-03-05
chore: Remove, ungroup various enums

duke7553:
After this PR, we still need to determine whether some of these are good candidates for being part of Files.Uwp instead of Files.Shared.


src/Files.Shared/Enums/SortOption.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        Name,
+        DateModified,
+        DateCreated,
+        Size,
+        FileType,
+        SyncStatus,
+        FileTag,
+        OriginalFolder,
+        DateDeleted
```
yaichenbaum, line 13: Can you add comments like in the group options class?

lukeblevins: @yair100 Those won't be necessary for this PR, as I didn't author these enums.

yaichenbaum: I understand but we have the opportunity to make an improvement that will take a couple of seconds.

yaichenbaum (approved):

Issue #8244, opened 2022-02-08
Some file types are not shown in the "New" menu


manfromarce:
### Description

Some file types (e.g. bitmap image and microsoft access database) are present in the Explorer "New" menu but are not displayed in Files (see screenshot below).

### Expected behavior

The same entries as Windows Explorer should be shown.

### Files Version

Version: 2.1.7.0

### Windows Version

Windows 11 21H2 22000.469

### Relevant Assets/Logs

Windows Explorer:
![Screenshot (64)](https://user-images.githubusercontent.com/45036600/153066420-2fa8931c-a6db-4c80-8368-34457d2f70b2.png)

Files:
![Screenshot (65)](https://user-images.githubusercontent.com/45036600/153066436-b5ea9a9a-40e4-4686-90d5-d8a6f6e52f86.png)

yaichenbaum:
This issue should be resolved with #8598

PR #8598, opened 2022-03-07
Show more file types in "New" context menu

manfromarce:
**Resolved / Related Issues**
- Partially solves #8244 

**Details of Changes**
This PR removes a check that prevented some files like .accdb (Microsoft Access) that have a template with a different extension from appearing in the "New" menu. The final file is still created with the correct extension.

I wasn't able to fix .txt and .bmp entries in Windows 11 because they don't have a ShellNew entry (this is somewhat related to the new Notepad and Paint apps).

**Validation**
- [x] Built and ran the app


gave92 (approved):
Nice, thanks

PR #8654, opened 2022-03-13
Draft Files.Backend.Item

cinqmilleans:
Here is the draft of the ListedItem refactor.
I created a new project because it will contain a lot of wrapped technical code not used by other projects. The "Item" and "Operation" backend should not have to communicate.
I use modern classes and interfaces such as IAsyncEnumerable, ImmutableArray or ReadOnlyObservableCollection.
This code is a draft. Nothing works, nothing is tested or finished. There is still a lot of work.


src/Files.Backend.Item/Item/Library/LibraryItem.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Linq;
+
+namespace Files.Backend.Item
+{
+    internal class LibraryItem : FileItem, ILibraryItem
```
d2dyno1, line 5: Is the interface `ILibraryItem` for `LibraryItem` really needed? I also noticed the same behavior in other places

cinqmilleans: Yes, the interface is required. LibraryItem is internal and therefore the dll only exposes ILibraryItem. There is therefore no dual use. This allows to have several implementations if necessary, for example to optimize, for the mockup, to add decorators for the log. If we expose in dll a class, we become its prisoner. This approach avoids unanticipated problems. For example, depending on whether the user activates the ShowFolderSize option or not, the same implementation is not returned. This makes it possible to have implementations, certainly more numerous, but simpler, more maintainable, more efficient. This avoids endless testing in implementations. We do the test only once in the factory.

cinqmilleans: It also makes it easier to test the application, by sending fake data to test the interface or easily sending edge cases to test stability.

cinqmilleans:
IAsyncEnumerable allows items to be sent one by one as they become available. This avoids the need for the providers to manage the sending batches or a limited number of results themselves. This allows the results to be displayed automatically as soon as they are received.
The immutable or readonlyobservablecollection classes make it possible to secure the code by preventing the caller from modifying the results. The goal is to avoid open objects in free access. It also makes the intent clearer for classes that use these exposed properties.

lukeblevins:
@cinqmilleans Anything I or @d2dyno1 can do to help speed this up? I'm eager to get Item in the backend because I have additional work depending on it.

cinqmilleans:
Here is the progress of ListedItem.

The main problem is to avoid redoing everything "on the side" at once, because this will be the source of a large number of bugs or regressions. It is also necessary to avoid that improvements made to the code currently are lost because of a refactoring which lasts too long. So I gave up the idea of ​​providing the code all at once. So I work by successive contributions to the main code.

The next major goal is to move the StorageItems (in FileSystem) to the backend. This requires removing all calls to user settings. StorageItems should be moveable in the backend, replacing uwp namespaces. These files are hard to read, and I rework them for that. I have already made prs for this (#8919 pending). Another will be released soon. When they are ready, we will migrate them.

@lukeblevins You can look at how to remove FtpItem dependencies from FtpStorageFolder. To have a clean code, ListedItem must depend on storage and not the opposite.

lukeblevins:
@cinqmilleans That's a better idea to me.

If possible, can you also open a new umbrella issue which has a list of every prerequisite task we intend to complete for the ListedItem replacement?

cinqmilleans:
@lukeblevins Ok, I will create this issue. But the status will be the same as the above post. Thanks for the help, it's much appreciated.

PR #8712, opened 2022-03-19
Notify when pinned items config changes

### LINKED ISSUE #8556: favorites do not update on other windows
### Description

when i add a favorite in one window it does not update the favorite lists in other windows. new list only propagates to windows created after the favorite is added.

### Steps To Reproduce

1. create (2) new windows
2. pin a favorite to one of the windows
3. note that favorite list on second window does not update
4. open 3rd window - favorite now shows up properly

### Expected behavior

favorites update in all windows when added in another.

### Files Version

Version: 2.1.15.0 OS Version: 10.0.19044.1526

### Windows Version

Edition	Windows 10 Pro Version	21H2 Installed on	‎12/‎25/‎2020 OS build	19044.1526 Experience	Windows Feature Experience Pack 120.2212.4170.0

### Relevant Assets/Logs

[debug.log](https://github.com/files-community/Files/files/8171765/debug.log)

jiejasonliu:
**Resolved / Related Issues**
- Resolves #8556

**Details of Changes**
- Added watcher to pinned items config
- Fixed bug with items staying pinned when multiple items unpinned at once
    - Caused by incorrect iteration order when mutating list while enumerating
- Fixed bug with incorrect pin order when favoriting multiple items
- Synchronize updates to `ChildItems` by using `BulkConcurrentObservableCollection`
   - Watcher callback needs to marshal `RefreshAsync()` to UI thread which 
      implies that modification to `ChildItems` now needs to be thread-safe

**Validation**
- [X] Built and ran the app

**Screenshots (optional)**
Files-8556

### DIFF
--- a/src/Files.Uwp/Controllers/SidebarPinnedController.cs
+++ b/src/Files.Uwp/Controllers/SidebarPinnedController.cs
@@ -1,12 +1,15 @@
 ﻿using Files.DataModels;
 using Files.Shared.Enums;
 using Files.Filesystem;
+using Microsoft.Toolkit.Uwp;
 using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
+using Windows.ApplicationModel.Core;
 using Windows.Storage;
+using Windows.Storage.Search;
 using Files.Shared.Extensions;
 
 namespace Files.Controllers
@@ -14,6 +17,10 @@ namespace Files.Controllers
     public class SidebarPinnedController : IJson
     {
         public SidebarPinnedModel Model { get; set; }
+
+        private StorageFileQueryResult query;
+
+        private bool suppressChangeEvent;
 
         public string JsonFileName { get; } = "PinnedItems.json";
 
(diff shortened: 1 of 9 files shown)


duke7553:
@jiejasonliu Thanks for the contribution! 

I think the idea to associate pinned items with a settings folder is a good one because we can easily watch for changes. 

One question: Would it be faster to use the TryGetChangeTracker() method on the StorageFolder for this? That way, we'd avoid creating a file query.

jiejasonliu:
> I think the idea to associate pinned items with a settings folder is a good one because we can easily watch for changes.

Just to clarify, this is already done, we currently store pinned items in `ms-appdata:///local/settings/PinnedItems.json`

> Would it be faster to use the TryGetChangeTracker() method on the StorageFolder for this? 

Creating a change tracker would require creating a trigger (i.e. a background task) and a lot more boilerplate. I'm not sure what would be faster. These changes were highly inspired by previous work done (see:  `TerminalController`). Should I proceed to try and use `TryGetChangeTracker()`?

duke7553:
@jiejasonliu I don't think it would require a trigger. You should be able to invoke the TryGetChangeTracker() on any StorageFolder and then use Enable(), unless I'm mistaken.

duke7553 (approved):
LGTM

PR #8716, opened 2022-03-20
Fix crash when "show folder sizes" is enabled

### LINKED ISSUE #8413: Files crashes when folder opens and "show folder sizes" is enabled
### Description

I don't really know how this happened but it happens every time i go to downloads

### Steps To Reproduce

1. Go to Downloads
2. Files crashes

### Expected behavior

It shouldn't crash

### Files Version

Version: 2.1.13.0 OS Version: 10.0.19042.1526

### Windows Version

Edition	Windows 10 Pro Version	20H2 Installed on	‎23.‎08.‎2020 OS build	19042.1526 Experience	Windows Feature Experience Pack 120.2212.4170.0

### Relevant Assets/Logs

[debug.log](https://github.com/files-community/Files/files/8097779/debug.log)
[debug_fulltrust.log](https://github.com/files-community/Files/files/8097780/debug_fulltrust.log)

gave92:
**Resolved / Related Issues**
- Closes #8413
- Removes some unused Helper classes 

**Details of Changes**
Add details of changes here.
The issue seems to be due to the fact that the file list is changed on a background thread (in `ApplyFilesAndFoldersChangesAsync()`) while it's being enumerated in `UpdateSelectionSize()`.
This PR adds a "thread-safe" enumerator to fix this issue. Alternative could have been returning a copy of the list (with ToList()) as in [1e758ef](https://github.com/files-community/Files/pull/8716/commits/1e758ef2cda59a5c74df38e7f955f41c7767aaf8). I'm somewhat unsure what the best option is.

**Validation**
- [x] Built and ran the app


src/Files.Uwp/App.xaml.cs
```diff
@@ -64,7 +64,6 @@ sealed partial class App : Application
         public static LibraryManager LibraryManager { get; private set; }
         public static FileTagsManager FileTagsManager { get; private set; }
         public static ExternalResourcesHelper ExternalResourcesHelper { get; private set; }
-        public static OptionalPackageManager OptionalPackageManager { get; private set; } = new OptionalPackageManager();
```
yaichenbaum, line 67: Is this being used?

yaichenbaum (approved):

src/Files.Uwp/App.xaml.cs
gave92: Not referenced anywhere 👍

jiejasonliu:
Just for documentation purposes, result of #3386 removing "snapshots" of the underlying collection.

I'll pitch in and argue to reimplement thread-safe enumeration. 
Most people would assume a "concurrent" collection should both be read and write safe.
We don't want to cause a similar race condition in the future because of this leaky abstraction.

gave92:
Do you find any issue with the PR as is? Do you have suggestions for reimplement thread-safe enumeration?
I think in the future we could switch to [ReaderWriterLockSlim](https://docs.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim?view=net-6.0) instead of `lock` as it supports multiple readers simultaneously or one writer.

jiejasonliu:
> Do you find any issue with the PR as is? Do you have suggestions for reimplementing thread-safe enumeration?

It's a bit of a nightmare to synchronize this without some type of drawback. 
The only alternative is to reintroduce the "snapshot" pattern but it's expensive for large lists (e.g. a directory with thousands of files), doing this will probably regress the PR I linked above. So... I think what you have is the correct direction to go.

> Would it make sense to switch to [ReaderWriterLockSlim](https://docs.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim?view=net-6.0) instead of lock as it supports multiple readers simultaneously?

I did some internal testing (not really conclusive) but we read about 20-25 times more than we write on an average session.
We would need to do some internal testing because despite our reads vs. writes, this would substantially increase overhead and might not be better overall.

PR #8792, opened 2022-03-28
Remove each x:uid

cinqmilleans:
This pr removes all use of x:Uid (considered obsolete). This pr is necessary to prepare the move of the localization to Share, itself essential to build the new backend.

Each call is replaced with helpers:ResourceString but uses the same text as before. A cleaning of the text keys to replace the duplicates by more generic texts would be appreciable but it is not done in this pr. 

**Validation**
- [ ] Built and ran the app


yaichenbaum:
@cinqmilleans can you test these changes for accessibility?

yaichenbaum (approved):
LGTM

PR #8816, opened 2022-03-30
LocalizationExtensions

cinqmilleans:
This pr moves all localization calls to a new GetLocalized extension method in Shared. This allows to use localization in Shared and Backend, and thus to continue the refactoring which was blocked by this problem.

Shared does the translations using a function injected into its initializer by App.cs. Resources are still stored Files.Uwp, but this can be changed for tidiness. However, I'm unfamiliar with resource usage and can't do it myself (if it needs to be done).


d2dyno1:
Why would we need to translate resources in Files.Shared? If something needs to be translated, better do it in Files.Backend or Files.Uwp

cinqmilleans:
@d2dyno1 The goal isn't to translate stuff in Shared. The objective is to have all the texts located in the same place. Thus, Backend or Uwp or future projects will use the same text source. It makes more sense for uwp to get its texts from Shared (which is very generic) than from Backend (which doesn't know anything about views). An alternative is that each project has its own texts, but this complicates a lot of things for the translators and the simplicity of the code.

I'll fix the merge in a few hours.

d2dyno1:
Why not put the service in Files.Backend? The Backend already doesn't know anything related to the implementation thanks to the service so putting it in Files.Shared is overcomplicating and unnecessary. Furthermore, the UI needs to inject its implementation into Files.Shared which shouldn't rely on any UI resources

d2dyno1:
tbh, I'd close this PR if it's not stopping anything from progressing. I think the current solution is good

cinqmilleans:
@d2dyno1 What solution do you propose so that the backend can access the texts?

d2dyno1:
The backend can already access texts through [ILocalizationService](https://github.com/files-community/Files/blob/main/src/Files.Backend/Services/ILocalizationService.cs) which has an extension `.ToLocalized()` inside  [LocalizationExtensions](https://github.com/files-community/Files/blob/main/src/Files.Backend/Extensions/LocalizationExtensions.cs).

In terms of other projects accessing resources -- they shouldn't need to, instead they should provide necessary data that'll be fetched by either Files.Backend or Files.Uwp and from there strings will be localized.

(I didn't want it to be named `GetLocalized()` because that method is already a part of CommunityToolkit package)

cinqmilleans:
@d2dyno1 It's ok for me then. I close.

PR #8826, opened 2022-04-01
Mitigate issue where drag & drop wasn't working in details view.

### LINKED ISSUE #8087: Files is Unable to Drag & Drop Files Into Folder Within Same Directory
### Description

FIles is unable to use drag and drop to move files into a folder that is in the same directory as the file being moved.

### Steps To Reproduce

1. Create a File
2. Create a Folder (In the same directory as the previously created File)
3. Attempt to drag the File into the Folder

### Expected behavior

Files should be able to move the file into the folder without issues.

### Files Version

Version: 2.1.5.0 OS Version: 10.0.22000.469

### Windows Version

Windows 11 22000.469

### Relevant Assets/Logs

Files-2022-01-31

### LINKED ISSUE #8677: Pick-up and drop doesn't work properly
### Description

When I try to pick up a file and put it in a folder, many times it doesn't give the option of dropping it in the correct folder. This issue doesn't occur every time but it occurs most of the time I try to do it.

### Steps To Reproduce

1. Go to files
2. Create 2-3 folders
3. Create a file
4. Pick up the file with the mouse
5. Try to move it to the desired folder and drop it.

### Expected behavior

I should be able to drop the file in the desired folder in one go.

### Files Version

Version: 2.1.15.0 OS Version: 10.0.22000.556

### Windows Version

Windows 11 Home 21H2 OS build	22000.556

### Relevant Assets/Logs

Screenshot (206)
Screenshot (206)

itsWindows11:
**Resolved / Related Issues**
- Closes #8087
- Closes #8677

**Details of Changes**
- Mitigated issue where drag & drop wasn't working in details view, in the same directory.
  - Delayed call for clearing selection to when the item is being dropped instead of being dragged over.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.Uwp/Views/LayoutModes/DetailsLayoutBrowser.xaml
+++ b/src/Files.Uwp/Views/LayoutModes/DetailsLayoutBrowser.xaml
@@ -217,7 +217,6 @@
                         VerticalContentAlignment="Stretch"
                         tui:ScrollViewerExtensions.EnableMiddleClickScrolling="{x:Bind IsMiddleClickToScrollEnabled, Mode=OneWay}"
                         x:FieldModifier="public"
-                        AllowDrop="{x:Bind InstanceViewModel.IsPageTypeSearchResults, Converter={StaticResource BoolNegationConverter}, Mode=OneWay}"
                         CanDragItems="True"
                         ContainerContentChanging="FileList_ContainerContentChanging"
                         DoubleTapped="FileList_DoubleTapped"
--- a/src/Files.Uwp/Interacts/BaseLayoutCommandImplementationModel.cs
+++ b/src/Files.Uwp/Interacts/BaseLayoutCommandImplementationModel.cs
@@ -525,8 +525,6 @@ namespace Files.Interacts
                 return;
             }
 
-            itemManipulationModel.ClearSelection();
-
             if (Filesystem.FilesystemHelpers.HasDraggedStorageItems(e.DataView))
             {
                 e.Handled = true;
--- a/src/Files.Uwp/Interacts/BaseLayoutCommandImplementationModel.cs
+++ b/src/Files.Uwp/Interacts/BaseLayoutCommandImplementationModel.cs
@@ -654,7 +654,7 @@ namespace Files.Interacts
                     CancellationTokenSource extractCancellation = new();
                     PostedStatusBanner banner = App.OngoingTasksViewModel.PostOperationBanner(
                         string.Empty,
-                        "ExtractingArchiveText".GetLocalized(),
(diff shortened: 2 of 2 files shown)


src/Files.Uwp/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -217,7 +217,6 @@
                         VerticalContentAlignment="Stretch"
                         tui:ScrollViewerExtensions.EnableMiddleClickScrolling="{x:Bind IsMiddleClickToScrollEnabled, Mode=OneWay}"
                         x:FieldModifier="public"
-                        AllowDrop="{x:Bind InstanceViewModel.IsPageTypeSearchResults, Converter={StaticResource BoolNegationConverter}, Mode=OneWay}"
```
yaichenbaum, line 220: Does this need to be removed?

itsWindows11: Will re-add, as that wasn't the cause.

itsWindows11: > Does this need to be removed?

I'll re-add it in a bit if it wasn't the cause of the issue.

itsWindows11: @yair100 that seemed to bring the issue back after bringing it again, so I had to remove it.

winston-de: Because the default value for AllowDrop is false, this change prevents dragging items and dropping them into the blank space to move them to the current directory.

itsWindows11: > Because the default value for AllowDrop is false, this change prevents dragging items and dropping them into the blank space to move them to the current directory. Are you sure this is the cause of the issue?

Yes, removing it makes dragging work from what i can see.

yaichenbaum (approved):

BanCrash (commented):
@itsWindows11 in my test it seems that this change doesn't fix the issue:
GIF 07-04-2022 15-34-32

Also, even if the issue is caused by the AllowDrop, you should check what is causing it instead just deleting the line since there are scenarios when it should be disabled (for example when you can't write on the folder).

itsWindows11:
I'm going to close this pull request temporarily, I need to sort some things and I will reopen when I'm done.

PR #8831, opened 2022-04-01
Auto refresh file list in network shares

### LINKED ISSUE #5869: SMB / MTP won't refresh after changes
### Description

When you make a change on a SMB or MTP folder, it won't get automatically refresh.

### Steps To Reproduce

1. Go to a SMB / MTP folder.
2. Do something like pasting a file
3. Check that if you don't manually refresh it won't get updated.

### Expected behavior

Should automatically refresh.

### Files Version

1.9.25.0

### Windows Version

### Relevant Assets

[debug.log](https://github.com/files-community/Files/files/7017844/debug.log)
[debug_fulltrust.log](https://github.com/files-community/Files/files/7017845/debug_fulltrust.log)

(GIF is for SMB, but it happens the same on MTP)
GIF 19-08-2021 23-36-23

gave92:
**Resolved / Related Issues**
- Closes partially #5869

**Details of Changes**
Add details of changes here.
- Listen for directory changes in network shares using FTP

**Validation**
- [x] Built and ran the app


yair100 (dismissed):

BanCrash:
@yair100 could you deploy a beta for testing this? This bug was sneaky and while debugging (even in Release) in VS the networks folders refreshed as expected, and for some reason with VS 2022 I can't create any longer packages to install it.

gave92:
Perhaps you can test the release azure build? Should be the same as the sideload version:
https://dev.azure.com/filescommunity/db556b89-8bf0-4c03-8053-434c8f8714ff/_apis/build/builds/11016/artifacts?artifactName=drop&api-version=7.0&%24format=zip

gave92:
A couple of notes:
- I've just noticed the linked issue was also about MTP folders. This PR does not help with them, only with network folders.
- Folders will stop update if you minimize and restore the app untill you refresh once. I'm still thinking how to fix this but should not be a blocking point.

yaichenbaum:
> Folders will stop updating if you minimize and restore the app untill you refresh once. I'm still thinking how to fix this but should not be a blocking point.

Can this be resolved by automatically triggering a refresh when the app resumes from being suspended?

gave92:
Yeah that should be the easiest path. Just have to refresh network folders and not all of them.

gave92:
Yay! Fixed point (2) without the need to refresh the items.

duke7553:
@gave92 Just curious if maybe in a separate PR we can evaluate whether a [StorageLibraryChangeTracker](https://docs.microsoft.com/en-us/uwp/api/Windows.Storage.StorageLibraryChangeTracker?view=winrt-22000) would suffice for MTP locations?

gave92:
We tried that didn't we? Turned out to not work for generic folders iirc?

duke7553:
@gave92 I thought we tried that for network attached storage locations. I heard from Microsoft it does work for local StorageFolders. I can try it and get back to you 😄

gave92:
You're right we tried for network folders. From a quick check, MTP does not work either I'm getting an "invalid path" exception when getting the tracker). Let me know if you have better luck 🤞.

src/Files.Launcher/MessageHandlers/Win32MessageHandler.cs
```diff
@@ -234,6 +242,92 @@ await Win32API.SendMessageAsync(connection, new ValueSet()
...
+                    { "Name", e.Name },
+                    { "Path", e.FullPath },
+                    { "Type", e.ChangeType.ToString() },
+                    { "WatcherID", sender.GetHashCode() },
+                };
+                if (e.ChangeType == WatcherChangeTypes.Created)
+                {
+                    using var folderItem = SafetyExtensions.IgnoreExceptions(() => new ShellItem(e.FullPath));
+                    if (folderItem == null) return;
+                    var shellFileItem = ShellFolderExtensions.GetShellFileItem(folderItem);
```
BanCrash, line 311: I can't test this since the file paste stops due to an exception. First I tried just installing the azure build, and when I saw the pasting didn't complete I tried debugging with VS, and this is the result:

image

I assume that the issue will be on this line since it's happening on GetShellFileItem, maybe is necessary another checking before calling the method?

BanCrash: @gave92

gave92: Could you confirm the following?
- that the file is correctly pasted (it appears if you refresh)
- what kind of drive you're copying to (SMB share?) and from where and what kind of item
- if renaming has the same issue while deleting works ok

BanCrash: @gave92 
> I'm having trouble reproducing. Could you confirm the following?
> 
> * if the file is correctly pasted (it appears if you refresh)

The .exe file not, have the size but corrupted:
image

About image files, the first get copied correctly, but after trying pasting another Files will crash and it won't get copied.

> * what kind of drive you're copying to (another PC shared folder?)

Synology shared folder using SMB protocol.

> * what kind of item you're copying

In the original test I've tried with an .exe, .gif and .jpg. Now I also tried with a .zip file and this format seems to work as expected.
BTW, I just realize that the .jpg/.gif error and .exe error is not the same. JPG/GIF throw this one:
image
('The process does not have access to the file because it is being used by another process. )

> * if renaming and deleting work ok

Yes, both operations seems to work ok and refresh as expected.

> 
> Also, could you check what happens if you change those lines (45-47) to:
> 
> ```cs
> //folderItem.Properties.TryGetValue<string>(
> //    Ole32.PROPERTYKEY.System.ItemNameDisplay, out var fileName);
> var fileName = Path.GetFileName(folderItem.Name); // Original file name
> ```

In this case, the error will move to lines 61-62. For continuing the test I modified the code to continue without crashing with this:
image

With all this changes, the setup seems to work:
image

Also it got automatically refresh and pasting 2 images one after the other won't make Files crash. The only issue is the expected, no file data:
image

After a manual refresh:
image

gave92: Thanks I think the issue is that the created event is raised before the file has finished writing.

gave92: 7b24ce7 should fix this issue.

BanCrash:
@gave92 this is just for network drives, right? I tried 2 FTP's servers and it doesn't refresh there.

On the network drive now it's working as expected btw.

gave92:
Good. Yes it's network drives only (and not even all network drives). The other devices (ftp, mtp) don't send filesystem notifications this way (or at all).
Solution for those would simply be to update the list when a file operation done from Files completes.

gave92:
Edited PR description to not close the linked issue, since it's a partial solution.

BanCrash (approved):
LGTM. Thank you @gave92 !

PR #8847, opened 2022-04-03
Possibly fix app hung on start

gave92:
**Resolved / Related Issues**
- Fixes an issue where the app occasionally hangs on start like:

**Details of Changes**
Add details of changes here.
Debugger is showing the app hung on this line L37 in FolderSizeProvider.cs => `private readonly CoreDispatcher dispatcher = CoreApplication.MainView.CoreWindow.Dispatcher;`.
Hunch is that the line is called too early on app startup, being invoked as a consequence of a static initializer in Drives.cs => `private static readonly IFolderSizeProvider folderSizeProvider = Ioc.Default.GetService<IFolderSizeProvider>();`).
I'm not certain this PR fixes the issue as the bug is hard to reproduce, but at worst it won't change anything.

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

Issue #8305, opened 2022-02-11
Image preview generation


iNether:
### What's the Problem?

Preview generation can not be disabled

### Solution/Idea

It would be nice to be able to disable preview generation for certain file types.
Case example: regular work with large image sequences that do not need to constantly generate and save previews

### Alternatives

.

Josh-65:
Same as #6421

Issue #8257, opened 2022-02-09
Always Show Icons, never thumbnails


A-A-Haggag:
### What's the Problem?

Always Show Icons, never thumbnails feautre

### Solution/Idea

Always Show Icons, never thumbnails feautre

### Alternatives

Always Show Icons, never thumbnails feautre

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish W | Must |
| This proposal will allow end users to accomplish X | Should |
| This proposal will allow developers to accomplish Y | Could |
| This proposal will allow end users to accomplish Z | Won't |

### Files Version

last one

### Windows Version

windows 11

### Comments

if someone could add this feature it would be great help

yaichenbaum:
Merging with #6421

Issue #8234, opened 2022-02-08
Option to hide thumbnails.


cinqmilleans:
### What's the Problem?

It is not possible to hide thumbnails of images/videos in layouts. This can be disturbing for confidentiality.

### Solution/Idea

We could add an option to display only the icon of the file.

### Alternatives

None

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish W | Must |
| This proposal will allow end users to accomplish X | Should |
| This proposal will allow developers to accomplish Y | Could |
| This proposal will allow end users to accomplish Z | Won't |

yaichenbaum:
Merging with #6421

PR #8888, opened 2022-04-09
Setting Show the thumbnails for media files

### LINKED ISSUE #6421: Add an option to always show icons, never thumbnails
### What's the Problem?

I strongly prefer the option to always show only icons of all files instead of thumbnails, no matter which layout I deem fit for any folder. Unfortunately, I do not believe there is such an option currently in Files and I did not find a previous issue regarding this matter.

### Solution/Idea

Add an option to always show icons instead of thumbnails.

### Alternatives

None.

### Priorities

_No response_

### Files Version

2.0.13.0

### Windows Version

10.0.19044.1237

### Additional comment

_No response_

cinqmilleans:
Add new setting "Show the thumbnails for media files".
When the settings is disabled, the media files show the file type icon.
Closes #6421 #8234 #8257 #8305

**Validation**
- [ ] Built and ran the app

**Screenshots (optional)**
ShowThumbnails1
ShowThumbnails2
ShowThumbnails3

### DIFF
--- a/src/Files.Uwp/Strings/en-US/Resources.resw
+++ b/src/Files.Uwp/Strings/en-US/Resources.resw
@@ -2757,6 +2757,6 @@ We use App Center to track which settings are being used, find bugs, and fix cra
     <value>ex: {0}, {1}</value>
   </data>
   <data name="SettingsFilesAndFoldersShowThumbnails" xml:space="preserve">
-    <value>Show the thumbnails for media files</value>
+    <value>Show thumbnails for media files</value>
   </data>
 </root>
--- a/src/Files.Uwp/Strings/en-US/Resources.resw
+++ b/src/Files.Uwp/Strings/en-US/Resources.resw
@@ -2757,6 +2757,6 @@ We use App Center to track which settings are being used, find bugs, and fix cra
     <value>ex: {0}, {1}</value>
   </data>
   <data name="SettingsFilesAndFoldersShowThumbnails" xml:space="preserve">
-    <value>Show thumbnails for media files</value>
+    <value>Show thumbnails</value>
   </data>
 </root>
--- a/src/Files.Uwp/Strings/en-US/Resources.resw
+++ b/src/Files.Uwp/Strings/en-US/Resources.resw
@@ -2753,7 +2753,10 @@ We use App Center to track which settings are being used, find bugs, and fix cra
   <data name="Universal" xml:space="preserve">
     <value>Universal</value>
   </data>
-    <data name="DateFormatSample" xml:space="preserve">
+  <data name="DateFormatSample" xml:space="preserve">
     <value>ex: {0}, {1}</value>
+  </data>
+  <data name="SettingsFilesAndFoldersShowThumbnails" xml:space="preserve">
+    <value>Show the thumbnails for media files</value>
   </data>
 </root>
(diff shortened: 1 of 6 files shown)


src/Files.Uwp/Strings/en-US/Resources.resw
```diff
@@ -2753,7 +2753,10 @@ We use App Center to track which settings are being used, find bugs, and fix cra
   <data name="Universal" xml:space="preserve">
     <value>Universal</value>
   </data>
-    <data name="DateFormatSample" xml:space="preserve">
+  <data name="DateFormatSample" xml:space="preserve">
     <value>ex: {0}, {1}</value>
   </data>
+  <data name="SettingsFilesAndFoldersShowThumbnails" xml:space="preserve">
+    <value>Show the thumbnails for media files</value>
```
d2dyno1, line 2760: No need for "the" 🙂

src/Files.Uwp/ViewModels/ItemViewModel.cs
```diff
@@ -387,13 +389,15 @@ private async void UserSettingsService_OnSettingChangedEvent(object sender, Sett
             switch (e.SettingName)
             {
                 case nameof(UserSettingsService.PreferencesSettingsService.ShowFileExtensions):
+                case nameof(UserSettingsService.PreferencesSettingsService.ShowThumbnails):
```
d2dyno1, line 392: Since it's a privacy feature I'm not sure if it needs to be logged. Need second opinion from @yair100

src/Files.Uwp/Strings/en-US/Resources.resw
cinqmilleans: It's done. My English is too light for these nuances.

src/Files.Uwp/ViewModels/ItemViewModel.cs
yaichenbaum: AppCenter doesn't track any personal data, so this is fine.

yaichenbaum:
@cinqmilleans does this work for other file types or is it only for media files?

cinqmilleans:
@yair100 There are currently 3 icon levels
- If the file is an image or a video, a thumbnail is generated to replace the icon. Other files may have them in the future: pdf, word, url, mp3 (cover) ... But they are always media (document, web, audio ...). If the file is not media, we can't generate a thumbnail, it doesn't make sense.
- If the file contains an icon, it is used (.exe).
- Otherwise, the file type icon.

This pr just disables thumbnails. This does not impact .exes. This pr therefore only concerns the media.

yaichenbaum:
@cinqmilleans I don't usually think of pdf files as media files. Would it be possible to change the wording to "Always show icons"?

cinqmilleans:
@yair100 PDFs are not currently affected. For now, and probably forever, we only have thumbnails for images and videos. I suggest the text "Show thumbnails for images and videos" (enabled by default). We will change the text later as needed if we add more types.

yair100:
> @yair100 PDFs are not currently affected. For now, and probably forever, we only have thumbnails for images and videos. I suggest the text "Show thumbnails for images and videos" (enabled by default). We will change the text later as needed if we add more types.

It depends on the file handler but there are multiple file types which support thumbnails, I think it's best to use generic words that aren't specific to the file type.

cinqmilleans:
I renamed the option "Show thumbnails".

yair100 (approved):

d2dyno1 (approved):

yaichenbaum:
@cinqmilleans thank you!

PR #8889, opened 2022-04-09
Clean IStorageItemWithPath

cinqmilleans:
Tiny pr to clean up IStorageItemWithPath and its 2 implementations. Refactoring only.
- Deletes unused sets.
- Use 1 property instead of 2 which returns the same instance.
- More readable code.

**Validation**
- [ ] Built and ran the app

### DIFF
--- a/src/Files.Uwp/Filesystem/StorageFileHelpers/StorageFileWithPath.cs
+++ b/src/Files.Uwp/Filesystem/StorageFileHelpers/StorageFileWithPath.cs
@@ -1,37 +1,22 @@
 ﻿using Files.Filesystem.StorageItems;
 using Windows.Storage;
+using IO = System.IO;
 
 namespace Files.Filesystem
 {
     public class StorageFileWithPath : IStorageItemWithPath
     {
-        public BaseStorageFile File
-        {
-            get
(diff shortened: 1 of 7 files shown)


src/Files.Uwp/Filesystem/StorageFileHelpers/StorageFileWithPath.cs
```diff
@@ -1,37 +1,22 @@
...
-        public FilesystemItemType ItemType => FilesystemItemType.File;
+        IStorageItem IStorageItemWithPath.Item => Item;
+        public BaseStorageFile Item { get; }
 
-        public StorageFileWithPath(BaseStorageFile file)
-        {
-            File = file;
-            Path = File.Path;
-        }
+        public FilesystemItemType ItemType => FilesystemItemType.Directory;
```
gave92, line 15: FilesystemItemType.File?

cinqmilleans: It is indeed an error. Thanks I fixed it.

d2dyno1 (approved):
lgtm

PR #8891, opened 2022-04-10
Fixed an issue preventing drag and drop from working 

devovercome:
**Resolved / Related Issues**
- Related #8677

**Details of Changes**
- Fixed: wrong AllowDrop flags set on items.
- Removed: code duplication on classed derived from BaseLayout. 

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**


devovercome:
I want to note that I made some extra changes in separate commit, called "Refactoring".
Please LMK are these changes suitable and good to go as single squashed commit, or should we revert refactoring commit.

Also, I think, the issue #8677 isn't done yet. The glitch still there, but I'm not sure how to fix the UI behavior.

yaichenbaum (approved):

gave92:
Nice, what's the change that fixes the issue? Passing the `listedItem` to `InitializeDrag`?

devovercome:
> Nice, what's the change that fixes the issue? Passing the `listedItem` to `InitializeDrag`?

Yes, I passed the `listedItem` down, because `GetItemFromElement(container)` was returning a wrong item, like a wild guess.
Meanwhile, on upside the `FileList_ContainerContentChanging` knows clearly which item is processing right now via `args.Item`, therefore, there is no reason to figure out which item needs re-initializing - just pass it down from the `args`.

PR #8893, opened 2022-04-11
Change preview pane play media toggle to 'ctrl + space'

### LINKED ISSUE #7705: Change ctrl + alt + p to something easier
### What's the Problem?

Hi I'm loving the program but I'm missing a key feature for me. When I'm moving through my sample library I would like to listen quickly to the audio. Right now I just move with the up and down keys and play by pressing CTRL + ALT + P, but that shortcut is just too hard. Thank you

Screenshot

### Solution/Idea

I would be great if we could change that shortcut to "space bar" for example or having an autoplay function.

### Alternatives

I've tried QuickLook but for some reason doesn't play files that are inside of Google Drive, if they are in the desktop for example it works fine, although the audio glitches a little bit sometimes and you have to replay the sample.

### Priorities

| Capability | Priority |
| :---------- | :------- |

### Files Version

_No response_

### Windows Version

_No response_

### Comments

_No response_

### LINKED ISSUE #8108: Shortcut to play audio only works when the focus is on the preview pane
- Turn on the preview pane
- Select and audio file
- ctrl + space
- See the error

jiejasonliu:
**Resolved / Related Issues**
- Resolves #7705
- Resolves #8108

**Details of Changes**
- Shortcut to toggle playback (`TogglePlaybackAccelerator`) is now `ctrl + space`
- Add code-behind to allow toggle in details/column layout
- `Space` was being suppressed and not propagated in `KeyboardAccelerator`, now handled in `PreviewKeyDown`
- Refactor `PreviewKeyDown` to be similar in structure with `KeyboardAccelerator_Invoked`

**Validation**
- [X] Built and ran the app
  - `` ctrl + ` `` still works as expected (open terminal with correct path) after refactor
  - `ctrl + space` works as expected (item selected, preview pane shown)
  - `space` key works when renaming and isn't disabled
  - **QuickLook** works as expected with `space` and as well as all the same interactions listed above

### DIFF
--- a/src/Files.Uwp/UserControls/FilePreviews/MediaPreview.xaml
+++ b/src/Files.Uwp/UserControls/FilePreviews/MediaPreview.xaml
@@ -11,9 +11,9 @@
     mc:Ignorable="d">
     <UserControl.KeyboardAccelerators>
         <KeyboardAccelerator
-            Key="P"
+            Key="Space"
             Invoked="TogglePlaybackAcceleratorInvoked"
-            Modifiers="Control,Menu" />
+            Modifiers="Control" />
     </UserControl.KeyboardAccelerators>
     <Grid CornerRadius="{StaticResource ControlCornerRadius}">
         <MediaPlayerElement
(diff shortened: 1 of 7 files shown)


winston-de (approved):
LGTM

Note: docs will need to be updated to reflect this change in the future

PR #8903, opened 2022-04-12
Codebase: Rename NavigationToolbar to AddressToolbar

### LINKED ISSUE #5470: Code improvement: Rename NavigationToolbar to AddressToolbar

puppetsw:
**Resolved / Related Issues**
- Closes #5470

**Details of Changes**
Add details of changes here.
- Renames NavigationToolbar user control to AddressToolbar.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.Uwp/ViewModels/ToolbarViewModel.cs
+++ b/src/Files.Uwp/ViewModels/ToolbarViewModel.cs
@@ -29,13 +29,13 @@ using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Controls.Primitives;
 using Windows.UI.Xaml.Input;
-using static Files.Uwp.UserControls.INavigationToolbar;
+using static Files.Uwp.UserControls.IAddressToolbar;
 using SearchBox = Files.Uwp.UserControls.SearchBox;
 using SortDirection = Files.Shared.Enums.SortDirection;
 
 namespace Files.Uwp.ViewModels
 {
-    public class ToolbarViewModel : ObservableObject, INavigationToolbar, IDisposable
+    public class ToolbarViewModel : ObservableObject, IAddressToolbar, IDisposable
     {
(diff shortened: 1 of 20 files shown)


puppetsw:
Not sure how I feel about this one.

lukeblevins:
@puppetsw Any concerns?

puppetsw:
> @puppetsw Any concerns?

Not anymore. I think renaming the toolbar view model makes sense as well, as it is shared by the address toolbar and inner navigation.

src/Files.Uwp/ViewModels/ToolbarViewModel.cs
```diff
@@ -329,7 +329,7 @@ public CurrentInstanceViewModel InstanceViewModel
 
         private PointerRoutedEventArgs pointerRoutedEventArgs;
 
-        public NavToolbarViewModel()
+        public ToolbarViewModel()
```
yaichenbaum, line 332: Should this be `AddressToolbarViewModel`?

puppetsw: The same ViewModel is used by the InnerNavigationToolbar as well. I thought a common name would be best, though I'm happy to change it?

yaichenbaum: I'm not familiar with what it's used for off hand and it may be beyond the scope of this PR but should each control have it's own view model attatched?

lukeblevins: @yair100 I think that would be out of scope for this PR. I think the question is whether they share the same ViewModel instance or if @puppetsw just meant they are separate instances with the same, generic name.

puppetsw: @lukeblevins I believe they are both using the same instance of that ViewModel.

yaichenbaum (approved):
Great work!

lukeblevins (approved):
Thanks @puppetsw !

PR #8927, opened 2022-04-16
Use correct resource for "show item type column" menu item in Details view

gave92:
**Details of Changes**
Add details of changes here.
- Use correct resource for "show item type column" menu item in Details view

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**

image


yaichenbaum (approved):
LGTM

PR #8960, opened 2022-04-19
Disable Move operation on folders when using StorageFile APIs

gave92:
**Resolved / Related Issues**
- Fixes a possible risk of data loss when moving directories

**Details of Changes**
Add details of changes here.
- A user reported data loss when moving a folder using Files. Folder was being moved to another location on the same drive but the files were being copied one-by-one. This suggests that the move operation was being carried out using the StorageFile API instead of the shell operations, which can happen if the fulltrust process has crashed.
- This PR disables the move operation on folders using the StorageFile  API which can result in data loss e.g. if the folder contains hidden items. A dialog is shown and the user can choose to copy or cancel instead.

**Validation**
- [x] Built and ran the app

**Screenshots**

### DIFF
--- a/src/Files.Uwp/Filesystem/FilesystemOperations/FilesystemOperations.cs
+++ b/src/Files.Uwp/Filesystem/FilesystemOperations/FilesystemOperations.cs
@@ -392,7 +392,9 @@ namespace Files.Uwp.Filesystem
 
                         if (fsResult)
                         {
-                            var fsResultMove = await FilesystemTasks.Wrap(() => MoveDirectoryAsync((BaseStorageFolder)fsSourceFolder, (BaseStorageFolder)fsDestinationFolder, fsSourceFolder.Result.Name, collision.Convert(), true));
+                            // Moving folders using Storage API can result in data loss, copy instead
+                            var fsResultMove = await FilesystemTasks.Wrap(() => CloneDirectoryAsync((BaseStorageFolder)fsSourceFolder, (BaseStorageFolder)fsDestinationFolder, fsSourceFolder.Result.Name, collision.Convert()));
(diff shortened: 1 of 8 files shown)


yair100 (approved):
LGTM

lukeblevins:
@gave92 @yair100 Is it worth creating another operation for this based on copy that deletes the source location when complete?

yair100:
> @gave92 @yair100 Is it worth creating another operation for this based on copy that deletes the source location when complete?

I think it makes sense to keep it how it is for now and focus on the refactor as WinApp SDK will resolve these kinds of issues.

gave92:
> Is it worth creating another operation for this based on copy that deletes the source location when complete?
@lukeblevins this is ablut what we do today, the problem is that StorageFile api does not list hidden folders and shortcuts, so you _think_ you copied everything, but actually you're losing some elements.

lukeblevins:
@yair100 "Move --> Copy + Delete source" would most definitely be a negligible amount of effort, but I guess that's not important if this case doesn't occur often enough. 

Definitely excited for a move to IFileOperation as this will give us the ability to build an accurate, reliable progress reporting system with IOperationsProgressDialog (which afaik doesn't have to be an actual dialog or window)

lukeblevins:
Hi @gave92 ,

I'm aware now that operations on the StorageFile API ignore hidden items. 

The solution I was proposing to keep something resembling the Move feature in this case was to replace it with copy and then we can delete the source location also. This could save us from telling the user to use copy instead of move.

Not a big deal to me either way

gave92:
Copying unfortunately has the same issue: hidden items are silently skipped.
Note that this PR affects only move with StorageApi which is rarely used, most of the time we use shell IFileOperation.

PR #8962, opened 2022-04-20
Update Files.vi.xlf

YT-Advanced:
Fix Vietnamese language

### DIFF
--- a/src/Files.Uwp/MultilingualResources/Files.vi.xlf
+++ b/src/Files.Uwp/MultilingualResources/Files.vi.xlf
@@ -36,7 +36,7 @@
         </trans-unit>
         <trans-unit id="Browse" translate="yes" xml:space="preserve">
           <source>Browse</source>
-          <target state="translated">Duyệt tìm</target>
+          <target state="translated">Duyệt</target>
         </trans-unit>
         <trans-unit id="nameColumn.Header" translate="yes" xml:space="preserve">
           <source>Name</source>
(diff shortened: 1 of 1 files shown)


src/Files.Uwp/MultilingualResources/Files.vi.xlf
```diff
@@ -276,23 +276,23 @@
...
+          <target state="translated">Ghim vào Ưa thích</target>
         </trans-unit>
         <trans-unit id="BaseLayoutItemContextFlyoutProperties.Text" translate="yes" xml:space="preserve">
           <source>Properties</source>
           <target state="needs-review-translation" state-qualifier="tm-suggestion">Thuộc tính</target>
         </trans-unit>
         <trans-unit id="WelcomeDialog.Title" translate="yes" xml:space="preserve">
           <source>Welcome to Files!</source>
-          <target state="needs-review-translation">Chào mừng đến Files!</target>
+          <target state="needs-review-translation">Chào mừng bạn đến với Files!</target>
```
BanCrash, line 287: You have some lines, like this, with the state "needs-review-translation". This state is added automatically when originally modifying the original string, or added manually by the translator when he/she is not sure about the result of the translation. Is the second scenario your case? If not, please replace in this line and all others "needs-review-translation" where is not the case with the state "translated".

BanCrash: I've forgot to mention you @YT-Advanced , sorry.

PR #8967, opened 2022-04-20
Do not change file date modified when setting tag

### LINKED ISSUE #8965: Date of modification changes when setting a tag on file/folder
### Description

When adding a tag to a file/folder, the date of modification comes true (countdown starts from zero seconds).

### Steps To Reproduce

1. Create new folder;
2. Wait some time;
3. Add tag any color;
4. Look at the modification date of a file/folder.

### Expected behavior

Countdown starts from zero seconds

### Files Version

Версия: 2.1.15.0 OS Version: 10.0.19044.1645

### Windows Version

Windows 10 21H2

gave92:
**Resolved / Related Issues**
- Closes #8965

**Details of Changes**
Add details of changes here.
- Restores file date modified after setting file tag

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Launcher/MessageHandlers/FileTagsHandler.cs
+++ b/src/Files.Launcher/MessageHandlers/FileTagsHandler.cs
@@ -33,7 +33,7 @@ namespace FilesFullTrust.MessageHandlers
 
         public static bool WriteFileTag(string filePath, string tag)
         {
-            var dateOk = GetDateModifiedTime(filePath, out var dateModified); // Backup date modified
+            var dateOk = GetFileDateModified(filePath, out var dateModified); // Backup date modified
             bool result = false;
             if (tag == null)
             {
@@ -52,7 +52,7 @@ namespace FilesFullTrust.MessageHandlers
             }
             if (dateOk)
             {
-                SetDateModifiedTime(filePath, dateModified); // Restore date modified
+                SetFileDateModified(filePath, dateModified); // Restore date modified
             }
             return result;
         }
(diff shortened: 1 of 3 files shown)


yaichenbaum (approved):
LGTM

yaichenbaum:
@gave92 I merged this with `main` but I'm wondering if we should revert it as the file really was modified and can potentially cause conflicts in other areas.

gave92:
File tags are written to an alternate data stream, but doing so also affects the "main" file modified date, despite it not having actually changed. I can understand how this could be annoying for users.
I don't think that changing the file modified date can cause issues, what are you thinking to?

BanCrash:
@gave92 what do you think about adding a setting to choose this option? I'm not sure but I think there are still backup softwares that uses the modification date to choose if they had to backup the file or not.

gave92:
@BanCrash that's a valid concern, even though here we're simply suppressing the update of the date modified when setting the tag. If the file actually gets modified the date will reflect that.
I wouldn't add a setting for this, I'd simply decide between the following:
- Revert this and let the date change. Changing file properties changes the date modified too so the date changing when setting a tag should not be come as a suprise.
- Keep this untill reports that it causes issue

BanCrash:
> @BanCrash that's a valid concern, even though here we're simply suppressing the update of the date modified when setting the tag. 

That was my point, if I add a tag to the file I want to get a backup of the file with the tag attached. If not, when restoring the file I'll have to reasign all my tags if the files weren't modified after.

I don't understand neither about the surprise of the modification date, it's the expected IMO.

PR #8971, opened 2022-04-21
Fixed an issue where the tab control had the wrong color

yaichenbaum:
A recent WinUI update added a second layer of mica to the tab control causing it to have the wrong transparncy in Files. I worked around this by removing one of the visual states from the default template.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.


winston-de (approved):
Gotta love retemplating controls to fix winui bugs.

LGTM

PR #8996, opened 2022-04-25
Updated Simplified Chinese Translation

imhe6:
Add translation of newly added strings and correct several mistakes.


yaichenbaum (approved):
LGTM

PR #8999, opened 2022-04-25
Fix spacing in status bar

gave92:
**Resolved / Related Issues**
This fixes a small issue where in the Status bar when no items are selected and folder size is enabled there's a 24px space (instead of 12px) between element count and size.

**Details of Changes**
Add details of changes here.
- Put margin on each column content instead of using ColumnSpacing property

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
Thank you

PR #8952, opened 2022-04-18
[Refactor] Decouple *Manager classes from the Sidebar

gave92:
**Details of Changes**
Add details of changes here.

Refactored "Manager" classes (DrivesManager, CloudDrivesManager, FileTagsManager, LibraryManager, NetworkDrivesManager, WSLDistroManager, SidebarPinnedModel) so that they are decoupled from the Sidebar.
- Only SidebarViewModel should access sidebar items -> moved SidebarItems collection from SidebarControl to SidebarViewModel and made it non static
- The "Manager" classes only load the items -> the SidebarItems collection is updated by SidebarViewModel  through events
- Reduced code duplication -> removed duplicate SyncSideBarItemsUI(), Update\*SectionVisibility(), Remove\*SideBarSection() methods
- Removed the SidebarItemsSemaphore

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #9002, opened 2022-04-25
Fix crash due to concurrent edits of sidebar lists

gave92:
**Resolved / Related Issues**
PR #8952 introduced a bug for which sometimes the app crashes on start due to the favorites list being concurrently modified.

**Details of Changes**
Add details of changes here.
- Add proper lock() statements when modifying favorites and other sidebar sections related lists (like was already the case with Drives)
- Remove a couple of extra "ToList()" calls when safe to do so.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Uwp/BaseLayout.cs
+++ b/src/Files.Uwp/BaseLayout.cs
@@ -367,7 +367,7 @@ namespace Files.Uwp
                 return;
             }
 
-            foreach (var item in items.ToList()) // ToList() is necessary
+            foreach (var item in items)
             {
                 if (item != null)
                 {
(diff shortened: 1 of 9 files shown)


src/Files.Uwp/BaseLayout.cs
```diff
@@ -367,7 +367,7 @@ public virtual void ResetItemOpacity()
                 return;
             }
 
-            foreach (var item in items.ToList()) // ToList() is necessary
```
gave92, line 370: ToList() in _not_ necessary here :) Only useful when either:
- the list is modified inside the forearch
- the foreach contains an "await"

lukeblevins (approved):
LGTM

d2dyno1 (approved):

PR #9010, opened 2022-04-26
Execute Files by protocol in launcher

### LINKED ISSUE #8951: Error on setting files as a default file manager to replace Windows Explorer
### Description

After I set the default file manager to Files (Settings>Experimental>Set Files as the default file manager), upon clicking Start>File Explorer at the sidebar an error window saying:
image
It couldn't find the Files app. Please fix! Thank you

### Steps To Reproduce

1. Open Files
2. Click Settings
3. Click Experimental tab
4. Enable 'Set Files as the default file manager'
5. Press Start
6. Click File Explorer 
Untitled
7. An error appears 
image

### Expected behavior

This is what happens:
image

### Files Version

Version: 2.1.24.0 OS Version: 10.0.19044.1645

### Windows Version

Edition	Windows 10 Education Version	21H2 Installed on	‎1/‎10/‎2021 OS build	19044.1645 Experience	Windows Feature Experience Pack 120.2212.4170.0

### Relevant Assets/Logs

[debug.log](https://github.com/files-community/Files/files/8506468/debug.log)
[debug_fulltrust.log](https://github.com/files-community/Files/files/8506469/debug_fulltrust.log)

gave92:
**Resolved / Related Issues**
- Closes #8951 (again)

**Details of Changes**
Add details of changes here.
This PR should fix the error that some users are reporting when setting Files as default. It does it by invoking Files by protocol ("files-uwp:") instead of command line ("files.exe") which appears to fix the issue.

image

It would be great if a couple of people could test this.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.OpenDialog/FilesLauncher/FilesLauncher.cpp
+++ b/src/Files.OpenDialog/FilesLauncher/FilesLauncher.cpp
@@ -172,7 +172,7 @@ int WINAPI WinMain(HINSTANCE hInstance,
 
 		std::wstring uriWithArgs = L"files-uwp:?cmd=" + str2wstr(wstring_to_utf8_hex(args));
 
-		std::wcout << L"Invoking: " << args << std::endl;
+		std::wcout << L"Invoking: " << args << L" = " << uriWithArgs << std::endl;
 
 		SHELLEXECUTEINFO ShExecInfo = { 0 };
 		ShExecInfo.cbSize = sizeof(SHELLEXECUTEINFO);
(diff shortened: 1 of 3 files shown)


gave92:
A user experiencing the issue has confirmed that the solution works.

winston-de (approved):
LGTM

note: 
if an older version of files (eg. non-preview version) is set as the handler for the files protocol, that version will open an empty window (black with no content).

yaichenbaum:
> LGTM
> 
> note: if an older version of files (eg. non-preview version) is set as the handler for the files protocol, that version will open an empty window (black with no content).

Should this be resolved before merging?

gave92:
Can't be resolved, but it's only an issue if you have 2 files versions installed at the same time, right @winston-de ?

winston-de:
Correct. Can be easily fixed by changing which version handles the protocol. Just wanted to point it out in case someone runs into the issue in the future.

PR #9018, opened 2022-04-27
Move CommonPaths to Shared

cinqmilleans:
Move CommonPaths to Shared.

- This pr moves CommonPaths to Shared. It is necessary to move the storage to the backend soon.
- The CommonPaths file has been restyled and its dictionary has become readonly.
- A cleanup has been done in the using files that depend on CommonPaths.
- This pr also moves the localization service and extension from Backend to Shared because CommonPaths uses it. This is its most natural place.

**Validation**
- [ ] Built and ran the app


src/Files.Shared/Extensions/LocalizationExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿using CommunityToolkit.Mvvm.DependencyInjection;
+using Files.Shared.Services;
+
+namespace Files.Shared.Extensions
+{
+    public static class LocalizationExtensions
```
d2dyno1, line 6: That's not supposed to be here 🤔

cinqmilleans: Sorry, I don't understand your comment.

d2dyno1: I mean, we decided that it should stay in Files.Backend

cinqmilleans: An alternative solution would be to place CommonPaths in Backend. It's more logical.

cinqmilleans:
@d2dyno1 I open a new pr to move CommonPaths to Backend. This is therefore closed.

PR #9035, opened 2022-04-29
fix(paste): Use FullTrustProcess operation to inspect clipboard items

lukeblevins:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #7928 

**Details of Changes**
Switch to using FullTustProcess to inspect the "data package" because copying certain items from Files sometimes has to add the items to the Clipboard via SetFileDropList() in the full trust component. When pasting the items, inspecting the resultant data package would only find 16 items. This could be an OS bug, but this way all operations involving paste are capped at output of 16 items.

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [x] Tested the changes for accessibility

**Screenshots (optional)**
N/A

PR #9014, opened 2022-04-27
Improved drag and drop with virtual storage items

lukeblevins:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #6941 
- Closes #8420 

**Details of Changes**
Add details of changes here.
- Create the concept of virtual storage item to help with dragging outside of Files app window to another app

## Special thanks to "High Caffeine Content" (@steventroughtonsmith) for making this and future PRs possible

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [x] Tested the changes for accessibility

PR #9048, opened 2022-04-30
feat(VirtualStorageItem): instantiation from filesystem path

lukeblevins:
**Resolved / Related Issues**
N/A

**Details of Changes**
Expands the capability of our new VirtualStorageItem impl to support instantiation solely from a filesystem path. For this feature, we leverage the fast FindFirstFileExFromApp() method to get enough item properties to construct the backing ListedItem which is used to set the properties on the IStorageItem contract. 

Additionally, a static data field for BasicProperties is created on the VirtualStorageItem class and is set, only initially, from a simple streamed storage file. This makes the present use of VirtualStorageFile (in BaseLayout for drag and drop) more concise. 

Lastly, path instantiation will allow us to leverage VirtualStorageItem for all filesystem operations, instead of having to create a StorageFile/Folder n number of times. This will likely speed up paste for storage item operations like move and copy because we inspect the underlying FileDropList (dealing with filesystem paths) instead of the DataPackage itself as of https://github.com/files-community/Files/pull/9035.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
N/A

### DIFF
--- a/src/Files.Uwp/BaseLayout.cs
+++ b/src/Files.Uwp/BaseLayout.cs
@@ -844,9 +844,8 @@ namespace Files.Uwp
             try
             {
                 // Only support IStorageItem capable paths
-                var itemList = e.Items.OfType<ListedItem>().Where(x => !(x.IsHiddenItem && x.IsLinkItem && x.IsRecycleBinItem && x.IsShortcutItem)).Select(x => new VirtualStorageItem(x));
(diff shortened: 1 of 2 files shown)


gave92 (approved):
Looks good.. unless we need to revert #9014 as per my comment :(

PR #9053, opened 2022-05-01
Service IDateTimeFormatter

cinqmilleans:
Replace DateTimeExtensions by service IDateTimeFormatter.

This pr is a refactor that profoundly changes the structure of date formatting. This pr releases ListedItem from this coupling to prepare its refactoring.

**Resolved / Related Issues**
DateTimeExtensions is a design issue: static class that exposes unused methods, difficult to add/modify a date format because everything is done in the same implementation, use of returnFormat as constructor parameter of many classes that shouldn't fit care about it, public method that returns 4 values in a tuple, read user parameters on each call.

**Details of Changes**
This pr creates a new IDateTimeFormatter service and its 3 user parameter implementations, and an implementation (provided by IServiceProvider) that still behaves like the user format.
DateTimeExtensions is removed.
All uses of returnFormat are removed. The other classes don't have to worry about it.
Use an event to change format.
The initialisation of the setting windows is simpler because of inheritance.

**Validation**
- [ ] Built and ran the app


cinqmilleans:
@BanCrash I made a small style change.

BanCrash (dismissed):
I don't know the opinion of @d2dyno1 , but this LGTM.

d2dyno1 (approved):
looks good

PR #9056, opened 2022-05-01
Manually get folder name in case of network share folder or drive

### LINKED ISSUE #7561: Network drives in bundles
### Description

The name of the folder is not diaplayed if the folder is a network drive, but clicking on the folder is working.

### Steps To Reproduce

Create a bundle
link a network drive as a folder

### Expected behavior

Display name of network drive in the bundle list

### Files Version

2.1.0.0

### Windows Version

Windows 11 21H2

### Relevant Assets/Logs

image

Bayrem00:
**Resolved / Related Issues**
- Closes #7561 ...

**Details of Changes**
Starting from .Net 5 as mentioned [here](https://github.com/dotnet/runtime/issues/48039), `GetFileName` does not get the Share folder name, but gets an empty string instead, which is how the bug arose.
I added specific handling for UNC paths if it stops at the share level (ie : \\127.0.0.1\share), and for Drive folders.

**Validation**
- [ X] Built and ran the app

### DIFF
--- a/src/Files.Uwp/ViewModels/Widgets/Bundles/BundleItemViewModel.cs
+++ b/src/Files.Uwp/ViewModels/Widgets/Bundles/BundleItemViewModel.cs
@@ -45,7 +45,23 @@ namespace Files.Uwp.ViewModels.Widgets.Bundles
         {
             get
             {
-                string fileName = System.IO.Path.GetFileName(this.Path);
+                string fileName;
+
+                // Network Share path
+                if (System.IO.Path.GetPathRoot(this.Path) == this.Path && this.Path.StartsWith(@"\\")) 
+                {
+                    fileName = this.Path.Substring(this.Path.LastIndexOf(@"\") + 1);
+                }
+                // Drive path
+                else if (System.IO.Path.GetPathRoot(this.Path) == this.Path)
+                {
+                    fileName = this.Path;
+                }
+                else
+                {
+                    fileName = System.IO.Path.GetFileName(this.Path);
+                }
+
 
                 if (fileName.EndsWith(".lnk", StringComparison.Ordinal) || fileName.EndsWith(".url", StringComparison.Ordinal))
                 {


src/Files.Uwp/ViewModels/Widgets/Bundles/BundleItemViewModel.cs
```diff
@@ -45,7 +45,23 @@ public string Name
...
-                string fileName = System.IO.Path.GetFileName(this.Path);
+                string fileName;
+
+                // Network Share path
+                if (System.IO.Path.GetPathRoot(this.Path) == this.Path && this.Path.StartsWith(@"\\")) 
+                {
+                    fileName = this.Path.Substring(this.Path.LastIndexOf(@"\") + 1);
+                }
+                // Drive path
+                else if (System.IO.Path.GetPathRoot(this.Path) == this.Path)
```
yaichenbaum, line 56: Do we need this `if` statement or is it already covered by the next one?

d2dyno1 (approved):

src/Files.Uwp/ViewModels/Widgets/Bundles/BundleItemViewModel.cs
```diff
@@ -45,7 +45,23 @@ public string Name
         {
             get
             {
-                string fileName = System.IO.Path.GetFileName(this.Path);
+                string fileName;
+
+                // Network Share path
+                if (System.IO.Path.GetPathRoot(this.Path) == this.Path && this.Path.StartsWith(@"\\"))
```
d2dyno1, line 51: Are you sure? This will check if the path starts with double "\\", not single "\"

Bayrem00: Yep, UNC paths start with ''\\'' , that is what i am checking for there.
If `this.Path` it is a UNC path and has no subfolders.

PR #9062, opened 2022-05-01
Fix copying a zip file not using Shell file operations

gave92:
**Details of Changes**
Add details of changes here.
- Fixes an issue for which file operations involving zip files do not use ShellFileOperations. This is because IsZipPath() returns true also for the zip file itself, not only for items inside.
- Add a `includeRoot` parameter IsZipPath() to exclude the zip file itself from the check

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

Issue #8649, opened 2022-03-13
Recycle bin doesn't ask for confirmation when dragging item to sidebar


Nexxol:
### Description

![image](https://user-images.githubusercontent.com/68960034/158043032-6eabf679-5e07-4eb6-a9e8-142798a1f03a.png)
(these are the folders I'm talking about)

If you go to wherever those system folders are located (mine is at the root of my B: drive) you can drag them into the recycle bin if you have recycle bin pinned to favourites.

If you drag it to the recycle bin it deletes all files inside of it even if you recover it in recycle bin, To recover these files you NEED to use recovery software. I Accidentally dragged my Videos folder into the recycle bin and when I restored it from recycle bin everything was gone.

### Steps To Reproduce

1. Have recycle bin pinned to the sidebar
2. Go to where those system folders are located
3. Drag them into the recycle bin on the sidebar
4. Try recovering them from the recycle bin
5. All files will be gone

### Expected behavior

You should not be able to delete these folders, they are system folders. File explorer does not allow you to and for good reason.
![image](https://user-images.githubusercontent.com/68960034/158043246-3ff3361c-4d7f-4210-9fce-f7dc4ea18e07.png)
(Windows file explorer) 

Below is a picture of me dragging Videos from  B: into Recycle Bin that is pinned on the sidebar. You should not be able to do this.
![image](https://user-images.githubusercontent.com/68960034/158043290-dec25551-b905-4241-8b32-5bb136f6194d.png)‌‌
(this is from Files)​‌‌‌‌​‌​​‌‌​‌​‌​​​​‌​​‌​​‌​‌‌​​‌‌​‌‌‌​‌​​‌​‌‌​‌​​‌‌‌‌​‌​‌​​‌‌

### Files Version

Version: 2.1.15.0 OS Version: 10.0.22572.100

### Windows Version

Windows 11 22H2 (OS Build 22572.100)

### Relevant Assets/Logs

![image](https://user-images.githubusercontent.com/68960034/158043325-d8145fda-2aa7-44f3-88bc-1b514e555e88.png)

yaichenbaum:
@Nexxol this is by design, you can delete these folders in File Explorer as well.

Nexxol:
@yair100  You cant. as shown in my screenshot 
![image](https://user-images.githubusercontent.com/68960034/158044235-d17c3b9b-ea9a-4045-b057-56f66a6fdd64.png)
[...]

PR #9081, opened 2022-05-04
Fix drop to recycle bin

cinqmilleans:
**Resolved / Related Issues**
When the option is activated, the deletion confirmation dialog is not displayed when deleting an element by drag&drop to the Recycle Bin. #8649

**Details of Changes**
The confirmation dialog is displayed (if the option is activated) during a drag&drop to the Recycle Bin.
Works with sidebar and dual panel.

**Validation**
- [ ] Built and ran the app


yaichenbaum (approved):
Thank you

PR #9082, opened 2022-05-04
Refactor EnumerateDrives

cinqmilleans:
Storage class EnumerateThingAsync methods have 2 issues fixed here.

1) The name assumes it returns an enumerable, but it returns nothing. It updates a property. I renamed it UpdateThingAsync for clarity.
2) The method reads the parameters to run or not according to user preferences. It is not the role of storage to handle this, but the role of the front. I moved this role to the caller which removes the dependency on UserSettingsService.

This is to move Storage to Backend, which should not depend on settings.

**Validation**
- [ ] Built and ran the app


BanCrash (approved):
LGTM

PR #9087, opened 2022-05-04
Fix EmptyRecycleBinButton state

cinqmilleans:
**Resolved / Related Issues**
When the RecycleBin is empty and appears at startup, the EmptyRecycleBin button can be active. This pr fixes this bug. This is visible with the "Continue where you left off" option when exiting Files on the RecycleBin tab.

**Details of Changes**
Changes the initial value of HasItem to False. This value is not modified when there is no element, this is the reason for the bug.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before/After
RecycleBin_before
RecycleBin_after

### DIFF
--- a/src/Files.Uwp/ViewModels/ToolbarViewModel.cs
+++ b/src/Files.Uwp/ViewModels/ToolbarViewModel.cs
@@ -1129,7 +1129,7 @@ namespace Files.Uwp.ViewModels
             }
         }
 
-        private bool hasItem = true;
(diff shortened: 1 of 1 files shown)


yaichenbaum:
@cinqmilleans can you verify that these changes build and mark the check box in the pull request description?

cinqmilleans:
It's done.

yaichenbaum (approved):

PR #7824, opened 2022-01-20
Removed extra controls from the Visual Tree

yaichenbaum:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #7820
- Related #issue...

**Details of Changes**
Add details of changes here.
- Added...
- Fixed...

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [ ] Tested the changes for accessibility

**Screenshots (optional)**
Add screenshots here.

PR #9090, opened 2022-05-04
feat(details-view): add option to size all columns to fit

### LINKED ISSUE #8384: Add option to size all columns to fit in the details layout

jiejasonliu:
**Resolved / Related Issues**
- Closes #8384
- Fix regression of #5506

**Details of Changes**
- Add `Size All Columns to Fit` to details menu flyout in detail view header
- Make `ResizeColumnToFit` safe to call programmatically, previously only safe to call when user clicks on visible column
- Improve heuristic to calculate appropriate width since measuring `TextBlock` is expensive
   - Measure longest text _first_ because we can expect that the width should be long enough for shorter texts 
   regardless if they have wider letters or not
   - This also allows the width to not overshoot, imagine if the first files were wide letters: 
   **(e.g. AAA, BBB, DDD)** --> this would skew the average letter width
- Measure `TextBlock` using sample instance to avoid `Arrange` or other metadata to pollute & make desired size incorrect
- Fix regression of "double click `GridSplitter` (divider)" in header to resize column
  - Caused by a change where the majority of `TextBlock` were no longer being nested in `Grid` 
  (#7824)

**Validation**
- [X] Built and ran the app

**Screenshots (optional)**

- feature (size all columns to fit):
Files-8384-menu-item

---

- regression fix (resize on double click divider):
Files-8384-regress-dbclick

### DIFF
--- a/src/Files.Uwp/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/src/Files.Uwp/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -689,6 +689,21 @@ namespace Files.Uwp.Views.LayoutModes
             e.Handled = true;
         }
 
+        private void SizeAllColumnsToFit_Click(object sender, RoutedEventArgs e)
+        {
+            if (!FileList.Items.Any())
+            {
+                return;
+            }
+
+            // for scalability, just count the # of public `ColumnViewModel` properties in ColumnsViewModel
+            int totalColumnCount = ColumnsViewModel.GetType().GetProperties().Count(prop => prop.PropertyType == typeof(ColumnViewModel));
+            for (int columnIndex = 1; columnIndex <= totalColumnCount; columnIndex++)
+            {
+                ResizeColumnToFit(columnIndex);
+            }
+        }
+
         private void ResizeColumnToFit(int columnToResize)
         {
             if (!FileList.Items.Any())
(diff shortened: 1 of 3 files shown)


src/Files.Uwp/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -700,16 +715,24 @@ private void ResizeColumnToFit(int columnToResize)
             {
                 1 => FileList.Items.Cast<ListedItem>().Select(x => x.ItemName?.Length ?? 0).Max(), // file name column
                 2 => FileList.Items.Cast<ListedItem>().Select(x => x.FileTagUI?.TagName?.Length ?? 0).Max(), // file tag column
-                3 => FileList.Items.Cast<RecycleBinItem>().Select(x => x.ItemOriginalPath?.Length ?? 0).Max(), // original path column
-                4 => FileList.Items.Cast<RecycleBinItem>().Select(x => x.ItemDateDeleted?.Length ?? 0).Max(), // date deleted column
+                3 => FileList.Items.Cast<ListedItem>().Select(x => (x as RecycleBinItem)?.ItemOriginalPath?.Length ?? 0).Max(), // original path column
```
gave92, line 718: Is this change just "stylistic"?

gave92 (dismissed):

yaichenbaum (changes requested):

yaichenbaum (approved):

yaichenbaum:
@jiejasonliu thank you!

PR #9098, opened 2022-05-06
Implement folder setting to sort directories and files together and add option to toolbar sort flyout

### LINKED ISSUE #7120: Add option to sort flyout to sort files and folders together
### What's the Problem?

Currently there is an option in the global settings to list and sort directories alongside files. Problem is that i wanna have the flexibility of using this only in specific folders. Having to temporarily change the global setting each time one wanna use it is quite annoying.

### Solution/Idea

Move the setting into the sort menu in the toolbar and have it only apply to the directory in question.

### Alternatives

An even better alternative would be https://github.com/files-community/Files/issues/5834 but it's broader. I'm thinking this new issue covers the most important feature of them all and might be easier to get through.

### Priorities

_No response_

### Files Version

2.0.40.0

### Windows Version

Windows 10 21H2 19044.1348

### Additional comment

_No response_

manfromarce:
**Resolved / Related Issues**
- Closes #7120.

**Details of Changes**
- The global setting to sort directories alongside files is removed and the option is loaded/saved per-folder (like "sort by" and ascending/descending)
- The option is added to the toolbar flyout

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
immagine

### DIFF
--- a/Files.sln
+++ b/Files.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.31911.196
+# Visual Studio Version 17
+VisualStudioVersion = 17.1.32421.90
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Files.FullTrust", "src\Files.FullTrust\Files.FullTrust.csproj", "{533F9E86-EE0A-4FCB-B70C-F29532C1B787}"
 	ProjectSection(ProjectDependencies) = postProject
--- a/Files.sln
+++ b/Files.sln
@@ -1,7 +1,6 @@
-﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.1.32421.90
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31911.196
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Files.FullTrust", "src\Files.FullTrust\Files.FullTrust.csproj", "{533F9E86-EE0A-4FCB-B70C-F29532C1B787}"
 	ProjectSection(ProjectDependencies) = postProject
--- a/src/Files.Uwp/UserControls/InnerNavigationToolbar.xaml
+++ b/src/Files.Uwp/UserControls/InnerNavigationToolbar.xaml
@@ -507,6 +507,7 @@
                         <MenuFlyoutSeparator />
                         <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.IsSortedAscending, Mode=TwoWay}" Text="{helpers:ResourceString Name=Ascending}" />
                         <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.IsSortedDescending, Mode=TwoWay}" Text="{helpers:ResourceString Name=Descending}" />
+                        <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.AreDirectoriesSortedAlongsideFiles, Mode=TwoWay}" Text="{helpers:ResourceString Name=SettingsListAndSortDirectoriesAlongsideFiles}" />
                     </MenuFlyout>
                 </AppBarButton.Flyout>
             </AppBarButton>
--- a/src/Files.Uwp/UserControls/InnerNavigationToolbar.xaml
+++ b/src/Files.Uwp/UserControls/InnerNavigationToolbar.xaml
@@ -507,7 +507,6 @@
                         <MenuFlyoutSeparator />
                         <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.IsSortedAscending, Mode=TwoWay}" Text="{helpers:ResourceString Name=Ascending}" />
                         <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.IsSortedDescending, Mode=TwoWay}" Text="{helpers:ResourceString Name=Descending}" />
-
                          <MenuFlyoutSeparator />
                          <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.AreDirectoriesSortedAlongsideFiles, Mode=TwoWay}" Text="{helpers:ResourceString Name=SettingsListAndSortDirectoriesAlongsideFiles}" />
                     </MenuFlyout>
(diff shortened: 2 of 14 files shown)


Files.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.31911.196
+# Visual Studio Version 17
+VisualStudioVersion = 17.1.32421.90
```
yaichenbaum, line 4: Is this necessary for the other changes?

src/Files.Uwp/Views/SettingsPages/Preferences.xaml
```diff
@@ -195,16 +195,6 @@
...
 
-                        <local:SettingsBlockControl Title="{helpers:ResourceString Name=SettingsListAndSortDirectoriesAlongsideFiles}" HorizontalAlignment="Stretch">
-                            <local:SettingsBlockControl.Icon>
-                                <FontIcon Glyph="&#xE174;" />
-                            </local:SettingsBlockControl.Icon>
-                            <ToggleSwitch
-                                AutomationProperties.Name="{helpers:ResourceString Name=SettingsListAndSortDirectoriesAlongsideFiles}"
-                                IsOn="{Binding ListAndSortDirectoriesAlongsideFiles, Mode=TwoWay}"
-                                Style="{StaticResource RightAlignedToggleSwitchStyle}" />
-                        </local:SettingsBlockControl>
```
yaichenbaum, line 206: It's up for discussion but I think we should keep this setting to configure the default behavior.

src/Files.Uwp/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -507,6 +507,7 @@
                         <MenuFlyoutSeparator />
                         <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.IsSortedAscending, Mode=TwoWay}" Text="{helpers:ResourceString Name=Ascending}" />
                         <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.IsSortedDescending, Mode=TwoWay}" Text="{helpers:ResourceString Name=Descending}" />
+                        <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.AreDirectoriesSortedAlongsideFiles, Mode=TwoWay}" Text="{helpers:ResourceString Name=SettingsListAndSortDirectoriesAlongsideFiles}" />
```
yaichenbaum, line 510:
```suggestion

                         <MenuFlyoutSeparator />
                         <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.AreDirectoriesSortedAlongsideFiles, Mode=TwoWay}" Text="{helpers:ResourceString Name=SettingsListAndSortDirectoriesAlongsideFiles}" />
```

Files.sln
manfromarce: No, it was unintentional

src/Files.Uwp/Views/SettingsPages/Preferences.xaml
yaichenbaum: The idea would be to just set the default value, if the user wants it to apply to all folders, they can adjust the individual folder setting.

manfromarce:
I've restored the global setting and it seems to work, except that already open tabs are not updated when the setting is changed.

yaichenbaum:
> I've restored the global setting and it seems to work, except that already open tabs are not updated when the setting is changed.

Is this somethign you can copy from the option to show/hide hidden items?

manfromarce:
The problem is that for each folder visited if there is no custom setting, the value of the global setting is saved. So when the user changes the global setting again it will have no effect on that folder

yaichenbaum:
> The problem is that for each folder visited if there is no custom setting, the value of the global setting is saved. So when the user changes the global setting again it will have no effect on that folder

@manfromarce that's expected, to have the setting apply on all items you need to turn off the option to use individual preferences.

manfromarce:
Ok, I've done some tests and the combination of global and individual settings seems to work properly

yaichenbaum:
> Ok, I've done some tests and the combination of global and individual settings seems to work properly

Awesome! Do you think the option to set the default value should be disabled when the option to use individual preferences is turned off?

src/Files.Uwp/ViewModels/ItemViewModel.cs
```diff
@@ -355,6 +362,16 @@ public bool IsSortedDescending
             }
         }
 
+        public bool AreDirectoriesSortedAlongsideFiles
+        {
+            get => !IsSortedAscending;
```
gave92, line 367: IsSortedAscending? Maybe folderSettings.SortDirectoriesAlongsideFiles?

gave92:
> > Ok, I've done some tests and the combination of global and individual settings seems to work properly
> 
> Awesome! Do you think the option to set the default value should be disabled when the option to use individual preferences is turned off?

Other options:
- the two options should be linked <-
- the new option in the toolbar should be hidden

src/Files.Uwp/ViewModels/ItemViewModel.cs
manfromarce: Yes, sorry

manfromarce:
It could be that when individual settings are disabled the new toolbar command is synced with the general setting (like show hidden/dot files)

yaichenbaum:
> It could be that when individual settings are disabled the new toolbar command is synced with the general setting (like show hidden/dot files)

That can work, we can change it later if it gets confusing.

manfromarce:
@yair100 @gave92 The settings should now be linked and refreshed correctly when individual settings are disabled, let me know what you think

yaichenbaum:
@manfromarce it's working great! We'll probably have to rearrange some options so that it's easier to see what's going on, but this is fantastic.
Do you want to add a setting to choose the default layout and default sort mode using the same logic?

yair100 (dismissed):
LGTM

src/Files.Uwp/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -507,6 +507,9 @@
                         <MenuFlyoutSeparator />
                         <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.IsSortedAscending, Mode=TwoWay}" Text="{helpers:ResourceString Name=Ascending}" />
                         <ToggleMenuFlyoutItem IsChecked="{x:Bind ViewModel.IsSortedDescending, Mode=TwoWay}" Text="{helpers:ResourceString Name=Descending}" />
+
```
yaichenbaum, line 510:
```suggestion
```

manfromarce:
OK, I think I can implement them.  We could maybe add a new group in settings because "files and folders" would have too many entries

yaichenbaum:
> OK, I think I can implement them. We could maybe add a new group in settings because "files and folders" would have too many entries

We can add a group for sorting preferences

yair100 (approved):

gave92 (approved):

PR #9115, opened 2022-05-07
Allow to set file tags from command line

### LINKED ISSUE #9033: Tagging a file programmatically
### What's the Problem?

Currently, there is no way to tag a file with the new tagging system if not by using the UI. It would be very useful to implement a some sort of way to do that via command line.

### Solution/Idea

For example it can be done through command line or, even better, by reading the already present metadata in windows that it uses  (for example implementing [EXIF](https://exiftool.org/TagNames/EXIF.html) into the whole tag system, but that would be a little out of scope.

The simplest method it'd be to use the command line.
`files -tag "C:\Path\To\File.txt" "Blue"` 

### Alternatives

Currently there aren't really any alternatives to interact programmatically with the tag system in Files, none that I know of at least.

### Files Version

_No response_

### Windows Version

_No response_

### Comments

_No response_

gave92:
**Resolved / Related Issues**
- Closes #9033

**Details of Changes**
Add details of changes here.
- Added "-tag" command line argument to allow setting/removing file tags from the command line
E.g `files /tag:blue "file1" "file2" "file3"` or `files -tag "orange" "file1" "file2"`
- Fixed an issue preventing command line args written as "/command" instead of "-command" from working
- Fixed an issue preventing shell folders from opening in explorer when "AlwaysOpenANewInstance=true"

**Notes**
Issuing the "-tag" command results in a Files window opening (and staying open). Given that there's no way to prevent the splash screen from appearing when invoking Files from the command line, I think it's better to leave the window open instead of causing it to repeatedly flash open and closed.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Backend/Services/Settings/IFileTagsSettingsService.cs
+++ b/src/Files.Backend/Services/Settings/IFileTagsSettingsService.cs
@@ -14,6 +14,8 @@ namespace Files.Backend.Services.Settings
 
         IEnumerable<FileTagViewModel> GetTagsByName(string tagName);
 
+        IEnumerable<FileTagViewModel> SearchTagsByName(string tagName);
+
         object ExportSettings();
 
         bool ImportSettings(object import);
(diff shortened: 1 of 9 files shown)


FukYaMudda (approved):

yaichenbaum:
> Issuing the "-tag" command results in a Files window opening (and staying open). Given that there's no way to prevent the splash screen from appearing when invoking Files from the command line, I think it's better to leave the window open instead of causing it to repeatedly flash open and closed.

This seems to be a fair tradeoff, we can improve this when we move to WinUI 3 in the future.

Issue #9093, opened 2022-05-05
Preview pane issues


z01ks:
### Description

Buttons on the top right (Selection options, Sort, Layout, Preview pane, More don't scale with the window when the Preview Pane is enabled and its size is maxed out to the left. Also, changing Preview Pane size doesn't work at that point. When I make the window bigger to the right so the button group appears on top right, I can again scale the preview pane freely. If the preview pane is big enough I just have to make Files windowed to lose my top right button group and the ability to move the preview pane.

Also I have another question. Is it possible for me to have the latest release on Windows 10? I am unable to update Win because they don't support my processor.

### Steps To Reproduce

1. Make Files Windowed.
2. Enable Preview Pane
3. Move Preview pane left edge all the way to the left.
4. Grab the right edge of the whole window and move it to the left.

### Expected behavior

You should see the button group disappear and at that point the preview pane left edge becomes unmovable.

### Files Version

2.1.24.0

### Windows Version

10.0.19044.1645

### Relevant Assets/Logs

n/a

cinqmilleans:
@z01ks @yair100 I was able to reproduce the bug. I will try to fix it.

z01ks:
Will it be possible to get this update on Windows 10 or is the latest Files
version only available on Win 11?

PR #9126, opened 2022-05-09
Fix preview size

cinqmilleans:
**Resolved / Related Issues**
When the application window is resized and the PreviewPane is opened, the content size shrinks to keep the same size for the PreviewPane. This is the proper behavior.
But if we reduce too much, the content reaches the minimum possible size but the size of the PreviewPane is not reduced. This cannot be seen directly because the content of the PreviewPane adapts to the visible part. But the grid is too big and this can be seen in the InnerToolbar which no longer displays the elements on the right, which are displayed outside the available space. It also causes the PreviewPane to get the wrong size if you close it and reopen it. #9093

**Details of Changes**
Add an event to adapt the size of the PreviewPane when the size of the window is modified and the content has its minimum size.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before/After

https://user-images.githubusercontent.com/46631671/167406957-d6ae5abc-b1c8-4903-a2ba-ab026bf0353e.mp4

https://user-images.githubusercontent.com/46631671/167405735-4f8e33e8-0c38-4499-9a89-c5abdc73800a.mp4


yaichenbaum:
@cinqmilleans before merging this, it looks like the issue is already resolved in the `main` branch. The visual state is broken though so we should still take that fix.

cinqmilleans:
@yair100 I'm not sure I understand your comment. This bug is not yet fixed in the main branch.

yaichenbaum:
@cinqmilleans I can't reproduce it locally, can you make sure you have the latest commit?

cinqmilleans:
@yair100 I just tested on the most up-to-date version of main, and there is the bug. To reproduce it, follow the instructions given here #9093. It takes a wide window, enlarge the PreviewPane almost to the maximum and reduce the window. You can see it in my "before" video at 11s.

yair100:
> @yair100 I just tested on the most up-to-date version of main, and there is the bug. To reproduce it, follow the instructions given here #9093. It takes a wide window, enlarge the PreviewPane almost to the maximum and reduce the window. You can see it in my "before" video at 11s.

I still can't reproduce the issue, looking at the screen recording, it appears that you're testing an older version.

cinqmilleans:
The video was made with the sideload version. The bug is still present in the latest version of main, as I just checked (commit 3df71f1f717ab67840bf10a04ef60cc3264caeab).

yaichenbaum:
> The video was made with the sideload version. The bug is still present in the latest version of main, as I just checked (commit [3df71f1](https://github.com/files-community/Files/commit/3df71f1f717ab67840bf10a04ef60cc3264caeab)).

Can you upload a screen recording of the version in `main` so I can compare the results?

cinqmilleans:
@yair100 Here.

https://user-images.githubusercontent.com/46631671/167614598-e980474e-e543-40fd-add1-1aa67d95daf2.mp4

yaichenbaum:
@cinqmilleans I see the issue now, I thought you were talking about the column sizes changing in the details pane. I think this can all be solved by adjusting the xaml controls so we're going to hold off on this until after v2.3.

cinqmilleans:
I fixed the merge, which was broken. I retested to validate the bug fix. This pr was pending release of version 2.3.

yaichenbaum:
Instead of handling this in codebehind, is this something we can handle with a visual state?

src/Files.Uwp/Views/MainPage.xaml.cs
```diff
@@ -383,6 +383,21 @@ private void Page_Loaded(object sender, RoutedEventArgs e)
             }
         }
 
+        private void Page_SizeChanged(object sender, SizeChangedEventArgs e)
+        {
+            switch (Pane?.Position)
+            {
+                case PanePositions.Right when ContentColumn.ActualWidth == ContentColumn.MinWidth:
+                    UserSettingsService.PaneSettingsService.VerticalSizePx += e.NewSize.Width - e.PreviousSize.Width;
```
d2dyno1, line 391: When you change the value of a setting, the whole cache is serialized. This will cause massive performance issues as well as storage strain

d2dyno1 (changes requested):

d2dyno1 (approved):
looks good

yaichenbaum:
> looks good

@d2dyno1 do you think we should use visual states here?

d2dyno1:
I guess using visual states there would only make things harder

yaichenbaum:
> I guess using visual states there would only make things harder

In what way? I find it easier when everything is done in xaml, it helps prevent surprises.

d2dyno1:
@yair100 Settings are saved in the code behind if you take a look, won't be possible with xaml only iirc

yair100 (approved):

PR #9146, opened 2022-05-11
Updated FR-fr

cinqmilleans:
Translates new texts into French. There are few texts, but the goal is to have complete coverage for v2.3.


yaichenbaum (approved):
lgtm

PR #9150, opened 2022-05-11
Rotate multiple images

puppetsw:
**Resolved / Related Issues**
- Closes ...

**Details of Changes**
Add details of changes here.
- Added ability to select multiple images and rotate left/right.
- Added ability to select multiple images and set as desktop background (create slideshow) Currently only works on primary monitor.

**Validation**
- [x] Built and ran the app

**Demo**
demo


puppetsw:
@yair100 would it be good to change the "set as background" text on the toolbar to "set as slideshow" when multiple images are selected?

yair100:
> @yair100 would it be good to change the "set as background" text on the toolbar to "set as slideshow" when multiple images are selected?

Either that or we can display a different button altogether.

yaichenbaum:
@puppetsw is this ready for review?

yaichenbaum:
What's the behavior if you have multiple file types selected in addition to the image?

puppetsw:
> What's the behavior if you have multiple file types selected in addition to the image?

The button should remain hidden.

yair100 (approved):

Issue #9159, opened 2022-05-13
The app only copies first 20 files from my android phone's DCIM folder


navedrasul:
### Description

I lost a lot of media because your app only copied first 20 files from my android phone's DCIM folder. I pressed Ctrl+a to select all of the files in the source folder and then pressed Ctrl+v in the destination folder. I thought that all of the files have been copied but it only copied first 20 files. I deleted the files in the source folder to free up the space. What would you say and do over this? 😐

### Steps To Reproduce

1. Go to a folder with more than 100 photos and videos files. Call this folder the source folder.
2. Press Ctrl+a to select all of the files in the source folder.
3. Go to another empty folder. Call this the destination folder.
4. Press Ctrl+v in the destination folder
5. Count the number of files copied to the destination folder.

### Expected behavior

All files should have been copied from source to the destination folder

### Files Version

2 

### Windows Version

Windows 11

### Relevant Assets/Logs

None

Josh-65:
Plase update the file version and windows version using Files settings > about > copy on the top item

PR #9162, opened 2022-05-13
Show failed banner if any items failed in a file operation

gave92:
**Resolved / Related Issues**
- Related #9159

**Details of Changes**
Add details of changes here.
- When using StorageAPI in file operations involving multiple items the last copy result overwrites the status shown in the banner. So if 29/30 operations fail but the 30th succeeds we show the "Success!" green banner, which is wrong.
- Changed callback function to only update the status if nothing has failed so far. 
- Remove the [Flags] attribute from enum whose values are not powers of 2 as required

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Shared/Enums/FileOperationType.cs
+++ b/src/Files.Shared/Enums/FileOperationType.cs
@@ -5,7 +5,6 @@ namespace Files.Shared.Enums
     /// <summary>
     /// Type of operation on Files filesystem that took place
     /// </summary>
-    [Flags]
     public enum FileOperationType : byte
     {
         /// <summary>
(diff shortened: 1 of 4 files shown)


yaichenbaum (approved):
LGTM

yaichenbaum:
@gave92 would you be intrested in documenting the code for file operations? It's pretty complex and the tinniest change can break things 😅

gave92:
I could try =P I wonder if we could add automated checks for file operations.

gave92:
> It's pretty complex and the tinniest change can break things

I totally agree. Perhaps we should require reviews/testing from multiple people before merging changes to file operations (I see you approved this pretty fast, now I'm provenly infallible, but still).

yaichenbaum:
We deffinitly can.

PR #9165, opened 2022-05-14
[Post 2.3] Show Google drive under Cloud section

### LINKED ISSUE #6321: Google drive for desktop not showing under cloud drives
### Description

image

### Steps To Reproduce

_No response_

### Expected behavior

It should come under Cloud drive section

### Files Version

Version: 2.0.16.0

### Windows Version

OS Version: 10.0.19042.1165

### Relevant Assets

_No response_

gave92:
**Resolved / Related Issues**
- Closes #6321

**Details of Changes**
Add details of changes here.
- Adds Google virtual drive to the Cloud sidebar section.

**Notes**
This adds google drive to the cloud section _but still shows it also under Drives section_.
The entry under cloud section is shown with the google drive icon .

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.Shared/CloudProvider.cs
+++ b/src/Files.Shared/CloudProvider.cs
@@ -10,6 +10,8 @@ namespace Files.Shared
 
         public string SyncFolder { get; set; }
 
+        public byte[] IconData { get; set; }
+
         public override int GetHashCode()
         {
             return $"{ID}|{SyncFolder}".GetHashCode();
--- a/src/Files.Uwp/Extensions/ImageSourceExtensions.cs
+++ b/src/Files.Uwp/Extensions/ImageSourceExtensions.cs
@@ -1,6 +1,9 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
+using System.Runtime.InteropServices.WindowsRuntime;
 using System.Threading;
 using System.Threading.Tasks;
+using Windows.Storage;
 using Windows.Storage.Streams;
 
 namespace Files.Uwp.Extensions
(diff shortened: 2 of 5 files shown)


DeveloperWOW64:
Don't you want to remove the icon for GD in the Drives section?

gave92:
I wouldn't remove it, no, but I'm open to feedback on this. While I agree it can be strange/wasteful to see it in two places:
- users may have hidden the cloud drive section and wouldn't find GD anymore
- I'm personally used to find it under Drives in explorer and on the this PC/ Files homepage
- it's annoying to code: we'd either need to delay showing the drives until we've detected cloud drives or the GD would have to be moved from one section to the other after detection

Lamparter:
> I wouldn't remove it, no, but I'm open to feedback on this. While I agree it can be strange/wasteful to see it in two places my concerns are:
> 
> * users may have hidden the cloud drive section and wouldn't find GD anymore after update
> * I'm personally used to find it under Drives in explorer and on the this PC/ Files homepage
> * it's annoying to code: we'd either need to delay showing the drives until we've detected cloud drives or the GD would have to be moved from one section to the other after detection

That's true - and any way to remove these minor concerns simply results in difficult code; I see your point.
What does @yair100 think?

yaichenbaum:
The downside to having it in both places is that users would think it's a bug. On the other hand, we don't want to add custom logic for Google Drive and delay the startup time. My opinion is that we should leave Google Drive as is and continue pointing users to Google regarding its choice to act as a drive instead of a cloud drive.

gave92:
I believe we have an approved feature request for the ability to hide/unpin drives and cloud drives. That would fix the duplication issue.
I personally don't find the double entry to be so strange, as by default in explorer Google drive will appear both under Drives and under Quick access sections. The different icon should help as well.
In any case, I'm not hung on this change :)

yaichenbaum:
> I personally don't find the double entry to be so strange, as by default in explorer Google drive will appear both under Drives and under Quick access sections. The different icon should help as well.

I wasn't aware it worked this way, should be fine in this case.

yair100 (approved):

DeveloperWOW64:
> I think it's still happening on Version: 2.3.0.0 (downloaded from Microsoft Store), my OS Version: 10.0.19044.1766 image

The fix is still rolling out to preview, not stable.

Issue #9147, opened 2022-05-11
Move the home item out of the favorites section


yaichenbaum:
### What's the Problem?

The home item isn't a regular favorite, but we put it under the favorites section to save space. This is no longer applicable since recent changes reduced the height of items in the sidebar.

### Solution/Idea

Move the home item out of the favorites section in the sidebar.

### Alternatives

Leave it as is

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will move the home item out of the favorites section Must |
| This proposal will allow users to hide the home item | Should |

### Files Version

v2.2

### Windows Version

Windows 11

cinqmilleans:
I create the pr.

esibruti:
@yair100 pr #9176 closes this issue

yair100:
> @yair100 pr #9176 closes this issue

#9176 completed some of the work for this issue but the second half still needs to be done.

PR #9166, opened 2022-05-14
Move the home button to the address bar

cinqmilleans:
Moves the Home button to the navigation bar, which is its most logical place. The MiddleClick opens a new tab. For this reason, this button is always active, even on the home page. #9147

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Home

### DIFF
--- a/src/Files.Uwp/DataModels/SidebarPinnedModel.cs
+++ b/src/Files.Uwp/DataModels/SidebarPinnedModel.cs
@@ -1,11 +1,11 @@
-﻿using Files.Shared.Extensions;
+﻿using CommunityToolkit.Mvvm.DependencyInjection;
+using Files.Backend.Services.Settings;
+using Files.Shared.Extensions;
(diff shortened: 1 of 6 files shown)


jiejasonliu:
can you verify that favorites reordering still works (Move one up/down, Move to top/bottom)?

last time I worked on this, `SidebarPinnedModel::MoveItem` assumes Home is there (I think) which may require this line to change:
`if (oldIndex >= 1 && newIndex >= 1 && newIndex <= FavoriteItems.Count)`

cinqmilleans:
There is indeed a bug, thank you. I will correct it.

yaichenbaum:
@cinqmilleans thank you for the PR, I should have specified in the issue that it should still be in the sidebar. The request is to move it on top of the favorites section.

cinqmilleans:
@yair100 No worries. I will create a new pr. I close this one.

PR #9188, opened 2022-05-16
Add support for Synology Drive Client

### LINKED ISSUE #8084: Synology Drive Client
### What's the Problem?

I use the Synology Drive Client to access my works NAS server. In the default windows file explorer this appears as an additional drive (much like DropBox or Google Drive) that can be easily accessed. However this does not appear in the Files explorer.

### Solution/Idea

Integrate Synology Drive Client to be included in the 'Cloud Drives'

### Alternatives

Currently I have to open the native file explorer or access it via my toolbar.

### Priorities

| Capability | Priority |
| :---------- | :------- |

### Files Version

2.1.5.0

### Windows Version

Windows 11 Home 22000.438

### Comments

_No response_

BanCrash:
**Resolved / Related Issues**
- Closes #8084

**Details of Changes**
- Added support for Synology Drive Client, based on GoogleCloud provider.

NOTE: I never used Synology Drive Client for sync before, just for backups, so while in my tests it's working I can't assure it will work for all scenarios since I had to figure out how the information is stored on the database, but maybe I figured it out wrong and it needs some readjustments. When this is added I would like to receive feedback from the users to check if it's working for them too.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

image


src/Files.Uwp/Filesystem/Cloud/Providers/SynologyDriveCloudProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+                    // Open the connection and execute the command
+                    con.Open();
+                    var reader = cmd.ExecuteReader();
+                    var results = new List<CloudProvider>();
+
+                    while (reader.Read())
+                    {
+                        // Extract the data from the reader
+                        var isSyncTask = reader["conn_id"]?.ToString() == "2";
+                        if (isSyncTask)
```
BanCrash, line 43: I set the value to a variable first and then check if it's true because I'm not sure if this conn_id just have 2 values or could have more. If other users confirm this is working this can be simplified with

 `if (reader["conn_id"]?.ToString() == "2")`

gave92:
Hi, I don't have the possibility to test this, but out of curiosity could you check if in registry under `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Desktop\NameSpace\" there is any subkey related to Synology folders?

BanCrash:
> Hi, I don't have the possibility to test this, but out of curiosity could you check if in registry under `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Desktop\NameSpace\` there is any subkey related to Synology folders?

There is one, but just for showing the name of the server, nothing about the folders:

image

Feel free to ask any other thing you think of.

gave92:
Thanks, I'm asking because that's where Cloud drives that integrate with Explorer register themselves (and indeed you've found Synology).
If you were to take the key name that has the reference to Synology (by key name I mean the "folder" name, below highlighted for Onedrive) and search under `HKEY_CLASSES_ROOT\CLSID\` you should find the same GUID with info about the synced folder (path would be `HKEY_CLASSES_ROOT\CLSID\<SYNOLOGY_GUID>\Instance\InitPropertyBag`).

image

This is what we do to detect a few cloud drives (OneDrive, Box, iCloudDrive, Dropbox, Nextcloud, MEGA, Amazon Drive).
*Note: I have nothing against parsing the database, I'm just curious if the above would work for synology too.

BanCrash:
> Thanks, I'm asking because that's where Cloud drives that integrate with Explorer register themselves (and indeed you've found Synology). If you were to take the key name that has the reference to Synology (by key name I mean the "folder" name, below highlighted for Onedrive) and search under `HKEY_CLASSES_ROOT\CLSID\` you should find the same GUID with info about the synced folder (path would be `HKEY_CLASSES_ROOT\CLSID\{<SYNOLOGY_GUID>}\Instance\InitPropertyBag`).
> 
> 
> 
> This is what we do to detect a few cloud drives (OneDrive, Box, iCloudDrive, Dropbox, Nextcloud, MEGA, Amazon Drive). *Note: I have nothing against parsing the database, I'm just curious if the above would work for synology too.

It seems that it only has a shortcut to a folder that have inside the last folder added, but that seems normal since in Windows file explorer for some reason it only shows the last created sync task:

image

image

However, as I suspected conn_id could have multiple values, now my new sync tasks have a 3, so they are not getting detected. I will do more tests and check the database to see if I can found how to know if a task is a backup task or a sync task.

BanCrash:
Ok, now it should work in each database. I got the connection id for other database, and in that database there is connection_type for knowing if a connection is for sync or for backup. 

I also added the host name to the drive folder as Windows does.

BTW, Files shows all synced folders while Windows only shows the last one:

image

image

gave92:
@BanCrash made a couple of small changes (replaced SynologyDriveConnection with a Tuple, skip invalid entries instead of returning no results). Hopefully I haven't broken anything. If you agree with the changes, this looks good.

gave92 (approved):

BanCrash (commented):
LGTM. Thank you! Just tested and still working as expected.

PR #9194, opened 2022-05-16
Hoist TaskCompletionSource outside the loop

hez2010:
To avoid unnecessary allocations.


yaichenbaum (approved):
This seems so obvious 😂

PR #9176, opened 2022-05-15
Move the Home item outside the favorites section

cinqmilleans:
Move the Home item outside the favorites section.
Had to fix the commands to move Favorites, as it assumed the first item was Home.

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [x] Tested the changes for accessibility

**Screenshots (optional)**
Home

PR #9206, opened 2022-05-17
Revert "Move the Home item outside the favorites section"

yair100:
Reverts files-community/Files#9176

Reverting this to be implemented after v2.3


cinqmilleans:
@yair100 It seems that the bug came from corrupted data during my development tests. My favorites contained the same paths several times. This is a case that shouldn't happen in production and therefore the bug shouldn't happen.

Can you confirm that the bug is not present for you?

If necessary, we can add code that eliminates possible duplicates at the opening of the session, but this case should not happen.

yaichenbaum:
@cinqmilleans it looks good on my end.

PR #9226, opened 2022-05-21
[Post 2.3] struct VolumeInfo

cinqmilleans:
**Resolved / Related Issues**
The VolumeId uniquely identifies a drive. The current code does not give us access to this information.

This information is necessary to be able to keep the sizes of the folders between sessions because the disks may have changed paths (especially removable disks).

This can also be useful information to display in disk properties (to be discussed).

**Details of Changes**
Add the VolumeInfo structure and its asynchronous factory which uses FullTrust code to retrieve the information.

This VolumeInfo is not yet used but will be used later to store folder sizes between sessions.

**Validation**
- [x] Built and ran the app


gave92:
The GetVolumeInformation win32 function is available in UWP. Could avoid having to call the fulltrust process for this.

cinqmilleans:
@gave92 Did you manage to use it? I tried several months ago. We can extract several pieces of information, but the volumeid is always empty. I had found information saying that you needed FullTrust rights to access this information. I just stayed and it returns me an empty string.

gave92:
Ah no I've havent tested, i was just trusting the documentation ;) I can give it a go and try

gave92:
OK, last attemp to avoid having to use the fulltrsut process for this 😆, does the following code retrieve the correct info?
```cs
var folder = await StorageFolder.GetFolderFromPathAsync(@"C:\Users\Marco Gavelli\source\repos\Files");
var extra = await folder.Properties.RetrievePropertiesAsync(new string[] { "System.FileFRN", "System.VolumeId" });
Debug.WriteLine($"File ID: {(ulong?)extra["System.FileFRN"]}");
Debug.WriteLine($"Volume ID: {(Guid)extra["System.VolumeId"]}");
```

cinqmilleans:
Sure enough, I was using the root. I managed to get the guid with your method. It also works for StorageFile. However, it has 2 problems:
1) Drive cannot be empty.
2) It does not work with removable drives. I couldn't get it to work for a folder in my usb drives. I get this guid with the FullTrust.

For performance, I can modify the factory to only use FullTrust when needed.

cinqmilleans:
@gave92 There are now 3 factories: FullTrustVolumeInfoFactory, StorageVolumeInfoFactory and VolumeInfoFactory. The latter uses the Storage method, but if it fails (for example if the drive is empty), it uses the FullTrust method.

gave92:
Ouch my suggestion was in the hope to reduce code not increase it xD.
Given the two limitations you listed above and the fact that StorageFile api is not exactly "fast" anyway we might just stick with Fulltrust.

src/Files.FullTrust/Helpers/DriveHelpers.cs
```diff
@@ -11,10 +11,10 @@ public static string GetVolumeId(string driveName)
             string name = driveName.ToUpperInvariant();
             string query = $"SELECT DeviceID FROM Win32_Volume WHERE DriveLetter = '{name}'";
 
-            using var searcher = new ManagementObjectSearcher(query);
-            foreach (ManagementObject item in searcher.Get()) // max 1 result because DriveLetter is unique.
+            using var cimSession = CimSession.Create(null);
```
gave92, line 14: Changed to use MMI which we already use, instead of adding reference to System.Management.

gave92 (approved):

cinqmilleans:
I approve.

PR #9228, opened 2022-05-21
Fix up button

### LINKED ISSUE #9227: The up button does not use the correct path for drive roots
### Description

Drive roots use a path such as "C:\\". This is the path that is used when opening a drive from the sidebar or other. But when the Up button needs to return to C:\\, it actually calls C:, it's that path in the address bar.

This leads to not finding the drive, for example when we want to display the properties of the disk.

### Steps To Reproduce

1. Display a subfolder of C:\\.
2. Select a view other than the column view.
3. Click on the up button.
4. Click on the address bar. It shows C: and not C:\\.
5. Open the context menu by clicking on an empty area.
6. Click Properties.
7. The popup does not appear because it does not match any Drive.Path.

### Expected behavior

The address bar path is C:\\.
The Properties popup opens.

### Files Version

last commit

### Windows Version

Windows 11

### Relevant Assets/Logs

I have already fixed this bug and I will post the pr.

cinqmilleans:
**Resolved / Related Issues**
Drive roots use a path such as "C:\\". This is the path that is used when opening a drive from the sidebar or other. But when the Up button needs to return to C:\\, it actually calls C:, it's that path in the address bar.

This leads to not finding the drive, for example when we want to display the properties of the disk.
Closes #9227

**Details of Changes**
This pr modifies the Up_Click method to add the trailing "\\" when the path ends with ":". This character is forbidden in paths, except to indicate the drive.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Uwp/Views/ModernShellPage.xaml.cs
+++ b/src/Files.Uwp/Views/ModernShellPage.xaml.cs
@@ -404,7 +404,7 @@ namespace Files.Uwp.Views
         {
             FilesystemViewModel?.UpdateSortOptionStatus();
         }
-        
+
         private void AppSettings_SortDirectoriesAlongsideFilesPreferenceUpdated(object sender, bool e)
         {
             FilesystemViewModel?.UpdateSortDirectoriesAlongsideFiles();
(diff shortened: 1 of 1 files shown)


gave92 (approved):

lukeblevins (approved):
LGTM

PR #9248, opened 2022-05-24
Set downloads folder to group by date

### LINKED ISSUE #6586: Set downloads folder to group by date
### What's the Problem?

Till now i was using the default grouping which is nothing but I just found out that you can group by date created/modified which is so useful and would really be nice to make it default like in regular file explorer for the downloads folder.

### Solution/Idea

Just make group by date modified default.

### Alternatives

Just make group by date modified default.

### Priorities

_No response_

### Files Version

v2.0.28.0

### Windows Version

Windows 11 21H2 22000.282

### Additional comment

_No response_

esibruti:
**Resolved / Related Issues**
- Closes #6586 

**Details of Changes**
Add details of changes here.
- With this code, the Downloads folder will be grouped by date modified on the first start.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
+++ b/src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
@@ -436,7 +436,10 @@ namespace Files.Uwp.ViewModels
                 ApplicationDataCompositeValue adcv = (ApplicationDataCompositeValue)dataContainer.Values[folderPath];
                 return LayoutPreferences.FromCompositeValue(adcv);
             }
-            else if (folderPath.Replace('_', '\\') == CommonPaths.DownloadsPath)
+
+            folderPath.Replace('_', '\\');
+
+            if (folderPath == CommonPaths.DownloadsPath)
             {
                 // Default for downloads folder is to group by date created
                 return new LayoutPreferences
--- a/src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
+++ b/src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
@@ -436,7 +436,7 @@ namespace Files.Uwp.ViewModels
                 ApplicationDataCompositeValue adcv = (ApplicationDataCompositeValue)dataContainer.Values[folderPath];
                 return LayoutPreferences.FromCompositeValue(adcv);
             }
-            else if (folderPath == CommonPaths.DownloadsPath)
+            else if (folderPath.Replace('_', '\\') == CommonPaths.DownloadsPath)
             {
                 // Default for downloads folder is to group by date created
                 return new LayoutPreferences
--- a/src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
+++ b/src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
@@ -352,7 +352,9 @@ namespace Files.Uwp.ViewModels
             if (userSettingsService.PreferencesSettingsService.AreLayoutPreferencesPerFolder)
             {
                 var layoutPrefs = ReadLayoutPreferencesFromAds(folderPath.TrimEnd('\\'));
-                return layoutPrefs ?? ReadLayoutPreferencesFromSettings(folderPath.TrimEnd('\\').Replace('\\', '_'));
+                return layoutPrefs ??
+                    ReadLayoutPreferencesFromSettings(folderPath.TrimEnd('\\').Replace('\\', '_')) ??
+                    GetDefaultPreferences(folderPath.TrimEnd('\\'));
             }
 
             return LayoutPreferences.DefaultLayoutPreferences;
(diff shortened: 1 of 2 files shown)


gave92:
We already have this logic in FolderSettingsViewModel 🤔
```cs
 
 ​            ​else​ ​if​ (​folderPath​ ​==​ ​CommonPaths​.​DownloadsPath​) 
 ​            { 
 ​                ​//​ Default for downloads folder is to group by date created 
 ​                ​return​ ​new​ ​LayoutPreferences 
 ​                { 
 ​                    ​...
 ​                    ​DirectoryGroupOption​ ​=​ ​GroupOption​.​DateCreated​, 
 ​                }; 
 ​            }
```
Doing it in ItemsViewModel would stop the user to change it to something else.

esibruti:
@gave92 I hadn't even noticed it 😅
So but that means that the issue was already resolved before... Now I'm confused.

gave92:
Lol confused as well as to why that issue exists 😆 Could be there's an issue preventing it to work (though it does work for me). Problem is you only notice it if you've never _ever_ changed the layout/sorting in Downloads folder. Reinstalling Files does not reset it so it's not easy to test.

gave92:
Aaah found the issue, indeed that logic does not work. That's because we're passing a wrong path to be compared with the downloads folder path: `ReadLayoutPreferencesFromSettings(folderPath.TrimEnd('\\').Replace('\\', '_')` -> We're triyng to compare `C:_Users_USER_Downloads` to `C:\Users\USER\Downloads`.

esibruti:
> Aaah found the issue, indeed that logic does not work. That's because we're passing a wrong path to be compared with the downloads folder path: `ReadLayoutPreferencesFromSettings(folderPath.TrimEnd('\\').Replace('\\', '_')` Ln355 -> We're triyng to compare `C:_Users_USER_Downloads` to `C:\Users\USER\Downloads`.

So the solution is to remove this replace or replace with '/'

gave92:
So changing "\\" with "_" is done for a reason, that is LocalSettings throw error if you try to use a key/setting name that contains "\\". We need to separate "ReadLayoutPreferencesFromSettings" which reads from LocalSettings and should use the paths with "_" from getting the default layout part that should use the "normal" path.

esibruti:
But that means on line 355 it should be comparing with the '\\' instead of the '_', right?

esibruti:
@gave92 When you can, do the review, I think I found the solution.

src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
```diff
@@ -436,7 +436,7 @@ private static LayoutPreferences ReadLayoutPreferencesFromSettings(string folder
                 ApplicationDataCompositeValue adcv = (ApplicationDataCompositeValue)dataContainer.Values[folderPath];
                 return LayoutPreferences.FromCompositeValue(adcv);
             }
-            else if (folderPath == CommonPaths.DownloadsPath)
+            else if (folderPath.Replace('_', '\\') == CommonPaths.DownloadsPath)
```
gave92, line 439: Ok lol this "works" but will fail if the original path does contain a "_", also should be done on line 453. My suggestion would be to spit this method into two. One that reads from settings and returns null if layout is not found [426-438] and a second that returns the default layout give the path [438-470].

esibruti: Two separate methods? so I can do if instead of else if on line 439. But before that line, since I already used the folderPath with '_', I would already replace and after line 439 down it would be '\\'.

gave92: Sorry something like (writing on the fly/untested):
```cs
public static LayoutPreferences GetLayoutPreferencesForPath(string folderPath)
        {
            IUserSettingsService userSettingsService = Ioc.Default.GetService<IUserSettingsService>();
            if (userSettingsService.PreferencesSettingsService.AreLayoutPreferencesPerFolder)
            {
                var layoutPrefs = ReadLayoutPreferencesFromAds(folderPath.TrimEnd('\\'));
                return layoutPrefs ?? 
                    ReadLayoutPreferencesFromSettings(folderPath.TrimEnd('\\').Replace('\\', '_')) ??
                    GetDefaultPreferences(folderPath.TrimEnd('\\')) ; <---- new method
            }

            return LayoutPreferences.DefaultLayoutPreferences;
        }

private static LayoutPreferences ReadLayoutPreferencesFromSettings(string folderPath)
        {
            if (string.IsNullOrEmpty(folderPath))
            {
                return LayoutPreferences.DefaultLayoutPreferences;
            }

            IUserSettingsService userSettingsService = Ioc.Default.GetService<IUserSettingsService>();
            ApplicationDataContainer dataContainer = localSettings.CreateContainer("LayoutModeContainer", ApplicationDataCreateDisposition.Always);
            folderPath = new string(folderPath.TakeLast(254).ToArray());
            if (dataContainer.Values.ContainsKey(folderPath))
            {
                ApplicationDataCompositeValue adcv = (ApplicationDataCompositeValue)dataContainer.Values[folderPath];
                return LayoutPreferences.FromCompositeValue(adcv);
            }

            return null;
}

private static LayoutPreferences GetDefaultPreferences(string folderPath) <--
{
            if (folderPath) == CommonPaths.DownloadsPath)
            {
                // Default for downloads folder is to group by date created
                return new LayoutPreferences
                {
                    ...
                };
            }
            else if (LibraryHelper.IsLibraryPath(folderPath))
            {
                // Default for libraries is to group by folder path
                return new LayoutPreferences
                {
                    .....
                };
            }
            else
            {
                return LayoutPreferences.DefaultLayoutPreferences; // Either global setting or smart guess
            }
        }
```

esibruti: It's done in the last commit

PR #9276, opened 2022-05-26
Download update in background

puppetsw:
**Resolved / Related Issues**
- Closes #issue...

**Details of Changes**
Add details of changes here.
- Changed update service to download available update in the background before showing the Update button. When the update has finished downloading the button shows to install the update.
- MSIX bundle is downloaded to the application folder.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Backend/Services/IUpdateService.cs
+++ b/src/Files.Backend/Services/IUpdateService.cs
@@ -15,7 +15,7 @@ namespace Files.Backend.Services
         /// </summary>
         bool IsUpdating { get; }
 
-        uint DownloadPercentage { get; }
+        int DownloadPercentage { get; }
 
         Task DownloadUpdates();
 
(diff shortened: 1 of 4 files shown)


puppetsw:
Would it be better to use Path.GetTempPath for the download location instead of the app folder?

yaichenbaum:
@puppetsw does this work for the store version as well?

puppetsw:
> @puppetsw does this work for the store version as well?

No, only the sideload version at the moment.

yaichenbaum:
> MSIX bundle is downloaded to the application folder.

How did this work before?

puppetsw:
> > MSIX bundle is downloaded to the application folder.
> 
> How did this work before?

The package manager downloaded it to whereever it does internally using the appinstaller. This PR takes that away and we handle it manually, so we can do it in the background.

yaichenbaum:
What happens if the app is closed before the download finishes?

puppetsw:
> What happens if the app is closed before the download finishes?

At the moment it is just terminated and starts again (not resume) at startup.

yaichenbaum (approved):

PR #9288, opened 2022-05-29
Refactor service ISizeProvider to Backend

cinqmilleans:
The service IFolderSizeProvider is refactored and moved to backend.

- Rename interface and eventargs for genericity.
- Use ulong instead long as all sizes in the backend.
- Use enum instead negative value for no value.
- Divides the implementation into multiple implementations to separate responsibilities (User parameter management, separation by drives, cache, and soon persistent storage).
- Uses one instance per drive for performance. Each disk has its own ConcurrentDictionary, so a disk with few files is not hampered by the many files in C:.
- Move NativeFindStorageItemHelper to backend because the service uses it.

Thanks to @gave92 for fixing and improving this code. All of these improvements are retained.


gave92:
This looks good. Could you just clarify the need for DrivesSizeProvider? The time for retrieving a value from a Dictionary does not depends on the size of the Dictionary.
> The Dictionary(Of TKey, TValue) generic class provides a mapping from a set of keys to a set of values. Each addition to the dictionary consists of a value and its associated key. Retrieving a value by using its key is very fast, close to O(1), because the Dictionary(Of TKey, TValue) class is implemented as a hash table.

cinqmilleans:
@gave92 Regarding the speed of reading a dictionary, I don't know if the size decreases the performance. I guess so, but let's say no. Another reason is to be able to use a different implementation depending on the drive.

I am currently working on another implementation of ISizeProvider which stores data persistently in an sqlite database. I was able to get a unique guid (recent commit) but there are cases where it doesn't work, e.g. network disk in linux. A solution to this problem may exist.

My idea is to use for each drive one of the 2 implementations (persistent or not). A single instance can be used for each of the 2 implementations. Soon DrivesSizeProvider will use a service to get the correct implementation for each drive.

gave92 (approved):

gave92:
Sounds good, for the database you may look to look at LiteDB which is how file tags are stored instead of SQLite.

cinqmilleans:
@gave92 I fixed a problem in this already approved pr. The cache could be cleared unnecessarily.

I try LiteDB.

gave92 (approved):

d2dyno1:
Hmm

PR #9299, opened 2022-05-30
After pasting, focus all pasted files/folders instead of just one

IgorKordiukiewicz:
**Resolved / Related Issues**
Fixes #8333 

**Details of Changes**

**Validation**
How did you test these changes?
- [x] Built and ran the app

Before:

https://user-images.githubusercontent.com/25117425/171063890-7f921d9c-7468-466e-80b1-3b69584bf49e.mp4

After:

https://user-images.githubusercontent.com/25117425/171063899-223eb360-da14-43b5-9dd4-38164e631681.mp4

PR #9304, opened 2022-05-31
Revert "After pasting, focus all pasted files/folders instead of just one"

yaichenbaum:
Reverts files-community/Files#9299


yaichenbaum:
FYI @IgorKordiukiewicz

IgorKordiukiewicz:
Yeah, I tried to fix it but since the files changes are processed separately, I don't think there is a way to reset `itemsAdded` properly.

jiejasonliu:
> I don't think there is a way to reset `itemsAdded` properly.

Yeah, unfortunately that's exactly why I never implemented this when I implemented selecting singular files.

If interested, there's the potential solution of keeping track of the items that are pasted before it's flattened into separate requests, then you can check if those items are a subset of `itemsAdded` and then select the items and reset the list then.

PR #9372, opened 2022-06-09
Change launch application from properties

### LINKED ISSUE #1581: Add support for changing which program a file opens in from the properties window
Show which application opens the file, with a button to change the application
Anotación 2020-07-26 180333

puppetsw:
**Resolved / Related Issues**
- Closes #1581

**Details of Changes**
Add details of changes here.
- Added new row in properties general showing application.
- Added button to change application (opens app picker)
- Note that at the moment the application name is only updated when the properties window gets focus. I wasn't sure of a way to detect when the Windows app picker has been closed or changed. Looking for suggestions on how to handle this.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image
image


yaichenbaum:
@puppetsw we still haven't decided if we want to implement this but if we do, the design calls for the option to be in the top section.
image

src/Files.FullTrust/Win32API.cs
```diff
@@ -122,6 +122,35 @@ string GetDesktopAssoc()
...
+
+            uint cOut = 0;
+            var fileExt = Path.GetExtension(fileName);
+
+            if (string.IsNullOrEmpty(fileExt))
+            {
+                return null;
+            }
+
+            if (ShlwApi.AssocQueryString(ShlwApi.ASSOCF.ASSOCF_VERIFY, ShlwApi.ASSOCSTR.ASSOCSTR_FRIENDLYAPPNAME, fileExt, null, null, ref cOut) != 1)
```
gave92, line 140: Could you check what happens if you try this on a file that has no associated program? By memory you may need to add the "ASSOCF_INIT_IGNOREUNKNOWN" flag to avoid getting wrong results. Haven't tested, does this API work when the associated app is a Store/UWP app?

src/Files.Uwp/Views/Pages/PropertiesGeneral.xaml.cs
```diff
@@ -115,5 +116,18 @@ private void DiskCleanupButton_Click(object sender, Windows.UI.Xaml.RoutedEventA
...
             }
         }
+
+        private async void ChangeApplicationButton_Click(object sender, RoutedEventArgs e)
+        {
+            var filePath = ViewModel.ItemPath + Path.DirectorySeparatorChar + ViewModel.ItemName;
+            await NavigationHelpers.OpenPath(filePath, AppInstance, openViaApplicationPicker: true);
+        }
+
+        private new async void GotFocus(object sender, RoutedEventArgs e)
```
gave92, line 126: An alternative is to use `SHChangeNotifyRegister` in fulltrust process to get notified when a file association changes. But I guess that would be unnecessarily complex.

puppetsw: If it's the proper way to do things it might be worth it. I'll look into it.

src/Files.FullTrust/Win32API.cs
puppetsw: At the moment it returns "Pick an application" as a string.

src/Files.Uwp/Views/Pages/PropertiesGeneral.xaml.cs
puppetsw: Okay I think you're right about the overly complex issue there, and I think it's above my skill level to implement correctly anyway.

src/Files.Uwp/ViewModels/SelectedItemsPropertiesViewModel.cs
```diff
@@ -107,9 +108,12 @@ public string ItemType
             {
                 ItemTypeVisibility = true;
                 SetProperty(ref itemType, value);
+                OnPropertyChanged(nameof(IsFileFolder));
             }
         }
 
+        public bool IsFileFolder => ItemType != "File folder";
```
gave92: Note that file type string is localized so this check only works is system language is English. Pushed a commit to change this.

src/Files.FullTrust/Win32API.cs
gave92: Thinking again I guess it's better to show the "Pick an application" instead of a blank space in case there's no associated program. I hadn't realized that Explorer shows that same string.

yaichenbaum:
@puppetsw let's move ahead with this feature 👍

gave92:
This looks good to me. The only sad thing is that there is no way to change the default app without opening the file. My understanding is that MS removed this possibility after windows 7.

yaichenbaum:
> This looks good to me. The only sad thing is that there is no way to change the default app without opening the file. My understanding is that MS removed this possibility after windows 7, so this is the best we can do.

Can't we change the registry value?

gave92:
No, or well not easily. Windows 10 added protection to stop programs from changing default file associations. I'm aware of this utility that can bypass that: http://kolbi.cz/blog/2017/10/25/setuserfta-userchoice-hash-defeated-set-file-type-associations-per-user/.

yaichenbaum:
Is there a shell api we can use?

yaichenbaum:
@ADeltaX seems like this is your kind of thing 😂
Any thoughts on how to use these flags from Files?

yaichenbaum:
@puppetsw I've been holding off on this because of the issue mentioned above. I think we can work around this by building our own open with dialog (I don't think it's that hard as we already have a lot of the code for the open with flyout menu). We will be modifying the registry, so we'll have to find a good way to do that.

puppetsw:
> @puppetsw I've been holding off on this because of the issue mentioned above. I think we can work around this by building our own open with dialog (I don't think it's that hard as we already have a lot of the code for the open with flyout menu). We will be modifying the registry, so we'll have to find a good way to do that.

I'll have a look at this

Issue #7321, opened 2021-12-19
Add support for adding drives to favorites section


ZaneDragonBorn:
### What's the Problem?

I can't favorite Drives or Files. This is kinda a bummer since it would give easier access to stuff. Also, you can't unfavorite stuff in the Home page.

### Solution/Idea

Allow them to be favorited. Only difference with a favorited file would be it is shown as a grid below favorited drives and folders.

### Alternatives

Nothing other than deal with it.

### Priorities

| Capability | Priority |
| :---------- | :------- |
| Favoriting of Drives | Must |
| Make it possible to toggle favorites in home menu | Should |
| Allow favoriting of files. | Could |
| This proposal will allow end users to accomplish Z | Won't |

### Files Version

Version: 2.0.34.0

### Windows Version

Windows 11 Home 21H2 22000.376

### Comments

Hope this gets added.

Naworm:
Please do it...

yaichenbaum:
This issue should be resolved with #9380

PR #9380, opened 2022-06-09
Allow to pin drives to favorites

IgorKordiukiewicz:
**Resolved / Related Issues**
- Related #7321

**Details of Changes**
Drives can now be pinned/unpinned to/from favorites either through sidebar or home page.

I used already existing localized texts, so should I add new ones where needed or leave it as it is?

**Validation**
- [x] Built and ran the app


yaichenbaum:
@IgorKordiukiewicz can't you already pin drives to favorites?

IgorKordiukiewicz:
The only way to do that now is by rightclicking on empty space in the disk's parent directory (but then if you rightclick the drive either in the sidebar or the drive's widget, there is no option to unpin it as compared to a folder).

Before:
rightclick_drive_before
After:
rightclick_drive_after

yaichenbaum:
> I used already existing localized texts, so should I add new ones where needed or leave it as it is?

It's good as is, we prefer reusing strings wherever possible.

yaichenbaum (approved):

PR #9391, opened 2022-06-11
Properties window is now shown near mouse cursor position

### LINKED ISSUE #9285: Show properties window near mouse cursor position
### What's the Problem?

In **File Explorer**, when you open the **Properties** of an item using <kbd>Alt</kbd>+<kbd>Enter</kbd> or from the context menu, the Properties window will show up at the bottom-right of the mouse except for when the mouse is near-edge of the screen, the window will adjust.

In **Files**, this is not possible as it will always show up to the last location of where you placed it.

### Solution/Idea

Make it show up near the mouse using its cursor position so users can expect where the window will show up.

### Alternatives

Show the Properties window on top of the current File instance like the image shown below.
Similar to a dialog window but not a modal window.

> The window here pops up in the center of the Files window
image

### Priorities

| Capability | Priority |
| :---------- | :------- |
| Show the **Properties** window near mouse position | Should |

### Files Version

2.3.0.0

### Windows Version

Windows 11 22H2 22621.1

### Comments

_No response_

IgorKordiukiewicz:
**Resolved / Related Issues**
- Closes #9285

**Details of Changes**

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Uwp/Helpers/FilePropertiesHelpers.cs
+++ b/src/Files.Uwp/Helpers/FilePropertiesHelpers.cs
@@ -82,6 +82,11 @@ namespace Files.Uwp.Helpers
                     {
                         // Set window size again here as sometimes it's not resized in the page Loaded event
                         appWindow.RequestSize(new Size(460, 550));
+
+                        DisplayRegion displayRegion = ApplicationView.GetForCurrentView().GetDisplayRegions()[0];
+                        Point pointerPosition = CoreWindow.GetForCurrentThread().PointerPosition;
+                        appWindow.RequestMoveRelativeToDisplayRegion(displayRegion,
+                            new Point(pointerPosition.X - displayRegion.WorkAreaOffset.X, pointerPosition.Y - displayRegion.WorkAreaOffset.Y));
                     }
                 }
                 else


yaichenbaum (approved):

gave92:
Curious: does this work when more than one monitor is connected?

IgorKordiukiewicz:
Yeah, subtracting display region's work area offset from pointer position handles that.

gave92:
Nice, the only issue I'm seeing is that this shows the properties window partially outside the screen bounds when the pointer is close to an edge.
Moving the call to RequestMoveRelativeToDisplayRegion after TryShowAsync appears to work correctly.
```cs
bool windowShown = await appWindow.TryShowAsync();
if (windowShown)
{
    // Set window size again here as sometimes it's not resized in the page Loaded event
    appWindow.RequestSize(new Size(460, 550));
    appWindow.RequestMoveRelativeToDisplayRegion(displayRegion,
        new Point(pointerPosition.X - displayRegion.WorkAreaOffset.X, pointerPosition.Y - displayRegion.WorkAreaOffset.Y));
}
```

IgorKordiukiewicz:
Updated

gave92 (approved):

yaichenbaum:
> Curious: does this work when more than one monitor is connected?

@gave92 good catch! I only tested this on a single monitor.

PR #9397, opened 2022-06-12
Removed code to handle GenerateUniqueName bug

### LINKED ISSUE #9277: File not showing up after renaming from conflict dialog (via drag & drop)
### Description

When dragging an image from Google Images to a folder, assume that the filename is already existing (in my case: **_"maxresdefault.png"_**), the conflict dialog will appear and will ask me to rename that file or do something else.
image

However, when I choose and confirm the **Generate new name** option, the file I dragged does not appear in the folder.
image

### Steps To Reproduce

1. Search any image from your browser
2. Drag that image to a folder
3. Drag the same image again from your browser to the folder (make sure to have both files share the same filename)
4. Conflict dialog should appear as expected
5. Choose **"Generate new name"** option and decide a new name for it.
6. Click **Continue** to confirm changes
7. Check if the 2nd image with a different name appears, and if it doesn't, that's the problem.

### Expected behavior

The renamed file should appear inside the folder along with the conflicting file.

### Files Version

2.3.0.0

### Windows Version

Windows 11 22H2 10.0.22621.1

### Relevant Assets/Logs

> The **_"asdfasdf.jpg"_** should be here.
image

_Note that this only happens when you drag an image from browser to the Files app._

puppetsw:
**Resolved / Related Issues**
- Closes #9277

**Details of Changes**
Add details of changes here.
- Removed code that doesn't seem necessary anymore as dragging from zip respects the GenerateUniqueName enum. This also fixes the issue with dragging from a browser.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.Uwp/Filesystem/FilesystemOperations/FilesystemOperations.cs
+++ b/src/Files.Uwp/Filesystem/FilesystemOperations/FilesystemOperations.cs
@@ -18,7 +18,6 @@ using Windows.UI.Xaml.Controls;
 using FileAttributes = System.IO.FileAttributes;
 using CommunityToolkit.Mvvm.DependencyInjection;
 using Files.Backend.Services;
-using System.Text.RegularExpressions;
 
 namespace Files.Uwp.Filesystem
 {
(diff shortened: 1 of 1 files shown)


puppetsw:
@gave92 Did you want to have a look at this one? I believe you added the original code to handle the bug that was there. From my testing it seems no longer necessary?

gave92:
Yep I'll check though I can't remember exactly what the problem was here. Like extracting the same file twice from .zip.

gave92:
Ok I was able to reproduce the issue. Steps are:.
- Open a .zip file _in windows explorer_
- Drag a file from the archive and drop it in a folder where an item with the same name already exists
- The existing file will get overwritten even if "generate new name" option was selected

Not sure what to do about that but sure I have to learn to write better comments -.-

puppetsw:
> Ok I was able to reproduce the issue. Steps are:
> 
> * Open a .zip file _in windows explorer_
> * Drag a file from the archive and drop it in a folder where an item with the same name already exists
> * The existing file will get overwritten even if "generate new name" option was selected
> 
> Not sure what to do about that but sure I have to learn to write better comments -.-

I'll have another look and see if there is a possible workaround, but if you weren't able to find anything I don't think I will. If that's the case do you want me to just update the comment with this PR or close it out?

puppetsw:
A bit of a hacky solution. What are your thoughts @gave92 ?

gave92:
> A bit of a hacky solution. What are your thoughts @gave92?

I don't like it 😜 but yeah it seems the only solution 👍

> I wonder if it's because the File object from the dragged zip file doesn't contain a path, so it isn't able to compare correctly.

Path being empty appears not to be the only factor. When dragging files from Phone storage, path is also empty but GenerateUniqueName works fine.

gave92:
@puppetsw do you mind if I tweak your solution slightly?

puppetsw:
> @puppetsw do you mind if I tweak your solution slightly?

Yeah go for it.

gave92:
Done, I've made changes so:
- There's no risk of a "race condition", e.g. of a file appearing between the call to TryGetItemAsync and the actual copy
- Avoids using StorageFolder methods as they may not work if the destination folder is another zip file or some other special location
- Should be a little faster if there is no conflict

puppetsw:
Looks much better now. I didn’t consider the race condition. Thanks for your help and input.

gave92 (approved):

PR #9435, opened 2022-06-19
Fix issues with drag&drop from special locations

gave92:
**Resolved / Related Issues**
- Fixes an issue preventing drag & drop from phone storage
- Fixes an issue causing a crash when dropping a file from a zip archive opened in Files and the destination already exists
- Fixes wrong filename when copy-pasting from a zip file opened in explorer and the destination already exists

**Details of Changes**
Add details of changes here.
- Fixed the GenerateUniquePath function so that a file named "test.txt" becomes "test (1).txt" instead of just "(1).txt" when copy-pasting from a zip file opened in explorer and the destination already exists.
- Take the most recent result for each source file at the end of a file operation. Fixes a crash due to duplicate key when calling "ToDictionary()" on file operation results.
- Removed dependency on `associatedInstance` from StorageHelpers. When drag&dropping the file does not come from the current instance. Fixes an issue preventing drag & drop from phone storage.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.FullTrust/Win32API.cs
+++ b/src/Files.FullTrust/Win32API.cs
@@ -505,7 +505,7 @@ namespace Files.FullTrust
 
                 for (ushort count = 1; File.Exists(uniquePath); count++)
                 {
-                    if (countMatch != null)
+                    if (countMatch.Success)
                     {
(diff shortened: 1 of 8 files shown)


yaichenbaum (approved):
LGTM

PR #9437, opened 2022-06-19
Use database to store layout preferences

gave92:
**Resolved / Related Issues**
I'm still thinking about this, opened to gather feedback mostly.
Pros:
- Allows to do stuff like "Reset all layout modes", "Set layout for all folders", etc..
These right know would require walking the HDD
- Don't leave traces when app is uninstalled

Cons:
- Reinstalling without exporting settings results in loss of layout prefs

**Details of Changes**
Add details of changes here.
- Store layout preferences in database instead of ADS of the folder
- Renaming or moving a folder preserves the layout mode
- No user-facing changes in functionality

**Validation**
- [x] Built and ran the app


yaichenbaum:
@gave92 are there any significant improvements over the current implementation?

gave92:
I'm still thinking about this.
Pros:
- Allows to do stuff like "Reset all layout modes", "Set layout for all folders", etc..
- Don't leave traces when app is uninstalled
Cons:
- Reinstalling without exporting settings results in loss of layout prefs

yaichenbaum:
> Reinstalling without exporting settings results in loss of layout prefs

Does the current method sync with File Explorer?

gave92:
No it doesn't, but saves the layout as alternate data stream of the folder. So it's not delete when you uninstall.

BanCrash:
> I'm still thinking about this, opened to gather feedback mostly.
> 
> Pros:
> 
> * Allows to do stuff like "Reset all layout modes", "Set layout for all folders", etc.. These right know would require walking the HDD
> * Don't leave traces when app is uninstalled
> 
> Cons:
> 
> * Reinstalling without exporting settings results in loss of layout prefs

I prefer the database way, IMO layout preferences are other type of settings, and so should be read from the settings instead of writing directly to the folder. Also that will give an easy way of make a fresh install.

zhuxb711:
Create an invisible "desktop.ini" in every folder is not unacceptable option

yaichenbaum:
> Create an invisible "desktop.ini" in every folder is not unacceptable option

That's not how the current method works

BanCrash:
Hi @gave92 , I know this is a draft and maybe never get merged but just in case I let you know that I've found an issue. If you have the same folder opened in two different tabs, changing the layout on one will not affect the other, even if you refresh the folder after. However, opening in a new tab will show the folder with the new layout.

GIF 23-06-2022 23-55-21

Sorry for using a Gif but Fluent Screen Recorder doesn't capture layout settings for some reason.

gave92:
Could you confirm whether `main ` bengalese differently?

BanCrash:
> Thanks for the feedback. Could you confirm whether `main` behaves differently?

No, you are right, It have the same behaviour, so it's unrelated to this PR.

yaichenbaum (approved):

src/Files.Uwp/Files.Uwp.csproj
```diff
@@ -235,6 +235,8 @@
     <Compile Include="Helpers\StorageSenseHelper.cs" />
     <Compile Include="Helpers\WindowDecorationsHelper.cs" />
     <Compile Include="Helpers\UniversalLogWriter.cs" />
+	<Compile Include="Helpers\LayoutPreferences\LayoutPreferences.cs" />
+	<Compile Include="Helpers\LayoutPreferences\LayoutPrefsDb.cs" />
```
chingucoding, line 239: nit: Indentation

src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
```diff
@@ -1,22 +1,30 @@
...
 
 namespace Files.Uwp.ViewModels
 {
     public class FolderSettingsViewModel : ObservableObject
     {
-        private static readonly ApplicationDataContainer localSettings = ApplicationData.Current.LocalSettings;
+        public static string LayoutSettingsDbPath => IO.Path.Combine(ApplicationData.Current.LocalFolder.Path, "layoutsettings.db");
+
+        private static readonly Lazy<LayoutPrefsDb> dbInstance = new(() => new LayoutPrefsDb(LayoutSettingsDbPath, true));
+        public static LayoutPrefsDb DbInstance => dbInstance.Value;
```
chingucoding, line 27: Out of curiosity, what would happen if for some reason DB creation takes ages or fails?

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
```diff
@@ -0,0 +1,143 @@
...
+
+namespace Files.Uwp.Helpers.LayoutPreferences
+{
+    public class LayoutPrefsDb : IDisposable
+    {
+        private readonly LiteDatabase db;
+
+        public LayoutPrefsDb(string connection, bool shared = false)
+        {
+            db = new LiteDatabase(new ConnectionString(connection)
```
chingucoding, line 14: Out of curiosity, was there a specific reason to use LiteDB?

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                Mode = shared ? FileMode.Shared : FileMode.Exclusive
+            }, new BsonMapper() { IncludeFields = true });
+        }
+
+        public void SetPreferences(string filePath, ulong? frn, LayoutPreferences prefs)
```
chingucoding, line 20: Maybe I miss a bit of context from not often working on Files but what does `frn` stand for?

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
```diff
@@ -0,0 +1,143 @@
...
+                if (prefs != null)
+                {
+                    // Insert new tagged file (Id will be auto-incremented)
+                    var newPref = new LayoutDbPrefs
+                    {
+                        FilePath = filePath,
+                        Frn = frn,
+                        Prefs = prefs
+                    };
+                    col.Insert(newPref);
```
chingucoding, line 37: Should we also call `col.ensureIndex(...)` to make lookup via the Filepath potentially faster?

src/Files.Uwp/Helpers/NativeFileOperationsHelper.cs
```diff
@@ -423,6 +448,21 @@ public static bool WriteBufferToFileWithProgress(string filePath, byte[] buffer,
             return result;
         }
 
+        // https://www.pinvoke.net/default.aspx/kernel32/GetFileInformationByHandleEx.html
```
chingucoding, line 451: Awesome that you include the link!

src/Files.Uwp/ViewModels/ColumnsViewModel.cs
```diff
@@ -269,5 +299,31 @@ public void TryMultiplySize(double factor)
...
+                    model.LengthIncludingGridSplitter.Value == this.LengthIncludingGridSplitter.Value &&
+                    model.UserLength.Value == this.UserLength.Value);
+            }
+            return base.Equals(obj);
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
+        }
```
chingucoding, line 327: As above, is this needed and if so, equals-hashcode contract might be a problem.

src/Files.Uwp/ViewModels/ColumnsViewModel.cs
```diff
@@ -129,28 +128,59 @@ private void SetColumnSizeProportionally(double factor)
...
+                    model.StatusColumn.Equals(this.StatusColumn) &&
+                    model.TagColumn.Equals(this.TagColumn));
+            }
+            return base.Equals(obj);
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
+        }
```
chingucoding, line 161: As mentioned earlier, do we need this and if so, equals-hashcode contract might cause issues if the DB relies on it.

src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
```diff
@@ -108,42 +116,20 @@ public bool IsLayoutModeChanging
...
-                    type = typeof(DetailsLayoutBrowser);
-                    break;
-            }
-            return type;
+                FolderLayoutModes.DetailsView => typeof(DetailsLayoutBrowser),
+                FolderLayoutModes.TilesView => typeof(GridViewBrowser),
+                FolderLayoutModes.GridView => typeof(GridViewBrowser),
+                FolderLayoutModes.ColumnView => typeof(ColumnViewBrowser),
+                _ => typeof(DetailsLayoutBrowser)
+            };
```
chingucoding, line 132: Great that you refactored this!

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPreferences.cs
```diff
@@ -0,0 +1,72 @@
...
+                    prefs.IsAdaptiveLayoutOverridden == this.IsAdaptiveLayoutOverridden &&
+                    prefs.ColumnsViewModel.Equals(this.ColumnsViewModel));
+            }
+            return base.Equals(obj);
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
+        }
```
chingucoding, line 70: Do we need this for DB stuff? In that case, maybe overriding the hashcode based on its own properties would be better since it would also ensure we adhere to the equals-hashcode contract. Having equals return true while the hashcodes are different might behave strangely.

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            }
+        }
+
+        public LayoutPreferences GetPreferences(string filePath = null, ulong? frn = null)
+        {
+            return _FindPreferences(filePath, frn)?.Prefs;
+        }
+
+        private LayoutDbPrefs _FindPreferences(string filePath = null, ulong? frn = null)
```
chingucoding, line 61: If we assign filePath to null, maybe we should also mark it as nullable?

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
gave92: It's NTFS "File Reference Number". It stays constant even if the filesystem path of a file or folder changes. I use it to keep the preferences intact when a folder is moved or renamed.

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
gave92: We use it for file tags, where it's needed to allow concurrent access to the DB from the UWP and FulltrustProcess processes. Since the dependency is there, I'm using it for layout prefs as well (I like the API more than e.g the SQLite one).

src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
gave92: I've looked at the source code for LiteDb and it does not do anything much in the constructor (not even creating the DB file). So this Lazy<> initialization is probably redundant.

src/Files.Uwp/Files.Uwp.csproj
gave92: Thanks, fixed.

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
gave92: Done, enabled nullable for this file.

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
gave92: Done, both on filepath and frn 👍

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPreferences.cs
gave92: It's used for this check, to avoid storing layout prefs in the DB if they match the default ones:
```cs
if (DbInstance.GetPreferences(folderPath, frn) is null)
{
    if (LayoutPreferences.DefaultLayoutPreferences.Equals(prefs))
    {
        return; // Do not create setting if it's default
    }
}
DbInstance.SetPreferences(folderPath, frn, prefs);
```
Not added GetHasCode() implementation, the way Resharper does it.

src/Files.Uwp/ViewModels/ColumnsViewModel.cs
gave92: Keeping comment below.

src/Files.Uwp/ViewModels/ColumnsViewModel.cs
gave92: Keeping comment below.

Laesx:
I much prefer the database way as well, because with the system as it currently stands it litters the folders a little too much and I have to put exceptions for my backups and for some other sync/file transfers, or when copying normally from explorer to say, an USB drive they get unhidden and are a pain.

You say there's an implementation to "set layout to all folders" I think it could be done a little like explorer does, if a folder mostly contains images it automatically selects that category for it and uses the layout you hasve set for that category to display it. 

Or at least be able to set a certain view for a folder AND all its subfolders.

src/Files.Shared/FileTagsDb.cs
```diff
@@ -167,8 +170,8 @@ public class TaggedFile
             [BsonId]
             public int Id { get; set; }
             public ulong? Frn { get; set; }
-            public string FilePath { get; set; }
-            public string Tag { get; set; }
+            public string FilePath { get; set; } = null!;
+            public string Tag { get; set; } = null!;
```
chingucoding, line 174: If those properties can be assigned null, should they be marked as nullable?

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
chingucoding: I think we should ensure the index if we insert, not if we look up the entities right? Or maybe I misunderstood what it does but creating the index while searching is not ideal, creating it before you run your query is better.

src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
```diff
@@ -690,6 +549,7 @@ public void OnDefaultPreferencesChanged(string folderPath, string settingsName)
                     break;
                 case nameof(UserSettingsService.PreferencesSettingsService.AreLayoutPreferencesPerFolder):
                     LayoutPreference = prefs;
+                    // TODO: update layout
```
chingucoding, line 552: Is this something being removed by the PR or was this never the case? Maybe this is worth tracking via an issue?

src/Files.Shared/FileTagsDb.cs
gave92: Better string.empty?

src/Files.Uwp/ViewModels/FolderSettingsViewModel.cs
gave92: Ah no this is just adding the comment. But worth opening an issue.

src/Files.Shared/FileTagsDb.cs
chingucoding: Yeah maybe that would be better.

src/Files.Uwp/Helpers/LayoutPreferences/LayoutPrefsDb.cs
gave92: Moving it after the insert sounds right, thanks. 👍

chingucoding (approved):

yaichenbaum:
> I think it could be done a little like explorer does, if a folder mostly contains images it automatically selects that category for it and uses the layout you hasve set for that category to display it.

@Laesx this is how the adaptive layout works but there is another option to turn off adaptive layout and to apply the same layout no matter what the contents are.

yaichenbaum (approved):
LGTM

PR #9454, opened 2022-06-22
Service IFullTrustAsker

cinqmilleans:
**Resolved / Related Issues**
The move from Storage to Backend is imminent. However, you need to be able to access FullTrust from the backend to be able to make the move.

**Details of Changes**
Adds the IFullTrustAsker service to wrap and simplify calls to FullTrust. This service is not currently called, but it is possible to simplify current calls using this service.

If all the calls are replaced, it will be possible to modify to include in the encapsulation elements that will no longer need to be in a public helper.

**Validation**
- [x] Built and ran the app


cinqmilleans:
Fixed merge

d2dyno1:
As mentioned on discord, we don't want to use this pattern since it'll make things harder when transitioning to WinUI 3

cinqmilleans:
@d2dyno1. I don't see how we can do without this service. We need to use FullTrust in another project because we don't have a workaround for Storage. @gave92 asks for this service on discord. This service will no longer be useful if we replace IStorageItem, but we won't be able to do that until WinUi3 and the transition isn't coming soon.

PR #9469, opened 2022-06-26
Fix issues with Sideload solution config

gave92:
Noticed a couple fo strange things in the Sideload configuration.

**Details of Changes**
Add details of changes here.
- Removed Deploy from Files.Uwp project in Sideload config. Only Files.Package should be deployed, right?
- Also changed Sideload to build for "arm64" instead of "Any CPU" to match Release config
- Changed OpenFileDialog project to compile in Release mode when in Sideload config (unimportant)

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
Thanks! Wish it was easier to create new configs 😆

PR #9475, opened 2022-06-28
Feature: Prompt users to review Files

yair100:
**Details of Changes**
Users who have opened the app more than 10 times will be prompted to review Files in the Microsoft Store. This prompt will only show once and will not come up again unless the app is reset or reinstalled.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Uwp/ServicesImplementation/Settings/ApplicationSettingsService.cs
+++ b/src/Files.Uwp/ServicesImplementation/Settings/ApplicationSettingsService.cs
@@ -1,0 +1,25 @@
+﻿using Files.Backend.Services.Settings;
+using Files.Shared.Enums;
+using Files.Uwp.Serialization;
+
+namespace Files.Uwp.ServicesImplementation.Settings
+{
+    internal sealed class ApplicationSettingsService : BaseObservableJsonSettings, IApplicationSettingsService
+    {
+        public bool WasPromptedToReview
+        {
+            get => Get(false);
+            set => Set(value);
+        }
+
+        public void ReportToAppCenter()
+        {
+           
+        }
+
+        public ApplicationSettingsService(ISettingsSharingContext settingsSharingContext)
+        {
+            RegisterSettingsContext(settingsSharingContext);
+        }
+    }
+}
--- a/src/Files.Uwp/Strings/en-US/Resources.resw
+++ b/src/Files.Uwp/Strings/en-US/Resources.resw
@@ -2789,4 +2789,10 @@ We use App Center to track which settings are being used, find bugs, and fix cra
   <data name="ShowAlternateStreams" xml:space="preserve">
     <value>Show Alternate Data Streams</value>
   </data>
+  <data name="ReviewFiles" xml:space="preserve">
+    <value>Review Files</value>
+  </data>
+  <data name="ReviewFilesContent" xml:space="preserve">
+    <value>Would you like to review Files?</value>
+  </data>
 </root>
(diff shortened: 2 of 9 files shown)


src/Files.Uwp/Views/MainPage.xaml.cs
```diff
@@ -75,9 +79,38 @@ public MainPage()
             ToggleCompactOverlayCommand = new RelayCommand(ToggleCompactOverlay);
             SetCompactOverlayCommand = new RelayCommand<bool>(SetCompactOverlay);
 
+            if (SystemInformation.Instance.TotalLaunchCount >= 10 & Package.Current.Id.Name == "49306atecsolution.FilesUWP" && !UserSettingsService.ApplicationSettingsService.WasPromptedToReview)
```
yaichenbaum, line 82: We should really create a separate service for this

src/Files.Uwp/ServicesImplementation/Settings/ApplicationSettingsService.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Files.Backend.Services.Settings;
+using Files.Shared.Enums;
+using Files.Uwp.Serialization;
+
+namespace Files.Uwp.ServicesImplementation.Settings
+{
+    internal sealed class ApplicationSettingsService : BaseObservableJsonSettings, IApplicationSettingsService
+    {
+        public bool WasPromptedToReview
```
gave92, line 9: I think like this the setting will be included when exporting settings. Not sure if it's a problem. And I'm not sure if/how we esclude some settings from the import/export.

src/Files.Uwp/Views/MainPage.xaml.cs
gave92: Perhaps we can combine the logic and the setting in a "ReviewPromptService" or similar.

gave92:
@yair100 i wont be home till next week. Feel free to assign this to someone else. Code looks good to me, but for the doubts above.

src/Files.Uwp/ServicesImplementation/Settings/ApplicationSettingsService.cs
yaichenbaum: That's good, we don't want to prompt multiple times.

src/Files.Uwp/Views/MainPage.xaml.cs
yaichenbaum: Definitely, the question is if we can get this in as is for now given the high impact it'll have in the store?

BUM2MILLIONAIRE:
Hi guys
Hope you enjoy Github I am new to the world of programming but I can see
that you are doing good work enjoy practice makes it perfect☺
Thank you
Ester

src/Files.Uwp/Strings/en-US/Resources.resw
```diff
@@ -2789,4 +2789,10 @@ We use App Center to track which settings are being used, find bugs, and fix cra
   <data name="ShowAlternateStreams" xml:space="preserve">
     <value>Show Alternate Data Streams</value>
   </data>
+  <data name="ReviewFiles" xml:space="preserve">
+    <value>Review Files</value>
+  </data>
+  <data name="ReviewFilesContent" xml:space="preserve">
+    <value>Would you like to review Files?</value>
```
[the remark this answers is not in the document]
yaichenbaum: I prefer the current wording as the yes/no action fits with the actual question. Just because someone enjoys Files, they might not be interested in reviewing it.

PR #9477, opened 2022-06-30
Optimized RD file

yaichenbaum:
Trimmed rd.xml file to reduce binary size and improve startup time. We'll have to verify that everything works correctly in release mode.


ricowalker (approved):

src/Files.Uwp/Properties/Default.rd.xml
```diff
@@ -21,16 +21,16 @@
       An Assembly element with Name="*Application*" applies to all assemblies in
       the application package. The asterisks are not wildcards.
     -->
-    <Namespace Name="SQLitePCL" Dynamic="Required All" />
```
gave92, line 24: I think this breaks SQLite, e.g. google drive detection. I'll check.

Issue #9479, opened 2022-06-30
Merge Files.Uwp\Helpers\NativeIoDeviceControlHelper.cs and Files.Uwp\Helpers\NativeFileOperationsHelper.cs


peteraritchie:
### What's the Problem?

Files.Uwp\Helpers\NativeIoDeviceControlHelper.cs and Files.Uwp\Helpers\NativeFileOperationsHelper.cs contain some of the same definitions creating duplicate code and is either redundant or introduces maintenance issues.

### Solution/Idea

Merge Files.Uwp\Helpers\NativeIoDeviceControlHelper.cs and Files.Uwp\Helpers\NativeFileOperationsHelper.cs to remove duplicate declarations and definitions.

### Alternatives

Leave it alone and accept the maintenance risk.

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish W | Must |
| This proposal will allow end users to accomplish X | Should |
| This proposal will allow developers to accomplish Y | Could |
| This proposal will allow end users to accomplish Z | Won't |

d2dyno1:
@peteraritchie Would you like to work on it? 🙂

PR #9484, opened 2022-07-02
Refactor storage native

cinqmilleans:
Refactor storage native before migration: Less redundancy, more consistency in filenames, only one item per file. Some elements remained duplicated in the backend. It is temporary until the migration. #9479

**Validation**
- [x] Built and ran the app


yaichenbaum:
@d2dyno1

d2dyno1:
Now that we've switched to WinUI 3, these native pinvokes are not needed anymore and can be replaced with Vanara library. It'd be more beneficial if we instead refactored helper classes and only used what Vanara has to offer

PR #9544, opened 2022-07-11
Introduce abstract storage layer

d2dyno1:
**Details of Changes**
This PR continues work on the infrastructure, making the storage layer more broad and flexible. The motivation for this change is to eliminate the convoluted file system code that Files is built on - this requires major changes both in frontend and backend part of the codebase. This work is a stretch goal to ultimately replace all storage calls with `Files.Sdk.Storage` abstractions. This PR is a work-in-progress.

**TODO**
*The following list is limited to this PR and doesn't include future changes*
- [x] Implement FtpFile/Folder
- [ ] ~~Implement NativeFile/Folder~~
- [ ] ~~Implement ShellFile/Folde~~
- [ ] ~~Implement SystemFile/Folder~~
- [ ] ~~Implement ZipFile/Folder~~

(Crossed out items will be done in following PRs)

**Validation**
- [x] Built and ran the app


cinqmilleans:
@d2dyno1 We agree on the essentials. I will clarify my vocabulary so we understand the same thing.

Storage. This is a technical layer that uses IStorageItem from uwp. This roughly corresponds to the current Files.Uwp.Filesystem namespace. It is a technical layer but is used at high level by the application (wrongly). Storage uses FullTrust but this is a limitation of uwp. I hadn't planned to use an abstraction for this layer because ListedItem will encapsulate it and it will be used a lot less. However, your suggestion to add an abstraction convinced me. This will allow us to easily replace this layer if necessary, especially if we leave uwp.

Filesystem. This is an abstraction that encapsulates Storage (and libraries such as FluentFTP) to be able to manipulate ftp cloud zip and other system files and folders. It roughly corresponds to ListedItem and operations (copy, rename, ...). Currently, there is no abstraction of operations. The ListedItem abstraction has major flaws and is not standardized. Operations directly use the storage layer, which should be avoided.

Filesystem contains 2 parts. The sdk (public) and the implementation (private, except enumeration services or factory to be able to instantiate the services). I had planned to place these 2 parts in the same project. The separation can be done later if necessary. I had planned to place Storage in the Files.Filesystem project, without abstraction, but encapsulated.

An important point is the breakdown of the application into several projects. I tend to create a lot of projects for an application to have a clear and impossible to violate architecture. I didn't apply it here because there may be team objections to creating multiple projects.

Here is a suggestion of some projects we could create.
Files.Sdk.Core (abstraction of basic services such as logger, localization, ...)
Files.Sdk.Storage (abstraction)
Files.Sdk.Filesystem (abstraction)
Files.Shared.Extensions (implementation of generic extensions of string, linq, ...)
Files.Backend.Storage (implementation)
Files.Backend.Filesystem (implementation)

d2dyno1:
Let's keep things simple for now. We'll try to start with this layout in mid:

Files.Sdk.Storage - storage abstractions
Files.Uwp.Storage - implementation of those abstractions (satisfying UWP platform limitations)

We can later on decide and separate the ListedItem layer (we want to avoid confusion and have a clear foundation that we can expand upon later).

The current goal is to separate the storage (- ftp, win32, shell, etc.) from the backend (- ViewModels). The abstractions for backend view models are out of scope of this PR 🙂

@cinqmilleans

d2dyno1:
I've implemented storage abstractions for FTP. We'll need to implement other classes this way as well :)

cinqmilleans:
This is ok for me.

cinqmilleans:
@yair100 @d2dyno1 @gave92 Here are some ideas about Storage. Storage is the part of the application that manages file systems (whatever they are). You need a Storage backend and a Storage viewModel.

The backend needs an abstraction, many implementations (which may change if we leave uwp), tests, etc. This layer must be self-contained, with no dependencies on user settings or localization. This part is not specific to Files. The viewmodel will present Storage in a more convenient way. It will be independent of the real file system (cloud, ftp, ...) but will use user settings and location. This part is specific to Files. The most important point is therefore the backend. A CommunityToolkit.Storage project is in the works, but we can't afford to wait for it (if it comes to fruition).

I propose to no longer integrate Storage into the Files project. It is a standalone project that could be used by another application (which is not a file manager). We can create a new Repository project "Files Community/Storage". This will contain projects for abstraction, uwp implementations, testing, and its own documentation. This ensures our own independence.

In parallel, we can collaborate with CommunityToolkit.Storage to integrate our own improvements. This with the aim, in the long term, of using it to replace our solution (which allows us to have a head start).

src/Files.Sdk.Storage/Enums/CreationCollisionOption.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Files.Sdk.Storage.Enums
+{
+    public enum CreationCollisionOption : byte
```
yaichenbaum, line 3:
```suggestion
    public enum CollisionOption : byte
```
Can this be simplified?

src/Files.Backend/Services/IFileExplorerService.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// A service that interacts with the system file explorer.
+    /// </summary>
+    public interface IFileExplorerService
+    {
+        /// <summary>
+        /// Opens the app folder.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that cancels this action.</param>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous operation.</returns>
+        Task OpenAppFolderAsync(CancellationToken cancellationToken = default);
```
yaichenbaum, line 18: Very good to see proper comments, nicely done!

src/Files.Sdk.Storage/Extensions/StorageExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Files.Sdk.Storage.Extensions
+{
+    public static class StorageExtensions
+    {
+        public static async Task CopyContentsToAsync(this IFile source, IFile destination, CancellationToken cancellationToken = default)
```
yaichenbaum, line 9: Should we add documentation/comments here?

src/Files.Uwp.Storage/Properties/Files.Uwp.Storage.rd.xml
```diff
@@ -0,0 +1,33 @@
...
+
+    3.  AttributeImplies
+        Use an AttributeImplies directive to indicate that your library needs to reflect over
+        types or methods decorated with an attribute.
+
+    For more information on writing Runtime Directives for libraries, please visit
+    https://go.microsoft.com/fwlink/?LinkID=391919
+-->
+<Directives xmlns="http://schemas.microsoft.com/netfx/2013/01/metadata">
+  <Library Name="Files.Uwp.Storage">
```
yaichenbaum, line 28: Can we add a parameter like `Required="Public"`?

src/Files.Uwp/Filesystem/StorageItems/FtpStorageFile.cs
```diff
@@ -39,7 +38,7 @@ public override string DisplayType
         }
 
         public override DateTimeOffset DateCreated { get; }
-        public override Storage.FileAttributes Attributes { get; } = Storage.FileAttributes.Normal;
+        public override Windows.Storage.FileAttributes Attributes { get; } = Windows.Storage.FileAttributes.Normal;
```
yaichenbaum, line 41: Can this be simplified?

src/Files.Uwp/Filesystem/StorageItems/FtpStorageFolder.cs
```diff
@@ -26,7 +25,7 @@ public sealed class FtpStorageFolder : BaseStorageFolder
         public override string FolderRelativeId => $"0\\{Name}";
 
         public override DateTimeOffset DateCreated { get; }
-        public override Storage.FileAttributes Attributes { get; } = Storage.FileAttributes.Directory;
+        public override Windows.Storage.FileAttributes Attributes { get; } = Windows.Storage.FileAttributes.Directory;
```
yaichenbaum, line 28: Same as previous comment

src/Files.Uwp/ServicesImplementation/FileExplorerService.cs
```diff
@@ -0,0 +1,71 @@
...
+            await Launcher.LaunchFolderAsync(ApplicationData.Current.LocalFolder).AsTask(cancellationToken);
+        }
+
+        /// <inheritdoc/>
+        public async Task OpenInFileExplorerAsync(ILocatableFolder folder, CancellationToken cancellationToken = default)
+        {
+            await Launcher.LaunchFolderPathAsync(folder.Path).AsTask(cancellationToken);
+        }
+
+        /// <inheritdoc/>
```
yaichenbaum, line 32: What does this stand for, is it a placeholder?

src/Files.Sdk.Storage/Enums/CreationCollisionOption.cs
d2dyno1: There's also NameCollisionOption which has some key differences.

src/Files.Backend/Services/IFileExplorerService.cs
d2dyno1: Copied them from SFFS 😅 
We should definitely comment our code as part of the refactor (and also include <inheritdoc/>)

src/Files.Uwp/Filesystem/StorageItems/FtpStorageFile.cs
d2dyno1: No, it cannot since we use another namespace which has the same class.
I intentionally don't want it to be simplified because it decreases readability and it **won't work** with Files.Sdk.Storage namespace!!

src/Files.Uwp/Filesystem/StorageItems/FtpStorageFolder.cs
d2dyno1: No

src/Files.Uwp/ServicesImplementation/FileExplorerService.cs
d2dyno1: It tells the VS that this method should inherit the comment from base method  (override method, interface method). I prefer it to stay as it also implies that the implemented method has a comment with it and if the code was changed, those changes should also be reflected in the comment

src/Files.Sdk.Storage/Extensions/StorageExtensions.cs
d2dyno1: Yes, but we'll add those later on (perhaps). Those methods can be volatile and be removed/modified often

src/Files.Uwp.Storage/Properties/Files.Uwp.Storage.rd.xml
d2dyno1: You could do that. I haven't dealt with `.rd.xml` files before

src/Files.Backend/Services/IFileExplorerService.cs
yaichenbaum: Can you add this to the code guidelines?

src/Files.Uwp.Storage/Properties/Files.Uwp.Storage.rd.xml
yaichenbaum: Trimming is something we should try to support

d2dyno1:
> The name CreateCopyOfAsync is inconsistent with MoveOfAsync. CopyOfAsync is simpler.

The current names are gramatically consistent

<br>

> It is cleaner to use enums instead of boolean as method arguments.

I don't know what arguments you're referring to

<br>

-
> The delete should be on an IModifiableStorage interface and delete the item itself. With the current model, you have to call the parent to delete the child.

The delete operation should be on parent folder since it needs to be *contained* within some container to be able to delete it. The parent folder also needs to be notified of the deletion

<br>

> It is not possible to know if an operation is possible before performing it. If the folder is read-only (in particular), we are forced to throw an exception. It's not efficient. Add CanCreateItemAsync, CanCopyItemAsync, ...

If a folder is read-only, it cannot be cast to `IModifiableFolder` in the first place.

<br>

> This interface does not allow advanced management, some of which are available on all other file managers (including File Explorer) or could bring great added value. (Pausing a copy, Percentage of completion).

These could be extended with additional interfaces or wrapped around interface like `IStorageTransfer` to allow for more in-depth control of the operation
@cinqmilleans

yair100 (dismissed):

yair100 (approved):

yaichenbaum:
@gave92 can I merge this?

gave92 (approved):

Issue #9546, opened 2022-07-12
Feature: Add chevron to folders in column layout


yaichenbaum:
### Solution/Idea

![image](https://user-images.githubusercontent.com/39923744/178402959-bd0c6814-8687-4f17-ad5f-d791c71ca8e7.png)

### Alternatives

NA

### Priorities

| Capability | Priority |
| :---------- | :------- |
| Add chevron to folders in the column layout | Must |

cinqmilleans:
How should cloud icons be managed? Should they be aligned vertically or positioned to the right?

![FolderChevron1](https://user-images.githubusercontent.com/46631671/178499979-d1cb868d-4642-4e76-b7d7-4c759cb8c2f2.png)
![FolderChevron2](https://user-images.githubusercontent.com/46631671/178499981-2a37c7d2-e63e-4bd1-901e-0a9601f423d1.png)

yaichenbaum:
That's a good question, we could hide the status for folders and only show it on the files but that doesn't seem like the best solution. Let's do it like the first screenshot for now and we can always revisit this later on.

yaichenbaum:
@cinqmilleans I gave this some more thought and I think we should hide the status in the column layout. We can still show it in a teaching tip, but I no longer think that it has a place in the list.

PR #9549, opened 2022-07-12
Adds chevron to folder in column view

cinqmilleans:
Adds chevron to folder in column view. #9546

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
FolderChevron_0
FolderChevron_1


mdtauk:
Might you consider changing the chevron to use the secondary or tertiary brushes.  They don't need to be quite so prominent because they are not directly interactable compared to the other items in the list.

cinqmilleans:
I don't know what the secondary or tertiary brush is. Here is a suggestion with opacity 0.6.
FolderChevron_2
FolderChevron_3

src/Files.Uwp/Views/LayoutModes/ColumnViewBase.xaml
```diff
@@ -332,6 +333,13 @@
                                                 Visibility="{x:Bind FileTagsUI, Converter={StaticResource EmptyObjectToObjectConverter}, Mode=OneWay}" />
                                         </StackPanel>
                                     </Grid>
+                                    <FontIcon
+                                        Grid.Column="3"
+                                        HorizontalAlignment="Right"
+                                        VerticalAlignment="Center"
+                                        FontSize="12"
+                                        Glyph="&#xE76C;"
+                                        Visibility="{x:Bind IsFolder}" />
```
yaichenbaum, line 342:
```suggestion
                                        x:Load="{x:Bind IsFolder}" />
```

yaichenbaum:
Note to self to test this with Accessibility Insights

cinqmilleans:
- The resources I'm being asked to use don't support the dark theme. It is not possible to have an adequate color in the 2 themes.
- Synchronization and color tag information is important and is available in other views. Perhaps it should be displayed before or after the text.

yaichenbaum:
> The resources I'm being asked to use don't support the dark theme. It is not possible to have an adequate color in the 2 themes.

I'll investigate this and send you the proper resource to use

> Synchronization and color tag information is important and is available in other views. Perhaps it should be displayed before or after the text.

This information can be displayed in a tooltip and the details pane but it doesn't fit in the file area anymore as it would be too crowded with it.

yaichenbaum:
@cinqmilleans does `TextFillColorSecondary` work?

yaichenbaum:
@cinqmilleans can you push the code so I can point you in the right direction? I think you might be using `StaticResource` instead of `ThemeResource`.

cinqmilleans:
Your answer helped me. I was using StaticResource instead of ThemeResource. With ThemeResource, it's ok in both themes. I changed the pr to use TextFillColorTertiary, because TextFillColorSecondary is not discrete enough.

yaichenbaum:
Can post an updated screenshot?

cinqmilleans:
Here.
Chevron1
Chevron2
Chevron3
Chevron4

yaichenbaum (approved):

yaichenbaum:
@cinqmilleans oh, one more thing before merging, did you test that the chevrons don't get in the way of the screen reader?

yaichenbaum (approved):

PR #9601, opened 2022-07-24
Fixed properties window in sideload preview

yaichenbaum:
**Details of Changes**
Add details of changes here.
Fixed an issue where the minimize and maximize buttons were visible in the properties window for the sideload preview build.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.Uwp/Helpers/WindowDecorationsHelper.cs
+++ b/src/Files.Uwp/Helpers/WindowDecorationsHelper.cs
@@ -20,6 +20,7 @@ namespace Files.Uwp.Helpers
                 {
                     "Files" => "xnYLj99c3vN9jFCZiDC6Rg==",
                     "FilesDev" => "Yoz9y1X66micskUKhFrJ5A==",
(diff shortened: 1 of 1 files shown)


src/Files.Uwp/Helpers/WindowDecorationsHelper.cs
```diff
@@ -20,6 +20,7 @@ public static void RequestWindowDecorationsAccess()
                 {
                     "Files" => "xnYLj99c3vN9jFCZiDC6Rg==",
                     "FilesDev" => "Yoz9y1X66micskUKhFrJ5A==",
+                    "FilesPreview_wvne1zexy08sa" => "jz4pZSdYrC05IyNfTaNwZQ==",
```
ahmed605, line 23: This is the package family name not the package name, the correct value should be `FilesPreview`

yaichenbaum: Good catch, this should be fixed in the latest commit.

ahmed605 (approved):

PR #9633, opened 2022-08-01
Add support for more archive types using 7zip

gave92:
**Resolved / Related Issues**
- Related #45

**Details of Changes**
Add details of changes here.
- Added support for browsing and editing rar, 7z, zip archives

TODO:
- [x] Fix renaming folders
- [x] Test in Release mode

**Validation**
- [x] Built and ran the app


yaichenbaum:
Is there anything I can do to help test this?

gave92:
Release mode now works 💪

@yair100 I'm not sure how to make so the users can choose whether to use Files or external program for archives. The current logic (using Files if the default app for zips is explorer) is ok for zips but not so much for other archives (7z, rar). Should we add an "Open archives in Files" setting? Or use Files only if set as default for that particular extension?

yaichenbaum:
How would an option to open archives in Files work? I'm pretty sure Microsoft prefers that we let users change the default on their own.
>  Or use Files only if set as default for that particular extension?
I'm leaning towards this but if there is a supported way to change the defaults for the other ones, it wouldn't hurt to add an option.

yaichenbaum:
Some additional feedback I got was that we can add shell extensions to File Explorer for extra visibility.

bbbhksgj (dismissed):

src/Files.Uwp/Helpers/FileExtensionHelpers.cs
```diff
@@ -47,14 +48,15 @@ public static bool IsPowerShellFile(string fileExtensionToCheck)
...
+            return new[] { ".zip", ".msix", ".appx", ".msixbundle", ".7z", ".rar", ".tar" }
+                .Contains(fileExtensionToCheck, StringComparer.OrdinalIgnoreCase);
+        }
 
-            return fileExtensionToCheck.Equals(".zip", StringComparison.OrdinalIgnoreCase) ||
-                   fileExtensionToCheck.Equals(".msix", StringComparison.OrdinalIgnoreCase) ||
-                   fileExtensionToCheck.Equals(".msixbundle", StringComparison.OrdinalIgnoreCase);
+        public static bool IsBrowsableZipFile(string filePath, out string ext)
+        {
+            ext = new[] { ".zip", ".7z", ".rar", ".tar" } // Only ext we want to browse
```
yaichenbaum, line 57: Can we add msix and appx to this list? They're basically self-extracting zip files.

src/Files.Package/Files.Package.wapproj
```diff
@@ -100,6 +100,12 @@
     <ProjectReference Include="..\Files.FullTrust\Files.FullTrust.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <Content Include="7z.dll">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
yaichenbaum, line 104: Does this work with the MIT license?

yair100 (dismissed):

src/Files.Uwp/Helpers/FileExtensionHelpers.cs
gave92: Not here, because then double clicking on them would browse the as archives instead of installing them. 
Above I believe I added msix and msixappxbundle to show the extract menu on them.

src/Files.Package/Files.Package.wapproj
gave92: I'm no lawyer but yeah it does. 7z.dll is distributed with LGPG license which is compatible with MIT.
This from 7zip website:
> Can I use the EXE or DLL files from 7-Zip in a Commercial Application?
Yes, but you are required to specify in your documentation (1) that you used parts of the 7-Zip program, (2) that 7-Zip is licensed under the GNU LGPL license and (3) you must give a link to www.7-zip.org, where the source code can be found.

yair100: We can add it to this page
image

yair100 (approved):
LGTM

bbbhksgj (approved):

src/Files.Uwp/Filesystem/BaseStorage/IBaseStorageFolder.cs
```diff
@@ -38,4 +39,11 @@ public interface IBaseStorageFolder : IStorageItem2, IStorageFolder, IStorageFol
         new BaseStorageFileQueryResult CreateFileQueryWithOptions(QueryOptions queryOptions);
         new BaseStorageFolderQueryResult CreateFolderQueryWithOptions(QueryOptions queryOptions);
     }
+
+    public interface ICreateFileWithStream
```
d2dyno1, line 43: What implements this interface?

d2dyno1, line 43: Is it needed? It adds additional complexity

gave92: Yeah it's needed because 7zip library requires a Stream with the contents when you create a new file to add to the archive. So the usual API steps "create an empty file", "open file as readwrite", "write contents" can't be used.

d2dyno1: 😫

d2dyno1 (approved):

PR #9669, opened 2022-08-07
Fix issue causing user settings to be repeatedly reloaded from file in some cases

gave92:
**Details of Changes**
Add details of changes here.
- Fixed an issue causing settings to be repeatedly read from file in some cases
- Moved public SemaphoreSlim from App.cs to StorageHistoryHelpers (only place it's used)

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

PR #9716, opened 2022-08-13
Fix google drive detection

gave92:
**Resolved / Related Issues**
- Fixes google drive detection

**Details of Changes**
Add details of changes here.
- Copy google drive's *.db-wal file in addition to *.db file
- Fixed rd.xml for SQLite

**Validation**
- [x] Built and ran the app


src/Files.Uwp/Properties/Default.rd.xml
```diff
@@ -21,7 +21,8 @@
       An Assembly element with Name="*Application*" applies to all assemblies in
       the application package. The asterisks are not wildcards.
     -->
-    <Assembly Name="SQLitePCL" Dynamic="Required Public" />
+    <Type Name="SQLitePCL.Batteries_V2" Dynamic="Required Public" />
```
yaichenbaum, line 24: What's this change for?

yaichenbaum (approved):
LGTM

src/Files.Uwp/Properties/Default.rd.xml
gave92: Does not work in Release mode otherwise 😕

Issue #1271, opened 2020-06-27
Feature: Display recycle bin size when hovering over the sidebar item


Rick-HM:
**Requirements**
- Update tooltip that's shown when hovering over recycle bin to display how much space it's taking up

PR #9734, opened 2022-08-16
Feature: Display drive details when hovering over the sidebar

### LINKED ISSUE #9170: Hide empty drive bays in the sidebar
### What's the Problem?

In Explorer, drives like disc drives and floppy drives are only shown in the sidebar when there is something in them. In Files, empty drive bays are still shown in the sidebar.

### Solution/Idea

When a drive bay is empty, Files should hide it from the sidebar (but it should still be visible on the Home screen).

### Alternatives

​

### Priorities

_No response_

### Files Version

Version 2.1.24.0

### Windows Version

Version 21H1, Build 10.0.19043.1645, Windows Feature Experience Pack 120.2212.4170.0

### Comments

_No response_

ferrariofilippo:
**Resolved**
- Closes #9170 

**Details of Changes**
-When inserting children in the DRIVE section, I check if these drives contains at least one file/folder 
-When a file/folder is created/pasted, the DRIVE section is updated

**Validation**
- Built and ran the app


yaichenbaum:
What happens if a user inserts a brand new usb drive?

ferrariofilippo:
Sorry I misunderstood issue's request.
Now it's fixed for Floppy Discs and CDs only

yaichenbaum:
You did a great job but I'm still trying to understand the goal of the feature request.

ferrariofilippo:
Honestly, I think that this feature, as asked in the issue, is pretty outdated. Nowadays, few people still use floppy disks/CDs.
At first, I thought this feature would aim at all removable drives. In this case, I would have used it in order to keep my sidebar tidy when I have many SDs/USB sticks plugged in.
I'll explain myself better: I like to take photos and in my backpack, I have like 15 SDs. Of these, in a day I might fill 4 or 5. When I come back home I don't remember which I used, so I plug them into my PC in batches of 6.
It's not a game-changing feature, but it would save me some time and I'd use it

yaichenbaum:
That is a valid use case but the home page displays the drive space, I'm wondering if that's a better way to see which one has items in it.

yaichenbaum:
We used to display the item size when hovering over the sidebar items, perhaps we can bring that back.

ferrariofilippo:
Ok, I'll work on this. Should I add the feature #1271 ?

yaichenbaum:
@ferrariofilippo that's correct, we need to also do this in a way that still displays the full item path.

ferrariofilippo:
I worked on this for a while. I found out that on the drive section there's no problem. Instead, calculating the size of "Favorites" folders takes too much time: calculating the size of 3 folders took me 5 minutes. I used the Files method, not StorageFolder.
Personally, I think we should add this feature only to drives.
Adding it to the recycle bin, might result in the same problem as with other folders: if you have too much trash it will take ages to calculate the size. 
Let me know what you think I should do

yaichenbaum:
I'm happy to do this for just drives 👍

yaichenbaum:
@ferrariofilippo I'm happy with the current solution but another alternative is to provide a setting for this. Whichever solution you like more is fine with me.

ferrariofilippo:
I don't think adding a setting for this feature is worthy. Let me explain: showing the size of drives has little impact on performance, since the size is always calculated. Moreover, adding a new setting may result in having too many of them.
On the UX side, the tooltip still remains really small, so it should not be a problem for the user.
My porposal is to add the feature as is and wait for users' opinion: if some of them find it annoying and they want a setting for it I'll add it

yaichenbaum:
> I don't think adding a setting for this feature is worthy. Let me explain: showing the size of drives has little impact on performance, since the size is always calculated. Moreover, adding a new setting may result in having too many of them. On the UX side, the tooltip still remains really small, so it should not be a problem for the user. My porposal is to add the feature as is and wait for users' opinion: if some of them find it annoying and they want a setting for it I'll add it

That's a fear decision, do you have any screenshots of how the tooltips look?

src/Files.Uwp/DataModels/NavigationControlItems/DriveItem.cs
```diff
@@ -69,7 +69,11 @@ public ByteSize FreeSpace
         public ByteSize SpaceUsed
         {
             get => spaceUsed;
-            set => SetProperty(ref spaceUsed, value);
+            set
+            {
+                SetProperty(ref spaceUsed, value);
+                HoverDisplayText = (Path.Contains("?", StringComparison.Ordinal) ? Text : Path) + $"\n{"PropertiesDriveUsedSpace/Text".GetLocalized()} {spaceUsed.ToSizeString()}";
```
yaichenbaum, line 75: Instead of including this in the setter, is it better practice to listen for the property change notification and to call a dedicated method for setting the HoverDisplayText?

ferrariofilippo:
Here's the tooltip

Screenshot (34)

src/Files.Uwp/DataModels/NavigationControlItems/DriveItem.cs
ferrariofilippo: Sorry, I didn't know

yaichenbaum: I'm not sure myself, maybe @d2dyno1 can giveus some direction.

yaichenbaum:
Looks good! It might be a good idea to add `Path:` before the path so that it's consistent with the size option.

src/Files.Uwp/DataModels/NavigationControlItems/DriveItem.cs
```diff
@@ -143,16 +139,15 @@ public bool ShowStorageSense
         public DriveItem()
         {
             ItemType = NavigationControlItemType.CloudDrive;
+            PropertyChanged += SizeChanged;
```
d2dyno1, line 142: No need for hooking the PropertyChanged event here, instead make a separate method that is called from `SpaceUsed { set; }` and `Path { set; }` and update `HoverDisplayText` from there.

Attaching a PropertyChanged event like that and not deattaching it will result in a memory leak

src/Files.Uwp/DataModels/NavigationControlItems/DriveItem.cs
```diff
@@ -247,6 +238,13 @@ public async Task LoadDriveIcon()
             }
             Icon = await IconData.ToBitmapAsync();
         }
+
+        private void SizeChanged(object sender, PropertyChangedEventArgs e)
```
d2dyno1, line 242: (Additional note, you should adhere to the original name - in this case `PropertyChanged`)

But please remove that method as per my first comment

d2dyno1 (changes requested):

src/Files.Uwp/DataModels/NavigationControlItems/DriveItem.cs
ferrariofilippo: Ok, I didn't know it. Thank you for tips

d2dyno1:
Could you also resolve conflicts?

ferrariofilippo:
Wait I screwed up with merging.
I'm sorry but I'm still learning how to use Git

d2dyno1:
You could open a new fresh PR with included changes :)

ferrariofilippo:
I should have fixed

d2dyno1 (dismissed):

src/Files.App/DataModels/NavigationControlItems/DriveItem.cs
```diff
@@ -246,6 +241,12 @@ public async Task LoadDriveIcon()
             }
             Icon = await IconData.ToBitmapAsync();
         }
+
+        private void UpdateHoverDisplayText()
+        {
+            HoverDisplayText = $"{"PropertiesItemPath/Text".GetLocalizedResource()} {(Path.Contains("?", StringComparison.Ordinal) ? Text : Path)}" +
```
yaichenbaum, line 247: We should have a generic resource for path, renaming `PropertiesItemPath` to path is probably the best way to do this.

yair100 (changes requested):

yaichenbaum:
Does this close #1271?

ferrariofilippo:
> Does this close #1271?
> 
> This isn't a requirement but using this UI in the tooltips for drives would be neat image

- No, it doesn't close #1271 
- Ok, I'll work on that UI

ferrariofilippo:
Screenshot (39)

Is this ok?

src/Files.App/DataModels/NavigationControlItems/DriveItem.cs
ferrariofilippo: Do you mean that it already exists and I should use it, or that I should edit Resources files?

yaichenbaum: I think you can edit the existing resource to just be called `Path`.

yaichenbaum:
That's exactly what I meant 💯

0x5bfa (dismissed):
Really cool!

Details of changes:
- Use new method for a localization resource of drive free space and capacity
- Rename a l10n resource "PropertiesItemPath.Text" with "Path" for general use case.
- Adding new custom tooltip with drive path, visualized space by ProgressBar, localized free space text for NavigationViewItem located in the app sidebar control

0x5bfa:
@yair100 for me good. But I have a few questions about the variable naming. ViewModel doesn't know at first look what `Text` represents and also desn't know what `HoverDisplayText` means and where it used. I think they should be more friendly name.

src/Files.App/UserControls/SidebarControl.xaml
```diff
@@ -89,8 +89,26 @@
...
+                                    Text="{x:Bind Text, Mode=OneWay}"/>
+                                
+                                <ProgressBar
+                                    Maximum="{x:Bind MaxSpace.GigaBytes}"
+                                    Value="{x:Bind SpaceUsed.GigaBytes}"
+                                    VerticalAlignment="Stretch"
+                                    VerticalContentAlignment="Stretch" />
+                                
+                                <TextBlock 
+                                    Text="{x:Bind HoverDisplayText, Mode=OneWay}"/>
```
yaichenbaum, line 108:
```suggestion
                                    Text="{x:Bind ToolTipText, Mode=OneWay}"/>
```
@onein528

0x5bfa (dismissed):
LGTM.

yaichenbaum:
- Is there supposed to be a progress bar when hovering over OneDrive? It would be really great if we can do this!
- Hovering over the network folder displays a blank tooltip.
- "Path" is redundant now that we have the progress bar.

ferrariofilippo:
> * Is there supposed to be a progress bar when hovering over OneDrive? It would be really great if we can do this!

Like this?
drive
(With OneDrive it shows me my PC's free space, if you can confirm that I'll open an issue)
<br>
> * Hovering over the network folder displays a blank tooltip.

Can you send me a pic of what you see? (Here's what I see)
Network
<br>
> * "Path" is redundant now that we have the progress bar.

Do you mean reverting `Path: C:\...` to simply `C:\...` for folders in the `FavouriteSection`

yaichenbaum:
> (With OneDrive it shows me my PC's free space, if you can confirm that I'll open an issue)

That should be fixed to show OneDrive storage or alternatively it can be hidden when it can't be fetched.

> Do you mean reverting Path: C:\... to simply C:\... for folders in the FavouriteSection

Yes

ferrariofilippo:
>That should be fixed to show OneDrive storage or alternatively it can be hidden when it can't be fetched.

As this issue involves even `Properties` window, I'll open an issue.
<br>
New tooltip desing (`Drives` and `CloudDrives`, should I add it for `NetworkFolders` too?):
Drive2

yaichenbaum:
> New tooltip desing (Drives and CloudDrives, should I add it for NetworkFolders too?):

If it's possible to fetch the drive space, that would be really useful. Otherwise, progress bar should be hidden. (icon on the tooltip is redundant)

ferrariofilippo:
> (icon on the tooltip is redundant)

Do you mean for `NetworkFolders` only, or for all sidebar's item?

yaichenbaum:
All sidebar's item 👍

src/Files.App/Filesystem/Cloud/CloudDrivesManager.cs
```diff
@@ -48,6 +49,13 @@ public async Task UpdateDrivesAsync()
                     Path = provider.SyncFolder,
                     Type = DriveType.CloudDrive,
                 };
+                try
+                {
+                    cloudProviderItem.Root = await StorageFolder.GetFolderFromPathAsync(cloudProviderItem.Path);
+                    _ = App.Window.DispatcherQueue.EnqueueAsync(() => cloudProviderItem.UpdatePropertiesAsync());
+                }
+                catch (Exception) { }
```
yaichenbaum, line 57: Can you log the exception?

ferrariofilippo: no problem

yair100 (approved):
LGTM

Issue #9306, opened 2022-05-31
Some icons don't show up on sidebar


Evgeen:
### Description

Icons for Recycle Bin, Libraries, Drives, Network Drives, Network aren't showing on sidebar. 

Buttons itself work and open assigned folders correctly, but only icons are missing. Icons however show up next to tab name when I open either Recycle Bin or Network.

I think it could be related with a fact that I have system installed on drive S: and not C: but other icons show up there and for tabs as I said above. I checked and for other Windows 10 users these icons are showing up, but it's not on my system.

### Steps To Reproduce

1. Open app
2. Look to sidebar
3. Find out icons missing

### Expected behavior

Mentioned icons should be displayed

### Files Version

Version: 2.3.0.0 OS Version: 10.0.19044.1739

### Windows Version

Windows 10 21H2 19044.1739

### Relevant Assets/Logs

**Sceenshots**
https://i.imgur.com/maSlgER.jpeg https://i.imgur.com/inAwjeJ.jpeg

**debug_fulltrust.log**
```
2022-05-31 22:51:08.9494|INFO|OnConnectionRequestReceived|Argument: GetSelectedIconsFromDLL
2022-05-31 22:51:08.9744|INFO|OnConnectionRequestReceived|Argument: DetectCloudDrives
2022-05-31 22:51:08.9760|INFO|OnConnectionRequestReceived|Argument: ShellLibrary
2022-05-31 22:51:08.9766|INFO|OnConnectionRequestReceived|Argument: NetworkDriveOperation
2022-05-31 22:51:09.1592|INFO|OnConnectionRequestReceived|Argument: GetIconWithoutOverlay
2022-05-31 22:51:09.1946|WARN|StartSTATask|The system cannot find the file specified. (0x80070002)
	System.IO.FileNotFoundException: The system cannot find the file specified. (0x80070002)
   at Vanara.PInvoke.HRESULT.ThrowIfFailed(String message)
   at Vanara.PInvoke.Shell32.ShellUtil.GetShellItemForPath(String path)
   at Vanara.Windows.Shell.ShellItem.Open(String path)
   at Files.FullTrust.Helpers.ShellFolderExtensions.GetShellItemFromPathOrPidl(String pathOrPidl)
   at Files.FullTrust.Win32API.GetFileIconAndOverlay(String path, Int32 thumbnailSize, Boolean getOverlay, Boolean onlyGetOverlay)
[...]

PR #9748, opened 2022-08-17
Fix for hardcoded OS drives  

### LINKED ISSUE #9307: Bug: Customization page doesn't have the correct path
### What's the Problem?

In properties window under Customize tab in the path to icon file, it doesn't use drive letter where my system is actually installed which in my case is S: but the path that is there by default is `C:\Windows\System32\SHELL32.dll` and it results for nothing to show there until I select correct path (`S:\Windows\System32\SHELL32.dll`)

### Solution/Idea

This window in regular explorer has path `%SystemRoot%\System32\shell32.DLL`, so an idea is basically having this path in this window in Files as well, which will result for icons to show up when you open that window without a need to select icon file in case if you have system installed not on C: drive.

### Alternatives

I don't think there can be any alternatives to it

### Priorities

| Capability | Priority |
| :---------- | :------- |
| Make sure drive letter in path will match with actual system drive | Could |

### Files Version

Version: 2.3.0.0 OS Version: 10.0.19044.1739

### Windows Version

Windows 10 21H2 19044.1739

### Comments

_No response_

DavidPerikala:
**Resolved / Related Issues**
- Closes #9307
- Related #9306

**Details of Changes**
Add details of changes here.
- So far, we have assumed the OS directory to be "C:\"
- Fixed by using `%systemroot%` env path to get the OS root directory

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Uwp/Helpers/UIHelpers.cs
+++ b/src/Files.Uwp/Helpers/UIHelpers.cs
@@ -95,8 +95,7 @@ namespace Files.Uwp.Helpers
 
         private static async Task<IEnumerable<IconFileInfo>> LoadSidebarIconResources()
         {
-            string systemRoot = Environment.GetEnvironmentVariable("Systemroot");
-            string imageres = Path.Combine(systemRoot, "System32", "imageres.dll");
+            string imageres = Path.Combine(CommonPaths.SystemRootPath, "System32", "imageres.dll");
             var imageResList = await UIHelpers.LoadSelectedIconsAsync(imageres, new List<int>() {
                     Constants.ImageRes.RecycleBin,
                     Constants.ImageRes.NetworkDrives,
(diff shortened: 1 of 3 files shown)


src/Files.Uwp/Helpers/UIHelpers.cs
```diff
@@ -94,7 +95,8 @@ public static async Task<BitmapImage> GetIconResource(int index)
 
         private static async Task<IEnumerable<IconFileInfo>> LoadSidebarIconResources()
         {
-            const string imageres = @"C:\Windows\System32\imageres.dll";
+            string systemRoot = Environment.GetEnvironmentVariable("Systemroot");
```
yaichenbaum, line 98: Can you add this path to `CommonPaths`? https://github.com/files-community/Files/blob/main/src/Files.Uwp/Helpers/CommonPaths.cs

DavidPerikala: Done

yaichenbaum (approved):
LGTM

PR #9754, opened 2022-08-17
Added option to set current directory setup as the default

ferrariofilippo:
**Details of Changes**
- Toolbar: Added a Button that let you show/hide the "Tags" column of "Details Layout". When you click it, the layout is fixed

**Validation**
- Built and ran the app


src/Files.Uwp/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -421,6 +421,19 @@
 
                 <local:ColoredIcon BaseLayerGlyph="&#xF02B;" OverlayLayerGlyph="&#xF02C;" />
             </AppBarButton>
+            <AppBarButton
```
yaichenbaum, line 424: @gave92 did you want at toolbar button or is it just that the option to show the tag column should follow the setting to use individual preferences?

gave92: The problem with following the setting to use individual preferences is that a user who does not want to use tags has to hide the column on each and every folder or turn off that setting, which _also_ locks the layout mode.
I guess we should allow the user to set separately:
1. a default layout mode <- this we have
2. a default sort and group option
3. for details mode: default shown columns (and column sizes?)

For [3] I was thinking about adding an "Set as default" button to the right-click menu that you use to hide/show columns in details view.

gave92: Perhaps we can hold this PR off a bit and think about what we want. Ideally I'd like something that works for any column, not just tags (e.g hide "Size" column everywhere).

yaichenbaum: > for details mode: default shown columns (and column sizes?)

We might want to create a section in settings for this and link to it in the context menu. We'll need more thought on how to do column sizes though.

ferrariofilippo: What about a percentage for the size?
We can create an UI like:
------------------------------------------------------------------
Column Name (Label) | Shown (CheckBox) | Size % (TextBox)
------------------------------------------------------------------
When the user saves this settings, we check if %es add up to 100.
If the sum is, for example, 80%, we may split the remaining 20% among all columns.


Another way of doing this might be:
- showing the user an empty detail view
- letting him edit it as he want (removing/resizing columns)
- saving data from this view (Columns shown and sizes)

gave92: For a start I think we could just use the "real"UI, by adding a "Set as default" button:



The user does not have to go to settings and can easily see how the view  will look like.

yaichenbaum: I like that 👍

gave92: Ok, pushed the UI proposed above here. Should work with all columns. Does not save columns width.

yaichenbaum (approved):
LGTM

src/Files.Uwp/UserControls/InnerNavigationToolbar.xaml
yaichenbaum: Before merging, I'm going to look at displaying these options in settings as well.

yaichenbaum (approved):

yaichenbaum:
@ferrariofilippo @gave92 thank you!

PR #9766, opened 2022-08-19
Fix - Selecting a tag on the sidebar doesn't highlight it

### LINKED ISSUE #9760: Selecting a tag on the sidebar doesn't highlight it
### Description

When clicked it shows home is selected not the tag.
image

### Steps To Reproduce

_No response_

### Expected behavior

Tag should be highlighed on sidebar

### Files Version

2.3.7.0

### Windows Version

10.0.19043.1889

### Relevant Assets/Logs

*

ferrariofilippo:
**Resolved / Related Issues**
- Closes #9760 

**Details of Changes**
- Added an 'if' that checks if the selected item is a FileTagItem and if true sets it as the path to highlight

**Validation**
- Built and ran the app

**Screenshots**
Screenshot (26)

### DIFF
--- a/src/Files.Uwp/Views/MainPage.xaml.cs
+++ b/src/Files.Uwp/Views/MainPage.xaml.cs
@@ -177,10 +177,6 @@ namespace Files.Uwp.Views
             if (SidebarAdaptiveViewModel.PaneHolder != null)
             {
                 var paneArgs = e.NavigationArg as PaneNavigationArguments;
-                if (SidebarAdaptiveViewModel.SidebarSelectedItem is FileTagItem tag)
-                {
-                    paneArgs.LeftPaneNavPathParam = tag.Path;
-                }
                 SidebarAdaptiveViewModel.UpdateSidebarSelectedItemFromArgs(SidebarAdaptiveViewModel.PaneHolder.IsLeftPaneActive ?
                     paneArgs.LeftPaneNavPathParam : paneArgs.RightPaneNavPathParam);
                 UpdateStatusBarProperties();
@@ -330,7 +326,8 @@ namespace Files.Uwp.Views
                             IsSearchResultPage = true,
                             SearchPathParam = "Home".GetLocalized(),
                             SearchQuery = tagPath,
-                            AssociatedTabInstance = shp
+                            AssociatedTabInstance = shp,
+                            NavPathParam = tagPath
                         });
                     }
                     return;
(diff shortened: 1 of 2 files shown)


src/Files.Uwp/Views/MainPage.xaml.cs
```diff
@@ -177,6 +177,10 @@ public void TabItemContent_ContentChanged(object sender, TabItemArguments e)
             if (SidebarAdaptiveViewModel.PaneHolder != null)
             {
                 var paneArgs = e.NavigationArg as PaneNavigationArguments;
+                if (SidebarAdaptiveViewModel.SidebarSelectedItem is FileTagItem tag)
```
yaichenbaum, line 180: Should this be in `SidebarViewModel`?

ferrariofilippo: I tried putting this 'if' inside the ViewModel, but the problem wasn't there. Everything is fine until this method gets called at the end of all other invocations.
When you click on a FileTag, paneArgs still has "Home" as the directory searched, and the one to be highlighted. 
This happens because, as you can see in Issue's screenshot, clicking on a Tag results in a search in the "Home".
By checking if the selectedItem is a FileTagItem and, if true, updating the path, I fixed the problem

yaichenbaum: I think the preferred solution would be to change the pane args to match the tag.

ferrariofilippo: Ok, I'll try to fix it

ferrariofilippo: I should have fixed it
Tell me if that's ok

gave92:
Nice one, thanks! I just added a null check b/c when you do a normal search NavPathParam is null.

gave92 (approved):

src/Files.Uwp/Views/ModernShellPage.xaml.cs
```diff
@@ -669,10 +669,11 @@ private async void ItemDisplayFrame_Navigated(object sender, NavigationEventArgs
                 ContentPage.ResetItemOpacity();
             }
             var parameters = e.Parameter as NavigationArguments;
+            var isTagSearch = parameters.NavPathParam is not null && parameters.NavPathParam.StartsWith("tag:"); // #9760
```
yaichenbaum, line 672:
```suggestion
            var isTagSearch = parameters.NavPathParam is not null && parameters.NavPathParam.StartsWith("tag:");
```

yaichenbaum (approved):

PR #9769, opened 2022-08-19
Feature: Open full path in column layout when opening items from the sidebar 

### LINKED ISSUE #7589: Option to open complete path of favorite folder in "Columns" view
### What's the Problem?

I have selected "Layout" option as "Columns"
Currently when I open a folder that is added to favorite from the favorites section, It opens only that folder in the leftmost column like below.

image

I would like some option in settings to open the folder including its parents in four columns with the target folder in rightmost column like seen below

image

### Solution/Idea

Please can you add option in settings to make this feature available.

### Alternatives

You can make this default method to open folder when Column is the selected layout.

### Priorities

| Capability | Priority |
| :---------- | :------- |

### Files Version

Version: 2.1.0.0 OS Version: 10.0.22000.434

### Windows Version

Windows 11 21H2 OS Build 22000.434

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**
- Closes #7589

**Details of Changes**
- In Appearance Settings: Added the option "Open Complete Path in Column Layout"
- In Resources.resw (en-US): Added the title for this option
- In ColumnViewBrowser: When opening a Path, if the option is enabled, the whole path of the folder is displayed. To achieve this I split the destination path in its directories, and I navigate through them until I reach my destination.

**Validation**
- Built and ran the app


yaichenbaum:
@ferrariofilippo instead of a setting, I think an option in the right click context menu will be enough. Feel free to ping me before starting on a new issue and we can come up with a plan to avoid extra work 🙂

yaichenbaum:
@ferrariofilippo can you resolve the merge conflicts?

ferrariofilippo:
I had problems with the merge so I deleted old commits and pushed a new one.

ferrariofilippo:
@yair100 is it ok now?

src/Files.App/UserControls/SidebarControl.xaml.cs
```diff
@@ -324,6 +329,12 @@ private List<ContextMenuFlyoutItemViewModel> GetLocationItemMenuItems(INavigatio
                     ID = "ItemOverflow",
                     Tag = "ItemOverflow",
                     IsHidden = true,
+                },
+                new ContextMenuFlyoutItemViewModel()
```
yaichenbaum, line 333: Can you move this under the open in new tab option?

yaichenbaum:
I took some time to play around with this option and found a couple of issues with the UX. If we have an option to open in the column layout users would expect an option to open in other layout as well and there really is any reason to do this. We could work around this by reverting to your original implementation that added a setting for this, but I don't think it's worth adding another setting. I think we can solve these issues by removing the context menu option (apologies for requesting it) and automatically opening the full path whenever the clicked item is set to use column layout.

ferrariofilippo:
You're right about the fact that users will expect the feature for all layouts, so having it in the context menu might be confusing. I didn't think of it in first place. 
The reason I first placed this as a setting is that this feature might slow the workflow.
Now, after some use of the feature, I changed my mind: the advantage of the column layout is that it lets you interact with directories much faster than others (eg. If I have to navigate 4 folders backwards I can do that in just 1 click, instead of 4). Opening from a specific folder makes column layout half-useless.
As you proposed, I'd like the feature to be always active.

ferrariofilippo:
@yaichenbaum is this ok now?

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2825,4 +2825,7 @@ We use App Center to track which settings are being used, find bugs, and fix cra
   <data name="Lockscreen" xml:space="preserve">
     <value>Lockscreen</value>
   </data>
+  <data name="OpenCompletePath" xml:space="preserve">
+    <value>Open full path</value>
```
yaichenbaum, line 2829: Can this be removed?

src/Files.App/UserControls/SidebarControl.xaml.cs
```diff
@@ -9,6 +9,8 @@
 using Files.App.Helpers;
 using Files.App.Helpers.ContextFlyouts;
 using Files.App.ViewModels;
+using Files.App.Views;
+using Files.App.Views.LayoutModes;
```
yaichenbaum, line 13: Is this needed?

yaichenbaum:
> @yaichenbaum is this ok now?

I still need to test it on my end but it sounds like the behavior is good 👍

yaichenbaum:
@ferrariofilippo is this for all sidebar items or just for pinned favorites?

ferrariofilippo:
> @ferrariofilippo is this for all sidebar items or just for pinned favorites?

I implemented it as the default opening mode for ColumnLayout so it works for all sidebar items too.
Anyway, I've just tested it again and I noted a bug when you try to open a Drive (two equal columns are opened), so now I'll work on that

yaichenbaum:
Sorry for the delay reviewing this, I'm going to review some of the other PRs before coming back to this.

yaichenbaum:
@ferrariofilippo we're putting this feature request on hold as we get a better understanding of the use case and how it's handled on other platforms.

ferrariofilippo:
Ok

hishitetsu:
I investigated the behavior in macOS Finder.
When a directory is opened in column view, Finder in principle opens columns starting from the root. The exception, however, seems to be for some directories, it opens columns starting from that directory. e.g. Documents, Downloads, Applications, etc.

Other differences I have noticed: 
* The navigation history also records directory moves within column view.
* When a parent directory is selected, the columns of the child directories is dismissed.

yaira2:
> When a directory is opened in column view, Finder in principle opens columns starting from the root. The exception, however, seems to be for some directories, it opens columns starting from that directory. e.g. Documents, Downloads, Applications, etc.

We should probably do the same, any folder pinned to favorites will start from the directory and everything else from root.

> When a parent directory is selected, the columns of the child directories is dismissed.

Don't we do this?

hishitetsu:
> We should probably do the same, any folder pinned to favorites will start from the directory and everything else from root.

Yes, I think so and #11274 would need to be closed and implemented differently.

> Don't we do this?

Probably not.
The Finder implementation is simpler, but the Files implementation is more convenient.

yaira2:
I see what you're saying, I'm in favor of using the simpler implementation because it's pretty hard to maintain what we have today, but I agree that it's more convenient.

hishitetsu:
So I think we should leave it as it is. If we change it, we will definitely get requests to change it back.

Issue #6920, opened 2021-11-15
Layout options for folder content display


Hefaistos68:
### What's the Problem?

Current layout for folder content is very spacious, nice to the eye but little useful for frequent file operations. Currently the display seems to about 50% larger than the default windows explorer density.

### Solution/Idea

Offer something like most email programs offer: 3 different density settings, like comfortable, compact and tense? Or at least a line height setting.

### Alternatives

No alternatives, currently it seems to be a fixed setting.

### Priorities

Would be nice to have and improve usability

### Files Version

Version: 2.0.34.0 

### Windows Version

OS Version: 10.0.19042.1348

gave92:
Merging with #1940

Issue #6439, opened 2021-10-11
Doenst remove label when eject device


joseehilton147:
### Description

![image](https://user-images.githubusercontent.com/53986027/136794334-5b3a0511-4bbc-4513-a264-b693bb10aa0f.png)

This device (pen drive) was removed like 20 minutes ago, but the label stay on files, on normal explorer doenst show.

Edit: I need to physically remove the device to stop showing on files, when I click on eject option on device i receive a alert It's safe to remove but still show there.

### Steps To Reproduce

1. Eject a device
2. Open files.

### Expected behavior

remove the device from files

### Files Version

Versão: 2.0.13.0 OS Version: 10.0.22471.1000

### Windows Version

Windows 11 Pro Insider Preview 22471.1000

yaira2:
Is this still an issue in v2.4?

PR #9770, opened 2022-08-19
Fix unmount iso/virtual drives after eject

### LINKED ISSUE #6072: Ejecting ISOs does not unmount the virtual drive.
### Description

If you eject an ISO in Windows Explorer, it will also remove the virtual drive. If you do it in Files, it won't remove the virtual drive. If you look in Explorer after ejecting the ISO, you will see an empty virtual drive.

### Steps To Reproduce

1. Go to the home page with the drives.
2. Eject the ISO.
3. View the drives in Explorer.

### Expected behavior

The virtual drive should also be ejected along with the ISO.

### Files Version

v2.0.11.0

### Windows Version

Windows 11 Home 21H2 22000.176

### Relevant Assets

_No response_

DavidPerikala:
**Resolved / Related Issues**
- Closes #6072

**Details of Changes**
Add details of changes here.
- Fixed Unmounting ISO files to remove label after ejecting.
- I could see that the case for ejecting is same as inserting? How? I guess it should be the same as Removing, right? Or is there a reason for that? 
- I updated and tested it. It works. 

**Validation**
- [x] Built and ran the app


gave92:
I think it was like that because when you have a "real" CDrom drive, when you eject a CD you don't want to remove the drive from the list.

gave92:
Plus if you do this I think the empty drive still appears in windows explorer.

DavidPerikala:
> I think it was like that because when you have a "real" CDrom drive (or even a daemon tools-like drive), when you eject a CD you don't want to remove the drive from the list, but just update the label.

Oh okay. Now I understand. 



> Plus even if you do this I think the empty drive will still appear in windows explorer.

oops yes. I see it :/
Any hint, how we can proceed here?

gave92:
Uh good question. I guess we could:
- check what happens if we can invoke the "Eject" verb of the explorer context menu from Files
- check if dismounting the iso from powershell removes the drive from explorer (e.g. with `Dismount-DiskImage -ImagePath "E:\Windows10.iso"`)

gave92:
Just checked [2]:
Calling e.g. `Dismount-DiskImage -ImagePath "E:\Windows10.iso"` will also remove the drive from explorer.
Guess this issue is the same as #6439 (for removable drives), we do not Eject the device/ISO correctly.

gave92:
Update: pushed a change to eject using shell. Appears to fix both #6072 and #6439

DavidPerikala:
Wow. I was looking in a different implementation but great, Thank you @gave92

gave92:
Thanks! Does it work on your end as well then?
I'm not super happy with this solution, I was hoping to accomplish this in UWP without depending on file explorer.
So if you've found a different solution feel free to revert my changes.

DavidPerikala:
Yes I tried and it works. 
It's ok. I guess we can use this fix for now?. But I will try an implementation and send a new PR for that it that works.

DavidPerikala:
@gave92 Hold on! It doesn't remove the pendrive. so #6439 is still reproducible. Can you verify once?

gave92:
I can confirm that on my end removing a usb key works (I have only one with which to try though). I'm on windows 10 (if that matters),
Do you get any error message? Do you get the "Safe to remove Hardware" windows notification?

DavidPerikala:
I am on win 11 & So there's safe to remove hardware notification & no error. But the drive is still there on the UI. Once I remove it physically it's removed. Debug log doesn't show any error.

gave92:
Does the drive appear in Explorer after you eject it from Files?

DavidPerikala:
> Does the drive appear in Explorer after you eject it from Files (before you remove it physically)?

No, It's gone from the Explorer. (before removing physically)



> Does the drive disappear from FIles if you use the code from `main` (without this PR)?

No, it doesn't disappear from Files on the main branch

gave92:
> No, it doesn't disappear from Files on the main branch

Thanks for confirming. If this PR does break anything I think we can merge it as it solves at least #6920.

src/Files.App/ViewModels/ItemViewModel.cs
```diff
@@ -407,6 +407,8 @@ public ItemViewModel(FolderSettingsViewModel folderSettingsViewModel)
             UserSettingsService.OnSettingChangedEvent += UserSettingsService_OnSettingChangedEvent;
             FileTagsSettingsService.OnSettingImportedEvent += FileTagsSettingsService_OnSettingImportedEvent;
             FolderSizeProvider.SizeChanged += FolderSizeProvider_SizeChanged;
+
+            _ = InitializeConnectionAsync(); // fire and forget
```
d2dyno1, line 411: I don't think it should belong in the constructor

gave92: Copied from YourHomeViewModel. Moved outside the constructor now.

gave92 (approved):

PR #9785, opened 2022-08-21
Feature: Added an option to set default columns

yaichenbaum:
**Details of Changes**
Add details of changes here.
- Added support for setting which columns are shown by default
- Removed the option to use individual preferences and replaced it with an easier to understand option "Force these preferences on all directories".

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image

### DIFF
--- a/src/Files.Uwp/Views/SettingsPages/Preferences.xaml
+++ b/src/Files.Uwp/Views/SettingsPages/Preferences.xaml
@@ -1,399 +1,455 @@
 ﻿<Page
-    x:Class="Files.Uwp.SettingsPages.Preferences"
-    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:UserControls="using:Files.Uwp.UserControls"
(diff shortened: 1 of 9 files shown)


src/Files.Uwp/Views/SettingsPages/Preferences.xaml
```diff
@@ -1,399 +1,455 @@
...
+								Style="{StaticResource RightAlignedToggleSwitchStyle}" />
+						</local:SettingsBlockControl>
+					</StackPanel>
+				</local:SettingsBlockControl.ExpandableContent>
+			</local:SettingsBlockControl>
+
+			<!--  Default folder preferences  -->
+			<local:SettingsBlockControl Title="{helpers:ResourceString Name=DefaultFolderPreferences}" HorizontalAlignment="Stretch">
+				<local:SettingsBlockControl.Icon>
+					<FontIcon Glyph="" />
```
gave92, line 265: Is this intended?

yaichenbaum: Should be fixed, good catch!

gave92 (approved):

PR #9787, opened 2022-08-21
Added button to reset layout preferences in settings

### LINKED ISSUE #9784: Feature: Add option to reset layout preferences
Add support for resetting the database with layout preferences.

gave92:
**Resolved / Related Issues**
- Closes #9784

**Validation**
- [x] Built and ran the app


src/Files.Uwp/Views/SettingsPages/Preferences.xaml
```diff
@@ -1,455 +1,477 @@
...
+                                    FontFamily="{ThemeResource SymbolThemeFontFamily}"
+                                    FontSize="16"
+                                    Glyph="&#xE001;"
+                                    Visibility="{Binding IsLayoutResetCheckmarkVisible}" />
+                                <Button
+                                    AutomationProperties.Name="{helpers:ResourceString Name=ResetLayoutPreferences}"
+                                    Command="{Binding ResetLayoutPreferencesCommand}"
+                                    Content="{helpers:ResourceString Name=ResetLayoutPreferences}" />
+                            </StackPanel>
+                        </local:SettingsBlockControl>
```
yaichenbaum, line 345: What do you think about adding this right after the "force layout" option on line 279?

gave92: The description for that setting says "these options", as in "the following options". I think I'd be strange to see the reset button among those.
Plus reset buttons are usually at the bottom.

yaichenbaum: Perhaps it can be on the header of the expander? Might be more intuitive over there.

yaichenbaum: image

gave92: Why not, the header needs to be changed to just "Folder preferences" though.

gave92: But the it's a bit to easy to press by mistake. If put there we also need to add a confirmation step.

yaichenbaum: It would still be a separate button so I think that would prevent accidental presses, kind of like the option to copy the file version.

gave92: We could do like github and force users to type something to confirm reset, like:
`Type "files_is_best" to reset layout pref`

yaichenbaum: I actually find those prompts a little annoying 😂 
Either way, that sort of prompt or just a teaching tip with an additional confirm button should do the trick.

src/Files.Uwp/Strings/en-US/Resources.resw
```diff
@@ -2807,4 +2807,10 @@ We use App Center to track which settings are being used, find bugs, and fix cra
   <data name="DisplayTypeColumn" xml:space="preserve">
     <value>Display Type column</value>
   </data>
+  <data name="Reset" xml:space="preserve">
+    <value>Reset</value>
+  </data>
+  <data name="ResetLayoutPreferences" xml:space="preserve">
+    <value>You are about to reset all folder layout preferences</value>
```
yaichenbaum, line 2814:
```suggestion
    <value>Are you sure you want to reset all folder layout preferences?</value>
```

yaichenbaum:
This still has the potential to be confusing as users might think they are resetting the settings rather than the database, but I think it's worth trying this.

src/Files.Uwp/Views/SettingsPages/Preferences.xaml
```diff
@@ -1,455 +1,481 @@
...
+                        FontSize="16"
+                        Glyph="&#xE001;"
+                        Visibility="{Binding IsLayoutResetCheckmarkVisible}" />
+                    <Button
+                        x:Name="ResetLayoutPreferencesButton"
+                        AutomationProperties.Name="{helpers:ResourceString Name=Reset}"
+                        Command="{Binding ShowResetLayoutPreferencesTipCommand}"
+                        Content="{helpers:ResourceString Name=Reset}" />
+                    <muxc:TeachingTip
+                        Title="{helpers:ResourceString Name=Reset}"
```
yaichenbaum, line 285: Can you add a separate resource with the title "Are you sure?"

src/Files.Uwp/Views/SettingsPages/Preferences.xaml
```diff
@@ -1,455 +1,481 @@
...
+                    VerticalAlignment="Center"
+                    Orientation="Horizontal"
+                    Spacing="12">
+                    <FontIcon
+                        Width="16"
+                        VerticalAlignment="Center"
+                        FontFamily="{ThemeResource SymbolThemeFontFamily}"
+                        FontSize="16"
+                        Glyph="&#xE001;"
+                        Visibility="{Binding IsLayoutResetCheckmarkVisible}" />
```
yaichenbaum, line 278: What are your thoughts on adding a fade in/fade out animation? Happy to push a commit for this if you agree.

gave92: Why not, feel free to do it, I suck with animations.

yaichenbaum: Sounds good, here's hoping I don't forget 😂

yaichenbaum (approved):
LGTM

Issue #9704, opened 2022-08-11
Folder contents are hidden when folder is double clicked in Columns layout.


nathanrboyer:
### Description

Folder contents should be shown to the right regardless of if I single click or double click on a folder in Columns layout. I am used to double clicking from the Details layout, so I often double click a folder in Columns layout and think it is empty.

### Steps To Reproduce

1. Choose Columns Layout.
2. Single click on a folder.
3. Observe that folder contents appear in the next column.
4. Double click on a different folder.
5. Observe that nothing appears in the next column.

### Expected behavior

I expect the contents of the selected item to appear in the next column of Columns Layout regardless of how that item was selected.

### Files Version

Version: 2.3.3.0 OS Version: 10.0.22000.856

### Windows Version

Windows 11 Pro 21H2 22000.856

### Relevant Assets/Logs

[debug_fulltrust.log](https://github.com/files-community/Files/files/9311900/debug_fulltrust.log)
[debug.log](https://github.com/files-community/Files/files/9311901/debug.log)

yaichenbaum:
The column layout only requires a single click, can you confirm that this solves the issue?

yaichenbaum:
Changing this to a feature request as this is by design.

Issue #9610, opened 2022-07-26
Feature: Auto focus first item in the Add Item modal


jameslashmar:
**Requirements**
- Auto focus the first option in the New Item model so that pressing the enter key automatically creates a new folder

manuel-gibaut-h:
Hi there, I think I found a solution to the problem, by adding an option on `PreferenceSettingService` and changing both `ModernShellPage` and `ColumnShellPage` to check if the option is enabled and redirect the shortcut "Ctrl + Shift + N" directly to `CreateFileFromDialogResultType` from `UIFilesystemHelpers` class, for the other part I've managed to focus the first item of AddItem ListView to be focused so it should be possible to travel the List via arrow keys.

The only problem I found so far is adding the Setting/Preference text to all the Resources.resw files, and if possible I would like to know what steps to follow or if it's not necessary on my part to complete that part.

Thanks on advance

@yair100

yaichenbaum:
@z3r0tru5tk4l1 we shouldn't need a setting for this as it should be the default behavior 🙂 
For the future, when adding a new string, adding the string to the en-us resource file is enough, the other strings are generated on our end.

manuel-gibaut-h:
@yair100 Thanks for the quick answer, final question on how to continue, should I revert the changes made to the Setting/Preference and both `ShellTypes` as well as the Resource.resw change I made in the en-us folder? And only leave the autofocus on the firs item of the AddItem ListView or do I just create a PR for the changes made?

Thanks in advanced

yaichenbaum:
> should I revert the changes made to the Setting/Preference and both ShellTypes as well as the Resource.resw change I made in the en-us folder? And only leave the autofocus on the firs item of the AddItem ListView

That would be ideal 👍

PR #9819, opened 2022-08-26
Focus on new folder by default when using ctrl + shift + n (#9704)

z3r0tru5tk4l1:
**Resolved / Related Issues**
Focus on "Folder" when using shortcut "Ctrl + Shift + N"

- Closes [https://github.com/files-community/Files/issues/9610](https://github.com/files-community/Files/issues/9610)

**Details of Changes**
- Added Method on `AddItemDialog.xaml.cs` that sets Focus on first ListViewItem which is "Folder"
- Added Loaded event on `AddItemDialog.xaml` to execute Focus function.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.Uwp/Dialogs/AddItemDialog.xaml
+++ b/src/Files.Uwp/Dialogs/AddItemDialog.xaml
@@ -1,12 +1,12 @@
 ﻿<ContentDialog
     x:Class="Files.Uwp.Dialogs.AddItemDialog"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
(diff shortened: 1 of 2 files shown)


yaichenbaum:
@z3r0tru5tk4l1 can I suggest a different approach where you bind the selected index to a property in the view model (assuming a view model exists)?

z3r0tru5tk4l1:
@yaichenbaum perfect, I'll try to look into it, I'll try to have news as soon as possible

z3r0tru5tk4l1:
@yaichenbaum sorry for the delay, Not sure if this is what you wanted, but instead of using an index which could be changed, I used the `ViewModel` Declared on the Class and used it to fetch the item with the Header `"Folder"` which will receive focus.

I validated the changes by changing the search of `"Folder"` to `"File"` and both items were successfully focused when using the 'AddItemDialog'

yaichenbaum:
@z3r0tru5tk4l1 I meant to create a property in the ViewModel called `SelectedIndex` and you could bind the `ListView` to this property.
Instead of selecting based on the header, you can simply set the value to `0` and it'll automatically select whatever the first item is.

yaichenbaum:
Merging with #10042

PR #9857, opened 2022-08-30
Get rid of FullTrust for ContextMenu and AppLaunch handlers

hez2010:
Let's do it in a progressive way.

As a beginning, aiming at ContextMenu and AppLaunch.

Opened an issue for traking: #9858


yaichenbaum:
@hez2010 this is great! Have you noticed any performance improvements?

hez2010:
> Have you noticed any performance improvements?

The loading speed of context menu is faster, though not that obvious. Also it doesn't require async loading, so the context menu will be fully loaded before showing.

yaichenbaum:
> so the context menu will be fully loaded before showing.

I'll have to try this out but the current behavior of dispalying the menu before it finished loading was done on purpose since it there was often a two or three second delay.

hez2010:
> there was often a two or three second delay.

I also noticed it. I'm making changes to make the context menu be loaded in the background.

hez2010:
> I'm making changes to make the context menu be loaded in the background.

Done.

yaichenbaum (dismissed):
The performance improvements here are amazing!

hez2010:
Added a commit to cancel context menu loading eagerly to avoid duplicated entries in the context menu being added if user quickly right-clicks twice.

yaichenbaum (approved):

src/Files.App/Files.App.csproj
```diff
@@ -23,6 +23,7 @@
     <NoWin32Manifest>True</NoWin32Manifest>
     <Platforms>x86;x64;arm64</Platforms>
     <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
+    <LangVersion>10</LangVersion>
```
lukeblevins, line 26: Nit: Is there really a need to include this explicitly?

src/Files.App/BaseLayout.cs
```diff
@@ -606,11 +606,10 @@ public async void BaseContextFlyout_Opening(object sender, object e)
                 if (!InstanceViewModel.IsPageTypeSearchResults && !InstanceViewModel.IsPageTypeZipFolder)
                 {
                     var shellMenuItems = await ContextFlyoutItemHelper.GetBaseContextShellCommandsAsync(currentInstanceViewModel: InstanceViewModel, workingDir: ParentShellPageInstance.FilesystemViewModel.WorkingDirectory, shiftPressed: shiftPressed, showOpenMenu: false, shellContextMenuItemCancellationToken.Token);
-                    if (shellContextMenuItemCancellationToken.IsCancellationRequested)
```
hez2010, line 609: The old logic for detecting cancellation is invalid because it is always checking the new instanitated cts.

hez2010:
I think this PR is ready for merge.
@yaichenbaum @lukeblevins PTAL.

yaichenbaum (approved):

gave92:
It seems it does not show all the menu items, e.g with a zip file 7zip menu is not shown.
image

PR #9868, opened 2022-08-31
Fix localized strings appearing empty in Compatibility properties

gave92:
**Details of Changes**
Add details of changes here.
- Fix the GetLocalizedResource() method to return empty string instead of throwing exception when the string resource is not found. This was the behavior on UWP and some code relies on this.

**Validation**
- [x] Built and ran the app


yaichenbaum:
Can we log the exception in app center so that we know to fix it?

gave92:
There's no exception to log :)

yaichenbaum (approved):

PR #9872, opened 2022-08-31
Fix: Fixed appcenter reporting

yaichenbaum:
**Resolved / Related Issues**
Fixed an issue where app center wasn't copied to the build directory

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Files.App.csproj
+++ b/src/Files.App/Files.App.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net6.0-windows10.0.22000.0</TargetFramework>
     <Platform Condition=" '$(Platform)' == '' ">x86</Platform>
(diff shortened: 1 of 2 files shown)


src/Files.App/Files.App.csproj
```diff
@@ -101,7 +101,9 @@
     </None>
     <None Update="Assets\terminal\terminal.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </None>
+    </None>    
+    <Content Include="Resources\AppCenterKey.txt" />
+    <Content Include="Resources\BingMapsKey.txt" />    
     <None Update="Resources\PreviewPanePropertiesInformation.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
gave92, line 108: Could you check if CopyToOutputDirectory is also needed? I had a problem with Themes working when deploying locally but not when building the package due to missing this.

yaichenbaum: I guess we won't know until it's packaged, I'll generate a build of this branch to test with.

yaichenbaum: It worked as intended 👍

PR #9885, opened 2022-09-03
Icon cache and Task.Run tweaks

hez2010:
**Details of Changes**
Add details of changes here.
- Introduced a cache for icon and overlay
- Avoid using `Task.Run` for long-running tasks

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Shell/Win32API.cs
+++ b/src/Files.App/Shell/Win32API.cs
@@ -1,6 +1,8 @@
-﻿using Files.Shared;
+﻿using Files.App.Helpers;
+using Files.Shared;
 using Files.Shared.Extensions;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -8,6 +10,7 @@ using System.Drawing;
 using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using System.Text;
@@ -16,6 +19,8 @@ using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using Vanara.PInvoke;
+using Windows.ApplicationModel.AppService;
+using Windows.Foundation.Collections;
 using Windows.System;
 
 namespace Files.App.Shell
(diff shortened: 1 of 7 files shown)


src/Files.App/ViewModels/ItemViewModel.cs
```diff
@@ -1878,7 +1878,7 @@ private void WatchForDirectoryChanges(string path, CloudDriveSyncStatus syncStat
 
             if (aProcessQueueAction == null) // Only start one ProcessOperationQueue
             {
-                aProcessQueueAction = Task.Run(() => ProcessOperationQueue(watcherCTS.Token, hasSyncStatus));
+                aProcessQueueAction = Win32API.StartSTATask(() => ProcessOperationQueue(watcherCTS.Token, hasSyncStatus));
```
gave92, line 1881: Not a problem but my understanding is that a dedicated thread makes sense only if you have blocking & non cpu-bound work (most of the work here is short running or asynchronous).

If instead there's a need to specify the task is long running perhaps I'd use Task.Factory.StartNew + LongRunning option, I'd reserve StarSTATask for operations that really require a STA thread + OLE (context menu, shell thumbnails..)

hez2010: Done.

gave92: Thanks, unfortunately Task.factory.StartNew has some bad default parameters, I believe the correct call to be smth like (note expecially the TaskScheduler):

`Task.Factory.StartNew(someAction, CancellationToken.None, TaskCreationOptions.DenyChildAttach | TaskCreationOptions.LongRunning, TaskScheduler.Default)`

gave92: Note: my knowledge about this topic is a bit poor, so feel free to disregard these suggestions / wskt for other oponions ;)

src/Files.App/Shell/Win32API.cs
```diff
@@ -16,13 +19,75 @@
...
+                Priority = ThreadPriority.Normal
+            };
+            thread.SetApartmentState(ApartmentState.STA);
+            thread.Start();
+            return tcs.Task;
+        }
+
+        public static Task StartSTATask(Action action)
+        {
+            var tcs = new TaskCompletionSource();
```
yaichenbaum, line 63:
```suggestion
            var taskCompletionSource = new TaskCompletionSource();
```
Can you do the same for the other variables? It helps make the code more readable.

src/Files.App/Shell/Win32API.cs
```diff
@@ -319,37 +426,37 @@ public static IList<IconFileInfo> ExtractSelectedIconsFromDLL(string file, IList
...
             if (icoCnt == null)
             {
                 return null;
             }
+
             int count = icoCnt.DangerousGetHandle().ToInt32();
-            int maxIndex = count - 1;
-            if (maxIndex == 0)
+            if (count <= 0)
             {
```
yaichenbaum, line 438: The brackets can be removed for single line if statements.

src/Files.App/Shell/Win32API.cs
```diff
@@ -362,7 +469,34 @@ public static bool SetCustomDirectoryIcon(string folderPath, string iconFile, in
...
+                _iconAndOverlayCache[folderPath] = new();
+            }
+            return success;
+        }
+
+        public static async Task<bool> SetCustomFileIconAsync(string filePath, string iconFile, int iconIndex = 0)
+        {
+            var connection = await AppServiceConnectionHelper.Instance;
+            if (connection != null)
+            {
```
yaichenbaum, line 484:
```suggestion
            if (connection == null)
                        return;
```
Returning early can help make the code more readable 👍

yaichenbaum (approved):

src/Files.App/ViewModels/ItemViewModel.cs
hez2010: I just checked the document and did't see why we need `TaskCreationOptions.DenyChildAttach`.

src/Files.App/Shell/Win32API.cs
```diff
@@ -428,16 +428,10 @@ public static IList<IconFileInfo> ExtractIconsFromDLL(string file)
             var iconsList = new List<IconFileInfo>();
             using var currentProc = Process.GetCurrentProcess();
             using var icoCnt = Shell32.ExtractIcon(currentProc.Handle, file, -1);
-            if (icoCnt == null)
-            {
-                return null;
-            }
+            if (icoCnt == null) return null;
```
yaichenbaum, line 431: I think `return` should still be on the next line, just without the brackets.

hez2010: Done.

yaichenbaum (approved):
LGTM

src/Files.App/ViewModels/ItemViewModel.cs
gave92: It would have been only to make it equivalent to Task.Run, but not necessary.

gave92: PS the important point was TaskScheduler.Default btw. Not passing it risks running the Task on the ui thread / causing a deadlock.

hez2010: Will fix this in following PR.

Issue #9793, opened 2022-08-22
Don't clear the keywords in search bar after searching


verdaderoken:
### What's the Problem?

In Explorer, when you search for something in the search bar, it would not clear the keywords:
![image](https://user-images.githubusercontent.com/88200096/185940377-07003263-4f11-499c-a1c1-9fb10e1f5a42.png)

However in Files, the keywords are cleared:
![image](https://user-images.githubusercontent.com/88200096/185940103-dae0322d-c758-4eaf-ad7e-eee1841255b9.png)

### Solution/Idea

Don't clear the keywords in the search bar after searching, just like in File Explorer.

This is to make it convenient for the user to not re-type the whole word in case they need to refine, modify or correct some spelling.

If the user is done searching (assumed that the keywords are retained), they can click the "**X**" button (or alternatively, the "Back" button that will act as the close button to escape from "Search" and go back to the normal view
![image](https://user-images.githubusercontent.com/88200096/185941871-c65d2d6a-21a1-4641-88c1-a9d845338adc.png)

### Alternatives

Maybe add recent searches so by clicking the keyword will autofill the search bar.

> from Explorer
![image](https://user-images.githubusercontent.com/88200096/185941346-34de01cb-f51d-4494-a2f4-3a1e5f354295.png)

### Priorities

| Capability | Priority |
| :---------- | :------- |
| Keywords should be retained in the search bar | Must |
| Clicking 'X' button in search bar should exit search mode and go back to normal view | Should |
| Add recent searches so by clicking the keyword will autofill the search bar. | Could |

### Files Version

v2.3.9.0

### Windows Version

10.0.22622.575

### Comments

I've already issued this but I think it's not the way it was fixed in #9311

ferrariofilippo:
I don't think we can add this feature as the search bar is a Windows Control, and it doesn't have any event raised on 'X' click.
I checked even EventArgs passed to OnTextChanged but there is no way to know if that button has been clicked
<html>
<body>


Clicking 'X' button in search bar should exit search mode and go back to normal view | Should
-- | --



</body>
</html>

The only way to add this feature would be to create our own control, but to my mind it's not worthy. Tell me what you think.

Anyway I'll work on the other 2 features

verdaderoken:
Thank you, @ferrariofilippo ! I agree with you on that 2nd feature. Shame that they don't have a click event on that clear button. Many thanks for making the other features implemented. I'll close this issue now :)

PR #9886, opened 2022-09-03
Feature: Display recent searches when clicking on the search box

### LINKED ISSUE #8104: add search history to search box
### What's the Problem?

add search history to the search box and also a button to remove searches

### Solution/Idea

add search history to the search box and also a button to remove searches

### Alternatives

add search history to the search box and also a button to remove searches

### Priorities

| Capability | Priority |
| :---------- | :------- |

### Files Version

2.1.5

### Windows Version

w10 21h1 64 bits

ferrariofilippo:
**Resolved / Related Issues**
- Related #9793 (Part 1 and 3)
- closes #8104

**Details of Changes**
- The SearchBox is no more cleaned after a query is submitted
- When the SearchBox is focused but empty, it shows the list of latest queries submitted. I limited this number to 3 to guarantee better performances. (Should I remove the file icon from these entries?)

**Comments**
- I couldn't implement the second feature requested, as this depends on a Windows Control

**Validation**
- Built and ran the app

**Screenshots**
Screenshot (32)
Screenshot (31)


src/Files.App/UserControls/SearchBox.xaml
```diff
@@ -73,7 +73,8 @@
         Text="{x:Bind SearchBoxViewModel.Query, Mode=TwoWay}"
         TextBoxStyle="{StaticResource TextBoxStyle}"
         TextChanged="SearchRegion_TextChanged"
-        UpdateTextOnSelect="False">
+        UpdateTextOnSelect="False"
+        GotFocus="SearchRegion_GotFocus">
```
yaichenbaum, line 77: Can we bind this event directly to the view model?

src/Files.App/ViewModels/SearchBoxViewModel.cs
```diff
@@ -65,14 +71,54 @@ public void SearchRegion_TextChanged(AutoSuggestBox sender, AutoSuggestBoxTextCh
...
+                // If the element is already contained, update its position
+                oldQueries.Remove(oldQueries.FirstOrDefault(q => q.ItemNameRaw == e.QueryText));
+
+                oldQueries.Insert(0, new ListedItem
+                {
+                    ItemNameRaw = e.QueryText
+                });
+
+                // Limit to last 3 queries to improve performance
+                if (oldQueries.Count > 3)
```
yaichenbaum, line 96:
```suggestion
                if (oldQueries.Count > 5)
```

verdaderoken:
image
For me, I think it might be better if the icon for the old queries is something like these search history icons:
image

yaichenbaum:
>  (Should I remove the file icon from these entries?)

@ferrariofilippo what are your thoughts on using the icon in the screenshot from @verdaderoken?

ferrariofilippo:
Ok, I'll fix that

yaichenbaum:
Should just be a regular glyph `&#xE81C;`

yaichenbaum (dismissed):
Excellent improvement!

I'd like to wait for another review before merging.

ferrariofilippo:
Ok np

src/Files.App/UserControls/SearchBox.xaml
```diff
@@ -58,7 +58,7 @@
 					<FontIcon
 						x:Name="EmptyIconGlyph"
 						x:Load="{x:Bind NeedsPlaceholderGlyph, Mode=OneWay}"
-						Glyph="&#xE7C3;" />
+						Glyph="{x:Bind Glyph, Mode=OneWay}" />
```
yaichenbaum, line 61: Is it possible to use a converter here? It would be nice to set the glyphs directly from the xaml. Alternatively, a comment with the two possible glyphs would also work.

src/Files.App/Converters/BoolToGlyph.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Files.App.Converters
+{
+	public class RecentFilesBoolToGlyph
+	{
+		public static object Convert(bool value)
+		{
+			return value ? "\xE81C" : "\xE7C3";
```
yaichenbaum, line 7: I should have specified, the values for the converter should be in the xaml file as a resource. It might be easier to just add a comment to the xaml file and skip using a converter all together.

src/Files.App/Converters/BoolToGlyph.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Files.App.Converters
+{
+	public class RecentFilesBoolToGlyph
```
d2dyno1, line 3: Please rename the file as well

d2dyno1 (changes requested):

src/Files.App/Converters/BoolToGlyph.cs
d2dyno1: Honestly, I'd make it a generic glyph converter where the converter parameter is the glyph:
`<glyphIfTrue>|<glyphIfFalse>`

ferrariofilippo: Ok, so I'll change this and I'll add a comment in the xaml to explian the two glyphs

yaichenbaum: > Honestly, I'd make it a generic glyph converter where the converter parameter is the glyph:

That's what I originally intended, thought we had a converter for this already.

ferrariofilippo: So, should I use the built-in converter (like in the picture) or should I add a brand new converter?

yaichenbaum: The built in one looks good, that's what I had in mind.

src/Files.App/Filesystem/ListedItem.cs
```diff
@@ -65,6 +65,8 @@ public bool NeedsPlaceholderGlyph
             set => SetProperty(ref needsPlaceholderGlyph, value);
         }
 
+        public bool IsRecentSearch { get; set; } = false;
```
yaichenbaum, line 68: Does this belong in ListedItem?

ferrariofilippo: Yes, I added it to differentiate recent searches from other `ListedItem`s (in order to show the correct glyph).

yaichenbaum: This is just me wondering but I don't think ListedItem is supposed to be coupled to search or have any knowledge of search.

ferrariofilippo: Do you have anything in mind to solve this problem?
To my mind, we could add a new class (called something like `ListedSuggestion`) that inherits from `ListedItem` and expands it with the `IsSearch` property.

yaichenbaum: Adding @d2dyno1

d2dyno1: Why do we even need to use/inherit from `ListedItem`? Isn't a search item a name and an icon?

ferrariofilippo: @d2dyno1

d2dyno1 (approved):

yaichenbaum (approved):

PR #9890, opened 2022-09-04
Fix performance issue with WinUI 3

hez2010:
It turns out that `ResourceMap` is incredibly slow in a WinUI 3 app. On my computer it needs almost 15ms for a single `TryGetValue` call.

Add a cache to fix the performance issue.


yaichenbaum (approved):
LGTM

Would it be possible to add a code comment explaining this change?

hez2010:
> Would it be possible to add a code comment explaining this change?

Yes. Will do it later.

PR #9944, opened 2022-09-07
Feature: ReadyToRun

yaichenbaum:
Added PublishReadyToRun

Do they need to be added to other PropertyGroups or is it good as is?

**Screenshots (optional)**
Add screenshots here.


hez2010:
We should hold this PR until WASDK 1.2 ships il trimming support. Otherwise it would result in an extremely huge binary.

yaichenbaum:
@hez2010 the package size is about 40mb larger than before I think it's okay to move forward with this considering we have a number of other things to do before releasing to stable.

hez2010 (approved):
LGTM

Issue #9858, opened 2022-08-30
Removal of FullTrust handlers


hez2010:
We are migrating to WindowsAppSDK, thus there's no longer need for fulltrust process for interop. Using this issue for tracking work-items of fulltrust handlers removal.

- [x] ApplicationLaunchHandler
- [x] ContextMenuHandler
- [ ] DesktopWallpaperHandler
- [ ] DriveHandler
- [ ] FileOperationsHandler
- [ ] FileTagsHandler
- [ ] IMessageHandler
- [ ] InstallOperationsHandler
- [ ] LibrariesHandler
- [ ] NetworkDrivesHandler
- [ ] QuickLookHandler
- [ ] RecentItemsHandler
- [ ] RecycleBinHandler
- [ ] Win32MessageHandler

lukeblevins:
@hez2010 I can take care of removing `RecentItemsHandler` once #9857 is merged because we'd want to have baked in support for a number of service implementations capable of listing/aggregating recent items from a multitude of sources down the road.

@yair100 FYI

yaichenbaum:
@lukeblevins I'm wondering if @jiejasonliu might want to do the `RecentItemsHandler` since they were just working on this feature a couple months ago.

lukeblevins:
@yair100 @jiejasonliu Please let me know.

jiejasonliu:
@lukeblevins @yair100 I'm probably not going to be able to get to this anytime soon due to school. If you can get around doing this in the next month or so, I encourage it.

I can offer to be a primary/secondary reviewer to any pull requests related to it and also feel free to DM me of any questions, I'm on the Files discord server (Jason#9955).

lukeblevins:
I'm almost done with the DriveHandler, so that one shouldn't be assigned to anyone else. @yair100

itsWindows11:
What is meant by removing here exactly?

itsWindows11:
I'm currently working on FileOperationsHandler but I have a question, do I migrate the methods and additional stuff from the FilePermissions and FileSystemAccessRule to the ones in Files.App?

lukeblevins:
@yair100 This can probably be pinned again.

I am working on the RecycleBin handler/watcher so that one can be assigned to me.

itsWindows11:
I can take a look at removing LibrariesHandler, probably not NetworkDrivesHandler because I don't have a network drive to test on.

yaichenbaum:
Love the enthusiasm! 

[...]

PR #9952, opened 2022-09-08
Removed InstallOperationsHandler

puppetsw:
**Resolved / Related Issues**
- Related #9858

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.


hez2010 (changes requested):
It's better to use `await process.WaitForExitAsync()` instead of `process.WaitForExit()`, so that the installation can be proceeded asynchronously.

puppetsw:
I’ll update shortly. Thanks!

src/Files.App/Shell/Win32API.cs
```diff
@@ -786,28 +786,17 @@ public static void OpenFolderInExistingShellWindow(string folderPath)
...
-            }
+            var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(30 * 1000));
+            using Process process = new Process();
+            process.StartInfo.FileName = "InfDefaultInstall.exe";
+            process.StartInfo.Verb = "runas";
+            process.StartInfo.UseShellExecute = true;
+            process.StartInfo.CreateNoWindow = true;
+            process.StartInfo.Arguments = $"{filePath}";
+            process.Start();
+            await process.WaitForExitAsync(cts.Token);
```
hez2010, line 799: Where is the boolean result going? In case that user cancels the UAC, even if we don't actually read the result of the method, we still need to surround the code in a try-catch block and catches `Win32Exception`.

hez2010 (changes requested):

hez2010 (dismissed):
Thank you. LGTM.

PR #9956, opened 2022-09-08
Feature: Hide minimize/maximize buttons on the properties window

### LINKED ISSUE #9931: Feature:  Hide minimize/maximize buttons on the properties window
**Requirements**
- Minimize and maximize buttons are hidden on the properties windows

ferrariofilippo:
**Resolved / Related Issues**
- Closes #9931 

**Changes**
- Now minimize and maximize buttons are hidden.
- Now the properties window can't be resized by the user

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Screenshot (40)

### DIFF
--- a/src/Files.App/Helpers/FilePropertiesHelpers.cs
+++ b/src/Files.App/Helpers/FilePropertiesHelpers.cs
@@ -59,8 +59,7 @@ namespace Files.App.Helpers
                     var propertiesWindow = new WinUIEx.WindowEx()
                     {
                         IsMaximizable = false,
-                        IsMinimizable = false,
-                        IsResizable = false
+                        IsMinimizable = false
                     };
                     var appWindow = propertiesWindow.AppWindow;
 
--- a/src/Files.App/Helpers/FilePropertiesHelpers.cs
+++ b/src/Files.App/Helpers/FilePropertiesHelpers.cs
@@ -1,17 +1,18 @@
 using Files.App.Dialogs;
-using Files.App.Views;
 using Files.App.Extensions;
+using Files.App.Views;
+using Microsoft.UI;
+using Microsoft.UI.Windowing;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Media.Animation;
 using System;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


yaichenbaum:
> Now the properties window can't be resized by the user

The user should still be able to resize the window

ferrariofilippo:
Ok, no problem

src/Files.App/Helpers/FilePropertiesHelpers.cs
```diff
@@ -22,35 +23,25 @@ public static async void ShowProperties(IShellPage associatedInstance)
...
+                var path = System.IO.Path.GetPathRoot(associatedInstance.FilesystemViewModel.CurrentFolder.ItemPath);
+                if (path is not null && path.Equals(associatedInstance.FilesystemViewModel.CurrentFolder.ItemPath, StringComparison.OrdinalIgnoreCase))
                     await OpenPropertiesWindowAsync(associatedInstance.FilesystemViewModel.CurrentFolder, associatedInstance);
-                }
                 else
-                {
                     await OpenPropertiesWindowAsync(App.DrivesManager.Drives
-                        .SingleOrDefault(x => x.Path.Equals(associatedInstance.FilesystemViewModel.CurrentFolder.ItemPath)), associatedInstance);
-                }
+                        .Single(x => x.Path.Equals(associatedInstance.FilesystemViewModel.CurrentFolder.ItemPath)), associatedInstance);
```
yaichenbaum, line 37: Why isn't this `SingleOrDefault`?

ferrariofilippo: The difference is that, by using `Single`, an exception is thrown if no item satisfies the conditions (`SingleOrDefault` would return `default`).
In this case, I replaced `SingleOrDefault` with `Single` because the condition is always satisfied by exactly one element.
Let me explain: the only way to enter this branch is by requesting the properties of a drive. To do so, you have to select it, so that drive must exist

yaichenbaum: Sounds reasonable

yaichenbaum (approved):
LGTM

Issue #9780, opened 2022-08-21
WinUI 3 progress


yaira2:
- [ ] #9929
- [ ] #10039
- [ ] #10422

verdaderoken:
- [ ] Title bar bug where another title bar shows above the tabs whenever a folder is opened or a tab is closed
![image](https://user-images.githubusercontent.com/88200096/187096333-60d7073c-6021-4de3-811d-ebf1a046bfdc.png)

https://user-images.githubusercontent.com/88200096/187096389-f0e11472-0c4d-4fff-9c7a-b038d522aee6.mp4

BanCrash:
Screenshot of: Context menu appears partially outside when right clicking an option on sidebar 

![imagen](https://user-images.githubusercontent.com/11770760/187232722-e7f908fa-720e-4d77-96f0-5a8f0c1fa2ee.png)

Josh-65:
> Screenshot of: Context menu appears partially outside when right clicking an option on sidebar
> 
> ![imagen](https://user-images.githubusercontent.com/11770760/187232722-e7f908fa-720e-4d77-96f0-5a8f0c1fa2ee.png)

That has always happend. i think it known on the WinUI github but they havent dont anything for it.

yaichenbaum:
Are you sure? I've only seen this behavior with the wasdk build.

Josh-65:
![image](https://user-images.githubusercontent.com/26977075/187310516-050cab79-bfb9-4c29-ab98-a8fbf39622d0.png)
![image](https://user-images.githubusercontent.com/26977075/187310590-6318f173-58ba-4eeb-bffe-0b0392ca89ce.png)


yep at least for the sidebar and trying it on the right of the screen has it show no were near the cursor

MarcAnt01:
![image](https://user-images.githubusercontent.com/43090263/188173995-7837b46d-7e18-4b0e-9f63-759c4351fd3f.png)
The preview feature seems to be broken

yaichenbaum:
>  Merge failure for shared merged PRI file: error 0x80070490. Fault in my computer?

@gave92 is this still happening?

gave92:
I think it was solved by removing arm build from the package.

Josh-65:
Tooltips like the status center one doesnt go away on its own or with the X button
Settings backdrop doesnt change the title bar
![image](https://user-images.githubusercontent.com/26977075/188321392-67e58f3c-1c91-43ed-ba97-9462434d0587.png)

[...]

PR #9966, opened 2022-09-09
Remove WindowDecorationsHelper, fix theme in properties window

### LINKED ISSUE #9911: Bug: Title bar close/minimize/maximize buttons color don't match the app theme
**Requirements**
- Title bar buttons should match the app theme rather than the system theme

gave92:
**Resolved / Related Issues**
- For #9780 ("Window Decoration api isn't working on the properties windows (we should remove the relevant code before release)")
- Closes #9911

**Changes**
- Removed AppWindowTitleBar.IsCustomizationSupported() check, not necessary as of winappsdk 1.2 (titlebar API supported down to 1809)
- Removed WindowDecorationsHelper, uwp only + same reason as above
- Removed code for custom titlebar close button (it was commented), we now use the standard close button
- Fix wrong background in properties window when theme changes
- Fix crash when clicking "share", remove call to GetForCurrentView() (uwp only)
- Restore code for changing titlebar buttons on theme change
- Improve positioning of properties window (do not go out of screen bounds)

**Validation**
- [x] Built and ran the app


src/Files.App/Helpers/FilePropertiesHelpers.cs
```diff
@@ -47,65 +47,51 @@ public static async Task OpenPropertiesWindowAsync(object item, IShellPage assoc
 
             if (ApiInformation.IsApiContractPresent("Windows.Foundation.UniversalApiContract", 8))
```
yaichenbaum, line 48: Can this if statement be removed? I don't think there is a scenario where this is not the case.

src/Files.App/UWPToWinAppSDKUpgradeHelpers.cs
```diff
@@ -23,6 +23,9 @@ interface IDataTransferManagerInterop
 
     public static class InteropHelpers
     {
+        public static readonly Guid DataTransferManagerInteropIID =
+            new Guid(0xa5caee9b, 0x8708, 0x49d1, 0x8d, 0x36, 0x67, 0xd2, 0x5a, 0x8d, 0xa0, 0x0c);
```
yaichenbaum, line 27: What does this do?

src/Files.App/Helpers/FilePropertiesHelpers.cs
gave92: Min version is 17763 which has UniversalApiContract 7

src/Files.App/UWPToWinAppSDKUpgradeHelpers.cs
gave92: Heck if I know :) Comes from the sample (it's the COM Guid of the DataTransferManager interface or something)

src/Files.App/Helpers/FilePropertiesHelpers.cs
yaichenbaum: I thought we were using 1809. We can keep this then 👍

gave92: Yeah 1809 == 10.0.17763.0

yaichenbaum:
> Restore code for changing titlebar buttons on theme change

@gave92 this seems to help for the main window, but the issue still exists with the property windows.

gave92:
Should be good now.

yaichenbaum (approved):
LGTM

Fahhd1400:
دمج طلب السحب

PR #9973, opened 2022-09-10
Fixed: Properties window opens behind main app window

### LINKED ISSUE #9926: Bug: Properties window opens behind main app window
**Steps to reproduce**
1. Right click file or folder
2. Click on properties
3. Observe that the properties window opens behind the main app window

**Requirements**
- Property window should be opened in the foreground

**Version information**
Version: 2.3.26.0
OS Version: 10.0.25193.1000

ferrariofilippo:
**Resolved / Related Issues**
- Closes #9926 

**Comments**
- The root of the problem was that, after the properties windows was opened, the focus returned to the main window because of the closing of the flyout context menu. I fixed the bug opening Properties after the menu gets closed
- If everything is ok, I'll proceed with code cleanup

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/IBaseLayout.cs
+++ b/src/Files.App/IBaseLayout.cs
@@ -25,5 +25,6 @@ namespace Files.App
         public SelectedItemsPropertiesViewModel SelectedItemsPropertiesViewModel { get; }
         public DirectoryPropertiesViewModel DirectoryPropertiesViewModel { get; }
         public BaseLayoutCommandsViewModel CommandsViewModel { get; }
+        public Microsoft.UI.Xaml.Controls.CommandBarFlyout ItemContextMenuFlyout { get; set; }
     }
 }
--- a/src/Files.App/IBaseLayout.cs
+++ b/src/Files.App/IBaseLayout.cs
@@ -1,6 +1,7 @@
 using Files.App.Filesystem;
 using Files.App.Interacts;
 using Files.App.ViewModels;
+using Microsoft.UI.Xaml.Controls;
 using System;
 using System.Collections.Generic;
 
@@ -25,6 +26,6 @@ namespace Files.App
         public SelectedItemsPropertiesViewModel SelectedItemsPropertiesViewModel { get; }
         public DirectoryPropertiesViewModel DirectoryPropertiesViewModel { get; }
         public BaseLayoutCommandsViewModel CommandsViewModel { get; }
-        public Microsoft.UI.Xaml.Controls.CommandBarFlyout ItemContextMenuFlyout { get; set; }
+        public CommandBarFlyout ItemContextMenuFlyout { get; set; }
     }
 }
--- a/src/Files.App/Interacts/BaseLayoutCommandImplementationModel.cs
+++ b/src/Files.App/Interacts/BaseLayoutCommandImplementationModel.cs
@@ -194,11 +194,17 @@ namespace Files.App.Interacts
 
         public virtual void ShowFolderProperties(RoutedEventArgs e)
         {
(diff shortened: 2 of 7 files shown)


gave92:
There is also a simpler alternative:
```cs
// Workaround for #9926
propertiesWindow.IsAlwaysOnTop = true;
propertiesWindow.DispatcherQueue.CreateTimer().Debounce(() =>
{
    propertiesWindow.Activate();
    propertiesWindow.IsAlwaysOnTop = false;
}, TimeSpan.FromSeconds(1));
```
Adding the above in OpenPropertiesWindowAsync() (@ line 82) appears to be enough to fix the issue.

ferrariofilippo:
> To be noted that there is also a simpler alternative:
> 
> ```cs
> // Workaround for #9926
> propertiesWindow.IsAlwaysOnTop = true;
> propertiesWindow.DispatcherQueue.CreateTimer().Debounce(() =>
> {
>     propertiesWindow.Activate();
>     propertiesWindow.IsAlwaysOnTop = false;
> }, TimeSpan.FromSeconds(1));
> ```
> 
> Adding the above in OpenPropertiesWindowAsync() (@ line 82) appears to be enough to fix the issue. It's a bit hacky though.

I took into account a solution like this, but there are two drawbacks: if, for any reason, the flyout context menu takes more than a second to close, the bug would remain unfixed (it would be a one-in-a-million case). The second problem is that the program might seem not optimized or inefficient because of that delay. I know it's only a second, but I'm obsessed with optimization.

gave92:
For "optimization" the simple solution is actually better since you don't wait for the menu to close before opening the window. I agree that it might not cover all cases (I'm not a fan of debounce times).

yaichenbaum:
Using the hwnd, can we force the new window to the foreground?

gave92:
With my solution I set it always on top for 1 second. It appears in front but it's not active. Not sure if we can force it to be active.

ferrariofilippo:
> Using the hwnd, can we force the new window to the foreground?

You can't really do that the way we'd want to.
Setting `IsAlwaysOnTop` to true means that the Properties Window can't be hidden by any other window (I don't think we want this).
Forcing the Prop. Window to be shown above others (using `BringToFront()` or `MoveInZOrderToTop()`) would be useless, because, as the context menu closes, the main window will be brought on top of others.
Mine and @gave92 solutions are the only viable.

src/Files.App/IBaseLayout.cs
```diff
@@ -25,5 +25,6 @@ public interface IBaseLayout : IDisposable
         public SelectedItemsPropertiesViewModel SelectedItemsPropertiesViewModel { get; }
         public DirectoryPropertiesViewModel DirectoryPropertiesViewModel { get; }
         public BaseLayoutCommandsViewModel CommandsViewModel { get; }
+        public Microsoft.UI.Xaml.Controls.CommandBarFlyout ItemContextMenuFlyout { get; set; }
```
yaichenbaum, line 28: Can this be simplified?

yaichenbaum (approved):

yaichenbaum (approved):
LGTM

yaichenbaum:
@ferrariofilippo does this work for the drives widget as well?

ferrariofilippo:
> @ferrariofilippo does this work for the drives widget as well?

Yes, it does

gave92:
@ferrariofilippo solution looks good to me. I just find it annoying that it has to be done in each place we have a context menu and each have slightly different code, but no biggie.

yaichenbaum:
Likewise but this is still a good solution

PR #9977, opened 2022-09-11
Feature: Added setting to select files and folders on mouse hover

### LINKED ISSUE #3682: Feature: Add setting to select files and folders when hovering over them with the mouse
**Requirements**
- Add setting to preferences setting page "Select files and folders when hovering over them with the mouse"
- Setting should be turned off by default

ferrariofilippo:
**Resolved / Related Issues**
- Closes #3682 

**Validation**
- [x] Built and ran the app


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2825,4 +2825,7 @@ We use App Center to track which settings are being used, find bugs, and fix cra
   <data name="BehaviorForOpeningFilesAndFolders" xml:space="preserve">
     <value>Behavior for opening files and folders</value>
   </data>
+  <data name="SelectFilesAndFoldersOnHovering" xml:space="preserve">
+	<value>Select files and folders hovering them</value>
```
yaichenbaum, line 2829: Tab spacing looks off, can you fix this?

src/Files.App/Views/LayoutModes/ColumnViewBase.xaml
```diff
@@ -192,6 +192,8 @@
 								CornerRadius="{StaticResource ControlCornerRadius}"
 								IsRightTapEnabled="True"
 								Loaded="Grid_Loaded"
```
yaichenbaum, line 194: Spacing should be set to tab spaces.

src/Files.App/Views/SettingsPages/Preferences.xaml
```diff
@@ -162,7 +162,14 @@
 								IsOn="{Binding ListAndSortDirectoriesAlongsideFiles, Mode=TwoWay}"
 								Style="{StaticResource RightAlignedToggleSwitchStyle}" />
 						</local:SettingsBlockControl>
-					</StackPanel>
+
+                        <local:SettingsBlockControl Title="{helpers:ResourceString Name=SelectFilesAndFoldersOnHovering}" HorizontalAlignment="Stretch">
```
yaichenbaum, line 166:
```suggestion
                        <local:SettingsBlockControl Title="{helpers:ResourceString Name=SelectFilesAndFoldersOnHover}" HorizontalAlignment="Stretch">
```

src/Files.App/Views/SettingsPages/Preferences.xaml
```diff
@@ -162,7 +162,14 @@
...
+                            <ToggleSwitch
+								AutomationProperties.Name="{helpers:ResourceString Name=SelectFilesAndFoldersOnHovering}"
+								IsOn="{Binding SelectOnHover, Mode=TwoWay}"
```
yaichenbaum, line 169:
```suggestion
								IsOn="{Binding SelectFilesOnHover, Mode=TwoWay}"
```

yaichenbaum:
Left some comments regarding spaces and property names but code looks code.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2825,4 +2825,7 @@ We use App Center to track which settings are being used, find bugs, and fix cra
   <data name="BehaviorForOpeningFilesAndFolders" xml:space="preserve">
     <value>Behavior for opening files and folders</value>
   </data>
+  <data name="SelectFilesAndFoldersOnHover" xml:space="preserve">
+    <value>Select files and folders hovering them</value>
```
yaichenbaum, line 2829:
```suggestion
    <value>Select files and folders when hovering over them</value>
```

yaichenbaum (dismissed):
LGTM

I did notice some issues where the items weren't selected but let's wait for the update ListView styles from WinUI before making this an issue.

yaichenbaum:
@gave92 do you want to look this over before I merge it?

gave92:
Ok, i should be able to go over it later today

src/Files.App/Views/SettingsPages/Preferences.xaml
```diff
@@ -162,7 +162,14 @@
 								IsOn="{Binding ListAndSortDirectoriesAlongsideFiles, Mode=TwoWay}"
 								Style="{StaticResource RightAlignedToggleSwitchStyle}" />
 						</local:SettingsBlockControl>
-					</StackPanel>
+
+                        <local:SettingsBlockControl Title="{helpers:ResourceString Name=SelectFilesAndFoldersOnHover}" HorizontalAlignment="Stretch">
```
gave92, line 166: I'd move this setting to the "Click to open settings" section below (perhaps changing the section name to "Behavior for opening and selecting files and folders"). @yair100?

src/Files.App/BaseLayout.cs
```diff
@@ -760,15 +714,16 @@ private void AddShellItemsToMenu(List<ContextMenuFlyoutItemViewModel> shellMenuI
             {
                 var openWith = contextMenuFlyout.SecondaryCommands.FirstOrDefault(x => x is AppBarButton abb && (abb.Tag as string) == "OpenWith") as AppBarButton;
                 var flyout = openWithOverflow.Flyout as MenuFlyout;
-                flyout.Items.Clear();
+
+                flyout!.Items.Clear();
```
gave92, line 718: I'm not convinced we should be using "!" so much just to get rid of some nullable warnings. E.g. this one could have been possibly better solved by:
```cs
var flyout = (MenuFlyout)openWithOverflow.Flyout;
flyout.Items.Clear();
```

src/Files.App/BaseLayout.cs
```diff
@@ -527,15 +497,15 @@ public void SetSelectedItemsOnNavigation()
...
 
-        private async void FolderSettings_GroupOptionPreferenceUpdated(object sender, GroupOption e)
+        private async void FolderSettings_GroupOptionPreferenceUpdated(object? sender, GroupOption e)
         {
             // Two or more of these running at the same time will cause a crash, so cancel the previous one before beginning
             groupingCancellationToken?.Cancel();
             groupingCancellationToken = new CancellationTokenSource();
             var token = groupingCancellationToken.Token;
-            await ParentShellPageInstance.FilesystemViewModel.GroupOptionsUpdated(token);
+            await ParentShellPageInstance!.FilesystemViewModel.GroupOptionsUpdated(token);
```
gave92, line 508: Similar to the comment below: is it correct to use "!" to get remove the nullable warning here? Perhaps we should add a null check instead or maybe ParentShellPageInstance should not be nullable in the first place.

src/Files.App/BaseLayout.cs
ferrariofilippo: Ok, I'll change that

src/Files.App/BaseLayout.cs
ferrariofilippo: I setted `ParentShellPageInstance` as nullable because the default value assigned to it is null (line @79).
Given this I think the null check is the best option. Anyway, if you think that it would be better to set it as non-nullable I'll change that

src/Files.App/Views/SettingsPages/Preferences.xaml
yaichenbaum: That sounds a bit confusing, let's keep it in this section for now.

gave92: That section would contain how the user interacts with items.. np.

src/Files.App/BaseLayout.cs
gave92: Honestly not sure. ParentShellPageInstance can indeed be null since it's not assigned in the constructor. On the other hand we use ParentShellPageInstance without null check pretty much everywhere because we assume that OnNavigatedTo() is called before everything else.

src/Files.App/Views/SettingsPages/Preferences.xaml
yaichenbaum: I want to eventually remove the files and folders section but I need some time to think it over.

src/Files.App/BaseLayout.cs
ferrariofilippo: I added `!` because of that,  I assumed that `OnNavigatedTo()` would be called always, so I didn't add the null checks because I thought it would be less performant

gave92: Nevermind, for now the current solution is fine 👍

gave92: At least for this PR I'd keep the current solution 👍. We can work on proper null checks in a future one.

gave92:
Sorry @ferrariofilippo I've pushed more changes than I wanted to. I'm gonna revert most of them shortly.
Aside from that the functionality works well. The limitations to keep in mind for improvements:
- Selection does not work when hovering over "empty" spaces of an element (noticeable in Tiles view)

- To match explorer hover selection should consider whether Control or Shift key are pressed in order to extend the selection

gave92 (approved):

yaichenbaum:
> Selection does not work when hovering over "empty" spaces of an element (noticeable in Tiles view)

I think we should fix this before merging

ferrariofilippo:
> - Selection does not work when hovering over "empty" spaces of an element (noticeable in Tiles view)
> 
It's a problem of DetailsLayout too.
I'll fix that

yaichenbaum:
> I find out there's a problem when hovering an element's borders

This seems to happen pretty often :thinking:

ferrariofilippo:
I read MS (Windows.UI.Xaml.Controls) ListView docs, and I learnt it doesn't support any `ItemHovered` event: you can only set a style for hovered items. Because of this, you have to use `PointerEntered` event of child items, which gets invoked only when known limits.

yaichenbaum:
Can this be solved by adding a border control to the ListView item and adding the PointerEntered event on that? I don't think it'll have that big of a performance hit but we should check the difference.

ferrariofilippo:
I'll try that

ferrariofilippo:
I found a way of fixing `PointerEntered` area for `TilesLayout`. Now I'm working on `DetailsLayout`.
Other layouts are fine

gave92:
I've pushed feee91f to propose a fix for hover on elements "empty space". It should work for all layout modes, with no need to change the xaml. Idea is to subscribe the pointer events on the ListViewItem container. What do you think?

ferrariofilippo:
I think it's great. Thank you for the fix!
It works perfectly

yaichenbaum (approved):
LGTM

src/Files.App/BaseLayout.cs
```diff
@@ -897,18 +897,22 @@ protected void FileList_ContainerContentChanging(ListViewBase sender, ContainerC
...
 			if (inRecycleQueue)
 			{
 				UninitializeDrag(container);
 			}
 			else
 			{
-				container.PointerEntered += ListedItem_PointerEntered;
-				container.PointerExited += ListedItem_PointerExited;
 				container.PointerPressed += FileListItem_PointerPressed;
+				if (UserSettingsService.PreferencesSettingsService.SelectFilesOnHover)
```
yaichenbaum, line 911: It appears that we need to listen when this property changes as well, otherwise the user needs to restart the app for the setting to take effect.

gave92: True. Refreshing the file list should be enough though.

yaichenbaum: Can probably do the same thing we do when users change the setting to show hidden items.

PR #9984, opened 2022-09-12
Fix shortcuts to folders opening outside Files

gave92:
**Resolved / Related Issues**
- Fixes an issue where lnk to folders open in explorer instead of Files

**Changes**
Change fields that must be serialized to json into properties. System.Text.Json.JsonSerializer does not serialize fields by default (unlike JsonConvert).

**Validation**
- [x] Built and ran the app


yaichenbaum (commented):
Good to know, thank you!

PR #9985, opened 2022-09-12
Feature: Removed QuickLookHandler

puppetsw:
**Resolved / Related Issues**
- Related #9858 

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.


src/Files.App/ViewModels/SettingsViewModel.cs
```diff
@@ -39,28 +39,19 @@ public SettingsViewModel()
...
-                    var (status, response) = await connection.SendMessageForResponseAsync(new ValueSet()
-                    {
-                        { "Arguments", "DetectQuickLook" }
-                    });
-                    if (status == AppServiceResponseStatus.Success)
-                    {
-                        App.AppModel.IsQuickLookSupported = response.Get("IsAvailable", false);
-                    }
-                }
+                App.AppModel.IsQuickLookSupported = QuickLookHelpers.CheckQuickLookAvailability();
```
yaichenbaum, line 47: Do you want to create a new startup task for this (should be delayed until everything else is loaded).

puppetsw: Yeah, I can do that. Just put it in the QL helper class?

src/Files.App/Helpers/QuickLookHelpers.cs
```diff
@@ -1,35 +1,68 @@
...
+namespace Files.App.Helpers;
+
+public static class QuickLookHelpers
 {
-    public static class QuickLookHelpers
+    public static Task ToggleQuickLook(IShellPage associatedInstance, bool switchPreview = false)
     {
-        public static async Task ToggleQuickLook(IShellPage associatedInstance, bool switchPreview = false)
+        if (!App.AppModel.IsQuickLookSupported || !associatedInstance.SlimContentPage.IsItemSelected || associatedInstance.SlimContentPage.IsRenamingItem)
+            return null;
```
gave92, line 14: Should be `return Task.CompletedTask;` I guess, or it will crash if you await it.

src/Files.App/Helpers/QuickLookHelpers.cs
```diff
@@ -1,35 +1,68 @@
...
+        if (!App.AppModel.IsQuickLookSupported || !associatedInstance.SlimContentPage.IsItemSelected || associatedInstance.SlimContentPage.IsRenamingItem)
+            return Task.CompletedTask;
+
+        App.Logger.Info("Toggle QuickLook");
+
+        string PipeName = $"QuickLook.App.Pipe.{WindowsIdentity.GetCurrent().User?.Value}";
+        string Message = switchPreview ? "QuickLook.App.PipeMessages.Switch" : "QuickLook.App.PipeMessages.Toggle";
+
+        using var client = new NamedPipeClientStream(".", PipeName, PipeDirection.Out);
+        client.Connect();
```
gave92, line 22: To be checked what happens if you close quicklook after starting Files. We might want to change this to `client.Connect(500);`.

src/Files.App/Helpers/QuickLookHelpers.cs
```diff
@@ -1,35 +1,68 @@
...
-            if (!App.AppModel.IsQuickLookSupported || !associatedInstance.SlimContentPage.IsItemSelected || associatedInstance.SlimContentPage.IsRenamingItem)
+            string pipeName = $"QuickLook.App.Pipe.{WindowsIdentity.GetCurrent().User?.Value}";
+            string pipeSwitch = "QuickLook.App.PipeMessages.Switch";
+
+            using var client = new NamedPipeClientStream(".", pipeName, PipeDirection.Out);
+            try
             {
-                return;
-            }
+                client.Connect(500);
```
gave92, line 40: We might want `await client.ConnectAsync(500)` instead. Since you now execute this directly instead of in a separate process Connect() might freeze the UI.

src/Files.App/ViewModels/SettingsViewModel.cs
yaichenbaum: I think that's good for now 👍

src/Files.App/Helpers/QuickLookHelpers.cs
puppetsw: I made a bunch of changes to the methods. Added in the await usings as well.

gave92: Wow `await using` is new to mee, cool.

puppetsw: C#8 I believe

src/Files.App/Helpers/QuickLookHelpers.cs
```diff
@@ -1,35 +1,69 @@
...
+        if (!App.AppModel.IsQuickLookSupported || !associatedInstance.SlimContentPage.IsItemSelected || associatedInstance.SlimContentPage.IsRenamingItem)
+            return;
+
+        App.Logger.Info("Toggle QuickLook");
+
+        string pipeName = $"QuickLook.App.Pipe.{WindowsIdentity.GetCurrent().User?.Value}";
+        string message = switchPreview ? "QuickLook.App.PipeMessages.Switch" : "QuickLook.App.PipeMessages.Toggle";
+
+        await using var client = new NamedPipeClientStream(".", pipeName, PipeDirection.Out);
+        await client.ConnectAsync(TIMEOUT);
```
gave92, line 24: With the added time out we also need try-catch(TimeoutException) like below.

src/Files.App/ViewModels/SettingsViewModel.cs
gave92: Is there an await missing on this line?

puppetsw: It wasn't at the time, but the methods have been changed, so I've added in.

src/Files.App/App.xaml.cs
```diff
@@ -206,7 +206,7 @@ private static async Task InitializeAppComponentsAsync()
                     SidebarPinnedController.InitializeAsync()
                 );
                 await Task.WhenAll(
-                    AppSettings.DetectQuickLook(),
+                    QuickLookHelpers.DetectQuickLook(),
```
yaichenbaum, line 209: Would it improve performance to defer this task until the first time the user presses the spacebar?

puppetsw: I imagine it would, but only slightly. I'll investigate and get back to you.

puppetsw: According to the profiler, it spends about 2ms in the detect method.

yaichenbaum: I think it's worth improving, every additional ms counts on startup especially on slower devices where it can take several times longer.

puppetsw: Good point. I’ll make the changes soon.

src/Files.App/Helpers/QuickLookHelpers.cs
```diff
@@ -12,9 +12,12 @@ public static class QuickLookHelpers
 
     public static async Task ToggleQuickLook(IShellPage associatedInstance, bool switchPreview = false)
     {
-        if (!App.AppModel.IsQuickLookSupported || !associatedInstance.SlimContentPage.IsItemSelected || associatedInstance.SlimContentPage.IsRenamingItem)
+        if (!associatedInstance.SlimContentPage.IsItemSelected || associatedInstance.SlimContentPage.IsRenamingItem)
             return;
 
+        if (!App.AppModel.IsQuickLookSupported)
+            await DetectQuickLook();
```
yaichenbaum, line 19: Is it better to make `IsQuickLookSupported` nullable so that the `DetectQuickLook` task only runs once?

yaichenbaum: @puppetsw pushed a change for this, can you look it over?

puppetsw: We can do it that way. I was thinking if QuickLook was installed while Files was running this would still allow you to use QuickLook without needing to restart Files. - Just means every time we press the spacebar on a folder it does the `DetectQuickLook`

src/Files.App/Helpers/QuickLookHelpers.cs
```diff
@@ -15,8 +15,11 @@ public static async Task ToggleQuickLook(IShellPage associatedInstance, bool swi
         if (!associatedInstance.SlimContentPage.IsItemSelected || associatedInstance.SlimContentPage.IsRenamingItem)
             return;
 
-        if (!App.AppModel.IsQuickLookSupported)
-            await DetectQuickLook();
+        if (App.AppModel.IsQuickLookSupported == null)
+			await DetectQuickLook();
```
puppetsw, line 19: LGTM. Just the formatting here I would fix myself.

src/Files.App/Helpers/QuickLookHelpers.cs
yaichenbaum: The task also runs when files are selected. It's not worth an extra operation to handle an edge case.
For what it's worth, @hez2010 is creating a built-in solution to replace this integration.

yaichenbaum: Pushed some changes to account for that scenario

yaichenbaum: It looks like the event is getting hit three times :thinking:

puppetsw: Might have to add the null check back in on the `await DetectQuickLookAvailability();` ?

yaichenbaum: The property is no longer nullable.

yaichenbaum (approved):
LGTM

PR #10031, opened 2022-09-21
Fix: Prevent crash on fresh startup

lukeblevins:
Presently, we assume that a null return value for `ReadFromFile()` is possible. This assumption causes a crash on startup when incorrectly attempting to deserialize the empty string returned on a fresh installation. This PR makes no changes to the `ReadFromFile()` method body, as that is out of scope, but instead adjusts the return type to not statically assert null values are possible. Additionally, we will now test for both null and `string.Empty` in `DefaultJsonSettingsDatabase` to fix the issue, while ensuring maximum compatibility with future changes to the `ReadFromFile()` method.

**Validation**
- [x] Built and ran the app


src/Files.App/Helpers/AppUpdater.cs
```diff
@@ -1,101 +0,0 @@
-using Files.App.Extensions;
```
yaichenbaum, line 1: Was this removed by mistake?

lukeblevins: @yair100 It was unused and replaced by a service implementation.

yaichenbaum (approved):
LGTM

PR #10044, opened 2022-09-22
Fixed: Properties window has the wrong drag region on the title bar

### LINKED ISSUE #9919: Bug: Properties window has the wrong drag region on the titlebar
**Steps to repro**
1. Open properties
2. Try dragging window around from different areas and see you can drag even from click regions

**Steps to fix**
- Click regions should not drag the window

ferrariofilippo:
**Resolved / Related Issues**
- Closes #9919 

**Changes Details**
I added a method that calculates `DragRectangles` for `TitleBar`. It's called once on load and then whenever you resize the window.
This method creates two rectangles:
- one over `NavigationViewItem`s (so that you can drag from the 8px margin)
- one on their right side

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Views/Pages/Properties.xaml.cs
+++ b/src/Files.App/Views/Pages/Properties.xaml.cs
@@ -41,11 +41,11 @@ namespace Files.App.Views
 			InitializeComponent();
 
 			var flowDirectionSetting = /*
-                TODO ResourceContext.GetForCurrentView and ResourceContext.GetForViewIndependentUse do not exist in Windows App SDK
-                Use your ResourceManager instance to create a ResourceContext as below. If you already have a ResourceManager instance,
-                replace the new instance created below with correct instance.
-                Read: https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/migrate-to-windows-app-sdk/guides/mrtcore
-            */new Microsoft.Windows.ApplicationModel.Resources.ResourceManager().CreateResourceContext().QualifierValues["LayoutDirection"];
+				TODO ResourceContext.GetForCurrentView and ResourceContext.GetForViewIndependentUse do not exist in Windows App SDK
+				Use your ResourceManager instance to create a ResourceContext as below. If you already have a ResourceManager instance,
+				replace the new instance created below with correct instance.
+				Read: https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/migrate-to-windows-app-sdk/guides/mrtcore
+			*/new Microsoft.Windows.ApplicationModel.Resources.ResourceManager().CreateResourceContext().QualifierValues["LayoutDirection"];
 
 			if (flowDirectionSetting == "RTL")
 				FlowDirection = FlowDirection.RightToLeft;
--- a/src/Files.App/Views/Pages/Properties.xaml.cs
+++ b/src/Files.App/Views/Pages/Properties.xaml.cs
@@ -1,269 +1,258 @@
+using CommunityToolkit.WinUI;
 using Files.App.DataModels.NavigationControlItems;
 using Files.App.Filesystem;
 using Files.App.Helpers;
 using Files.App.Helpers.XamlHelpers;
 using Files.App.ViewModels;
 using Files.App.ViewModels.Properties;
-using CommunityToolkit.WinUI;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Input;
+using Microsoft.UI.Xaml.Markup;
+using Microsoft.UI.Xaml.Navigation;
+using Microsoft.UI;
+using Microsoft.UI.Windowing;
 using System;
 using System.Threading;
 using Windows.Foundation.Metadata;
 using Windows.System;
 using Windows.UI;
-using Windows.UI.Core;
-using Microsoft.UI.Xaml;
-using Microsoft.UI.Xaml.Controls;
-using Microsoft.UI.Xaml.Input;
-using Microsoft.UI.Xaml.Markup;
-using Microsoft.UI.Xaml.Media;
-using Microsoft.UI.Xaml.Media.Animation;
-using Microsoft.UI.Xaml.Navigation;
-using Microsoft.UI;
-using Microsoft.UI.Windowing;
+using Windows.Graphics;
 
 namespace Files.App.Views
 {
-    public sealed partial class Properties : Page
-    {
-        private CancellationTokenSource tokenSource = new CancellationTokenSource();
-        private ContentDialog propertiesDialog;
+	public sealed partial class Properties : Page
+	{
+		private CancellationTokenSource? tokenSource = new CancellationTokenSource();
+		private ContentDialog propertiesDialog;
 
-        private object navParameterItem;
-        private IShellPage AppInstance;
+		private object navParameterItem;
+		private IShellPage AppInstance;
 
-        private ListedItem listedItem;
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/Pages/Properties.xaml.cs
```diff
@@ -1,269 +1,282 @@
...
+				) ? Visibility.Visible : Visibility.Collapsed;
+			base.OnNavigatedTo(e);
+		}
+
+		private async void Properties_Loaded(object sender, RoutedEventArgs e)
+		{
+			AppSettings.ThemeModeChanged += AppSettings_ThemeModeChanged;
+			if (ApiInformation.IsApiContractPresent("Windows.Foundation.UniversalApiContract", 8))
+			{
+				SetDragRegion();
```
yaichenbaum, line 79: Is this duplicate? I'm pretty sure the Window size changed event is automatically called when the Window is opened.

src/Files.App/Views/Pages/Properties.xaml.cs
```diff
@@ -1,269 +1,282 @@
...
+				appWindow.Destroying += AppWindow_Destroying;
+				appWindow.Changed += AppWindow_Changed;
```
yaichenbaum, line 81: Shouldn't this be the size changed event?

yaichenbaum:
I think this needs to be applied to the security properties window as well.

src/Files.App/Views/Pages/Properties.xaml.cs
ferrariofilippo: If you meant subscribing the new method to `SizeChanged` event, I can't do that as AppWindow events are only `Closing`, `Destroying` and `Changed`.
I would not rename `AppWindow_Changed` as this is a general event, raised in many cases. 
Because of the above, I'll rename `SetDragRegion` to `AppWindow_SizeChanged`

src/Files.App/Views/Pages/Properties.xaml.cs
ferrariofilippo: Actually, this isn't a duplicate as I'm subscribing `AppWindow_Changed` inside `Properties_Loaded`. When the latter is raised, the window is already opened.

src/Files.App/Views/Pages/Properties.xaml.cs
yaichenbaum: That's correct, we do something similar on the MainPage

ferrariofilippo:
> I think this needs to be applied to the security properties window as well.

Can you tell me how to open that window, so that I can test it?

ferrariofilippo:
I just found out that `DragRegion` should be re-calculated even when a `NavigationViewItem` is selected, because when this happens now, drag region is still wrong



Swap between the two doesn't raise `SizeChanged` event

yaichenbaum:
> Can you tell me how to open that window, so that I can test it?

It's under the advanced permissions button on the permissions page.

ferrariofilippo:
> think this needs to be applied to the security properties window as well.

Is this page supposed to have `Minimize/Maximize` buttons

ferrariofilippo:
> I think this needs to be applied to the security properties window as well.

This isn't needed as `AdvancedPermission` has no `NavigationView`, so `DragRegion` is right by default

yaichenbaum:
> Is this page supposed to have Minimize/Maximize buttons

No

src/Files.App/Views/Pages/Properties.xaml.cs
```diff
@@ -1,269 +1,300 @@
...
+					}
+				}
+			}
+			appWindow.TitleBar.SetDragRectangles(new RectInt32[]
+			{
+				new RectInt32(
+					0,
+					0,
+					navigationViewWidth,
+					(int)NavigationView.ActualOffset.Y),
```
yaichenbaum, line 116: This might be easier to read if it's all on one line

src/Files.App/Views/Pages/Properties.xaml.cs
```diff
@@ -1,269 +1,300 @@
...
+				propertiesDialog = DependencyObjectHelpers.FindParent<ContentDialog>(this);
+				propertiesDialog.Closed += PropertiesDialog_Closed;
+			}
+		}
+
+		private void NavigationView_SizeChanged(object? sender, SizeChangedEventArgs? e)
+		{
+			int navigationViewWidth = 0;
+			bool overflowAdded = false;
+			foreach (NavigationViewItem item in NavigationView.MenuItems)
```
yaichenbaum, line 95: This seems like an expensive calculation, it might be better to collapse the items that are hidden so that they don't take any space.

ferrariofilippo: The problem is that `NavigationView`, by default, takes more space than it's needed.
For example when there are only the first two tabs opened (and all other tabs are collapsed) NavView width is 320 while the two tabs total width is 209.
I was thinking of calculating `ActualWidth` once when the page is loaded and then when the selected tab changes, but if a page is hidden and the page get resized, DragRegion would be off

yaichenbaum: Can you see if setting x:Load to false helps? For some reason hidden items are still taking up space but x:Load should solve that.

ferrariofilippo: Just learnt `x:Load` can't be used on `NavigationViewItem` (it throws a `XAML Parsing` exception).

yaichenbaum: Can you add some comments to the code explaining the work arounds?

src/Files.App/Views/Pages/Properties.xaml.cs
```diff
@@ -107,18 +112,14 @@ private void NavigationView_SizeChanged(object? sender, SizeChangedEventArgs? e)
 					}
 				}
 			}
+
+			// Sets properties window drag region.
 			appWindow.TitleBar.SetDragRectangles(new RectInt32[]
```
yaichenbaum, line 117: Does this work on Windows 10? @gave92

gave92: Nope, SetDragRectangles has currently no effect on win10. But I'm still able to drag the window from the area around the close button. Good enough till they fix it, using the same workaround we added in MainPage.cs would be overly complex I guess.

src/Files.App/Views/Pages/PropertiesSecurity.xaml.cs
```diff
@@ -95,7 +95,11 @@ private async void OpenAdvancedProperties()
                     }, new SuppressNavigationTransitionInfo());
 
                     // Initialize window
-                    var propertiesWindow = new WinUIEx.WindowEx();
+                    var propertiesWindow = new WinUIEx.WindowEx()
```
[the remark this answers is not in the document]
yaichenbaum: This change is good, was probably an oversight on the UWP version.

src/Files.App/Views/Pages/Properties.xaml.cs
yaichenbaum: Sounds good, we can leave as is

yaichenbaum (approved):

yaichenbaum:
> one on their right side

Looks like this issue is still there

ferrariofilippo:
Do you mean in this area?

ferrariofilippo:
Should I add this area too?

yaichenbaum:
image

PR #10083, opened 2022-09-29
Fix: Fixed issue where preview of ANSI text didn't work

gave92:
**Resolved / Related Issues**
- Fixes an issue where preview for a text file with ANSI encoding (default for notepad) does not work

**Changes**
- Use StreamReader instead of DataReader as it auto-detects encoding

**Validation**
- [x] Built and ran the app


yaichenbaum (approved):
LGTM

PR #10085, opened 2022-09-29
Fix: Fixed an issue where Ctrl+X wasn't working

gave92:
**Resolved / Related Issues**
- Fixes an issue where cutting an item failed due to DisatcherQueue being null.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Helpers/UIFilesystemHelpers.cs
+++ b/src/Files.App/Helpers/UIFilesystemHelpers.cs
@@ -20,6 +20,8 @@ using Windows.Storage;
 using Files.Backend.Enums;
 using Windows.System;
 using Microsoft.UI.Xaml.Controls;
+using CommunityToolkit.Mvvm.DependencyInjection;
+using Files.Backend.Services;
 
 namespace Files.App.Helpers
 {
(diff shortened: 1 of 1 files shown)


src/Files.App/Helpers/UIFilesystemHelpers.cs
```diff
@@ -48,7 +50,7 @@ public static async void CutItem(IShellPage associatedInstance)
 
                 try
                 {
-                    var dispatcherQueue = DispatcherQueue.GetForCurrentThread();
+                    var threadingService = Ioc.Default.GetRequiredService<IThreadingService>();
```
yaichenbaum, line 53: Can you add a comment explaining why this is needed? It's helpful to prevent someone from changing this in the future 🙂

yaichenbaum (approved):
LGTM

src/Files.App/Helpers/UIFilesystemHelpers.cs
gave92: Could you confirm that Ctrl+X is currently _not_ working for you on main?

yaichenbaum: It's working for me on `main` but Windows 10 and different specs could be at play.

yaichenbaum: Should I merge?

gave92: Mmm better to wait to merge this one then b/c this change should not be necessary (and I could swear it used to work).

PR #10089, opened 2022-09-30
Feature: Added option to change default layout mode

### LINKED ISSUE #4155: Feature: Add option to choose the default layout mode
I like to view my files in medium or large icons, but it's defaulted at list view. 

To have an option to default the view for _all_ folders would be great. I only recently found out that the legacy File Explorer has such an option (Set folders to preferred size, then go to view > options > view > apply to folders > apply/okay). If there were an option in the settings that's more straightforward ("layout mode: [large/medium/small/tiles/details]") that would be awesome!

Not sure of any other potential alternatives.

**Additional context**
N/A

| Capability | Priority |
| :---------- | :------- |
| This proposal will add a combo box to the folder preferences section allowing users to select the default layout mode | Must |

image

yaichenbaum:
**Resolved / Related Issues**
- Closes #4155

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
+++ b/src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
@@ -1,135 +1,137 @@
 using CommunityToolkit.Mvvm.DependencyInjection;
-using Files.Backend.Services.Settings;
+using Files.App.Filesystem;
 using Files.App.ViewModels;
 using Files.App.ViewModels.Previews;
+using Files.Backend.Services.Settings;
+using Files.Shared.Enums;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using Windows.Storage;
-using System.Collections.Generic;
-using Files.App.Filesystem;
 
 namespace Files.App.Helpers
 {
-    public static class AdaptiveLayoutHelpers
-    {
-        public static bool PredictLayoutMode(FolderSettingsViewModel folderSettings, string path, IList<ListedItem> filesAndFolders)
-        {
-            IUserSettingsService userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
+	public static class AdaptiveLayoutHelpers
+	{
+		public static bool PredictLayoutMode(FolderSettingsViewModel folderSettings, string path, IList<ListedItem> filesAndFolders)
+		{
+			IUserSettingsService userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
 
-            if (!userSettingsService.PreferencesSettingsService.ForceLayoutPreferencesOnAllDirectories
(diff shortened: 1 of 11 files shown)


yaichenbaum:
There seems to be an issue where the app doesn't follow this setting when the option to force preferences are turned off.

yaichenbaum:
@ferrariofilippo can you look this over? I want to get it merged before #10117. This requires some stress testing, try to break it 🙂

ferrariofilippo:
I tried everything I thought of, and it seems the only bug is the one we already know.

I think the user should be able to decide even among grid layout sizes.
At the moment this feature uses the last size selected by the user when `ForceOnAllDirectories` is enabled.
If you want to reproduce:
- Reset preferences
- Enable `ForceOnAllDirectories`
- Set grid layout size
- Disable `ForceOnAllDirectories`
- Change `DefaultLayout` to `Grid`

I think this should be changed because a user would need to do all the above only to change the default items' size for the grid layout.

yaichenbaum:
There are a range of sizes for the grid layouts, perhaps we can support a couple presets in the drop down (small, medium, large). I'm not sure how to do that right now as we don't have enums for those.

ferrariofilippo:
> we can support a couple presets in the drop down (small, medium, large)

I think that's enough.

yaichenbaum:
@ferrariofilippo are okay merging this before those additional options are added?

yaichenbaum:
Merged this as is so that we can merge the default column options, should be okay if we add the options later on.

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
```diff
@@ -1,135 +1,137 @@
...
-                && folderSettings.IsAdaptiveLayoutEnabled
-                && !folderSettings.IsLayoutModeFixed)
-            {
-                Action layoutDetails = () => folderSettings.ToggleLayoutModeDetailsView(false);
-                Action layoutTiles = () => folderSettings.ToggleLayoutModeTiles(false);
-                Action layoutGridView = () => folderSettings.ToggleLayoutModeGridView(folderSettings.GridViewSize);
+			if (!userSettingsService.PreferencesSettingsService.ForceLayoutPreferencesOnAllDirectories
+				&& folderSettings.IsAdaptiveLayoutEnabled
+				&& !folderSettings.IsLayoutModeFixed
+				&& folderSettings.LayoutMode == FolderLayoutModes.Adaptive)
```
gave92, line 23: Note that this prevents adaptive layout mode from working because folderSettings.LayoutMode is never set to Adaptive.  Even in adaptive mode folderSettings.LayoutMode is still either details or grid or tiles. The above check IsAdaptiveLayoutEnabled is what we want.

Issue #6768, opened 2021-11-07
Bug: Draging files over other items makes it look like they're being selected


martsum:
### Description

Dragging and dropping files behaviour is inconsistent and unexpected.

### Steps To Reproduce

1. Drag a file or folder over another item
2. Other items will sometimes become 'highlighted' with the accent color
3. When releasing the mouse, the other items will continue to stay highlighted.

### Requirements

When dragging a file/folder over other items, they should not become highlighted...

### Files Version

2.0.34

### Windows Version

Windows 11 21H2 22000.282

marcinhordziej:
Having the same issue on 2.1.13.0 on W11.

alishtamimi:
Any updates on this issue? Is it in progress to solve or not?

Josh65-2201:
Asking @yair100 for approval.

sthivaios:
I have just checked this, and the issue is solved, at least for me.

I am using the PRE version of "Files". I suggest updating the app (if it is not up to date) and checking it again.

So, for me, the items did not get hilighted.

alishtamimi:
It is almost solved, but yet, there is a problem when you drag a file and change its place for example, it will remain highlighted even if you click on another file.

sthivaios:
@alishtamimi 

Oh yes! I just tried that and I can confirm that his happens...

https://user-images.githubusercontent.com/109022579/195158682-fa24afa2-6973-4514-89cb-9e27078d3816.mp4

yaichenbaum:
The issue is resolved in the details layout but it's still present in the other layouts.

yaichenbaum:
The issue is resolved in `main`.

Issue #9594, opened 2022-07-21
Feature: Remove accent color from non-selected items in the column's layout


yaira2:
### Solution/Idea

![image](https://user-images.githubusercontent.com/39923744/180116588-d02aee63-ef52-4bdb-901d-fd8d084e1b14.png)

### Requirements

- Remove accent color from non-selected items in the column's layout
- Non folders should still have the regular accented color section state

### Files Version

v2.3.3

PR #10097, opened 2022-10-01
Fixed: Draging files over items select them

ferrariofilippo:
**Resolved / Related Issues**
- Related: #6768 

**Comments**
- I fixed the bug delaying the selection at the end of the timer. This way, the hovered item is selected just before it's opened.
- There's still a problem with `ColumnsLayout`, where items have `SelectedStyle` even when they're not. I believe the bug affects `ListViewBase` since the code is the same as in `GridViewBase`, where the feature is working.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/BaseLayout.cs
+++ b/src/Files.App/BaseLayout.cs
@@ -790,8 +790,6 @@ namespace Files.App
 			{
 				deferral = e.GetDeferral();
 
-				ItemManipulationModel.SetSelectedItem(item);
-
 				if (dragOverItem != item)
 				{
 					dragOverItem = item;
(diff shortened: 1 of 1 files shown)


yaichenbaum (approved):
LGTM

yaichenbaum:
> There's still a problem with ColumnsLayout, where items have SelectedStyle even when they're not. I believe the bug affects ListViewBase since the code is the same as in GridViewBase, where the feature is working

Doesn't DetailsLayout use ListViewBase as well?

ferrariofilippo:
If you hover over a folder for 1s while dragging an item, Files selects it and opens it.
If you use `DetailsLayout` this is not a problem, because Files loads a new pane.
With `ColumnsLayout`, the selected folder is opened in a new pane, but the previous one remains open. The bug happens here.

ferrariofilippo:
I just found out that you have already tracked this issue in #9594

yaichenbaum:
https://github.com/files-community/Files/issues/9594 is a different issue

Issue #1089, opened 2020-06-17
Feature: Improving the recycle bin experience


yaichenbaum:
- [ ] Recycle bin could also use different icons for when it's empty or not #9195
- [x] I was testing file restoration, and all are restored without extension
- [x] Show original location for items in the recycle bin
- [ ] #9946

SOI7:
Apart from the obvious need of an Empty Bin and a Restore File function, one grip I have with the Recycle Bin atm is that it gets lost inside the libraries tree, and since it's not a regular folder, it should be inside a separatde area, maybe either like the Drives section or ditching the library layout and use it as a simple button somewhere

gave92:
Thanks for the feedback.
Both "Empty Bin" and "Restore File" functions are already available (through right click context menu).
After the latest modifications the recycle bin link on the sidebar will appear as the second item (right after th quick access/home link) but this can problably be improved 👍

XPoppyX:
I think an option to empty the recycle bin from the sidebar

gave92:
@XPoppyX perhaps a "empty bin" button on the topbar would be better? Like instead of the disabled "plus" button.

You're right the recycle bin changing position on the sidebar is not ok. I guess it should either be always at the bottom or in a separate section.

XPoppyX:
> perhaps a "empty bin" button on the topbar would be better? Like instead of the disabled "plus" button.

I want to avoid having to open the recycle bin first to be able to empty

gave92:
I see so perhaps an option to emtpy the bin when you right click on the recycle bin entry on the sidebar

yaichenbaum:
When deleting and item or emptying recycle bin, there should be a confirmation dialog.

SOI7:
Recycle bin could also use different icons for when it's empty or not. Also, right now I need to refresh it to see changes after emptying it

XPoppyX:
I was testing file restoration, and all are restored without extension

when it is deleted:
![Anotación 2020-06-28 090022](https://user-images.githubusercontent.com/61259627/85948263-ea0afc00-b91d-11ea-8eb8-7da3d21f86f1.png)

after restoring:
![Anotación 2020-06-28 090100](https://user-images.githubusercontent.com/61259627/85948265-eb3c2900-b91d-11ea-92b2-9ca7e09cf271.png)

tsvietOK:
@XPoppyX Can't reproduce, v0.11

XPoppyX:
[...]

PR #10113, opened 2022-10-03
Feature: Recycle Bin changing icons in the side bar

### LINKED ISSUE #9195: Highlight whether the Recycle Bin is empty or not.
### What's the Problem?

To know if the Recycle Bin is empty or not, you have to open it, or open its contextual menu.

### Solution/Idea

You can add information about the item from the Recycle Bin to favorites. This can be done in many ways: different icon, bold text, number of items in parentheses when there are any, or any different style.

### Alternatives

None

### Priorities

| Capability | Priority |
| :---------- | :------- |

### Files Version

_No response_

### Windows Version

_No response_

### Comments

_No response_

QuaintMako:
**Resolved / Related Issues**
- Closes #9195 
- Related #1089 

**What have been done**
- An event is now raised when the recycle bin is modified.
- A new icon for the empty recycle bin has been indexed.
- A light refactoring in `src/Files.App/DataModels/SidebarPinnedModel.cs` has been done to the function `AddItemToSidebarAsync`.

**Validation**
- [X] Built and ran the app

**Screenshots (optional)**

When the recycle bin is empty:
empty

When the recycle bin is not empty:
filled


src/Files.App/Constants.cs
```diff
@@ -43,7 +42,8 @@ public static class ImageRes
 			public const int WindowsDrive = 36;
 			public const int ThisPC = 109;
 			public const int NetworkDrives = 25;
-			public const int RecycleBin = 55;
+			public const int RecycleBin = 54;
```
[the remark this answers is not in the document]
QuaintMako: Yes, it's the icon when the recycle bin contains items.
Do you want me to rename the constant to FullRecycleBin ?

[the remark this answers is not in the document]
yaichenbaum: I think it's fine as is, was just confirming.

src/Files.App/DataModels/SidebarPinnedModel.cs
```diff
@@ -280,6 +282,21 @@ public async Task AddItemToSidebarAsync(string path)
...
 		}
 
+		private async Task<byte[]?> RetrieveItemIconData(string itemPath, FilesystemResult<BaseStorageFolder> result)
+		{
+			byte[]? iconData = null;
+
+			if (result)
+			{
+				iconData = await FileThumbnailHelper.LoadIconFromStorageItemAsync(result.Result, 24u, Windows.Storage.FileProperties.ThumbnailMode.ListView);
+				iconData ??= await FileThumbnailHelper.LoadIconFromStorageItemAsync(result.Result, 24u, Windows.Storage.FileProperties.ThumbnailMode.SingleItem);
```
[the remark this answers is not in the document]
yaichenbaum: That's fine, let's remove the SingleItem since ListView should be fine.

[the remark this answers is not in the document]
yaichenbaum: Thanks!

src/Files.App/Constants.cs
```diff
@@ -28,7 +28,6 @@ public static class KnownImageFormats
 		{
 			public const string BITMAP_IMAGE_FORMAT = "bitmapimage";
 		}
-
 		public static class ImageRes
```
yaichenbaum, line 31: Line 31 removed by accident?

QuaintMako: I believe so. My bad.

src/Files.App/DataModels/SidebarPinnedModel.cs
```diff
@@ -280,6 +282,20 @@ public async Task AddItemToSidebarAsync(string path)
 			AddLocationItemToSidebar(locationItem);
 		}
 
+		private async Task<byte[]?> RetrieveItemIconData(string itemPath, FilesystemResult<BaseStorageFolder> result)
+		{
+			byte[]? iconData = null;
+
+			if (result)
```
d2dyno1, line 289: Can you check for and retrieve result in the calling function and not this one?

src/Files.App/Helpers/RecycleBinHelpers.cs
```diff
@@ -76,6 +80,7 @@ public async Task EmptyRecycleBin()
             if (result == ContentDialogResult.Primary)
             {
                 Shell32.SHEmptyRecycleBin(IntPtr.Zero, null, Shell32.SHERB.SHERB_NOCONFIRMATION | Shell32.SHERB.SHERB_NOPROGRESSUI);
+                RaiseRecycleBinChangedEvent();
```
d2dyno1, line 83: You can use null check
```csharp
RecycleBinChanged?.Invoke(null, null);
```

src/Files.App/Helpers/UIHelpers.cs
```diff
@@ -143,5 +144,14 @@ private static IEnumerable<IconFileInfo> LoadSidebarIconResources()
 
             return imageResList;
         }
+
+        public static int GetAdaptedRecycleBinIconIndex()
+        {
+            if (new RecycleBinHelpers().RecycleBinHasItems())
```
d2dyno1, line 150: Can you avoid creating new instances?

src/Files.App/ViewModels/SidebarViewModel.cs
```diff
@@ -542,6 +543,26 @@ public async void EmptyRecycleBin(RoutedEventArgs e)
 			await RecycleBinHelpers.S_EmptyRecycleBin();
 		}
 
+		public async void OnRecycleBinChanged(object sender, EventArgs e)
+		{
+			var recycleBinLocationItem = SideBarItems
```
d2dyno1, line 548: I don't know if that Linq makes any sense. Wouldn't it be better if it was a single `foreach` loop? 🙂

src/Files.App/ViewModels/SidebarViewModel.cs
```diff
@@ -236,6 +236,7 @@ public SidebarViewModel()
 			SideBarItems = new BulkConcurrentObservableCollection<INavigationControlItem>();
 			EmptyRecycleBinCommand = new RelayCommand<RoutedEventArgs>(EmptyRecycleBin);
 			UserSettingsService.OnSettingChangedEvent += UserSettingsService_OnSettingChangedEvent;
+			RecycleBinHelpers.RecycleBinChanged += OnRecycleBinChanged;
```
d2dyno1, line 239: Unsure, but the event should probably be detached to avoid memory leaks (unless we don't have to do it)

d2dyno1 (changes requested):

src/Files.App/DataModels/SidebarPinnedModel.cs
QuaintMako: Could do it, yet by doing so we remove a security in this function. May be a problem if it's called somewhere else in the future. I'm doing the modification nonetheless.

src/Files.App/Helpers/UIHelpers.cs
QuaintMako: Currently no, the while RecycleBinHelper is kind of a mess to that regard. There is static and not static methods. But it's a good time as any to make a refactoring of that helper.

src/Files.App/Helpers/RecycleBinHelpers.cs
QuaintMako: Changed inside the function that handles the event raising. But would you rather have the event raised directly without any method maybe?

src/Files.App/ViewModels/SidebarViewModel.cs
QuaintMako: Done, even though I have no idea how to make it a one `foreach` loop only.

src/Files.App/Helpers/UIHelpers.cs
QuaintMako: Made RecycleBin a static class, think it will address the issue.

src/Files.App/Helpers/RecycleBinHelpers.cs
d2dyno1: Yes. There's no need for additional method to raise this event

d2dyno1:
@QuaintMako Could you resolve conflicts before it's merged?

src/Files.App/Helpers/RecycleBinHelpers.cs
QuaintMako: We actuallty do need it to raise the event outside of the class. We need to raise it from `RestoreItemsFromTrashAsync` and `DeleteItemsAsync` from `FilesystemHelpers.cs`.

QuaintMako: @d2dyno1 Should I do something to indeed remove the need of the method entirely or should it stay that way? 

Removing it would means to move methods from the `FileSystemHelpers` to the `RecycleBinHelpers`, albeit it wouldn't be their optimal locations.

d2dyno1: I think moving it would make sense (?). In addition, it should be moved to instantiable helper (because it exposes an event) but that'd be part of a bigger refactor out of scope for this PR.

engrproama: I think that moving it is better and good

QuaintMako:
Marking that PR as a draft for now, I won't be able to take care of it for a few days.

src/Files.App/Helpers/RecycleBinHelpers.cs
QuaintMako: Not sure how to actually implement this. I have been turning the issue in my head for the past few days, but I cannot find a solution that feels actually clean for it. _Consider this message as a formal request for help._

After giving it some thoughts, the methods from `FilesystemHelpers.cs` should **not** be moved. They are in their right places, they are not only linked to the recycle bin. 

A possibility would to raise events in `FilesystemHelpers.cs` when `Delete` or `Restore` are called, and have methods to update the recycle bin icon subscribe to them. But then, an issue comes from the **RecycleBinHelpers.EmptyRecycleBin()** method that does not use the `Delete` method to clear the recycle bin.

I must be missing something obvious, and I'd like to see this PR through.

QuaintMako:
Would it be possible to get a `help requested` label on this PR please?

QuaintMako:
After discussions with Yair on Discord, I'm closing this PR for now. Waiting for the big refactoring coming up soon, which should solve the event issue.

PR #10174, opened 2022-10-11
Feature: Display prompt when user doesn't have permission to access a folder

### LINKED ISSUE #1194: Feature: Display modal when trying to view directories that require admin access
**Describe the request**
I've multiple users on my PC and in some scenarios I want to be open a file or folder in another user. When I access another users' folder in File Explorer for the first time, it will show me a small dialog where I can get access (Standard users need admin password). In Files UWP, you can only access another users' folder after you get access permission via File Explorer or the properties window. When you try to access them without permission, Files UWP will simply say the folder is empty. This isn't really a Files UWP bug since Windows returns no items when the user doesn't have permission however we should check whether the user has permission or not and we should add the current user to the permission list for that folder.

**Requirements**
- Display modal explaining that the user does not have permission to access the location
- Modal should have a "close" action
- Modal should have an "edit permissions" action

### LINKED ISSUE #7856: Feature: Add an access denied error when drive/folder is blocked because of permissions

I have set a drives permissions for a user account to not allow access, it works but is not very clear to the user that there are permissions stoping them from seeing the drives contents, I presume this happends for folders too.

### Steps to reproduce

1. Open drives properties page (don't do it on the C drive it will brake Windows)
2. set a user account to deny on the full control option
3. accses the drive with that denied account
4. see the "folder is empty" message

### Requirements

- Display access denied error instead of opening the drive

### Files Version

2.3.11.0

### Windows Version

10.19043.1899

### Comments

_No response_

puppetsw:
* Moved FilePermissions methods from FullTrust to helpers.
* Updated SecurityProperties to use new permission helpers.
* Display modal when user doesn't have access to a folder.

**Resolved / Related Issues**
- Closes #1194
- Closes #7856

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Screenshot 2022-10-16 152353


Josh-65:
It properly would make more sense to have the primary button run Files as admin then attempt to open the folder first. If it doesn't have access still, then show the edit permissions button.

src/Files.App/Files.App.csproj
```diff
@@ -175,6 +175,7 @@
     <PackageReference Include="CommunityToolkit.WinUI.Notifications" Version="7.1.2" />
     <PackageReference Include="CommunityToolkit.WinUI.UI.Behaviors" Version="7.1.2" />
     <PackageReference Include="CommunityToolkit.WinUI.UI.Controls" Version="7.1.2" />
+    <PackageReference Include="Tulpep.ActiveDirectoryObjectPicker" Version="3.0.11" />
```
yaichenbaum, line 178: 🤔 

I'm hesitant to add another third-party package, hard much effort is involved in doing this ourselves?

puppetsw: I didn’t really look into it because it was the same one in the fulltrust project. But I’ll get back to you.

yaichenbaum: I didn't realize we were already using this, definitely not a blocker but still something to remove if possible.

puppetsw: Looks to be a bit of work involved going by the project source. I would be hesitant to just try re-create it. Maybe someone else can chime in and give their opinion?

yaichenbaum: Fully agree to leave as is right now, eventually we should be at the point where we can implement the code ourselves but it doesn't make sense to do that right now.

yaichenbaum:
> If the primary button is clicked on the modal the PropertiesSecurityAdvanced view is displayed. (I'm not sure the best way to handle this)

Do we need to display an edit permissions option? It might be cleaner to simply inform the user that they don't have permission.

puppetsw:
> > If the primary button is clicked on the modal the PropertiesSecurityAdvanced view is displayed. (I'm not sure the best way to handle this)
> 
> Do we need to display an edit permissions option? It might be cleaner to simply inform the user that they don't have permission.

Probably not. They could go to the folder properties themselves and change it there.  From memory explorer has an option to just take ownership.

yaichenbaum:
> They could go to the folder properties themselves and change it there

Let's do this for now as I'm not convinced the modal needs a shortcut. We can continue gathering feedback and add one in the future if needed.

puppetsw:
> > They could go to the folder properties themselves and change it there
> 
> Let's do this for now as I'm not convinced the modal needs a shortcut. We can continue gathering feedback and add one in the future if needed.

Sounds good. Are you happy with the current text? or would like it changed?

yaichenbaum:
Perhaps this would be better
Title: "Access denied"
Content "You don't have access to open this folder."

src/Files.App/Views/Pages/PropertiesSecurity.xaml.cs
```diff
@@ -76,7 +76,7 @@ public override void Dispose()
 
         }
 
-        private void OpenAdvancedProperties()
+        public void OpenAdvancedProperties()
```
yaichenbaum, line 79:
```suggestion
        private void OpenAdvancedProperties()
```

src/Files.App/Helpers/NavigationHelpers.cs
```diff
@@ -261,9 +262,16 @@ private static async Task<FilesystemResult> OpenDirectory(string path, IShellPag
...
-			bool IsShortcut = path.EndsWith(".lnk", StringComparison.Ordinal) || path.EndsWith(".url", StringComparison.Ordinal); // Determine
+			bool isShortcut = path.EndsWith(".lnk", StringComparison.Ordinal) || path.EndsWith(".url", StringComparison.Ordinal); // Determine
 
-			if (IsShortcut)
+			// Check for folder permission
+			if (!FolderHelpers.CheckFolderAccessWithWin32(path))
+			{
+				await DialogDisplayHelper.ShowDialogAsync(DynamicDialogFactory.GetFor_FolderNoPermission());
+				return (FilesystemResult)false;
+			}
```
yaichenbaum, line 272: Do we have to explicitly check for access? I'm pretty sure the navigation apis will throw for access denied.

yaichenbaum:
@puppetsw does this resolved #7856?

src/Files.App/Helpers/NavigationHelpers.cs
puppetsw: Good point. I’ll check this.

puppetsw:
> @puppetsw does this resolve #7856?

I’m not sure. I’ll investigate.

src/Files.App/Helpers/NavigationHelpers.cs
puppetsw: I can't get my head around it. If there is an exception, is must be handled somewhere already. I guess we don't want to be checking each folder before we open them if the user has permission or not though.

yaichenbaum: cc @gave92

yaichenbaum:
What's the status of this PR?

puppetsw:
> What's the status of this PR?

It doesn't look like it solves #7856 but I haven't been able to test properly yet.

src/Files.App/Helpers/NavigationHelpers.cs
gave92: Sry missed this comment. I've just checked with `C:\Users\Administrator` and the is actually no exception being thrown (ItemViewModel/EnumerateItemsFromStandardFolderAsync() would be the place). The Storage API does not return any error, it just makes the folder appear empty. I'm positive this was not the case with UWP, we even have code for showing a dialog:
```cs
else if (res == FileSystemStatusCode.Unauthorized)
{
  //TODO: proper dialog
  await DialogDisplayHelper.ShowDialogAsync(
    "AccessDenied".GetLocalizedResource(),
    "SubDirectoryAccessDenied".GetLocalizedResource());
  return -1;
}
```

gave92: Using `CheckFolderAccessWithWin32` is probably not ok because I imagine it would show access denied when browsing e.g phone storage.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2841,6 +2841,9 @@
   <data name="DeleteShortcutDescription" xml:space="preserve">
     <value>The target destination cannot be found {0}. Do you want to delete this shortcut?</value>
   </data>
+  <data name="AccessDeniedToFolder" xml:space="preserve">
+    <value>You don't have access to open this folder.</value>
```
yaira2, line 2845:
```suggestion
    <value>You don't have permission to access this folder.</value>
```

yaira2 (approved):
LGTM

PR #10188, opened 2022-10-12
Fix: Icons not showing up in tab control.

### LINKED ISSUE #9929: Bug: Icons are no longer shown on the tab control
**Requirements**
- Display the icon of current directory on the tab control

QuaintMako:
**Resolved / Related Issues**
- Closes #9929 

**What has been done**

It seems that the icons not showing up on TabViewModel in WinUI3 is a known problem on the Microsoft-UI side. (See #9929 discussion for more informations.)

A (temporary?) solution is to switch to Symbol Icons. We no longer have beautiful icons access, but we can get back to having icons.

If this is not a viable solution, I may look into another workaround, but I haven't the smallest clue if it's possible.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**


yaichenbaum:
I think we should continue looking for another workaround. Is it posible to add our own icon control to the template?

QuaintMako:
Will dig into it.

QuaintMako:
Closing PR while looking for an alternative.

PR #10240, opened 2022-10-17
Fix: Fixed an issue where the name column had a max width

### LINKED ISSUE #10230: Bug: File name column has a max width
### Description

I can't stretch the file name column past a certain width This is very confusing to me. Uninstalling and reinstalling is still the same.

### Steps To Reproduce
1. Open folder in the details layout
2. Try resizing the name column as wide as possible

### Steps to fix
- Remove the max width

### Files Version
v3.3.12

### Windows Version

Win 10 21H2

ferrariofilippo:
**Resolved / Related Issues**
- Closes #10230 

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
@@ -299,8 +299,7 @@
 									<ColumnDefinition Width="Auto" />
 									<ColumnDefinition
 										x:Name="Column2"
-										Width="{x:Bind ColumnsViewModel.NameColumn.Length, Mode=OneWay}"
-										MaxWidth="{x:Bind ColumnsViewModel.NameColumn.MaxLength, Mode=OneWay}" />
+										Width="{x:Bind ColumnsViewModel.NameColumn.Length, Mode=OneWay}" />
 									<ColumnDefinition Width="Auto" />
 									<ColumnDefinition
 										x:Name="Column3"
--- a/src/Files.App/ViewModels/ColumnsViewModel.cs
+++ b/src/Files.App/ViewModels/ColumnsViewModel.cs
@@ -26,7 +26,10 @@ namespace Files.App.ViewModels
             set => SetProperty(ref tagColumn, value);
         }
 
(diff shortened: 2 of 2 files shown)


src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -299,8 +299,7 @@
 									<ColumnDefinition Width="Auto" />
 									<ColumnDefinition
 										x:Name="Column2"
-										Width="{x:Bind ColumnsViewModel.NameColumn.Length, Mode=OneWay}"
-										MaxWidth="{x:Bind ColumnsViewModel.NameColumn.MaxLength, Mode=OneWay}" />
```
yaichenbaum, line 303: Should `MaxLength` be removed from the ViewModel?

ferrariofilippo: Personally, I would remove this constraint. I don't think it's a problem if the user has very wide columns.

yaichenbaum: Most columns should have this constraint, NameColumn is the exception. Is it possible to set the default for the NameColumn MaxLength to *?

ferrariofilippo: We can set `double.MaxValue` as default value

yaichenbaum: We don't actually have to bind to it in the xaml but I want the c# value to reflect it.

ferrariofilippo: What about something like this:

        private ColumnViewModel nameColumn = new ColumnViewModel()
        {
            NormalMaxLength = double.PositiveInfinity
        };

yaichenbaum: Maybe just set the max width to 1000 😂

yaichenbaum (approved):
LGTM

PR #10242, opened 2022-10-17
Feature: Improved the tooltip when hovering over drives

### LINKED ISSUE #10214: Show drives size in sidebar

To get the available size of a drive, you need to access the home page or the properties of the drive. It's annoying to know which disk to move files to.

It would be convenient to display it in the sidebar. I attach an example below. I also changed the recent tooltip to be non-redundant and consistent with properties. This could be an setting (Don't show/Only chart/Used size/Free size/Total size).

### Requirements

- Just edit the template.

### Files Version

Version: 2.3.48.0 OS Version: 10.0.22621.674

### Windows Version

Windows 11 22621.674

### Comments

Here a screenshot.
drive1
drive2

cinqmilleans:
**Resolved / Related Issues**
Changes the sidebar drives template to match the property menu 
Closes #10214.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
tooltip_light
tooltip_dark

before/after
fixbefore
fixafter

### DIFF
--- a/src/Files.App/Views/Pages/PropertiesGeneral.xaml
+++ b/src/Files.App/Views/Pages/PropertiesGeneral.xaml
@@ -441,7 +441,7 @@
 				x:Name="PropertiesDriveCapacity"
 				Grid.Row="2"
 				Grid.Column="1"
-				Padding="25,0,0,0"
+				Padding="28,0,0,0"
 				x:Load="{x:Bind ViewModel.DriveCapacityVisibility, Mode=OneWay}"
 				Style="{StaticResource PropertyName}"
 				Text="{helpers:ResourceString Name=PropertiesDriveCapacity/Text}" />
--- a/src/Files.App/Views/Pages/PropertiesGeneral.xaml
+++ b/src/Files.App/Views/Pages/PropertiesGeneral.xaml
@@ -441,7 +441,7 @@
 				x:Name="PropertiesDriveCapacity"
 				Grid.Row="2"
 				Grid.Column="1"
-				Padding="28,0,0,0"
+				Padding="25,0,0,0"
 				x:Load="{x:Bind ViewModel.DriveCapacityVisibility, Mode=OneWay}"
 				Style="{StaticResource PropertyName}"
 				Text="{helpers:ResourceString Name=PropertiesDriveCapacity/Text}" />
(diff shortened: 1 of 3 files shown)


yaichenbaum:
@cinqmilleans thank you for the PR. Can you confirm how were these changes validated?

src/Files.App/Views/Pages/PropertiesGeneral.xaml
```diff
@@ -441,7 +441,7 @@
 				x:Name="PropertiesDriveCapacity"
 				Grid.Row="2"
 				Grid.Column="1"
-				Padding="28,0,0,0"
+				Padding="25,0,0,0"
```
yaichenbaum, line 444: This was very specific, is there any reason it was changed?

cinqmilleans: As explained above, there is a gap of 3 pixels compared to the other 2 texts. The rectangle uses 25 pixels (15 + right margin 10). With this change, texts are aligned vertically. This is visible on the before/after screenshorts.

yaichenbaum: Can you send a screenshot showing the issue? There might be a way to solve it while keeping the divisible by four rule.

cinqmilleans: There is already a screenshot above (before/after). For the multiple of 4, we can change the rectangle, or its margin, because it is not a multiple of 4.

yaichenbaum: I don't the issue in the screenshots

cinqmilleans: Here are the same images with a red line to clarify the problem. We clearly see that the 3rd text is not aligned with the 2 others. Looking at the code, it is clear that the gap should be 25 (used by the rectangle) and not 28.

fixbefore
fixafter

yaichenbaum: I'd like to do this differently, can you revert this change and I'll follow up with a separate PR?

yaichenbaum (approved):
LGTM

PR #10243, opened 2022-10-17
Feature: Added extract archive button inside archives

### LINKED ISSUE #9822: Feature: Add extract archive button to the toolbar inside when browsing inside archives
### What's the Problem?

It's not very clear how to extract archives once they're already open

### Requirements 
- Added "Extract archive" button to the toolbar, this should open a modal asking where the user wants to extract the archive.
- Modal should have a message "Where do you want to extract this archive?"
- Modal should have a read only text box that defaults to the parent path + the archive name
- Modal should have a browse button where the user can choose a different location
- Modal should have an extract button
- Modal should have a cancel button

ferrariofilippo:
**Resolved / Related Issues**
- Closes #9822

**Comments**
I recycled the existing code and adapted it so that it works in both cases (inside and outside archives).
If you think I should implement this feature with brand new code, I'll find another workaround.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Helpers/FileExtensionHelpers.cs
+++ b/src/Files.App/Helpers/FileExtensionHelpers.cs
@@ -11,12 +11,10 @@ namespace Files.App.Helpers
         /// <param name="fileExtensionToCheck">The file extension to check.</param>
         /// <returns><c>true</c> if the fileExtensionToCheck is an image;
         /// otherwise, <c>false</c>.</returns>
-        public static bool IsImageFile(string fileExtensionToCheck)
+        public static bool IsImageFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
                 return false;
-            }
 
             return fileExtensionToCheck.Equals(".png", StringComparison.OrdinalIgnoreCase) ||
                    fileExtensionToCheck.Equals(".jpg", StringComparison.OrdinalIgnoreCase) ||
@@ -46,20 +44,18 @@ namespace Files.App.Helpers
         /// <param name="fileExtensionToCheck">The file extension to check.</param>
         /// <returns><c>true</c> if the fileExtensionToCheck is a zip bundle file;
         /// otherwise <c>false</c>.</returns>
-        public static bool IsZipFile(string fileExtensionToCheck)
+        public static bool IsZipFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
                 return false;
-            }
 
             return new[] { ".zip", ".msix", ".appx", ".msixbundle", ".7z", ".rar", ".tar" }
                 .Contains(fileExtensionToCheck, StringComparer.OrdinalIgnoreCase);
         }
 
-        public static bool IsBrowsableZipFile(string filePath, out string ext)
+        public static bool IsBrowsableZipFile(string? filePath, out string? ext)
         {
-            if (string.IsNullOrEmpty(filePath))
+            if (string.IsNullOrWhiteSpace(filePath))
             {
                 ext = null;
                 return false;
@@ -70,12 +66,10 @@ namespace Files.App.Helpers
             return ext is not null;
         }
 
-        public static bool IsInfFile(string fileExtensionToCheck)
+        public static bool IsInfFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
                 return false;
-            }
 
             return fileExtensionToCheck.Equals(".inf", StringComparison.OrdinalIgnoreCase);
         }
@@ -87,12 +81,10 @@ namespace Files.App.Helpers
         /// <returns><c>true</c> if the fileExtensionToCheck is a font file;
         /// otherwise <c>false</c>.</returns>
         /// <remarks>Font file types are; fon, otf, ttc, ttf</remarks>
-        public static bool IsFontFile(string fileExtensionToCheck)
+        public static bool IsFontFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
                 return false;
-            }
 
             return fileExtensionToCheck.Equals(".fon", StringComparison.OrdinalIgnoreCase) ||
                      fileExtensionToCheck.Equals(".otf", StringComparison.OrdinalIgnoreCase) ||
@@ -107,12 +99,10 @@ namespace Files.App.Helpers
         /// <returns><c>true</c> if the fileExtensionToCheck is a shortcute file;
         /// otherwise <c>false</c>.</returns>
         /// <remarks>Font file type is .lnkf</remarks>
-        public static bool IsShortcutFile(string fileExtensionToCheck)
+        public static bool IsShortcutFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
                 return false;
-            }
 
             return fileExtensionToCheck.Equals(".lnk", StringComparison.OrdinalIgnoreCase);
         }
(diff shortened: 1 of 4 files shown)


src/Files.App/Helpers/FileExtensionHelpers.cs
```diff
@@ -11,12 +11,10 @@ public static class FileExtensionHelpers
         /// <param name="fileExtensionToCheck">The file extension to check.</param>
         /// <returns><c>true</c> if the fileExtensionToCheck is an image;
         /// otherwise, <c>false</c>.</returns>
-        public static bool IsImageFile(string fileExtensionToCheck)
+        public static bool IsImageFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
```
QuaintMako, line 16: `if` block needs to have the curly braces I believe.

src/Files.App/Helpers/FileExtensionHelpers.cs
```diff
@@ -46,20 +44,18 @@ public static bool IsPowerShellFile(string fileExtensionToCheck)
         /// <param name="fileExtensionToCheck">The file extension to check.</param>
         /// <returns><c>true</c> if the fileExtensionToCheck is a zip bundle file;
         /// otherwise <c>false</c>.</returns>
-        public static bool IsZipFile(string fileExtensionToCheck)
+        public static bool IsZipFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
```
QuaintMako, line 49: `if `block without curly braces as well here.

src/Files.App/Helpers/FileExtensionHelpers.cs
```diff
@@ -70,12 +66,10 @@ public static bool IsBrowsableZipFile(string filePath, out string ext)
             return ext is not null;
         }
 
-        public static bool IsInfFile(string fileExtensionToCheck)
+        public static bool IsInfFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
```
QuaintMako, line 71: Curly braces missing as well.

src/Files.App/Helpers/FileExtensionHelpers.cs
```diff
@@ -87,12 +81,10 @@ public static bool IsInfFile(string fileExtensionToCheck)
         /// <returns><c>true</c> if the fileExtensionToCheck is a font file;
         /// otherwise <c>false</c>.</returns>
         /// <remarks>Font file types are; fon, otf, ttc, ttf</remarks>
-        public static bool IsFontFile(string fileExtensionToCheck)
+        public static bool IsFontFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
```
QuaintMako, line 86: The curly braces are missing here too.

src/Files.App/Helpers/FileExtensionHelpers.cs
```diff
@@ -107,12 +99,10 @@ public static bool IsFontFile(string fileExtensionToCheck)
         /// <returns><c>true</c> if the fileExtensionToCheck is a shortcute file;
         /// otherwise <c>false</c>.</returns>
         /// <remarks>Font file type is .lnkf</remarks>
-        public static bool IsShortcutFile(string fileExtensionToCheck)
+        public static bool IsShortcutFile(string? fileExtensionToCheck)
         {
-            if (string.IsNullOrEmpty(fileExtensionToCheck))
-            {
+            if (string.IsNullOrWhiteSpace(fileExtensionToCheck))
```
QuaintMako, line 104: Curly braces needed.

src/Files.App/ViewModels/ToolbarViewModel.cs
```diff
@@ -1150,9 +1150,11 @@ public List<ListedItem> SelectedItems
...
-		public bool CanExtract => SelectedItems is not null && SelectedItems.Any() && SelectedItems.All(x => FileExtensionHelpers.IsZipFile(x.FileExtension)) && !InstanceViewModel.IsPageTypeRecycleBin;
-		public string ExtractToText => CanExtract ? SelectedItems.Count > 1 ? string.Format("ExtractToChildFolder".GetLocalizedResource(), $"*{Path.DirectorySeparatorChar}") : string.Format("ExtractToChildFolder".GetLocalizedResource() + "\\", Path.GetFileNameWithoutExtension(selectedItems.First().Name)) : "ExtractToChildFolder".GetLocalizedResource();
-		public bool IsMultipleArchivesSelected => CanExtract && SelectedItems.Count > 1;
+		public bool CanExtract => IsArchiveOpened ? (SelectedItems is null || !SelectedItems.Any()) : IsArchiveSelected;
+		public string ExtractToText => IsArchiveSelected ? SelectedItems.Count > 1 ? string.Format("ExtractToChildFolder".GetLocalizedResource(), $"*{Path.DirectorySeparatorChar}") : string.Format("ExtractToChildFolder".GetLocalizedResource() + "\\", Path.GetFileNameWithoutExtension(selectedItems.First().Name)) : "ExtractToChildFolder".GetLocalizedResource();
```
QuaintMako, line 1154: I'd move this line lower, around the `OpenInTerminal` method line 1164 so there is not a `public string` method amongst the `public bool` evaluators.

src/Files.App/ViewModels/ToolbarViewModel.cs
```diff
@@ -1150,9 +1150,11 @@ public List<ListedItem> SelectedItems
...
+		public bool IsArchiveOpened => FileExtensionHelpers.IsZipFile(Path.GetExtension(pathControlDisplayText));
+		public bool IsArchiveSelected => SelectedItems is not null && SelectedItems.Any() && SelectedItems.All(x => FileExtensionHelpers.IsZipFile(x.FileExtension)) && !InstanceViewModel.IsPageTypeRecycleBin;
+		public bool IsMultipleArchivesSelected => IsArchiveSelected && SelectedItems.Count > 1;
```
QuaintMako, line 1157: Shouldn't it check if all the selected items are archives as well?

ferrariofilippo: which one?

src/Files.App/Helpers/FileExtensionHelpers.cs
QuaintMako: My bad!

src/Files.App/ViewModels/ToolbarViewModel.cs
QuaintMako: Sorry, still a bit new on reviews, seems I selected more lines than I intended too.

I meant `IsMultipleArchivesSelected`, shouldn't it check that all the items selected are only archives?

ferrariofilippo: If you mean that `IsMultipleArchivesSelected` should check if all items are archives, that is already checked by `IsArchiveSelected`

QuaintMako: Alright, so we got it covered. Wouldn't be the place I expect to such a test, but it works.

ferrariofilippo: Actually I didn't want to name it `IsArchiveSelected`, but I couldn't find any other more suitable name.

src/Files.App/ViewModels/ToolbarViewModel.cs
ferrariofilippo: Ok, I'll change that

src/Files.App/ViewModels/ToolbarViewModel.cs
QuaintMako: I'd propose to rename it `HasSelectionOneArchiveOnly` and then `IsMultipleArchivesSelected` into `HasSelectionMultipleArchivesOnly`.

Then extract `SelectedItems is not null && SelectedItems.Any() && SelectedItems.All(x => FileExtensionHelpers.IsZipFile(x.FileExtension))  && !InstanceViewModel.IsPageTypeRecycleBin` into a method named `IsSelectionArchivesOnly`.

You will end up with :

`public bool HasSelectionOneArchiveOnly => IsSelectionArchivesOnly && SelectedItems.Count == 1`
and 
`public bool HasSelectionMultipleArchivesOnly => IsSelectionArchivesOnly && SelectedItems.Count > 1`

QuaintMako (approved):
LGTM.

yaichenbaum:
I noticed an issue where selecting a zip file and opening the toolbar flyout shows disabled options
image

I also noticed you added an option to open the destination folder when complete. I really like this added touch! It appears that the toolbar button to extract the archive is still visible after it navigates to the extracted items.

ferrariofilippo:
> I noticed you added an option to open the destination folder when complete.

Actually this feature already existed before my commit.
I'll try to fix the two bugs as quickly as possible

PR #10258, opened 2022-10-20
Fix: Fixed issue where drives had the wrong order in the sidebar

### LINKED ISSUE #10252: Bug: Drives section has the wrong order
### Description

The Drives section sometimes appears in the wrong order (with c: last), while the home page is correct.
The order should be by letter and not by discovery. The C: drive sometimes appears last.

### Steps To Reproduce

For C:, I do not know. It seems random, half the time.

The bug can be seen as follows:
1) Connect a USB key.
2) Connect a second USB key.
3) Remove the first USB key.
4) Reconnect the first USB key.
It appears with the initial letter but in last position. The order is no longer correct.

### Steps to fix

Drives are added in the order of discovery. The position should be forced when adding a drives. I can do it.

### Files Version

dev #eabf72e

### Windows Version

Windows 11 22621.674

### Relevant Assets/Logs

drives1

cinqmilleans:
**Resolved / Related Issues**
Detected drives are always displayed at the end. It makes more sense to display them in order according to the path. Sometimes internal drives are not displayed in path order.  

Closes #10252.

This pr keeps drives sorted by path in the sidebar and homepage.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/UserControls/Widgets/DrivesWidget.xaml.cs
+++ b/src/Files.App/UserControls/Widgets/DrivesWidget.xaml.cs
@@ -7,6 +7,7 @@ using Files.App.Helpers;
 using Files.App.Helpers.XamlHelpers;
 using Files.App.ViewModels.Widgets;
 using Files.Backend.Services.Settings;
+using Files.Shared.Extensions;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Controls.Primitives;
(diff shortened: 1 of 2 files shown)


yaichenbaum:
@cinqmilleans can you verify how these changes were validated?

yaichenbaum:
Btw, if this closes an issue, you can edit the PR description to look like this and it'll link the PR to the issue 
image

src/Files.App/UserControls/Widgets/DrivesWidget.xaml.cs
```diff
@@ -25,8 +26,8 @@
 
 namespace Files.App.UserControls.Widgets
 {
-    public class DriveCardItem : ObservableObject, IWidgetCardItem<DriveItem>
-    {
+    public class DriveCardItem : ObservableObject, IWidgetCardItem<DriveItem>, IComparable<DriveCardItem>
+	{
```
yaichenbaum, line 30: Is spacing set to tabs?

cinqmilleans: The indentation is fixed.

yaichenbaum (approved):
LGTM

PR #10260, opened 2022-10-20
Fix: Fixed issue with the text alignment in the properties window

cinqmilleans:
**Resolved / Related Issues**
Fix the bad position of the text Capacity in the properties of drives.
The correct padding is 25, not 28, as the rectangle (width 15 + margin 10).

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before/After
fixbefore
fixafter


yaichenbaum:
The width should be changed to 16 and the margin to 12

cinqmilleans:
The changes are made. I applied the same change to the tooltip for consistency.
Capacity1
Capacity2

yaichenbaum (approved):
LGTM

PR #10269, opened 2022-10-21
Feature: Display drive letter of the connected network drives

### LINKED ISSUE #10261: Feature: Display drive letter of the connected network drives

It is possible to assign a letter to a network drive. It then appears in the local drives but without displaying its letter. I don't know if it's a bug or if it's intentional. It would be convenient to display its drive letter like any other drive. I see no reason not to.

### Requirements

Modify DriveItem.Text for network drives.

### Files Version

dev commit 4e3b8db

### Windows Version

Windows 11 22621.674

### Comments

Drive Z:NVIDIA_SHIELD does not display its letter Z:.
Drive1
Drive2
Drive3

cinqmilleans:
**Resolved / Related Issues**
The access path of network disks which have a local access path is not displayed in their label, contrary to fixed disks. This is so in the item name received by .net and FileExplorer has the same problem. This pr adds this missing path for consistency and readability
Closes #10261.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before/After
Drive1
fix2
fix4
fix5

### DIFF
--- a/src/Files.App/DataModels/NavigationControlItems/DriveItem.cs
+++ b/src/Files.App/DataModels/NavigationControlItems/DriveItem.cs
@@ -152,7 +152,7 @@ namespace Files.App.DataModels.NavigationControlItems
 			if (imageStream != null)
 				item.IconData = await imageStream.ToByteArrayAsync();
 
-			item.Text = root.DisplayName;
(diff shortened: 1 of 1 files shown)


yaichenbaum (approved):
LGTM

PR #10274, opened 2022-10-22
Fix: Fixed progress ring style in drive tooltip

cinqmilleans:
**Resolved / Related Issues**
This pr fixes the style of displaying drive sizes (Property and Tooltip).
- In the tooltip, the background of the ring was missing.
- In the tooltip, the 3 texts did not use exactly the same style.
- In light mode, the color of the free space legend did not match the color used by the ring. The color of the ring has been used because it is more readable and this corresponds to the progressbar of the Drive widget.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before Property
Before_Property_Dark
Before_Property_Light
After Property
After_Property_Dark
After_Property_Light
Before Tooltip
Before_Tooltip_Dark
Before_Tooltip_Light
After Tooltip
After_Tooltip_Dark
After_Tooltip_Light

### DIFF
--- a/src/Files.App/Views/Pages/PropertiesGeneral.xaml
+++ b/src/Files.App/Views/Pages/PropertiesGeneral.xaml
@@ -21,11 +21,9 @@
 				<ResourceDictionary x:Key="Light">
 					<SolidColorBrush x:Key="ProgressRingBackgroundThemeBrush" Color="#c3c3c3" />
 				</ResourceDictionary>
-
 				<ResourceDictionary x:Key="Dark">
 					<SolidColorBrush x:Key="ProgressRingBackgroundThemeBrush" Color="#191919" />
 				</ResourceDictionary>
-
 				<ResourceDictionary x:Key="HighContrast">
 					<SolidColorBrush x:Key="ProgressRingBackgroundThemeBrush" Color="#c3c3c3" />
 				</ResourceDictionary>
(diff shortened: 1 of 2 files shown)


QuaintMako:
LGTM.

src/Files.App/Views/Pages/PropertiesGeneral.xaml
```diff
@@ -360,7 +360,7 @@
 					Width="16"
 					Height="16"
 					Margin="0,0,12,0"
-					Fill="{ThemeResource SystemAccentColorLight2}"
+					Fill="{Binding Foreground, ElementName=SpaceUserProgressRing}"
```
yaichenbaum, line 363: What was wrong with the previous fill?

cinqmilleans: In light mode, the color of the ring and its legend do not match for the space used. The SystemAccentColorLight2 color is used for dark mode but not for light mode. This change ensures that the legend always matches the ring.

Before
Before_Property_Light
After
After_Property_Light

src/Files.App/Views/Pages/PropertiesGeneral.xaml
```diff
@@ -293,12 +291,13 @@
 				<ColumnDefinition Width="Auto" />
 			</Grid.ColumnDefinitions>
 			<Grid.RowDefinitions>
-				<RowDefinition Height="Auto" />
-				<RowDefinition Height="Auto" />
-				<RowDefinition Height="Auto" />
+				<RowDefinition Height="*" />
+				<RowDefinition Height="*" />
+				<RowDefinition Height="*" />
```
yaichenbaum, line 296: Does this need to be changed?

cinqmilleans: This guarantees that the 3 texts use the same height, which was not the case before the correction of the size of the 3rd text. It doesn't change anything in practice anymore because it's fixed, but it says in the code that they should be the same height.

yaichenbaum: What if the text is wrapped onto a different line? I think Auto works better here.

yaichenbaum (approved):
LGTM

src/Files.App/Views/Pages/PropertiesGeneral.xaml
cinqmilleans: You are right. It's back to auto.

PR #10285, opened 2022-10-23
Remove OpenProtectedFileForRead function

gave92:
**Resolved / Related Issues**
- Removes unnecessary OpenProtectedFileForRead function. The function was used when opening files from inside an archive due to UWP limitations but it's not necessary anymore.

**Validation**
- [x] Built and ran the app
- [x] Launch a file from inside an archive and check it still works


yaichenbaum (approved):
LGTM

PR #10295, opened 2022-10-25
Fix: Fixed issue where the sidebars right click menu would sometimes get cut off

cinqmilleans:
**Resolved / Related Issues**
In the sidebar, the context menu may not be entirely visible. This pr changes its placement so that it is no longer obscured by the edge of the window.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before
ContextBefore
After
ContextAfter


yaichenbaum (approved):
LGTM

PR #10299, opened 2022-10-26
Fix: Remove icons from extract flyout

heftymouse:
**Resolved / Related Issues**
None

**Changes made**
This PR removes the icons from the archive extraction flyout, since they take up space and add clutter while adding little information.

**Validation**
- [x] Built and ran the app


src/Files.App/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -1,935 +1,924 @@
 ﻿<UserControl
-	x:Class="Files.App.UserControls.InnerNavigationToolbar"
```
yaichenbaum, line 2: Spacing shouldn't be changed (should be set to tabs)

yaichenbaum (changes requested):

yaichenbaum (approved):
LGTM

PR #10302, opened 2022-10-26
Code Quality: Misc fixes and bumping dependencies

hez2010:
Although we cannot enable trimming for now.

### DIFF
--- a/builds/azure-pipelines-release.yml
+++ b/builds/azure-pipelines-release.yml
@@ -106,7 +106,13 @@ jobs:
   - task: UseDotNet@2
     inputs:
       packageType: sdk
-      version: 6.0.400
+      version: 7.0.x
+      includePreviewVersions: true
+
+  - task: CmdLine@2
+    inputs:
+      script: |
+        for /f %%a in ('dir /b /a:d %localappdata%\Microsoft\VisualStudio\17*') do echo UsePreviews=True>%localappdata%\Microsoft\VisualStudio\%%a\sdk.txt
 
   - task: NuGetToolInstaller@1
 
@@ -220,7 +226,13 @@ jobs:
   - task: UseDotNet@2
     inputs:
       packageType: sdk
-      version: 6.0.400
+      version: 7.0.x
+      includePreviewVersions: true
+
+  - task: CmdLine@2
+    inputs:
+      script: |
+        for /f %%a in ('dir /b /a:d %localappdata%\Microsoft\VisualStudio\17*') do echo UsePreviews=True>%localappdata%\Microsoft\VisualStudio\%%a\sdk.txt
 
   - task: NuGetToolInstaller@1
 


hez2010:
Ready for review

src/Files.OpenDialog/FilesLauncher/FilesLauncher.vcxproj
```diff
@@ -381,12 +381,12 @@
...
+    <Import Project="..\..\..\packages\Microsoft.Windows.ImplementationLibrary.1.0.220914.1\build\native\Microsoft.Windows.ImplementationLibrary.targets" Condition="Exists('..\..\..\packages\Microsoft.Windows.ImplementationLibrary.1.0.220914.1\build\native\Microsoft.Windows.ImplementationLibrary.targets')" />
   </ImportGroup>
   <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
     <PropertyGroup>
-      <ErrorText>Questo progetto fa riferimento a uno o più pacchetti NuGet che non sono presenti in questo computer. Usare lo strumento di ripristino dei pacchetti NuGet per scaricarli. Per altre informazioni, vedere http://go.microsoft.com/fwlink/?LinkID=322105. Il file mancante è {0}.</ErrorText>
+      <ErrorText>このプロジェクトは、このコンピューター上にない NuGet パッケージを参照しています。それらのパッケージをダウンロードするには、[NuGet パッケージの復元] を使用します。詳細については、http://go.microsoft.com/fwlink/?LinkID=322105 を参照してください。見つからないファイルは {0} です。</ErrorText>
```
QuaintMako, line 388: Maybe the text error should be in English?

src/Files.Shared/Extensions/LinqExtensions.cs
```diff
@@ -41,7 +41,7 @@ public static class LinqExtensions
 			return defaultValue;
 		}
 
-		public static Task<TValue?> GetAsync<TKey, TValue>(this IDictionary<TKey, Task<TValue?>> dictionary, TKey key, Func<Task<TValue?>> defaultValueFunc)
+		public static Task<TValue?> GetAsync<TKey, TValue>(this IDictionary<TKey, Task<TValue?>> dictionary, TKey key, Func<Task<TValue?>> defaultValueFunc) where TKey : notnull
 		{
 			if (dictionary is null || key is null)
```
QuaintMako, line 46: Does the `key is null` in the condition line 46 still makes sense with the `where TKey : notnull` ?

builds/azure-pipelines.yml
```diff
@@ -62,7 +62,7 @@ steps:
 - task: UseDotNet@2
   inputs:
     packageType: sdk
-    version: 6.0.400
+    version: 7.0.0
```
yaichenbaum, line 65: @hez2010 do you know which version we need?

hez2010: Oops I forgot to update. We can simply use `7.0.*` and set `include-prerelease: true`.

hez2010: # Resolved.

src/Files.Shared/Extensions/LinqExtensions.cs
hez2010: `where TKey : notnull` doesn't enforce non-nullable unless we treat all nullable related warnings as errors.

builds/azure-pipelines.yml
yaichenbaum: Still failing :thinking:
We might need to specify the exact version number

hez2010: The exact sdk version number is `7.0.100-rc.2.22477.23`.

hez2010: Why MSBuild cannot find the installed sdk?

hez2010: Now things go right. 
We need to write a `UsePreviews=True` to sdk.txt in VS user data directory to enable preview SDK adoption for VS.

yaichenbaum (approved):
LGTM

PR #10318, opened 2022-10-28
Feature: Improve loading animations on the home page

### LINKED ISSUE #8441: Feature: Improve loading animations on the home page
### Description

When switching to the home page it takes a while for the folders widget (at the top) to load the folders. After they load, they just pop in making everything on the screen move suddenly. Also my computer is pretty powerful, so I don't think it's caused by a lack of resources.

### Steps To Reproduce

1. Switch to a page other than the Home page
2. Switch back to the Home page
3. Observe that the folder widget doesn't load the folders immediately, and they pop in once loaded

### Expected behavior

They should act identically to the drives widget. The drives are already there when the page opens, and they smoothly animate down.

### Files Version

Version: 2.1.13.0 OS Version: 10.0.22000.527

### Windows Version

Windows 11 21H2 22000.527

### Relevant Assets/Logs

Here is an example of what I am talking about:
Folder Widget

This doesn't seem like the kind of thing where logs would be helpful but let me know if you need them.

QuaintMako:
**Resolved / Related Issues**
- Closes #8441 

**What has been done**
- Switching from `BulkConcurrentObservableCollection` to `ObservableCollection`.
- Minor refactoring.

**Validation**
- [X] Built and ran the app

**Screenshots**
AnimationHome

### DIFF
--- a/src/Files.App/UserControls/Widgets/DrivesWidget.xaml.cs
+++ b/src/Files.App/UserControls/Widgets/DrivesWidget.xaml.cs
@@ -80,7 +80,7 @@ namespace Files.App.UserControls.Widgets
 
         public event PropertyChangedEventHandler PropertyChanged;
 
-        public static ObservableCollection<DriveCardItem> ItemsAdded = new ObservableCollection<DriveCardItem>();
+        public static ObservableCollection<DriveCardItem> ItemsAdded = new();
 
         private IShellPage associatedInstance;
 
--- a/src/Files.App/UserControls/Widgets/FolderWidget.xaml.cs
+++ b/src/Files.App/UserControls/Widgets/FolderWidget.xaml.cs
@@ -1,13 +1,21 @@
-using Files.App.Filesystem;
-using Files.App.Helpers;
-using Files.Backend.Services.Settings;
-using Files.App.ViewModels.Widgets;
-using Files.App.Extensions;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.DependencyInjection;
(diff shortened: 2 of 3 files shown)


QuaintMako:
Setting this as a draft for several reasons :
1- The result on my computer looks a bit choppy. I'd appreciate if someone could test it on their machine.
2- Switching from `BulkCollection `to `ObservableCollection` fixed the issue. But why was it Bulk in the first place? Need to be investigated. Still wrapping my head around what a `BulkCollection` is.

hez2010:
> Still wrapping my head around what a BulkCollection is.

It's our own implementation of `ObservableCollection` which supports bulk inserting multiple items without need of trigger `CollectionChanged` event when every single item gets inserted. This is a significant optimization to avoid UI freeze while dealing with plenty of data.

QuaintMako:
> It's our own implementation of `ObservableCollection` which supports bulk inserting multiple items without need of trigger `CollectionChanged` event when every single item gets inserted, and it's implemented in a thread-safe manner. This is a significant optimization to avoid UI freeze while dealing with plenty of data.
> 
> We prefer to use `BulkConcurrentObservableCollection` rather than `ObservableCollection` in Files.

Thanks for the explanation, I see the utility then. 

In the case of this issue, maybe it's viable to use ObservableCollection? We haven't a lot of data to deal with, only six elements. In not, we'd need to somehow get `BulkCollection` to be "usable by animations" on the UI.

hez2010:
I think it's fine to switch it here.

yaichenbaum (approved):
LGTM

Issue #10320, opened 2022-10-28
Bug: Localization switching back to Default on DetailsBrowser


QuaintMako:
### Description

When opening a view in Details Browser, when the view is done loading, all the items localization switch back to the default language.

### Steps To Reproduce

1. Go into Settings and set any language other than default (to better see the issue)
2. Switch back to any tab and open a folder with several different object.
3. Observe the localization going from selected language to the default language.

### Steps to fix

- The localization stays to the default language.

### Files Version

Version : 2.3.48.0 OS

### Windows Version

W11 21H2 Version: 10.0.22000.1098

### Relevant Assets/Logs

![BrokenLoca](https://user-images.githubusercontent.com/110472580/198686371-e99dcbae-294d-4180-9482-5dc57cd4e663.gif)

QuaintMako:
Looking at the gif, there may be a bunch of stuff broken regarding the localization. Can I see the buttons under the Favorites and the date of modification. But it may be because my computer's language is set to English.

yaichenbaum:
Favorites work off of the folder names

yaichenbaum:
See #1586 for reference

yaichenbaum:
We're going to leave this as is

PR #10324, opened 2022-10-29
Fix: Use date from Windows culture instead of Files culture

cinqmilleans:
**Resolved / Related Issues**
Some dates may display in the Windows culture (language and format) instead of the Files culture. This is fixed. 

The issue mentions other location issues. These are special folders whose names are provided by Windows. We cannot modify these texts according to the language in a simple way. I think it is useless to replace these texts according to the language.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/ServicesImplementation/DateTimeFormatter/AbstractDateTimeFormatter.cs
+++ b/src/Files.App/ServicesImplementation/DateTimeFormatter/AbstractDateTimeFormatter.cs
@@ -1,13 +1,14 @@
-using Files.Shared.Services.DateTimeFormatter;
 using Files.App.Extensions;
+using Files.Shared.Services.DateTimeFormatter;
 using System;
 using System.Globalization;
+using Windows.Globalization;
 
 namespace Files.App.ServicesImplementation.DateTimeFormatter
 {
-    internal abstract class AbstractDateTimeFormatter : IDateTimeFormatter
+	internal abstract class AbstractDateTimeFormatter : IDateTimeFormatter
     {
-        private readonly Calendar calendar = new CultureInfo(CultureInfo.CurrentUICulture.Name).Calendar;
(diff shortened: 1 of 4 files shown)


src/Files.App/ServicesImplementation/DateTimeFormatter/AbstractDateTimeFormatter.cs
```diff
@@ -1,13 +1,14 @@
...
 using Files.App.Extensions;
+using Files.Shared.Services.DateTimeFormatter;
 using System;
 using System.Globalization;
+using Windows.Globalization;
 
 namespace Files.App.ServicesImplementation.DateTimeFormatter
 {
-    internal abstract class AbstractDateTimeFormatter : IDateTimeFormatter
+	internal abstract class AbstractDateTimeFormatter : IDateTimeFormatter
```
QuaintMako, line 9: The formatting is slightly off here.

QuaintMako (approved):

src/Files.App/ServicesImplementation/DateTimeFormatter/AbstractDateTimeFormatter.cs
cinqmilleans: The indentation is repaired.

QuaintMako (approved):

yaichenbaum:
@cinqmilleans can link you link this to an issue?
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

yaichenbaum:
This doesn't appear to resolve #10320 🤔

cinqmilleans:
Sure enough, I only saw the dates issue, but not the Type column issue. I was able to reproduce the problem and I will prepare a fix. However it seems that this is another problem and it will be another pr.

yaichenbaum:
Seems like two separate issues, LGTM.

PR #10326, opened 2022-10-29
Code quality: Compare null with 'is' instead of '=='

cinqmilleans:
**Resolved / Related Issues**
Comparing the null value using "is/is not" instead of "==/!=" is more efficient because it avoids comparing using the Equals method (which can be replaced by IEquatable<>, take resources, ...). This pr replaces all occurrences.

**Validation**
- [x] Built and ran the app


src/Files.App/Filesystem/StorageEnumerators/UniversalStorageEnumerator.cs
```diff
@@ -88,9 +88,9 @@ public static class UniversalStorageEnumerator
...
                         {
                             var fileEntry = await AddFileAsync(item.AsBaseStorageFile(), currentStorageFolder, cancellationToken);
-                            if (fileEntry != null)
+                            if (fileEntry is not null)
                             {
-                                if (defaultIconPairs != null)
+                                if (defaultIconPairs is not null)
                                 {
                                     if (!string.IsNullOrEmpty(fileEntry.FileExtension))
                                     {
```
QuaintMako, line 96: Not the initial scope of the PR, but if you could take 5mn to fuse the nested `if `here, that'd be awesome!

QuaintMako (approved):
LGTM, all the changes have their correct equivalence.

QuaintMako:
Related to #4180

yaichenbaum (commented):
Looks good!

It might be interesting to run a benchmark and see the difference.

yaichenbaum (approved):

cinqmilleans:
Merge fixed.

hez2010 (approved):
LGTM

PR #10338, opened 2022-10-30
Fix: Fixed issue where type column wasn't translated

### LINKED ISSUE #10320: Bug: Localization switching back to Default on DetailsBrowser

cinqmilleans:
**Resolved / Related Issues**
The column Type is displayed in the Windows language. This is because ViewModel replaces the value of ItemType (which is created and located by Files) with the value DisplayType (which is managed by Windows). This pr removes this override so that the display is the one created by Files. Closes #10320.

**Validation**
- [x] Built and ran the app


QuaintMako (dismissed):
LGTM.

yaichenbaum:
The overrides from Windows are usually better, this is the before and after
image

image

cinqmilleans:
We have several choices:
1) Do not modify anything. The column will always display in the Windows language. Users using Files in a different language are probably extremely rare. In addition, they are accustomed to or understand the Windows language.
2) Select a list of cases to deal with ourselves (Folder, ...) and not manage the others.
3) Improve our code to better handle all cases. It requires a lot of work and knowing all the files.
4) Manage everything ourselves as it is currently done.

I propose 1) which requires no work and satisfies the user.

yaichenbaum:
> Do not modify anything. The column will always display in the Windows language. Users using Files in a different language are probably extremely rare. In addition, they are accustomed to or understand the Windows language.

What would this look like?

cinqmilleans:
It's closed this pr and continue as before. The column appears in the Windows language. For most users this does not change anything because Windows and Files will be in the same language.

PR #10339, opened 2022-10-30
Fix: Fixed column Type text

cinqmilleans:
**Resolved / Related Issues**
Column names no longer end with ":", except for the Type column. This pr fixes this in English. The translations will have to adapt because the problem exists in the other languages.

I did not create an issue because it would have taken me as long as creating the pr.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before
Columns
After
Columns2


yaichenbaum (approved):
LGTM

PR #10354, opened 2022-10-31
Fix: toggle default explorer setting

gave92:
**Resolved / Related Issues**
- Fixes an issue where toggling the "set Files as default" setting is not reliable (toggling it off would set files as default). In WinUI3 apparently the toggle state is updated after the Toggled event is invoked.

**Validation**
- [x] Built and ran the app


lukeblevins:
This is a very good find.

yaichenbaum (approved):
LGTM

PR #10358, opened 2022-10-31
Code Quality: Refactored TerminalController

cinqmilleans:
**Resolved / Related Issues**
Modify TerminalController, fixed warnings, code more readable, 

**Validation**
- [x] Built and ran the app


yaichenbaum:
Can we get rid of Terminal controller? @d2dyno1

src/Files.App/Controllers/TerminalController.cs
```diff
@@ -15,198 +15,192 @@
...
+		private async Task StartWatchConfigChangeAsync()
+		{
+			var queryOptions = new QueryOptions
+			{
+				ApplicationSearchFilter = "System.FileName:" + JsonFileName
+			};
+
+			var settingsFolder = await ApplicationData.Current.LocalFolder.GetFolderAsync("settings");
+			var query = settingsFolder.CreateFileQueryWithOptions(queryOptions);
+			query.ContentsChanged += Query_ContentsChanged;
```
QuaintMako, line 109: Shouldn't it be unsubscribed somewhere?

src/Files.App/Controllers/TerminalController.cs
```diff
@@ -15,198 +15,192 @@
...
+			catch {}
+		}
+
+		private async Task<TerminalFileModel> GetDefaultTerminalFileModel()
+		{
+			try
+			{
+				var uri = new Uri(defaultTerminalPath);
+				var defaultFile = await StorageFile.GetFileFromApplicationUriAsync(uri);
+				var defaultContent = await FileIO.ReadTextAsync(defaultFile);
```
QuaintMako, line 139: Those lines are repetition of lines l.117 to 1.119 with different names for the variables. Worth creating a method of them both.

src/Files.App/Controllers/TerminalController.cs
```diff
@@ -15,198 +15,192 @@
...
+			var terminalDefs = new Dictionary<Terminal, bool>
+			{
+				{
+					new Terminal
+					{
+						Name = "Windows Terminal",
+						Path = "wt.exe",
+						Arguments = "-d .",
+						Icon = string.Empty,
+					},
```
QuaintMako, line 170: Maybe overkill, but I'd suggest to create a method that construct the Terminal instance with given arguments. In case we want to add someday an argument to the process, we will have only to modify one function rather than three constructors.

QuaintMako (commented):
A few suggestions to make the code a bit cleaner. Otherwise, awesome refactoring!

src/Files.App/Controllers/TerminalController.cs
cinqmilleans: I modified the Terminal class to add constructors and fixed the warnings.

src/Files.App/Controllers/TerminalController.cs
cinqmilleans: I added an unsubscribe after the event.

src/Files.App/Controllers/TerminalController.cs
cinqmilleans: I added a method.

yaichenbaum:
We want to completly get rid of the controller as it doesn't fir the rest of the code architecture.

src/Files.App/DataModels/Terminal.cs
```diff
@@ -4,16 +4,20 @@ namespace Files.App.DataModels
 {
 	public class Terminal
 	{
+		public Terminal() {}
+		public Terminal(string name, string path, string arguments = "", string icon = "")
+			=> (Name, Path, Arguments, Icon) = (name, path, arguments, icon);
```
yaichenbaum, line 9: Can you split onto multiple lines?

cinqmilleans: It's done.

src/Files.App/DataModels/Terminal.cs
```diff
@@ -4,16 +4,27 @@ namespace Files.App.DataModels
 {
 	public class Terminal
 	{
+		public Terminal()
+		{
+		}
+		public Terminal(string name, string path, string arguments = "", string icon = "")
```
yaichenbaum, line 10:
```suggestion
		
		public Terminal(string name, string path, string arguments = "", string icon = "")
```

yaichenbaum (approved):

PR #10365, opened 2022-10-31
Fixed crash that would occur when checking for updates

gave92:
**Resolved / Related Issues**
- Fix crash when checking for updates. IUpdateService methods are supposed to be called from the UI thread.
- Wait for file transfers to complete before terminating process
- Restored call to UpdateTagsDb()

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/App.xaml.cs
+++ b/src/Files.App/App.xaml.cs
@@ -283,6 +283,8 @@ namespace Files.App
 		{
 			// Save application state and stop any background activity
 
+			await Task.Yield(); // Method can take a long time, make sure the window is hidden
+
 			SaveSessionTabs();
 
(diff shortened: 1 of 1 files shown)


hez2010:
I don't think all code in `InitializeAppComponentsAsync` should run on the UI thread. This will make the Window slower to show.
How about making IUpdateService only run on the UI thread?

yaichenbaum:
The check for update code could also be deferred until everything else is loaded.

gave92:
The code in initializeblabla() is not called all on ui thread. There's a Task.Run in there. Update check is deferred untill everything else is loaded.

hez2010 (approved):
LGTM

PR #10384, opened 2022-11-02
Fix: Resolving soft-lock issue in operations.

### LINKED ISSUE #10383: Soft-locking operations when executing at least two operations after each other
### Description

We can soft-lock Files' operations (copy, paste, empty etc...) pretty easily. It's pretty detrimental, since it forces the user to reset the software when they want to do more than two operations in a row.

### Steps To Reproduce

1. Copy any file in a folder. It works.
2. Copy another file in the same folder. The operation doesn't end, yet it appears in the `Ongoing Tasks` pane.

### Steps to fix

- Resolve the soft-locking of the operations.

### Files Version

Version 2.3.57.0

### Windows Version

W11 21H2 22000.1098

### Relevant Assets/Logs

SoftLock

QuaintMako:
**Resolved / Related Issues**
- Closes #10383 

**What has been done**
- Moved back the `WaitForCompletion()` method inside `EnsureSettingsAndConfigurationAreBootstrapped()`.

**Validation**
- [X] Built and ran the app

### DIFF
--- a/src/Files.App/App.xaml.cs
+++ b/src/Files.App/App.xaml.cs
@@ -158,8 +158,6 @@ namespace Files.App
 			WSLDistroManager ??= new WSLDistroManager();
 			FileTagsManager ??= new FileTagsManager();
 			SidebarPinnedController ??= new SidebarPinnedController();
-
-			FileOperationsHelpers.Init();
 		}
 
 		private static async Task StartAppCenter()
--- a/src/Files.App/App.xaml.cs
+++ b/src/Files.App/App.xaml.cs
@@ -158,6 +158,8 @@ namespace Files.App
 			WSLDistroManager ??= new WSLDistroManager();
 			FileTagsManager ??= new FileTagsManager();
 			SidebarPinnedController ??= new SidebarPinnedController();
+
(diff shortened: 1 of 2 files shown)


QuaintMako:
I'm gonna be honest I have **no idea** why and how it solves the issue. I'm mostly poking around to solve this issue since it's kind of a bug-breaking problem.

With these informations, I think @gave92 will be able to see to the issue much better than I. (If you are OK with taking a look at it.)

gave92:
Uops didn't mean to force-push here. Anyway @QuaintMako could you test if my commit also solves the issue? The difference is in where the static variables in FileOperationsHelpers get initialized. Still checking why that matters.

gave92:
OK, it matters because if the taskbar object is initialized while inside StartSTATask() then it's created as STA and can't be used in other threads.

yaichenbaum:
Does this impact startup time?

gave92:
Nope!

yaichenbaum (approved):
LGTM

yaichenbaum:
Thought I saw app.xaml.cs in there 😂

PR #10385, opened 2022-11-02
Fix: Fixed crash that would occur when holding down ctrl + t

### LINKED ISSUE #2145: Bug: App crashes when opening a lot of tabs
**Steps to reproduce**
- Open many tabs very quickly by holding CTRL + T.

**Requirements**
Fix crash when opening many tabs

[debug_fulltrust.txt](https://github.com/files-community/Files/files/5434978/debug_fulltrust.txt)
[debug.txt](https://github.com/files-community/Files/files/5434979/debug.txt)

QuaintMako:
**Resolved / Related Issues**
- Closes #2145 

**Validation**
- [x] Built and ran the app


ferrariofilippo (approved):
LGTM
I tested this and it solved the crash

yaichenbaum:
It solves the crash but I'm wondering if we can have this logic apply to all keyboard shortcuts.

QuaintMako:
> It solves the crash but I'm wondering if we can have this logic apply to all keyboard shortcuts.

We may be able to do that by extending #10416 and developping a kind of shortcut executer side by side.

yaichenbaum:
Would it make sense to target the branch that has @cinqmilleans work?

cinqmilleans:
The problem should no longer arise with the new hotkey management.
Before: The shortcut (ctrl + 2, ...) is added when the tab is created. I guess this is limited to 9 or 10 tabs.
After: The shortcut is created when a hotkey is assigned. If the user (or the default) assigned ctrl+2 to the command ShowTab2 (example), the action is assigned even if tab 2 does not exist. When the hotkey is triggered, the action tries to display tab 2 if it exists, otherwise nothing. Hotkeys are only modified when the hotkeys (and not the command) is modified.

We can validate this pr which will correct the problem immediately (I have not tested it).

gave92:
Wondering if just changing AddNewInstanceAcceleratorCommand from RelayCommand to AsyncRelayCommand and making AddNewInstanceAccelerator `async Task` would solve the issue.

QuaintMako:
Does not longer fix the issue. See #2145.

PR #10392, opened 2022-11-03
Revert "Fixed issue with Files launcher not getting copied to output …

yaichenbaum:


gave92 (approved):

gave92:
The project format has not changed. Why should it cause issue when installing? Was that project being built before? Need to check an azure build where the launcher was working properly.

yaichenbaum:
It stopped working when we switched to WinAppSdk (it worked when we removed ARM64). It started working again when this was removed (accidentally).

gave92:
We still need this to fix the launcher though.. so is it an issue of building the FilesLuancher project for ARM64 then?

yaichenbaum:
That's what it looks like, can we use a regular csproj for this?

gave92:
CSproj if for CSharp, vcproj is for c++. Not sure why would the project format matter..

yaichenbaum:
It's also possible the installation issue was from the changes to the pipeline

gave92:
Perhaps we could go back to not build that project and just include the FilesLauncher.exe in the repository

yaichenbaum:
Sounds like a good solution

PR #10403, opened 2022-11-04
Fix: Fixed accessibility tests

yaichenbaum:
- Fixed issues with the grid splitters for the details layout
- Added folder creation test
- Fixed accessibility issues in the settings dialog


tests/Files.InteractionTests/Tests/CreateFolderTest.cs
```diff
@@ -0,0 +1,60 @@
...
+			TestHelper.InvokeButtonByName("Folder");
+
+			// Check for axe issues in the prompt
+			AxeHelper.AssertNoAccessibilityErrors();
+
+			// Type the folder name
+			action.SendKeys("Test Folder").Build().Perform();
+
+			// Click on the set name button
+			TestHelper.InvokeButtonByName("Set Name");
```
marcelwgn: This test only works for me if "Name" is in lower case, @yair100 does the test work for you this way?

chingucoding (changes requested):

tests/Files.InteractionTests/Tests/CreateFolderTest.cs
yaichenbaum: It works but it still might be a good idea for me to switch to ID.

chingucoding (approved):
LGTM

PR #10405, opened 2022-11-04
Implement Oem plus key for Alt+Shift++ (New pane) shortcut

### LINKED ISSUE #9508: Bug: Alt+Shift++" shortcut only works with Numpad Plus key
### What's the Problem?

It seems that the **New Pane**'s shortcut <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>+</kbd> only works with Numpad plus key. The other <kbd>+</kbd> key near the <kbd>Backspace</kbd> isn't working.
image

---
_(This is just an extra)_
Currently, the **New Pane** & **Close Pane**'s shortcut keys are so far away from each other.
|Name| Shortcut|
| - | - | 
| New Pane | <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>+</kbd>  |
| Close Pane | <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>W</kbd>  |

### Solution/Idea

Add support for <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>+ (Plus)</kbd> from number row, not just for NumpadPlus key

### Alternatives

I would suggest to rebind the shortcut key for **New Pane** nearer to **Close Pane**'s <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>W</kbd> hotkey.
For example: <kbd>Shift</kbd>+<kbd>W</kbd>

### Priorities

_No response_

### Files Version

2.3.1.0

### Windows Version

10.0.22621.160

### Comments

_No response_

manfromarce:
The New Pane keyboard shortcut (Alt+Shift++) is now also executed with the "oem" plus key, in addition to the numpad plus key.

**Resolved / Related Issues**
- Closes #9508 

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Views/PaneHolderPage.xaml.cs
+++ b/src/Files.App/Views/PaneHolderPage.xaml.cs
@@ -194,7 +194,9 @@ namespace Files.App.Views
 			}
 		}
 
-		public PaneHolderPage()
+        public readonly VirtualKey PlusKey = (VirtualKey)187;
+
+        public PaneHolderPage()
 		{
 			this.InitializeComponent();
 			App.Window.SizeChanged += Current_SizeChanged;
@@ -322,6 +324,7 @@ namespace Files.App.Views
 					break;
 
 				case (false, true, true, VirtualKey.Add): // alt + shift + "+" open pane
+				case (false, true, true, (VirtualKey)187):
 					if (UserSettingsService.MultitaskingSettingsService.IsDualPaneEnabled)
 					{
 						if (string.IsNullOrEmpty(NavParamsRight?.NavPath))
--- a/src/Files.App/Views/PaneHolderPage.xaml
+++ b/src/Files.App/Views/PaneHolderPage.xaml
@@ -181,5 +181,9 @@
 			Key="Add"
 			Invoked="KeyboardAccelerator_Invoked"
 			Modifiers="Menu,Shift" />
-	</Page.KeyboardAccelerators>
+        <KeyboardAccelerator
+			Key="{x:Bind PlusKey}"
+			Invoked="KeyboardAccelerator_Invoked"
+			Modifiers="Menu,Shift" />
+    </Page.KeyboardAccelerators>
 </Page>


src/Files.App/Views/PaneHolderPage.xaml.cs
```diff
@@ -194,7 +194,9 @@ public bool IsCurrentInstance
 			}
 		}
 
-		public PaneHolderPage()
+        public readonly VirtualKey PlusKey = (VirtualKey)187;
```
QuaintMako, line 197: I suppose the variable here is used to be able to bind it in the .xaml file. But would it be possible to use a simple key, like in l.181 of `PaneHolderPage.xaml`?

QuaintMako (commented):
Suggested a few changes to make the code cleaner, thanks a lot for your contribution!

I cannot test the function on Files for the moment sadly.

src/Files.App/Views/PaneHolderPage.xaml.cs
```diff
@@ -322,6 +324,7 @@ private void KeyboardAccelerator_Invoked(KeyboardAccelerator sender, KeyboardAcc
 					break;
 
 				case (false, true, true, VirtualKey.Add): // alt + shift + "+" open pane
+				case (false, true, true, (VirtualKey)187):
```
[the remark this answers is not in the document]
manfromarce: Unfortunately the [VirtualKey enum](https://learn.microsoft.com/en-us/uwp/api/windows.system.virtualkey?view=winrt-22621) doesn't have Plus or minus but only Add and Subtract.

src/Files.App/Views/PaneHolderPage.xaml.cs
manfromarce: The Key property must be a VirtualKey, so again it doesn't support the plus key directly. I don't know if there is a cleaner workaround.

src/Files.App/Views/PaneHolderPage.xaml.cs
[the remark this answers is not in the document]
QuaintMako: In that case, does the switch still works if you use the variable you defined line 197? So we don't repeat the cast.

src/Files.App/Views/PaneHolderPage.xaml.cs
QuaintMako: Not that I can think of sadly.

src/Files.App/Views/PaneHolderPage.xaml.cs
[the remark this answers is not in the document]
manfromarce: It didn't work because it is not seen as a constant value, but I've found another way to bind so that we only have one cast.

QuaintMako (dismissed):
LGTM. Thanks for the changes!

src/Files.App/Views/PaneHolderPage.xaml.cs
```diff
@@ -322,6 +324,7 @@ private void KeyboardAccelerator_Invoked(KeyboardAccelerator sender, KeyboardAcc
 					break;
 
 				case (false, true, true, VirtualKey.Add): // alt + shift + "+" open pane
+				case (false, true, true, PlusKey):
```
yaichenbaum, line 327:
```suggestion
				case (false, true, true, (VirtualKey)187;):
```

Is a constant needed if it's only used once?

QuaintMako: It is used as well in the `PaneHolderPage.xaml`

yaichenbaum (changes requested):
Code looks good but there is a coupe of spacing issues, can you make sure you have the file set to use tabs?

manfromarce:
> Code looks good but there is a coupe of spacing issues, can you make sure you have the file set to use tabs?

Yes, I didn't notice it was "mixed". It should be fixed now

yaichenbaum (approved):
LGTM

PR #10413, opened 2022-11-06
Feature: Show modal when extracting password protected archives

### LINKED ISSUE #9948: Feature: Show modal when extracting password protected archives
**Requirements**
- Display modal with a password field
- Add cancel option
- Add extract archive option

ferrariofilippo:
**Resolved / Related Issues**
- Closes #9948 

**Change List**
- User is prompted for password when extracting an encrypted archive
- I created a method that tells if an archive is encrypted
- I moved the creation of `destinatioFolder` to prevent the folder from being created even when the user presses `Cancel`
- I added translations for the `Extract` button

**Others**
- There is a bug with `7z` archives, but it also exists in the release version (I'm opening a different issue to track that)

**Comments**
Note that with the `SevenZip` library we can pass the password only in the constructor

**Validation**
- [x] Built and ran the app

Dialog2
Dialog1
Dialog3


yaichenbaum:
Do you think we can reuse the existing "extract to" dialog and add the logic for this?

ferrariofilippo:
> Do you think we can reuse the existing "extract to" dialog and add the logic for this?

I think I can add the password field to the dialog and make it visible only when the archive is encrypted. Is that ok?

yaichenbaum:
I think that's acceptable, we'll probably add additional optiosn in the future so having it one dialog will be helpful.

ferrariofilippo:
Now when the archive is encrypted it looks like this

Dialog

yaichenbaum:
I'd expect the other options to only show if you click "Extract To"

ferrariofilippo:
> I'd expect the other options to only show if you click "Extract To"

It works like that

yaichenbaum:
I think we'll need to make a couple of changes to the dialog to make it more generic
1. Change the title to "Extract Archive"
2. Move the check box to the last row

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2008,7 +2008,7 @@
     <value>View documentation.</value>
   </data>
   <data name="DecompressArchiveDialog.Title" xml:space="preserve">
```
yaichenbaum, line 2010:
```suggestion
  <data name="Extract Archive" xml:space="preserve">
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2856,4 +2856,7 @@
   <data name="BinEmptyingFailed" xml:space="preserve">
     <value>Failed to empty recycle bin</value>
   </data>
+  <data name="ArchivePasswordDialogDescription" xml:space="preserve">
+    <value>Insert archive password</value>
```
yaichenbaum, line 2860:
```suggestion
  <data name="ArchivePassword" xml:space="preserve">
    <value>Archive password</value>
```

We should probably add this header above the text box

ferrariofilippo: What if I set it as the placeholder for the text box? It would be much cleaner than the header

yaichenbaum: It's usually recommended to have a header as placeholders are only good as long as the text box is empty.

yaichenbaum (dismissed):
Looks good UI wise.

yaichenbaum (approved):
LGTM

Issue #9559, opened 2022-07-13
Feature: Add support for key remapping


yair100:
### Solution/Idea

As Files gets more advanced and the number of shortcuts keeps growing, it'll be useful to allow keyboard remapping to allow users to configure shortcuts to work the way they like it.

![image](https://github.com/files-community/Files/assets/39923744/f9e173ce-c229-4ee3-b08c-60f695f186c6)

### Requirement

This proposal will add a settings page called "Actions"
This page will contain a list of actions
This proposal will allow configuring the shortcut or gesture to configure each action
This proposal will allow users to reset all actions to the default

### Files Version

v2.3

cinqmilleans:
Here is a custom KeyPicker that I could integrate into Files. What do you think?

https://user-images.githubusercontent.com/46631671/179416381-91b554f2-69b4-4a67-90ec-a1b8e063ede1.mp4

yaichenbaum:
Something like that, I need to create the design but you should be able to work on the functionality in the meantime.

yaira2:
Remapping actions is already supported if you manually edit the settings file
1. Open Settings > Advanced > Edit settings file
2. Right click the system tray icon for Files and click "Quit"
3. Go back to the settings file and find the section for "Actions"
4. The list of available actions can be found [here](https://github.com/files-community/Files/blob/main/docs/rich-commands.md)

Here is an example remapping the copy action to `ctrl` + `j`
```json
"Actions": {
    "CopyItem": "ctrl+j"
 },
```

0x5bfa:
![Image](https://github.com/files-community/Files/assets/62196528/9a0c6f9d-c0a6-4af2-b032-93c36f3b422b)

A page would be great? Working on it.

yaira2:
@0x5bfa nice! I would suggest calling the page "Actions" as it's supposed to be more than just keyboard shortcuts.

yaira2:
The requirements listed above are still accurate but I don't have any specific design in mind for the UI.

yaira2:
A big shoutout to @0x5bfa for completing this large task 🎉

yaira2:
> This proposal will allow configuring the shortcut or gesture to configure each action

Custom gestures are out of scope for this task, but we'll continue exploring to see if there is a path forward for them.

PR #10416, opened 2022-11-06
Draft Key Manager

cinqmilleans:
A requested feature is to be able to configure keyboard shortcuts (existing or to add some to commands that don't have any by default). This is a proposed method to centralize keyboard shortcuts in a manager. This is a preliminary step to then add a configuration page. This pr is a draft, a proof of concept because it was too complex to explain without showing any code. It only implements a few example commands.

Technically,
1) The HotKey structure is a VirtualKey/VirtualKeyModifiers pair, it includes a specific ToString and parser (ex: F1, Ctrl+Shift+1, ...)
2) IKeyboardAction represents an action that can be assigned to hotkey. Each action has its own implementation.
3) (I)KeyboardManager manages the execution of actions and the assignment of actions to custom shortcuts. It is initialized with the list of actions.
4) All shortcuts are moved to MainPage, but KeyboardAccelerators are populated by the manager.
5) Menus can use the manager to display the shortcut in a tooltip or menu entry.
6) Modifying a shortcut applies without rebooting (but current code has no example where it is done).
7) The listview manages some shortcuts themselves. We had to capture the event to prevent it from triggering and call the manager to trigger the shortcut manually. You will have to do this for other actions.
8) We use KeyboardAccelerator to intercept only active shortcuts. It is no longer necessary to find the correct command according to the keys pressed, we have it directly according to the hotkey.

It will be necessary to manage all the commands before managing the storage and the modification of the shortcuts. The manager will have to be divided into 2 classes, one to manage user shortcuts and one to manage the updating of controls.

**Validation**
- [x] Built and ran the app


yaichenbaum:
I think this is the opposite of what we're going for, you shared a UI concept a while back that I really liked, and I think we should be working towards that. Let's continue planning in the relevant issues before continuing with the PR. https://github.com/files-community/Files/issues/9559#issuecomment-1186572147

cinqmilleans:
It will be necessary to store all the hotkeys (user and default) to prevent assigning the same shortcut twice.
It will also be necessary to be able to apply the hotkeys. Currently they are hard-coded in views and using binding does not work for KeyboarAccelerator (no update of a modified observable value). With the current code, when we modify a hotkey, we have to change in several places (the view and the code) and it differs according to the hotkey. Some shortcuts are also duplicated (ModernShellPage and ColumnShellPage).

This pr centralizes the management of hotkeys so that everything is done in one place. Actions have a label, a default value and a code that will be used for storage. It will be very easy then to add the interface containing the pickers.

src/Files.App/App.xaml.cs
```diff
@@ -123,6 +124,7 @@ private IServiceProvider ConfigureServices()
 				.AddSingleton<IThreadingService, ThreadingService>()
 				.AddSingleton<ILocalizationService, LocalizationService>()
 				.AddSingleton<ICloudDetector, CloudDetector>()
+				.AddSingleton<IKeyboardManager, KeyboardManager>()
```
yaichenbaum, line 127:
```suggestion
				.AddSingleton<IKeyboardAcceleratorrManager, KeyboardManager>()
```

src/Files.App/Keyboard/Actions/ClearSelectionAction.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Files.App.ViewModels;
+
+namespace Files.App.Keyboard.Actions
+{
+	internal class ClearSelectionAction : IKeyboardAction
```
yaichenbaum, line 5: Instead of IKeyboardActions, we'd like to have generic command files. It's essentially the same thing but with a more generic interface.

src/Files.App/Keyboard/Actions/ToggleLayoutColumnsAction.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Files.App.Extensions;
+using Files.App.ViewModels;
+using Windows.System;
+
+namespace Files.App.Keyboard.Actions
+{
+	internal class ToggleLayoutColumnsAction : IKeyboardAction
```
yaichenbaum, line 7: We should probably group the layout commands together but doesn't have to be right away as it's easy to make this change.

src/Files.App/Keyboard/Actions/ClearSelectionAction.cs
yaichenbaum: Perhaps this should just be a regular command so that we can use the commands in other areas besides keyboard shortcuts.

src/Files.App/Keyboard/Actions/ToggleLayoutColumnsAction.cs
cinqmilleans: What do you mean ? The parameters should allow you to find an order quickly (filter textbox, categories, ...). Is this?

src/Files.App/Keyboard/Actions/ClearSelectionAction.cs
cinqmilleans: It is an excellent idea. The word command has a specific meaning in .net. Maybe IAction, because we don't have to handle CanExecute and CanExecuteChange, but other properties. This can also be used for certain mouse shortcuts or custom buttons or custom menu entries.

src/Files.App/App.xaml.cs
cinqmilleans: I mentioned in the description that it was in progress. It will indeed be cleaner with 2 managers. I mentioned in the description that it was in progress. It will indeed be cleaner with 2 managers. I will do this this weekend at the latest.

src/Files.App/Keyboard/Actions/ClearSelectionAction.cs
yaichenbaum: @d2dyno1

src/Files.App/Keyboard/Actions/ToggleLayoutColumnsAction.cs
yaichenbaum: Meaning the class should be `ToggleLayoutCommands` and there should be separate methods for `Columns` and `Details` etc.

cinqmilleans: This goes against the interface and forces the manager to make special cases for the various orders. What's wrong with having one class per action? If we add a layout, it will suffice to add an action, without the manager knowing anything about it. There will be hundreds of actions, all with the same minimal interface.

yaichenbaum: There isn't anything wrong but it's also a good idea to group related tasks, at the very least, they should have their own folder.

cinqmilleans: Yes I will add folders.

cinqmilleans:
@yair100 @d2dyno1 
I continued my analysis (thank you for your comments) and here are some food for thought.

Problems with the current implementation.
1) It is not possible to customize keyboard shortcuts, menus or buttons.
2) Actions are not centralized. Modifying a glyph requires modifying it everywhere (contextual menu, toolbar, ...)

The problems of this draft.
1) The execution is done in the action and the action is triggered globally. This forces the action to know all the context to know who to do (for example in which panel).
2) It forces to provide the context before the call because the call has no argument.
3) Displaying action data in controls is complex because controls do not have direct access to actions.

Proposed solution.
1) IAction describes an action without its behavior (code, label, description, hotkey, glyph, ...). Each action has its implementation.
2) The IActionViewModel interface (which has only one implementation) provides an action for the UI (label, command, icon, keyboardaccellerator, ...).
3) The IActionManager service contains all the ActionViewModel (one per property). The UI uses it to display the labels of buttons, menus, fill in keyboard shortcuts, ... No more repeating icons or labels.
4) This service has a method to trigger an action (by its name) and an event triggered at each action.
Components subscribe to the event and optionally perform the action (and stop the propagation).

With this system, we can keep the processing of actions closer to their location. This system can be implemented gradually.

cinqmilleans:
The method proposed in my last post poses insurmountable problems. I advanced on the functionality by simplifying the code and using commands. I will update soon.

yaira2:
@cinqmilleans

PR #10418, opened 2022-11-07
Tests: Added copy paste test

yaichenbaum:
**Resolved / Related Issues**
- Added a test to copy and paste a folder

**Validation**
- [x] Built and ran the app
- [x] Ran the test project

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/tests/Files.InteractionTests/Tests/FolderTests.cs
+++ b/tests/Files.InteractionTests/Tests/FolderTests.cs
@@ -1,6 +1,8 @@
 ﻿using OpenQA.Selenium;
 using OpenQA.Selenium.Interactions;
+using System;
 using System.Threading;
+using static System.Collections.Specialized.BitVector32;
 
 namespace Files.InteractionTests.Tests
 {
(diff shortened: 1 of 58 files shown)


tests/Files.InteractionTests/Tests/FolderTests.cs
```diff
@@ -12,34 +12,7 @@ public class FolderTests
...
-
-			// Press the enter key
-			action = new Actions(SessionManager.Session);
-			action.SendKeys(Keys.Enter).Perform();
-
-			// Wait for items to be deleted
-			Thread.Sleep(1000);
-
-
+		{
```
chingucoding, line 15: Should we still call the folder test in the cleanup to be sure?

yaichenbaum: What do you mean?

chingucoding: If the test crashes for some reason, we would still leave the test folder on disk right? Calling the delete folder test could help us clean that up. Alternatively, if we know for sure where the test folder is created, we could also delete it using C# file API.

chingucoding (approved):

Issue #10359, opened 2022-10-31
Code Quality: Dead code and types


QuaintMako:
### Description

There are a bunch of dead code and types in the source code, probably reminiscent of former structures.  While it's not a huge issue in itself, it does create some clutter around the codebase.

### Concerned code

Whole solution

### Gains

- A cleaner codebase.
- A better application of the YAGNI principle.

### Requirements

- Inspect the code and remove methods and objects that are artifacts of former structure.

QuaintMako:
Running some tools can help us find the dead code pretty easily, it shouldn't take a lot of work to remove.

yaichenbaum:
> Running some tools can help us find the dead code pretty easily, it shouldn't take a lot of work to remove.

We need to evaluate each case because we sometimes keep things around for reference. Are you interested in opening a PR and I'll go over it to see if there is anything we want to keep?

thebarshablog:
I'd like to take care of this issue, incase it has not been assigned.

yaira2:
It's preferred to make these changes in increments (usually when you're already working on the file with the code) as opposed to a single PR that takes care of it all at once.

thebarshablog:
For sure, I'll keep tab of each thing that requires attention and will do so according.

yaira2:
@QuaintMako do we need a dedicated issue to track this or should it just be a general goal while working on the codebase?

QuaintMako:
I think we can close it now.

PR #10442, opened 2022-11-09
Code Quality : Removal of dead code and types

QuaintMako:
**Resolved / Related Issues**
- Related to #10359 

**What has been done**
- Removed classes, methods and fields used nowhere.

**Validation**
- [X] Built and ran the app


src/Files.App/Converters/Converters.cs
```diff
@@ -85,37 +85,6 @@ public abstract class ValueConverter<TSource, TTarget>
 		}
 	}
 
-	/// <summary>
-	/// The base class for converting instances of type T to object and vice versa.
-	/// </summary>
-	public abstract class ToObjectConverter<T>
```
yaichenbaum, line 91: We should probably keep these, what do you think?

src/Files.App/Helpers/ContextFlyoutItemHelper.cs
```diff
@@ -47,8 +47,8 @@ public static Task<List<ContextMenuFlyoutItemViewModel>> GetBaseContextShellComm
 
 		public static List<ContextMenuFlyoutItemViewModel> Filter(List<ContextMenuFlyoutItemViewModel> items, List<ListedItem> selectedItems, bool shiftPressed, CurrentInstanceViewModel currentInstanceViewModel, bool removeOverflowMenu = true)
 		{
-			items = items.Where(x => Check(item: x, currentInstanceViewModel: currentInstanceViewModel, selectedItems: selectedItems, shiftPressed: shiftPressed)).ToList();
-			items.ForEach(x => x.Items = x.Items?.Where(y => Check(item: y, currentInstanceViewModel: currentInstanceViewModel, selectedItems: selectedItems, shiftPressed: shiftPressed)).ToList());
+			items = items.Where(x => Check(item: x, currentInstanceViewModel: currentInstanceViewModel, selectedItems: selectedItems)).ToList();
```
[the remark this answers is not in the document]
QuaintMako: This would be an error from me. Needed to remove only an unused parameter, not the whole line. I'll fix it as soon as possible.

src/Files.App/Converters/Converters.cs
QuaintMako: Let's restore them in that case. I tried to not remove unused code that would have real value, but this one has eluded me.

QuaintMako: Added back.

src/Files.App/Helpers/PackageHelper.cs
```diff
@@ -1,47 +0,0 @@
-using System;
-using System.Diagnostics;
-using System.Threading.Tasks;
-using Windows.System;
-
-namespace Files.App.Helpers
```
yaichenbaum, line 6: Let's restore this helper

yaira2 (commented):
LGTM

PR #10473, opened 2022-11-15
Fix: Fixed issue where the context menu would sometimes stay open

cinqmilleans:
**Resolved / Related Issues**
The context menu does not close when executing a command in a submenu. This pr fixes this.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before

https://user-images.githubusercontent.com/46631671/201942336-8f8e8b2e-6f8f-41dd-b489-ea64ba8a1c2c.mp4

After

https://user-images.githubusercontent.com/46631671/201942330-027bb914-e128-4af0-b895-d9d03a06f798.mp4


src/Files.App/BaseLayout.cs
```diff
@@ -587,14 +596,23 @@ private async Task LoadMenuItemsAsync()
...
+			primaryElements
+				.OfType<AppBarButton>()
+				.ForEach(button => button.Click += closeHandler);
+			secondaryElements
+				.OfType<AppBarButton>()
+				.Select(item => item.Flyout)
+				.OfType<MenuFlyout>()
+				.SelectMany(menu => menu.Items)
+				.OfType<MenuFlyoutItem>()
+				.ForEach(button => button.Click += closeHandler);
```
QuaintMako, line 614: Those lines are a repetition of l.541 to l.553.
Would it be possible to fuse them inside a method?

cinqmilleans: It's done.

src/Files.App/BaseLayout.cs
```diff
@@ -606,6 +600,24 @@ private async Task LoadMenuItemsAsync()
 			}
 		}
 
+		private void AddCloseHandler(IList<ICommandBarElement> primaryElements, IList<ICommandBarElement> secondaryElements)
+		{
+			// Workaround for WinUI (#5508)
```
yaira2, line 605: Can you include a link to the issue?

cinqmilleans: The comment comes from a previous fix for primary elements. This pr fixes the submenus of the secondary elements. Is it worth keeping this comment? What link should I put?

yaira2: If it's still a workaround, we should link to the issue on WinUI

PR #10484, opened 2022-11-15
Fix: Fixed issue where nav menu had extra margin in the properties window

cinqmilleans:
**Resolved / Related Issues**
Properties tabs use an unnecessary right margin. Some tabs do not appear. This removes the margin.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Before
PropertyBefore
After
PropertyAfter

### DIFF
--- a/src/Files.App/Views/Pages/Properties.xaml
+++ b/src/Files.App/Views/Pages/Properties.xaml
@@ -39,7 +39,7 @@
 		<NavigationView
 			x:Name="NavigationView"
 			Grid.Row="0"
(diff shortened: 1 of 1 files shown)


yaira2:
The margin is to provide a drag region.

cinqmilleans:
I reduced the margin from 120 to 32. This seems sufficient to me and often allows to have the 3 tabs instead of 2.

yaira2:
The number 120 is to match the expected behavior in Windows.

PR #10505, opened 2022-11-17
Fix: Fixed issue where 7z archives were not extracted correctly

### LINKED ISSUE #10414: Bug: 7z archives are not extracted correctly
### Description

When I try to extract a 7z archive, only the base folder is extracted.

The problem is related to line # 63 of `Files.App.Helpers.ZipHelpers`.
The base folder is not considered as a directory. 
Because of this line # 124 is evaluated to true and the method returns.

### Steps To Reproduce

1. Start the extraction of a 7z archive
2. See that only the base folder is created

### Steps to fix

?

### Files Version

### Windows Version

### Relevant Assets/Logs

Folder

`xStage` is a folder but `IsDirectory` is set to false

ferrariofilippo:
**Resolved / Related Issues**
- Closes #10414

**Description**
- 7zip library doesn't distinct files and firectories. Because of this, they are all added to `fileEntries`
- I fixed this by removing the directories from the list after they are created

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Helpers/ZipHelpers.cs
+++ b/src/Files.App/Helpers/ZipHelpers.cs
@@ -74,8 +74,8 @@ namespace Files.App.Helpers
 			if (zipFile is null)
 				return;
 			//zipFile.IsStreamOwner = true;
-			List<ArchiveFileInfo> directoryEntries = new List<ArchiveFileInfo>();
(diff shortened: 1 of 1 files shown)


yaira2 (approved):
LGTM

PR #10513, opened 2022-11-19
Feature: Allow resizing the status column

### LINKED ISSUE #10432: Feature: Allow resizing the status column
### Description

`Status` column cannot be resized by the user.

### Steps To Reproduce

1. Open OneDrive folder
2. Try to resize `Status` column

### Steps to fix

User should be able resize Status column

### Files Version

### Windows Version

### Relevant Assets/Logs

Headers

ferrariofilippo:
**Resolved / Related Issues**
- Closes #10432 

**Changes**
- I added a column with `width=0`, since I discovered that you can't resize the last column
- I removed `BsonIgnore` from `NormalMaxLength`, as this prevented length constraints from being applied to custom folders

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/ViewModels/ColumnsViewModel.cs
+++ b/src/Files.App/ViewModels/ColumnsViewModel.cs
@@ -188,7 +188,6 @@ namespace Files.App.ViewModels
 
 		private double normalMaxLength = 800;
 
-		[LiteDB.BsonIgnore]
 		public double NormalMaxLength
 		{
 			get => normalMaxLength;
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
@@ -339,7 +339,8 @@
 									x:Name="Column10"
 									Width="{x:Bind ColumnsViewModel.StatusColumn.Length, Mode=OneWay}"
 									MaxWidth="{x:Bind ColumnsViewModel.StatusColumn.MaxLength, Mode=OneWay}" />
-								<ColumnDefinition x:Name="Column11" Width="0" />
+								<ColumnDefinition Width="Auto" />
+								<ColumnDefinition Width="0" />
 							</Grid.ColumnDefinitions>
 
 							<Grid Grid.Column="0" />


src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -339,7 +339,8 @@
 									x:Name="Column10"
 									Width="{x:Bind ColumnsViewModel.StatusColumn.Length, Mode=OneWay}"
 									MaxWidth="{x:Bind ColumnsViewModel.StatusColumn.MaxLength, Mode=OneWay}" />
-								<ColumnDefinition x:Name="Column11" Width="0" />
+								<ColumnDefinition Width="Auto" />
```
yaira2, line 342: This doesn't seem correct, does the size get saved to the database?

ferrariofilippo: Yes it does. That line defines the grid splitter column

yaira2: Is the column supposed to have a name?

ferrariofilippo: I'll remove it since it's not needed

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -339,7 +339,8 @@
...
+								<ColumnDefinition Width="0" />
```
yaira2, line 343: A little confusing because there are a lot of other columns, but why do the others bind to the view model and this one just has a hardcoded width?

ferrariofilippo: The column with `width=0` is needed since you can't actually resize the last column of a grid. That column will never have a width different from `0`. It's only purpose is to be there so that we can resize the last useful column (`status`)

src/Files.App/ViewModels/ColumnsViewModel.cs
```diff
@@ -188,7 +188,6 @@ public double MaxLength
 
 		private double normalMaxLength = 800;
 
-		[LiteDB.BsonIgnore]
```
yaira2, line 191: Was this supposed to be removed?

yaira2 (approved):
LGTM

PR #10524, opened 2022-11-20
Feature: Added advanced options when creating archives

### LINKED ISSUE #10327: Feature: Add compression options

Currently, data compression has no options other than file name. Some options are very practical (7z format, compression level, password, ...). Having all this integrated adds significant added value (no need to install third-party software, better integrated context menu, etc.). However, it must be kept simple so as not to confuse the user.

The compression popup only asks for the file name. We could add a button to display options. These options should be explained, for example to understand the advantages and disadvantages of each format.

### Requirements

- Right clicking on any item (or multiple items) displays a menu item to "Add to archive"
   - Clicking this menu item displays a sub flyout with an option to add "folder.zip" (folder is the name of the parent directory
   - There should be a second item in this sub flyout that says "Add to archive". This displays a popup where the user can choose the format and other advanced options.
- Advanced options modal should have the title "Add to archive"
   - Users can choose the archive name
   - Users can pick the format
   - Users can set a password
   - Users can choose the compression level

### Files Version

Version : 2.3.52.0 OS Version: 10.0.22621.675

### Windows Version

Windows 11 22621.674

### Comments

I can make an example view before implementation to confirm the concept.

cinqmilleans:
**Resolved / Related Issues**
This pr improves the creation of compressed archives. Closes #10327.
1) The dialog has been redone and adds additional possibilities. I tried to stay practical and not technical.
2) The context menu has been unified with the decompression and offers additional possibilities.
3) There is no more distinction of the name of the archive according to whether one compresses an element or several.
4) The progress percentage indication is updated at each percent. Previously, it could never be updated for a large folder.
5) The compression code has been moved and is no longer static. It should be moved to the backend but the library needs to be updated first.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Archive3
Archive4
Archive5
Archive1
Archive2


yaira2:
image

I have a couple of questions,
1. Is there a reason for having compress when selecting an existing archive?
2. Can the split options be hidden when they aren't available?
3. Icons in the dialog aren't necessary, can they be removed?
4. What does the "Do not compress" option do?
5. Using `x:Load` instead of `Visibility` is a good idea and will automatically fix the extra space at the bottom of the dialog.

src/Files.App/Dialogs/CompressArchiveDialog.xaml
```diff
@@ -0,0 +1,169 @@
+﻿<ContentDialog
+	x:Class="Files.App.Dialogs.CompressArchiveDialog"
```
yaira2, line 2: Should this be called "CreateArchiveDialog"?

cinqmilleans:
1) I changed the rule. The Compression part no longer appears when the selection is a single file which is an archive.
2) It's changed.
3) Icons have been removed.
4) There are 6 levels of compression (Ultra, High, Normal, Low, Fast, None). Currently, the compression is always in Ultra. It's handy to have None because it's very fast to store many files in one to transfer or store. Media files or documents are often already compressed but recompressing them is slow and unnecessary. I put back the 6 levels because, after testing, the Ultra level is very slow (40 minutes for my extreme test, against 4 minutes in Normal). The default mode, including without going through the dialog, is Normal because it is the best compression/speed compromise. We can optionally not display each level for simplicity (Ultra/Normal/Fast/None).
5) The extra space was intentional to avoid changing the window size. I changed the behavior to no longer have that empty space. I applied x:Load.

Archive6
Archive7

src/Files.App/Dialogs/CompressArchiveDialog.xaml
cinqmilleans: It's done.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
```diff
@@ -0,0 +1,164 @@
...
+								TextWrapping="Wrap"
+								Visibility="{Binding IsDropDownOpen, ElementName=FileFormatSelector}" />
+						</Grid>
+					</DataTemplate>
+				</ComboBox.ItemTemplate>
+			</ComboBox>
+		</Grid>
+
+		<StackPanel Spacing="4">
+			<Grid Style="{StaticResource GridSettingStyle}">
```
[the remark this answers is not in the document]
cinqmilleans: This grid is used to display the background of each setting, as in the UserControl SettingsBlockControl. Using the style on the UserControl does not work.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
```diff
@@ -0,0 +1,164 @@
...
+						DisplayMemberPath="Label"
+						ItemsSource="{x:Bind CompressionLevels}"
+						SelectedValue="{x:Bind CompressionLevel, Mode=TwoWay}"
+						SelectedValuePath="Key" />
+				</settings:SettingsDisplayControl>
+			</Grid>
+			<Grid
+				x:Name="SplittingSizeLayout"
+				Style="{StaticResource GridSettingStyle}"
+				Visibility="Collapsed">
```
yaira2, line 105: Let's switch to `x:Load` and bind to a property in the ViewModel.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2859,4 +2859,76 @@
   <data name="ExtractToPath" xml:space="preserve">
     <value>Path</value>
   </data>
+  <data name="Archive" xml:space="preserve">
+    <value>Archive</value>
+  </data>
+  <data name="CreateArchive" xml:space="preserve">
+    <value>Create a archive</value>
```
yaira2, line 2866:
```suggestion
    <value>Create archive</value>
```

yaira2 (changes requested):

src/Files.App/Strings/en-US/Resources.resw
cinqmilleans: It's done.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
cinqmilleans: It's done

hez2010:
Will do the review tomorrow. I'm running out of bandwidth today :)

cinqmilleans:
A few minor edits.
1) Opening the expander now automatically gives focus to the password field.
2) Combobox are too short for SplittingSize texts. To correct this problem, the combobox no longer displays the media when it is closed. The size becomes a minimum size because the "Do not split" text will be longer in many languages and it was truncated.

Other available formats are GZip, BZip2, Tar and XZ. I don't really know these formats but only Tar is managed by Files in decompression.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
```diff
@@ -0,0 +1,160 @@
...
+			<TextBlock
+				Grid.Row="0"
+				Grid.Column="0"
+				VerticalAlignment="Center"
+				Text="{helpers:ResourceString Name=Format}" />
+			<ComboBox
+				x:Name="FileFormatSelector"
+				Grid.Row="0"
+				Grid.Column="1"
+				MinWidth="140"
```
yaira2, line 76: Can you hardcode a width instead? `160` is probably enough.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
```diff
@@ -0,0 +1,160 @@
...
+				IsEnabled="{x:Bind ViewModel.CanSplit, Mode=OneWay}"
+				ItemsSource="{x:Bind ViewModel.SplittingSizes}"
+				SelectedItem="{x:Bind ViewModel.SplittingSize, Mode=TwoWay}"
+				SelectedValuePath="Key">
+				<ComboBox.ItemTemplate>
+					<DataTemplate>
+						<StackPanel Orientation="Horizontal" Spacing="4">
+							<TextBlock Text="{Binding Label}" />
+							<TextBlock Text="{Binding Separator}" Visibility="{Binding IsDropDownOpen, ElementName=SplittingSizeSelector}" />
+							<TextBlock Text="{Binding Description}" Visibility="{Binding IsDropDownOpen, ElementName=SplittingSizeSelector}" />
```
yaira2, line 125: Is description really needed? Files supports all these formats, so it seems a little redundant.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
cinqmilleans: It's done.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
cinqmilleans: The description concerns the type of media (cd, dvd, ...), not the format. I used a description to avoid the text being too long when closed.

Archive8
Archive9

yaira2:
> Other available formats are GZip, BZip2, Tar and XZ. I don't really know these formats but only Tar is managed by Files in decompression.

Can we add support for rar and tar?

cinqmilleans:
You can add tar easily. For rar, you need another library, but also check the licenses. Free compression/decompression softwares doesn't seem to support it.

yaira2:
Let's support whatever the current library supports 👍

src/Files.App/Dialogs/CreateArchiveDialog.xaml
```diff
@@ -0,0 +1,160 @@
...
```
hez2010, line 125: Can you use `x:Bind` instead? You can set `x:DataType` on `DateTemplate` to enable `x:Bind` here.

src/Files.App/Files.App.csproj
```diff
@@ -66,6 +66,9 @@
     </Content>
     <Content Remove="Assets\FilesHome.png" />
   </ItemGroup>
+  <ItemGroup>
+    <None Remove="Dialogs\CreateArchiveDialog.xaml" />
+  </ItemGroup>
```
hez2010, line 71: It's redundant.

src/Files.App/Files.App.csproj
```diff
@@ -120,4 +123,9 @@
     <TrimmerRootAssembly Include="CommunityToolkit.WinUI.UI.Controls.Media" />
     <TrimmerRootAssembly Include="CommunityToolkit.WinUI.UI.Controls.Primitives" />
   </ItemGroup>
+  <ItemGroup>
+    <Page Update="Dialogs\CreateArchiveDialog.xaml">
+      <XamlRuntime>$(DefaultXamlRuntime)</XamlRuntime>
+    </Page>
+  </ItemGroup>
```
hez2010, line 130: It's redundant.

src/Files.App/Filesystem/Archive/ArchiveCreator.cs
```diff
@@ -0,0 +1,119 @@
...
+				EncryptHeaders = true,
+				PreserveDirectoryRoot = sources.Length > 1,
+				EventSynchronization = EventSynchronizationStrategy.AlwaysAsynchronous,
+			};
+			compressor.Compressing += Compressor_Compressing;
+
+			try
+			{
+				var files = sources.Where(source => File.Exists(source)).ToArray();
+				var directories = sources.Where(source => System.IO.Directory.Exists(source)).ToArray();
```
hez2010, line 93:
```suggestion
				var directories = sources.Where(source => System.IO.Directory.Exists(source));
```

src/Files.App/Interacts/BaseLayoutCommandImplementationModel.cs
```diff
@@ -576,52 +577,120 @@ public async Task CreateFolderWithSelection(RoutedEventArgs e)
...
-					return;
-				archivePath = Path.Combine(
-					associatedInstance.FilesystemViewModel.WorkingDirectory,
-					$"{(string)archiveDialog.ViewModel.AdditionalData}.zip");
-			}
+			if (sources.Length is 0)
+				return (sources, string.Empty, string.Empty);
+
+			string directory = associatedInstance.FilesystemViewModel.WorkingDirectory.Normalize();
+			string fileName = Path.GetFileName(sources.Length is 1 ? sources[0] : directory);
```
hez2010, line 645: nit:

```suggestion
			string fileName = Path.GetFileName(sources is [source] ? source : directory);
```

src/Files.App/Interacts/IBaseLayoutCommandImplementationModel.cs
```diff
@@ -98,6 +98,10 @@ public interface IBaseLayoutCommandImplementationModel : IDisposable
 
 		Task CompressIntoArchive();
 
+		Task CompressIntoZip();
+
+		Task CompressIntoSevenZip();
+
```
hez2010, line 104: Can we merge them into `CompressIntoArchive` which uses a parameter to decide the archive format?

src/Files.App/Interacts/BaseLayoutCommandsViewModel.cs
```diff
@@ -174,6 +176,10 @@ private void InitializeCommands()
 
 		public ICommand CompressIntoArchiveCommand { get; private set; }
 
+		public ICommand CompressIntoZipCommand { get; private set; }
+
+		public ICommand CompressIntoSevenZipCommand { get; private set; }
+
```
hez2010, line 182: Can we merge them into `CompressIntoArchiveCommand `?

hez2010 (changes requested):

src/Files.App/Interacts/BaseLayoutCommandImplementationModel.cs
cinqmilleans: This does not compile as the feature is in preview. Maybe I need to activate something.

src/Files.App/Filesystem/Archive/ArchiveCreator.cs
cinqmilleans: It's done.

src/Files.App/Interacts/IBaseLayoutCommandImplementationModel.cs
cinqmilleans: I prefer to use multiple methods because the role of this parameter would not be intuitive. CompressIntoArchive displays a dialog with several options. CompressIntoZip directly creates the archive without any options. Optionally, we can group CompressIntoZip and CompressIntoSevenZip together. But these functions should correspond to methods (see below).

src/Files.App/Interacts/BaseLayoutCommandsViewModel.cs
cinqmilleans: A command should never use a parameter. In particular this case which concerns very different behaviors. Using commands without parameters is needed to customize hotkeys (another feature in progress).

src/Files.App/Files.App.csproj
cinqmilleans: It's removed.

src/Files.App/Files.App.csproj
cinqmilleans: It's removed.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
cinqmilleans: I don't use DataType because the type is not public but nested to the dialog. This class does not need to be public as it is an implementation detail of the dialog. It is therefore not possible to use this type as a datatype in xaml.

cinqmilleans:
I added support for other formats for compression. These formats do not support splitting. I will try to add them also in unzip (including browse content). This will be done in another pr.

cinqmilleans:
I disabled the other formats. bzip2 and xz don't seem to keep filenames. gz and tar do not support passwords. To decide what to do.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
hez2010: XAML doesn't recognize nested classes, but the class can be non-public. You can move the class out of the contained class and use x:DataType. We did the same for other classes a while ago.

cinqmilleans: It's done.

src/Files.App/Dialogs/CreateArchiveDialog.xaml
```diff
@@ -0,0 +1,161 @@
...
+				x:Name="FileFormatSelector"
+				Grid.Row="0"
+				Grid.Column="1"
+				Width="160"
+				ItemsSource="{x:Bind ViewModel.FileFormats}"
+				SelectedItem="{x:Bind ViewModel.FileFormat, Mode=TwoWay}"
+				SelectedValuePath="Key">
+				<ComboBox.ItemTemplate>
+					<DataTemplate>
+						<TextBlock Text="{Binding Label}" />
```
hez2010, line 83: Still one left over here.

cinqmilleans: This template was not necessary. I removed it. Thank you for your reviews. It's very nice of you.

hez2010 (approved):
LGTM. Thank you!

yaira2 (approved):

PR #10541, opened 2022-11-23
Feature: Wrap file names in the grid layout

### LINKED ISSUE #1033: Feature: Wrap file names in the grid layout
**Requirements**
- File names should wrap up to three lines in the grid view layouts.
- Selected items names should wrap as many lines as necessary to show the complete name

GabrieleMessina:
**Resolved / Related Issues**
- Closes #1033

**Changes**
- Add Selected property to ListedItem class;
- Increased file name TextBlock's height for showing 3 lines when on 100% system text scale;
- Add ScrollViewer to selected ListedItem to be able to read the entire name if it's long;

**Validation**
- [x] Built and ran the app

**Screenshots**
image
image
image


yaira2:
What happens if you have multiple rows? Does the item height increase for all the rows even if there aren't multiple lines of text?

GabrieleMessina:
@yair100 
> What happens if you have multiple rows? Does the item height increase for all the rows even if there aren't multiple lines of text?

Yes, because i believe the row height is given by the first element, not by the highest. So i've set the height of the TextBlock to 60.

image

yaira2:
This is going to be an issue because the first row is taller than it needs to be...

GabrieleMessina:
I understand, but i don't know how to fix this. Do you have any suggestions? 

I've checked, and the AdaptiveGridView doesn't give us any control over row height and i can't access or find the first element for each row because I don't know how many items there are in each one.

yaira2:
I'm not really sure, I've tried different solutions but none of them works as expected.

PR #10554, opened 2022-11-24
Fix: Open archives in existing tab if files is set as default app

### LINKED ISSUE #10529: Feature: Open archives in existing tab

Every time I open an archive it opens in a new tab which I don't want as it stop the back button working which I would expect to take me back to the main folder.

### Requirements

- Double clicking or opening an archive should open it in the existing tab

### Files Version

2.3.61.0

### Windows Version

10.0.22621.875

### Comments

_No response_

mafra99:
**Resolved / Related Issues**
- Closes #10529

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

**Explanation**

Zip files were not recognized as browsable zip because this method returned false if Files is set as default app.

`assoc` contains the full path to the associated app, for example “C:\\Program Files\\WindowsApps\\FilesPreview_2.3.61.0_x64__wvne1zexy08sa\\Files.App\\Files.exe”

The package `FamilyName` is in debug mode for example “FilesDev_ykqwq8d6ps0ag”

I also thought about comparing it with the current executing assembly path or to check if the associated path contains the Package Name, but that would not work in debug mode or if multiple files apps are installed (Store, Sideload, …)

### DIFF
--- a/src/Files.App/Filesystem/StorageItems/ZipStorageFolder.cs
+++ b/src/Files.App/Filesystem/StorageItems/ZipStorageFolder.cs
@@ -100,7 +100,7 @@ namespace Files.App.Filesystem.StorageItems
 				var assoc = await NativeWinApiHelper.GetFileAssociationAsync(filePath);
 				if (assoc is not null)
 				{
-					return assoc == Package.Current.Id.FamilyName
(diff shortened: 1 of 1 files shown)


yaira2 (approved):
LGTM

PR #10561, opened 2022-11-24
Check zip assoc against FamilyName

gave92:
**Resolved / Related Issues**
- Fixes an issue with #10554

On my PC `assoc` contains exactly `FilesDev_ykqwq8d6ps0ag` (FamilyName). So the PR for me actually causes the issue it's supposed to solve.
Not sure where the difference is (OS version?) but leaving both checks (FamilyName + Files.exe path) should be good for everyone.

**Validation**
- [x] Built and ran the app


yaira2 (approved):
LGTM

PR #10572, opened 2022-11-26
Fix: Fixed issue where app would freeze when deleting a folder that had an open file

### LINKED ISSUE #10530: Bug: Files freezes when deleting a folder that has an open file
### Description

Sometime Files will crash but the windows will just be frozen and have to be killed from task manager. I've only had this when deleting folders with no specific area.

### Steps To Reproduce

1. Open file in folder
2. Delete the folder
3. Files will freeze but not closed

### Steps to fix

1. Display error message with name of the open item instead of freezing

### Files Version

2.3.61.0

### Windows Version

10.0.22621.875

### Relevant Assets/Logs

[debug.log](https://github.com/files-community/Files/files/10050903/debug.log)

gave92:
**Resolved / Related Issues**
- Closes #10530

When it could not find which process was locking a file, code was calling ToList() on a null object.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Filesystem/FilesystemOperations/ShellFilesystemOperations.cs
+++ b/src/Files.App/Filesystem/FilesystemOperations/ShellFilesystemOperations.cs
@@ -723,7 +723,7 @@ namespace Files.App.Filesystem
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


yaira2 (approved):
LGTM

PR #10574, opened 2022-11-26
Fix: Fixed crash that would occur when opening multiple files at once

### LINKED ISSUE #10423: Bug: App crash while opening multiple files at once
### Description

The app would go unresponsive and crash. It is when you selected multiple images and open them

Also, if you try to open the selected images using **"Open With"**, the modal will pop up a lot of times. It is supposed to treat the selected items as one so that the modal will only show once. Even after selecting "Photos" as an app to open the files, it would only open the last item in the selection, not the whole highlighted items

### Steps To Reproduce

**A.**
1. Select multiple images
2. If your default photo viewer is Photos, press enter to open and it will crash.

**B.**
1. Select multiple images
2. Right-click and select **Open With**.
3. A number of modal windows will pop up and disappear for a moment
4. Select "Photos"
5. Only a single image is opened in Photos app.

### Steps to fix

- Fix the crash when opening multiple images using the selection
- **Open With** modal for multiple items that have similar file types should open once, not multiple times popping up.

### Files Version

2.3.58.0

### Windows Version

10.0.22623.875

### Relevant Assets/Logs

There's no error showing in the log.

gave92:
**Resolved / Related Issues**
- Closes #10423

Shell COM objects are not happy when passed between threads.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Shell/ContextMenu.cs
+++ b/src/Files.App/Shell/ContextMenu.cs
@@ -101,86 +101,70 @@ namespace Files.App.Shell
 
 		public async static Task<ContextMenu> GetContextMenuForFiles(string[] filePathList, Shell32.CMF flags, Func<string, bool> itemFilter = null)
 		{
-			List<ShellItem> shellItems = new List<ShellItem>();
-			try
-			{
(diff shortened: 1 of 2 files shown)


yaira2 (approved):
LGTM

PR #10250, opened 2022-10-19
Code: Implement inter-process synchronization for LiteDB

hez2010:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #10248
- Fixes https://github.com/files-community/Files/pull/10008#pullrequestreview-1144304212

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [ ] Tested the changes for accessibility

**Screenshots (optional)**
Add screenshots here.

PR #10008, opened 2022-09-17
Convert FileOperationsHandler + FileTagsHandler to helpers

itsWindows11:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Related #4180.
- Related #9858

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [ ] Tested the changes for accessibility

**Screenshots (optional)**
Add screenshots here.

PR #10577, opened 2022-11-27
Fix: Fixed issue where tags loaded slowly

gave92:
**Resolved / Related Issues**
- Since #10250 file tags are loaded noticeably slower. This PR restores the previous code. LiteDB dependency is downgraded to v4 as v5 does not handle concurrent access from multiple processes for packaged apps.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Filesystem/FileTagsHelper.cs
+++ b/src/Files.App/Filesystem/FileTagsHelper.cs
@@ -1,13 +1,11 @@
 using Common;
 using Files.App.Filesystem.StorageItems;
+using Files.App.Helpers;
 using Files.App.Shell;
 using Files.Shared.Extensions;
 using System;
-using System.IO;
 using System.Linq;
-using System.Runtime.InteropServices.ComTypes;
 using System.Threading.Tasks;
-using Vanara.PInvoke;
 using Windows.Storage;
 using Windows.Storage.FileProperties;
 using IO = System.IO;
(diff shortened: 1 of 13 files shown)


hez2010:
Even with LiteDB v4, it still does not work and causes crashing on multiple files instance startup. 
How about replacing LiteDB with SQLite?

gave92:
I can't reproduce the issue with v4. Lite sb should not lock the file with shared connection mode. SqLite is even worse for concurrency iirc.

hez2010:
Try open two files instances and then do file operation (such as Copy, Move, Delete...) on one.

hez2010:
The problem was here before we bump the LiteDB: https://github.com/files-community/Files/pull/10008#pullrequestreview-1144304212

gave92:
Have you verified with this PR? Code is a bit different from #10008.
> Try opening two files instances and then do file operation (such as Copy, Move, Delete...) on one.

Appears ok for me

hez2010:
I think we need to test the LiteDB manually (maybe somewhere other than Files) to make sure a dbfile can be opened from different processes without any issue.

gave92:
The problem in #10008 was this one:
image

gave92:
> I remembered that after merging https://github.com/files-community/Files/pull/10008 we were not able to do the above things until https://github.com/files-community/Files/pull/10250 got merged. Maybe something has changed later.

I'll take this as "I've tested your PR and it does not crash so it's worth more testing" :)

> I think we need to test the LiteDB manually (maybe somewhere other than Files) to make sure a dbfile can be opened and operated from different processes simultaneously without any issue.

Good idea 👍

gave92:
Here is a console program that runs a bunch of read/write to the DB. Launching the exe a few times concurrently appears OK. Let me know if you want to run more tests.
[LiteDBTest.zip](https://github.com/files-community/Files/files/10098361/LiteDBTest.zip)

hez2010:
> Here is a console program that runs a bunch of read/write to the DB. Launching the exe a few times concurrently appears OK. Let me know if you want to run more tests. [LiteDBTest.zip](https://github.com/files-community/Files/files/10098361/LiteDBTest.zip)

Well seems that the concurrency issue no longer persists. It's good to me.

gave92:
Wow I've just tried the test app with LiteDB v5 and read/write perfo is _awful_ when using Shared mode (1 second per each write). That is probably the cause of slow loading of tags. Yuk LiteDB v5!

hez2010 (approved):
LGTM.

Issue #10342, opened 2022-10-30
Feature: Custom themes v2


yaichenbaum:
## Rewriting custom themes

After discussing with the other contributors, it's been decided to rewrite custom themes in a more optimized way. What's clear is that the actual format doesn't matter to most users as they won't be editing the actual files. The deciding factor was that json is easier to work with code wise and is also easier to import/export with other settings.

## Implementation details

Theme profiles will be stored in a `themes.json` file, this is more efficient than storing them directly in the `settings.json` file. The `settings.json` file will only contain the selected theme name.

```json
{
  "Themes": [
    {
      "name": "Green Grass",
      "RootBackgroundBrush": "AA3B5A1F",
      "SolidBackgroundFillColorBase": "223412",
      "SolidBackgroundFillColorSecondary": "3B5A1F",
      "SolidBackgroundFillColorTertiary": "54802C",
      "AccentColor": "599124"
    },
    {
      "name": "Luxury Red",
      "RootBackgroundBrush": "AA3B5A1F",
      "SolidBackgroundFillColorBase": "223412",
      "SolidBackgroundFillColorSecondary": "3B5A1F",
      "SolidBackgroundFillColorTertiary": "54802C",
      "AccentColor": "599124"
    }
  ]
}
```

## UX

In addition to displaying themes in the setting dialog, we will have to add an "Edit themes" button. This will open the json file in the default editor and allow users to modify/create themes.

### Concerned code

NA

### Gains

NA

### Requirements

NA

mdtauk:
How will you handle things like Mica and Acrylic?

Does the AccentColor override the system provided colour?  And if so, how do you tell it to use the system colour?
`` "AccentColor": "SystemAccentFillSecondary"; ``
Something like this maybe? If you can parse the value when its not ARGB.

Have you considered adding support for background images maybe?  Either for the IconView or for the Window Background...

The recent support for JSON Themes with Terminal is a good comparison.

yaichenbaum:
> How will you handle things like Mica and Acrylic?

The app background automatically has mica (should fall back to acrylic on Windows 10). We might need to add an opacity value for each brush.

> Does the AccentColor override the system provided colour? And if so, how do you tell it to use the system colour?

The system will fall back to the default system accent color if a proper value isn't provided.

> Have you considered adding support for background images maybe? Either for the IconView or for the Window Background...

Providing a path for `RootBackgroundBrush` instead of a color should work.

yaichenbaum:
[...]

PR #10587, opened 2022-11-27
Fix: Theme is not applied on launch

### LINKED ISSUE #10586: Bug: Theme is not applied on launch
### Description

When using themes in Files, the theme is not applied on some app launches. Sometimes it would lose the theme after you close the app and open it again.

I had to hit Alt+I to open Settings and it automatically reloads the theme I set.

### Steps To Reproduce

1. Set a custom theme like Bronze or Green Grass or any theme that is noticeably different from the Default theme.
image

2. Close the app
3. Open the app again and it's back to the default theme. (If the theme is retained, close and open again a few more times)
image
Sometimes it would retain its theme but most of the time it doesn't.

My workaround for this is to hit Alt+I and the theme will refresh.

### Steps to fix

- Load the theme properly when launching the app

### Files Version

2.3.63

### Windows Version

10.0.22623.885

### Relevant Assets/Logs

None

gave92:
**Resolved / Related Issues**
- Closes #10586

Reload resources when applying theme, like we do everywhere else.
Simple fix, can be put in while #10342 is being worked on.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Helpers/ExternalResourcesHelper.cs
+++ b/src/Files.App/Helpers/ExternalResourcesHelper.cs
@@ -97,6 +97,7 @@ namespace Files.App.Helpers
 					{
 						Themes.Add(theme);
 					}
+					ForceReloadResourceFile();
 					return true;
 				}
 				return false;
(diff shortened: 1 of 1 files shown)


QuaintMako (approved):
LGTM for a temporary fix.

src/Files.App/Helpers/ExternalResourcesHelper.cs
```diff
@@ -108,6 +109,22 @@ public async Task<bool> TryLoadThemeAsync(AppTheme theme)
 			}
 		}
 
+		/// <summary>
```
yaira2, line 112: Isn't there already a method for this? Can we reuse it?

gave92: With this there are 3 similar methods, but they're all a bit different and I didn't want to revalidate them 😅

PR #10596, opened 2022-11-28
Feature: Add support for choosing the default sorting and grouping options

### LINKED ISSUE #7082: Feature: Add option to set default sorting options
### Requirements
- Add combo box drop down to folder settings page similar to the default layout options
   - Settings can still be overridden on individual directories 

### Files Version

Version: 2.0.34.0

### Windows Version

 OS Version: 10.0.19044.1348

### Additional comment

_No response_

manfromarce:
**Resolved / Related Issues**
- Closes #7082

**Validation**
- [x] Built and ran the app


yaira2:
Locking this PR as the feature request requirements haven't been finalized. Let's continue dicussing in the issue.

yaira2:
Requirements have been sorted out

manfromarce:
Ok, if I understand correctly I need to remove the separate "override sort settings" option and use the existing Override layout option, and move the new settings to the Folders tab. Should I also move "Sort directories alongside files" which was already in Preferences?

yaira2:
> use the existing Override layout option, and move the new settings to the Folders tab.

Correct

> Should I also move "Sort directories alongside files" which was already in Preferences?

Let's leave as is for now

src/Files.App/ViewModels/FolderSettingsViewModel.cs
```diff
@@ -287,47 +287,59 @@ public ColumnsViewModel ColumnsViewModel
 		private static LayoutPreferences GetLayoutPreferencesForPath(string folderPath)
 		{
 			IUserSettingsService userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
-			if (userSettingsService.FoldersSettingsService.EnableOverridingFolderPreferences)
+
+			var layout = LayoutPreferences.DefaultLayoutPreferences;
+
+			if (userSettingsService.FoldersSettingsService.EnableOverridingFolderPreferences || userSettingsService.LayoutSettingsService.EnableOverridingSortingPreferences)
```
QuaintMako, line 293: You're using these two variables two times. One here and one in l.302 and l.323.
You'd be better off saving it inside variables to not access them two times.

QuaintMako, line 293: You could also reverse the condition of the `if` to avoid a nesting level with an earlier `return layout` after it.

src/Files.App/ViewModels/FolderSettingsViewModel.cs
```diff
@@ -348,6 +360,32 @@ public static void SetLayoutPreferencesForPath(string folderPath, LayoutPreferen
...
+
+			if (!userSettingsService.LayoutSettingsService.EnableOverridingSortingPreferences)
+			{
+				// Do not save OriginalPath as global sort option (only works in recycle bin)
+				if (prefs.DirectorySortOption != SortOption.OriginalFolder &&
+					prefs.DirectorySortOption != SortOption.DateDeleted &&
+					prefs.DirectorySortOption != SortOption.SyncStatus)
+				{
+					userSettingsService.LayoutSettingsService.DefaultDirectorySortOption = prefs.DirectorySortOption;
+				}
```
QuaintMako, line 372: One liner `if`s should not have curly brackets.

src/Files.App/ViewModels/FolderSettingsViewModel.cs
```diff
@@ -348,6 +360,32 @@ public static void SetLayoutPreferencesForPath(string folderPath, LayoutPreferen
...
+				if (prefs.DirectoryGroupOption != GroupOption.OriginalFolder &&
+					prefs.DirectoryGroupOption != GroupOption.DateDeleted &&
+					prefs.DirectoryGroupOption != GroupOption.FolderPath &&
+					prefs.DirectoryGroupOption != GroupOption.SyncStatus)
+				{
+					userSettingsService.LayoutSettingsService.DefaultDirectoryGroupOption = prefs.DirectoryGroupOption;
+				}
```
QuaintMako, line 379: One liner `if`s should not have curly brackets.

src/Files.App/ViewModels/SettingsViewModels/PreferencesViewModel.cs
```diff
@@ -112,6 +112,9 @@ public PreferencesViewModel()
 
 			AddDateTimeOptions();
 			SelectedDateTimeFormatIndex = (int)Enum.Parse(typeof(DateTimeFormats), DateTimeFormat.ToString());
+			SelectedDefaultSortingIndex = DefaultSortOption == SortOption.FileTag ? 5 : (int)DefaultSortOption;
```
QuaintMako, line 115: Where does the `5` comes from? Maybe you could put it inside a well named variable to avoid the magic constant effect?

src/Files.App/ViewModels/SettingsViewModels/PreferencesViewModel.cs
```diff
@@ -112,6 +112,9 @@ public PreferencesViewModel()
...
+			SelectedDefaultSortDirectionIndex = (int)DefaultSortDirection;
+			SelectedDefaultGroupingIndex = DefaultGroupOption == GroupOption.FileTag ? 6 : (int)DefaultGroupOption;
```
QuaintMako, line 117: Same as the previous point about the `6`.

src/Files.App/ViewModels/SettingsViewModels/PreferencesViewModel.cs
```diff
@@ -475,6 +491,87 @@ public bool ListAndSortDirectoriesAlongsideFiles
...
+		private int selectedDefaultSortingIndex;
+		public int SelectedDefaultSortingIndex
+		{
+			get => selectedDefaultSortingIndex;
+			set
+			{
+				if (SetProperty(ref selectedDefaultSortingIndex, value))
+				{
+					OnPropertyChanged(nameof(SelectedDefaultSortingIndex));
+					DefaultSortOption = value < 5 ? (SortOption)value : SortOption.FileTag;
```
QuaintMako, line 503: Same as above about the magic constant. If the `5` here has the same use as line 115, putting it inside a `private readonly int` field could solve the issue quite nicely.

src/Files.App/ViewModels/SettingsViewModels/PreferencesViewModel.cs
```diff
@@ -475,6 +491,87 @@ public bool ListAndSortDirectoriesAlongsideFiles
...
+		private int selectedDefaultGroupingIndex;
+		public int SelectedDefaultGroupingIndex
+		{
+			get => selectedDefaultGroupingIndex;
+			set
+			{
+				if (SetProperty(ref selectedDefaultGroupingIndex, value))
+				{
+					OnPropertyChanged(nameof(SelectedDefaultGroupingIndex));
+					DefaultGroupOption = value < 6 ? (GroupOption)value : GroupOption.FileTag;
```
QuaintMako, line 531: `6` has the name problem and possible solution as `5` on last point.

QuaintMako (commented):
LGTM outside of those small propositions to make the code cleaner.

src/Files.App/ViewModels/SettingsViewModels/PreferencesViewModel.cs
manfromarce: This is a workaround to hide "sync status" and other options which only work in cloud folders or recycle bin. I tried swapping sync status and tag in SortOption and GroupOption, but it should be done in toolbar and context menu too, otherwise the wrong option is selected there.

manfromarce:
The latest commit can be reverted if we prefer to leave the order as is.

src/Files.App/ServicesImplementation/Settings/FoldersSettingsService.cs
```diff
@@ -207,6 +207,25 @@ public bool CalculateFolderSizes
 			get => Get(false);
 			set => Set(value);
 		}
+
+		public SortDirection DefaultDirectorySortDirection
```
yaira2, line 211: We don't plan on this option for now

src/Files.App/ViewModels/SettingsViewModels/PreferencesViewModel.cs
QuaintMako: If it's a work around, I again highly suggest using a named variable to explain its use. Or at least a comment next to it.

src/Files.App/ServicesImplementation/Settings/FoldersSettingsService.cs
manfromarce: Ok, I will remove it

QuaintMako (commented):
Aside of the comments still up, LGTM in general code-wise.

src/Files.App/Helpers/ContextFlyoutItemHelper.cs
```diff
@@ -292,6 +284,14 @@ public static List<ContextMenuFlyoutItemViewModel> GetBaseLayoutMenuItems(Curren
...
 						},
 						new ContextMenuFlyoutItemViewModel()
+						{
+							Text = "SyncStatus".GetLocalizedResource(),
+							IsChecked = itemViewModel.IsSortedBySyncStatus,
+							Command = new RelayCommand(() => itemViewModel.IsSortedBySyncStatus = true),
+							ShowItem = currentInstanceViewModel.IsPageTypeCloudDrive,
+							ItemType = ItemType.Toggle
+						},
+						new ContextMenuFlyoutItemViewModel()
```
yaira2, line 294: Why was this changed?

src/Files.App/ServicesImplementation/Settings/FoldersSettingsService.cs
yaira2: It should be kept in LayoutSettingsService but not here as we don't want to expose the setting.

yaira2, line 211: We shouldn't be exposing this to users right now

src/Files.Shared/Enums/SortOption.cs
```diff
@@ -7,9 +7,9 @@ public enum SortOption : byte
 		DateCreated,
 		Size,
 		FileType,
-		SyncStatus,
 		FileTag,
-		OriginalFolder,
-		DateDeleted
+		SyncStatus, // Cloud drive
```
yaira2, line 11: Why were these reordered?

manfromarce: This was to separate the options available for all folders from those specific to recycle bin or cloud drives (which need to be hidden from global settings), but I can revert this changes as well as those about default sort direction

src/Files.App/Helpers/ContextFlyoutItemHelper.cs
manfromarce: This was necessary after reordering the enum (please see below)

src/Files.Shared/Enums/SortOption.cs
yaira2: That would be preferred 👍

manfromarce:
I should be done with the requested changes.

src/Files.App/ServicesImplementation/Settings/FoldersSettingsService.cs
```diff
@@ -207,6 +207,19 @@ public bool CalculateFolderSizes
...
 			set => Set(value);
 		}
+
+		public SortOption DefaultDirectorySortOption
+		{
+			get => (SortOption)Get((long)SortOption.Name);
+			set => Set((long)value);
+		}
+
+		public GroupOption DefaultDirectoryGroupOption
```
yaira2, line 217:
```suggestion
		public GroupOption DefaultGroupOption
```

src/Files.App/ServicesImplementation/Settings/FoldersSettingsService.cs
```diff
@@ -207,6 +207,19 @@ public bool CalculateFolderSizes
 			get => Get(false);
 			set => Set(value);
 		}
+
+		public SortOption DefaultDirectorySortOption
```
yaira2, line 211:
```suggestion
		public SortOption DefaultSortOption
```

src/Files.App/ViewModels/SettingsViewModels/FoldersViewModel.cs
```diff
@@ -239,6 +244,60 @@ public bool CalculateFolderSizes
...
+		private int selectedDefaultGroupingIndex;
+		public int SelectedDefaultGroupingIndex
+		{
+			get => selectedDefaultGroupingIndex;
+			set
+			{
+				if (SetProperty(ref selectedDefaultGroupingIndex, value))
+				{
+					OnPropertyChanged(nameof(SelectedDefaultGroupingIndex));
+					DefaultGroupOption = value == FileTagGroupingIndex ? GroupOption.FileTag : (GroupOption)value;
```
yaira2, line 270:
```suggestion
					FoldersSettingsService.DefaultDirectorySortOption = value == FileTagGroupingIndex ? GroupOption.FileTag : (GroupOption)value;
```
```suggestion
					DefaultGroupOption = value == FileTagGroupingIndex ? GroupOption.FileTag : (GroupOption)value;
```
If we do this, we can remove the `DefaultSortOption` property from line 275.

manfromarce: You probably meant FoldersSettingsService.DefaultGroupOption

yaira2: Yeah

yaira2: Can you do the same for Sorting?

src/Files.App/ViewModels/SettingsViewModels/FoldersViewModel.cs
```diff
@@ -10,6 +10,9 @@ public class FoldersViewModel : ObservableObject
 	{
 		private IUserSettingsService UserSettingsService { get; } = Ioc.Default.GetRequiredService<IUserSettingsService>();
 
+		private readonly int FileTagSortingIndex = 5;
+		private readonly int FileTagGroupingIndex = 6;
```
yaira2, line 14: What are these values for?

src/Files.App/ViewModels/SettingsViewModels/FoldersViewModel.cs
manfromarce: Yes

src/Files.App/ViewModels/SettingsViewModels/FoldersViewModel.cs
manfromarce: They are used to select the correct option in the comboboxes. Sync status, original path and delete date are hidden because they aren't available for regular folders, so if Tags is set as default the casting would return an invalid index. At first I swapped FileTag and SyncStatus but that enum is referenced from many places.

yaira2: Can you add a comment to the code?

manfromarce: Yes, I'll do it

yaira2 (approved):
LGTM

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!-- 
+    Microsoft ResX Schema 
+
```
QuaintMako, line 5:
```suggestion
  <!--
    Microsoft ResX Schema

```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -1,17 +1,17 @@
...
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
     associated with the data types.
-
+    
     Example:
-
+
```
QuaintMako, line 14:
```suggestion

```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -26,36 +26,36 @@
         <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
         <comment>This is a comment</comment>
     </data>
-
-    There are any number of "resheader" rows that contain simple
+
```
QuaintMako, line 29:
```suggestion

```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -26,36 +26,36 @@
         <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
         <comment>This is a comment</comment>
     </data>
-
-    There are any number of "resheader" rows that contain simple
+                
+    There are any number of "resheader" rows that contain simple
```
QuaintMako, line 30:
```suggestion
    There are any number of "resheader" rows that contain simple
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -26,36 +26,36 @@
...
-
-    Each data row contains a name, and value. The row also contains a
-    type or mimetype. Type corresponds to a .NET class that support
-    text/value conversion through the TypeConverter architecture.
-    Classes that don't support this are serialized and stored with the
+    
+    Each data row contains a name, and value. The row also contains a 
+    type or mimetype. Type corresponds to a .NET class that support 
+    text/value conversion through the TypeConverter architecture. 
+    Classes that don't support this are serialized and stored with the
```
QuaintMako, line 36:
```suggestion

    Each data row contains a name, and value. The row also contains a
    type or mimetype. Type corresponds to a .NET class that support
    text/value conversion through the TypeConverter architecture.
    Classes that don't support this are serialized and stored with the
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -26,36 +26,36 @@
...
+    type or mimetype. Type corresponds to a .NET class that support 
+    text/value conversion through the TypeConverter architecture. 
+    Classes that don't support this are serialized and stored with the 
     mimetype set.
-
-    The mimetype is used for serialized objects, and tells the
-    ResXResourceReader how to depersist the object. This is currently not
+    
+    The mimetype is used for serialized objects, and tells the 
+    ResXResourceReader how to depersist the object. This is currently not
```
QuaintMako, line 40:
```suggestion

    The mimetype is used for serialized objects, and tells the
    ResXResourceReader how to depersist the object. This is currently not
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -26,36 +26,36 @@
...
     extensible. For a given mimetype the value must be set accordingly:
-
-    Note - application/x-microsoft.net.object.binary.base64 is the format
-    that the ResXResourceWriter will generate, however the reader can
+    
+    Note - application/x-microsoft.net.object.binary.base64 is the format 
+    that the ResXResourceWriter will generate, however the reader can 
     read any of the formats listed below.
-
+
```
QuaintMako, line 46:
```suggestion

```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -26,36 +26,36 @@
...
     read any of the formats listed below.
-
+    
     mimetype: application/x-microsoft.net.object.binary.base64
-    value   : The object must be serialized with
+    value   : The object must be serialized with 
             : System.Runtime.Serialization.Formatters.Binary.BinaryFormatter
             : and then encoded with base64 encoding.
-
+
```
QuaintMako, line 51:
```suggestion

```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -26,36 +26,36 @@
...
     mimetype: application/x-microsoft.net.object.binary.base64
-    value   : The object must be serialized with
+    value   : The object must be serialized with 
             : System.Runtime.Serialization.Formatters.Binary.BinaryFormatter
             : and then encoded with base64 encoding.
-
+    
     mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with
+    value   : The object must be serialized with
```
QuaintMako, line 53:
```suggestion
    value   : The object must be serialized with
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -26,36 +26,36 @@
...
+    
     mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with
+    value   : The object must be serialized with 
             : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
             : and then encoded with base64 encoding.
 
     mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array
+    value   : The object must be serialized into a byte array
```
QuaintMako, line 58:
```suggestion
    value   : The object must be serialized into a byte array
```

QuaintMako (approved):
LGTM, aside of the small formatting I suggested above to remove unneeded spaces.

QuaintMako (approved):

PR #10597, opened 2022-11-28
Fix: Prevent Visual Studio from stopping when deleting an item

cinqmilleans:
**Resolved / Related Issues**
In debug, Visual sSudio stops at a Win32Exception necessary to know if the recycle bin can be used. This pr disables Visual Studio stops for this exception.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Helpers/FileOperationsHelpers.cs
+++ b/src/Files.App/Helpers/FileOperationsHelpers.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


QuaintMako (approved):
LGTM.

yaira2:
1. How can I reproduce this issue?
2. Can you tell Visual Studio to ignore it locally? I'm not sure we should add this to the code.

cinqmilleans:
Open Files in Debug mode, select a file, press Delete. Visual Studio stops at a Win32Exception. The code can only raise an exception regardless of the result. It is probably not possible to do otherwise.

It's fine for Visual Studio to break for a Win32Exception, but not this one. Developers shouldn't change the options, just definitely point out not to stop at this exception.

yaira2:
I can't reproduce this
Fyi @d2dyno1

QuaintMako:
I can reproduce it, do you have the right exceptions activated in Visual Studio?

yaira2:
How can I check?

QuaintMako:
> How can I check?

Visual Studio, Press Ctrl+Alt+E and check if the Win32Exceptions are checked.

In case: https://learn.microsoft.com/en-us/visualstudio/debugger/managing-exceptions-with-the-debugger?view=vs-2022#tell-the-debugger-to-break-when-an-exception-is-thrown

yaira2:
It's checked but it's possible I ignored this specific error.

yaira2 (approved):
LGTM

PR #10618, opened 2022-12-03
Code Quality: Refactor SidebarViewModel

cinqmilleans:
**Resolved / Related Issues**
As part of the viewmodel refactoring, here is the refactoring of the class SidebarViewModel. Added an interface, removed unused code, improved member visibility, fixed warnings, improved readability and improved member order.

**Validation**
- [x] Built and ran the app


src/Files.App/ViewModels/SidebarViewModel.cs
```diff
@@ -6,230 +6,133 @@
...
-
 		public NavigationViewDisplayMode SidebarDisplayMode
 		{
 			get => sidebarDisplayMode;
 			set
 			{
 				if (SetProperty(ref sidebarDisplayMode, value))
-				{
 					OnPropertyChanged(nameof(IsSidebarCompactSize));
-					UpdateTabControlMargin();
```
QuaintMako, line 59: Why is this method removed here?

src/Files.App/ViewModels/SidebarViewModel.cs
```diff
@@ -6,230 +6,133 @@
...
-			Manager_DataChanged(SectionType.WSL, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
-			Manager_DataChanged(SectionType.FileTag, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
+			var resetEventArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset);
+			Manager_DataChanged(SectionType.Favorites, resetEventArgs);
+			Manager_DataChanged(SectionType.Library, resetEventArgs);
+			Manager_DataChanged(SectionType.Drives, resetEventArgs);
+			Manager_DataChanged(SectionType.CloudDrives, resetEventArgs);
+			Manager_DataChanged(SectionType.Network, resetEventArgs);
+			Manager_DataChanged(SectionType.WSL, resetEventArgs);
+			Manager_DataChanged(SectionType.FileTag, resetEventArgs);
```
QuaintMako, line 136: Can all the `new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset` be safely represented by one object where they used to be different objects before?

src/Files.App/ViewModels/SidebarViewModel.cs
```diff
@@ -240,91 +143,110 @@ public SidebarViewModel()
...
+			=> item is not null && !item.IsEmpty && item.IsDefaultLocation;
 
 		private async Task AddElementToSection(INavigationControlItem elem, LocationItem section, int index = -1)
 		{
 			if (elem is LibraryLocationItem lib)
 			{
-				if (IsLibraryOnSidebar(lib) &&
-					await lib.CheckDefaultSaveFolderAccess() &&
-					!section.ChildItems.Any(x => x.Path == lib.Path))
+				if (IsLibraryOnSidebar(lib) && await lib.CheckDefaultSaveFolderAccess() && !section.ChildItems.Any(x => x.Path == lib.Path))
```
QuaintMako, line 249: A nested `if`` could be eliminated here.

QuaintMako (commented):
Aside from small questions and details, LGTM.

src/Files.App/ViewModels/SidebarViewModel.cs
cinqmilleans: This method updates a property that is never read. The property has been removed.

yaira2:
Just a side note but SidebarController doesn't fit with MVVM and should be removed at some point.

src/Files.App/ViewModels/SidebarViewModel.cs
cinqmilleans: Yes because it is read-only and the NewItems and OldItems properties are null.

src/Files.App/ViewModels/SidebarViewModel.cs
cinqmilleans: I do not agree. If elem is LibraryLocationItem but does not satisfy the other conditions, it should not pass in the last "else if" below.

QuaintMako (dismissed):
LGTM.

src/Files.App/Controllers/SidebarPinnedController.cs
```diff
@@ -18,7 +18,7 @@ public class SidebarPinnedController : IJson
 	{
 		public SidebarPinnedModel Model { get; set; }
 
-		public EventHandler<NotifyCollectionChangedEventArgs> DataChanged;
+		public EventHandler<NotifyCollectionChangedEventArgs>? DataChanged;
```
yaira2, line 21: When would this be null?

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -363,11 +364,6 @@ private void ToggleSidebarCollapsedState(KeyboardAcceleratorInvokedEventArgs? e)
 			e!.Handled = true;
 		}
 
-		private void SidebarControl_Loaded(object sender, RoutedEventArgs e)
-		{
-			SidebarAdaptiveViewModel.UpdateTabControlMargin(); // Set the correct tab margin on startup
```
yaira2, line 368: Is this no longer used?

src/Files.App/ViewModels/SidebarViewModel.cs
```diff
@@ -345,264 +267,205 @@ private async Task AddElementToSection(INavigationControlItem elem, LocationItem
...
-				case nameof(UserSettingsService.AppearanceSettingsService.IsSidebarOpen):
-					if (UserSettingsService.AppearanceSettingsService.IsSidebarOpen != IsSidebarOpen)
-					{
-						OnPropertyChanged(nameof(IsSidebarOpen));
-					}
+				case nameof(IAppearanceSettingsService.IsSidebarOpen):
+					OnPropertyChanged(nameof(IsSidebarOpen));
 					break;
-				case nameof(UserSettingsService.AppearanceSettingsService.ShowFavoritesSection):
+				case nameof(IAppearanceSettingsService.ShowFavoritesSection):
```
yaira2, line 440: This isn't right
@d2dyno1

src/Files.App/Controllers/SidebarPinnedController.cs
cinqmilleans: It can't be null. This property is nullable because it is used as an event and unsubscribing causes a warning if the event is not nullable.
This property is not an event because it is triggered from another class. It's not the best way to do it but that's not the purpose of this pr.

src/Files.App/Views/MainPage.xaml.cs
cinqmilleans: UpdateTabControlMargin only updates a never-read value. Dead code is removed.

src/Files.App/ViewModels/SidebarViewModel.cs
cinqmilleans: Why is this not right?

d2dyno1: Looks good

src/Files.App/Controllers/SidebarPinnedController.cs
yaira2: We're trying to stay away from making properties/events nullable unless necessary. The preferred solution is to solve the underlying problem. In this case, we'd need to merge the controller with the view model.

yair100 (commented):
Aside from the requested changes on the sidebar controller, this is looking very good.

src/Files.App/Controllers/SidebarPinnedController.cs
cinqmilleans: This change is to avoid a warning in SidebarViewModel. The controller and the model will have to be refactored but it is a big job and it is not the object of this pr. The behavior must be analyzed because the code is not clean.

yaira2: It's best to leave as it was, simply removing the warning doesn't solve the issue.

d2dyno1: Nope. Nullable should always be applied to events unless it has a default, non-null value

cinqmilleans: I set this DataChanged back to non-nullable. The SidebarPinnedModel/Controller logic needs to be reviewed (I'll do it in January). The other DataChanged have been converted to an event, which makes more sense.

src/Files.App/ViewModels/SidebarViewModel.cs
```diff
@@ -240,91 +143,110 @@ public SidebarViewModel()
...
+				.Where(x => !string.IsNullOrWhiteSpace(x.Path))
+				.Concat(SideBarItems
+					.Where(x => (x as LocationItem)?.ChildItems is not null)
+					.SelectMany(x => ((LocationItem)x).ChildItems)
+					.Where(x => !string.IsNullOrWhiteSpace(x.Path))
+				)
+				.ToList();
+
+			if (string.IsNullOrEmpty(arg))
+				return;
```
ferrariofilippo, line 159: I think it would make more sense if these two lines were moved to the beginning of the method

ferrariofilippo (commented):
I left a comment on code logic, aside from that LGTM

yaira2:
@cinqmilleans sorry for the delay on this PR, we made some changes to the sidebar during this time, and it looks like there are some merge conflicts. Can you look and see if they can be resolved?

ferrariofilippo (approved):
LGTM

src/Files.App/ViewModels/ISidebarViewModel.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.UI.Xaml.Controls;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Windows.Input;
+
+namespace Files.App.ViewModels
+{
+	public interface ISidebarViewModel : INotifyPropertyChanged, IDisposable
```
yaira2, line 12: Can you add docs/comments to the interface?

src/Files.App/DataModels/SidebarPinnedModel.cs
```diff
@@ -265,7 +265,7 @@ private void AddLocationItemToSidebar(LocationItem locationItem)
 				insertIndex = lastItem is not null ? favoriteList.IndexOf(lastItem) + 1 : 0;
 				favoriteList.Insert(insertIndex, locationItem);
 			}
-			controller?.DataChanged?.Invoke(SectionType.Favorites, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Add, locationItem, insertIndex));
+			controller?.DataChanged.Invoke(SectionType.Favorites, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Add, locationItem, insertIndex));
```
d2dyno1, line 268: Why remove null check?

src/Files.App/ViewModels/ISidebarViewModel.cs
d2dyno1: Why is sidebar viewmodel an interface?

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -247,9 +247,10 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
 			SidebarControl.SidebarItemNewPaneInvoked += SidebarControl_SidebarItemNewPaneInvoked;
 		}
 
-		private async void SidebarControl_SidebarItemDropped(object sender, SidebarItemDroppedEventArgs e)
+		private async void SidebarControl_SidebarItemDropped(object _, SidebarItemDroppedEventArgs e)
```
d2dyno1, line 250: Why remove sender parameter?

yaira2:
@cinqmilleans

cinqmilleans:
@yair100 This pr is too old. You have to redo everything because the code has changed too much. I close it.

PR #10661, opened 2022-12-09
Codebase quality: Refactor entire Files.App

0x5bfa:
**Validation**
- [x] Built and ran the app

**Screenshots (optional)**

_None_

**Refactoring rules**

- The previous/next line before/after a control syntax such as `if`, `while`, and `for` is most often a line break
```cs
var something = new Class1();

if (a.Length == 0)
{
}

return null;
```
- All lamba expressions for properties and mothods except Linq expressions are broke to the next line.
```cs
public bool IsNull()
    => Content.Type == null;
```
- There're line breaks between properties, mothods, fields. but there's no line break between the properties and fields in the binding relationship.
```cs
public EventHandler<NotifyCollectionChangedEventArgs> DataChanged;

private readonly List<WslDistroItem> distros = new();

---

private readonly List<DriveItem> drives = new();
public IReadOnlyList<DriveItem> Drives
{
    get => drives;
    set => SetProperty(ref drives, value);
}
```
- The last line of all files ends with only a line break.
```cs
        }
    }
}
[CRLF]
```
- The before line of the `return` statement is line break
```cs
var something = new Class1();

return null;
```


0x5bfa:
@yair100 Before taking a look at ListViewExtensions of CommunityToolkit, I would like you to review this.

onein528:
Files gonna be great code quality!

PR #10671, opened 2022-12-10
Exit renaming with F2

cinqmilleans:
**Resolved / Related Issues**
On the keyboard, it is possible to enter renaming mode with the F2 key, but you must use Escape to exit without validating (or Enter to exit and validate).
This pr allows you to exit without validating with F2, which is easier in the event of an error.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
@@ -224,7 +224,7 @@ namespace Files.App.Views.LayoutModes
 
 		private void RenameTextBox_KeyDown(object sender, KeyRoutedEventArgs e)
 		{
-			if (e.Key == VirtualKey.Escape)
+			if (e.Key is VirtualKey.Escape or VirtualKey.F2)
 			{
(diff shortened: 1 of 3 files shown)


yaira2:
> This pr allows you to exit without validating with F2, which is easier in the event of an error.

I'm not against the shortcut but what do you mean easier in the event of an error??

QuaintMako:
> I'm not against the shortcut but what do you mean easier in the event of an error??

Probably if hitting `F2` by mistake. The gain of time from F2 to Esc would be minimal, but I'm more curious about an accessibility point of view. I'm not versed in accessiblity, but is mapping the on/off of a specific behaviour to the same shortcut a good idea?

On a sidenote, the new shortcut works on the application as intended and code LGTM.

yaira2:
> Probably if hitting F2 by mistake. The gain of time from F2 to Esc would be minimal though.

If this is the scenario, we should just ignore the shortcut when the user is renaming a file.

> I'm more curious about an accessibility point of view. I'm not versed in accessiblity, but is mapping the on/off of a specific behaviour to the same shortcut a good idea?

This would be a concern

PR #10686, opened 2022-12-13
Fix: Fixed issue where the prompt to review would crash the app

yaira2:
**Resolved / Related Issues**
- Fixed issue where the prompt to review would crash the app [636092730u](https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/636092730u/overview)

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Views/MainPage.xaml.cs
+++ b/src/Files.App/Views/MainPage.xaml.cs
@@ -73,12 +73,6 @@ namespace Files.App.Views
 			ToggleCompactOverlayCommand = new RelayCommand(ToggleCompactOverlay);
 			SetCompactOverlayCommand = new RelayCommand<bool>(SetCompactOverlay);
 
-			if (SystemInformation.Instance.TotalLaunchCount >= 15 & Package.Current.Id.Name == "49306atecsolution.FilesUWP" && !UserSettingsService.ApplicationSettingsService.WasPromptedToReview)
-			{
-				PromptForReview();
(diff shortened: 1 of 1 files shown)


src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -328,6 +324,17 @@ private void Page_Loaded(object sender, RoutedEventArgs e)
 			FindName(nameof(InnerNavigationToolbar));
 			FindName(nameof(horizontalMultitaskingControl));
 			FindName(nameof(NavToolbar));
+
+			// Prompt user to review app in the Store
+			if
+			(
+				SystemInformation.Instance.TotalLaunchCount >= 15 &
+				Package.Current.Id.Name == "49306atecsolution.FilesUWP" &&
```
QuaintMako, line 332: Extracting the string value inside a constant?

yaira2: Is there a constant?

flaviabanchieri: > Extracting the string value inside a constant?

I think there's no constant, or, case there's one, where is the constant?

yaira2: Looks like this is the only place where we use the package name so we can keep as is.

QuaintMako (approved):
LGTM.

PR #10728, opened 2022-12-15
Fix: Fixed crash that would occur when switching between Network and Recycle Bin

### LINKED ISSUE #10715: App may crash when clicking on Network or Recycle Bin in Sidebar
### Description

This seems to be an issue appearing randomly when using the Sidebar.

I have been able to provoke it only on the Recycle Bin and the Network Drives so far.

### Steps To Reproduce

1. Alternate clicking between items in the Sidebar
2. When clicking on the Recycle Bin or a Network Drive, the application may crash.

### Steps to fix

- Clicking on Network or Recycle Bin should not provoke a crash.

### Files Version

Version: 2.4.1.0 OS Version: 10.0.19045.2251

### Windows Version

W10 Home 22H2 19045.2251

### Relevant Assets/Logs

[debug.log](https://github.com/files-community/Files/files/10231572/debug.log)

QuaintMako:
**Resolved / Related Issues**
- Closes #10715

**What has been done**
- Wrapped the acquisition of the ShellItem inside a `try...catch...` so the exception raised when a path not valid would be given is no longer making the application crash.

**Validation**
- [x] Built and ran the app


src/Files.App/Filesystem/StorageItems/ShellStorageFile.cs
```diff
@@ -85,10 +85,17 @@ public static IAsyncOperation<BaseStorageFile> FromPathAsync(string path)
 			return Task.FromResult<BaseStorageFile>(null).AsAsyncOperation();
 		}
 
-		private static ShellFileItem GetFile(string path)
+		private static ShellFileItem? GetFile(string path)
```
yaira2, line 88: Why would this be null?

src/Files.App/Filesystem/StorageItems/ShellStorageFile.cs
```diff
@@ -85,10 +85,17 @@ public static IAsyncOperation<BaseStorageFile> FromPathAsync(string path)
...
-			using var shellItem = ShellFolderExtensions.GetShellItemFromPathOrPidl(path);
-			return ShellFolderExtensions.GetShellFileItem(shellItem);
+			try
+			{
+				using var shellItem = ShellFolderExtensions.GetShellItemFromPathOrPidl(path);
+				return ShellFolderExtensions.GetShellFileItem(shellItem);
+			}
+			catch
+			{
+				return default;
```
yaira2, line 97: What circumstances will cause it to catch? Let's add a comment to the code.

src/Files.App/Filesystem/StorageItems/ShellStorageFile.cs
QuaintMako: The default value of ShellFileItem is null, which could be now raised in the `catch`.

yaira2 (approved):
LGTM

gave92:
If it can wait I'd like to take a look at this.

QuaintMako:
@gave92 solution is much cleaner. The issue I offer felt like a makeshift fix, thanks for having looking into it !

QuaintMako (commented):
LGTM.

gave92:
Pushed too early, still missing something. You confirm it solves the issue?

yaira2:
> as I found it's needed in some cases.

Do you know where it's still needed? Might be useful to add a comment.

gave92:
> Do you know where it's still needed? Might be useful to add a comment.

Done!

yaira2 (approved):
LGTM

PR #10772, opened 2022-12-19
Codebase quality: Use helper method for checking .lnk and .url extensions

gave92:
**Resolved / Related Issues**
- No issue, only code quality

Created helper method to check if a file path is a shortcut or url file instead of having checks for the extension in 20 places. Also makes sure to use OrdinalIgnoreCase instead of Ordinal.

**Validation**
- [x] Built and ran the app


yaira2 (approved):
LGTM

PR #10782, opened 2022-12-20
Code Quality: Refactored AdaptiveLayoutHelpers

cinqmilleans:
**Resolved / Related Issues**
Refactor AdaptiveLayoutHelpers, which applies the layout according to the folder. The behavior has not changed.

This refactoring makes it easier to modify rules, which are not always suitable. For example, a video folder is not displayed in grid.

**Validation**
- [x] Built and ran the app


src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
```diff
@@ -3,124 +3,128 @@
...
+using static Files.App.Constants.AdaptiveLayout;
+using IO = System.IO;
 
 namespace Files.App.Helpers
 {
 	public static class AdaptiveLayoutHelpers
 	{
+		private static readonly IFoldersSettingsService foldersSettingsService = Ioc.Default.GetRequiredService<IFoldersSettingsService>();
+
 		public static bool PredictLayoutMode(FolderSettingsViewModel folderSettings, string path, IList<ListedItem> filesAndFolders)
```
QuaintMako, line 20: I find the name of this method a bit missleading. Does it predict what layer mode should be used? Or does it evaluates if the layer mode should be predicted? 
Could use a rename while we're at refactoring it.

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
```diff
@@ -3,124 +3,128 @@
...
+			static bool IsFolder(ListedItem item)
+				=> item.PrimaryItemAttribute is StorageItemTypes.Folder;
+
+			static bool IsImage(ListedItem item)
+				=> !string.IsNullOrEmpty(item.FileExtension)
+				&& ImagePreviewViewModel.ContainsExtension(item.FileExtension.ToLowerInvariant());
+
+			static bool IsMedia(ListedItem item)
+				=> !string.IsNullOrEmpty(item.FileExtension)
+				&& MediaPreviewViewModel.ContainsExtension(item.FileExtension.ToLowerInvariant());
```
QuaintMako, line 120: This is out of this PR, but we should get all this `IsFolder`, `IsImage` and the likes inside a common helper and mutualize its use.

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
```diff
@@ -3,124 +3,128 @@
...
+		}
+
+		private enum Layouts
+		{
+			None,
+			Detail,
+			Grid,
 		}
```
QuaintMako, line 128: Is there no Layouts type enum somewhere else? If not, does it really belong here and not inside a specialized enum stored inside its own file?

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
QuaintMako: Actually such an helper exists : `FileExtensionHelpers`.
Would be a good opportunity to enrich it!

cinqmilleans: There is a difference in the behavior of these 2 methods. FileExtensionHelpers checks 4 extensions to display some commands (Rotate, ...). ImagePreviewViewModel checks 8 extensions to display the item as an image in the preview. It makes sense that .ico is displayed as an image in Preview but cannot be set as Wallpaper. For our case, it will be necessary to redefine the criteria (a file made up of videos is displayed in detail instead of grid). This pr was made to facilitate these behavior changes with a more readable and modifiable code. Maybe add a HasThumbnail property.

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
cinqmilleans: This enum is very specific because it contains the value None. This value is used to check the next case or do nothing otherwise. This is a class implementation trick that should have no impact on the rest of the code.

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
cinqmilleans: It's worse because this method modifies the active layout but its name suggests that it only chooses it. I will modify it. The return isn't clear either (true if a new layout was chosen, even if it's the same).

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
QuaintMako: A comment about this trick may be useful here, so it's not replicated anywhere else by accident if the knowledge get lost.

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
QuaintMako: Alright, this is something we want to keep track of, I'll copy part of it to #10831 since it may feed the discussion.

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
cinqmilleans: The method became void ApplyAdaptativeLayout. The return parameter was never used.

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
cinqmilleans: A comment have been added.

yaira2:
Can you outline the changes that were made?

src/Files.App/Helpers/AdaptiveLayoutHelpers.cs
```diff
@@ -3,124 +3,126 @@
...
+
+		public static void ApplyAdaptativeLayout(FolderSettingsViewModel folderSettings, string path, IList<ListedItem> filesAndFolders)
 		{
-			IUserSettingsService userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
+			if (!foldersSettingsService.EnableOverridingFolderPreferences)
+				return;
+			if (string.IsNullOrWhiteSpace(path))
+				return;
+			if (folderSettings.IsLayoutModeFixed || !folderSettings.IsAdaptiveLayoutEnabled)
+				return;
```
ferrariofilippo, line 27: I think it would be more efficient to have only one `if` with many `ORs`

yaira2: I think it's okay like this, makes it more readable.

yaira2:
@cinqmilleans can you resolve the merge conflicts?

QuaintMako (approved):

yaira2:
@cinqmilleans thank you

PR #10795, opened 2022-12-21
Fix: Fixed crash that would occur when updating from the store

puppetsw:
**Resolved / Related Issues**
- Fixes issue with store versions crashing when trying to show store update dialog window.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/ServicesImplementation/UpdateService.cs
+++ b/src/Files.App/ServicesImplementation/UpdateService.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Windows.Services.Store;
+using WinRT.Interop;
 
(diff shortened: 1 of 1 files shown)


src/Files.App/ServicesImplementation/UpdateService.cs
```diff
@@ -106,6 +107,10 @@ private async Task GetUpdatePackages()
 			try
 			{
 				_storeContext ??= await Task.Run(StoreContext.GetDefault);
+
+				var hwnd = WindowNative.GetWindowHandle(App.Window);
+				InitializeWithWindow.Initialize(_storeContext, hwnd);
```
gave92, line 112: You can use directly App.WindowHandle here I think.

yaira2 (approved):
LGTM

PR #10832, opened 2022-12-23
Codebase: Removed class Disposable

cinqmilleans:
**Resolved / Related Issues**
The abstract class Disposable is very generic. This class is moved to Shared so that it can be used in the SDK.

**Validation**
- [x] Built and ran the app


src/Files.Shared/Disposable.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace Files.App.Shell
+namespace Files.Shared
 {
 	public abstract class Disposable : IDisposable
```
d2dyno1, line 5: I think we should get rid of it

cinqmilleans: It's removed.

src/Files.App/Helpers/FileOperationsHelpers.cs
```diff
@@ -914,7 +914,13 @@ public void WaitForCompletion()
 				operationsCompletedEvent.WaitOne();
 			}
 
-			protected override void Dispose(bool disposing)
+			public void Dispose()
+			{
+				Dispose(true);
+				GC.SuppressFinalize(this);
```
d2dyno1, line 920: I'd remove that method and put everyting that's in `Dispose(bool)` here

yaira2: @cinqmilleans

PR #10839, opened 2022-12-24
Fix: Fixed issue where the network icon was blank

### LINKED ISSUE #10811: Bug: Missing network icon
### Description

When i open the app, the network icon is missing.

### Steps To Reproduce

-

### Steps to fix

- Display folder icon

### Files Version

Wersja: 2.4.4.0 OS Version: 10.0.22621.963

### Windows Version

Windows 11 22621.963

### Relevant Assets/Logs

Zrzut ekranu 2022-12-22 135549

hecksmosis:
**Resolved / Related Issues**
- Closes #10811 

**Validation**
- [*] Built and ran the app

**Screenshots (optional)**
Add screenshots here.


src/Files.App/DataModels/NavigationControlItems/DriveItem.cs
```diff
@@ -224,7 +224,14 @@ public async Task LoadDriveIcon()
 			if (IconData is null)
 			{
 				if (!string.IsNullOrEmpty(DeviceID))
+				{
+					if (string.Equals(DeviceID, "network-folder"))
```
QuaintMako, line 228: I'm not a huge fan of having three imbricated `if`s like this. Even if it's a detail, it grows the cognitive complexity by quite a margin. 

Maybe we could remove a level of nesting by inverting the first `if` like this at the beginning of the method?

```
if (IconData is not null)
{
	Icon = await IconData.ToBitmapAsync();
	return;
}
```

src/Files.App/DataModels/NavigationControlItems/DriveItem.cs
```diff
@@ -224,7 +224,14 @@ public async Task LoadDriveIcon()
...
+					{
+                        			Icon = await UIHelpers.GetIconResource(Constants.ImageRes.Folder);
+						return;
```
QuaintMako, line 231: This is really me being picky on the details here, but all the method is around fetching the `IconData` before having a final `Icon = await IconData.ToBitmapAsync();` at the end of the method.

I'd suggest to add an `async UIHelpers.GetIconData(int)` as the `async UIHelpers.GetIconResource(int)` that exists already, making it returning only the IconData, and then let the final instruction take care of it.

Besides, it will ensure that `IconData` is not null in case of `(string.Equals(DeviceID, "network-folder")` being true.

QuaintMako (commented):
Some discussions around the implementations, where I'm being really picky.

Otherwise, looks good to me!

QuaintMako:
Heckmosis' fix works for me. Just to be sure, is this the right icon we want @yair100 ?

yaira2:
Nope, it should use the standard folder icon.

hecksmosis:
@QuaintMako are you using a different icon set? Mine looks like this:
image

QuaintMako:
> @QuaintMako are you using a different icon set? 

The difference comes from W10 and W11.

src/Files.App/DataModels/NavigationControlItems/DriveItem.cs
```diff
@@ -221,17 +223,17 @@ public int CompareTo(INavigationControlItem other)
...
-				if (IconData is null)
-				{
-					var resource = UIHelpers.GetIconResourceInfo(Constants.ImageRes.Folder);
-					IconData = resource?.IconData;
-				}
-			}
+                Icon = await IconData.ToBitmapAsync();
+				return;
+            }
+
```
QuaintMako, line 231: The formatting here is spacing instead of tabs

hecksmosis: Should I change the `IconData is not null` and invert it to not repeat the `Icon = await IconData.ToBitmapAsync()`

QuaintMako (approved):
LGTM, with the right icon for W10.

PR #10847, opened 2022-12-26
Fix: Fixed issue where the wrong recent item was being removed

### LINKED ISSUE #10820: Bug: Removing an item from recent list doesn't work with PowerToys locksmith
### Description

If I have PowerToys file locksmith enabled it will open that on the file trying to remove. If I don't have it on then it doesn't remove the selected item but the last one in the list

### Steps To Reproduce

1. Right click an item in a recent list
2. Click remove
3. See last file from list is removed

### Steps to fix

- Remove selected item from list

### Files Version

2.4.8.0

### Windows Version

10.0.22621.963

### Relevant Assets/Logs

[debug.log](https://github.com/files-community/Files/files/10291334/debug.log)

gave92:
**Resolved / Related Issues**
- Closes part of #10820

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
+++ b/src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
@@ -112,7 +112,10 @@ namespace Files.App.UserControls.Widgets
 							for (; idx < recentFiles.Count; idx++) // Add new items (top of the list)
 							{
 								if (idx >= recentItemsCollection.Count || !recentFiles[idx].Equals(recentItemsCollection[idx]))
-									await AddItemToRecentListAsync(recentFiles[idx], idx);
+								{
+									if (!AddItemToRecentList(recentFiles[idx], idx)) // Not a new item
+										break;
+								}
 								else
 									break;
 							}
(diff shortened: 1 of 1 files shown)


src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -147,13 +150,16 @@ private async Task UpdateRecentsList(NotifyCollectionChangedEventArgs e)
...
 		/// </summary>
 		/// <param name="recentItem">The recent item to be added</param>
-		private async Task AddItemToRecentListAsync(RecentItem recentItem, int index = -1)
+		private bool AddItemToRecentList(RecentItem recentItem, int index = -1)
 		{
 			if (!recentItemsCollection.Any(x => x.Equals(recentItem)))
 			{
 				recentItemsCollection.Insert(index < 0 ? recentItemsCollection.Count : Math.Min(index, recentItemsCollection.Count), recentItem);
-				await recentItem.LoadRecentItemIcon();
+				_ = recentItem.LoadRecentItemIcon()
```
gave92, line 158: Optional change, can revert if asked: do not wait for recent item icon to be loaded.

yaira2: Looks good

hez2010:
Changes LGTM, but this doesn't fix the original issue. I can still reproduce the issue with the fix in this PR.

gave92:
The issue about powertoys or the issue of not removing the correct item?
If the second, could you share a video?

hez2010:
> The issue about powertoys or the issue about not removing the correct item? If the second, could you share a screen recording?

The issue about powertoys. Removing items from the recent list can still trigger the locksmith open, and the item didn't get  removed in the end.

gave92:
Ah ok here I'm only fixing the issue with powertoys off. I'm changing the description to not close the issue.

hez2010 (approved):
LGTM.

PR #10760, opened 2022-12-18
Fixes & improvements for Recent Items

gave92:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Fixes the issues listed in the comments below

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #10867, opened 2022-12-28
Performance: Optimized the recent files widget

jiejasonliu:
**Resolved / Related Issues**
- in general, this PR adds performance improvements through figuring out the type of change that occurs
  - more than often, changes only occur to one file (add/updated/removed); if it deviates, we can do a simple O(N) refresh
- also fixes a refresh bug from #10760
  - it seems that the prior solution falls short when a recent item is updated or when multiple changes occur, sometimes the OS batches changes to the recent items

**Validation**
- [X] Built and ran the app


jiejasonliu:
pinging @gave92 since you've been working around this area

src/Files.App/Filesystem/RecentItems.cs
```diff
@@ -205,6 +206,49 @@ public Task<bool> UnpinFromRecentFiles(RecentItem item)
 			}));
 		}
 
+		private NotifyCollectionChangedEventArgs GetChangedActionEventArgs(IReadOnlyList<RecentItem> oldItems, IList<RecentItem> newItems)
+		{
+			var intersection = oldItems.Intersect(newItems);
```
gave92, line 211: I'm not sure this works that way you intend it to. Intersect returns the items which are in common between two lists, not the ones which are different.
Also RecentItem did not propertly implement IEquatable so `intersection` was always empty. I've fixed this in 
595ac69.

gave92 (changes requested):

src/Files.App/Filesystem/RecentItems.cs
gave92: Edit: I've fixed GetChangedActionEventArgs() in c6d3cb5 the way I think it's right. Could you take a look?

jiejasonliu: totally right, i meant set difference not intersection 😅 

your changes look right to me, thanks for taking a lot at this!

gave92 (approved):

yaira2 (approved):
LGTM

PR #10886, opened 2022-12-30
Fix: Fixed issue where changing file permission for multiple files simultaneously would crash the app

### LINKED ISSUE #10870: Bug: Changing file permission for multiple files simultaneously crashes app
### Description

When setting permission ( readonly, etc. ) with Files' built-in properties dialog, Files will crash and exit after clicking OK button to apply settings.
This only happens when multiple items were chosen ( >= 2 items ), but no problems when dealing with single one item.

However, permissions of selected items will still be set properly even though Files crashes.

### Steps To Reproduce

1. Go to any path that contains multiple items.
2. Select multiple items ( >=2 items ), and click "Properties" on the menu bar or context menu to bring out properties dialog
3. Modify any kind of permission ( noo matter the type ) and click OK.
4. See the error

### Steps to fix

Looks like this is a standalone issue, no prerequisites should be accomplished before dealing with this one.
Log file is able to collect relevant error ( see the following )

### Files Version

Version： 2.4.8.0 OS Version: 10.0.22621.963

### Windows Version

Windows 11 22H2 22621.963

### Relevant Assets/Logs

2022-12-29 10:48:46.8620|INFO|CheckForUpdates|STORE: Checking for updates...
2022-12-29 10:48:48.2666|INFO|IsRunningOnArm|Running on ARM: False
2022-12-29 10:49:10.3062|ERROR|AppUnhandledException|Object reference not set to an instance of an object.
   at Files.App.Views.PropertiesGeneral.SaveChangesAsync(ListedItem item)
   at Files.App.Views.Properties.OKButton_Click(Object sender, RoutedEventArgs e)
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.<>c__DisplayClass2_0.<Post>b__0()
--- End of stack trace from previous location ---
   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|20_0(Int32 hr)
   at ABI.Windows.ApplicationModel.Core.IUnhandledErrorMethods.Propagate(IObjectReference _obj)
   at Microsoft.AppCenter.Utils.ApplicationLifecycleHelperWinUI.<.ctor>b__0_3(Object sender, UnhandledErrorDetectedEventArgs eventArgs)
--- End of stack trace from previous location ---
   at Microsoft.AppCenter.Utils.ApplicationLifecycleHelperWinUI.<.ctor>b__0_3(Object sender, UnhandledErrorDetectedEventArgs eventArgs)
   at ABI.System.EventHandler`1.Do_Abi_Invoke[TAbi](Void* thisPtr, IntPtr sender, TAbi args)

hecksmosis:
**Resolved / Related Issues**
- Closes #10870 

**Validation**
- [x] Built and ran the app


src/Files.App/Views/Pages/PropertiesGeneral.xaml.cs
```diff
@@ -22,9 +22,9 @@ public sealed partial class PropertiesGeneral : PropertiesTab
 
 		public PropertiesGeneral() => InitializeComponent();
 
-		public override async Task<bool> SaveChangesAsync(ListedItem item)
+		public override async Task<bool> SaveChangesAsync(ListedItem? item)
 		{
-			ViewModel.ItemName = ItemFileName.Text; // Make sure Name is updated
+			if (item is not null) ViewModel.ItemName = ItemFileName.Text; // Make sure Name is updated, except if there are multiple items
```
yaira2, line 27:
```suggestion
			if (item is not null)
					ViewModel.ItemName = ItemFileName.Text; // Make sure Name is updated, except if there are multiple items
```

yaira2 (dismissed):

QuaintMako:
While this does fix the crash for me, I'm not sure if the fix is the correct solution. 

It reminds me of #10728 where I fixed a crash with a surface fix, but Gave92 did some exploration and fixed an bigger underlying problem.

Here we have a null object, while the `Properties.xaml.cs` can hold the several objects in the page inside its `navParameterItem` field it seems. Mave the `SaveChangesAsync` should take a list of `ListedItem` instead of one? I am not sure, it's only a surface level analysis here.

I'd suggest to take this fix, for it fixes a pretty important problem : if the application crashes, you simply lose your files. And it's not gonna be in the recycle bin from what I can see. But we need to dive further into the problem and fix the root of the cause.

gave92:
Indeed fairly sure this would break renaming Drives

QuaintMako (approved):
LGTM.

src/Files.App/Views/Pages/Properties.xaml.cs
```diff
@@ -58,7 +56,7 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
 			var args = e.Parameter as PropertiesPageNavigationArguments;
 			AppInstance = args.AppInstanceArgument;
 			navParameterItem = args.Item;
-			listedItem = args.Item as ListedItem;
+			var listedItem = args.Item as ListedItem;
```
jiejasonliu, line 59: maybe to avoid further confusion we can completely get rid of this and instead of `listedItem is not null` we replace them all with `args.Item is ListedItem`?

it's not obvious because `args.Item` is of type object and can anything from a ListedItem, List<ListedItem>, a subclass, or something entirely different

hecksmosis: I'm not sure about this, as we would have to use `((ListedItem)args.Item)` quite a lot, which can be less confusing

hecksmosis: Wouldn't we have to use `(ListedItem)args.Item` quite a lot then?

gave92: Better? [89bcf75](https://github.com/files-community/Files/pull/10886/commits/89bcf753e1cb0215ca84c088015327e52f565ed3)

src/Files.App/Views/Pages/Properties.xaml.cs
```diff
@@ -56,18 +56,23 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+					(listedItem.IsShortcut && !listedItem.IsLinkItem)) ? Visibility.Visible : Visibility.Collapsed;
+				TabCompatibility.Visibility = (
+						FileExtensionHelpers.IsExecutableFile(listedItem is ShortcutItem sht ? sht.TargetPath : listedItem.FileExtension, true)
+					) ? Visibility.Visible : Visibility.Collapsed;
+				TabSecurity.Visibility = !listedItem.IsLibrary && !listedItem.IsRecycleBinItem ? Visibility.Visible : Visibility.Collapsed;
+			}
+			else if (args.Item is DriveItem)
+			{
+				TabSecurity.Visibility = Visibility.Visible;
+			}
```
QuaintMako, line 75: _Very_ minor, but several data are evaluated several times. Would store them in a variable be of any use instead? Speaking of `listedItem.IsLibrary`, `listedItem.IsShortcut`,`listedItem.FileExtension`.

hecksmosis: Ok

gave92: Ugh a bit excessive considering FileExtension is just a string property but ok :)

src/Files.App/Views/Pages/Properties.xaml.cs
jiejasonliu: yeah, definitely better imo

hecksmosis:
Is it ok now?

QuaintMako (approved):
LGTM.

gave92 (approved):

PR #10888, opened 2022-12-30
Fix HorizontalMultitaskingControl RightPaddingColumn.Width

cinqmilleans:
**Resolved / Related Issues**
It happened to me that the appWindowTitleBar.RightInset value was negative. This value is provided by the sdk, we do not calculate it. This may cause an exception because GridLength does not accept a negative value. This pr fixes that problem.

**Validation**
- [ ] Built and ran the app

### DIFF
--- a/src/Files.App/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
+++ b/src/Files.App/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
@@ -125,7 +125,7 @@ namespace Files.App.UserControls.MultitaskingControl
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


yaira2:
How was it negative?

cinqmilleans:
The HorizontalMultitaskingControl constructor throws an exception. This has no important consequence because it is intercepted by the xaml. However, there are 3 problems.
1) This exception is resource intensive.
2) If a future commit adds code to the end of this constructor, it cannot be executed randomly.
3) This may interfere with debugging another feature. The code should not cause any easily avoidable exceptions.

yaira2:
How can I reproduce this issue?

cinqmilleans:
I don't know how to reproduce it, but it's happened to me before.

yaira2 (approved):
LGTM

PR #10294, opened 2022-10-25
Fix: Fixed issue where it wasn't working to open properties from blank area of a folder

cinqmilleans:
**Resolved / Related Issues**
This pr mod improves the Property window.
- Fix the bug of the context menu which does not display the window for the current folder. closes #10289.
- The dialog always appears at top, like the settings. It's more consistent. This also prevents the window from appearing in the background, which sometimes happens in debug.
- It is now possible to open the properties of the current folder from the quick actions when nothing is selected. It's very useful.
- The code used to detect if folder is a drive has been simplified.
- The FilePropertiesHelpers class has been made more readable. Warnings have been fixed.

Cloud drives are displayed as ordinary folders but it will be possible to display them as drives when properties show remote sizes.

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [ ] Tested the changes for accessibility

PR #10897, opened 2023-01-01
Fix: Fixed issue where properties window was always on top

yaira2:
**Resolved / Related Issues**
- Fixed issue where properties window was always on top

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Helpers/FilePropertiesHelpers.cs
+++ b/src/Files.App/Helpers/FilePropertiesHelpers.cs
@@ -66,7 +66,6 @@ namespace Files.App.Helpers
 
 				var propertiesWindow = new WinUIEx.WindowEx
 				{
-					IsAlwaysOnTop = true,
 					IsMinimizable = false,
 					IsMaximizable = false,
 					MinWidth = 460,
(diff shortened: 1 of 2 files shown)


gave92:
Didn't we have a problem where Properties would open behind Files?

yaira2:
The issue is back (can't reproduce it consistently) but it introduced a bigger problem where the window was on top of everything else. This was most noticeable when trying to open the advanced permissions window.

gave92:
I think the idea was to put the window always on top only for a short time as a workaround, but we removed/ forgot the code to make it stop being on top after open.
Perhaps reopen tbe old issue regarding opening behind if any?

yaira2:
https://github.com/files-community/Files/issues/9926

ferrariofilippo:
Just checked the commit history and it seems that the line you removed (`IsAlwaysOnTop = true`) was added in #10294 
I think we could solve this problem implementing @gave92 solution proposed in #9973

gave92:
Oh you're right, I assumed that setting "alwaysontop" was the fix for #9926 but it wasn't. The fix from #9973 is still in place and appears to be working. I can't repro #9926 even after removing "alwaysontop".

PR #10900, opened 2023-01-02
Append app name to title

### LINKED ISSUE #10899: Feature: Append app name to title

Unlike UWP, setting the window title in WinUI 3 doesn't append the app name. This makes it harder to distinguish a folder window opened in Files from one opened in File Explorer.

### Requirements

When setting title, set it as folder name - Files.

### Files Version

2.4.0.0

### Windows Version

10.0.22000.1335

### Comments

_No response_

JaiganeshKumaran:
**Resolved / Related Issues**
- Closes #10899

Instead of hardcoding the app name "Files", it takes the name from the SystemInformation class instead. In dev mode the title will be "Folder name - Files - Dev". I would recommend changing "Files - Dev" to "Files (Dev)" or simply "Files Dev" because it's cleaner.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -1,4 +1,5 @@
 using CommunityToolkit.Mvvm.DependencyInjection;
+using CommunityToolkit.WinUI.Helpers;
 using Files.App.CommandLine;
 using Files.App.Filesystem;
 using Files.App.Helpers;
@@ -45,7 +46,7 @@ namespace Files.App
 		private void EnsureEarlyWindow()
 		{
 			// Set title
-			AppWindow.Title = "Files";
+			AppWindow.Title = SystemInformation.Instance.ApplicationName;
 
 			// Set icon
 			AppWindow.SetIcon(Path.Combine(Package.Current.InstalledLocation.Path, Constants.AssetPaths.Logo));
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -46,7 +46,7 @@ namespace Files.App
 		private void EnsureEarlyWindow()
 		{
 			// Set title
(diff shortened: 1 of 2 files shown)


src/Files.App/ViewModels/MainPageViewModel.cs
```diff
@@ -233,7 +234,7 @@ public async void UpdateInstanceProperties(object navigationArg)
 			if (AppInstances.Count > 1)
 				windowTitle = $"{windowTitle} ({AppInstances.Count})";
 			if (navigationArg == SelectedTabItem?.TabItemArguments?.NavigationArg)
-				App.GetAppWindow(App.Window).Title = windowTitle;
+				App.GetAppWindow(App.Window).Title = $"{windowTitle} - {SystemInformation.Instance.ApplicationName}";
```
yaira2, line 237:
```suggestion
				App.GetAppWindow(App.Window).Title = $"{windowTitle} - Files";
```

src/Files.App/MainWindow.xaml.cs
```diff
@@ -45,7 +46,7 @@ public MainWindow()
 		private void EnsureEarlyWindow()
 		{
 			// Set title
-			AppWindow.Title = "Files";
+			AppWindow.Title = SystemInformation.Instance.ApplicationName;
```
yaira2, line 49:
```suggestion
			AppWindow.Title = "Files";
```

yaira2 (approved):
LGTM

src/Files.App/MainWindow.xaml.cs
JaiganeshKumaran: Why? I do want it to distinguish between Dev, Preview and the release version.

yaira2: The icon helps with that

PR #10467, opened 2022-11-14
Code Quality: Refactoring of SidebarControl.xaml.cs

QuaintMako:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Related to #9895 

**What has been done**
- Refactored `src/Files.App/UserControls/SidebarControl.xaml.cs` extensively.
- Merged several duplicated sections into methods.
- Reduced the nesting levels.
- Nullable references annotations added.
- Formatting.

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #10910, opened 2023-01-03
Fix: Fixed crash from #10467

QuaintMako:
**Resolved / Related Issues**
- Related #10467 - A crash would occurs for some users when expanding/collapsing the sidebar since the refactoring, because of an access on a null object.

**What has been done**
- Added back the null check that seemingly prevents the crash to occur for the concerned users.

### DIFF
--- a/src/Files.App/UserControls/SidebarControl.xaml.cs
+++ b/src/Files.App/UserControls/SidebarControl.xaml.cs
@@ -448,7 +448,7 @@ namespace Files.App.UserControls
 				return;
 			}
 
-			var navigationPath = args.InvokedItemContainer.Tag.ToString();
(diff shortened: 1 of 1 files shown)


hecksmosis:
Why is the check failing?

QuaintMako:
> Why is the check failing?

Seemingly an issue on our certificates on Azure : 
```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VisualStudio\v17.0\AppxPackage\Microsoft.AppXPackage.Targets(4704,5): Error APPX0108: The certificate specified has expired. For more information about renewing certificates, see http://go.microsoft.com/fwlink/?LinkID=241478.```

PR #10936, opened 2023-01-05
Feature: Switched to StackPanel in the Details Layout template

yaira2:
**Resolved / Related Issues**
- Switched to StackPanel in the Details Layout template for better performance

**Issues**
These changes broke support for double clicking the grid splitter to resize the columns. This should be resolved before merging.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -595,11 +595,11 @@ namespace Files.App.Views.LayoutModes
 
 		#endregion IDisposable
 
-		private void Grid_Loaded(object sender, RoutedEventArgs e)
+		private void StackPanel_Loaded(object sender, RoutedEventArgs e)
 		{
 			// This is the best way I could find to set the context flyout, as doing it in the styles isn't possible
 			// because you can't use bindings in the setters
-			DependencyObject item = VisualTreeHelper.GetParent(sender as Grid);
+			DependencyObject item = VisualTreeHelper.GetParent(sender as StackPanel);
 			while (item is not ListViewItem)
 				item = VisualTreeHelper.GetParent(item);
 			if (item is ListViewItem itemContainer)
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -595,11 +595,11 @@ public override void Dispose()
 
 		#endregion IDisposable
 
-		private void Grid_Loaded(object sender, RoutedEventArgs e)
+		private void StackPanel_Loaded(object sender, RoutedEventArgs e)
 		{
 			// This is the best way I could find to set the context flyout, as doing it in the styles isn't possible
 			// because you can't use bindings in the setters
-			DependencyObject item = VisualTreeHelper.GetParent(sender as Grid);
+			DependencyObject item = VisualTreeHelper.GetParent(sender as StackPanel);
```
d2dyno1, line 602: Cast to DependencyObject instead of StackPanel. It'll help avoid such instances like this one

yaira2: This fixes the double click to size issue?

jiejasonliu: unfortunately not, resizing columns is broken in general because the logic depends on `Grid.Column` to know the correct `TextBlock` to calculate from

here's my naive attempt at fixing it, I'm sure there's a better way to avoid refactors breaking it, but [this is the best I can think of atm](https://github.com/jiejasonliu/Files/commit/59de71980f810e2a826729dd7b6a4f4d9bab5418) and will unblock your PR

yaira2: I'm fine with using this workaround, do you want to open a PR that targets this branch?

jiejasonliu: done

Issue #10171, opened 2022-10-10
Bug: Drag and drop fails when receiving app uses storage apis


Leoijeon:
### Description

In the windows file explorer, you can drag file to the xshell to upload file. The Files app do not support that.
Installed from Microsoft store. Not using a preview version.
![GIF](https://user-images.githubusercontent.com/30471002/194866370-6a4b70a8-26ba-4b86-abea-4b066e169104.gif)

### Steps To Reproduce

1. open xshell and files app
2. drag a file to xshell

### Steps to fix

- Use better drag/drop api

### Files Version

Version: 2.3.12.0 OS Version: 10.0.19044.2006

### Windows Version

Windows 10 Pro Version	21H2 OS build	19044.2006 Experience	Windows Feature Experience Pack 120.2212.4180.0

### Relevant Assets/Logs

[debug.log](https://github.com/files-community/Files/files/9745907/debug.log)
[debug_fulltrust.log](https://github.com/files-community/Files/files/9745910/debug_fulltrust.log)

yaichenbaum:
This is due to some API limitations left over from UWP, we're doing work to resolve this and it's high on our priority list.

hishitetsu:
@yair100 I don't think this issue should be closed yet. Although #11728 has increased the number of applications that can drag and drop, it still does not allow drag and drop to the xshell mentioned in this issue.

yaira2:
Can you open a new issue for this?

PR #10949, opened 2023-01-07
Fix: Fixed issues with drag & drop from Files to other apps

gave92:
**Resolved / Related Issues**

This PR should solve the issues we have when dragging items from Files to other apps. This should e.g work for UWP apps, the inbox Mail app, File explorer & other win32 apps.
Note: it references System.Windows.Forms, it's not mandatory but it makes things much simpler. If this works we can worry about removing all references to it later.

**Validation**
- [x] Built and ran the app


yaira2:
Fixes #10171?

yaira2 (approved):
LGTM

gave92:
Mmm it looks like this fixed drag&drop to some apps (WhatsApp) but breaks others (WordPad). And both fail to drag&drop to notepad or xshell (#10171)

yaira2:
What do we need to fix this issue?

gave92:
I'm stumped. Best to revert this for now on stable anyway.

PR #10966, opened 2023-01-09
Feature: Added shortcut for opening folders in secondary pane

### LINKED ISSUE #10907: Feature: Shortcut to open selected folder in Secondary Pane

Adding a shortcut to open a folder in the secondary pane would help make the secondary pane more useful, and accelerate its use without having to go into the context menu and selecting the option manually. Similar to the use of Ctrl + Enter to open a folder in a new tab.

### Requirements

- `ctrl` + `shift` + `enter` should open folder in new pane

### Files Version

2.4.8.0

### Windows Version

Version	10.0.22621 Build 22621

### Comments

_No response_

QuaintMako:
**Resolved / Related Issues**
- Closes #10907 

**What has been done**
- Made the process to open a folder in the dual pane into a method inside `NavigationHelpers`.
- Added the keyboard accelerator to both Grid View and Details View. (Not in Column View).
- Added a null check in `FileList_PreviewKeyDown` for each view.
- Fixed the missing icon on the flyout menu to open in dual pane option. (Used the one from the Settings)

**Validation**
- [x] Built and ran the app

**Screenshots**

With icon:
IconFlyoutMenu

Without icon:
image


hecksmosis:
Aside from that LGTM

QuaintMako:
> Aside from that LGTM and works for me

Aside from? I suppose your review did not get through.

hecksmosis:
Does it not appear?

hecksmosis:
Wouldn't it be better to rename the function to OpenInSecondaryPane for consistency?

src/Files.App/Helpers/NavigationHelpers.cs
```diff
@@ -36,6 +37,19 @@ public static Task<bool> OpenTabInNewWindowAsync(string tabArgs)
...
+				return;
+
+			IUserSettingsService userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
+			if (!userSettingsService.PreferencesSettingsService.IsDualPaneEnabled)
+				return;
+
+			if (listedItem is not null)
+				associatedInstance.PaneHolder?.OpenPathInNewPane((listedItem as ShortcutItem)?.TargetPath ?? listedItem.ItemPath);
+		}
+
```
ferrariofilippo, line 52: Why here `listedItem` is null checked twice?
I think you should also change `IUserSettingsService` to `var`

ferrariofilippo (changes requested):

src/Files.App/Helpers/NavigationHelpers.cs
yaira2: Good catch, looks like it can be removed

src/Files.App/Helpers/NavigationHelpers.cs
```diff
@@ -36,6 +37,19 @@ public static Task<bool> OpenTabInNewWindowAsync(string tabArgs)
 			return Launcher.LaunchUriAsync(folderUri).AsTask();
 		}
 
+		public static void OpenInSecondaryPane(IShellPage associatedInstance, ListedItem listedItem)
+		{
+			if(associatedInstance is null || listedItem is null)
+				return;
+
+			IUserSettingsService userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
```
yaira2, line 45: Can the service be declared at the top of the class?

ferrariofilippo (approved):
LGTM

yaira2 (approved):

src/Files.App/Helpers/ContextFlyoutItemHelper.cs
```diff
@@ -688,10 +688,10 @@ public static List<ContextMenuFlyoutItemViewModel> GetBaseItemMenuItems(BaseLayo
 				new ContextMenuFlyoutItemViewModel()
 				{
 					Text = "BaseLayoutItemContextFlyoutOpenInNewPane/Text".GetLocalizedResource(),
-					Glyph = "\uE117",
+					Glyph = "\xF117",
```
d2dyno1, line 691: What does it change?

yaira2: The previous glyph had a typo so it didn't load

d2dyno1 (approved):

PR #10967, opened 2023-01-09
Fix: Prevent creating folders with forbidden names

### LINKED ISSUE #9522: Bug: Files is allowing forbidden folder names (CON, NUL, etc)
### Description

It seems that Files doesn't prevent the user from creating files or folders with forbidden names that regular File Explorer would otherwise prevent.

It should prevent the user from creating files or folders named (in either lowercase or uppercase):

CON
PRN
AUX
NUL
COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, COM0
LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, LPT9, LPT0

con01
con02

### Steps To Reproduce

1. Open Files
2. Go to a place where you can freely create folders or files
3. Create a folder or a file with a forbidden name (list in the description)
4. Open file explorer and create a folder or a file with the same name
5. File explorer will not allow the creation of a folder or a file with that name

### Steps to fix

- Display text when forbidden folders names are entered into the new folder modal
- Display error content dialog when users try renaming a folder to con

### Files Version

2.3.3.0

### Windows Version

Windows 11 21H2 22000.778

### Relevant Assets/Logs

n/a

hecksmosis:
**Resolved / Related Issues**
- Closes #9522 

**Validation**
- [x] Built and ran the app


src/Files.App/Filesystem/FilesystemOperations/FilesystemOperations.cs
```diff
@@ -53,7 +53,7 @@ public async Task<(IStorageHistory, IStorageItem)> CreateAsync(IStorageItemWithP
 			FileSystemProgress fsProgress = new(progress, true, FileSystemStatusCode.InProgress, 1);
 			fsProgress.Report();
 
-			try
+				try
```
yaira2, line 56: ?

src/Files.App/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
```diff
@@ -96,6 +96,13 @@ public async Task<(ReturnResult, IStorageItem)> CreateAsync(IStorageItemWithPath
 			var progress = new Progress<FileSystemProgress>();
 			progress.ProgressChanged += (s, e) => returnStatus = returnStatus < ReturnResult.Failed ? e.Status!.Value.ToStatus() : returnStatus;
 
+			if (string.IsNullOrWhiteSpace(source.Name)
+				|| ContainsRestrictedCharacters(source.Name)
+				|| ContainsRestrictedFileName(source.Name))
+			{
+				return (ReturnResult.Failed, null);
```
yaira2, line 103: Does this display a dialog?

hecksmosis: No, but in CreateAsync no error displays a dialog

hecksmosis: Should I change it? @yair100

yaira2: What does the error dialog look like?

hecksmosis: The error dialog could be something like this, if we use existing strings already in resources.resw
image

yaira2: Strings would have to be modified but this is the right idea.

hecksmosis: Ok, I'll fix it now

hecksmosis: Done

gave92:
Why are not just checking for invalid names in the create file dialog?

yaira2:
> Why are not just checking for invalid names in the create file dialog?

Need to cover rename to as well.

hecksmosis:
Rename already has the check

hecksmosis:
If we cover this in the dialog we would have to check in multiple places, wouldn't we?

yaira2:
> Rename already has the check

What dialog is displayed there? We should reuse as much code as possible.

hecksmosis:
I've noticed there isn't actually a check for rename. (I've added one)

src/Files.App/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
```diff
@@ -538,6 +548,16 @@ public async Task<ReturnResult> RenameAsync(IStorageItemWithPath source, string
...
 
+			if (string.IsNullOrWhiteSpace(newName)
+				|| ContainsRestrictedCharacters(newName)
+				|| ContainsRestrictedFileName(newName))
+			{
+				await DialogDisplayHelper.ShowDialogAsync(
+					"ErrorDialogThisActionCannotBeDone".GetLocalizedResource(),
+					"ErrorDialogNameNotAllowed".GetLocalizedResource());
+				return ReturnResult.Failed;
+			}
```
yaira2, line 559: Is it worth creating a common method for this code?

hecksmosis: The return types change, so I think we should leave it as is

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2877,4 +2877,7 @@
   <data name="ReleaseNotes" xml:space="preserve">
     <value>Release Notes</value>
   </data>
+  <data name="ErrorDialogNameNotAllowed" xml:space="preserve">
+    <value>The item name specfied is invalid</value>
```
yaira2, line 2881:
```suggestion
    <value>The specified item name is invalid</value>
```

src/Files.App/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
```diff
@@ -538,6 +548,16 @@ public async Task<ReturnResult> RenameAsync(IStorageItemWithPath source, string
 			var progress = new Progress<FileSystemProgress>();
 			progress.ProgressChanged += (s, e) => returnStatus = returnStatus < ReturnResult.Failed ? e.Status!.Value.ToStatus() : returnStatus;
 
+			if (string.IsNullOrWhiteSpace(newName)
+				|| ContainsRestrictedCharacters(newName)
+				|| ContainsRestrictedFileName(newName))
```
QuaintMako, line 553: This condition is tested twice (at least) in the file. It should be extracted inside a method called `IsSuitedForFileName(string name)` or something akin.

src/Files.App/Filesystem/FilesystemOperations/Helpers/FilesystemHelpers.cs
```diff
@@ -538,6 +548,16 @@ public async Task<ReturnResult> RenameAsync(IStorageItemWithPath source, string
...
+			{
+				await DialogDisplayHelper.ShowDialogAsync(
+					"ErrorDialogThisActionCannotBeDone".GetLocalizedResource(),
+					"ErrorDialogNameNotAllowed".GetLocalizedResource());
+				return ReturnResult.Failed;
```
QuaintMako, line 558: This block is almost the same as l.103 to l.107 aside of the return time. Would it be possible to fuse them? I'm not sure how, it's not huge. Except if you find an idea right away, do not spend too much on it.

hecksmosis: I don't think there is an easy way to merge them without modifying changing a lot of code

QuaintMako: Agreeing with you on this one, let's not overdo it.

QuaintMako (dismissed):

QuaintMako (approved):

yaira2:
@hecksmosis thank you

PR #11013, opened 2023-01-14
Feature: Show error when a shortcut can't be created

### LINKED ISSUE #10987: Feature: "Can't create shortcut" modal

Currently, there is no modal for a failed shortcut creation

An example of this is when you try to create a shortcut for an item in the root drive.
image

### Requirements

- Display a modal for a failed **Create Shortcut** operation
- Ask the user if they want to place the shortcut to the Desktop instead.

### Files Version

2.4.21

### Windows Version

22623.1095

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**
- Closes #10987 

**Details**
- Fixed a bug with shortcut's names created from modal (lines 82-86 `CreateShortcutDialogViewModel.cs`
- Added `Can't create shortcut` modal (I'm not sure if it should be displayed when `File/Folder` properties are edited)

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2877,4 +2877,10 @@
   <data name="ReleaseNotes" xml:space="preserve">
     <value>Release Notes</value>
   </data>
+  <data name="CannotCreateShortcutDialogTitle" xml:space="preserve">
+    <value>Files is not authorized to create a shortcut here.</value>
```
yaira2, line 2881:
```suggestion
    <value>Files can't create a shortcut here.</value>
```

ferrariofilippo:
> I'm not sure if it should be displayed when File/Folder properties are edited

@yair100 what about this?

yaira2:
What do you mean by this?

ferrariofilippo:
At the moment, when the user opens `Properties` of a shortcut, he/she can edit its attributes. When this happens, Files tries to update the shortcut. If this is placed in a protected directory (such as `C:\`) nothing happens and the user doesn't get any error.

What I'm asking is whether the `Create on desktop` modal should be displayed in this specific case.
If the answer is no, should we inform the user that he/she needs to run Files as admin to successfully edit the shortcut?

yaira2:
I didn't realize it can be worked around if the user runs as administrator. We should update the message accordingly. Could try something like this "This action requires administrator privileges", "Please run Files as administrator to create this shortcut".

ferrariofilippo:
> I didn't realize it can be worked around if the user runs as administrator. We should update the message accordingly.

I think we should give the user the ability to create it on Desktop: if, for any reason , he can't run Files as admin, he should still be able to create that shortcut.

The modal might be like:
Title: `This action requires administrator privileges`
Message: `Please, run Files as administrator to create this shortcut or create it on Desktop`

ferrariofilippo:
I have to inform you that I have temporarily removed the binding from `ItemDestinationPath`. Now this property is being updated manually. I did this because I noticed that the Binding was not working anymore. In fact, if you try to create a shortcut (even in stable version), you'll see that it doesn't work

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2877,6 +2877,12 @@
   <data name="ReleaseNotes" xml:space="preserve">
     <value>Release Notes</value>
   </data>
+  <data name="CannotCreateShortcutDialogTitle" xml:space="preserve">
+    <value>This action requires administrator privileges.</value>
+  </data>
+  <data name="CannotCreateShortcutDialogMessage" xml:space="preserve">
+    <value>Please, run Files as administrator or create it on Desktop</value>
```
yaira2, line 2884:
```suggestion
    <value>Please run Files as administrator or create the shortcut on the Desktop instead.</value>
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2871,6 +2871,12 @@
   <data name="ReleaseNotes" xml:space="preserve">
     <value>Release Notes</value>
   </data>
+  <data name="CannotCreateShortcutDialogTitle" xml:space="preserve">
+    <value>This action requires administrator privileges.</value>
+  </data>
+  <data name="CannotCreateShortcutDialogMessage" xml:space="preserve">
+    <value>Please run Files as administrator or create the shortcut on the Desktop instead.</value>
```
yaira2, line 2878:
```suggestion
    <value>Please run Files as administrator or create the shortcut somewhere else instead.</value>
```

yair100 (dismissed):
LGTM

src/Files.App/Strings/en-US/Resources.resw
ferrariofilippo: @yair100 at the moment if the creation of the shortcut fails and the user hits `Create` the shortcut will be placed on the Desktop

yaira2: Thanks! I edited the strings to clarify that.

PR #11020, opened 2023-01-16
Feature: Improved the background color preview

yaira2:
**Resolved / Related Issues**
image

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/ValueConverters/ColorToSolidColorBrushValueConverter.cs
+++ b/src/Files.App/ValueConverters/ColorToSolidColorBrushValueConverter.cs
@@ -1,0 +1,30 @@
+﻿using Microsoft.UI.Xaml.Data;
+using Microsoft.UI.Xaml.Media;
+using System;
+using Windows.UI;
+
+namespace Files.App.ValueConverters
+{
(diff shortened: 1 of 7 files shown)


src/Files.App/ValueConverters/ColorToSolidColorBrushValueConverter.cs
```diff
@@ -0,0 +1,30 @@
...
+	{
+		public object? Convert(object value, Type targetType, object parameter, string language)
+		{
+			if (null == value)
+				return null;
+
+			if (value is Color)
+			{
+				Color color = (Color)value;
+				return new SolidColorBrush(color);
```
jiejasonliu, line 18: we'd probably want to convert the alpha channel to `255` to preserve behavior? 
or is it intentional that we want it to look opaque?

```cs
return new SolidColorBrush(Color.FromArgb(255, color.R, color.G, color.B));
```

yaira2: This is intentional, it's designed this way so that the app still shows the mica background.

Issue #11025, opened 2023-01-16
DetailsLayout Margin/Padding is not set correctly


ferrariofilippo:
### Description

When using details layout content shows in random columns. In the screenshot below, you can see that `Size` is displayed under `Date Modified`
This may be caused by margin/padding or by the switch from `Grid` to `StackPanel`

### Steps To Reproduce

1. Open a folder using DetailsLayout
2. See that file info is not aligned into columns

### Steps to fix

- Content should fit into columns
<em>or</em>
- Fix Margin/Padding

### Files Version

Version: 2.4.21.0 OS Version: 10.0.22621.1105

### Windows Version

Windows 11 Home Version 22H2 22621.963

### Relevant Assets/Logs

![1](https://user-images.githubusercontent.com/102259289/212634245-b5f1a7cc-55ca-4a96-ad4e-af2a1666b844.png)

yaira2:
This looks very wrong, what happens if you go to settings reset folder overrides?

ferrariofilippo:
Nothing changes, it remains as in the screenshot

yaira2:
I can't reproduce the issue, does this happen for all folders? Is there a specific column type that causes it?

Josh-65:
Happens to me when the tag column is enable and the columns span the window. Not as much as the screenshot but probably worsen on higher resolution displays.

ferrariofilippo:
Hiding tags column solve the problem for all folders but the recycle bin, which still appears as in the screenshot.
When tags column is visible, spacing (except in the recycle bin which looks as in the screenshot above) is a little bit off

![3](https://user-images.githubusercontent.com/102259289/212844402-cd017aed-1aa5-4fa1-980a-b6fcd7860c9c.png)

hecksmosis:
@yair100 Don't issues get automatically sorted into the task board?

Josh-65:
they have to be manually added

yaira2:
Is this only in recycle bin?

hecksmosis:
> they have to be manually added

I think this can be automated with Issue Bot

hecksmosis:
> Is this only in recycle bin?

Yes

ferrariofilippo:
> ![3](https://user-images.githubusercontent.com/102259289/212844402-cd017aed-1aa5-4fa1-980a-b6fcd7860c9c.png)

This is not recycle bin

yaira2:
[...]

PR #10585, opened 2022-11-27
Fix: Original path and date deleted column sizes aren't saved

gave92:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #10535

Remember status, date deleted, original path columns settings when "enable individual preferences" is off. Same code as for all the other columns.
Also adds missing "Show sync status column" checkbox when right-clicking on details view header.

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #10560, opened 2022-11-24
Only save necessary info to folder preferences

gave92:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- None, small optimization

Currently we are serializing unnecessary/calculated fields of ColumnsViewModel to the preference DB. This PR adds [BsonIgnore] to those.

Before (each column per each folder):
```json
"NameColumn": 
{
  "LengthIncludingGridSplitter": 
  {
    "Value": 248.0,
    "GridUnitType": "Pixel",
    "IsAbsolute": true,
    "IsAuto": false,
    "IsStar": false
  },
  "Length": 
  {
    "Value": 240.0,
    "GridUnitType": "Pixel",
    "IsAbsolute": true,
    "IsAuto": false,
    "IsStar": false
  },
  "UserCollapsed": false,
  "UserLength": 
  {
    "Value": 240.0,
    "GridUnitType": "Pixel",
    "IsAbsolute": true,
    "IsAuto": false,
    "IsStar": false
  },
  "UserLengthPixels": 240.0
},
```

After:
```json
"NameColumn": 
{
  "UserCollapsed": false,
  "UserLengthPixels": 240.0
}
```

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #10239, opened 2022-10-17
Fix: Fixed issue where columns sometimes had no width

ferrariofilippo:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Related #10117 #9867 

**Comments**
There was a bug involving hidden columns.
Steps to reproduce:
- Hide at least one column
- Set column widths as default
- Show one of the hidden columns
- See that its width is 0px

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #11027, opened 2023-01-16
Fix: Fixed issue with resizing the status column

### LINKED ISSUE #10432: Feature: Allow resizing the status column

### LINKED ISSUE #11103: Bug: Recycle bin original path and data deleted columns are set to 0 width
### Description

When I edit any column in details view out of the recycle bin original path and date deleted will be set to 0

### Steps To Reproduce

1. Turn off Files settings > Folder > Enable overriding folder preferences in individual directories
2. Go to recycle bin and set either original path or date deleted
3. Go to any other folder and edit size of a column
4. Go back to recycle bin and see both columns at 0 width

### Steps to fix

- Keep width of columns when others are edited.

### Files Version

2.4.25.0

### Windows Version

10.0.22621.1105

### Log file

[debug.log](https://github.com/files-community/Files/files/10527647/debug.log)

ferrariofilippo:
…olumns

**Resolved / Related Issues**
- Closes #11103 
- Closes #10432
- Related #10585 
- Related #10560
- Related #10117
- Related #10239

**Comments**
- I added a null check in `Win32Shell` to avoid an exception that was getting me mad
- To fix this bug we need to save the default size only when the column is visible. We have two ways to achieve this:
  - We can check if each column is actually visible (is not collapsed and has width > 0) -> This is what i proposed in this PR
  - We can use the less readable and less understandable, but faster approach I first proposed in #10117 (see `Unresolved Conversation`). I may add a comment to the first width setter that specify when width = 0

Let me know what you think

**Validation**
- [x] Built and ran the app


QuaintMako (approved):

yaira2:
How is it different than any of the other columns that can be collapsed?

ferrariofilippo:
> How is it different than any of the other columns that can be hidden, can't we reuse the logic?

The problem was that `Hidden` columns were not `Collapsed`, and so their sizes were saved. I changed the logic so that when a column is hidden it's collapsed too.

yaira2:
Does this fix #11025?

ferrariofilippo:
> Does this fix #11025?

It doesn't

yaira2 (approved):
LGTM

PR #11028, opened 2023-01-16
Fix: Fixed issue with using arrow keys while renaming a file

### LINKED ISSUE #9691: Renaming a file using the arrow keys to navigate words in its name, will start selecting other files
### Description

When I'm renaming a file, if I use the arrow keys to navigate between words of its name, the file rename will abort as soon as I navigate to the last or first characters of the name (including extension) and Files start selecting nearby files. 

### Steps To Reproduce

1. Rename file
2. Use left arrow key continuously to rename first characters of filename or right arrow key to rename last characters
3. Files aborts rename operation
4. Files starts selecting other files/folders with the arrow keys

### Expected behavior

Files should only leave the rename operation after an explicit Save or Abort by the **user**

### Files Version

Versão: 2.3.3.0 

### Windows Version

OS Version: 10.0.22621.317 (Win11 22H2 Release Preview)

### Relevant Assets/Logs

[debug.log](https://github.com/files-community/Files/files/9299528/debug.log)
[debug_fulltrust.log](https://github.com/files-community/Files/files/9299529/debug_fulltrust.log)

ferrariofilippo:
… other files

**Resolved / Related Issues**
- Closes #9691

**Validation**
- [x] Built and ran the app


yaira2:
Does this need to be applied to the other layouts? Why was the search control modified?

ferrariofilippo:
> as soon as I can I'll do it for details and columns.

Left/Right arrows work. Up/Down select files above/below. I'll fix this

ferrariofilippo:
Last commit doesn't fully resolve the problem. There's a bug when you're at the beginning of a word and press `Down Arrow`. Nothing happens in this case and also in the opposite one (you're at the end and press `Up Arrow`). At the moment I know that in this case `KeyDown` event is not raised. We can't resolve the issue adding an event handler to `PreviewKeyDown` since this event gives no information about modifiers

PR #11031, opened 2023-01-17
Codebase: Refactored themes

yaira2:
**Resolved / Related Issues**
- Switched property in view model and resource factory to string. This accounts for empty strings which indicate the default value. If we didn't use blank strings to indicate default, we'd have to provide a separate value for light and dark mode.
- Added `LoadAppResources` method to `AppThemeResourcesHelper`
- Call `LoadAppResources` from the ViewModel instead of the View. 

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/ValueConverters/StringToSolidColorBrushValueConverter.cs
+++ b/src/Files.App/ValueConverters/StringToSolidColorBrushValueConverter.cs
@@ -1,20 +1,21 @@
-﻿using Microsoft.UI.Xaml.Data;
+﻿using CommunityToolkit.WinUI.Helpers;
+using Microsoft.UI.Xaml.Data;
 using Microsoft.UI.Xaml.Media;
 using System;
 using Windows.UI;
 
 namespace Files.App.ValueConverters
 {
-	public class ColorToSolidColorBrushValueConverter : IValueConverter
+	public class StringToSolidColorBrushValueConverter : IValueConverter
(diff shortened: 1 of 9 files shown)


src/Files.App/ValueConverters/StringToSolidColorBrushValueConverter.cs
```diff
@@ -1,20 +1,21 @@
...
 		public object? Convert(object value, Type targetType, object parameter, string language)
 		{
 			if (null == value)
 				return null;
 
-			if (value is Color)
+			if (value is string)
 			{
-				Color color = (Color)value;
+				Color color = ColorHelper.ToColor((string)value);
```
jiejasonliu, line 18: minor nit, but you can avoid this cast using pattern matching.
```cs
if (value is string colorString) 
{
    Color color = ColorHelper.ToColor(colorString);
}

yaira2: Is there a benefit to it or just neater code?

jiejasonliu: just for consistency! this type of pattern matching to cast is predominant in the codebase

jiejasonliu (approved):

PR #11034, opened 2023-01-17
Code Quality: Shell page refactoring

ferrariofilippo:
This PR aims to refactor `ColumnShellPage` and `ModernShellPage`. They have a lot of code in common and I think we should create a base class to avoid code duplication and simplify future maintenance. 
This is still a draft, I have to check that everything works as expected.
@lukeblevins I'd be glad if you could help me with this refactoring


QuaintMako:
@ferrariofilippo is there work left on this PR or could it be theorically ready for merging?

ferrariofilippo:
Theoretically it's ready but @yair100 asked me to wait a review from @lukeblevins

src/Files.App/Views/BaseShellPage.cs
```diff
@@ -0,0 +1,739 @@
...
+
+		protected readonly StorageHistoryHelpers storageHistoryHelpers;
+
+		protected readonly CancellationTokenSource cancellationTokenSource;
+
+		protected IDialogService DialogService { get; } = Ioc.Default.GetRequiredService<IDialogService>();
+
+		protected IUserSettingsService UserSettingsService { get; } = Ioc.Default.GetRequiredService<IUserSettingsService>();
+
+		protected IUpdateService UpdateSettingsService { get; } = Ioc.Default.GetRequiredService<IUpdateService>();
```
QuaintMako, line 52: Would need to be converted into `protected readonly` fields.

src/Files.App/Views/BaseShellPage.cs
```diff
@@ -0,0 +1,739 @@
...
+			this.PointerPressed += CoreWindow_PointerPressed;
+
+			/*
+			TODO UA307 Default back button in the title bar does not exist in WinUI3 apps.
+			The tool has generated a custom back button in the MainWindow.xaml.cs file.
+			Feel free to edit its position, behavior and use the custom back button instead.
+			Read: https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/migrate-to-windows-app-sdk/case-study-1#restoring-back-button-functionality
+			*/
+			//SystemNavigationManager.GetForCurrentView().BackRequested += ModernShellPage_BackRequested;
+
```
QuaintMako, line 219: This line will be broken when uncommented.

src/Files.App/Views/BaseShellPage.cs
```diff
@@ -0,0 +1,739 @@
...
+		/**
+		 * Some keys are overriden by control built-in defaults (e.g. 'Space').
+		 * They must be handled here since they're not propagated to KeyboardAccelerator.
+		 */
+		protected void ShellPage_PreviewKeyDown(object sender, KeyRoutedEventArgs args)
+		{
+			var ctrl = Microsoft.UI.Input.InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Control).HasFlag(CoreVirtualKeyStates.Down);
+			var shift = Microsoft.UI.Input.InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Shift).HasFlag(CoreVirtualKeyStates.Down);
+			var alt = Microsoft.UI.Input.InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Menu).HasFlag(CoreVirtualKeyStates.Down);
```
QuaintMako, line 269: `using Microsoft.UI.Input`

src/Files.App/Views/BaseShellPage.cs
```diff
@@ -0,0 +1,739 @@
...
+					if (SlimContentPage?.SelectedItem?.PrimaryItemAttribute == StorageItemTypes.Folder)
+						path = SlimContentPage.SelectedItem.ItemPath;
+
+					// TODO open path in Windows Terminal
+					args.Handled = true;
+					break;
+
+				case (false, false, false, true, VirtualKey.Space): // space, quick look
+																	// handled in `CurrentPageType`::FileList_PreviewKeyDown
+					break;
```
QuaintMako, line 288: Is this case truly needed ?

src/Files.App/Views/BaseShellPage.cs
```diff
@@ -0,0 +1,739 @@
...
+		public virtual void Forward_Click()
+		{
+			var incomingPageContent = ItemDisplay.ForwardStack[ItemDisplay.ForwardStack.Count - 1];
+			var incomingPageNavPath = incomingPageContent.Parameter as NavigationArguments;
+			incomingPageNavPath.IsLayoutSwitch = false;
+			if (incomingPageContent.SourcePageType != typeof(WidgetsPage)) // Update layout type
+				InstanceViewModel.FolderSettings.GetLayoutType(incomingPageNavPath.IsSearchResultPage ? incomingPageNavPath.SearchPathParam : incomingPageNavPath.NavPathParam);
+			SelectSidebarItemFromPath(incomingPageContent.SourcePageType);
+			ItemDisplay.GoForward();
+		}
```
QuaintMako, line 521: Those two methods could be refined to use the same logic to fetch the layout type. Only the first line and the last few lines would be different in the Forward and Backward.

QuaintMako (commented):
A first quick review of BaseShell. Looks overall promising, but I need to now look at the two existing ShellPage to make sure the code removed have lost no instruction. 
Thanks for this PR again !

yair100:
> Theoretically it's ready but @yair100 asked me to wait a review from @lukeblevins

I think we can move forward and if Luke is able to he'll add his review.

ferrariofilippo:
@yair100 with last commit I fixed a `TODO` (support for `Open in terminal` shortcut). I did that in this PR to avoid having to rewrite the code after merging this PR. Do you think I should open a different PR?
BTW, I think I should add a setting to choose the `Default Terminal` (would also be useful for Terminal Integration)

yaira2:
> Do you think I should open a different PR?

Should be in a separate PR

> BTW, I think I should add a setting to choose the Default Terminal (would also be useful for Terminal Integration)

We used to have this but removed it when Windows added a built-in setting.

QuaintMako (dismissed):
LGTM, I think I haven't seen any missing code. Saw a few things that could be cleaned up that I'll fix an upcoming PR, but it's out of the scope.

Thanks Ferrario for that chunky work!

yair100 (dismissed):
Hoping we can gid rid of modern shell page altogether in the near future

QuaintMako:
> Hoping we can gid rid of modern shell page altogether in the near future

Let's open an issue for that one and write down what would be needed to. The new command system would be a step in the right direction for one.

Issue #7967, opened 2022-01-27
Automatically delete/discard illegal filename characters when *pasting* a string on a *new* file/folder


lightningHoneypot:
### What's the Problem?

In Files Explorer, every time when you paste a string that contains: 

- :
- &
- "
- ?
- <
- (>)
- (#)
- {
- }
- % 
- ~
- /
- \

It will automatically delete the string **and** displays the following message
> file names can't contain the following characters: & " ? < > # { } % ~ / \.

If the user tries again by deleting and pasting the same string, File Explorer will automatically post the string of text without those characters as if the user didn't copy them in the first place. (see video 2)

Currently, Files (Community) does not support this feature, but this idea could help people like for example: a student that is organizing their folders when an instructor releases new materials by the week. Small but significant features that are in Files Explorer should be implemented into Files (Community) so the program could be seen as a replacement to the venerable but dated Files Explorer.

https://user-images.githubusercontent.com/88814097/151285775-efbd8a5d-b18a-402c-81e8-f786b8d13c9f.mov

https://user-images.githubusercontent.com/88814097/151285778-d48a250b-6266-464d-8c5c-05aeecf8c354.mov

### Solution/Idea

When the user pastes a string when naming a new file/folder with these illegal filename characters:

- :
- &
- "
- ?
- <
- (>) 
- (#) 
- {
- }
- % 
- ~
- /
- \

It should automatically delete the string **and** displays the following message
> file names can't contain the following characters: & " ? < > # { } % ~ / \.

If the user tries again by deleting and pasting the same string, Files should automatically post the string of text without those characters as if the user didn't copy them in the first place. (see video 2)

### Alternatives

There is currently no alternative to what I describe other than not implementing it, it's a small feature but it could help some people.

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow end users to not manually delete a copied text that has an illegal filename character(s) when naming files/folders | Must |
| If the user tries again by deleting and pasting the same string, Files (Community) will automatically post the string of text without those illegal filename characters as if the user didn't copy them in the first place, just like Files Explorer. | Must/Should |

### Files Version

Version: 2.1.2.0 OS Version: 10.0.19043.1466

### Windows Version

Windows 10 21H1 OS Build 19043.1466

yaichenbaum:
This issue should be resolved with #8204

PR #11040, opened 2023-01-19
Fix: Fixed create file button when entering text with Chinese input flyout

### LINKED ISSUE #9563: Bug: Create file button is disabled when using Chinese input flyout
### Description

When I try to create a folder or file and try to type the name in Chinese, I can't create it immediately(the button below remains grey.).
Rather, I have to delete one character to make the ‘create’ button available.

### Steps To Reproduce

1. Use input flyout to enter text into the name field without actually typing with the keyboard
2. See that the create file button is disabled

### Expected behavior

- Enable create file button no matter the input type

### Files Version

Version: 2.3.3.0 OS Version: 10.0.22000.795

### Windows Version

Windows 11 21H2 22000.795

### Relevant Assets/Logs

problem

hecksmosis:
**Resolved / Related Issues**
- Closes #9563 

**Validation**
- [x] Built and ran the app


src/Files.App/Helpers/DynamicDialogFactory.cs
```diff
@@ -103,6 +103,18 @@ public static DynamicDialog GetFor_RenameDialog()
 				}
 			};
 
+			inputText.TextChanged += (textBox, args) =>
+			{
+				if (!string.IsNullOrWhiteSpace(inputText.Text))
```
yaira2, line 108: Does this check for invalid characters?

hecksmosis: No, but as the text has already been changed we cannot revert to the old one.

hecksmosis: @yair100 What should we do then?

yaira2: > No, but as the text has already been changed we cannot revert to the old one.

The best thing might be to do a little refactoring and remove the TextChanging event so the behavior is consistent. We can disable the create button if the name isn't valid and display an explanation in a TextBlock.

src/Files.App/Helpers/DynamicDialogFactory.cs
```diff
@@ -103,6 +103,18 @@ public static DynamicDialog GetFor_RenameDialog()
...
+				{
+					dialog.ViewModel.DynamicButtonsEnabled = DynamicDialogButtons.Primary | DynamicDialogButtons.Cancel;
+				}
+				else
+				{
+					dialog.ViewModel.DynamicButtonsEnabled = DynamicDialogButtons.Cancel;
+				}
```
QuaintMako, line 115: Can be simplified (here without the proper indentation):  
```suggestion
dialog.ViewModel.DynamicButtonsEnabled = string.IsNullOrWhiteSpace(inputText.Text) 
    ? DynamicDialogButtons.Cancel 
    : DynamicDialogButtons.Cancel | DynamicDialogButtons.Primary ;
```

src/Files.App/Helpers/DynamicDialogFactory.cs
```diff
@@ -57,50 +57,26 @@ public static DynamicDialog GetFor_ShortcutNotFound(string targetPath)
...
-					else
-					{
-						dialog.ViewModel.DynamicButtonsEnabled = DynamicDialogButtons.Cancel;
-					}
-
-					tipText.Opacity = 0.0d;
-				}
+				tipText.Opacity = FilesystemHelpers.ContainsRestrictedCharacters(inputText.Text) ? 1.0d : 0.0d;
+				dialog!.ViewModel.DynamicButtonsEnabled = string.IsNullOrWhiteSpace(inputText.Text) || FilesystemHelpers.ContainsRestrictedCharacters(inputText.Text) 
+														? DynamicDialogButtons.Cancel : DynamicDialogButtons.Primary | DynamicDialogButtons.Cancel;
```
QuaintMako, line 79:
```suggestion
														? DynamicDialogButtons.Cancel 
														: DynamicDialogButtons.Primary | DynamicDialogButtons.Cancel;
```

QuaintMako: Jump a line.

src/Files.App/Helpers/DynamicDialogFactory.cs
```diff
@@ -57,50 +57,27 @@ public static DynamicDialog GetFor_ShortcutNotFound(string targetPath)
...
-						dialog.ViewModel.DynamicButtonsEnabled = DynamicDialogButtons.Primary | DynamicDialogButtons.Cancel;
-					}
-					else
-					{
-						dialog.ViewModel.DynamicButtonsEnabled = DynamicDialogButtons.Cancel;
-					}
-
-					tipText.Opacity = 0.0d;
-				}
+				tipText.Opacity = FilesystemHelpers.ContainsRestrictedCharacters(inputText.Text) ? 1.0d : 0.0d;
```
yaira2, line 77: x:Load instead?

hecksmosis: It was used with opacity previously, that's why I decided to leave it that way.

hecksmosis: How would you use x:load here (sorry I am quite new to xaml)

yaira2: We can create a Boolean and bind the x:Load property of the text to the property.

hecksmosis: But there is no actual xaml for the text, is there?

hecksmosis: I think we can do this with Visibility if you prefer but as there is no xaml I don't think we can use x:Load

hecksmosis: Also, if we make the textblock unload or make it disappear, the dialog box size changes abruptly, and I don't think that looks good

yaira2: Forgot about the size changing, let's leave as is then.

yaira2 (dismissed):
LGTM

jiejasonliu:
this change will regress #7967, just wanted to point out if we're okay with that

yaira2:
> this change will regress #7967, just wanted to point out if we're okay with that

Thanks for the heads up, this is okay.

src/Files.App/Helpers/DynamicDialogFactory.cs
```diff
@@ -57,50 +57,27 @@ public static DynamicDialog GetFor_ShortcutNotFound(string targetPath)
...
+				dialog!.ViewModel.DynamicButtonsEnabled = string.IsNullOrWhiteSpace(inputText.Text) || FilesystemHelpers.ContainsRestrictedCharacters(inputText.Text)
```
QuaintMako, line 78: Shouldn't the test test also for `ContainsRestrictedFileName(name)` ?
If so, we could use here the `IsValidForFilename` method that is introduced in #10967 to simplify the test.

hecksmosis: I think we should merge #10967 first and then we can use the method

QuaintMako: @yair100

QuaintMako: @hecksmosis your move to update and use the method now!

hecksmosis: Done

src/Files.App/Helpers/DynamicDialogFactory.cs
```diff
@@ -66,18 +66,18 @@ public static DynamicDialog GetFor_RenameDialog()
...
 			inputText.TextChanged += (textBox, args) =>
 			{
-				tipText.Opacity = FilesystemHelpers.ContainsRestrictedCharacters(inputText.Text) ? 1.0d : 0.0d;
-				dialog!.ViewModel.DynamicButtonsEnabled = string.IsNullOrWhiteSpace(inputText.Text) || FilesystemHelpers.ContainsRestrictedCharacters(inputText.Text) 
-														? DynamicDialogButtons.Cancel 
-														: DynamicDialogButtons.Primary | DynamicDialogButtons.Cancel;
+				tipText.Opacity = FilesystemHelpers.IsValidForFilename(inputText.Text) ? 0.0d : 1.0d;
+				dialog!.ViewModel.DynamicButtonsEnabled = FilesystemHelpers.IsValidForFilename(inputText.Text)
+														? DynamicDialogButtons.Primary | DynamicDialogButtons.Cancel
+														: DynamicDialogButtons.Cancel;
```
QuaintMako, line 80: We can win one test by stocking the value in a variable before hand and reusing them.
```suggestion
				var isInputValid = FilesystemHelpers.IsValidForFilename(inputText.Text);
				tipText.Opacity = isInputValid ? 0.0d : 1.0d;
				dialog!.ViewModel.DynamicButtonsEnabled = isInputValid 
														? DynamicDialogButtons.Primary | DynamicDialogButtons.Cancel
														: DynamicDialogButtons.Cancel;
```

QuaintMako (commented):
That should be the last bit I can squeeze by being picky. LGTM otherwise.

yaira2 (approved):

PR #11063, opened 2023-01-23
Feature: Use the tab key to select the next file while renaming

### LINKED ISSUE #6675: Enable the use of the Tab key to select next file while renaming
### What's the Problem?

The default files explorer enables us to rename the next file by pressing the Tab key while renaming. By pressing Shift + Tab, we can rename the file before the one currently being renamed. This is a very useful feature when renaming many files located in the same folder.

While not renaming, the tab key could also be set to navigate the selection between files, so acting like arrow up/down currently does.

### Solution/Idea

Add the shortcuts with Tab and Shift + Tab to select next/previous file in renaming mode.

### Alternatives

An alternative would be to add the possibility to rename the next file by pressing the down arrow, and the previous one with the up arrow.

### Priorities

| Capability | Priority |
| :---------- | :------- |

### Files Version

v2

### Windows Version

Windows 10

### Additional comment

_No response_

### LINKED ISSUE #9691: Renaming a file using the arrow keys to navigate words in its name, will start selecting other files

ferrariofilippo:
**Resolved / Related Issues**
- Closes #9691
- Closes #6675

**Comments**
There's a bug when you're at the beginning of a word and press Down Arrow. Nothing happens (also when you're at the end and press Up Arrow). Neither `KeyDown` nor `PreviewKeyDown` events are raised. `GridView` layout is not affected by the bug (but `Tiles` is)
I'll try to find a solution

**Validation**
- [x] Built and ran the app


QuaintMako (dismissed):
LGTM and works on my machine.

I do have the issue where the down arrow doesn't do anything when at the start of the name as well. It'd be nice to fix it indeed.

yaira2:
I think key + up/down should finish the rename and start renaming the item above/below it.

ferrariofilippo:
> Actually, that's tab/shift tab

Added. I had to use `User32.dll` to get `Shift` state because `InputKeyboard.GetKeyStateForCurrentThread()` caused some issues

ferrariofilippo:
> I do have the issue where the down arrow doesn't do anything when at the start of the name as well.

`Ctrl+Up/Down` arrow works. This may be a WinUI bug as it involves all textboxes (e.g. new file modal or properties window)

src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
```diff
@@ -194,27 +197,60 @@ protected virtual void RenameTextBox_LostFocus(object sender, RoutedEventArgs e)
...
+				case VirtualKey.Left:
+					e.Handled = textBox.SelectionStart == 0;
+					break;
+				case VirtualKey.Right:
+					e.Handled = (textBox.SelectionStart + textBox.SelectionLength) == textBox.Text.Length;
+					break;
+				case VirtualKey.Tab:
+					textBox.LostFocus -= RenameTextBox_LostFocus;
+					textBox.Text = OldItemName;
+					EndRename(textBox);
```
yaira2, line 231: Shouldn't this commit the name?

ferrariofilippo: Yes, it make sense. I'll need some time to implement it, since committing a name causes the file list to refresh

ferrariofilippo:
I changed `CommitRename` from `async void` to `async Task` to avoid a bug where files/folders got duplicated.
I need to test this new feature further, but it seems to work.

ferrariofilippo:
There's a problem with `ListViewBase.ContainerFromItem()` which sometimes returns null even when the item exists. Once I figured out how to fix this, there should be no more bugs.

ferrariofilippo:
Everything should work now

QuaintMako (approved):
LGTM and works for me.

hecksmosis (approved):
LGTM

yaira2:
@ferrariofilippo can you confirm this resolves #6675?

ferrariofilippo:
Yes

yaira2 (approved):
LGTM

hecksmosis (approved):
LGTM and seems to work

yaira2:
I tested, it not only seems to work but it's a great improvement 😅

PR #11069, opened 2023-01-25
Feature: Display item size in tooltip

yaira2:
**Changes**
- Added item size to the tooltip
- Display correct tooltip in the column layout

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Filesystem/ListedItem.cs
+++ b/src/Files.App/Filesystem/ListedItem.cs
@@ -15,6 +15,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.IO;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Windows.Storage;
--- a/src/Files.App/Filesystem/ListedItem.cs
+++ b/src/Files.App/Filesystem/ListedItem.cs
@@ -49,6 +49,9 @@ namespace Files.App.Filesystem
 				return $"{"ToolTipDescriptionName".GetLocalizedResource()} {Name}{Environment.NewLine}" +
 					$"{"ToolTipDescriptionType".GetLocalizedResource()} {itemType}{Environment.NewLine}" +
 					$"{"ToolTipDescriptionDate".GetLocalizedResource()} {ItemDateModified}" +
+					(!String.IsNullOrWhiteSpace(FileSize)
+						? $"{Environment.NewLine}{"ToolTipDescriptionSize".GetLocalizedResource()} {FileSize}"
+						: string.Empty) +
 					(SyncStatusUI.LoadSyncStatus
 						? $"{Environment.NewLine}{"syncStatusColumn/Header".GetLocalizedResource()}: {syncStatusUI.SyncStatusString}"
 						: string.Empty);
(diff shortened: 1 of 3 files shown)


src/Files.App/Filesystem/ListedItem.cs
```diff
@@ -49,6 +49,9 @@ public string ItemTooltipText
 				return $"{"ToolTipDescriptionName".GetLocalizedResource()} {Name}{Environment.NewLine}" +
 					$"{"ToolTipDescriptionType".GetLocalizedResource()} {itemType}{Environment.NewLine}" +
 					$"{"ToolTipDescriptionDate".GetLocalizedResource()} {ItemDateModified}" +
+					(!String.IsNullOrWhiteSpace(FileSize)
+						? $"{Environment.NewLine}{"ToolTipDescriptionSize".GetLocalizedResource()} {FileSize}"
+						: string.Empty) +
```
QuaintMako, line 54: Worth using a string builder here.

QuaintMako (commented):
Can this process be costy? Is the size calculated on the fly? I cannot check right now, but if so, would be good to check for optimisations.

src/Files.App/Filesystem/ListedItem.cs
yaira2: It's using the exiting property, so I think it's okay. I'd rather leave a refactor to a separate PR so as not to overcomplicate it right now.

ferrariofilippo: > Worth using a string builder here.

I agree. And also, can you replace `String.IsNullOrWhiteSpace` with `string.IsNullOrWhiteSpace`?

src/Files.App/Filesystem/ListedItem.cs
```diff
@@ -49,6 +49,9 @@ public string ItemTooltipText
...
 					(SyncStatusUI.LoadSyncStatus
 						? $"{Environment.NewLine}{"syncStatusColumn/Header".GetLocalizedResource()}: {syncStatusUI.SyncStatusString}"
 						: string.Empty);
```
QuaintMako, line 57: This should do the trick and still working. Like so, no need to come back for a refactoring we'll never make 'cause it will be forgotten.

```suggestion
				var tooltipBuilder = new StringBuilder();
				tooltipBuilder.AppendLine($"{"ToolTipDescriptionName".GetLocalizedResource()} {Name}");
				tooltipBuilder.AppendLine($"{"ToolTipDescriptionType".GetLocalizedResource()} {itemType}");
				tooltipBuilder.Append($"{"ToolTipDescriptionDate".GetLocalizedResource()} {ItemDateModified}");
				if(!string.IsNullOrWhiteSpace(FileSize))
					tooltipBuilder.Append($"{Environment.NewLine}{"ToolTipDescriptionSize".GetLocalizedResource()} {FileSize}");
				if(SyncStatusUI.LoadSyncStatus)
					tooltipBuilder.Append($"{Environment.NewLine}{"syncStatusColumn/Header".GetLocalizedResource()}: {syncStatusUI.SyncStatusString}");

				return tooltipBuilder.ToString();
```

QuaintMako (commented):
Suggestion of fix.
Otherwise, LGTM and works on my computer.

src/Files.App/Filesystem/ListedItem.cs
QuaintMako: See new comment.

QuaintMako (approved):

ferrariofilippo (approved):
LGTM and works

PR #11110, opened 2023-01-29
Feature: Move shell extensions to sub menu by default

yaira2:
**About this change**
The goal of this change is to have a more predictable height for the context menu. Once we find a way to load extensions faster, we should look into reverting this change.

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.


QuaintMako:
> Once we find a way to load extensions faster, we should look into reverting this change.

Please be sure to create an issue to keep track of this.

yaira2:
@QuaintMako opened #11112

PR #11169, opened 2023-02-05
Fix: Fixed quick access widgets when there are few items

gave92:
**Resolved / Related Issues**

Fixes an issue where the recent folders widgets misbehaves when there are few pinned folder.
Current code works assuming the non-pinned items are the last 4 which is not always the case.
This PR looks instead at the "System.Home.IsPinned" property.

**Validation**
- [x] Built and ran the app


yaira2:
Fyi @hecksmosis

hecksmosis (approved):
LGTM, didn't know the property existed

PR #11185, opened 2023-02-06
Feature: Show confirmation dialog when the undo operation is to delete items.

### LINKED ISSUE #11174: Bug: Cut and paste into the source folder causes source folder to be deleted
### Description

I just tried to copy some files from desktop... First of all, for some reason it copied all files, even though I only selected 7 or 8 files... Issue number one. I then cut the files and copied them into another folder. But not all files could be copied into the folder because the folder had the same name.

So after cut & paste failure, I tried to do CTRL+Z to undo the cut & paste to get my files back to the desktop. Well, all my files are gone, irreversibly deleted...

### Steps To Reproduce

1. Cut target folder
2. Paste target folder inside itself
4. See error prompt
5. Use ctrl + z to undo
6. Notice that target folder was deleted

### Steps to fix
Skip cutting the source folder

### Files Version

Versi: 2.4.25.0

### Windows Version

OS Version: 10.0.22623.1250

### Log file

[debug.log](https://github.com/files-community/Files/files/10636714/debug.log)

**Is there any way to recover my files?**

hishitetsu:
**Resolved / Related Issues**
- Closes #11174

**Details**
A confirmation dialog will appear when the undo operation is to delete items to prevent unexpected deletion.

**TODO**
- [X] ~~The setting should be separate from "Show confirmation dialog when deleting items"~~. No new settings this time.
- [X] When cancelling the undo operation, ~~the history is erased even though it is not actually undone. Is that right?~~ keep the history.
- [X] On the confirmation dialog, we can choose to put it in the trash instead of deleting it. ~~Is that right?~~ It's ok.

**Validation**
- [X] Built and ran the app

### DIFF
--- a/src/Files.App/Filesystem/StorageHistory/Helpers/StorageHistoryHelpers.cs
+++ b/src/Files.App/Filesystem/StorageHistory/Helpers/StorageHistoryHelpers.cs
@@ -22,13 +22,17 @@ namespace Files.App.Filesystem.FilesystemHistory
 				{
 					return ReturnResult.InProgress;
 				}
+				bool keepHistory = false;
 				try
 				{
-					return await operations.Undo(App.HistoryWrapper.GetCurrentHistory());
+					ReturnResult result = await operations.Undo(App.HistoryWrapper.GetCurrentHistory());
+					keepHistory = result is ReturnResult.Cancelled;
+					return result;
 				}
 				finally
 				{
-					App.HistoryWrapper.DecreaseIndex();
+					if (!keepHistory)
+						App.HistoryWrapper.DecreaseIndex();
 					semaphore.Release();
 				}
 			}
--- a/src/Files.App/Filesystem/StorageHistory/StorageHistoryOperations.cs
+++ b/src/Files.App/Filesystem/StorageHistory/StorageHistoryOperations.cs
@@ -38,13 +38,13 @@ namespace Files.App.Filesystem.FilesystemHistory
 				case FileOperationType.CreateNew: // Opposite: Delete created items
 					if (!IsHistoryNull(history.Source))
 					{
-						return await helpers.DeleteItemsAsync(history.Source, false, true, false);
+						return await helpers.DeleteItemsAsync(history.Source, true, true, false); // Show a dialog to prevent unexpected deletion
 					}
 					break;
 				case FileOperationType.CreateLink: // Opposite: Delete created items
 					if (!IsHistoryNull(history.Destination))
 					{
-						return await helpers.DeleteItemsAsync(history.Destination, false, true, false);
+						return await helpers.DeleteItemsAsync(history.Destination, true, true, false); // Show a dialog to prevent unexpected deletion
 					}
 					break;
 				case FileOperationType.Rename: // Opposite: Restore original item names
@@ -61,7 +61,7 @@ namespace Files.App.Filesystem.FilesystemHistory
 				case FileOperationType.Copy: // Opposite: Delete copied items
 					if (!IsHistoryNull(history.Destination))
 					{
-						return await helpers.DeleteItemsAsync(history.Destination, false, true, false);
+						return await helpers.DeleteItemsAsync(history.Destination, true, true, false); // Show a dialog to prevent unexpected deletion
 					}
 					break;
 				case FileOperationType.Move: // Opposite: Move the items to original directory
--- a/src/Files.App/Filesystem/StorageHistory/StorageHistoryOperations.cs
+++ b/src/Files.App/Filesystem/StorageHistory/StorageHistoryOperations.cs
@@ -1,4 +1,6 @@
+using CommunityToolkit.Mvvm.DependencyInjection;
 using Files.App.Helpers;
+using Files.Backend.Services.Settings;
 using Files.Shared.Enums;
 using Files.Shared.Extensions;
 using System;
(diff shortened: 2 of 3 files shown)


yaira2:
I'm still a little confused as to why ctrl + z is deleting the folder, we should really match File Explorers behavior in that case and skip cutting the folder.

> On the confirmation dialog, we can choose to put it in the trash instead of deleting it. Is that right?

Can we show the regular confirm delete dialog? This already has the logic and indicates that it'll be permanently deleted.

> The setting should be separate from "Show confirmation dialog when deleting items".

Perhaps a new setting for "Show confirmation dialog when permanently deleting items?

hishitetsu:
> I'm still a little confused as to why ctrl + z is deleting the folder

In Files, creating a new folder like executing the following command.
MKDIR _AnyFoderName_
But in Explorer, like this.
MKDIR "New folder";RENAME _AnyFoderName_
So ctrl+z behaves delete in Files and rename back in Explorer.
If we don't rename the new folder in Explorer, ctrl+z behaves delete like Files.

> Can we show the regular confirm delete dialog? This already has the logic and indicates that it'll be permanently deleted.

Yes, this PR will do so.
But I am concerned that in this dialog I can choose to put it in the trash instead of permanently deleting it.

> Perhaps a new setting for "Show confirmation dialog when permanently deleting items".

I think we should a settings for "Show confirmation dialog when the undo operation is to delete items" as the title of this PR.
"Show confirmation dialog when permanently deleting items" might be a good idea, but I think there should be separate settings for normal delete and ctrl+z delete.

yaira2:
> In Files, creating a new folder like executing the following command.
MKDIR AnyFolderName
But in Explorer, like this.
MKDIR "New folder";RENAME "New folder" AnyFolderName
So ctrl+z behaves delete in Files and rename back in Explorer.
If we don't rename the new folder in Explorer, ctrl+z behaves delete like Files.

Would it be desirable to copy this behavior so that it's consistent?

> Yes, this PR will do so.
But I am concerned that in this dialog I can choose to put it in the trash instead of permanently deleting it.

Unchecking the "permanently" delete will send to trash, does this work?

> I think we should add a settings for "Show confirmation dialog when the undo operation is to delete items" as the title of this PR.
"Show confirmation dialog when permanently deleting items" might be a good idea, but I think there should be separate settings for normal delete and ctrl+z delete.

My concern is that it's confusing to explain what this does.

hishitetsu:
> Would it be desirable to copy this behavior so that it's consistent?

I don't think so. I think the folder rather should be deleted, not rename to "New folder" when undoing. Instead, it would be better to show a confirmation dialog if necessary.

> Unchecking the "permanently" delete will send to trash, does this work?

Yes.

> My concern is that it's confusing to explain what this does.

How about adding the following sub-settings?

Show confirmation dialog when deleting items - On or Off
 -> Also when the undo operation is to delete items - On or Off

yaira2:
> Show confirmation dialog when deleting items - On or Off
-> Also when the undo operation is to delete items - On or Off

I think it's still confusing, I think we should release the change without a setting and if users request an option to turn it off, we can reevaluate the behavior.

hishitetsu:
Ok, the rest is a matter of the operation history.
In this PR, the history is deleted regardless of whether or not the user canceled. If the user cancels, should the history remain?
If the history remains, it may be more consistent but the user can't undo further past operations.

yaira2:
Which one is safer?

hishitetsu:
I committed a new code to keep the history. So I think it is ready to merge.

hishitetsu:
Sorry, this code may not be good. I will try to fix it immediately.

hishitetsu:
This time it's okay.

src/Files.App/Filesystem/StorageHistory/Helpers/StorageHistoryHelpers.cs
```diff
@@ -22,13 +22,17 @@ public async Task<ReturnResult> TryUndo()
...
-					return await operations.Undo(App.HistoryWrapper.GetCurrentHistory());
+					ReturnResult result = await operations.Undo(App.HistoryWrapper.GetCurrentHistory());
+					keepHistory = result is ReturnResult.Cancelled;
+					return result;
 				}
 				finally
 				{
-					App.HistoryWrapper.DecreaseIndex();
+					if (!keepHistory)
+						App.HistoryWrapper.DecreaseIndex();
```
yaira2, line 35:
```suggestion
						App.HistoryWrapper.DecreaseIndex();

```

yaira2:
It looks like the issue with ctrl + z was fixed but I'm not seeing a confirmation prompt

hishitetsu:
Really? When creating a folder and performing ctrl+z, I see the dialog.
image

yaira2:
That's right, setting to show confirmation dialog is turned on as well but that shouldn't matter.

yaira2:
I rebuilt the branch and it's working now, LGTM

src/Files.App/Filesystem/StorageHistory/StorageHistoryOperations.cs
```diff
@@ -38,13 +42,13 @@ public async Task<ReturnResult> Undo(IStorageHistory history)
 				case FileOperationType.CreateNew: // Opposite: Delete created items
 					if (!IsHistoryNull(history.Source))
 					{
-						return await helpers.DeleteItemsAsync(history.Source, false, true, false);
+						return await helpers.DeleteItemsAsync(history.Source, UserSettingsService.FoldersSettingsService.ShowConfirmDeleteDialog, true, false);
```
yaira2, line 45: When I said we should try without a separate setting, I meant that we should show the confirmation regardless of the setting since it's not immediately clear that it will delete items. If user feedback comes in that they still want to hide the prompt in this case, we can add a separate setting and let them choose.

hishitetsu: I understand. I will do so.

hishitetsu: Now fixed.

yaira2 (approved):
LGTM

PR #11197, opened 2023-02-07
Fix: Additional fix to show a confirmation dialog when undoing

hishitetsu:
**Resolved / Related Issues**
- This is an additional fix for #11185 

**Details**
Unfortunately, #11185 did not show a confirmation dialog on undo when the setting to show the dialog is off because DeleteItemAsync method always confirms the setting internally.
In this PR, the method parameter is changed to be able to show the dialog ignoring the setting.

**Validation**
- [X] Built and ran the app


hishitetsu:
Sorry, I will rewrite the code to be simpler.

Issue #3722, opened 2021-02-25
Feature: Add hashes tab to the properties window


PMExtra:
### Describe the solution
[HashTab](http://implbits.com/products/hashtab/) is a plugin of Windows Explorer.

![image](https://user-images.githubusercontent.com/11289158/109131328-a78c0e80-778d-11eb-9186-78b11bcd1ff6.png)

I want the same feature that build in Files.

### Requirements
- Add hashes tab to the properties windows
- Display hashes (starting with CRC32, MD5 and SHA-1) in the list (should follow the design pattern in the details tab)

### Files version

2.3.12.0

### Windows version

10.0.19043.2006

PMExtra:
@yair100 Please let me know if you'd like to implement it or welcome the PR.

yaichenbaum:
@PMExtra A PR would be great, would you like to work on this?

Josh-65:
Thought to add since 7zip is now integrated the hash features might be in it too. The 7zip program support these.
![image](https://user-images.githubusercontent.com/26977075/185786097-078fdbde-bfda-4c50-9282-95078797a127.png)

Josh65-2201:
Asking @yair100 for approval.

yaichenbaum:
What kinds of settings are available in HashTab?

sthivaios:
> What kinds of settings are available in HashTab?

I just installed HashTab on a VM.

It has the following features :
- A link to license the software, that seems like it doesn't work anymore
- A settings link, which opens the following window : https://user-images.githubusercontent.com/109022579/195403031-21ea85ba-4dba-4685-aee0-5b45d8b4de4c.png
- A "Hash Comparison" text box, and a "Compare a file..." button
- The "main" thing that shows the hash values.


That is a screenshot of the full window : 

![image](https://user-images.githubusercontent.com/109022579/195403307-10ea5b02-e907-4a30-9c90-d178ed324c13.png)



PS : You might have meant a completely different think by saying "What kind of settings are available in HashTab?", but, that is what I understood... I don't have any idea how to code so I cannot get involved anymore in this thread.

yaichenbaum:
I think we can start with the most common hashes and track requests for additional ones later one. That being said, we need a design before starting on this.

PR #11214, opened 2023-02-08
Perf: Small performance improvement for CalculateChecksumForPath

vcsjones:
This is a small performance improvement for `CalculateChecksumForPath`:

1. MD5 hashes are stack-buffer size friendly, so switch to the buffer-writing `MD5.HashData` and write to the stack.
2. Replace `BitConverter.ToString(..).Replace(..)` with `Convert.ToHexString`. This eliminates a 32 character string allocation.

Note that this method could be further optimized to make it near allocation-less at the cost of readability, adding complexity. and using unsafe code. But it isn't clear to me how "hot" this code path is. Right now the allocations scale with the length of the path. However, these changes are small and don't harm the readability of the method.

Benchmarks:

| Method | PathLength |       Mean |   Error |  StdDev |   Gen0 | Allocated |
|------- |----------- |-----------:|--------:|--------:|-------:|----------:|
|  After |        260 |   815.4 ns | 5.43 ns | 5.07 ns | 0.0591 |     376 B |
| Before |        260 | 1,054.6 ns | 2.77 ns | 2.59 ns | 0.0839 |     536 B |

**Validation**
- [X] Built and ran the app

**Screenshots (optional)**
Not really applicable.

### DIFF
--- a/src/Files.Shared/Helpers/ChecksumHelpers.cs
+++ b/src/Files.Shared/Helpers/ChecksumHelpers.cs
@@ -9,9 +9,9 @@ namespace Files.Shared.Helpers
 		public static string CalculateChecksumForPath(string path)
 		{
 			var buffer = Encoding.UTF8.GetBytes(path);
-			var hash = MD5.HashData(buffer);
-
-			return BitConverter.ToString(hash).Replace("-", string.Empty);
+			Span<byte> hash = stackalloc byte[MD5.HashSizeInBytes];
+			MD5.HashData(buffer, hash);
+			return Convert.ToHexString(hash);
 		}
 	}
 }


yaira2:
For what it's worth, I don't think we're actually using this right now but it's a good reminder for us to implement #3722

0x5bfa:
@yair100 To use stackalloc will be helpful if they are going to calculate considerably long binary data like file content. In this case, they calculate path.

hishitetsu (approved):
LGTM, but actually, this code does not appear to be used at all.

> it's a good reminder for us to implement #3722 

Yes. In fact, thanks to @vcsjones, we were able to implement #3722! I appreciate it very much.
https://github.com/dotnet/runtime/pull/63757

yaira2:
@vcsjones thank you!

PR #11237, opened 2023-02-10
Fix: Fixed selecting wrong items when creating or renaming an item

### LINKED ISSUE #11092: Bug: Creating, renaming, or deleting an item causes Files to select additional items
### Description

When creating a new folder and giving it a name (and using the preview version of Files), it instead makes a folder called "New Folder", and then selects other items.

https://user-images.githubusercontent.com/28498877/214926209-4850e2c7-5f17-4221-8ff5-8246e65a0c27.mp4

This issue is more apparent in bigger directories, as there's more files and folders to select.

### Steps To Reproduce

1. Use the preview version
2. Make a new folder
3. Try to name it
4. Confirm

### Steps to fix

- Actually name the folder
- Don't select other folders / items after confirmation (maybe it's tied to the feature where you press A and you select the first file / folder that starts with A)

### Files Version

2.4.24.0

### Windows Version

Windows 11 22H2 22621.1105

### Log file

[debug.log](https://github.com/files-community/Files/files/10512569/debug.log)

hishitetsu:
**Resolved / Related Issues**
- Closes #11092

**Details**
`ItemContainer.Content` appears to refer to the old item before the item was created or renamed.
Because of this, checked and unchecked events select wrong items.
These events are for (un)selecting (un)checked items and are not needed in this logic, so I disabled these events temporarily.

Deleteing items also selects another item, but this appears to be the intended specification so it is left as is at this time.
(ItemViewModel.cs:1850)

**Validation**
- [X] Built and ran the app

### DIFF
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -714,7 +714,16 @@ namespace Files.App.Views.LayoutModes
 			{
 				var checkbox = container.FindDescendant("SelectionCheckbox") as CheckBox;
 				if (checkbox is not null)
+				{
+					// Temporarily disable events to avoid selecting wrong items
+					checkbox.Checked -= ItemSelected_Checked;
+					checkbox.Unchecked -= ItemSelected_Unchecked;
+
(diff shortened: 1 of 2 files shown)


yaira2 (approved):
LGTM

PR #11249, opened 2023-02-11
Feature: Add a UI to edit, delete and create tags

### LINKED ISSUE #9588: Feature: Add a UI to edit, delete and create tags
This issue is open for discussion.
image

- Should the UI for creating tag have its own dialog or should it be part of settings?
- Should you be able to right click the tags section on the sidebar to create a new tag or should you have to open the manage tags UI and do it from there?

**Requirements**
- Add a list under experimental settings for adding/deleting tags
- There should be a way to edit existing tags as well

ferrariofilippo:
**Resolved / Related Issues**
- Closes #9588

**Supported Features**
- Add a new tag (name = "New Tag" and random color)
- Rename an existing tag
- Delete an existing tag
- Change color of an existing tag (Color Picker as for app bg)
- Sidebar section, Tags Widget and right click context menu are all synced

**Notes**
- `CommandBar.MinWidth=250` is needed to avoid a UI bug involving localization
- `ColorHelpers.FromHex()` now works even when alpha is specified
- I created a method to remove a `GoTo` in `AdvancedViewModel`
- `FileTagsSettingsService.GetTagAndIndex()` is basically an `IndexOf` which does not work (always returns -1)
- I'm not sure why (I'll check this out), but to edit `FileTagList` you need to reassign it

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
image


yaira2:
This is amazing! Could of suggestions for improving the UX,

- Instead of a circle, maybe we can use this same design we do in the in appearance settings. This designs makes it a little easier to tell that you can click to change the color.
image

- For renaming tags, perhaps there can be an edit button at the top of the list alongside add and remove.

src/Files.App/ServicesImplementation/Settings/FileTagsSettingsService.cs
```diff
@@ -72,6 +74,47 @@ public IEnumerable<TagViewModel> SearchTagsByName(string tagName)
 			return FileTagList.Where(x => x.Name.StartsWith(tagName, StringComparison.OrdinalIgnoreCase));
 		}
 
+		public void CreateNewTag()
+		{
+			var newTag = new TagViewModel(
+				"NewTag",
+				"#9EA3A1",
```
yaira2, line 81: Can we generate a random color?

42-star:
Yes please

hecksmosis (dismissed):

ferrariofilippo:
Actually, there seems to be a bug (Tags color in contextmenu are different from sidebar)

yaira2:
I'm planning to test this myself but can you also update the screenshot and supported features list in the description?

yaira2:
> Rename an existing tag (Edit button or F2 or Double Click)

Can we simplify this to the edit button? If it's more intuitive, the edit button can be added to the ListItem template. Clicking the edit button will enable rename mode and display a check button to complete the rename.

I don't think we need to support F2 and Double Click here.

yaira2:
We can add a delete button to the ListItem template as well.

ferrariofilippo:
> We can add a delete button to the ListItem template as well.

Done

yaira2:
Can the list view be made a little wider?

src/Files.App/Views/SettingsPages/Advanced.xaml.cs
```diff
@@ -1,12 +1,139 @@
...
+		private string oldTagName = string.Empty;
+
+		private TagViewModel? renamingTag;
+
 		public Advanced()
 		{
 			InitializeComponent();
 		}
+
+		private void RenameTextBox_KeyDown(object sender, KeyRoutedEventArgs e)
```
yaira2, line 22: What's this for? Is it still needed with the button to commit the rename?

yaira2:
@ferrariofilippo I made a couple of tweaks to the UI, one thing that can be improved is instead of changing the visibility of elements from code behind, we can use x:Load. This makes it easier to understand how the xaml will behave in different scenarios.

src/Files.App/Views/SettingsPages/Advanced.xaml.cs
ferrariofilippo: I can remove `Esc` and `Enter`. Still, I think we should handle `arrows`. If we don't, when the user is at the beginning/end of the textbox and he presses up/down, other items will be selected

ferrariofilippo:
> one thing that can be improved is instead of changing the visibility of elements from code behind, we can use x:Load. 

I binded `Visibility` instead of `x:Load` to simplify things

src/Files.App/Views/SettingsPages/Advanced.xaml.cs
yaira2: I'd prefer it if we stuck with the edit and confirm button for this. We can keep enter though.

ferrariofilippo: > I'd prefer it if we stuck with the edit and confirm button for this.

That is what I meant, we have to handle `Up/Down` arrows to avoid selecting other items.

> We can keep enter though.

I'll reintroduce that

yaira2:
Can you use commands instead of the click event?

yaira2:
I made some UI improvements, can you change the behavior so the color only takes affect when you save changes?

src/Files.App/Views/SettingsPages/Advanced.xaml.cs
yaira2: I implemented this 👍

yaira2:
I improved the way new tags are added but it still needs some work.
- Choosing the color when creating a new tag isn't supported yet
- The save changes button should only be enabled if there is a valid name in the text box

ferrariofilippo:
> - can you change the behavior so the color only takes affect when you save changes?
> - Choosing the color when creating a new tag isn't supported yet
> - The save changes button should only be enabled if there is a valid name in the text box

Done

ferrariofilippo:
> Can you use commands instead of the click event?

To do that we should bind to `ListedTagViewModel` and then link commands each time we create an instance of it. I don't think this would be more readable. Anyway, if you believe we should stick to that, I'll refactor the code.

ferrariofilippo:
Just found a bug: app crashes if you click `Remove` twice in a row

ferrariofilippo:
> This results in the `Set` method, which stores the values in the JSON file, not being called.

That's the case, thank you. I didn't check how `Get` worked and I thought we had a cached instance

ferrariofilippo:
> > Can you use commands instead of the click event?
> 
> To do that we should bind to `ListedTagViewModel` and then link commands each time we create an instance of it. I don't think this would be more readable. Anyway, if you believe we should stick to that, I'll refactor the code.

@yair100 ?

yaira2:
Let's see what @d2dyno1 wants

src/Files.App/ServicesImplementation/Settings/FileTagsSettingsService.cs
```diff
@@ -73,6 +76,44 @@ public IEnumerable<TagViewModel> SearchTagsByName(string tagName)
...
+		public void CreateNewTag(string newTagName, string color)
+		{
+			var newTag = new TagViewModel(
+				newTagName,
+				color,
+				Guid.NewGuid().ToString());
+
+			var oldTags = FileTagList.ToList();
+			oldTags.Add(newTag);
+			FileTagList = oldTags;
```
yaira2, line 88: Can we add directly to FileTagsList?

src/Files.App/ServicesImplementation/Settings/FileTagsSettingsService.cs
```diff
@@ -73,6 +76,44 @@ public IEnumerable<TagViewModel> SearchTagsByName(string tagName)
...
+
+		public void DeleteTag(string uid)
+		{
+			var (_, index) = GetTagAndIndex(uid);
+			if (index == -1)
+				return;
+
+			var oldTags = FileTagList.ToList();
+			oldTags.RemoveAt(index);
+			FileTagList = oldTags;
```
yaira2, line 114: Same thing here, why are we replacing the list?

src/Files.App/ServicesImplementation/Settings/FileTagsSettingsService.cs
ferrariofilippo: > I think RemoveAt only modifies the current object, but does not set the property (does this make sense?). This results in the Set method, which stores the values in the JSON file, not being called.

yaira2:
> The save changes button should only be enabled if there is a valid name in the text box

This is still a issue

ferrariofilippo:
> This is still a issue

Which are invalid names? I filtered empty names and names starting/ending with a `.`

yaira2:
Empty names

yaira2:
Did you apply this to the new tag section?

ferrariofilippo:
> Did you apply this to the new tag section?

Yes, but I'll check

yaira2:
It's working now, maybe I was on an old commit

yaira2:
@ferrariofilippo one last thing, if you switch to a different tag without saving the previous edits, can you reset the preview color on the left?

ferrariofilippo:
My bad I removed a needed line with one of the latest commit

ferrariofilippo:
Wait there's still a bug

ferrariofilippo:
Everything should be fine now

yaira2:
@ferrariofilippo thank you for your patience as I figured out the UI, I wasn't really sure what I wanted it to look like going into this but I'm happy with how it turned out.

yair100 (dismissed):
LGTM

ferrariofilippo:
> @ferrariofilippo thank you for your patience as I figured out the UI, I wasn't really sure what I wanted it to look like going into this but I'm happy with how it turned out.

There's no problem, I love tags and I'd do anything to make them more powerful

src/Files.App/ViewModels/SettingsViewModels/AdvancedViewModel.cs
```diff
@@ -299,5 +317,72 @@ private FileOpenPicker InitializeWithWindow(FileOpenPicker obj)
...
+		{
+			get => isNameValid;
+			set => SetProperty(ref isNameValid, value);
+		}
+
+		public void Reset()
+		{
+			Name = string.Empty;
+			Color = ColorHelpers.RandomColor();
+		}
```
QuaintMako, line 386: Worth moving into its own file to not clutter the class.

src/Files.App/Views/SettingsPages/Advanced.xaml.cs
```diff
@@ -1,12 +1,104 @@
...
+		private void RemoveTag_Click(object sender, RoutedEventArgs e)
+		{
+			ViewModel.DeleteExistingTag((ListedTagViewModel)((Button)sender).DataContext);
+		}
+
+		private void NewTagTextBox_TextChanged(object sender, TextChangedEventArgs e)
+		{
+			var text = ((TextBox)sender).Text;
+			ViewModel.NewTag.Name = text;
+			ViewModel.NewTag.IsNameValid = !(string.IsNullOrWhiteSpace(text) || text.EndsWith('.') || text.StartsWith('.'));
```
QuaintMako, line 101: This name validation is repeated twice, worth making it a `private bool` method.

QuaintMako (commented):
A few minor points. I haven't been able to test it yet. I'll add the approval once I can.

src/Files.App/ViewModels/SettingsViewModels/AdvancedViewModel.cs
yaira2: Where should the file go?

ferrariofilippo: Maybe in `DataModels` or `Backed.ViewModels.FileTags`?

QuaintMako: I would have suggested still in `ViewModels`, but in an entire new class. But after looking at the code from within the Visual Studio, I see the class is only used insside the `AvancedViewModel`, so if we're sure it will never be used elsewhere, it may stay there.
May I ask what is the use of the `Backed` namespace?

ferrariofilippo: I forgot an `N` it should have been `BackeNd.ViewModels.FileTags`, which is where other Tags view models are

QuaintMako:
A few ideas while testing the UI I can think of, to be validated by @yair100 :
- Having a option when right clicking on the `Tags` item on the sidebar to open directly to the setting to edit the tags would be nice.
- When editing a tag, pressing `Esc` should stop the edition, not close the setting panel.
- Having to travel to and from the action buttons may be cumbersome if a lot of manipulation has do be done. Maybe getting them closer would be nicer?

Screenshot_1

- I can create two tags with the same name. Sure, the guid are differents, but is this a wanted behavior?
- Deleting a tag should display a dialog box to ensure it's not an error.
- Once it's no longer an experimental feature, maybe move it to an entirely new setting tab?
- Displaying some kind of warning when the name is not usable, a red box around the name perhaps ? Also, explain somehow why it's not valid (tooltip?). Otherwise, the user won't be able to know what's wrong.

Screenshot_2

Otherwise @ferrariofilippo very smooth code and nice UI! Thanks for taking care of this, it's gonna be a **huge** addition when it comes live!

src/Files.App/ViewModels/SettingsViewModels/AdvancedViewModel.cs
QuaintMako: In that case, that might be a good idea!

yaira2:
> Having a option when right clicking on the Tags item on the sidebar to open directly to the setting to edit the tags would be nice.

We don't have any deep linking like this in the app right now so best to leave as is but can revisit later on.

> When editing a tag, pressing Esc should stop the edition, not close the setting panel.

This can be added

> Having to travel to and from the action buttons may be cumbersome if a lot of manipulation has do be done. Maybe getting them closer would be nicer?

Unlike editing files, I don't think tags are something that are going to be modified a lot, nevertheless, it wouldn't hurt to center them.

yaira2:
> I can create two tags with the same name. Sure, the guid are differents, but is this a wanted behavior?

No, should probably be added to the validation.

> Deleting a tag should display a dialog box to ensure it's not an error.

A flyout can do the trick, can you open an issue to track?

> Once it's no longer an experimental feature, maybe move it to an entirely new setting tab?

That's the plan

> Displaying some kind of warning when the name is not usable, a red box around the name perhaps ? Also, explain somehow why it's not valid (tooltip?). Otherwise, the user won't be able to know what's wrong.

Good idea, can you open an issue to track?

yaira2:
@QuaintMako can you open some issues to track these? I'd like to merge this as is so that we can move forward. Most of those are quick changes and shouldn't take long to implement.

QuaintMako:
Alright, I'll do so soon.

QuaintMako (approved):

yaira2:
Feel free to skip the description at the top of the issues since I already know what they are referring to. Just make sure to include the requirements.

PR #11243, opened 2023-02-11
Feature: Disable show more options menu item until shell items are loaded

hecksmosis:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #11236

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [x] Tested the changes for accessibility

**Screenshots (optional)**
Add screenshots here.

PR #11270, opened 2023-02-13
Fix: Fixed issues with recent context menu changes

hishitetsu:
**Resolved / Related Issues**
This is a PR of some fixes for recent context menu changes. (#11243 etc.)

**Details**
* Fixed an issue that could cause a hang when right-clicking on a widget
* Fixed an issue in which "Loading more options" did not disappear in some cases
* Some tweaks to the overflow separator

**Validation**
- [X] Built and ran the app


yaira2 (approved):
Tested and works as expected, LGTM

Issue #6902, opened 2021-11-13
Cannot continue with the folder tree in columns layout after relaunching Files


XuShenLZ:
### Description

I always admire the "columns layout" of the file explorer in Mac system. Now I can finally use it with Files! When seeing this feature in Files v2, I immediately switched to Files as my default explorer as a **paid Directory Opus user for years**!

However, I found that with columns layout, you cannot continue with the entire folder tree after relaunching Files. 

In "Setting -> On Startup -> When launching Files", I choose "Continue where you left off".

![image](https://user-images.githubusercontent.com/31999415/141657537-c902beae-bf6a-44e5-add8-4d1ab2fcb138.png)

Then I navigate a few levels down through my disk with "Columns Layout". In the example below, I start the tab with "Downloads", and navigate to "Downloads/DiskGenius/docs". But the title of this tab is still "Downloads".

![image](https://user-images.githubusercontent.com/31999415/141657579-d522a08c-8882-41c4-9998-ee66a89cb101.png)

So when I exit Files and relaunch it, it only start with the "Downloads" and I have to navigate to all subfolders again.

![image](https://user-images.githubusercontent.com/31999415/141657649-80dbf7f6-6b34-405f-9383-b4561df239ac.png)

### Steps To Reproduce

1. Go to "Setting -> On Startup -> When launching Files"
2. Select "Continue where you left off"
3. Open a new tab and start with a "root" location
4. Select "Columns" Layout
5. Navigate down a few subfolders
6. Close Files
7. Relaunch Files
8. I can only get the "root" location back, not the last subfolder I have visited

### Expected behavior

Files should remember the expanded folder tree in columns layout. After exiting Files and relaunch, I should have the entire folder tree back, rather than the root folder only.

### Files Version

2.0.34.0

### Windows Version

Windows 11 Pro 21H2 22000.318

sstavridis:
Is there any way to apply the columns layout to be the default view for all the folders?
For example in File Explorer you can go to view and select Large Icons then you can to Options and on the View tab you check Apply to folders.

Josh-65:
Is this still an issue?

XuShenLZ:
Yes. Everytime I close and restart Files, the folder tree in column layout will root from the latest folder. The entire tree is not preserved

PR #11274, opened 2023-02-13
Feature: Improve navigations in Column View

### LINKED ISSUE #11128: Feature: Improve backwards navigation in Column View

I like the Column View, but it is frustrating that when I move to another folder, the Back button does not return to the directory where I previously was, but to the root of the Column View.

Steps to reproduce:
1. Set the view to Column View.
2. Use the left pane to navigate to C:\ from the Drives list.
3. Navigate to C:\Program Files (x86)\Microsoft\Edge (or whatever else).
4. Use the left pane to navigate to the Downloads folder from the Favorites list.
5. Press the Back button.
6. It returns to the C:\ folder, rather than C:\Program Files (x86)\Microsoft\Edge.

### Requirements

- All opened directories in the Column View should be kept so that after clicking the Back button, the navigation should return to the deepest directory visited before.

### Files Version

2.4.25

### Windows Version

Windows 11 Pro 22H2 22621.1105

### Comments

_No response_

hishitetsu:
**Resolved / Related Issues**
- Closes #11128 

**Details**
All opened directories in the Column View are now kept after clicking the Back or Forward button.
However, it is hard to keep focus, so the root directory is focused after clicking the button.

**Validation**
- [X] Built and ran the app

### DIFF
--- a/src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
@@ -70,7 +70,7 @@ namespace Files.App.Views.LayoutModes
 					NavPathParam = column.NavPathParam
 				});
 				navigationArguments.Column = ColumnHost.ActiveBlades.IndexOf(newblade);
-				navigationArguments.NavPathParam = column.NavPathParam;
+				navigationArguments.ColumnPathParam = column.NavPathParam;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


yaira2:
Does this help for #6902?

hishitetsu:
I tried, but #6902 has not been improved.

hishitetsu:
I have converted this PR to draft as it seems to make #6902 more unpredictable.

hishitetsu:
Ok, #6902 has not been improved yet, but has returned to the same behavior as before. It looks like another, more complicated fix is needed to fix #6902, so I have returned it to ready for review status.

ferrariofilippo:
> Ok, #6902 has not been improved yet, but has returned to the same behavior as before. It looks like another, more complicated fix is needed to fix #6902, so I have returned it to ready for review status.

#9769 code might be useful to fix #6902

hishitetsu:
@ferrariofilippo Thank you, but the problem with #6902 is that the path of each column has to be saved on exit as tab information. This seems more difficult than with the back/forward buttons.

ferrariofilippo:
> @ferrariofilippo Thank you, but the problem with #6902 is that the path of each column has to be saved on exit as tab information. This seems more difficult than with the back/forward buttons.

That code used to open all the baldes up to the root folder, we might save the first and the last one and then use that algo to achieve our goal

hishitetsu:
Right now we can only store one path as tab information, but if we can extend that, we can certainly use this code.
Anyway, #6902 is a similar but different problem to this one, so I think this PR is ok.

yaira2:
In addition to the focus, the opened directories aren't selected

hishitetsu:
Sorry, is "selected" different than "focused"?

hishitetsu:
Ah, I understand. I guess I need to fix it a bit more.

hishitetsu:
I close this PR as it should be addressed along with #7589.

PR #11192, opened 2023-02-06
Feature: Added shortcut to open folders in terminal

ferrariofilippo:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes `TODO` task

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [ ] Tested the changes for accessibility

PR #11281, opened 2023-02-13
Feature: Added shortcut to open Windows Terminal as administrator 

### LINKED ISSUE #11278: Feature: Open Windows Terminal as admin using ctrl + shift + `

On Windows 10, on MS File Explorer (MSFE) you can open a shell terminal by using the combo ALT->F->S->(A|R) A for Admin or R for Regular User.  
This is **extremely** handy as one can open a shell as admin on the path we want to work with with ease, especially for software installation etc.  

image

On Windows 11, the brilliant minds of Microsoft removed any of the ALT keys from MSFE, because reasons.  

I've started using Files, I like the way it works, and having this feature would be great. 

### Requirements

- This proposal will require the addition of a hot-key or sequence of keys that would open a (power)shell terminal on the current location as administrator. 

### Files Version

2.4.33.0

### Windows Version

Win 10 - 10.0.19044.2486

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**
- Closes #11278 

**Notes**
- Fixes a bug with #11192 which would not open the path if the default terminal was PowerShell

**Validation**
- [x] Built and ran the app


yaira2 (approved):
LGTM

yaira2:
Instead of requiring Files to be admin, can display a UAC prompt?

ferrariofilippo:
There's a bug with latest commit, I'll try to figure out why

yaira2 (approved):
LGTM
Can you add to the documentation?

src/Files.App/Views/ColumnShellPage.xaml.cs
```diff
@@ -230,7 +230,7 @@ private async void ColumnShellPage_PreviewKeyDown(object sender, KeyRoutedEventA
 
 			switch (c: ctrl, s: shift, a: alt, t: tabInstance, k: args.Key)
 			{
-				case (true, false, false, true, (VirtualKey)192): // ctrl + ` (accent key), open terminal
+				case (true, _, false, true, (VirtualKey)192): // ctrl + ` (accent key), open terminal
```
francisco-lafe, line 233: Shouldn't it say "true" instead of "_"?

francisco-lafe: I think the way this should be done is with another case where Shift is 'true' and then there pass the "-Verb runas" as an argument.

ferrariofilippo: I don't think so, it would only create duplicate code

francisco-lafe: Alright, I don't know if a following commit trampled over these changes, but the feature for "CTRL+SHIFT+`" is not working. 
#11443

ferrariofilippo: I'll check

francisco-lafe: > I'll check

I did, but unsure if the feature was moved elsewhere. On commit:
https://github.com/files-community/Files/commit/dc676d30a6e01836683e9bde4bed88ac92f65be5

ferrariofilippo: Moved elsewhere but still it should work. I'll see I can find what is causing the problem

ferrariofilippo: Can you confirm that it was working before that commit?

ferrariofilippo: Ok I found the root and I feel so stupid ...

francisco-lafe: > Ok I found the root and I feel so stupid ...

You're smart enough to have found it! 
Great work 😄

ferrariofilippo: I wouldn't have without your help

PR #11304, opened 2023-02-14
Feature: Move open with to main menu when right clicking recent files

### LINKED ISSUE #11298: Feature: Move "open with" to main menu when right clicking recent files

The recent files widget has an option to "open with" but it's in the overflow menu. Moving it to the main menu would improve the user experience and consistency between the layouts.

### Requirements

- This proposal will display "open with" in the main menu when right clicking in the recent files widget

### Files Version

v2.4.34

### Windows Version

Windows 11

### Comments

_No response_

hecksmosis:
**Resolved / Related Issues**
- Closes #11298

**Validation**
- [x] Built and ran the app


yaira2:
Is this working for you? I don't see the option and it's crashing as well.

hecksmosis:
This works fine for me

hecksmosis:
One of the few times CI doesn't spontaneously combust

yaira2:
It's still not working for me, maybe someone else can test

hecksmosis:
I cannot reproduce this, don't know why this is happening

hecksmosis:
Does this happen in main?

yaira2:
`main` is fine

yaira2:
@hishitetsu does this work for you?

yaira2:
@hecksmosis in the meantime, can you share a screen recording of this in action?

hishitetsu:
Sorry, I'm away from home and can't debug right now, but `showOpenWithMenu: true` was not specified in the other widgets and was causing the shell extension not to load, so I removed this option.

hecksmosis:
Oh, that's why then

hecksmosis:
@hishitetsu I looked into this and it seems to work but only if you don't await it

hecksmosis:
I'll send a screen recording

hecksmosis:
[f1fcd40d-7694-4ae8-b9c3-5314700cc655.webm](https://user-images.githubusercontent.com/55754091/219298289-5a824c84-5f1b-4d53-9ea1-3e1c2356fb9d.webm)

hishitetsu:
I have identified the cause of the error.
Once #11325 is merged, the error should no longer occur.

yaira2:
It's working now but it looks like it's missing the placeholder that's supposed to show when the sub menu is loading. Also, there seems to be an issue in menu flyouts where the colored icons don't work so we might have to use a regular icon for now.

hecksmosis:
> It's working now but it looks like it's missing the placeholder that's supposed to show when the sub menu is loading. Also, there seems to be an issue in menu flyouts where the colored icons don't work so we might have to use a regular icon for now.

@yair100 Do you mean a placeholder for the open with menu so it doesn't change size?

yaira2:
Yes

hecksmosis:
Ok, on it

hecksmosis:
What icon should we use?

yaira2:
`\uE17D`

hecksmosis:
Does this have a color part or should I use \uF047 and \uF048? (These are the ones used in BaseLayout)

hecksmosis:
Done

src/Files.App/Helpers/ShellContextMenuHelper.cs
```diff
@@ -255,7 +255,15 @@ public static async Task LoadShellMenuItems(string path, CommandBarFlyout itemCo
 					var openWithItem = shellMenuItems.Where(x => (x.Tag as Win32ContextMenuItem)?.CommandString == "openas").ToList().FirstOrDefault();
 					if (openWithItem is not null)
 					{
+						openWithItem.ColoredIcon = new ColoredIconModel()
+						{
+							BaseLayerGlyph = "\uF049",
+							OverlayLayerGlyph = "\uF04A",
+						};
```
yaira2, line 262: Oh, color icon is working here?

yair100 (approved):
LGTM

src/Files.App/Helpers/ShellContextMenuHelper.cs
hecksmosis: Found out about this and decided to try it

hecksmosis: It's added in the MenuFlyoutItem before it's converted into an AppBarButton

yaira2: Can we do the same for the "new" flyout? We have a colored icon for the shortcut option but it wasn't working to use the icon there.

PR #11310, opened 2023-02-15
Fix: Fixed closing app when the item flyout is open to hang

### LINKED ISSUE #10110: Bug: Closing app when menu flyouts are open causes hang
### Description

Closing the app when the right click content menu is still open causes the application to hang for a couple of seconds before it closes.

### Steps To Reproduce

1. Right click on a file or folder
2. Click close button on the title bar
3. Observe that the application hangs for a couple of seconds before closing

### Steps to fix

- Fix hang that occurs when opening the context menu before closing the app

### Files Version

v.2.3.39

### Windows Version

Windows 11

### Relevant Assets/Logs

No crash is detected in the log files

hishitetsu:
**Resolved / Related Issues**
- Closes #10110

**Details**
For some reason, the application does not close properly when a CommandBarFlyout is open, so the open flyout is forced to close before closing app, and then the application is closed as the event after closing the flyout.

**Validation**
- [X] Built and ran the app

### DIFF
--- a/src/Files.App/UserControls/Widgets/FileTagsWidget.xaml.cs
+++ b/src/Files.App/UserControls/Widgets/FileTagsWidget.xaml.cs
@@ -65,6 +65,7 @@ namespace Files.App.UserControls.Widgets
 		private void Item_RightTapped(object sender, RightTappedRoutedEventArgs e)
 		{
 			var itemContextMenuFlyout = new CommandBarFlyout { Placement = FlyoutPlacementMode.Full };
+			itemContextMenuFlyout.Opening += (sender, e) => App.LastOpenedFlyout = sender as CommandBarFlyout;
 			if (sender is not StackPanel tagsItemsStackPanel || tagsItemsStackPanel.DataContext is not FileTagsItemViewModel item)
 				return;
 			itemContextMenuFlyout.ShowAt(tagsItemsStackPanel, new FlyoutShowOptions { Position = e.GetPosition(tagsItemsStackPanel) });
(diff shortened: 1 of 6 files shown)


src/Files.App/UserControls/Widgets/FileTagsWidget.xaml.cs
```diff
@@ -65,6 +65,7 @@ private async void FileTagItem_ItemClick(object sender, ItemClickEventArgs e)
 		private void Item_RightTapped(object sender, RightTappedRoutedEventArgs e)
 		{
 			var itemContextMenuFlyout = new CommandBarFlyout { Placement = FlyoutPlacementMode.Full };
+			itemContextMenuFlyout.Opening += (sender, e) => App.LastOpenedFlyout = sender as CommandBarFlyout;
```
yaira2, line 68: Should we unsubscribe when it closes?

hishitetsu: I think there is no need to do so because a new instance of the flyout is created each time and is not reused.

hishitetsu: I have verified that the flyout works after showing flyouts several times on the home screen and also that the application exits without issue. Should I still check something?

yaira2 (approved):
LGTM

src/Files.App/UserControls/Widgets/FileTagsWidget.xaml.cs
hishitetsu: Memory leaks from not subscribing would occur if the publisher continues to remain. But since the publisher is disposed every time, memory leaks should not occur. If a memory leak were to occur, it would happen without subscribing to the event.

PR #11327, opened 2023-02-16
RichCommand: Base and commands ShowFileExtensions/ShowHiddenItems

cinqmilleans:
**Resolved / Related Issues**
This pr adds the rich command manager, and 2 basic actions. This is a first pr. Features and commands will be added step by step hotkey, glyph, context, ...)

The actions (IActions and IToggleAction) make it easy to implement a command.
The commands (IRichCommands) gives properties and methods useful to views. IRichCommand should never be implemented.

The actions ShowFileExtensions and ShowHiddenItems are already used by the menu.

**Validation**
- [x] Built and ran the app


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
```
yaira2, line 3: Spacing shouldn't be changed in this file

src/Files.App/Commands/IRichCommand.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading.Tasks;
+using System.Windows.Input;
+
+namespace Files.App.Commands
+{
+	public interface IRichCommand : ICommand, INotifyPropertyChanging, INotifyPropertyChanged
+	{
+		CommandCodes Code { get; }
+
+		string Label { get; }
```
yaira2, line 12: Should we have a separate string for automation properties? Maybe one for tooltips as well.

cinqmilleans: I notice that the automation.name always have the same text as the label. Concerning the particular case of commands, there is no reason for automation.name to be different from the name of the command. You can specify that it is a command (Command Show file extensions). This makes it possible not to complicate the actions, it suffices to add a property to RichCommand.

For the tooltip, it will probably be necessary to add a Description property to the actions. It will probably also be useful for UIs to manage commands (hotkeys, and maybe future custom menus).

Does this fit?

yaira2: That's in line with what I'm thinking. Perhaps we can automatically generate the tooltip by combining the label and hotkey.

src/Files.App/Strings/en-US/Resources.resw
cinqmilleans: It's fixed.

yaira2 (approved):
LGTM

src/Files.App/Commands/IRichCommand.cs
cinqmilleans: It's done like this in the draft, there is a LabelWithHotKey property at IRichCommand. This is not the case here because the hotkeys will come later.

d2dyno1 (approved):

PR #11335, opened 2023-02-17
Fix: Fixed issue where online-only files deleted before confirmation

### LINKED ISSUE #11318: Bug: Deleting online-only files from OneDrive removes files regardless of the confirmation prompt
### Description

Deleting online-only files from OneDrive removes files regardless of the confirmation prompt. This is actually the correct behavior considering File Explorer does the same exact thing, but I think it's preferable to let the user choose and follow the option we have in settings.

### Steps To Reproduce

1. Open OneDrive
2. Create any type of file
3. Right click and choose the option to free up space
4. Delete file
5. See that it's deleted before the user confirms

### Steps to fix

- Allow the user to choose

### Files Version

v2.4.34

### Windows Version

Windows 11

### Log file

NA

hishitetsu:
**Resolved / Related Issues**
- Closes #11318 

**Details**
Online-only files cannot be tried for deletion probably because of Vanara or Win32 API, so they are treated as to be permanently deleted.

**Validation**
- [X] Built and ran the app

### DIFF
--- a/src/Files.App/Helpers/FileOperationsHelpers.cs
+++ b/src/Files.App/Helpers/FileOperationsHelpers.cs
@@ -157,9 +157,7 @@ namespace Files.App.Helpers
 				}
 
 				if (!tryDelete)
-				{
 					return (true, shellOperationResult);
-				}
 
 				var deleteTcs = new TaskCompletionSource<bool>();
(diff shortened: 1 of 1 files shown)


yaira2 (approved):
LGTM

PR #11344, opened 2023-02-17
Feature: Improve Column View

### LINKED ISSUE #6902: Cannot continue with the folder tree in columns layout after relaunching Files

### LINKED ISSUE #7589: Option to open complete path of favorite folder in "Columns" view

### LINKED ISSUE #11128: Feature: Improve backwards navigation in Column View

hishitetsu:
**Resolved / Related Issues**
- Closes #6902 
- Closes #7589
- Closes #11128

**Details**
* When a directory is opened in column view, any folder pinned to favorites starts from the directory and everything else from root.
* When the navigation button is clicked or the application relaunches to show column view, the last opened directory is displayed (not the last focused directory).

**Validation**
- [X] Built and ran the app

### DIFF
--- a/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
@@ -107,14 +107,29 @@ namespace Files.App.Views.LayoutModes
 		{
 			if (eventArgs.Parameter is NavigationArguments navArgs)
 			{
-				navArgs.FocusOnNavigation = (navArgs.AssociatedTabInstance as ColumnShellPage)?.ColumnParams?.Column == 0; // Focus filelist only if first column
 				columnsOwner = (navArgs.AssociatedTabInstance as FrameworkElement)?.FindAscendant<ColumnViewBrowser>();
+				navArgs.FocusOnNavigation = (navArgs.AssociatedTabInstance as ColumnShellPage)?.ColumnParams?.Column == columnsOwner?.NumberOfColumns() - 1;
+				if (!navArgs.FocusOnNavigation)
+				{
+					FileList.ContainerContentChanging += HighlightPathDirectory;
+				}
 			}
 
 			base.OnNavigatedTo(eventArgs);
 
 			FolderSettings.GroupOptionPreferenceUpdated -= ZoomIn;
 			FolderSettings.GroupOptionPreferenceUpdated += ZoomIn;
+		}
+
+		private void HighlightPathDirectory(ListViewBase sender, ContainerContentChangingEventArgs args)
+		{
+			if (args.Item is ListedItem item && columnsOwner?.OwnerPath() is string ownerPath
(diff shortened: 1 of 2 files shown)


yaira2:
It looks like the arrow key navigation for switching between columns is broken, it seems that this issue can be fixed if we close columns that aren't focused.

hishitetsu:
Fixed

yaira2 (dismissed):
LGTM

PR #11351, opened 2023-02-18
Fix: A small fix to Improve Column View

hishitetsu:
**Resolved / Related Issues**
This is a small fix for #11344 .
#11344 would probably still work, but theoretically it might cause timing problems, so I fixed it.

**Validation**
- [X] Built and ran the app


yaira2 (approved):
LGTM

PR #11367, opened 2023-02-19
RichCommand: HotKeys + Open/FullScreen

cinqmilleans:
**Resolved / Related Issues**
Added to RichCommands.
1) Support for hotkeys (default and custom). The code to modify the custom hotkey is already present but it must not be used before having converted all the commands, otherwise there is a risk of having 2 commands on the same hotkey.
2) Added AutomationName property. Currently, returns the label. This property allows you to already use it in views. We can therefore modify its content later if we wish.
3) Renamed existing commands to always use a verb. The behavior is more explicit.
4) Conversion of OpenHelp and ToggleFullscreen commands, accessible only by HotKey (F1/F11).

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Actions/Global/ToggleFullScreenAction.cs
+++ b/src/Files.App/Actions/Global/ToggleFullScreenAction.cs
@@ -1,0 +1,37 @@
+﻿using Files.App.Commands;
+using Files.App.Extensions;
+using Microsoft.UI.Windowing;
+using System.Threading.Tasks;
+using Windows.System;
+
+namespace Files.App.Actions
+{
+	internal class ToggleFullScreenAction : IToggleAction
+	{
+		public string Label => "FullScreen".GetLocalizedResource();
+
+		public HotKey HotKey => new(VirtualKey.F11);
+
+		public bool IsOn
+		{
+			get
+			{
+				var window = App.GetAppWindow(App.Window);
+				return window.Presenter.Kind is AppWindowPresenterKind.FullScreen;
+			}
+		}
+
+		public Task ExecuteAsync()
+		{
+			var window = App.GetAppWindow(App.Window);
+
+			var newKind = window.Presenter.Kind is AppWindowPresenterKind.FullScreen
+				? AppWindowPresenterKind.Overlapped
+				: AppWindowPresenterKind.FullScreen;
+
+			window.SetPresenter(newKind);
+
+			return Task.CompletedTask;
+		}
+	}
+}
--- a/src/Files.App/Actions/Global/ToggleFullScreenAction.cs
+++ b/src/Files.App/Actions/Global/ToggleFullScreenAction.cs
@@ -8,9 +8,9 @@ namespace Files.App.Actions
 {
 	internal class ToggleFullScreenAction : IToggleAction
 	{
-		public string Label => "FullScreen".GetLocalizedResource();
+		public string Label { get; } = "FullScreen".GetLocalizedResource();
 
-		public HotKey HotKey => new(VirtualKey.F11);
+		public HotKey HotKey { get; } = new(VirtualKey.F11);
 
 		public bool IsOn
 		{
--- a/src/Files.App/Actions/IAction.cs
+++ b/src/Files.App/Actions/IAction.cs
@@ -1,10 +1,13 @@
-﻿using System.Threading.Tasks;
+﻿using Files.App.Commands;
+using System.Threading.Tasks;
 
 namespace Files.App.Actions
 {
 	public interface IAction
 	{
 		string Label { get; }
+
+		HotKey HotKey => HotKey.None;
 
 		bool IsExecutable => true;
 
--- a/src/Files.App/Commands/HotKey.cs
+++ b/src/Files.App/Commands/HotKey.cs
@@ -1,0 +1,157 @@
+﻿using System;
+using System.Linq;
+using System.Text;
+using Windows.System;
+
+namespace Files.App.Commands
+{
+	public readonly struct HotKey : IEquatable<HotKey>
+	{
+		public static HotKey None { get; } = new(VirtualKey.None, VirtualKeyModifiers.None);
+
+		public bool IsNone => Key is VirtualKey.None;
+
+		public VirtualKey Key { get; } = VirtualKey.None;
+		public VirtualKeyModifiers Modifiers { get; } = VirtualKeyModifiers.None;
+
+		public HotKey(VirtualKey key) : this(key, VirtualKeyModifiers.None) { }
+		public HotKey(VirtualKey key, VirtualKeyModifiers modifiers)
+		{
+			if (key is VirtualKey.None)
+				return;
+
+			if (IsModifier(key))
+				throw new ArgumentException("The key cannot be a modifier.", nameof(key));
+
+			Key = key;
+			Modifiers = modifiers;
+
+			static bool IsModifier(VirtualKey key)
+				=> key is VirtualKey.Menu or VirtualKey.LeftMenu or VirtualKey.RightMenu
+				or VirtualKey.Control or VirtualKey.LeftControl or VirtualKey.RightControl
+				or VirtualKey.Shift or VirtualKey.LeftShift or VirtualKey.RightShift
+				or VirtualKey.LeftWindows or VirtualKey.RightWindows;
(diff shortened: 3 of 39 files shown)


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2955,4 +2955,10 @@
   <data name="ShowHiddenItems" xml:space="preserve">
     <value>Show hidden items</value>
   </data>
+  <data name="Help" xml:space="preserve">
+    <value>Aide</value>
```
yaira2, line 2959: Shouldn't this be help?

src/Files.App/Commands/Manager/HotKeyChangedEventHandler.cs
```diff
@@ -0,0 +1,4 @@
+﻿namespace Files.App.Commands
+{
+	public delegate void HotKeyChangedEventHandler(ICommandManager manager, HotKeyChangedEventArgs e);
```
yaira2, line 3: This feels over engineered, does it really need it's own class?

src/Files.App/Strings/en-US/Resources.resw
yaira2: Or maybe it can say "Open documentation"

cinqmilleans: I wrote it in French. It is fixed.

src/Files.App/Commands/Manager/HotKeyChangedEventHandler.cs
cinqmilleans: It is not a class, only a delegate.

yaira2: True, but the main point is that it doesn't need a separate file.

cinqmilleans: The delegate is removed.

yaira2 (dismissed):
LGTM

src/Files.App/Commands/HotKey.cs
```diff
@@ -0,0 +1,157 @@
...
+				"1" => VirtualKey.Number1,
+				"2" => VirtualKey.Number2,
+				"3" => VirtualKey.Number3,
+				"4" => VirtualKey.Number4,
+				"5" => VirtualKey.Number5,
+				"6" => VirtualKey.Number6,
+				"7" => VirtualKey.Number7,
+				"8" => VirtualKey.Number8,
+				"9" => VirtualKey.Number9,
+				"Pad0" => VirtualKey.NumberPad0,
```
d2dyno1, line 99: Do we use Pad(s)?

src/Files.App/Actions/IAction.cs
```diff
@@ -1,11 +1,14 @@
...
+﻿using Files.App.Commands;
+using System.Threading.Tasks;
 
 namespace Files.App.Actions
 {
 	public interface IAction
 	{
 		string Label { get; }
 
+		HotKey HotKey => HotKey.None;
```
d2dyno1, line 10: Could you replace it with `{ get; }`?

src/Files.App/Commands/HotKey.cs
yaira2: Not right now but we'll need this if we're going to support remapping.

src/Files.App/Actions/IAction.cs
cinqmilleans: Isn't it better to avoid assigning in memory since the value is static?

d2dyno1: It is assigned in the memory anyway. The code above uses **default interface implementation**. I'd suggest replacing it with `{ get; }` and assigning it in the implementation (or not, assuming that `HotKey.None` *is* the default value in which case it'll be set automatically for you)

src/Files.App/Actions/Global/ToggleFullScreenAction.cs
```diff
@@ -0,0 +1,37 @@
...
+using Files.App.Extensions;
+using Microsoft.UI.Windowing;
+using System.Threading.Tasks;
+using Windows.System;
+
+namespace Files.App.Actions
+{
+	internal class ToggleFullScreenAction : IToggleAction
+	{
+		public string Label => "FullScreen".GetLocalizedResource();
```
d2dyno1, line 11:
```suggestion
		public string Label { get; } = "FullScreen".GetLocalizedResource();
```

`=>` **is not the same as** `{ get; }`!!!
In this case using `=>` will execute `GetLocalizedResource()` *every time* the property is accessed, whereas using `{ get; }` will make it readonly and assign it *only once*

src/Files.App/Actions/Global/ToggleFullScreenAction.cs
```diff
@@ -0,0 +1,37 @@
...
+
+		public HotKey HotKey => new(VirtualKey.F11);
```
d2dyno1, line 13: Same here and in other places too

src/Files.App/Commands/Manager/CommandManager.cs
```diff
@@ -134,12 +190,14 @@ private void Action_PropertyChanging(object? _, PropertyChangingEventArgs e)
 						break;
 				}
 			}
-			private void Action_PropertyChanged(object? sender, PropertyChangedEventArgs e)
```
d2dyno1, line 137: Please don't discard `sender` parameters 🙂

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -36,6 +38,7 @@ namespace Files.App.Views
 	public sealed partial class MainPage : Page, INotifyPropertyChanged
 	{
 		public IUserSettingsService UserSettingsService { get; } = Ioc.Default.GetRequiredService<IUserSettingsService>();
+		public ICommandManager Commands = Ioc.Default.GetRequiredService<ICommandManager>();
```
d2dyno1, line 41:
```suggestion
		public ICommandManager Commands { get; } = Ioc.Default.GetRequiredService<ICommandManager>();
```

d2dyno1 (changes requested):

src/Files.App/Views/MainPage.xaml.cs
cinqmilleans: You are right. Thanks.

src/Files.App/Actions/IAction.cs
cinqmilleans: It is modified in the classes, but not in the interface because it generates a compilation error.

src/Files.App/Actions/Global/ToggleFullScreenAction.cs
cinqmilleans: It's done.

src/Files.App/Actions/Global/ToggleFullScreenAction.cs
cinqmilleans: It's done.

src/Files.App/Commands/Manager/CommandManager.cs
cinqmilleans: It's done.

yaira2:
@cinqmilleans can you confirm you tested the actions to make sure the work the same as before?

cinqmilleans:
Yes, I confirm.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2955,6 +2955,12 @@
   <data name="ShowHiddenItems" xml:space="preserve">
     <value>Show hidden items</value>
   </data>
+  <data name="Help" xml:space="preserve">
+    <value>Help</value>
+  </data>
+  <data name="FullScreen" xml:space="preserve">
+    <value>Full screen</value>
```
yaira2, line 2962:
```suggestion
    <value>Toggle full screen</value>
```

src/Files.App/Actions/IAction.cs
d2dyno1: Yeah, that's why I think it should be assigned in the implementation and not in the interface by setting the default value

yaira2 (approved):

PR #11382, opened 2023-02-20
Feature: Added menu option to restore all items from recycle bin

### LINKED ISSUE #10048: Restore recycle bin on flyout menu on sidebar

When right clicking the Recycle Bin on the sidebar, the Empty Action is available. The Restore Action should be available as well.

### Requirements

- This proposal will accomplish a more fluid experience with the Recycle Bin

### Files Version

2.3.28.0

### Windows Version

W11Pro 21H2 22000.978

### Comments

_No response_

QuaintMako:
**Resolved / Related Issues**
- Closes #10048 

**Validation**
- [X] Built and ran the app

**Note**
- The `Restore All Items` action from within the recycle bin has been tested as well, it works as expected.

**Screenshots**
RestoreSidebar


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -123,6 +123,9 @@
   <data name="SideBarOpenInNewWindow.Text" xml:space="preserve">
     <value>Open in new window</value>
   </data>
+  <data name="SideBarRestoreRecycleBin.Text" xml:space="preserve">
```
yaira2, line 126:
```suggestion
  <data name="RestoreAllItems" xml:space="preserve">
```

yaira2:
Can you try using the rich commands that were recently introduced?

src/Files.App/Strings/en-US/Resources.resw
QuaintMako: This key is already used in for the recycle bin inner toolbar. 
The problem is that all the words are capitalized, and we don't want that capitalization in the flyout menu. 
I could make a `FirstCharToUpper()` method to use it, but I'm not sure how it would react for non-ascii characters (like Chinese).

yaira2: How about SidebarRestoreAllItems? The `.Text` isn't needed.

QuaintMako: Alright, will fix that!

src/Files.App/Helpers/RecycleBinHelpers.cs
```diff
@@ -109,8 +109,7 @@ public static async Task RestoreRecycleBin(IShellPage associatedInstance)
 
 			if (result == ContentDialogResult.Primary)
 			{
-				associatedInstance.SlimContentPage.ItemManipulationModel.SelectAllItems();
-				await RestoreItem(associatedInstance);
+				Vanara.Windows.Shell.RecycleBin.RestoreAll();
```
yaira2, line 112: Can you use CsWin32 instead of Vanara?

QuaintMako: > Can you use CsWin32 instead of Vanara?

When you say CsWin32, what DLL/namespace are you referring to? I found a project named CsWin32 on Github by Microsoft, but I reckon that's not it?

yaira2: https://github.com/microsoft/CsWin32

QuaintMako: @onein528 you seem to know how CsWin32 works, would you happen to know if there is a way to restore the recycle bin from it?

onein528: No. I didn’t get how that works about string handling. They dont have StringBuilder and have PWSTR. And I couldn’t understand how can I assign them.

QuaintMako:
I'm closing this PR as #11472 will reduce the work needed, plus removes the need to use either Vanara or CsWin32 to restore all of the recycle bin.

PR #11411, opened 2023-02-22
Fix: Catch COMException wherever it occurs in shell extension loading

hishitetsu:
**Resolved / Related Issues**
- This is a follow-up to #11363 

**Details**
* COMException is now caught wherever it occurs in shell extension loading. In this case, the context menu will appear without the shell extension.
* `GetContextMenuForFolder()` has been removed as it does not seem to be called from anywhere.

**Validation**
- [X] Built and ran the app

### DIFF
--- a/src/Files.App/Shell/ContextMenu.cs
+++ b/src/Files.App/Shell/ContextMenu.cs
@@ -157,37 +157,6 @@ namespace Files.App.Shell
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


yaira2:
Is it just me or does the menu load faster now 🤔

hishitetsu:
> Is it just me or does the menu load faster now 🤔

That's strange, because nothing should change unless COMException is thrown.

yaira2:
Maybe a placebo but what matters is that it's working a lot better now.

PR #11415, opened 2023-02-22
RichCommand: Glyph + ContextMenu + EmptyRecycleBinAction

cinqmilleans:
**Resolved / Related Issues**
* Usage of glyphs in IAction. This simplifies the use of icons with the FontIcon and ColoredIcon properties of IRichCommand.
* Simplifies the creation of context menus. It is possible to pass a RichCommand to ContextMenuFlyoutItemViewModel to initialize it.
* Added RichCommand EmptyRecycleBin, which uses glyphs. This command is used everywhere, including the sidebar. Old commands are deleted. This new action does not yet use the IsExecutable property. This requires the context class and will be the subject of a future pr.
* A behavior has changed slightly. In the context menu of the RecycleBin page, the RecycleBin command is no longer active if the RecycleBin is empty. This seemed like a bug to me.
* A bug has been fixed in the parser of HotKey (not yet used).
* The next step is to add the context. This is the last part of the command base. Then, it will be necessary to convert the current commands (actions + use in the views).

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
+++ b/src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
@@ -1,0 +1,16 @@
+﻿using Files.App.Commands;
+using Files.App.Extensions;
+using Files.App.Helpers;
+using System.Threading.Tasks;
+
+namespace Files.App.Actions
+{
+	internal class EmptyRecycleBinAction : IAction
+	{
+		public string Label { get; } = "EmptyRecycleBin".GetLocalizedResource();
+
+		public RichGlyph Glyph { get; } = new RichGlyph("\uEF88", fontFamily: "RecycleBinIcons");
+
+		public async Task ExecuteAsync() => await RecycleBinHelpers.EmptyRecycleBin();
+	}
+}
--- a/src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
+++ b/src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
@@ -11,6 +11,9 @@ namespace Files.App.Actions
 
 		public RichGlyph Glyph { get; } = new RichGlyph("\uEF88", fontFamily: "RecycleBinIcons");
 
-		public async Task ExecuteAsync() => await RecycleBinHelpers.EmptyRecycleBin();
+		public async Task ExecuteAsync()
+		{
+			await RecycleBinHelpers.EmptyRecycleBin();
+		}
 	}
 }
--- a/src/Files.App/Commands/RichGlyph.cs
+++ b/src/Files.App/Commands/RichGlyph.cs
@@ -30,11 +30,14 @@ namespace Files.App.Commands
 			if (IsNone)
 				return null;
 
-			return new FontIcon
+			var icon = new FontIcon
 			{
 				Glyph = BaseGlyph,
-				FontFamily = ToFontFamily(),
 			};
+			if (!string.IsNullOrEmpty(FontFamily))
+				icon.FontFamily = (FontFamily)Current.Resources[FontFamily];
+
+			return icon;
 		}
 
 		public ColoredIcon? ToColoredIcon()
@@ -42,12 +45,15 @@ namespace Files.App.Commands
 			if (IsNone)
 				return null;
 
-			return new ColoredIcon
+			var icon = new ColoredIcon
 			{
 				BaseLayerGlyph = BaseGlyph,
 				OverlayLayerGlyph = OverlayGlyph,
-				FontFamily = ToFontFamily(),
 			};
+			if (!string.IsNullOrEmpty(FontFamily))
+				icon.FontFamily = (FontFamily)Current.Resources[FontFamily];
+
+			return icon;
 		}
 	}
 }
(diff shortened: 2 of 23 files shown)


yaira2:
Can you add support for OpacityIcon? We're currently only using them in the toolbar but we're going to want to use them in the context menu as well.

src/Files.App/ViewModels/ContextMenuFlyoutItemViewModel.cs
```diff
@@ -84,6 +85,36 @@ public class ContextMenuFlyoutItemViewModel
 		public bool ShowLoadingIndicator { get; set; }
 
 		public bool IsHidden { get; set; }
+
+		public ContextMenuFlyoutItemViewModel()
+		{
+		}
+		public ContextMenuFlyoutItemViewModel(IRichCommand command)
```
yaira2, line 92: Small nit, can you add an empty line?

cinqmilleans:
I haven't looked at this class but probably it's possible. I will watch for the next iteration.

src/Files.App/UserControls/SidebarControl.xaml.cs
```diff
@@ -208,16 +203,12 @@ private List<ContextMenuFlyoutItemViewModel> GetLocationItemMenuItems(INavigatio
 					Command = RestoreLibrariesCommand,
 					ShowItem = options.IsLibrariesHeader
 				},
-				new ContextMenuFlyoutItemViewModel()
+				new ContextMenuFlyoutItemViewModel(commands.EmptyRecycleBin)
 				{
-					Text = "BaseLayoutContextFlyoutEmptyRecycleBin/Text".GetLocalizedResource(),
-					Glyph = "\uEF88",
-					GlyphFontFamilyName = "RecycleBinIcons",
```
yaira2, line 215: Is the font family name still needed?

src/Files.App/Helpers/ContextFlyoutItemHelper.cs
```diff
@@ -565,14 +563,10 @@ private static bool Check(ContextMenuFlyoutItemViewModel item, CurrentInstanceVi
...
-				new ContextMenuFlyoutItemViewModel()
+				new ContextMenuFlyoutItemViewModel(commands.EmptyRecycleBin)
 				{
-					Text = "BaseLayoutContextFlyoutEmptyRecycleBin/Text".GetLocalizedResource(),
-					Glyph = "\uEF88",
-					GlyphFontFamilyName = "RecycleBinIcons",
-					Command = commandsViewModel.EmptyRecycleBinCommand,
+					IsEnabled = RecycleBinHelpers.RecycleBinHasItems(),
 					ShowItem = !itemsSelected && currentInstanceViewModel.IsPageTypeRecycleBin,
-					ShowInRecycleBin = true,
```
yaira2, line 575: Why was this removed?

yaira2:
@cinqmilleans please fill out validation steps and add any other steps you used. This helps us know what to test for when reviewing the PR.

src/Files.App/Helpers/ContextFlyoutItemHelper.cs
cinqmilleans: ShowInThing properties are set to true in the constructor. The ShowItem property has the final say on whether the item is displayed or not. Here it is enough since ShowItem tests IsPageTypeRecycleBin. With contexts, ShowItem will no longer be needed because ShowItem is initialized with IsExecutable in the constructor.

src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Files.App.Actions
+{
+	internal class EmptyRecycleBinAction : IAction
+	{
+		public string Label { get; } = "EmptyRecycleBin".GetLocalizedResource();
+
+		public RichGlyph Glyph { get; } = new RichGlyph("\uEF88", fontFamily: "RecycleBinIcons");
+
+		public async Task ExecuteAsync() => await RecycleBinHelpers.EmptyRecycleBin();
```
d2dyno1, line 14: Could you add brackets `{ ... }`?

src/Files.App/Commands/RichGlyph.cs
```diff
@@ -0,0 +1,59 @@
...
+	{
+		public static RichGlyph None { get; } = new(string.Empty);
+
+		public bool IsNone => string.IsNullOrEmpty(BaseGlyph);
+
+		public string BaseGlyph { get; }
+		public string OverlayGlyph { get; }
+		public string FontFamily { get; }
+
+		public RichGlyph(string baseGlyph, string overlayGlyph = "", string fontFamily = "")
```
d2dyno1, line 18: Same here, and in other places too

src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
cinqmilleans: It's done.

src/Files.App/Commands/RichGlyph.cs
cinqmilleans: It's done.

yaira2 (approved):
LGTM

d2dyno1 (approved):

PR #11444, opened 2023-02-23
Fix: Fixed issue with open terminal as admin not working

ferrariofilippo:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #11443 
- Related #11281 
- Related #11034 

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [ ] Tested the changes for accessibility

PR #11445, opened 2023-02-23
Feature: Open in Terminal RichCommand

ferrariofilippo:
**Notes**
- Work in progress
- @cinqmilleans I need your help: how can I pass a parameter to `Execute` without messing up all your code?

**Validation**
- [x] Built and ran the app


cinqmilleans:
Hello @ferrariofilippo . This is a feature that will be coming soon.

The solution proposed by @mafra99, which at first glance seems the right one, poses too many problems and greatly reduces the interest of richcommands because the parameter would only be given when the command is executed.
* The action can no longer manage its descriptions (Label, ...) and it's still the view that will have to do it.
* The action can no longer manage its state (IsExecutable), which differs depending on the parameter.
* The different versions will not be able to be in the command manager, because only the views will know the parameters.
* The command should be called as an ICommand using the CommandParameter. Can't call it with Tapped, which richcommands can do.
* It is not possible to assign a hotkey to its commands since the action would not know which parameter to use.

Here is the solution that will be implemented.
* ICommandManager will have a new way to register an custom action (and also to delete it).
* It will therefore be possible in code behind to add several richcommands using the same Action class. The parameter will be passed in its constructor.
* The commands created will be visible in the command manager and the user will be able to assign them a hotkey.
* The Action class can itself manage all its properties (label, IsExecutable, hotkey by default) because each instance knows its parameter.
* To save a new action, you will need to give it a unique key. This key will save custom hotkeys between sessions.
* These custom commands will not be accessible directly through a property of ICommandManager. It will be necessary to use the instance of IRichCommand provided by registering the action.
* You can find a custom command using its unique key.

All comments are welcome.

cinqmilleans:
I just read your code, and I better understand your need. Actions need to be able to observe the application to run but also to update properties. This is currently too difficult to do. The solution will be to use contexts.

A context is an observable service that easily returns useful information to actions. There will be several depending on the type of action. I am just preparing the most important context which will make it very easy to know the active folder, the selected elements and others. It should be released this weekend and with this context you will have everything you need.

ferrariofilippo:
That's great, thank you

cinqmilleans:
@ferrariofilippo You can now use the new IContentPageService which will provide you with what you need.

ferrariofilippo:
We can't bind that directly to a `KeyboardAccelerator` since there's no `VirtualKey` for `` ` ``

ferrariofilippo:
Yes

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2601,6 +2601,12 @@
   <data name="ApplyToAllConflictingItems" xml:space="preserve">
     <value>Apply this action to all conflicting items</value>
   </data>
+  <data name="OpenTerminal" xml:space="preserve">
+    <value>Open in terminal</value>
+  </data>
+  <data name="OpenTerminalAsAdmin" xml:space="preserve">
+    <value>Open in terminal as admin</value>
```
yaira2, line 2608:
```suggestion
    <value>Open in terminal as administrator</value>
```

hecksmosis (dismissed):
LGTM

src/Files.App/Views/BaseShellPage.cs
```diff
@@ -284,19 +285,10 @@ protected void ShellPage_PreviewKeyDown(object sender, KeyRoutedEventArgs args)
 				// Ctrl + ` (accent key), open terminal
 				case (true, _, false, true, (VirtualKey)192):
```
yaira2, line 286: Isn't this supposed to be removed?

ferrariofilippo: > We can't bind that directly to a `KeyboardAccelerator` since there's no `VirtualKey` for `` ` ``

src/Files.App/Strings/en-US/Resources.resw
ferrariofilippo: I won't be able to fix that until Saturday

src/Files.App/Views/BaseShellPage.cs
yaira2: @cinqmilleans any solutions for this?

cinqmilleans: I could not test because my keyboard does not have a direct ` key. If you use this shortcut, whole countries will not have access to it.

I'm not sure what the problem is you are having. You should not use KeyboardAccelerator yourself. KeyboardAccelerators are handled automatically at MainPage. Just give a HotKey for it to be used. It is also displayed in context menus automatically using ContextFlyoutItemViewModelBuilder. To display the Hotkey in another menu, you can use the HotKeyText property of IRichCommand. It does not display anything when there is no HotKey.

It would also be necessary to modify the parser and tostring of the struct HotKey to use the string `. It will look prettier in the menus.

cinqmilleans: I realize that there are a few subtleties to be aware of. I will write a short guide to implementing actions and using commands. We will put this in the documentation.

ferrariofilippo: > I could not test because my keyboard does not have a direct ` key. If you use this shortcut, whole countries will not have access to it.

You don't need to press specifically `` ` `` to use that shortcut. Windows localize \` by default, for example I can use `ò` to trigger this shortcut

ferrariofilippo: @yair100 this is the same way it was implemented in #11502 see file `src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs`

cinqmilleans:
@yair100 I can't help as I don't have that key on my keyboard. I recommend using another. It will probably be necessary to list the acceptable touches in all cultures.

ferrariofilippo:
You don't need that key, I can access the shortcut using the ò key. Windows will localize the shortcut, you can use the same that you use in VS code to open terminal

cinqmilleans:
@ferrariofilippo A fix has been committed.

This hotkey will be displayed as "Ctrl+`" in the menus. You can also parse it explicitly [HotKey.Parser("Ctrl+`")] or implicitly [(HotKey)"Ctrl+`"]. It's better to use the explicit constructor for performance.

You should completely remove the BaseShellPage code. The hotkey will be managed automatically by CommandManager. The difference with the admin version will be automatic because it is not the same hotkey.

cinqmilleans:
I will look at this in more detail tomorrow.

cinqmilleans:
Here are my remarks.
* If I start opening a folder in administrator mode, I get a popup from Windows to confirm. If I cancel, there is a Win32Exception which is not caught by the catch and crashes the application.
* The user has no way of knowing his commands (waiting for the command viewer). Maybe add these commands to the context menu. It's wider than this pr, but here's an idea. Use an Open menu in the context menu with a submenu (here/in a new tab/pane/windows, in a terminal, in a terminal (admin)). To discuss.
* If multiple folders are selected, it only opens the first one. Open them all or none.
* The label of OpenTerminalAsAdmin is not correct because it is the label of OpenTerminal which is used when the instance is of type IAction, even if the class is OpenTerminalAsAdmin. Use virtual/override instead of new string.
* Replace new Glyph() with new() in both instances (to avoid a warning).
* Use (is "") instead of (== string.Empty). This avoids using the equality operator.
* * This action is always executable because IsExecutable is not set. However, the keyboard shortcut can open a terminal which indicates an error if it is used in the wrong place (homepage, recyclebin).
Implement IsExecutable to filter use cases. No need to do this in ExecuteAsync, which will never be called when IsExecutable is false.
* This action is not observable. It doesn't matter if it's only a hotkey or a menu. If we add it to an always visible button, it will be problematic. Notify IsExecutable changes by subscribing to IContentPageContext changes (PageType,HasSelection,SelectedItem)

ferrariofilippo:
Thank you, I'll work on that later
~I remember @yair100 told me not to add this option to the context menu, but maybe they changed their mind~ That's actually wrong, see #11444

ferrariofilippo:
> * This action is not observable. It doesn't matter if it's only a hotkey or a menu. If we add it to an always visible button, it will be problematic. Notify IsExecutable changes by subscribing to IContentPageContext changes (PageType,HasSelection,SelectedItem)

I won't subscribe to `HasItem` and `SelectedItem` since if there's no selected item, it will open base folder

ferrariofilippo:
This is not finished yet, I still need to open all paths in a single wt instance, but I need to figure out how to know which shell is running to do this

cinqmilleans:
If you open the terminal at the root of a drive, the terminal indicates that the folder is invalid. You must use the path C:\ instead of C: for drives, but especially not for other folders.

ferrariofilippo:
> If you open the terminal at the root of a drive, the terminal indicates that the folder is invalid. You must use the path C:\ instead of C: for drives, but especially not for other folders.

Actually the issue is that the back slash escapes the following characther

yaira2:
@cinqmilleans is this ready to merge?

cinqmilleans:
The problems:
* Methods with poor visibility (private/public).
* IsExecutable difficult to read.
* Add a limit to avoid opening 1000 folders per accident.
* Update too frequent because not filtered.
* If a selected item is not a folder, it opens the parent folder. The most logical thing is to do nothing.
* Do not use ShellPage when values are available directly in Context. This avoids null problems.

Here is an improved version. The 4 must be replaced by a constant to be placed in the constants file.

```
using CommunityToolkit.Mvvm.ComponentModel;
using CommunityToolkit.Mvvm.DependencyInjection;
using Files.App.Commands;
using Files.App.Contexts;
using Files.App.Extensions;
using System;
using System.ComponentModel;
using System.Diagnostics;
using System.Linq;
using System.Threading.Tasks;
using Windows.Storage;
using Windows.System;

namespace Files.App.Actions
{
	internal class OpenTerminalAction : ObservableObject, IAction
	{
		private readonly IContentPageContext context = Ioc.Default.GetRequiredService<IContentPageContext>();

		public virtual string Label { get; } = "OpenTerminal".GetLocalizedResource();

		public virtual HotKey HotKey { get; } = new((VirtualKey)192, VirtualKeyModifiers.Control);

		public RichGlyph Glyph { get; } = new("\uE756");

		private bool isExecutable;
		public bool IsExecutable => isExecutable;

		public OpenTerminalAction()
		{
			isExecutable = GetIsExecutable();
			context.PropertyChanged += Context_PropertyChanged;
		}

		public Task ExecuteAsync()
		{
			var startInfo = GetProcessStartInfo();
			if (startInfo is not null)
			{
				App.Window.DispatcherQueue.TryEnqueue(() =>
				{
					try
					{
						Process.Start(startInfo);
					}
					catch (Win32Exception)
					{
					}
				});
			}
			return Task.CompletedTask;
		}

		protected virtual ProcessStartInfo? GetProcessStartInfo()
		{
			var paths = GetPaths();
			if (paths.Length is 0)
				return null;

			return new()
			{
				FileName = "wt.exe",
				Arguments = string.Join(" ; nt ", paths.Select(path => $"-d \"{path}\"")),
			};
		}

		private string[] GetPaths()
		{
			if (context.HasSelection)
			{
				return context.SelectedItems
					.Where(item => item.PrimaryItemAttribute is StorageItemTypes.Folder)
					.Select(item => item.ItemPath)
					.ToArray();
			}
			else if (context.Folder is not null)
			{
				return new string[1] { context.Folder.ItemPath };
			}
			return Array.Empty<string>();
		}

		private bool GetIsExecutable()
		{
			if (context.PageType is ContentPageTypes.None or ContentPageTypes.Home or ContentPageTypes.RecycleBin or ContentPageTypes.ZipFolder)
				return false;
			if (!context.HasSelection)
				return context.PageType is not ContentPageTypes.SearchResults && context.Folder is not null;
			if (context.SelectedItems.Count > 4)
				return false;
			return context.SelectedItems.Any(item => item.PrimaryItemAttribute is StorageItemTypes.Folder);
		}

		private void Context_PropertyChanged(object? sender, PropertyChangedEventArgs e)
		{
			switch (e.PropertyName)
			{
				case nameof(IContentPageContext.PageType):
				case nameof(IContentPageContext.Folder):
				case nameof(IContentPageContext.SelectedItems):
					SetProperty(ref isExecutable, GetIsExecutable(), nameof(IsExecutable));
					break;
			}
		}
	}
}

```

ferrariofilippo:
> @ferrariofilippo The problems:

I didn't apply `GetProcessStartInfo()` changes since they would cause the previous bug with drives

ferrariofilippo:
> The 4 must be replaced by a constant to be placed in the constants file.

I changed it to `5` to be more consistent with `Open in New Tab` and similar

yaira2:
Anything else @cinqmilleans or is this ready to merge?

cinqmilleans:
@yair100 @ferrariofilippo It is mandatory to select a folder to activate the hotkey. It doesn't work on parent folder without seletion. Yet the action handles this case everywhere except in GetIsExecutable. Modify this method to enable it.

PR #11472, opened 2023-02-26
Rich command: Context + Restore

cinqmilleans:
**Resolved / Related Issues**
This pr adds the first and main context used by actions. A context is an observable service that easily provides actions with properties that are difficult to know. To observe properties of the current page, you must notify the properties when it changes or when a parent or ancestor object changes.

* Added ContentPageContext. The data is always from the current page. Selected items are easily notified (whether it's a selection or a page change).
* Added RestoreRecycleBin and RestoreAllRecycleBin commands.
* Improved EmptyRecycleBin with context.

**Validation**
- [x] Built and ran the app


src/Files.App/Contexts/ContentPage/IContentPageContext.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Files.App.Filesystem;
+using System.Collections.Generic;
+using System.ComponentModel;
+
+namespace Files.App.Contexts
+{
+	public interface IContentPageContext : INotifyPropertyChanged
```
d2dyno1, line 7: What is this interface for?

cinqmilleans: This interface is a context service which gives the active content of the page (The type of page, the selected elements, ...).
The properties of this context are useful to most commands and must be observable for the command to notify its state changes.

If I want to know the selected items, currently I have to observe ToolbarViewModel, but this instance itself can change
without notification if I switch tabs. It is therefore necessary to observer several "floors" of an object and update each time the floor changes. The context makes its information easy to obtain and notifies of changes.

d2dyno1: I guess it's okay, but on the other hand it bothers me because we're just further spreading the use of IShellPage...

cinqmilleans: What do you mean ?

src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
```diff
@@ -1,19 +1,52 @@
...
 
 		public RichGlyph Glyph { get; } = new RichGlyph("\uEF88", fontFamily: "RecycleBinIcons");
 
+		public bool IsExecutable
+		{
+			get
+			{
+				if (context.PageType is ContentPageTypes.RecycleBin)
+					return context.HasItem;
+				return RecycleBinHelpers.RecycleBinHasItems();
```
QuaintMako, line 26: Can we just use a simple `RecycleBinHelpers.RecycleBinHasItems()` here? If you did it, I suppose it's because the `context.hasItem` is lighter to execute?

src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
```diff
@@ -1,19 +1,52 @@
...
+			}
+		}
+
+		public EmptyRecycleBinAction()
+		{
+			context.PropertyChanged += Context_PropertyChanged;
```
QuaintMako, line 32: No need to unsubscribe somewhere?

src/Files.App/Actions/FileSystem/RestoreAllRecycleBinAction.cs
```diff
@@ -0,0 +1,53 @@
...
+
+		public RichGlyph Glyph { get; } = new RichGlyph("\xE777");
+
+		public bool IsExecutable
+		{
+			get
+			{
+				if (context.PageType is ContentPageTypes.RecycleBin)
+					return context.HasItem;
+				return RecycleBinHelpers.RecycleBinHasItems();
```
QuaintMako, line 26: Same as for EmptyRecycleBin

src/Files.App/Actions/FileSystem/RestoreAllRecycleBinAction.cs
```diff
@@ -0,0 +1,53 @@
...
+			}
+		}
+
+		public RestoreAllRecycleBinAction()
+		{
+			context.PropertyChanged += Context_PropertyChanged;
```
QuaintMako, line 32: Same as for EmptyRecycleBin

src/Files.App/Actions/FileSystem/RestoreRecycleBinAction.cs
```diff
@@ -0,0 +1,46 @@
...
+
+		public string Label { get; } = "Restore".GetLocalizedResource();
+
+		public RichGlyph Glyph { get; } = new RichGlyph("\xE777");
+
+		public bool IsExecutable => context.PageType is ContentPageTypes.RecycleBin && context.SelectedItems.Any();
+
+		public RestoreRecycleBinAction()
+		{
+			context.PropertyChanged += Context_PropertyChanged;
```
QuaintMako, line 25: Same has EmptyRecycleBin

src/Files.App/Actions/FileSystem/RestoreAllRecycleBinAction.cs
cinqmilleans: Same as for EmptyRecycleBin

src/Files.App/Actions/FileSystem/RestoreRecycleBinAction.cs
cinqmilleans: Same as for EmptyRecycleBin

src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
cinqmilleans: Yes, context.HasItem is 5-10 times faster. But this is a trick that can only be used if the RecycleBin is displayed in the current page.

src/Files.App/Commands/Manager/CommandManager.cs
```diff
@@ -121,6 +122,9 @@ private class ActionCommand : ObservableObject, IRichCommand
 			private readonly Lazy<ColoredIcon?> coloredIcon;
 			public ColoredIcon? ColoredIcon => coloredIcon.Value;
```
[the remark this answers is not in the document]
cinqmilleans: It's done.

cinqmilleans:
I updated to resolve an incompatibility with the latest commits. This forced me to integrate the management of OpacityIcon

* RichGlyph can be of type OpacityIcon
* ColoredIcon is removed.
* Restore(All) uses OpacityIcon.
* Added ContextMenuFlyoutItemViewModelBuilder. This new class avoids having to read the properties of the command when it is not displayed in the menu, to improve performance.

When all the commands have been converted, it will be possible to replace ContextMenuFlyoutItemViewModel with an interface and 3 implementations to improve code and performance.

Known bug: In the context menu, the Restore entry of a RecycleBin item has no icon. However, the element does have an OpacityIcon content. I don't think the bug is related to this code.

yaira2:
> Known bug: In the context menu, the Restore entry of a RecycleBin item has no icon. However, the element does have an OpacityIcon content. I don't think the bug is related to this code.

I noticed this as well, I think we can leave as is for now (same issue in `main`)

src/Files.App/Contexts/ContentPage/IContentPageContext.cs
yaira2: @d2dyno1 is this good to go?

src/Files.App/App.xaml
```diff
@@ -6,8 +6,6 @@
 	<Application.Resources>
 		<ResourceDictionary>
 			<FontFamily x:Key="CustomGlyph">/Assets/Custom Glyphs/Custom-Glyphs.ttf#Custom-Glyphs</FontFamily>
-			<FontFamily x:Key="RecycleBinIcons">/Assets/Custom Glyphs/recyclebin-fabric-glyph.ttf#Fabric MDL2 Assets</FontFamily>
-			<FontFamily x:Key="ColoredIconFontFamily">/Assets/Custom Glyphs/Colored-Icons.ttf#Untitled1</FontFamily>
```
yaira2, line 10: @winston-de 😂

yaira2 (approved):
LGTM

PR #11502, opened 2023-02-28
Rich command: Selection

cinqmilleans:
**Resolved / Related Issues**
*Adds commands SelectAll and ClearSelection and InvertSelection and MultiSelect. The menu is disabled if the command has no effect.
* Adds properties HasSelection and SelectedItem to IContentPageContext for to make it easier. SelectedItem is null when the selection does not contain one and only one element.
* The Ctrl+A hotkey is handled in the file lists and the appropriate command is reissued, to be able to reassign this hotkey.
* Rich commands use a new HotKeyText property, which is null when the hotkey is empty. Otherwise, the menus display None instead of displaying nothing.

**Validation**
- [ ] Built and ran the app

### DIFF
--- a/src/Files.App/Commands/CommandCodes.cs
+++ b/src/Files.App/Commands/CommandCodes.cs
@@ -20,11 +20,17 @@
 		RestoreRecycleBin,
 		RestoreAllRecycleBin,
 
-    // Start
+		// selection
+		MultiSelect,
+		SelectAll,
+		InvertSelection,
+		ClearSelection,
+
+		// start
 		PinToStart,
 		UnpinFromStart,
 
-		// Favorites
+		// favorites
 		PinItemToFavorites,
 		UnpinItemFromFavorites,
 	}
--- a/src/Files.App/Strings/en-US/Resources.resw
+++ b/src/Files.App/Strings/en-US/Resources.resw
@@ -153,13 +153,13 @@
   <data name="Skip" xml:space="preserve">
     <value>Skip</value>
   </data>
-  <data name="NavToolbarSelectAll.Text" xml:space="preserve">
+  <data name="SelectAll" xml:space="preserve">
     <value>Select All</value>
   </data>
-  <data name="NavToolbarInvertSelection.Text" xml:space="preserve">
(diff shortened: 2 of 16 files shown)


src/Files.App/Commands/CommandCodes.cs
```diff
@@ -20,11 +20,17 @@ public enum CommandCodes
...
+		SelectAll,
+		InvertSelection,
+		ClearSelection,
+
+		// start
 		PinToStart,
 		UnpinFromStart,
 
-		// Favorites
+		// favorites
```
yaira2, line 33: Why this change?

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2604,4 +2604,7 @@
   <data name="Save" xml:space="preserve">
     <value>Save</value>
   </data>
+  <data name="MultiSelect" xml:space="preserve">
+    <value>Multiselect</value>
```
yaira2, line 2608: Why is there a command for this?

cinqmilleans: This command is no longer used in the menu. However, the command selector can display commands that do not have direct access. It is one of his interests. In addition, it allows you to assign it a custom hotkey. Personally, this command will be the first to have one when possible.

src/Files.App/Commands/CommandCodes.cs
yaira2: The other sections should be capital as well.

src/Files.App/Strings/en-US/Resources.resw
yaira2: This is a setting though, not an action.

cinqmilleans: Toggle a setting can be an action. This is very practical when you often modify certain settings.

src/Files.App/Commands/CommandCodes.cs
cinqmilleans: It's done.

yaira2:
Thank you

yaira2 (approved):

PR #11515, opened 2023-03-01
Feature: Introduce AnimatedIcon to backward/forward buttons

onein528:
### Description

As the title

### Motivation and Context

There's an animated icon for backward glyph. And I rotate it and introduced it to the forward button.

- Removed unnecessary Grid nesting.
- Introduce AnimatedIcon to the Backward button.
- Introduce AnimatedIcon to the Forward button.
- Fix border-less button style(Completely removed border by setting BorderThickness to 0).
- Add some comments.

### Validation

- [x] Built and ran the app

### DIFF
--- a/src/Files.App/UserControls/AddressToolbar.xaml
+++ b/src/Files.App/UserControls/AddressToolbar.xaml
@@ -19,6 +19,7 @@
 	d:DesignWidth="800"
 	Loading="NavToolbar_Loading"
 	mc:Ignorable="d">
+
 	<UserControl.Resources>
 		<ResourceDictionary>
 			<converters1:BoolNegationConverter x:Key="BoolNegationConverter" />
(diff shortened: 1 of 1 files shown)


src/Files.App/UserControls/AddressToolbar.xaml
```diff
@@ -34,7 +35,7 @@
 						TargetType="Button">
 						<Setter Property="Padding" Value="0" />
 						<Setter Property="Background" Value="Transparent" />
-						<Setter Property="BorderBrush" Value="Transparent" />
+						<Setter Property="BorderThickness" Value="0" />
```
yaira2, line 38: This breaks the border brush on hover.

yaira2: It causes contrast issues in light mode

onein528: Hmmm. Right, then I will revert them.

PR #11483, opened 2023-02-27
Fix: Added notice for third party licenses

hishitetsu:
Since Files contains third-party software that is not under the MIT license, it should be clearly stated that a different license applies to them.
I also think it should be added to third party licenses display on the app's settings screen, but first I submit a pull request for the license file.

Edit:
Now LICENSE is unchanged. I have added NOTICE.md to clarify third-party software licenses.

PR #11531, opened 2023-03-02
Feature: Display NOTICE.md in the About page

hishitetsu:
**Resolved / Related Issues**
- Related to #11483 

**Details**
This may not be necessary since @d2dyno1 is working on the license page, but I submit it for reference.
NOTICE.md is included in the application package instead of referencing the online one since the third-party libraries used may change from version to version.
I also removed the link to the license of QuickLook since it is not included in Files and merely has the feature to call it.

**Validation**
- [X] Built and ran the app

**Screenshots**
image


onein528:
FIiles don’t have display full license term. So if you’d like to add 3rd party notice in Files, display only package name by using Hyperlink Button.

hishitetsu:
Displaying third party licenses is to comply with their licenses, so it is different from whether or not Files itself has the feature to display its own license.
For example, Windows Terminal does not have the feature to display its own license, but it does have third party notices.
image
(Sorry for Japanese)

onein528:
Me too Japanese, in fact. this PR's problem is redundancy.

hishitetsu:
> Me too Japanese, in fact.

I knew it! So I uploaded the screenshot in Japanese 😃 

> this PR's problem is redundancy.

That is a point to be discussed.
In Windows Terminal, the browser is opened to show the HTML file converted from NOTICE.md as third-party notices. I don't think this is a good idea, so I have embedded NOTICE.md in the About page.
In any case, users will not care much about third-party licenses, so I think it would be sufficient to display NOTICE.md as it is.

yaira2:
@hishitetsu can you fix the merge conflicts?

src/Files.App/Views/Settings/AboutPage.xaml
```diff
@@ -15,7 +15,8 @@
 		<StackPanel
 			HorizontalAlignment="Stretch"
 			VerticalAlignment="Stretch"
-			Spacing="4">
+			Spacing="4"
+			Loading="{x:Bind ViewModel.About_Loading}">
```
yaira2, line 19: Instead of loading when the page loads, would it make sense to load it when the licenses section is expanded?

hishitetsu: Done.

hishitetsu:
> @hishitetsu can you fix the merge conflicts?

Successfully merged the upstream main.

src/Files.App/Views/Settings/AboutPage.xaml.cs
```diff
@@ -17,5 +21,17 @@ public AboutPage()
 
 			ViewModel = new AboutViewModel();
 		}
+
+		private void ThirdPartyLicenses_Tapped(object sender, TappedRoutedEventArgs e)
```
yaira2, line 25: Is this called when the control is loaded?

hishitetsu: This is called when Third party licenses SettingsBlockControl is clicked and MarkdownTextBlock is to expand.

onein528: There isn't Expanded event?

hishitetsu: Seems not. It did not appear in IntelliSense.

onein528: You can use Click event, instead.

https://github.com/files-community/Files/blob/3712ddf262fd37147b306682e270dc9c6fefc211/src/Files.App/UserControls/Settings/SettingsBlockControl.xaml.cs#L133-L141

yaira2 (approved):
LGTM

src/Files.App/Views/Settings/AboutPage.xaml.cs
hishitetsu: @onein528 Thank you! Fixed.

src/Files.App/Views/Settings/AboutPage.xaml.cs
```diff
@@ -17,5 +20,17 @@ public AboutPage()
 
 			ViewModel = new AboutViewModel();
 		}
+
+		private void ThirdPartyLicenses_Click(object sender, bool e)
+		{
+			if (e && ViewModel.ThirdPartyNotices is null)
+				ViewModel.LoadThirdPartyNotices();
+		}
```
onein528, line 28:
```suggestion
		private void ThirdPartyLicenses_Click(object sender, RoutedEventArgs args)
		{
			if (ViewModel.ThirdPartyNotices is null)
				ViewModel.LoadThirdPartyNotices();
		}

```

onein528 (changes requested):

src/Files.App/Views/Settings/AboutPage.xaml.cs
hishitetsu: `bool` is the correct event argument here. `RoutedEventArgs` would result in a compile error.

onein528: Opps, you right :)

PR #11546, opened 2023-03-04
Feature: Update checkbox behavior in details layout to replace icon

chingucoding:
This PR updates the behavior of the selection checkboxes in Details layout. Insteaed of listening to the app settings, now the "select all checkbox" will always be visible and the rows checkboxes will replace the icon if the user is hovering over the row or any row is selected.

**Validation**
- [X] Built and ran the app

**Screenshots (optional)**
Add screenshots here.


yaira2:
@chingucoding can you fix the merge conflicts and update from `main`?

yaira2:
@chingucoding it looks like the xaml got messed up in the merge, can you take a look?

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -547,7 +550,7 @@
 								Visibility="{x:Bind ColumnsViewModel.StatusColumn.Visibility, Mode=OneWay}" />
 
 							<controls:GridSplitter
-								Grid.Column="20"
+								Grid.Column="19"
```
yaira2, line 553: Does changing the columns affect resizing them?

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -561,247 +565,272 @@
...
+									IsRightTapEnabled="True"
+									Loaded="StackPanel_Loaded"
+									Orientation="Horizontal"
+									ToolTipService.ToolTip="{x:Bind ItemTooltipText, Mode=OneWay}">
+									<Grid
+										Width="24"
 										HorizontalAlignment="Left"
-										VerticalAlignment="Bottom"
-										x:Load="{x:Bind IsShortcut}">
-										<uc:OpacityIcon Style="{StaticResource ColorIconShortcut}" />
```
yaira2, line 637: Was this removed on purpose?

chingucoding: Looks like a bad merge, fixed now.

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
chingucoding: Given that we removed a column, my guess would be no and some testing also indicate resizing still works.

src/Files.App/Views/Settings/FoldersPage.xaml
```diff
@@ -238,18 +238,6 @@
 					Style="{StaticResource RightAlignedToggleSwitchStyle}" />
 			</local:SettingsBlockControl>
 
-			<!--  Show checkboxes next to file names  -->
-			<local:SettingsBlockControl Title="{helpers:ResourceString Name=SettingsCheckboxesToSelectItems}" HorizontalAlignment="Stretch">
```
yaira2, line 242: Can you remove the string?

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
yaira2: Double clicking a grid splitter to resize the column doesn't seem to work. Same with the size all to fit option.

chingucoding: Are those behaviors currently present in the store version? I'm not sure what "double clickign" the grid splitter is supposed to do. The behavior for both things you mentioned seems to be the same compared to the store version 🤔

yaira2: Compare with `main`

yaira2: This behavior was broken in the store version and fixed in `main`

chingucoding: I think I fixed this now. The whole column logic is a real pain to work with though...

yaira2: You're telling me 😂

yaira2 (approved):
LGTM

PR #11399, opened 2023-02-21
Feature: Remember previously selected option in conflicts dialog

hishitetsu:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #11266 

**Details**
* When the Apply to All option is other than Custom and the Continue button is pressed, the option that was selected is saved.
* Fixed a bug that caused the Apply to All option to be applied to non-conflicted files.
* The flyout-related code did not appear to be used, so it was removed.

**Validation**
How did you test these changes?
- [X] Built and ran the app
- [X] Tested the changes for accessibility

PR #11557, opened 2023-03-05
Fix: Use x:Load in conflicts dialog again

hishitetsu:
**Resolved / Related Issues**
- Related to #11399 

**Details**
In #11399, I used `Visibility` instead of `x:Load` because the initial value of the combo box could not be set correctly.
But I found a way to set the initial value correctly for `x:Load`, so I changed `Visibility` back to `x:Load`.

**Validation**
- [X] Built and ran the app


yaira2 (approved):
LGTM

PR #11565, opened 2023-03-05
Fix: Fixed PageContext + DisplayPageContext

cinqmilleans:
**Resolved / Related Issues**
The ContentPageContext class contains a bug. IShellPage is not always up to date, especially when opening and then closing the column view. This pr notably fixes this by using a new method to update IShellPage. Do not update actions that already use the context.

* New context PageContext which only manages the update of IShellPage, quite complex.
* Simplified ContentPageContext using PageContext.
* New context DisplayPageContext, to manage the layout, sorting and group (Commands will arrive in the next pr, to facilitate the review).

**Validation**
- [x] Built and ran the app


src/Files.App/Contexts/DisplayPage/LayoutTypes.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Files.App.Contexts
+{
+	public enum LayoutTypes : ushort
```
yaira2, line 3: Isn't there already an enum for this?

cinqmilleans: Not really. There are FolderLayoutModes (DetailsView/TilesView/ColumnView/GridView/Adaptive) and GridViewSizeKind (Small/Medium/Large). It is more convenient for commands to simply view or modify a single enum.

cinqmilleans:
Copied in #11669.

PR #11552, opened 2023-03-04
Code Quality: Reorganze namespaces

onein528:
### Description

Reorganize namespaces. For more information, see Details of Changes

### Motivation and Context

I found that there're a lot of namespaces that need to be reorganized and there're namespaces that do not need to be separated and I merged them into the same.

### Details of Changes

Previous|New|Rationale
---|---|---
`Files.App.Controllers.IJson`|*Remove*|Unused
`Files.App.ValueConverters`|`Files.App.Converters`|Same thing
`Files.App.AppModels`|`Files.App.DataModels`|Same thing
`Files.App.CommandLine`|`Files.Backend.CommandLine`|Be able to be moved to Backend
`UserControls/Pane/PreviewPane`|`UserControls/PreviewPane`|No need to be another folder
UserControls' ViewModels(`Files.App.ViewModels.*`)|`Files.App.ViewModels.UserControls`|Organized

### Validation

- [x] Built and ran the app
- [x] Tested the changes for accessibility

PR #11574, opened 2023-03-06
Fix: Fixed XAML data binding failure

Poker-sang:
**Resolved / Related Issues**
- Closes 

**Details**
When opening a file list, `CheckBox.Width`'s data binding failed for there's no `GridLengthToDouble` converter

**Validation**
- [x] Built and ran the app


Poker-sang:
> How can I reproduce this issue?

I just build and run the app, and the issue comes up. I guess it can be reproduced when using details view and enabling checkboxes?

Because `GridLength` cannot convert to `double` implicitly, I thought this should be fixed sooner or later. 🤔

yaira2:
I think there is an existing converter that we can use.

chingucoding (dismissed):
lgtm

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
```diff
@@ -571,7 +572,7 @@
 
 								<CheckBox
 									x:Name="SelectionCheckbox"
-									Width="{Binding ElementName=SelectAllColumn, Mode=OneWay, Path=Width}"
```
chingucoding, line 574: I think we can probably even remove this line given the visibility setter at the bottom.

src/Files.App/Converters/GridLengthToDouble.cs
```diff
@@ -0,0 +1,20 @@
...
+		return value is GridLength { IsAbsolute: true } gridLength 
+			? gridLength.Value 
+			: throw new InvalidCastException("GridLength with a \"Star\"('*') or \"Auto\" value cannot be converted to double");
+	}
+
+	public object ConvertBack(object value, Type targetType, object parameter, string language)
+	{
+		return new GridLength((double)value);
+	}
+}
```
QuaintMako, line 20: The class should be contained inside the namespace I believe?

Poker-sang: I am sorry that I don't understand. Do you mean I should put this class in another namespace?

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
Poker-sang: It seems like you are right.  `CheckBox.Width` is actually the width of the label of the `CheckBox`. So I Remove this line.

Poker-sang:
I removed the binding, and merged `Files.App.Converters` and `Files.App.ValueConverters` namespace. I thought they were duplicated.

Poker-sang:
I removed the binding, and merged `Files.App.Converters` and `Files.App.ValueConverters` namespace. I thought they were duplicated.

src/Files.App/Converters/GridLengthToDouble.cs
QuaintMako: We usually use the notation with curly brackets for the namespace scopes on File. Meaning your code should be like this

```csharp
namespace Files.App.Converters
{
    public class GridLengthToDouble : IValueConverter { ... }
}
```

Rather than: 

```csharp
namespace Files.App.Converters;

public class GridLengthToDouble : IValueConverter { ... }
```

Poker-sang: Ok. I'll get right on it.

hecksmosis:
> I removed the binding, and merged `Files.App.Converters` and `Files.App.ValueConverters` namespace. I thought they were duplicated.

This is already being done in #11552

Poker-sang:
> This is already being done in #11552

I also simplify some code in the converter classes. But I can restore it if necessary.

yaira2:
@Poker-sang thank you for your contribution, in general we prefer to limit the scope of PRs to the related issue/feature. As per the contribution guidelines, it's strongly recommended to open an issue and wait for an org member to approve it before opening the PR. This is more for the contributors benefit than ours because it helps everyone be on the same page and avoids duplicate work.
Files is a very active community project, so sticking to these guidelines helps keep track of which work is already in progress and which stuff still need to be picked up 🙂

Poker-sang:
I am sorry. I just encountered a XAML failure and tried to fix it. If this PR was duplicated, please close it. I will pay more attention to this from now on.

yaira2:
The data binding fix is good but the name spaces changes are being done in another PR.

QuaintMako (dismissed):

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml
Poker-sang: Oh yes, there is no need for it. I will remove it.

yaira2 (approved):
LGTM

PR #11578, opened 2023-03-06
RichCommand: TogglePreviewPane

cinqmilleans:
**Resolved / Related Issues**
Adds new command: TogglePreviewPane

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Actions/Show/TogglePreviewPaneAction.cs
+++ b/src/Files.App/Actions/Show/TogglePreviewPaneAction.cs
@@ -1,0 +1,39 @@
+﻿using CommunityToolkit.Mvvm.ComponentModel;
+using Files.App.Commands;
+using Files.App.Extensions;
+using Files.App.ViewModels;
+using System.ComponentModel;
+using System.Threading.Tasks;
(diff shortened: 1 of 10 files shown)


src/Files.App/Actions/Show/TogglePreviewPaneAction.cs
```diff
@@ -0,0 +1,39 @@
...
+		public bool IsOn => viewModel.IsEnabled;
+
+		public TogglePreviewPaneAction()
+		{
+			viewModel.PropertyChanged += ViewModel_PropertyChanged;
+		}
+
+		public Task ExecuteAsync()
+		{
+			viewModel.IsEnabled = !IsOn;
```
QuaintMako, line 29: Shouldn't we check if ViewModel is not null ?

cinqmilleans: This property is defined only once when launching the application. It cannot be null.

QuaintMako (approved):

yaira2:
Thanks

PR #11588, opened 2023-03-07
Fix: Fixed OpacityIcon visibility

### LINKED ISSUE #11540: Bug: Empty and Restore button disappear when opening the recycle bin twice
### Description

Empty and Restore button is not displayed when opening the recycle bin twice.
Furthermore, the app crashes when trying to open another folder from this state.
image

### Steps To Reproduce

1. Open the recycle bin and you see Empty and Restore button.
2. Open another folder.
3. Open the recycle bin again.
4. Empty and Restore button disappear this time.
5. If trying to open another folder from this state, the app will crash.

### Steps to fix

Show the buttons every time and avoid the crash.

### Files Version

2.4.49.0

### Windows Version

10.0.22621.1265

### Log file

[debug.log](https://github.com/files-community/Files/files/10883319/debug.log)

cinqmilleans:
**Resolved / Related Issues**
Fixes #11540.
XAML treats bindings of a Content attribute badly. We have to pass the content as a tag. This pr adds IRichCommand.OpacityStyle. This allows to display the icon without knowing the style used. The downside is that you have to know if it's a fonticon or opacityicon. The advantage is to be able to change the size (this will be used for the layout icons). This solution is acceptable because the icon will rarely change, and never by the user. Additionally, FontIcon may be phased out in favor of the more generic OpacityStyle.

I also removed IsEnabled properties to IsExecutable. By default, the command already notifies its state using IsExecutable, and XAML reacts to it.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Commands/IRichCommand.cs
+++ b/src/Files.App/Commands/IRichCommand.cs
@@ -1,4 +1,5 @@
 ﻿using Files.App.UserControls;
+using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Input;
 using System.ComponentModel;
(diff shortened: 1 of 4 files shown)


yaira2:
@hishitetsu can you test?

hishitetsu:
Tested and worked. LGTM

PR #11589, opened 2023-03-07
Fix: Fixed SelectAll HotKey

cinqmilleans:
**Resolved / Related Issues**
Views still have code to perform SelectAll with the shortcut Ctrl+A. This pr removes this code because the SelectAll action triggers this hotkey. This avoids a bug if we change the hotkey of the action or the action of the hotkey.

**Validation**
- [x] Built and ran the app


yaira2 (approved):
LGTM

PR #11608, opened 2023-03-08
Fix: Fixed issue where toolbar icons were mistakenly enabled on startup

hishitetsu:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Related to #11588 

**Details**
In #11588, `IsEnabled` properties were removed in favor of `IsExecutable`. However, without `IsEnabled`, `IsExecutable` is not evaluated on initialization and icons are shown as 'Enabled', so Cut, Copy and Delete still need `IsEnabled` property.
I missed this behavior when I tested #11588. Sorry.

**Validation**
How did you test these changes?
- [X] Built and ran the app
- [X] Tested the changes for accessibility

PR #11612, opened 2023-03-08
Simplify commands cut copy delete

cinqmilleans:
**Resolved / Related Issues**
Improves the commands (warnings, optimisation).
Context.HasSelection is much better suited (for IsExecutable). It can't be null, is false on the home page, and is only updated when needed.

**Validation**
- [ ] Built and ran the app


cinqmilleans:
The commit #11608, a few minutes ago, is not correct. The button'state is OK. The property IsEnabled is is unnecessary because it is included in the command.

hecksmosis (approved):
LGTM

yaira2 (approved):
Thank you

PR #11539, opened 2023-03-03
Fix: Fixed issue where delete button didn't work on toolbar

hishitetsu:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #11534 

**Validation**
How did you test these changes?
- [X] Built and ran the app
- [X] Tested the changes for accessibility

PR #11614, opened 2023-03-08
Fix: Fixed OpacityIcon state

### LINKED ISSUE #11605: Bug: Delete has the wrong color
### Description

When inactive, the Delete icon is colored, unlike all the others. However, it is not possible to click on it.

### Steps To Reproduce

1. Display Home.
2. The icon delete is colored.

### Steps to fix

Modify the style of the Delete icon when it is inactive.

### Files Version

2.4.49.0

### Windows Version

10.0.19045.2604

### Log file

Home Page
one
Detail Page
two

cinqmilleans:
**Resolved / Related Issues**
The state of OpacityIcon is updated whenever its parent control's IsEnabled changes. But it is not initialized. It can therefore be active when its control is not.

This is the case when starting the application on the homepage. The Cut, Copy and Delete toolbar icons are shown as active while the button is inactive.

This pr updates the state of OpacityIcon when loaded. Fixes #11605.
This pr is not the same as #11539, which fixes the button. This pr fixes the icon.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/UserControls/OpacityIcon.xaml.cs
+++ b/src/Files.App/UserControls/OpacityIcon.xaml.cs
@@ -6,7 +6,10 @@ namespace Files.App.UserControls
 {
 	public sealed partial class OpacityIcon : Control
 	{
-		public OpacityIcon() => InitializeComponent();
+		public OpacityIcon()
+		{
+			InitializeComponent();
+		}
 
 		private void IsEnabledChange(DependencyObject sender, DependencyProperty dp)
 		{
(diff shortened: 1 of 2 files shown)


src/Files.App/UserControls/OpacityIcon.xaml.cs
```diff
@@ -2,38 +2,30 @@
...
-
 namespace Files.App.UserControls
 {
 	public sealed partial class OpacityIcon : Control
 	{
-		public OpacityIcon()
-		{
-			InitializeComponent();
-		}
+		public OpacityIcon() => InitializeComponent();
```
yaira2, line 9: This should have brackets

cinqmilleans: It's done.

yaira2 (approved):
LGTM

chingucoding (approved):

PR #11620, opened 2023-03-08
Fix: Fixed context menus in parent folders

### LINKED ISSUE #11615: Bug: Context menu is broken.
### Description

The context menu for the parent folder no longer works. This seems to be coming from the commands that compression. The Label property updates based on the content, but may throw an error.

For the specific case of these commands, I recommend having a generic label, and modifying the label in the menu display. It will be better for the command handler.

### Steps To Reproduce

1 Open the context menu in a base folder

### Steps to fix

Fix the compression commands

### Files Version

Last commit

### Windows Version

Win11

### Log file

None

ferrariofilippo:
**Resolved / Related Issues**
- Closes #11615 

**Details**
I think this soultion is better than having a `Generic Label`, since the latter would not prevent the throwing of an exception (caused by trying to access the first item of an empty collection)

**Validation**
- [x] Built and ran the app


yaira2 (approved):
LGTM

PR #11637, opened 2023-03-09
Display commands hotkeys

cinqmilleans:
**Resolved / Related Issues**
* Fix bad LabelWithHotKey when HotKey is none.
* Fix commands who misdeclares their hotkey (field instead property, so no effect).
* Restore accessKey for cut and copy
* Display HotKey in tooltip for commands in toolbar

AccessKey is not part of RichCommand because it is an element of the interface, not of the command. It is accessible for other elements (menu, textbox, ...). There are no plans to make it editable.

The Delete command does not work because its Execute is incorrect. It will be for another pr.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Commands/Manager/CommandManager.cs
+++ b/src/Files.App/Commands/Manager/CommandManager.cs
@@ -167,7 +167,7 @@ namespace Files.App.Commands
 			public CommandCodes Code { get; }
 
 			public string Label => action.Label;
-			public string LabelWithHotKey => $"{Label} ({CustomHotKey})";
+			public string LabelWithHotKey => !customHotKey.IsNone ? $"{Label} ({CustomHotKey})" : Label;
 			public string AutomationName => Label;
 
 			public RichGlyph Glyph => action.Glyph;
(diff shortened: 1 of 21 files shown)


src/Files.App/Commands/Manager/CommandManager.cs
```diff
@@ -238,6 +238,10 @@ public ActionCommand(CommandManager manager, CommandCodes code, IAction action)
 					notifyPropertyChanging.PropertyChanging += Action_PropertyChanging;
 				if (action is INotifyPropertyChanged notifyPropertyChanged)
 					notifyPropertyChanged.PropertyChanged += Action_PropertyChanged;
+
+				if (code is CommandCodes.CutItem)
```
yaira2, line 242: What's this?

yaira2 (dismissed):
LGTM

src/Files.App/Commands/Manager/CommandManager.cs
cinqmilleans: It's fixed.

PR #11666, opened 2023-03-11
Fix: Fixed null error in RotateLeft/Right

cinqmilleans:
**Resolved / Related Issues**
I had exceptions in these classes to use ShellPage when it is null. Here is a fix.
I improved readability and used distinct to avoid dealing with the same extension multiple times.

**Validation**
- [x] Built and ran the app


QuaintMako:
Good catch, I did not fall on it when doing them! Thanks!

yaira2 (approved):
LGTM

Issue #11481, opened 2023-02-26
Generalizing Rich Command Usage


QuaintMako:
### Description

Thanks to @cinqmilleans' job, Files has now a cleaner way to handle commands across the solution. Rich commands now need to be implemented throughtout the whole solution.

This issue is used to keep track of which commands are made into Rich Commands.

### Concerned code

The following commands must be transformed into Rich Commands (See `BaseLayoutCommandsViewModel`, `ContextFlyoutItemHelper` and `ToolbarViewModel`):

- [x] Clear Selection (#11502)
- [x] Close Selected Tab (#11916)
- [x] Close Pane (#12002)
- [x] Compress Into Archive (#11572)
- [x] Compress Into Seven Zip (#11572)
- [x] Compress Into Zip (#11572)
- [x] Copy Item (#11504)
- [x] Copy Path Of Selected Items (#11774)
- [ ] Create Folder With Selection
- [x] Create New File
- [x] Create New Folder (#11493)
- [x] Create Shortcut (#11493)
- [x] Create Shortcut From Dialog (#11493)
- [x] Cut Item (#11504)
- [x] Decompress Archive (#11723)
- [x] Decompress Archive Here (#11723)
- [x] Decompress Archive To Child Folder (#11723)
- [x] Delete Item (#11504)
- [x] Duplicate Tab (#11596)
- [x] Empty Recycle Bin (#11415 )
- [x] Enter compact overlay (#11821)
- [x] Exit compact overlay (#11821)
- [x] Format Drive #11831
- [x] Gridview Size Decrease (#11678)
- [x] Gridview Size Increase (#11678)
- [x] Group Layer Sort (#11678)
- [x] Install Font (#11780)
- [x] Install Inf Driver (#11812)
- [x] Invert Selection (#11502)
- [x] Multi-Selection (#11502)
- [x] Navigate back (#11829)
- [x] Navigate forward (#11829)
- [x] Navigate up (#11829)
- [x] Navigate To Next Tab (#11916)
- [x] - [x] Navigate To Previous Tab (#11916)
- [ ] Open Directory In New Pane
- [ ] Open Directory In New Tab
- [x] Open File Location
- [ ] Open In New Window Item
- [x] Open In Terminal (#11445)
- [x] Open Item (#11667)
- [x] Open Item With Application Picker (#11667)
- [x] Open New Tab (#11596)
- [x] Open Parent Folder (#11667)
- [x] Open new pane (#12002)
- [x] Open settings #11846
[...]

PR #11667, opened 2023-03-11
Feature: Add RichCommands Open, Open With, Open Parent Folder

hecksmosis:
**Resolved / Related Issues**
- Related to #11481 
- Adds Tag to ContextMenuFlyoutItemViewModelBuilder

**Validation**
- [x] Built and ran the app


src/Files.App/Actions/FileSystem/OpenItemAction.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+	internal class OpenItemAction : ObservableObject, IAction
+	{
+		private readonly IContentPageContext context = Ioc.Default.GetRequiredService<IContentPageContext>();
+
+		public string Label => "Open".GetLocalizedResource();
+
+		public RichGlyph Glyph => new(opacityStyle: "ColorIconOpenFile");
+
+		public bool IsExecutable => context.HasSelection && context.SelectedItems.Count <= 10;
```
yaira2, line 20: What's the number 10? Should there be a constant variable for it?

hecksmosis: It was there in ContextFlyoutItemHelper, I assume it's more or less an arbitrary value to prevent the app from opening too many files at once

yaira2: If we still want to keep it, we should create a constant

hecksmosis: I'll do that

yaira2:
Should the code for these actions be grouped together into a single file?

yaira2:
The layout and sort commands were grouped together and we may do the same for other commands as well.

hecksmosis:
Ok, I'll join them

yaira2:
@hecksmosis can you fix the merge conflicts?

hecksmosis:
Done

QuaintMako (approved):
LGTM.

PR #11670, opened 2023-03-11
Feature: RichCommands Cut/Copy/Paste

cinqmilleans:
**Resolved / Related Issues**
* New rich command PasteItem.
* Fix visibility of private methods.
* The Paste command is inconsistent. In the context menu, the selection must be empty or a single folder of type element. In the toolbar, if there are several elements selected, we copy to the parent folder. Now Paste is always disabled if the selection is more than one item or the item is not a folder.
* The Paste AccessKey is now V instead P. This corresponds to Cut(X) and Copy(C). This adds the accessKey P to the PreviewPane if need.
* Use commands Cut/Copy/Paste in the context menu.
* Paste is always visible in the context menu, It's more intuitive. It was visible in search results (useless) but not in RecycleBin. It was visible in search results (useless) but not in trash. Now Cut/Copy/Paste go together with less chance of getting it wrong.
* Fix the new text Cut.

**Validation**
- [x] Built and ran the app


src/Files.App/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -105,7 +105,7 @@
 					Width="Auto"
 					MinWidth="40"
 					AccessKey="X"
-					AutomationProperties.AutomationId="InnerNavigationToolbarCutButton"
+					AutomationProperties.AutomationId="{x:Bind Commands.CutItem.AutomationName}"
```
yaira2, line 108: AutomationId should not be part of the command.

cinqmilleans: You are right. It's done.

yaira2 (approved):
LGTM

yaira2:
@cinqmilleans can you fix the build error?

PR #11584, opened 2023-03-07
Feature: Introduce NavigationViewItemButtonStyle to Settings dialog

onein528:
### Description

Introducing new NavView style to Settings pages. (In another PR, to Properties pages)

### Motivation and Context

Basing on a new concept style of NavigationView, I am going to make them real.

### Details of changes

- Add NavigationViewButtonStyle
- Refactor SettingsDialog
- Add missing namespace 'Views' in 'Files.App.Settings'

### Validation

- [x] Built and ran the app
- [x] Tested the changes for accessibility

#### Main style

image

#### Collapsed style

image

PR #11672, opened 2023-03-11
Feature: Introduce the new NavigationViewItem style to Properties window

0x5bfa:
### Description

Part of the renovation.

### Motivation and Context

- #11584

### Tasks

- [x] Change style
- [x] Change style to be able to display two different icon corresponding states
- [x] Collapse pane on smaller windows
- [x] Make default window size wider

### Validation

- [x] Built and ran the app

### Screenshots

Default mode
image

Compact mode
image


onein528:
> Change style to be able to display two different icon corresponding states

I can do this primitively(I can do fast) like this:
```xaml
<StackPanel Orientation="Horizontal">
    <Grid>
        <OpacityIcon Visibility="{x:Bind IsSelected, Converter...}" /> <!--  Unselected  -->
        <OpacityIcon Visibility="{x:Bind IsSelected}" /> <!--  Selected  -->
    </Grid>

    <TextBlock Text="{x:Bind Name}" />
</StackPanel>
```

But Idk, maybe still adding VisualState is more better (have to add codes a lot, though).

yaira2:
VisualStates are better and less xaml overall.

src/Files.App/Views/Properties/MainPropertiesPage.xaml
```diff
@@ -33,49 +33,31 @@
...
 			x:Name="MainPropertiesWindowNavigationView"
 			Grid.Row="1"
-			Width="80"
 			IsBackButtonVisible="Collapsed"
 			IsPaneToggleButtonVisible="False"
 			IsSettingsVisible="False"
 			IsTitleBarAutoPaddingEnabled="False"
 			MenuItemsSource="{x:Bind NavViewItems, Mode=TwoWay}"
-			OpenPaneLength="88"
+			OpenPaneLength="200"
```
yaira2, line 60: Is this what we used in settings?

onein528: 260 in settings. should be the same?

yaira2: We can try with 200 👍

0x5bfa:
Do you @yair100 me try that in here PR?

yaira2:
I already implemented it in propeties but it's missing the new styles

src/Files.App/Views/Properties/MainPropertiesPage.xaml
```diff
@@ -37,127 +38,104 @@
...
 
 								<!--  Icon  -->
 								<Viewbox
-									Width="24"
-									Height="24"
+									Width="16"
+									Height="16"
 									VerticalAlignment="Center">
-									<usercontrols:OpacityIcon IsSelected="{x:Bind IsSelected, Mode=OneWay}" Style="{x:Bind OpacityIconStyle, Mode=OneWay}" />
+									<uc:OpacityIcon IsSelected="{x:Bind IsSelected, Mode=OneWay}" Style="{x:Bind OpacityIconStyle}" />
```
yaira2, line 79:
```suggestion
									<uc:OpacityIcon IsSelected="{x:Bind IsSelected, Mode=OneWay}" Style="{x:Bind OpacityIconStyle, Mode=OneWay}" />
```

yaira2: We always define mode

onein528: Can you explain why for me? I'd like to understand.

yaira2: It makes the xaml more readable

onein528: I definitely recommend we should not do that in DataTemplate. This cause a warning `WMC1506: OneWay bindings require at least one of their steps to support raising notifications when their value changes`.

yaira2: Template could be an exception for `OneWay` but I believe it's needed for `TwoWay`.

onein528: Yes, if a data have to be needed two way bind, have to add `Mode=TwoWay`. But should not add `Mode=OneWay` in one way binding. (In DataTemplate)

onein528: Default Mode is OneTime in x:Bind. OpacityIconStyle is being set in its class's constructor so I believe it's not needed. But your decision actually whether we should add OneWay for each binds in TwoWay template.

yaira2: I think it's better for readability and forces developers to double check that they set the correct mode.

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
```diff
@@ -98,6 +98,20 @@ private async void Page_Loaded(object sender, RoutedEventArgs e)
 			}
 		}
 
+		private void MainPropertiesPage_SizeChanged(object sender, SizeChangedEventArgs e)
+			=> UpdateDialogLayout();
```
yaira2, line 102: Coding convention in Files is to use brackets

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
```diff
@@ -388,6 +403,13 @@ public bool IsSelected
 			get => _isSelected;
 			set => SetProperty(ref _isSelected, value);
 		}
+
+		private bool _isCompacted;
+		public bool IsCompacted
```
yaira2, line 408: What is this property for?

src/Files.App/Helpers/FilePropertiesHelpers.cs
```diff
@@ -58,10 +58,7 @@ public static async Task OpenPropertiesWindowAsync(object item, IShellPage assoc
 
 			if (isUniversal)
 			{
-				var frame = new Frame
-				{
-					RequestedTheme = ThemeHelper.RootTheme
-				};
+				var frame = new Frame { RequestedTheme = ThemeHelper.RootTheme };
```
yaira2, line 61: Coding convention is not to use single line here.

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
onein528: The labels such as General, Customisation will not be collapsed automatically when NavigationView mode is LeftCompact. To fix that, detect compact mode and collapse text with this property.

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
onein528: Does that mean that we should not lambda expression for methods in anywhere?

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
yaira2: Where is it set?

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
yaira2: There might be exceptions but not in this context.

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
yaira2: Shouldn't it be `isCompact`?

src/Files.App/Helpers/FilePropertiesHelpers.cs
yaira2: @onein528

yair100 (approved):

onein528:
Oh thank you

PR #11676, opened 2023-03-12
RichCommand: Added support for multiple hot keys

cinqmilleans:
**Resolved / Related Issues**
This pr allows to use several hotkeys for the same action. Added SecondHotKey and MediaHotKey in addition to HotKey. They are optional.
If SecondHotKey is set, it also displays in menus (eg: Ctrl+T,Ctrl-Shit+B.
If MediaHotkey is set, it does not display. This allows to assign the optional media keys of some keyboards (GoBack, GoHome, ...).

The code to modify the HotKeys, which was not yet used, has been removed. It had been designed for a single hotkey per command. A multikey version will come later.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Actions/IAction.cs
+++ b/src/Files.App/Actions/IAction.cs
@@ -10,6 +10,8 @@ namespace Files.App.Actions
 		RichGlyph Glyph => RichGlyph.None;
 
 		HotKey HotKey => HotKey.None;
+		HotKey SecondHotKey => HotKey.None;
+		HotKey MediaHotKey => HotKey.None;
 
 		bool IsExecutable => true;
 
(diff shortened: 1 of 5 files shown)


src/Files.App/Actions/IAction.cs
```diff
@@ -10,6 +10,8 @@ public interface IAction
 		RichGlyph Glyph => RichGlyph.None;
 
 		HotKey HotKey => HotKey.None;
+		HotKey SecondHotKey => HotKey.None;
```
yaira2, line 13: While you're at it, can you add support for a third hot key?

cinqmilleans: It's done.

yaira2 (approved):
LGTM

PR #11657, opened 2023-03-10
Code quality: Introduced WidgetsPageContext

ferrariofilippo:
**Details**
This PR introduces `WidgetsPageContext` which will be needed to use `RichCommands` from the `WidgetsPage`
This context will be useful to implement features such as `Open Properties`, `Copy`, ..., which requires the right clicked item
Once it will be merged, I'll remove `TagsContext` from #11598 since it won't be needed anymore

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [ ] Tested the changes for accessibility

PR #11677, opened 2023-03-12
Code quality: Introduced SidebarContext

ferrariofilippo:
**Details**
This PR introduces `SidebarContext` which will be needed to implement RichCommand for sidebar items (such as `OpenInNewPane`, `OpenInNewTab`, `UnpinItem`, ...)
@cinqmilleans I'm once again asking for your review and help

**Validation**
- [x] Built and ran the app


cinqmilleans:
The goal is for actions to behave differently when a sidebar context menu is open. RightClickedItemChanged is only triggered when the context menu opens, not when it closes, so you have to notify when closed
(RightClickedItemChanged should announce that the item becomes null again).

Booleans can easily be replaced by classes.
public bool IsDriveItem => rightClickedItem is DriveItem;
public DriveItem? OpenDriveItem => rightClickedItem as DriveItem;
This makes IsDriveItemPinned useless.

ferrariofilippo:
> The goal is for actions to behave differently when a sidebar context menu is open. RightClickedItemChanged is only triggered when the context menu opens, not when it closes, so you have to notify when closed (RightClickedItemChanged should announce that the item becomes null again).

I'll implement this also in `WidgetsContext`

cinqmilleans:
@ferrariofilippo 

There is a structural problem with this context. The actions will have to test 2 contexts each time. For example, Properties can apply to the parent folder, selection, or sidebar. The action does not need to know where the item is. We could further simplify the contexts.

ContentPageContext can return active items, be it selection, parent folder or sidebar. Most actions will no longer need to double check as currently (if HasSelection use this else use that). It will take an enum to know where the applied element is located (for example, one can only rename a selected element). But this will simplify the code most of the time.

What do you think ?

ferrariofilippo:
> The actions will have to test 2 contexts each time. 

Actually it would be three, we would have also `WidgetsContext`

> ContentPageContext can return active items, be it selection, parent folder or sidebar. Most actions will no longer need to double check as currently (if HasSelection use this else use that). It will take an enum to know where the applied element is located (for example, one can only rename a selected element). But this will simplify the code most of the time.
> 
> What do you think ?

Yes, that seems reasonable. As I mentioned earlier, we should also merge WidgetsContext into our main context. Should I start working on this or would you rather do it on your own?

I'll mark this and #11657 as Drafts so that you can easily access them if you need to. Once the final PR is merged, I'll close them

cinqmilleans:
@ferrariofilippo The problem that these elements are of different classes. Let's think about the best solution before intervening.

ferrariofilippo:
We can access Item property of the widget which is a INavigationControlItem

cinqmilleans:
@ferrariofilippo @yair100 This solution will not work because it has a flaw. For example Properties is disabled on the homepage. If when we open the context menu of a favorite, the action updates according to this element, the toolbar button will be activated. This behavior is not desirable. For Paste, we used 2 commands, but this complicates the commands in the palette.

A solution may be to use temporary commands for some menus. The menu creation code instantiates an action with a parameter (a specific unobservable class). This class is not visible in the palette. Then, we use CommandManager to get an instance of IRichCommand from this action. It may be a simpler implementation as no need to update it. The menu uses it and then throws it away.

These actions won't need any context, everything will be passed in the constructor. As these are rich commands, we can use the builder to generate the menu.

cinqmilleans:
I have a better idea. I will implement it for the tab menu, it will be more understandable.

ferrariofilippo:
> What's the latest on this?

I'll close this as soon as @cinqmilleans ends their work

cinqmilleans:
I have thought a lot about this problem.  I think I have a practical solution.  I will open an issue to validate it before implementation.

yaira2:
@cinqmilleans what's the latest on this?

PR #11678, opened 2023-03-12
RichCommand: Layout Sort Group

cinqmilleans:
**Resolved / Related Issues**
Add commands Layout Sort Group.

When the option "Sync layout and sorting preferences across directories", the adaptive layout is disable in toolbar but not in context menu. It's fixed.

The layout context menu replaces icons by toggle indicator. This lets you know which layout is active.

**Validation**
- [x] Built and ran the app


yaira2:
Can you group related actions together instead of creating separate files for each one?

cinqmilleans:
It's done.

src/Files.App/Actions/Display/LayoutAction.cs
```diff
@@ -0,0 +1,179 @@
...
+				SetProperty(ref isOn, Context.LayoutType == LayoutType, nameof(IsOn));
+
+			if (e.PropertyName is not null)
+				OnContextChanged(e.PropertyName);
+		}
+
+		protected virtual void OnContextChanged(string propertyName) { }
+	}
+
+	internal class LayoutPreviousAction : IAction
```
yaira2, line 130: What does this do?

cinqmilleans: It is a command that allows you to navigate between all the layouts. This is not used but may be available in the future command handler or with a new or custom hotkey.

yaira2: We already have `ctrl` + `+` and `ctrl` + `-`.

yaira2:
> The layout context menu replaces icons by toggle indicator. This lets you know which layout is active.

Doesn't MenuItem have support for icon as well?

cinqmilleans:
Yes, but only a fonticon, not an opacityicon. The columns layout uses opacityicon. I tested with the tool icon + font icon, it's not aesthetic. I'll send a screenshot later.

src/Files.App/Actions/Display/LayoutAction.cs
cinqmilleans: I replaced Previous/Next with DecreaseSize/IncreaseSize.

cinqmilleans:
Here is a screenshot. We can find a trick for the image of columns.
Layout

yaira2:
I'm going to be updating the other icons in the flyout as well, do you have any ideas on fixing the issue?

cinqmilleans:
It will probably be necessary to create a UserControl derived from ToggleMenuFlyoutItem (as it was done for the buttons). In the preview, the Columns icon is already missing in this context menu. I propose to deal with this problem in a future pr. I will commit the code with the check + icon.

cinqmilleans:
It's done.

yaira2:
@cinqmilleans can you fix the merge conflict?

cinqmilleans:
It's done.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -1419,7 +1419,7 @@
   <data name="DateModifiedLowerCase" xml:space="preserve">
     <value>Date modified</value>
   </data>
-  <data name="NavToolbarArrangementOptionOriginalFolder.Text" xml:space="preserve">
```
yaira2, line 1422: Can you search to see if this is still being used? Same for the other strings that were removed.

cinqmilleans: I had already checked and I just checked again.

PR #11691, opened 2023-03-12
Feature: Added selected state to opacity icon

yaira2:
**Resolved / Related Issues**
- Added IsSelected state to opacity icon

**Validation**
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/UserControls/OpacityIcon.xaml.cs
+++ b/src/Files.App/UserControls/OpacityIcon.xaml.cs
@@ -11,15 +11,28 @@ namespace Files.App.UserControls
 			InitializeComponent();
 		}
 
+
+		public bool IsSelected
+		{
+			get => (bool)GetValue(IsSelectedProperty);
+			set => SetValue(IsSelectedProperty, value);
+		}
+
+		public static readonly DependencyProperty IsSelectedProperty =
+			DependencyProperty.Register(nameof(IsSelected), typeof(bool), typeof(OpacityIcon), new PropertyMetadata(null));
+
 		private void IsEnabledChange(DependencyObject sender, DependencyProperty dp)
 		{
 			string state = sender.GetValue(dp) is false ? "Disabled" : "Normal";
-			VisualStateManager.GoToState(this, state, true);
+
+			if (state == "Normal" && IsSelected)
+				VisualStateManager.GoToState(this, "Selected", true);
+			else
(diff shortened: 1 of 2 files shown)


src/Files.App/UserControls/OpacityIcon.xaml.cs
```diff
@@ -11,15 +11,28 @@ public OpacityIcon()
...
+
 		private void IsEnabledChange(DependencyObject sender, DependencyProperty dp)
 		{
 			string state = sender.GetValue(dp) is false ? "Disabled" : "Normal";
-			VisualStateManager.GoToState(this, state, true);
+
+			if (state == "Normal" && IsSelected)
+				VisualStateManager.GoToState(this, "Selected", true);
+			else
+				VisualStateManager.GoToState(this, state, true);
```
chingucoding, line 31: I think something like this would be better to read:
```c#
if(IsSelected) 
    VisualStateManager.GoToState(this, "Selected", true);
if(sender.GetValue(IsEnabledProperty))
    VisualStateManager.GoToState(this, "Normal", true);
VisualStateManager.GoToState(this, "Disabled", true)
```

chingucoding (commented):
I assume there is no use case for selected and simultanously disabled right?

src/Files.App/UserControls/OpacityIcon.xaml.cs
yaira2: > I assume there is no use case for selected and simultaneously disabled right?

That's right, but it can be selected and enabled at the same time in which `Selected` should have priority.

yaira2: @chingucoding have you tested if that works? I didn't test but the logic doesn't look right.

chingucoding: I did not and it also wouldn't. I think the following would work as expected:
```c#
			if(sender.GetValue(dp) is false)
				VisualStateManager.GoToState(this, "Disabled", true);
			else if (IsSelected)
				VisualStateManager.GoToState(this, "Selected", true);
			else
				VisualStateManager.GoToState(this, "Normal", true);
```

yaira2: This looks good, thanks!

chingucoding (approved):

PR #11715, opened 2023-03-14
Fix: Fixed checkbox visibility for multiselect

cinqmilleans:
**Resolved / Related Issues**
In detail view, the selection checkbox disappears when manuelly unchecked. This pr leaves it visible but unchecked.

In grid view, there is a slight blinking of the element when unchecking the selection checkbox (which disappears briefly then comes back). This pr leaves it visible to avoid this visual issue.

**Validation**
- [x] Built and ran the app

### DIFF
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -741,7 +741,15 @@ namespace Files.App.Views.LayoutModes
 		private void ItemSelected_Unchecked(object sender, RoutedEventArgs e)
 		{
 			if (sender is CheckBox checkBox && checkBox.DataContext is ListedItem item && FileList.SelectedItems.Contains(item))
+			{
+				FileList.SelectionChanged -= FileList_SelectionChanged;
+
 				FileList.SelectedItems.Remove(item);
+				if (SelectedItems is not null && SelectedItems.Contains(item))
+					SelectedItems.Remove(item);
+
+				FileList.SelectionChanged += FileList_SelectionChanged;
+			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -741,7 +741,15 @@ private void ItemSelected_Checked(object sender, RoutedEventArgs e)
...
 			if (sender is CheckBox checkBox && checkBox.DataContext is ListedItem item && FileList.SelectedItems.Contains(item))
+			{
+				FileList.SelectionChanged -= FileList_SelectionChanged;
+
 				FileList.SelectedItems.Remove(item);
+				if (SelectedItems is not null && SelectedItems.Contains(item))
+					SelectedItems.Remove(item);
+
+				FileList.SelectionChanged += FileList_SelectionChanged;
+			}
```
chingucoding, line 752: Instead of trying to try ignore the selection event when unchecking a checkbox, I think the proper solution would be to introduce a new VSM VisualState Group for the selection states and switch between the "Selected" and "Unselected" while also maintaining a "Hover" and "Normal" state.

chingucoding (changes requested):

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
cinqmilleans: Good idea.

cinqmilleans: @chingucoding I don't know much about VSMs. Can you create a new pr to fix this bug?

chingucoding: Sure if you prefer that I can create a new PR to fix that.

PR #11720, opened 2023-03-14
Fix: Don't show DriveUnpluggedDialog error for recent files

cinqmilleans:
**Resolved / Related Issues**
Sometimes a recent file is no longer available because its disk is unplugged. The homepage show an error dialog. It's annoying because the user does not try to access it but receives this message every time.

This pr no longer shows the dialog for recent files.

**Validation**
- [x] Built and ran the app


yaira2 (approved):
LGTM

PR #11724, opened 2023-03-15
Code Quality: Simplify ColorPicker control with `ColorPickerButton`

Poker-sang:
**Details**
We can use new `ColorPickerButton` control to simplify code. It has better appearance and easy to maintain the code.

But both `ColorPicker` and `ColorPickerButton` under namespace `CommunityToolkit.WinUI.UI.Controls` have a XAML binding failure problem. I cannot help it and have to wait for _CommunityToolkit_ to fix.

**Validation**
- [x] Built and ran the app

**Screenshots**
Old control:
image

New one:
image

XAML Binding failure (Unable to fix now):
Error: Converter failed to convert value of type 'null' to type 'Color'; BindingExpression: Path='Color' DataItem='CommunityToolkit.WinUI.UI.Controls.ColorPicker'; target element is 'Microsoft.UI.Xaml.Controls.GridView' (Name='null'); target property is 'SelectedValue' (type 'Object').

### DIFF
--- a/src/Files.App/Views/Settings/AppearancePage.xaml
+++ b/src/Files.App/Views/Settings/AppearancePage.xaml
@@ -126,28 +126,7 @@
 				<local:SettingsBlockControl.Icon>
 					<FontIcon Glyph="&#xE771;" />
 				</local:SettingsBlockControl.Icon>
-				<Button Padding="0" AutomationProperties.Name="{helpers:ResourceString Name=Background}">
-					<StackPanel Orientation="Horizontal">
-						<Border
-							Width="24"
-							Height="24"
-							Margin="4"
-							Background="{ThemeResource App.Theme.BackgroundBrush}"
-							CornerRadius="4" />
-						<FontIcon
-							Margin="8,4,10,4"
-							FontSize="12"
-							Glyph="&#xE019;" />
-					</StackPanel>
-					<Button.Flyout>
-						<Flyout>
-							<toolkit:ColorPicker
-								IsAlphaEnabled="True"
-								IsColorSpectrumVisible="False"
-								Color="{x:Bind ViewModel.AppThemeBackgroundColor, Mode=TwoWay}" />
-						</Flyout>
-					</Button.Flyout>
-				</Button>
+				<toolkit:ColorPickerButton AutomationProperties.Name="{helpers:ResourceString Name=Background}" SelectedColor="{x:Bind ViewModel.AppThemeBackgroundColor, Mode=TwoWay}" />
 				<local:SettingsBlockControl.ExpandableContent>
 					<GridView
 						Padding="8"
--- a/src/Files.App/Views/Settings/TagsPage.xaml
+++ b/src/Files.App/Views/Settings/TagsPage.xaml
@@ -182,28 +182,10 @@
 												</TextBox.Resources>
 											</TextBox>
 
-											<Button
+											<controls:ColorPickerButton
 												Margin="8,0"
-												Padding="0"
-												AutomationProperties.Name="{helpers:ResourceString Name=TagColor}">
-												<StackPanel Orientation="Horizontal">
-													<Border
-														Width="20"
-														Height="20"
-														Margin="4"
-														Background="{x:Bind NewColor, Mode=OneWay}"
-														CornerRadius="4" />
-													<FontIcon
-														Margin="8,4,10,4"
-														FontSize="12"
-														Glyph="&#xE019;" />
-												</StackPanel>
-												<Button.Flyout>
-													<Flyout>
-														<controls:ColorPicker Color="{x:Bind NewColor, Mode=TwoWay}" />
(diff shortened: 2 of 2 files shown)


yaira2:
Thank you for the PR, I wasn't aware of this control. (In the future, it's recommended to open an issue to discuss a change before working on it).

Overall this looks good but there are two styling issues we should fix before merging
1. The button height should be a little taller to match the current control.
2. The arrow is too large, it should look the same as the current control.

src/Files.App/Views/Settings/AppearancePage.xaml
```diff
@@ -126,28 +126,7 @@
...
-						<Flyout>
-							<toolkit:ColorPicker
-								IsAlphaEnabled="True"
-								IsColorSpectrumVisible="False"
-								Color="{x:Bind ViewModel.AppThemeBackgroundColor, Mode=TwoWay}" />
-						</Flyout>
-					</Button.Flyout>
-				</Button>
+				<toolkit:ColorPickerButton AutomationProperties.Name="{helpers:ResourceString Name=Background}" SelectedColor="{x:Bind ViewModel.AppThemeBackgroundColor, Mode=TwoWay}" />
 				<local:SettingsBlockControl.ExpandableContent>
```
ferrariofilippo, line 130: Is there any way to keep `IsColorSpectrumVisible="False"`?

ferrariofilippo (changes requested):
You changed only one of the two color pickers in `TagsPage`, you can find the other one @ line 82.
~Adding a default value to `ListedTagViewModel.newColor` would prevent *XAML Exceptions* in TagsPage~

Poker-sang:
> You changed only one of the two color pickers in `TagsPage`, you can find the other one @ line 82. Adding a default value to `ListedTagViewModel.newColor` would prevent _XAML Exceptions_ in TagsPage

@ferrariofilippo Sorry, can you provide more information? I had tried but failed, I thought it is bug from CommunityToolkit.

Poker-sang:
> Thank you for the PR, I wasn't aware this control existed. (In the future, it's recommended to open an issue to discuss a change before working on it).
> 
> Overall this looks good but there are two styling issues we should fix before merging
> 
> 1. The button height should be a little taller to match the current control.
> 2. The arrow is too large, it should look the same as the current control.

@yair100 I'm sorry. And because CommunityToolkit.WinUI.Controls is based on WinUI2 (or UWP? I'm not sure), so the appearance is not fluent design.
So I thought we shall use `Microsoft.UI.Xaml.Controls.ColorPicker`, do you think so?

ferrariofilippo:
> @ferrariofilippo Sorry, can you provide more information? I had tried but failed, I thought it is bug from CommunityToolkit.

My bad, I got the exception with a little delay in debug mode and I thought it was fixed

Poker-sang:
> You changed only one of the two color pickers in `TagsPage`, you can find the other one @ line 82. 

Thanks for mentioning. I would fix it after deciding which ColorPicker to use.

Poker-sang:
It looks like we cannot avoid the Binding failure unless using `Microsoft.UI.Xaml.Controls.ColorPicker` instead. Do you think we should do this? @yair100

yaira2:
I went with the toolkit control because it had a nicer UX.

Poker-sang:
emm OK. But we have to wait for toolkit fix the bug.

yaira2:
Is there an issue on the Toolkit repo?

Poker-sang:
@yair100 Which control causes the bug? Is there a relevant issue or details

yaira2:
It seems to be from the ColorPickerButton, changing the theme from one of the built-in options should properly apply the theme but it seems to only work if you choose a color from the color picker flyout.

yaira2:
> I am not sure if ColorPicker in ToolKit has the same bug. I haven't written for so long that I've forgotten.

ColorPicker (from the Toolkit) doesn't have this issue.

Poker-sang:
Exactly. It is a bug appears in ColorPickerButton. When changed  `SelectedColor` property, it won't trigger ColorChanged event. So I suppose to just use ColorPicker. That's good.

Poker-sang:
> ColorPicker (from the Toolkit) doesn't have this issue.

Yes, I just checked it. But they all have binding failure


So I prefer `Microsoft.UI.Xaml.Controls.ColorPicker`

yaira2:
> Microsoft.UI.Xaml.Controls.ColorPicker

The WinUI color picker takes up a lot more space whereas the Toolkits is a lot nicer.

Poker-sang:
Then I think we'll be fine with the current solution.🤔

Poker-sang:
> We'll still need to switch back to ColorPicker to fix this issue.

Sorry I didn't get it. The PR is closed and not merged. So it should already be using ColorPicker by now

yaira2:
Ah you're right, the issue most be from a different change.

PR #11729, opened 2023-03-15
Feature: Improved commands for Pin/Unpin

cinqmilleans:
**Resolved / Related Issues**
Improves commands Pin/UnPin (text, perf, IsExecutable criteria more complete, IsExecutable notify even from FileExplorer).
Fix a modified list exception during its enumeration (when IsExecutable is updated while the favorites list is modified).

   Closes #issue...

**Validation**
- [x] Did you remove any strings from the en-us resource file?
   - [x] Did you search the solution to see if the string is still being used? 


yaira2:
> Are there any other steps that were used to validate these changes?

Can you fill out this section? It helps me know what to test.

cinqmilleans:
There is no visible change for the user, just more complete code (for example if we add a button, its state will always be correct, it was not the case before). There are also hard-to-predict exceptions that have been fixed.

yaira2:
Is there a good way to test the state?

cinqmilleans:
Edit bookmarks in FileExplorer, and see if this is reflected in the context menu. The state is no longer checked when the menu is opened, it is an updated value when it detects a modification.

src/Files.App/Actions/Favorites/PinItemAction.cs
```diff
@@ -3,53 +3,82 @@
...
 		{
-			context.PropertyChanged += Context_PropertyChanged;
+			if (context.HasSelection)
+			{
+				var items = context.SelectedItems.Select(x => x.ItemPath).ToArray();
+				await service.PinToSidebar(items);
+			}
+			else if (context.Folder is not null)
+			{
+				await service.PinToSidebar(context.Folder.ItemPath);
```
hecksmosis, line 45: I think we should remove the curly braces

src/Files.App/Actions/Favorites/UnpinItemAction.cs
```diff
@@ -3,57 +3,80 @@
...
 		{
-			context.PropertyChanged += Context_PropertyChanged;
+			if (context.HasSelection)
+			{
+				var items = context.SelectedItems.Select(x => x.ItemPath).ToArray();
+				await service.UnpinFromSidebar(items);
+			}
+			else if (context.Folder is not null)
+			{
+				await service.UnpinFromSidebar(context.Folder.ItemPath);
```
hecksmosis, line 44: As above

src/Files.App/Actions/Favorites/PinItemAction.cs
cinqmilleans: It's done.

src/Files.App/Actions/Favorites/UnpinItemAction.cs
cinqmilleans: It's done.

src/Files.App/Actions/Favorites/PinItemAction.cs
yaira2: We usually use curly braces if there is an else statement.

hecksmosis: Oh, I wasn't aware of that

src/Files.App/Actions/Favorites/UnpinItemAction.cs
cinqmilleans: I put the style back as before.

src/Files.App/Actions/Favorites/PinItemAction.cs
cinqmilleans: I put the style back as before.

yaira2 (approved):
LGTM

PR #11750, opened 2023-03-17
Feature: Auto update feature of FilesLauncher

### LINKED ISSUE #11746: Feature: Keep FilesLauncher.exe up to date on users devices

Whenever we make changes to the FilesLauncher file, existing users need to toggle the default file manager setting off and on for it to apply. Automatically updating this file will make sure users always have the latest improvements.

### Requirements

- Keep the FilesLauncher.exe up to date when users update Files

### Files Version

v2.4.40

### Windows Version

Windows 11

### Comments

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #11746 

**Validation**
- [X] Did you check for accessibility? You can use Accessibility Insights for this.
   You can try this feature by editing or deleting FilesLaucher.exe.sha256 in the LocalAppData folder.
   A log is output when updated.

**Details**
* The update process runs when the app is launched with the default file manager setting turned on.
* This feature checks the file stored the SHA256 hash value of FilesLauncher.exe, rather than FilesLauncher.exe itself.
* FilesLaucher.exe.sha256 is generated during the build process of FilesLauncher.


src/Files.App/App.xaml.cs
```diff
@@ -226,6 +226,7 @@ private static async Task InitializeAppComponentsAsync()
 			var updateService = Ioc.Default.GetRequiredService<IUpdateService>();
 			await updateService.CheckForUpdates();
 			await updateService.DownloadMandatoryUpdates();
+			await updateService.CheckAndUpdateFilesLauncherAsync();
```
yaira2, line 229: Does this run after startup is complete?

hishitetsu: Called asynchronously during startup.

yaira2: Can we delay it until after startup? It probably doesn't take long but we might as well delay it.

yaira2 (approved):
LGTM

PR #11753, opened 2023-03-17
Feature: RichCommands Paste

cinqmilleans:
**Resolved / Related Issues**
Copy the pr #11670, approved but the checks fail for an unknown reason.
I split the pr, this part is only Cut/Copy. This is the second and final part, the Paste command.

**Validation**


yaira2:
@cinqmilleans can you fix the conflicts?

yaira2 (dismissed):
Thank you

yaira2:
@cinqmilleans do the tests past if you run them locally?

hishitetsu:
I figured out why the test fails.
If you select an item and press the copy button and then the paste button, a conflict dialog appears instead of automatically renaming and copying the item. The test doesn't assume that a dialog will appear.
Also, the paste button cannot be pressed when any file is selected.

cinqmilleans:
Regarding the 2nd difference, it is voluntary. The toolbar buttons should always have the same behavior as in the context menu. If a folder is selected, I can paste into it with the toolbar. Currently, you have to go through its contextual menu or open the folder.

Regarding the dialog box, I corrected that.

cinqmilleans:
The collision issue already exists in the current application.
1) Open a folder containing a folder and a file.
2) Copy the file
3) Open the context menu of the sub-folder (same parent as the copied file).
4) Paste into folder.
5) A collision popup appears. However, there is no real collision.

In this pr, copying in the same folder (deselecting everything), does not display a popup but automatically renames the file.
The problem seems to come from FilesystemHelpers.GetCollision(). That's another issue for another pr, if the issue is confirmed.

hishitetsu:
Sorry, I misunderstood a bit. The test fails because it expects the folder to be pasted in the same location as the original folder.
In File Explorer, the paste button in the toolbar copies the folder in the opened folder, not in the selected folder. It is also available while a file is selected. I think it should work the same as Explorer i.e. current Files.

The behavior of the conflict dialog was certainly not changed.

hishitetsu:
Maybe two commands are needed: paste to current folder and paste to selected folder.

cinqmilleans:
It's done.

hishitetsu:
Thank you. It is much better, but there is still one point where the behavior is different.
When I open a context menu without item selection, the paste button is disabled. It should be enabled and work the same way as the paste button on the toolbar.

cinqmilleans:
You are right. I added this. Thank you.

hishitetsu:
Nice. Everything seems to be working as expected.

PR #11763, opened 2023-03-19
Fix: Fix Issue with unselecting row immediately hiding checkbox

chingucoding:
**Resolved / Related Issues**
- Issue where unselecting a row hides checkbox instead of showing it unselected, supersedes #11715 @cinqmilleans FYI

Current issue is that the row is no longer highlighted when unselecting but I dont know a good workaround for this right now.
**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Views/LayoutModes/GridViewBrowser.xaml
+++ b/src/Files.App/Views/LayoutModes/GridViewBrowser.xaml
@@ -215,7 +215,7 @@
 
 					<VisualStateManager.VisualStateGroups>
 						<VisualStateGroup x:Name="CheckboxVisibilityStates">
-							<VisualState x:Name="Normal" />
+							<VisualState x:Name="HideCheckbox" />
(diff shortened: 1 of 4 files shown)


yaira2 (approved):
LGTM

src/Files.App/Views/LayoutModes/GridViewBrowser.xaml
```diff
@@ -215,7 +215,7 @@
 
 					<VisualStateManager.VisualStateGroups>
 						<VisualStateGroup x:Name="CheckboxVisibilityStates">
-							<VisualState x:Name="Normal" />
+							<VisualState x:Name="HideCheckbox" />
```
yaira2, line 218: Can you do this for the tiles layout as well?

chingucoding: Sure thing

PR #11771, opened 2023-03-19
RichCommand: QuickLook

yaira2:
**Resolved / Related Issues**
   Closes #issue...

**Validation**
   1. Install and open QuickLook
   2. Open Files and select an item
   3. Press the space bar and see if the QuickLook preview opens
   4. Select a different item with the arrow keys or mouse and see if the preview updates
   5. Select multiple items and press the space bar, verify that the items remain selected.

**Screenshots (optional)**
Add screenshots here.


src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
```diff
@@ -170,7 +170,7 @@ protected virtual async void FileList_SelectionChanged(object sender, SelectionC
 			SelectedItems = ListViewBase.SelectedItems.Cast<ListedItem>().Where(x => x is not null).ToList();
 
 			if (SelectedItems.Count == 1 && App.AppModel.IsQuickLookAvailable)
-				await QuickLookHelpers.ToggleQuickLook(ParentShellPageInstance, true);
+				await QuickLookHelpers.ToggleQuickLook(SelectedItem.ItemPath, true);
```
yaira2, line 173: @cinqmilleans how can I trigger the action here?

src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
```diff
@@ -286,14 +286,6 @@ protected override async void FileList_PreviewKeyDown(object sender, KeyRoutedEv
 				FilePropertiesHelpers.ShowProperties(ParentShellPageInstance);
 				e.Handled = true;
 			}
-			else if (e.Key == VirtualKey.Space)
-			{
-				if (!IsRenamingItem && !ParentShellPageInstance.ToolbarViewModel.IsEditModeEnabled)
-				{
-					e.Handled = true;
```
yaira2, line 293: This e.Handled = true is very important, it prevents the selection from being reset if multiple items are selected.

src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
cinqmilleans: Declare the service in the class.
`private readonly ICommandManager commands = Ioc.Default.GetRequiredService<ICommandManager>();`

The method simply becomes this. This code checks that the action is IsExecutable, and executes it if possible.
```
protected virtual async void FileList_SelectionChanged(object sender, SelectionChangedEventArgs e)
{
    await commands.LaunchQuickLook.ExecuteAsync();
}
```

yaira2: Can I pass an argument?

src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
cinqmilleans: I don't know what your problem is. Could you explain.

src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
hecksmosis: No, I think you have to use contexts to get the value that you need

src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
yaira2: It's alright, I worked around this be keeping the code behind.

src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
yaira2: Changing the selection of an item (without pressing the space key) is supposed to toggle QuickLook but with a different command. I guess we'll need a separate action for this?

cinqmilleans: No. IContentPageContext allows you to be notified when the selection changes, and to know the selection and the parent folder. The action can act accordingly.

cinqmilleans:
There is a bug with the Space key. If you select an image file, then click on the search bar, you cannot use Space, it opens QuickLook. I will create another pr to fix this which is related to centralized hotkey management.

yaira2:
> There is a bug with the Space key. If you select an image file, then click on the search bar, you cannot use Space, it opens QuickLook. I will create another pr to fix this which is related to centralized hotkey management.

Does this affect the ctrl + a shortcut as well?

cinqmilleans:
Regarding quicklook. While testing, I had the reflex to try to close it with the Escape key. Maybe add this hotkey to exit. (Command quit/exit, which can close active popup, settings, quicklook, properties,... ). It's another pr but I note the idea.

yaira2:
> Regarding quicklook. While testing, I had the reflex to try to close it with the Escape key. Maybe add this hotkey to exit. (Command quit/exit, which can close active popup, settings, quicklook, properties,... ). It's another pr but I note the idea.

Good idea 👍

PR #11704, opened 2023-03-13
Fix: Fixed issue where delete button would delete file while editing filename

ferrariofilippo:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #11696 

**Validation**
How did you test these changes?
- [x] Built and ran the app
- [ ] Tested the changes for accessibility

PR #11791, opened 2023-03-21
Allow Textbox to use keyboard despite hotkeys

cinqmilleans:
**Resolved / Related Issues**
Some hotkeys should be disabled in textboxes, especially if there are no modifiers. This pr filters the keys that can trigger hotkeys in a textbox.

**Validation**
Test the hotkeys in the search field (ctrl+a, space, home, F11 ...)

### DIFF
--- a/src/Files.App/Commands/HotKeyHelpers.cs
+++ b/src/Files.App/Commands/HotKeyHelpers.cs
@@ -1,0 +1,59 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+using Windows.System;
+
+namespace Files.App.Commands
+{
+	internal static class HotKeyHelpers
+	{
+		private static readonly ImmutableArray<VirtualKey> globalKeys = new List<VirtualKey>
+		{
+			VirtualKey.GoHome,
+			VirtualKey.GoBack,
+			VirtualKey.GoForward,
+			VirtualKey.Application,
+			VirtualKey.Favorites,
+			VirtualKey.Search,
+			VirtualKey.Refresh,
+			VirtualKey.Pause,
+			VirtualKey.Sleep,
+			VirtualKey.Print,
+			VirtualKey.F1,
+			VirtualKey.F2,
+			VirtualKey.F3,
+			VirtualKey.F4,
+			VirtualKey.F5,
+			VirtualKey.F6,
+			VirtualKey.F7,
+			VirtualKey.F8,
+			VirtualKey.F9,
+			VirtualKey.F10,
+			VirtualKey.F11,
+			VirtualKey.F12,
+			VirtualKey.F13,
+			VirtualKey.F14,
+			VirtualKey.F15,
+			VirtualKey.F16,
+			VirtualKey.F17,
+			VirtualKey.F18,
+			VirtualKey.F19,
+			VirtualKey.F20,
+			VirtualKey.F21,
+			VirtualKey.F22,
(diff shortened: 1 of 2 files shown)


src/Files.App/Commands/HotKeyHelpers.cs
```diff
@@ -0,0 +1,59 @@
...
+		private static readonly ImmutableArray<HotKey> textBoxHotKeys = new List<HotKey>
+		{
+			new HotKey(VirtualKey.X, VirtualKeyModifiers.Control), // Cut
+			new HotKey(VirtualKey.C, VirtualKeyModifiers.Control), // Copy
+			new HotKey(VirtualKey.V, VirtualKeyModifiers.Control), // Paste
+			new HotKey(VirtualKey.A, VirtualKeyModifiers.Control), // Select all
+			new HotKey(VirtualKey.Z, VirtualKeyModifiers.Control), // Cancel
+		}.ToImmutableArray();
+
+		public static bool IsGlobalKey(this VirtualKey key) => globalKeys.Contains(key);
```
yaira2, line 56: What's global keys for?

cinqmilleans: These are keys that can be used in direct access throughout the application.  mainly F1 F2 and specific keys on certain keyboards.  Goback favorites

cinqmilleans: If the user assigns a command to the B key without modifiers, it is still necessary to be able to type B in a text field without activating the command.

yaira2: Oh, so global keys are just shortcuts that aren't supported when a text box is active?

yaira2: I'm confused, what happens if I remap F1 to be the shortcut to copy files?

cinqmilleans: It is a global key.  So we can use this hotkeys everywhere even in a textbox.  If a file is selected you go to the search field and press F1 the copy command will be triggered on the file.  It makes sense because F1 is useless in textbox unlike most keys.

src/Files.App/Commands/HotKeyHelpers.cs
```diff
@@ -0,0 +1,59 @@
...
+		}.ToImmutableArray();
+
+		private static readonly ImmutableArray<HotKey> textBoxHotKeys = new List<HotKey>
+		{
+			new HotKey(VirtualKey.X, VirtualKeyModifiers.Control), // Cut
+			new HotKey(VirtualKey.C, VirtualKeyModifiers.Control), // Copy
+			new HotKey(VirtualKey.V, VirtualKeyModifiers.Control), // Paste
+			new HotKey(VirtualKey.A, VirtualKeyModifiers.Control), // Select all
+			new HotKey(VirtualKey.Z, VirtualKeyModifiers.Control), // Cancel
+		}.ToImmutableArray();
```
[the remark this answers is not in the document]
cinqmilleans: Keys without modifiers are all captured by the textbox (except some function keys listed in GlobalKeys). So Delete backspace and space are captured by the textbox and won't trigger the corresponding action.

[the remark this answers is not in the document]
ferrariofilippo: Actually they are caught #11704

Issue #11800, opened 2023-03-22
Bug: Keyboard shortcuts aren't working properly


yaira2:
### Description

Keyboard shortcuts like `delete` don't work when you first select an item, requiring users to press the keyboard shortcut a second time.

### Steps To Reproduce

1. Navigate between a couple of different folders
2. Select an item and press delete
3. See if the delete dialog shows
4. Press delete again and see that the dialog shows this time
5. Click on a blank space to deselect the file
6. Select the file again and see that the issue is back

### Requirements

- Fix issue where shortcuts don't work

### Files Version

v2.4.55

### Windows Version

Windows 11

### Log file

NA

ferrariofilippo:
~@yair100 that's my fault (related to #11743 ). The underlying issue is that `OnPreviewKeyUp` is not always fired.~
~(Lines causing the issue are 259 -> 264 of `MainPage.xaml.cs`)~
~We have two ways to fix this, I'm not sure which one is the best:~
~- we can handle it as we do when renaming items (`keyReleased = (GetKeyState((int)e.Key) & KEY_DOWN_MASK) == 0;`~
~- we can filter which command we want to be executed only once (which I think is the best). We can achieve this as follows `keyReleased = e.Key is not VirtualKey.T or VirtualKey.P or ...;`~

~Ignore this for the moment, the bug may be related to something else. I've been testing it for an hour and I can't find a real pattern to reproduce it consistently. Even with your `Steps to reproduce` sometimes it works and sometimes it doesn't~

The issue seems to be that `currentModifiers` sometimes is set even when no modifier is pressed. In my case this happens when I use `Alt + Tab` to switch window. `Menu` is set when you hit `Alt` but it isn't unset when you release it because you are on a different window.

@cinqmilleans I suggest evaluating `Modifiers` status (or at least `Menu`) just before we execute the command. We may use `GetKeyState(int key)` (User32.dll) to do that. What do you think?

yaira2:
This is still an issue

hishitetsu:
Reproduce by the following steps:
1. Select an item.
2. Press Delete key.
3. You will see a dialog. Press Cancel.
4. Deselect the item.
5. Select an item again.
6. Press Delete key.
7. The dialog will not appear this time.

ferrariofilippo:
The problem is that `Delete` opens a modal and because of that `OnPreviewKeyUp` is called on that modal and not on the MainPage. This issue should involve all the actions that open modals.
[...]

PR #11805, opened 2023-03-22
KeyModifiers read only when necessary.

cinqmilleans:
**Resolved / Related Issues**
Our way of determining KeyModifiers is not always correct. This pr determines KeyModifiers when needed, based on the state of system-provided keys, instead of updating it manually in real time.

I also removed an unused EventArgs.

**Validation**
Test the hotkeys


ferrariofilippo:
> Does this fix #11800?

Yes

PR #11811, opened 2023-03-22
Fix: Fixed issue where progress was not updated when creating an archive 

ferrariofilippo:
**Resolved / Related Issues**
- [x] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers..

**Archive Name**
- Operation banner was not reporting the name of the archive being created. This PR fixes that issue

**Archive Progress**
- This PR doesn't completely resolve the issue (I'll try to open a different PR in SevenZip repo to fix that). 
- If there are at least two folders, progress will be shown correctly (4 folders: 0% -> 25% -> 50% -> 75% -> Completed)

**Validation**
- [ x Did you implement any design changes to an existing feature?
   1. Select at least two folders
   2. Create an archive
   3. See that progress is now being reported


src/Files.App/Filesystem/Archive/ArchiveCreator.cs
```diff
@@ -11,8 +11,15 @@ namespace Files.App.Filesystem.Archive
 {
 	public class ArchiveCreator : IArchiveCreator
 	{
+		private int itemsAmount = 1;
```
yaira2, line 14: What's this for?

ferrariofilippo: It's needed to calculate a weighted percentage of progress. If you compress multiple folders, each one will report its own progress. Using `itemsAmount` we can report a realistic progress.
To get the actual progress, I use this formula: 

    [oldPercentage + (percentageDelta / itemsAmount)] 

which means that if we are compressing 2 items and we have compressed 50% of the first one, the total progress will be:
   
    [0% + (50% / 2)] = 25%

yaira2: Can you add a comment to the code?

yaira2 (approved):
LGTM

PR #11813, opened 2023-03-23
Bug: Fixed issue where Rotate Image & Set as Wallpaper were sometimes not displayed

### LINKED ISSUE #11810: Bug: IsExecutable for Set as Background action is broken
### Description

The IsExecutable state isn't updating properly for the Set as Background action. It looks like this issue affect the rotate image actions as well.

### Steps To Reproduce

1. Navigate to a folder with images as well as other file types
2. Click any file that isn't a supported image
3. Click an image file and see how the rotate toolbar buttons are disabled
4. See that the Set as Background toolbar button isn't even visible

### Requirements

- Fix IsExecutable state for the rotate buttons
- Fix visibility of the Set as Background button

### Files Version

v2.4.55

### Windows Version

Windows 11

### Log file

NA

mafra99:
**Resolved / Related Issues**
   Closes #11810

**Validation**
   1. Rotate & Select As Background Image are hidden if no item is selected
   2. Rotate & Select As Background Image are hidden if a non image file is selected
   3. Rotate & Select As Background Image are visible if an image is selected
   4. Set as desktop slideshow is visible if multiple images are selected


yaira2:
> It might be better if the ViewModel would update the state itself so you don't have to call the method CheckAllFileExtensions in all three Actions but I couldn't find an easy way to implement this

Once we're finished with implementing the actions, we can look into improving the way they access the context as well as the code it uses to execute the logic.

ferrariofilippo:
> It might be better if the ViewModel would update the state itself so you don't have to call the method `CheckAllFileExtensions` in all three Actions but I couldn't find an easy way to implement this

What about an abstract parent `BaseSetAsAction` so that we can share the code?

ferrariofilippo:
> This does not change the fact that the method `CheckAllFileExtensions` is called in each action. But it is a good idea to avoid the duplicate code. I will create the abstract class later

Saving the result in a static property?

src/Files.App/Actions/Content/ImageEdition/RotateRightAction.cs
```diff
@@ -1,65 +1,21 @@
...
-		public bool IsExecutable => IsContextPageTypeAdaptedToCommand()
-						&& (context.ShellPage?.SlimContentPage?.SelectedItemsPropertiesViewModel?.IsSelectedItemImage ?? false);
-
-		public RotateRightAction()
-		{
-			context.PropertyChanged += Context_PropertyChanged;
-		}
-
-		public async Task ExecuteAsync()
+		public RotateRightAction() : base()
```
d2dyno1, line 17: No need to explicitly call `base()`

yaira2 (dismissed):
LGTM

hishitetsu:
Please rename this too.
https://github.com/files-community/Files/blob/344a8901eda766e33d2669a56a7f9d19b55e7cdf/src/Files.App/Commands/CommandCodes.cs#L80-L82

yaira2 (approved):
LGTM

Issue #11786, opened 2023-03-21
Feature: Add a description to each `IAction`


hishitetsu:
### What feature or improvement do you think would benefit Files?

I think a description of what the action does would be helpful when implementing the command palette or shortcut customization features.

### Requirements

- `IAction` must have a property named `Description`
- `Description` property of each action must have a description of what the action does

### Files Version

2.4.54.0

### Windows Version

Windows 11 22H2

yaira2:
> Also it would be nice to have an alternate label, because some commands have the same label, Paste for example, and are indistinguishable when listed by labels.

We'll only show commands for the context so I don't think this will be an issue.

hishitetsu:
I am wondering if alternate labels would be helpful to list shortcuts like this.

| Key | Action | Description |
| --- | ------- | ------------ |
| Ctrl+C | Copy | Copy items to clipboard |
| Ctrl+Shift+C | Copy path | Copy path of item to clipboard |
| Ctrl+X | Cut | Cut items to clipboard |
| Ctrl+V | Paste | Paste items from clipboard to current folder |
| Ctrl+Shift+V | Paste to selected | Paste items from clipboard to selected folder |

yaira2:
Isn't the description enough

hishitetsu:
Maybe so. I'll change requirements to the description only.

cinqmilleans:
Here is what I envision.
* Add optionnel Description, for tooltip.
* Add Name in addition to Label. Name is immutable (but localized). Name is the default value of Label if undefined.
* The label may change (for example, (un)compressing actions change it to include the folder depending on the selection.)
This is the name that will be displayed in the command palette.
* Add Category (immutable). Some actions have similar names. For example "Ascending" can apply to sort or group.
This category will be an enum, containing the value None. If there is a category, it will be displayed with a different style before ordering. This helps to facilitate the search filter and to group together similar commands.

We will have for example
[bold]Sort By[/bold] Name
[bold]Sort By[/bold] Size
[bold]Group By[/bold] Name
[bold]Group By[/bold] Size
[bold]Selection[/bold] Select all
[bold]Selection[/bold] Invert selection
[bold]Selection[/bold] Clear selection

yaira2:
@cinqmilleans the descriptions have been added,

PR #11815, opened 2023-03-23
Feature: Added `Description` property to `IAction`

hishitetsu:
**Resolved / Related Issues**

**Validation**

**Notes**
* `Description` property _must_ be implemented in each action because actions that cannot be described should not exist. If we want to show or not show the description depending on the action or context, we should have a separate flag.
* For now, all descriptions other than copy, cut, and paste are placeholders. These should be replaced with appropriate descriptions before any additional features related to the command can be implemented. #11786 will be closed when all descriptions are implemented.

### DIFF
--- a/src/Files.App/Actions/IAction.cs
+++ b/src/Files.App/Actions/IAction.cs
@@ -21,15 +21,15 @@ namespace Files.App.Actions
 		RichGlyph Glyph => RichGlyph.None;
 
 		/// <summary>
-		/// Primary hotkey to execute actions
+		/// Primary hotkey to execute the action
 		/// </summary>
 		HotKey HotKey => HotKey.None;
 		/// <summary>
-		/// Secondary hotkey to execute actions
+		/// Secondary hotkey to execute the action
 		/// </summary>
 		HotKey SecondHotKey => HotKey.None;
 		/// <summary>
-		/// Tertiary hotkey to execute actions
+		/// Tertiary hotkey to execute the action
 		/// </summary>
 		HotKey ThirdHotKey => HotKey.None;
 		/// <summary>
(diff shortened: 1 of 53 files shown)


src/Files.App/Actions/IAction.cs
```diff
@@ -5,17 +5,47 @@ namespace Files.App.Actions
 {
 	public interface IAction
 	{
+		/// <summary>
+		/// A label for display in context menus and toolbars
```
yaira2, line 9: It can be displayed in the command palette as well.

src/Files.App/Actions/IAction.cs
```diff
@@ -5,17 +5,47 @@ namespace Files.App.Actions
...
+		/// Glyph information to display icon
+		/// </summary>
 		RichGlyph Glyph => RichGlyph.None;
 
+		/// <summary>
+		/// A hotkey to execute actions
+		/// </summary>
 		HotKey HotKey => HotKey.None;
+		/// <summary>
+		/// Another hotkey to execute actions
```
yaira2, line 28:
```suggestion
		/// Secondary hotkey to execute actions
```

src/Files.App/Actions/IAction.cs
hishitetsu: I rather think that the command palette should display the description.
image

yaira2: 👍

yaira2 (approved):
LGTM

I think we should create the descriptions in a spreadsheet or table so that they can be reviewed side by side before we add them to the code.

hishitetsu:
> I think we should create the descriptions in a spreadsheet or table so that they can be reviewed side by side before we add them to the code.

In specs directory?
https://github.com/files-community/Files/tree/main/specs

hishitetsu:
OK. I'll make a table of actions.

PR #11818, opened 2023-03-23
Fix: Fixed issue where new tab icon disappears when window is smaller than 640px

### LINKED ISSUE #11797: Bug: New tab icon disappears when window is smaller than 640 pixels
### Description

New tab icon disappears when window is smaller than 640 pixels
### Steps To Reproduce

1. Open Files
2. Reduce Files to a small window size and noticed the new tab button hides

### Requirements

Being able to click on (+) to create a new tab even when Files is not in full screen.

### Files Version

2.4.40.0

### Windows Version

10.0.22621.1413

### Log file
[debug.log](https://github.com/files-community/Files/files/11034318/debug.log)

ferrariofilippo:
**Resolved / Related Issues**
   Closes #11797 

**Changes**
- Now `New Tab` button is always visible
- Fixes a bug where you couldn't close tabs (using `x` button) when window was smaller than 640px

**Validation**
   1. **NOTE** To test changes you have to build Files as release or comment out line 128 of `DragZoneHelper.cs` (XAML toolbar may interfere with taps)
   2. Resize the window so that it it smaller than 640px
   3. Try to create a tab using the `+` button

### DIFF
--- a/src/Files.App/Actions/Navigation/NewTabAction.cs
+++ b/src/Files.App/Actions/Navigation/NewTabAction.cs
@@ -12,6 +12,8 @@ namespace Files.App.Actions
 
 		public HotKey HotKey { get; } = new(VirtualKey.T, VirtualKeyModifiers.Control);
 
+		public RichGlyph Glyph { get; } = new("\uE710");
+
 		public Task ExecuteAsync() => MainPageViewModel.AddNewTabAsync();
 	}
 }
--- a/src/Files.App/Actions/Navigation/NewTabAction.cs
+++ b/src/Files.App/Actions/Navigation/NewTabAction.cs
@@ -10,6 +10,8 @@ namespace Files.App.Actions
 	{
 		public string Label { get; } = "NewTab".GetLocalizedResource();
 
+		public string Description => "TODO: Need to be described.";
+
 		public HotKey HotKey { get; } = new(VirtualKey.T, VirtualKeyModifiers.Control);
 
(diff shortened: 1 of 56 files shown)


src/Files.App/Actions/Navigation/NewTabAction.cs
```diff
@@ -12,6 +12,8 @@ internal class NewTabAction : IAction
 
 		public HotKey HotKey { get; } = new(VirtualKey.T, VirtualKeyModifiers.Control);
 
+		public RichGlyph Glyph { get; } = new("\uE710");
```
yaira2, line 15: Is this the right icon?

ferrariofilippo: It's the same we were using before.

yaira2: We had a custom icon which was removed when I got rid of the custom icon font.

ferrariofilippo: How can I get it?

yaira2: It's not available anymore, I think we should remove this for now.

ferrariofilippo: That would remove the icon from the `TabView` control as well. I'll remove it only from the context menu, if this is what you mean

yaira2 (approved):
Thank you

Issue #6085, opened 2021-09-08
Add Toolbar button to show context menu (of selected file[s])


DrPepperBianco:
### What's the Problem?

Using a touch screen I always missed the possibility to open context menus easily

### Solution/Idea

In the toolbar, where Copy, Cut, Paste, Rename, and Delete is located, there could be another button, that just opens the context menu of the selected file(s).

When using a touch screen, you have to "hold" a touch click to open a context menu. I always find this a not perfect solution. Such button would make life easier for touch users like me.

### Alternatives

In MS Office, if you select text (or a cell etc.), next to the selection an automatic short context menu appears. Maybe this wouldn't be a bad idea.

Like, if the file is selected with touch, the context menu (or a short context menu, with expand button or so) appears automatically next to it.

If we use this approach, the context menu should (like in office) be not so prominent, like opacity<1-ish. Also, the context menu shouldn't prevent the user from doing other stuff (like selecting the next file etc.)

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow touch users to open context menu without long click. | Must |
| This proposal should not make the work experience of users without touch devices worse! | Should |

### Files Version

v2.0

### Windows Version

Windows 10 Pro - 21H1 - 19043.1165

### Additional comment

Side note: Windows 8 had a very practical solution for this. When Windows 8 intially added Tiles in their start screen, you could scroll sideways. But if you swiped the tiles up and down you could select/unselect the tiles. In Windows 8 that also meant that the "context menu" (in Windows 8 it was a bar that appeared at the bottom of the screen) automatically appeared, when a Tile was selected.

It was not very "beautiful" but was actually very practical.

When in Windows 10 the whole convention shifted from sidescroller to vertical scroller again, this method of selecting tiles was dropped. From then on it was harder to select tiles.

Anyway, in the "Desktop World" (like classical Explorer) it was always the same (at least from Windows 7 onwards), long click/tap to open the context menu.

Depending on the circumstances it can be troublesome to do this long tap.

josh-65:
I like this idea but the alternative mini context menu is quite annoying for me. I haven't used it in Office but in Edge. Its always gotten in the way for me.

winston-de:
> In MS Office, if you select text (or a cell etc.), next to the selection an automatic short context menu appears. Maybe this wouldn't be a bad idea.
> 
> Like, if the file is selected with touch, the context menu (or a short context menu, with expand button or so) appears automatically next to it.
> 
[...]

PR #11839, opened 2023-03-25
Command OpenContextMenu 

cinqmilleans:
**Resolved / Related Issues**
 * New command OpenContextMenu for open the context menu for the active element (item of layout, tab, ...).
 * Use the hotkey Ctrl+Enter.
 * Works for any interface element that has a ContextFlyout.
 * The homepage and sidebar elements don't have a ContextFlyout because the menu is generated by the RightTapped event. The action contains specific code for these 2 cases. To use this command on these elements, it is necessary to select them with the keyboard, and it is difficult to do. A pr to facilitate keyboard selection of homepage and sidebar elements is in preparation.
 * There is an incompatibility between hotkeys and flyouts. When a flyout is opened MainPage events are no longer fired. When you open the flyout with a hotkey, the keyreleased state is not modified when you release the key. MainPage contains a specific line of code to fix this. However, this problem may arise for other commands (OpenSettings, NewFolder, ...). A more permanent solution must be provided.

**Validation**
Test the hotkey Ctrl+Enter on item with contextmenu (selected by keyboard or mouse).


yaira2:
I'm not convinced we need this, most keyboards have a menu key built for this purpose.

cinqmilleans:
* This key is not present on my laptop, yet recent and professional. It is only available on my usb keyboard.
* This does not work in the sidebar and the homepage, because the menus are created dynamically by the RightTapped event, impossible to simulate on the keyboard (to my knowledge).
* The automatic placement by this extra key is strange, especially in the Detail view. The menu appears very to the right instead of being positioned at the beginning.
* The command will allow to have a personalized hotkey, including a single key. It is practical in case of handicap.

src/Files.App/Actions/Open/OpenContextMenuAction.cs
```diff
@@ -0,0 +1,79 @@
...
+			var position = new Point(point.X + 16, point.Y + 16);
+
+			if (element.DataContext is INavigationControlItem)
+			{
+				var sidebar = element.FindAscendant<SidebarControl>();
+				if (sidebar is not null)
+				{
+					await sidebar.OpenContextMenuAsync(element, position);
+					return;
+				}
```
ferrariofilippo, line 57: You can remove the `return` here and also remove the brackets

src/Files.App/Actions/Open/OpenContextMenuAction.cs
```diff
@@ -0,0 +1,79 @@
...
+			{
+				if (App.LastOpenedFlyout is CommandBarFlyout { IsOpen: true })
+					return false;
+
+				var element = GetFocusedElement();
+				if (element is null)
+					return false;
+
+				return element.ContextFlyout is FlyoutBase { IsOpen: false }
+					|| element.DataContext is INavigationControlItem or WidgetCardItem;
```
ferrariofilippo, line 37:
```suggestion
				return element.ContextFlyout is FlyoutBase { IsOpen: false } ||
					element.DataContext is INavigationControlItem or WidgetCardItem;
```

ferrariofilippo (commented):
At the moment this shortcut is not working for tags widget and recent files. I think there's a conflict with another shortcut: pressing `Ctrl + Enter` will open items.

Also just got this exception when trying to open sidebar context menu (doesn't always happen)
[debug.log](https://github.com/files-community/Files/files/11071584/debug.log)

src/Files.App/Actions/Open/OpenContextMenuAction.cs
cinqmilleans: I don't understand this change.

ferrariofilippo: In most of the codebase we align `&&`s and `||`s on the previous line

src/Files.App/Actions/Open/OpenContextMenuAction.cs
cinqmilleans: It's done but I left the brackets, in accordance with the required style.

cinqmilleans:
The fix for KeyReleased has been removed as it is no longer needed, thanks to a bug fix.

cinqmilleans:
I fixed this command for RecentItems. There are still the tags to manage but it will be done in a day or two. Do not commit this pr.

ferrariofilippo:
@cinqmilleans I was looking through the opened issues and I found this one #6085. You may consider implementing it

yaira2:
@ferrariofilippo there are some pretty good alternatives in that issue so we're probably not going to implement that feature as originally requested.

src/Files.App/Actions/Open/OpenContextMenuAction.cs
```diff
@@ -0,0 +1,86 @@
...
+using Windows.System;
+
+namespace Files.App.Actions
+{
+	internal class OpenContextMenuAction : IAction
+	{
+		public string Label { get; } = "OpenContextMenu".GetLocalizedResource();
+		public string Description => "TODO: Need to be described.";
+
+		public HotKey HotKey { get; } = new(VirtualKey.Enter, VirtualKeyModifiers.Control);
```
yaira2, line 22: Can you set the default to the `menu` key?

cinqmilleans: It's done.

cinqmilleans:
The command now works in widgets. The hotkey is Menu+Enter.

src/Files.App/Actions/Open/OpenContextMenuAction.cs
```diff
@@ -0,0 +1,96 @@
...
+using Windows.System;
+
+namespace Files.App.Actions
+{
+	internal class OpenContextMenuAction : IAction
+	{
+		public string Label { get; } = "OpenContextMenu".GetLocalizedResource();
+		public string Description => "TODO: Need to be described.";
+
+		public HotKey HotKey { get; } = new(VirtualKey.Enter, VirtualKeyModifiers.Menu);
```
yaira2, line 23: Can you remove the enter key?

yaira2: > Perhaps you mean the key that displays the menu

This is what I meant, and when we add support for remapping you'll be able to switch to Ctrl+Enter but the default should be the regular menu/application key.

cinqmilleans: This key is not present on all keyboards. Why not use the 2 default hotkeys ?

yaira2: Do you know of any other apps using Ctrl+Enter for this?

cinqmilleans: None. This is an opportunity to stand out.

yaira2: Can you add it to the list of shortcuts on the website?

cinqmilleans: I don't know enough node.js to do this quickly. Surely someone else can do this. Currently I am preparing a pr to modify hotkeys and store custom hotkeys in a litedatabase.

ferrariofilippo: Actually you have only to edit a markdown file (See https://github.com/files-community/Website/pull/413 for reference)

yaira2: > Currently I am preparing a pr to modify hotkeys and store custom hotkeys in a litedatabase.

Wouldn't json be better?

yaira2: This is out of scope for this pull request, the main thing is to update the file on the website documenation to include the shortcut.

cinqmilleans: I put the shortcuts Ctrl+Enter and the Application key.

cinqmilleans:
Merge fixed.

PR #11860, opened 2023-03-27
Localized Hotkeys

cinqmilleans:
**Resolved / Related Issues**
* Hotkey texts are localizable. The keys were translated before using richcommands. This makes it possible not to have to translate the same key to each hotkey that uses it.
* Available keys are limited to a whitelist. This ensures that you always have a label available. This label is prettier than a simple transposition of the enum.
* Removed unused Hotkey.Parse. He was too complex with localization. It is unlikely to be useful as hotkey storage will need to be language insensitive.
* Some keys have a more readable label (XButton1 -> Mouse4).

**Validation**


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2667,4 +2667,142 @@
   <data name="ExitCompactOverlayDescription" xml:space="preserve">
     <value>Exit compact overlay</value>
   </data>
+  <data name="Key.Menu" xml:space="preserve">
+    <value>Menu</value>
```
yaira2, line 2671: Can you add a description to the hotkeys? The same description can be used for all of them.

src/Files.App/Commands/HotKeyHelpers.cs
```diff
@@ -1,13 +1,146 @@
...
+			[VirtualKey.X] = "X",
+			[VirtualKey.Y] = "Y",
+			[VirtualKey.Z] = "Z",
+			[VirtualKey.Add] = "+",
+			[VirtualKey.Subtract] = "-",
+			[VirtualKey.Multiply] = "*",
+			[VirtualKey.Divide] = "/",
+			[(VirtualKey)188] = ",",
+			[(VirtualKey)190] = ".",
+			[(VirtualKey)192] = "`",
```
ferrariofilippo, line 122: Shouldn't we localize them as well?
In Italian `(VirtualKey)192` is `ò`, it would make more sense if that was the key displayed

cinqmilleans: @ferrariofilippo @yair100 These particular keys (oem) must not be localized because they depend on the user layout and not on his language. You can have an Italian keyboard and use Windows and Files in English for example. I found a piece of code that returns the key used for these oems depending on the layout used. I will update the code.

yaira2: General guidance from Microsoft is not to translate hot keys. We don't have to follow this guidance but something to keep in mind.

cinqmilleans: @ferrariofilippo It's done, You should see ò now. It is not displayed, it will be necessary to display Commands.OpenTerminal.LabelWithHotkey somewhere in the application to test.

ferrariofilippo: I'll test it asap

src/Files.App/Strings/en-US/Resources.resw
cinqmilleans: It's done.

src/Files.App/Commands/HotKeyHelpers.cs
ferrariofilippo: > @ferrariofilippo It's done, You should see ò now. It is not displayed, it will be necessary to display Commands.OpenTerminal.LabelWithHotkey somewhere in the application to test.

Everything works fine. Thank you for your work 👍

ferrariofilippo (dismissed):
LGTM

Jay-o-Way:
Totally off-topic, but let's re-think the use of words like whitelist/blacklist being synonims for good/bad 😇

yaira2 (approved):
LGTM

PR #11871, opened 2023-03-28
Fixed: Opening file with enter key resets the selected file

### LINKED ISSUE #11870: Bug: Opening file with enter key resets the selected file
### Description

Using the enter key to open a file changes the selected item in the list.

### Steps To Reproduce

1. Select the last file in a list
2. Use the enter key to open it
3. Observe that the item loses its selection and the first item in the list is selected instead

### Requirements

- Selection shouldn't change when opening a file

### Files Version

v2.4.57

### Windows Version

Windows 11

### Log file

NA

ferrariofilippo:
**Resolved / Related Issues**
   Closes #11870

**Validation**
   1. Select the last file in a folder (DetailsLayout - other layouts were already working)
   2. Hit enter


hishitetsu:
I was working on the same issue.
This fix is simpler and does not risk regression, but I think it would be better to use RichCommand.

PR #11890, opened 2023-03-29
Feature: Translate shell folders names & avoid possible bugs

ferrariofilippo:
**Resolved / Related Issues**
   Closes N/A

**Change List**
- Now Shell Folders have a tooltip in the sidebar (Before it was empty)
- Path box will show `This PC` instead of the whole Shell folder path
- Prevent paths such as `RecycleBin/..` which will probably cause issues
- Localize Shell Folders names in `JumpList`, `TabName`, `Path box`
- Removed `LocalSettings` translation of `RecycleBin`

**Validation**
   1. Open app
   2. Navigate to shell folders

### DIFF
--- a/src/Files.App/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
+++ b/src/Files.App/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
@@ -399,8 +399,10 @@ namespace Files.App.Filesystem
 		{
 			if (shPath.StartsWith(CommonPaths.RecycleBinPath, StringComparison.OrdinalIgnoreCase))
 				return CommonPaths.RecycleBinPath;
+
 			if (shPath.StartsWith(CommonPaths.MyComputerPath, StringComparison.OrdinalIgnoreCase))
 				return CommonPaths.MyComputerPath;
+
 			if (shPath.StartsWith(CommonPaths.NetworkFolderPath, StringComparison.OrdinalIgnoreCase))
 				return CommonPaths.NetworkFolderPath;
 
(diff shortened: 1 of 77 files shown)


src/Files.App/Filesystem/StorageFileHelpers/StorageFileExtensions.cs
```diff
@@ -389,6 +395,20 @@ private static string ResolvePath(string path, bool isFtp)
 			return pathBuilder.ToString();
 		}
 
+		private static string ResolveShellPath(string shPath)
```
yaira2, line 398: Should this be a separate helper?

yaira2: Making a separate helper will allow us to reuse the code.

src/Files.App/ViewModels/Settings/PreferencesViewModel.cs
```diff
@@ -605,9 +605,16 @@ public string Text
 				if (Path == "Home")
 					return "Home".GetLocalizedResource();
 
-				return (Path == CommonPaths.RecycleBinPath)
-					   ? ApplicationData.Current.LocalSettings.Values.Get("RecycleBin_Title", "Recycle Bin")
-					   : Path;
+				if (Path.Equals(CommonPaths.RecycleBinPath, StringComparison.OrdinalIgnoreCase))
```
yaira2, line 608: Can you use the helper?

ferrariofilippo: This is actually different from the helper I've just pushed. I'll add another one

yaira2 (approved):
LGTM

Issue #11737, opened 2023-03-16
Feature: Only open single tab if user holds down ctrl + t


yaira2:
### What feature or improvement do you think would benefit Files?

Holding down `ctrl` + `t` will keep opening tabs until the user releases the keys. In order to reduce the risk of crashing the app, we should only open a single tab. Depending on the context, we should do this for other shortcuts as well.

### Requirements

- Only open a single tab when holding down `ctrl` + `t`
- Only close one tab when holding down `ctrl` + `w`
- Only refresh tab once when holding down `ctrl` + `r`
- Only go back once when holding down `alt` + `left`
- Only go forward once when holding down `alt` + `right`

### Files Version

v2.4

### Windows Version

Windows 11

QuaintMako:
I took the liberty to edit the feature tracking so we have a task for each need.
@yair100 the issue should be renamed to better reflect the work needed, something like "Feature: Execute navigation between tabs only once on navigation commands"

yaira2:
Thank you

ferrariofilippo:
> Only close one tab when holding down ctrl + w
> Only refresh tab once when holding down ctrl + r

These two are already implemented

ferrariofilippo:
I'll check but I won't be able to do that until Monday

yaira2:
I haven't seen this issue in a while.

PR #11916, opened 2023-03-31
RichCommands: Switch tabs command

### LINKED ISSUE #6111: Bug: Changing tab with keyboard doesn't update address bar
### Description

When you use a keyboard shortcut to change to the previous or next tab (Ctrl + Shift + Tab or Ctrl + Tab), sometimes it will change the folder contents but not the toolbar and address bar, other times it won't change at all...

In the gif, check for example that the second time changing from home to Music, it will keep the widgets button instead the folder buttons, and the address bar says "Nueva pestaña" instead my Music path.

### Steps To Reproduce

1. Open several tabs with differents folder.
2. Start to change tabs with the keyboard shortcut (Ctrl + Shift + Tab or Ctrl + Tab)
3. Check the address bar and see that sometimes it shows an incorrect address.

NOTE: For some reason apening settings and closing it seems to fix the issue until you open a new tab. 

### Expected behavior

It should show the proper address and toolbar.

### Files Version

V 2.0.12.0

### Windows Version

OS Version: Windows 10 21H1 10.0.19043.1202

### Relevant Assets

GIF 13-09-2021 15-55-22

Log files:
[debug_fulltrust.log](https://github.com/files-community/Files/files/7154784/debug_fulltrust.log)
[debug.log](https://github.com/files-community/Files/files/7154785/debug.log)

ferrariofilippo:
**Resolved / Related Issues**
   Related #11481 
   Closes #6111
   Related #11737 (You can hold down `Ctrl + Tab` to switch more than once)

**Details**
- `Ctrl + Tab` and `Ctrl + Shift + Tab` shortcuts
- Seems to fix #6111 (I can't reproduce it so I think is fixed)
- Should I introduce a common base action to reduce code dupication or is it overkill?
-`Ctrl + W` or `Ctrl + F4` to close current tab

**Validation**
   1. Open some tabs
   2. Swap active tab using shortcuts

### DIFF
--- a/src/Files.App/Actions/Navigation/CloseSelectedTabAction.cs
+++ b/src/Files.App/Actions/Navigation/CloseSelectedTabAction.cs
@@ -23,9 +23,10 @@ namespace Files.App.Actions
 
 		public RichGlyph Glyph { get; } = new();
 
-		public bool IsExecutable => context.Control is not null && 
-									context.TabCount > 0 && 
-									context.CurrentTabItem is not null;
+		public bool IsExecutable =>
+			context.Control is not null && 
+			context.TabCount > 0 && 
+			context.CurrentTabItem is not null;
 
 		public CloseSelectedTabAction()
 		{
--- a/src/Files.App/Actions/Navigation/CloseSelectedTabAction.cs
+++ b/src/Files.App/Actions/Navigation/CloseSelectedTabAction.cs
@@ -1,0 +1,53 @@
+﻿using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.DependencyInjection;
+using Files.App.Commands;
+using Files.App.Contexts;
+using Files.App.Extensions;
+using System.ComponentModel;
+using System.Threading.Tasks;
+using Windows.System;
+
+namespace Files.App.Actions
+{
+	internal class CloseSelectedTabAction : ObservableObject, IAction
+	{
+		private readonly IMultitaskingContext context = Ioc.Default.GetRequiredService<IMultitaskingContext>();
+
+		public string Label { get; } = "CloseTab".GetLocalizedResource();
+
+		public string Description { get; } = "TODO: Need to be described.";
+
+		public HotKey HotKey { get; } = new(VirtualKey.W, VirtualKeyModifiers.Control);
+
+		public HotKey SecondHotKey { get; } = new(VirtualKey.F4, VirtualKeyModifiers.Control);
+
+		public RichGlyph Glyph { get; } = new();
+
+		public bool IsExecutable => context.Control is not null && 
+									context.TabCount > 0 && 
+									context.CurrentTabItem is not null;
+
+		public CloseSelectedTabAction()
+		{
+			context.PropertyChanged += Context_PropertyChanged;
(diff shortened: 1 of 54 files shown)


src/Files.App/Actions/Navigation/SwapTabLeftAction.cs
yaira2, line 1: What about "NextTab" and "PreviousTab"? "Swap" sounds like you're switching the order of tabs.

yaira2:
> (You can hold down Ctrl + Tab to switch more than once)

Apparently Edge only allows it once, do you think we should do the same?

ferrariofilippo:
> Apparently Edge only allows it once, do you think we should do the same?

After some testing I think we should do that too. You can't really predict where you will land if you keep pressing `Ctrl + Tab` (unless we introduce some kind of delay so that you could choose where to stop)

src/Files.App/Actions/Navigation/CloseSelectedTabAction.cs
```diff
@@ -0,0 +1,53 @@
...
+
+		public HotKey HotKey { get; } = new(VirtualKey.W, VirtualKeyModifiers.Control);
+
+		public HotKey SecondHotKey { get; } = new(VirtualKey.F4, VirtualKeyModifiers.Control);
+
+		public RichGlyph Glyph { get; } = new();
+
+		public bool IsExecutable => context.Control is not null && 
+									context.TabCount > 0 && 
+									context.CurrentTabItem is not null;
```
0x5bfa, line 28:
```suggestion
		public bool IsExecutable =>
			context.Control is not null && 
			context.TabCount > 0 && 
			context.CurrentTabItem is not null;
```

src/Files.App/Actions/Navigation/PreviousTabAction.cs
```diff
@@ -0,0 +1,44 @@
...
+			multitaskingContext.PropertyChanged += MultitaskingContext_PropertyChanged;
+		}
+
+		public Task ExecuteAsync()
+		{
+			if (App.AppModel.TabStripSelectedIndex is 0)
+				App.AppModel.TabStripSelectedIndex = multitaskingContext.TabCount - 1;
+			else
+				App.AppModel.TabStripSelectedIndex--;
+			return Task.CompletedTask;
```
0x5bfa, line 35:
```suggestion

			return Task.CompletedTask;
```

Add a new line break before return statement

src/Files.App/Actions/Navigation/CloseSelectedTabAction.cs
```diff
@@ -0,0 +1,53 @@
...
+
+		public CloseSelectedTabAction()
+		{
+			context.PropertyChanged += Context_PropertyChanged;
+		}
+
+		public Task ExecuteAsync()
+		{
+			context.Control!.CloseTab(context.CurrentTabItem);
+			return Task.CompletedTask;
```
0x5bfa, line 38:
```suggestion

			return Task.CompletedTask;
```

Add a new line break before return statement

0x5bfa (commented):
Coding concerns

yaira2 (approved):
LGTM

PR #11851, opened 2023-03-26
Feature: Always apply most current layout

lukeblevins:
**Resolved / Related Issues**
- [x] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
   Closes #801

**Validation**
How did you test these changes?
- [x] Did you build the app and test your changes?
- [x] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [ ] Did you implement any design changes to an existing feature?
   - [ ] Was this change approved?
- [ ] Are there any other steps that were used to validate these changes?
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
N/A

PR #11965, opened 2023-04-04
Fix: Fixed regressions caused by "Always apply most current layout"

hishitetsu:
_Targeted to the servicing branch_

#11851 has caused some regressions. This PR will fix them.
* Fixed issue with layout changes sometimes not working properly
* Fixed "NotAFolder" error when returning to the home page with the back button after changing the layout

**Resolved / Related Issues**
   Closes #issue...

**Validation**
   1. Layout changes should always work correctly. In 2.4.60, layout changes sometimes causes blank screen or no change.
   2. "NotAFolder" error shouldn't occur when returning to the home page with the back button after changing the layout.

### DIFF
--- a/src/Files.App/Views/BaseShellPage.cs
+++ b/src/Files.App/Views/BaseShellPage.cs
@@ -503,7 +503,8 @@ namespace Files.App.Views
 		{
 			foreach (PageStackEntry entry in ItemDisplay.BackStack.ToList())
 			{
-				if (entry.Parameter is NavigationArguments args && args.NavPathParam is not null and not "Home")
+				if (entry.Parameter is NavigationArguments args && 
+					args.NavPathParam is not null and not "Home")
 				{
 					var correctPageType = FolderSettings.GetLayoutType(args.NavPathParam, false);
 					if (!entry.SourcePageType.Equals(correctPageType))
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/BaseShellPage.cs
```diff
@@ -503,9 +503,9 @@ public void ResetNavigationStackLayoutMode()
 		{
 			foreach (PageStackEntry entry in ItemDisplay.BackStack.ToList())
 			{
-				if (entry.Parameter is NavigationArguments args)
+				if (entry.Parameter is NavigationArguments args && args.NavPathParam is not null and not "Home")
```
yaira2, line 506:
```suggestion
				if (entry.Parameter is NavigationArguments args &&
				args.NavPathParam is not null and not "Home")
```

src/Files.App/Views/BaseShellPage.cs
```diff
@@ -518,9 +518,9 @@ public void ResetNavigationStackLayoutMode()
 
 			foreach (PageStackEntry entry in ItemDisplay.ForwardStack.ToList())
 			{
-				if (entry.Parameter is NavigationArguments args)
+				if (entry.Parameter is NavigationArguments args && args.NavPathParam is not null and not "Home")
```
yaira2, line 521:
```suggestion
				if (entry.Parameter is NavigationArguments args &&
				args.NavPathParam is not null and not "Home")
```

yaira2 (approved):
LGTM

PR #11976, opened 2023-04-05
Feature: Keep timestamps when extracting an archive

### LINKED ISSUE #11259: Feature: Keep timestamps when extracting an archive

In Files, timestamps of the file from an extracted archive are set to the current date and time.
But in File Explorer and many other archivers, timestamps are set to the original date and time stored in the archive file.
I think Files should keep timestamps too.

### Requirements

Timestamps are set to the original date and time stored in the archive file, like File Explorer and many other archivers.

### Files Version

Version: 2.4.31.0

### Windows Version

Windows 11 22H2 22621.1105

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**
   Closes #11259

**Notes**
`CreationTime` might be greater than `LastWriteTime` (try to edit a file and then copy it). This doesn't make sense to me, so I think we should adapt the `CeationTime` (Explorer does this too)

**Validation**
   1. Extract an archive
   2. Open extracted files properties


QuaintMako (approved):
LGTM, I couldn't find a way to break it.

PR #11986, opened 2023-04-05
Fix: Fixed issue where column view jumps from left to right

### LINKED ISSUE #11977: Bug: Column view scroll jumps from left to right when using arrow keys for navigation
### Description

annoying contents sway

### Steps To Reproduce

While in columns view, press the up/down arrows to change directory. The contents of the entire Files windows sway left and then right. Does not happen if you click/select folders with the mouse. See the attached gif.

### Requirements

I would expect the arrow keys to behave the same as the mouse in this case. I was about to check the contents of multiple folders in the same root and then I found that I need to do it with the mouse. The window contents sway is annoying and the whole animation takes what is seems like an eternity to complete.

### Files Version

2.4.60.0

### Windows Version

W11 22H2 22621.1413

### Log file

This is a functionality bug. The app itself does not crash.

ferrariofilippo:
**Resolved / Related Issues**
   Closes #11977

**Validation**
   1. Open a folder using columns layout
   2. Navigate using arrow keys


yaira2:
Do the right and left arrow keys still scroll?

ferrariofilippo:
> Do the right and left arrow keys still scroll?

Horizontally yes

yaira2 (approved):
LGTM

PR #11997, opened 2023-04-06
Feature: Add Invalid Archive Name & Invalid Shortcut Location Warning

### LINKED ISSUE #12044: Feature: Add invalid name and path warnings to modals

At the moment, when you try to create a shortcut, if you enter a non existing path, `Create` button will be disabled but no reason is shown.
Thre's a similar issue when creating archives: you can enter whatever name you want (even using forbidden chars), and the `Create` button will always be enabled. If the name contains forbidden chars, the creation of the archive will fail and no reason will be given (only `Couldn't create the archive`)
Finally, there's a warning tip when we you try to create a file with an invalid name. This is a simple text block and looks different from the one used in tag creation. I think we should always use the same style to be consistent

### Requirements

- Add a warning tip when the shortcut path doesn't exist
- Add a warning tip when the archive name isn't valid
- Change the style of the warning tip of the `Rename` dynamic dialog 

### Files Version

2.4.61.0

### Windows Version

Windows 11 Pro 22H2 10.0.22621.1485

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**
   Closes #12044 

**Validation**
   1. Open app and try to create an archive
   2. Clear archive name

**Screenshots (optional)**


yaira2:
Can you link an issue for this?

ferrariofilippo:
I'm having some troubles with `rename dialog` because of an exception, I'll try to find a workaround

src/Files.App/Dialogs/CreateArchiveDialog.xaml
```diff
@@ -29,11 +35,16 @@
 			BorderThickness="1"
 			ColumnSpacing="8"
 			CornerRadius="4">
+			<Grid.RowDefinitions>
+				<RowDefinition Height="Auto" />
+				<RowDefinition Height="Auto" />
+			</Grid.RowDefinitions>
```
yaira2, line 41: Is this still needed?

ferrariofilippo: No, I'll remove it

yaira2 (approved):
LGTM

Issue #3652, opened 2021-02-20
Bug: Launching Files from PowerShell prevents entering additional commands


HydrogenDeuterium:
**Describe the bug**
run files in powershell terminal,the terminal is fixed and could do nothing before close the window.

**To Reproduce**
call files through pwsh(cmd no simailar problem)

**Expected behavior**
terminal open the window,and ready to my next input

**Screenshots**
![image](https://user-images.githubusercontent.com/36532302/108600281-45b85700-73d1-11eb-87e2-87711998e151.png)

**Desktop (please complete the following information):**
 - OS Version: win 10 2004
 - App version: [v0.11.0]1.2.2.0

**Additional context**
when i choose to run with"start-process files",it works as i wished.
another uwp app works so,and the developer asked me to run it with start-process,
i think it's a tiny issue and easy to fix(unforturnly,i know lettle about uwp app develop,otherwise i will fix it myself)

enisn:
As a workaround using launching with 'start' works well

```bash
start files
```

Josh-65:
Is this still an issue?

yaira2:
Yes

Lukiluc29:
I reproduced your action and whether I write "files" or "start files" it opens with version 2.4.80

heftymouse:
I looked into this some more, and it seems it's not an issue exclusive to Files - I can reproduce this on many packaged apps made with a variety of frameworks and languages. I'm still not clear on what exactly is causing this.

hishitetsu:
I found a PowerShell issue on this.
https://github.com/PowerShell/PowerShell/issues/9970

hishitetsu:
As a side effect of Leave app running, if Files is launched from Powershell while it is in the background, the prompt returns immediately.

yaira2:
Considering this issue is with the platform, and the fact that the issue doesn't occur when Files is open in the background, I think we no longer need to track this issue. For further updates, we suggest subscribing to the issue on the PowerShell repo https://github.com/PowerShell/PowerShell/issues/9970.

PR #12005, opened 2023-04-07
Feature: Auto restore tabs if the app crashes

### LINKED ISSUE #5993: Feature: Auto restore tabs if the app crashes
### What's the Problem?

With #5985, it will only remember the tabs of the last opened window, even if you have more than one opened.

### Requirements
Auto restore tabs on first launch after the app crashes

### Files Version

1.9.32.0 / #5985

### Windows Version

### Additional comment

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #5993

**Validation**


QuaintMako:
With the current implementation, does the user's original setting is preserved after the restart?

hishitetsu:
It should because the current implementation is the same mechanism as when changing languages.
I'm currently working on further improvements to automatically restart the app without clicking notifications.

hishitetsu:
Automatic restart is now implemented.
When `Launcher.LaunchUriAsync()` is called, the method somehow doesn't return even though a new app is launched, so I use a timer to force it to end.

src/Files.App/MainWindow.xaml.cs
```diff
@@ -136,10 +136,6 @@ public async Task InitializeApplication(object activatedEventArgs)
 					break;
 
 				case IToastNotificationActivatedEventArgs eventArgsForNotification:
-					if (eventArgsForNotification.Argument == "report")
-					{
-						await Windows.System.Launcher.LaunchUriAsync(new Uri(Constants.GitHub.BugReportUrl));
-					}
 					break;
```
QuaintMako, line 139: Is thetre any need to keep an empty switch case?

ferrariofilippo:
> When `Launcher.LaunchUriAsync()` is called, the method somehow doesn't return even though a new app is launched

It might be related with #3652

hishitetsu:
As it is, if I get an error when I view a particular page, I get endless restarts.

It needs a little more improvement.

hishitetsu:
> > When `Launcher.LaunchUriAsync()` is called, the method somehow doesn't return even though a new app is launched
> 
> It might be related with #3652

But it returns normally when changing language.
Maybe the UI thread is going haywire because it is after the exception occurred.
https://github.com/files-community/Files/blob/29c35b0d0434a28e69d8f6438e68b0752fb70162/src/Files.App/UserControls/RestartControl.xaml.cs#L51-L57

ferrariofilippo:
> Maybe the UI thread is going haywire because it is after the exception occurred.

Yes that might be the case

hishitetsu:
> As it is, if I get an error when I view a particular page, I get endless restarts.

I addressed it. Now ready for review.

src/Files.App/App.xaml.cs
```diff
@@ -448,19 +449,41 @@ private static void AppUnhandledException(Exception ex, bool shouldShowNotificat
...
 			};
 
 			// Create the toast notification
 			var toastNotif = new ToastNotification(toastContent.GetXml());
 
 			// And send the notification
 			ToastNotificationManager.CreateToastNotifier().Show(toastNotif);
+
+			// Restart the app
+			var userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
```
QuaintMako, line 468: This assignation is done several times through the class, why not hoilst it on top of it in `private readonly` field?

QuaintMako:
@hishitetsu may I ask you what was the process you used to provoke the crash of the app to test it myself?

hishitetsu:
> @hishitetsu may I ask you what was the process you used to provoke the crash of the app to test it myself?

I tested that by embedding the code that throws the Exception for debugging purposes.

QuaintMako (approved):
It works as expected for me.

src/Files.App/ServicesImplementation/Settings/PreferencesSettingsService.cs
```diff
@@ -63,6 +63,12 @@ public List<string> LastSessionTabList
 			set => Set(value);
 		}
 
+		public List<string> LastCrashedTabList
```
yaira2, line 66: How is this different from `LastSessionTabList`?

src/Files.App/App.xaml.cs
```diff
@@ -448,19 +449,41 @@ private static void AppUnhandledException(Exception ex, bool shouldShowNotificat
...
+			var lastSessionTabList = userSettingsService.PreferencesSettingsService.LastSessionTabList;
+
+			if (userSettingsService.PreferencesSettingsService.LastCrashedTabList?.SequenceEqual(lastSessionTabList) ?? false)
+			{
+				// Avoid infinite restart loop
```
yaira2, line 473: Are there any known situations where the app can get into a restart loop?

src/Files.App/ServicesImplementation/Settings/PreferencesSettingsService.cs
hishitetsu: `LastSessionTabList` is updated to "Home" tab at startup, presumably in case of a crash without notification. So I added another one that is not updated except during crashes.

src/Files.App/App.xaml.cs
hishitetsu: No. Not all crashes are handled with `AppUnhandledException`, but in which case the `LastSessionTabList` is updated to "Home" so that the crashed page is not displayed again.

yaira2 (dismissed):
LGTM

yaira2:
@hishitetsu can you resolve the merge conflicts?

yaira2 (dismissed):

src/Files.Backend/Services/Settings/IGeneralSettingsService.cs
```diff
@@ -46,6 +46,11 @@ public interface IGeneralSettingsService : IBaseSettingsService, INotifyProperty
 		/// </summary>
 		List<string> LastSessionTabList { get; set; }
 
+		/// <summary>
+		/// A list containing all paths to tabs last crashed.
```
yaira2, line 50:
```suggestion
		/// A list containing paths of the tabs from the previous session that crashed.
```

Issue #9871, opened 2022-08-31
Avoid unnecessary async state machine generation


hez2010:
We have plenty of code like below:

```csharp
async IAsyncOperation<T> Foo()
{
    return await AsyncInfo.Run(async () => 
    {
        // no await in the body at all
        return foo;
    });
}
```

which doesn't actually await anything in an async method. Instead, we should rewrite them to

```csharp
IAsyncOperation<T> Foo()
{
    return Task.FromResult(foo).AsAsyncOperation();
}
```

This can avoid unnecessary async state machine be generated to reduce GC pressure.

QuaintMako:
Linking it the performance issues maybe? Anyone taking up that issue may want to dive deeper into optimization.

yaichenbaum:
> Linking it the performance issues maybe? Anyone taking up that issue may want to dive deeper into optimization.

Good idea

QuaintMako:
I believe there is no longer any `return await AsyncInfo.Run` methods inside the codebase.

Does this issue also concerns any `return AsyncInfo.Run` that does not have any `await` in its instructions?

itsWindows11:
There are still stuff which could be changed in the codebase, might need re-opening this issue?

hez2010:
Can I assume this issue has been resolved?

itsWindows11:
It should be fully resolved after merging the PR, no problem.

PR #12010, opened 2023-04-07
Refactor: Continued work on reducing async state machines

itsWindows11:
**Resolved / Related Issues**

- Continues work on #9871.

**Validation**
   - Try doing any operation with archives, bundles and file operations. This includes extracting archives, adding/renaming bundles, or creating a shortcut.

**Screenshots (optional)**
Add screenshots here.


src/Files.App/Actions/Content/Archives/CompressIntoSevenZipAction.cs
```diff
@@ -37,7 +37,7 @@ public async Task ExecuteAsync()
 				FileFormat = ArchiveFormats.SevenZip,
 			};
 
-			await ArchiveHelpers.CompressArchiveAsync(creator);
```
hez2010, line 43: I'm a bit worry about this kind of changes (removing `async` from a method that actually has `await` statements), because it will omit some frames from the stack trace and lead to poor diagnosability. 
But I would like to hear how other people thinking about it. @yair100

yaira2: @0x5bfa @d2dyno1

0x5bfa: > omit some frames from the stack trace and lead to poor diagnostics.

I’m not sure how that’ll affect. The await should be attached when that async method have one or more try statement(s). Unless it, I believe we are able to remove await.

yaira2: @hez2010 how would you go about this?

yaira2: 👍

itsWindows11:
Will do in a few hours, I'm busy right now.

src/Files.App/Actions/Content/Archives/CompressIntoSevenZipAction.cs
itsWindows11: I don't think changing that will affect much in terms of diagnosability, any crash or issue would most likely be in `ArchiveHelpers.CompressArchiveAsync`.

itsWindows11:
@yair100 Should be done now, feel free to merge.

yaira2:
@itsWindows11 thank you! Can you fill out the section with steps to test these changes? It should include any areas that might be affected.

itsWindows11:
There are so many ways these changes could be tested, but I'll try to fill in some now.

itsWindows11:
Just finished with filling in the steps (it's not really steps anyway but it involves testing different features).

src/Files.App/App.xaml.cs
```diff
@@ -117,14 +117,14 @@ private static Task StartAppCenter()
...
-		private static async Task InitializeAppComponentsAsync()
+		private static Task InitializeAppComponentsAsync()
 		{
 			var userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
 			var addItemService = Ioc.Default.GetRequiredService<IAddItemService>();
 			var generalSettingsService = userSettingsService.GeneralSettingsService;
 
 			// Start off a list of tasks we need to run before we can continue startup
-			await Task.Run(async () =>
+			return Task.Run(async () =>
```
hez2010, line 127: I don't think this behavior is parity with the existing one. I remembered there used to be a crash related this kind of change (the Update process must happen in the main thread), so can you revert the change here?
We may revisit here later.

src/Files.App/Dialogs/AddItemDialog.xaml.cs
```diff
@@ -27,7 +27,7 @@ public AddItemDialog()
 			InitializeComponent();
 		}
 
-		public new async Task<DialogResult> ShowAsync() => (DialogResult)await base.ShowAsync();
+		public new Task<DialogResult> ShowAsync() => base.ShowAsync().AsTask().ContinueWith(t => (DialogResult)t.Result);
```
hez2010: I don't like this kind of change (i.e. `ContinueWith` instead of chained `await`) as it reduces the readability a lot. There're also many such changes in other places through this PR.
/cc: @yair100

src/Files.App/Shell/ContextMenu.cs
```diff
@@ -163,17 +165,16 @@ public async Task InvokeItem(int itemID)
 			}
 		}
 
-		public static async Task WarmUpQueryContextMenuAsync()
+		public static Task WarmUpQueryContextMenuAsync()
 		{
-			var thread = new ThreadWithMessageQueue();
-			await thread.PostMethod(() =>
+			using var thread = new ThreadWithMessageQueue();
+			return thread.PostMethod(() =>
```
hez2010, line 171: This will cause the `thread` being disposed early unexpectedly.

hez2010 (changes requested):
Otherwise LGTM.

src/Files.App/App.xaml.cs
itsWindows11: Just reverted the changes here.

src/Files.App/ViewModels/Settings/AboutViewModel.cs
```diff
@@ -100,10 +100,10 @@ public void CopyWindowsVersion()
...
 		}
 
-		public async Task SupportUs()
-		{
-			await Launcher.LaunchUriAsync(new Uri(Constants.GitHub.SupportUsUrl));
-		}
+		public Task SupportUs()
+        {
+            return Launcher.LaunchUriAsync(new Uri(Constants.GitHub.SupportUsUrl)).AsTask();
+        }
```
hez2010, line 106: Indents (spaces vs tabs)

hez2010 (dismissed):
LGTM. Thank you!

src/Files.App/ViewModels/Settings/AboutViewModel.cs
yaira2: Fixed

yair100 (approved):
Thank you

PR #12025, opened 2023-04-09
Codebase: Add AppEnvironment enum

### LINKED ISSUE #11959: Code Quality: Add Environment Enum AppModel
**Description**
There are a number of different environments that Files might be running under.
- WindowsDev
- WindowsSideload
- WindowsStore
- WindowsSideloadPreview

The app icon, package identity, update server and AppCenter token can differ between each environment. Be able to check the environment once and save the state in AppModal will simplify the code base and allow us to add specific logic for different environments.

hecksmosis:
**Resolved / Related Issues**
   Closes #11959
- (probably) needs more usages

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.


lukeblevins (changes requested):
We should be using the [IHostEnvironment.EnvironmentName](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostenvironment.environmentname) property to set the environment like this:

```csharp
Host.CreateDefaultBuilder(args)
    .UseEnvironment("Sideload")
    // ...
```

yaira2:
> We should be using the [IHostEnvironment.EnvironmentName](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostenvironment.environmentname) property to set the environment like this:

We'll need to add some more configurations, right now sideload and sideload preview use the same configuration but they have different environments.

src/Files.App/AppModels/AppEnvironment.cs
0x5bfa, line 1: This file can be moved to Backend?

hecksmosis: Yes, I think so

hecksmosis:
> We should be using the [IHostEnvironment.EnvironmentName](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostenvironment.environmentname) property to set the environment like this:
> 
> ```cs
> Host.CreateDefaultBuilder(args)
> #if SIDELOAD
>     .UseEnvironment("Sideload")
> #else
>     .UseEnvironment("Store")
> #endif
>     // ...
> ```

How would this be done with the enum?

yaira2:
@hecksmosis we might have to rework this, Luke's suggestion would be instead of an enum but it won't work unless we add additional configurations.

hecksmosis:
> @hecksmosis we might have to rework this, Luke's suggestion would be instead of an enum but it won't work unless we add additional configurations to the project. I think it's best to use an enum for now and hold off on implementing `IHostEnvironment.EnvironmentName`.

So should we merge this for the time being or do I modify it?

yaira2:
@hecksmosis we can merge but I think we should move the code out of app.xaml. The question is to where...

heftymouse:
Would it be better to set the value using preprocessor directives set by MSBuild instead of reading from the display name?

yaira2:
> Would it be better to set the value using preprocessor directives set by MSBuild instead of reading from the display name?

Once it's setup then yes, but currently we don't have a directive for each config.

yaira2:
@hecksmosis I created the new project configurations so we can switch away from using the display name now.

hecksmosis:
> @hecksmosis I created the new project configurations so we can switch away from using the display names.

What are the environment name strings?

yaira2:
- Store
- Stable
- Preview
- Debug
- Release

hecksmosis:
Is release the dev version?

yaira2:
Release and Debug are both dev versions.

hecksmosis:
Ok

src/Files.App/App.xaml.cs
```diff
@@ -73,6 +74,7 @@ public partial class App : Application
 
 		public static string AppVersion = $"{Package.Current.Id.Version.Major}.{Package.Current.Id.Version.Minor}.{Package.Current.Id.Version.Build}.{Package.Current.Id.Version.Revision}";
 		public static string LogoPath;
+		public static AppEnvironment AppEnv;
```
yaira2, line 77: This doesn't really belong in App.xaml.cs but we can patch it up in another PR.

yaira2: Ideally we can build a service that has values for `Environment`, `LogoPath`, `Version` etc.

src/Files.App/Helpers/EnvHelpers.cs
```diff
@@ -0,0 +1,42 @@
...
+				AppEnvironment.WindowsStore;
+#elif PREVIEW
+				AppEnvironment.WindowsSideloadPreview;
+#elif STABLE
+				AppEnvironment.WindowsSideload;
+#elif DEBUG
+				AppEnvironment.Debug;
+#else 
+				AppEnvironment.WindowsDev;
+#endif
```
yaira2, line 30:
```suggestion
#if STORE
				AppEnvironment.Store;
#elif PREVIEW
				AppEnvironment.Preview;
#elif STABLE
				AppEnvironment.Stable;
#elif DEBUG
				AppEnvironment.Debug;
#else 
				AppEnvironment.Dev;
#endif
```

src/Files.App/Helpers/EnvHelpers.cs
```diff
@@ -0,0 +1,42 @@
...
+
+			var path = env switch
+			{
+				AppEnvironment.Release => Constants.AssetPaths.DevLogo,
```
yaira2, line 34:
```suggestion
				AppEnvironment.Dev=> Constants.AssetPaths.DevLogo,
```

hecksmosis: This would mean changing the EnvironmentName to Dev, because it's set as the string representation of the enum variant. Do we want to do that?

src/Files.App/App.xaml.cs
hecksmosis: Ok

hecksmosis: > This doesn't really belong in App.xaml.cs but we can patch it up in another PR.

Should I do it after this is merged?

yaira2: Yes, it's out of scope for this PR.

src/Files.App/Helpers/EnvHelpers.cs
yaira2: You're right, can you replace the `release` and `debug` enum with `Dev`.

hecksmosis: Ok

yaira2 (approved):
LGTM

PR #12033, opened 2023-04-09
Fix: Some icons are not displayed

ferrariofilippo:
**Resolved / Related Issues**
   Regression #11928

**Validation**
   1. Open app and see if `Delete` icon is shown
   2. Open context menu and if `Delete` and `Paste` icons are shown

### DIFF
--- a/src/Files.App/Actions/BaseUIAction.cs
+++ b/src/Files.App/Actions/BaseUIAction.cs
@@ -1,27 +1,17 @@
 ﻿using CommunityToolkit.Mvvm.ComponentModel;
-using Files.App.Commands;
 using Files.App.Helpers;
 using System.ComponentModel;
-using System.Threading.Tasks;
 
 namespace Files.App.Actions
 {
-	internal abstract class BaseUIAction : ObservableObject, IAction
+	internal abstract class BaseUIAction : ObservableObject
 	{
-		public abstract string Label { get; }
-
-		public abstract string Description { get; }
-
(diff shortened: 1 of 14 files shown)


hishitetsu (approved):
LGTM

src/Files.App/Actions/BaseUIAction.cs
```diff
@@ -11,6 +12,8 @@ internal abstract class BaseUIAction : ObservableObject, IAction
 
 		public abstract string Description { get; }
 
+		public virtual RichGlyph Glyph { get; } = RichGlyph.None;
+
```
hishitetsu, line 16: HotKey should be declared as well. Otherwise, hotkeys will not be enabled in the same way as icons.
I rather think IAction should be implemented in each concrete Action, not in BaseUIAction. This way, there is no need to declare the same default implementation in IAction and BaseUIAction.

yaira2: @ferrariofilippo

ferrariofilippo: > HotKey should be declared as well. Otherwise, hotkeys will not be enabled in the same way as icons. I rather think IAction should be implemented in each concrete Action, not in BaseUIAction. This way, there is no need to declare the same default implementation in IAction and BaseUIAction.

Yes it makes sense. I didn't think that we had to declare fiels in BaseUIAction too when I first created it, anyway I'll revert it now since it will make things clearer

hishitetsu (approved):
LGTM

PR #12043, opened 2023-04-11
Feature: Added swipe gesture for back/forward navigation

### LINKED ISSUE #100: Feature: Implement back/forward touch gestures
Would be nice to have horizontal overscroll touchpad gestures for going back and forward, like in Chrome and Edge.

**Requirements**
- Swiping left/right navigates backwards or forwards.
- Works with touchpad gestures
- Works with touch screen gestures

heftymouse:
**Resolved / Related Issues**

   Closes #100

**Validation**
   1. Open app ...
   2. Click settings button ...

If you have the time you're encourage to clone this PR and see if the gestures 'feel right' to you. Please leave a comment if there's anything that could be improved.


heftymouse:
Please ignore the fact that there's 23 extra commits I accidentally pressed the wrong button

src/Files.App/Views/NavigationInteractionTracker.cs
```diff
@@ -0,0 +1,205 @@
...
+
+			public void RequestIgnored(InteractionTracker sender, InteractionTrackerRequestIgnoredArgs args)
+			{
+
+			}
+
+			public void ValuesChanged(InteractionTracker sender, InteractionTrackerValuesChangedArgs args)
+			{
+
+			}
```
ferrariofilippo, line 196: What are these methods needed for?

heftymouse: They're only there to comply with IInteractionTrackerOwner

src/Files.App/Views/NavigationInteractionTracker.cs
```diff
@@ -0,0 +1,205 @@
...
+		{
+			get
+			{
+				_props.TryGetBoolean("CanGoBack", out bool val);
+				return val;
+			}
+			set
+			{
+				_props.InsertBoolean("CanGoBack", value);
+			}
```
ferrariofilippo, line 41: As above

src/Files.App/Views/NavigationInteractionTracker.cs
```diff
@@ -0,0 +1,205 @@
...
+
+			ElementCompositionPreview.SetIsTranslationEnabled(_backIcon, true);
+			ElementCompositionPreview.SetIsTranslationEnabled(_forwardIcon, true);
+			_rootVisual = ElementCompositionPreview.GetElementVisual(_rootElement);
+			_backVisual = ElementCompositionPreview.GetElementVisual(_backIcon);
+			_forwardVisual = ElementCompositionPreview.GetElementVisual(_forwardIcon);
+
+			_props = _rootVisual.Compositor.CreatePropertySet();
+			_props.InsertBoolean("CanGoForward", false);
+			_props.InsertBoolean("CanGoBack", false);
```
ferrariofilippo, line 73: As above

src/Files.App/Views/NavigationInteractionTracker.cs
```diff
@@ -0,0 +1,205 @@
...
+		private UIElement _forwardIcon;
+
+		private Visual _rootVisual;
+		private Visual _backVisual;
+		private Visual _forwardVisual;
+
+		private InteractionTracker _tracker = null!;
+		private VisualInteractionSource _source = null!;
+		private InteractionTrackerOwner _trackerOwner = null!;
+		private CompositionPropertySet _props = null!;
```
ferrariofilippo, line 55:
```suggestion
		private InteractionTracker _tracker;
		private VisualInteractionSource _source;
		private InteractionTrackerOwner _trackerOwner;
		private CompositionPropertySet _props;
```

You don't need to declare them as `not null` since you assign them in the `constructor`
You may also consider to mark them as `readonly`

src/Files.App/Views/NavigationInteractionTracker.cs
```diff
@@ -0,0 +1,205 @@
...
+
+			var forwardResistance = CompositionConditionalValue.Create(compositor);
+			var forwardResistanceCondition = compositor.CreateExpressionAnimation("tracker.Position.X > 0 && tracker.Position.X < 96");
+			forwardResistanceCondition.SetReferenceParameter("tracker", _tracker);
+			var forwardResistanceValue = compositor.CreateExpressionAnimation("source.DeltaPosition.X * (1 - sqrt(1 - square((tracker.Position.X / 96) - 1)))");
+			forwardResistanceValue.SetReferenceParameter("source", _source);
+			forwardResistanceValue.SetReferenceParameter("tracker", _tracker);
+			forwardResistance.Condition = forwardResistanceCondition;
+			forwardResistance.Value = forwardResistanceValue;
+
```
ferrariofilippo, line 119: This is almost the same code repeated twice. You may create a method where you pass in `resistanceConditionString` and `resistanceValueString`

src/Files.App/Views/NavigationInteractionTracker.cs
```diff
@@ -0,0 +1,205 @@
...
+			}
+
+			public void CustomAnimationStateEntered(InteractionTracker sender, InteractionTrackerCustomAnimationStateEnteredArgs args)
+			{
+
+			}
+
+			public void IdleStateEntered(InteractionTracker sender, InteractionTrackerIdleStateEnteredArgs args)
+			{
+				if (!_shouldBounceBack) return;
```
ferrariofilippo, line 159:
```suggestion
				if (!_shouldBounceBack)
				    return;
```

ferrariofilippo (commented):
I left some suggestions, tell me what you think about them
Overall it looks good and it works for me

src/Files.App/Views/NavigationInteractionTracker.cs
ferrariofilippo: You may group them (maybe at the end of the class) and add a comment to explain why they are there

ferrariofilippo (approved):
LGTM and works for me!

heftymouse:
The UX for the indicator thingy needs to be improved. Currently, it's just a circle that comes out from the side of the page, with no indication of how far to go to trigger a navigation. I really like the way that Chrome does it but it'd be better to have some original element as well.

yaira2:
@heftymouse I noticed you added a setting to turn off this behavior, is there a reason you think we need a setting?

heftymouse:
One could find themselves accidentally triggering it if they use a lot of
force when scrolling which could get annoying

d2dyno1:
Sorry, I don't have a device with touchpad

hishitetsu:
I tried. It is very cool and works well for me!
It would be better if it could refresh the page when swiping down.

> One could find themselves accidentally triggering it if they use a lot of
force when scrolling which could get annoying

I have never had an unintended swipe, but I think there should be an option.

yaira2:
> It would be better if it could refresh the page when swiping down.

We used to have this but it wasn't widely used and was unreliable, users navigate back and forth a lot more than they refresh.

hishitetsu:
> We used to have this but it wasn't widely used and was unreliable

I didn't know it existed before. With this implementation, it appears to be reliable.

yaira2:
It wouldn't hurt to give it another try 👍

src/Files.App/Files.App.csproj
```diff
@@ -84,7 +84,7 @@
         <PackageReference Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.22621.756" />
         <PackageReference Include="SevenZipSharp" Version="1.0.0" />
         <PackageReference Include="SQLitePCLRaw.bundle_green" Version="2.1.4" />
-        <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.2.230313.1" />
+        <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.3.230331000" />
```
yaira2, line 87:
```suggestion
        <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.2.230313.1" />
```

yaira2:
I agree that it's too sensitive, we can solve this by only triggering the action if the layout is already scrolled to the left or right.

yaira2:
I think this need to be disabled on the column layout.

src/Files.App/Views/ModernShellPage.xaml
```diff
@@ -51,12 +52,48 @@
...
 
-	<Frame
-		x:Name="ItemDisplayFrame"
-		HorizontalAlignment="Stretch"
-		x:FieldModifier="public"
-		BorderBrush="{x:Bind CurrentInstanceBorderBrush, Mode=OneWay}"
-		BorderThickness="{x:Bind CurrentInstanceBorderThickness, Mode=OneWay}"
-		Navigated="ItemDisplayFrame_Navigated" />
+	<Grid HorizontalAlignment="Stretch" wct:UIElementExtensions.ClipToBounds="True">
+		<Grid
```
yaira2, line 56: Does ellipse support child elements? If it does, maybe we can eliminate this grid.

heftymouse: Ellipse isn't a ContentControl so it doesn't, but the ellipse can be eliminated by setting the same properties on the grid instead.

yaira2: What about `Border`?

heftymouse: I've changed it to Border

yaira2 (approved):
LGTM

PR #12050, opened 2023-04-11
Fix: Fixed Ctrl+A in the rename textbox

### LINKED ISSUE #12040: Bug: Ctrl + A shortcut isn't selecting text when renaming files

Hello, I work with the extensions of the files shown, when I press F2 to change the file name, I also press CTRL+A to edit the whole name of the file. Or copy the name of the file. 
On Windows Explorer, pressing F2 enables the change name for the file (files does this too) and then CTRL+A selects the entire text on the text box. 

image

### Requirements

- This proposal will achieve the functionallity of the combination CTRL+A to select the entire text inside the textbox

### Files Version

2.4.61.0

### Windows Version

10.0.19044.2728

### Comments

_No response_

hishitetsu:
_Targeted to the servicing branch_

**Resolved / Related Issues**
   Closes #12040 

**Validation**

### DIFF
--- a/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
@@ -260,6 +260,9 @@ namespace Files.App.Views.LayoutModes
 
 			if (ctrlPressed && e.Key is VirtualKey.A)
 			{
+				if (IsRenamingItem)
+					return;
+
 				e.Handled = true;
 
 				var commands = Ioc.Default.GetRequiredService<ICommandManager>();
--- a/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
@@ -252,7 +252,7 @@ namespace Files.App.Views.LayoutModes
 
 		protected override async void FileList_PreviewKeyDown(object sender, KeyRoutedEventArgs e)
 		{
-			if (ParentShellPageInstance is null)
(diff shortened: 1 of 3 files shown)


ferrariofilippo:
I'm wondering if we should disable all the shortcuts when renaming items...

hishitetsu:
> I'm wondering if we should disable all the shortcuts when renaming items...

Certainly. However, this fix is necessary anyway, since the code is intended to disable the ListView native shortcut and redirect to RichCommand.

QuaintMako (approved):
LGTM, it solves the issue.

QuaintMako:
Regarding the disabling of all shortcuts when renaming, that is possibly a good idea. 
I'd suggest opening an feature request for it to open discussions around it.

ferrariofilippo:
> Certainly. However, this fix is necessary anyway, since the problem here is not that Select All Items is invoked, but that Select All Text is not invoked. The code is intended to disable the ListView native shortcut and redirect to RichCommand.

I'm sorry probably I was not very clear. That comment meant that we could move `if (IsRenamingItem) return;` to the top of the method

ferrariofilippo:
> Regarding the disabling of all shortcuts when renaming, that is possibly a good idea. I'd suggest opening an feature request for it to open discussions around it.

Just opened #12053

hishitetsu:
> I'm sorry probably I was not very clear. That comment meant that we could move `if (IsRenamingItem) return;` to the top of the method

I see. I will address it that way later.

hishitetsu:
> I'm sorry probably I was not very clear. That comment meant that we could move `if (IsRenamingItem) return;` to the top of the method

I addressed it.

PR #12074, opened 2023-04-15
RichCommands: SearchUnindexedItems

ferrariofilippo:
**Resolved / Related Issues**
   Related #11481 

**Validation**
   1. Open app
   2. Search something that is not indexed
   3. Click on `Search unindexed items`


yaira2:
@ferrariofilippo can you split the PR to help have a clearer commit history?

ferrariofilippo:
@yair100 I saw you told me to split this PR, but now I no longer see the comment (did you delete it?), should I still open a separate PR for PermanentDeletion?

yaira2:
It's preferred but I don't want to be too picky.

yaira2:
I'm not convinced this should have an action, it's likely to be confused between the setting and the button on the search page.

ferrariofilippo:
> I'm not convinced this should have an action, it's likely to be confused between the setting and the button on the search page.

I think we could rename it, but we definitely need to add this action. Otherwise we would need to keep `IBaseLayoutCommandImplementation` just for this action.

0x5bfa (dismissed):
First introduction of rich commands in layout modes!

yaira2:
@ferrariofilippo after giving this some thought, I decided that we should remap the button on the search to open the settings page where users can toggle the setting for themselves.

The settings dialog doesn't have support for deep linking to specific pages (not yet) so we're going to have to hold off on this.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3178,6 +3178,9 @@
   <data name="ShowCheckboxesWhenSelectingItems" xml:space="preserve">
     <value>Show checkboxes when selecting items</value>
   </data>
+  <data name="SearchUnindexedItemsDescription" xml:space="preserve">
+    <value>Include unindexed items in search results</value>
```
yaira2, line 3182: What about "Search for unindexed items"?

yair100 (approved):
LGTM

yair100 (approved):
LGTM

yair100 (approved):
LGTM

PR #11898, opened 2023-03-29
Feature: Close unfocused blades

ferrariofilippo:
**Resolved / Related Issues**
- [x] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
   Closes None

**Validation**
How did you test these changes?
- [x] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [x] Did you implement any design changes to an existing feature?
   - [x] Was this change approved?
- [x] Are there any other steps that were used to validate these changes?
   1. Open a path such as `C:\Windows\System32` in columns layout
   2. Click on `Windows` (Single click to open option)

PR #12124, opened 2023-04-19
Fix: Fixed cannot select Files in columns view

### LINKED ISSUE #12090: Bug: Can't select some files in column view
### Description

When trying to select some files in column view by clicking them, Files closes all subsequent subfolders opened in columns and returns to the first column.

https://user-images.githubusercontent.com/2331058/232351898-5d8615ef-9948-4ab1-bed3-3e98c491582b.mp4

### Steps To Reproduce

1. Open some folders in column view
2. Try to select files of different names and types (I haven't narrowed down the exact issue)
3. Some of them will cause the bug to happen

### Requirements

- eg. Fix crash that occurs when...
- eg. Correct action happens when clicking...

### Files Version

2.4.62.0

### Windows Version

10.0.19044.2846

### Log file

There are no logs about the bug in the debug.log file, but I attached it anyways. The bug happened at on 2023-04-16 around 9 PM.

[debug.log](https://github.com/files-community/Files/files/11245022/debug.log)

ferrariofilippo:
**Resolved / Related Issues**
   Closes #12090 

**Description**
- I simplified the code to find the last column that should remain open (I don't know why I wrote that complex code in the first place). Now it looks for the blade with the same path as the requested one.
- As mentioned in #11898, there's a bug with double click to open

**Notes**
- Please, test these changes deeply since they may cause many regressions
- It may happen that the user creates a loop of columns (like: `C:\Users\username\Desktop\shortcut_to_Users\username\Desktop\something_else`). In this case, if the user clicks an item in `Desktop`, should we always close all columns after the first `Desktop`?

**Validation**
**Test 1**
   1. Open app and a folder using colums view
   2. Use `Open folders with single click in columns view`
   3. Open a shortcut to a folder (try with a shortcut to a sub-subfolder of the current directory, a shortcut to a parent folder of the current and with one to a random folder)
   4. Select a file
**Test 2**
   1. Repeat all the steps above using `double click to open`
   2. Navigate between folders using `Up/Down arrows` to ensure there's no regression from #11986   
**Test 3**
   1. Open a path such as C:\Windows\System32 in columns layout
   2. Click on Windows (Single click to open option) (Check that there's no regression from #11898)

### DIFF
--- a/src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
@@ -441,6 +441,9 @@ namespace Files.App.Views.LayoutModes
 
 		private void CloseUnnecessaryColumns(ColumnParam column)
 		{
+			if (string.IsNullOrEmpty(column.NavPathParam))
+				return;
+
 			var relativeIndex = -1;
 
 			if (column.Source is not null)
@@ -455,7 +458,7 @@ namespace Files.App.Views.LayoutModes
 
 			if (relativeIndex is -1)
 				while (relativeIndex < ColumnHost.ActiveBlades.Count &&
-					(!column.NavPathParam?.Equals(GetWorkingDirOfColumnAt(++relativeIndex)) ?? false)) ;
+					!column.NavPathParam.Equals(GetWorkingDirOfColumnAt(++relativeIndex))) ;
 
 			if (relativeIndex >= 0 && relativeIndex < ColumnHost.ActiveBlades.Count)
 			{
--- a/src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
@@ -444,7 +444,7 @@ namespace Files.App.Views.LayoutModes
 			if (string.IsNullOrEmpty(column.NavPathParam))
 				return;
 
-			var relativeIndex = -1;
+			var relativeIndex = column.Column is not 0 ? column.Column : -1;
 
 			if (column.Source is not null)
 			{
@@ -457,6 +457,7 @@ namespace Files.App.Views.LayoutModes
 			}
 
 			if (relativeIndex is -1)
+				// Get the index of the blade with the same path as the requested
 				while (relativeIndex < ColumnHost.ActiveBlades.Count &&
 					!column.NavPathParam.Equals(GetWorkingDirOfColumnAt(++relativeIndex))) ;
 
--- a/src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
@@ -457,11 +457,16 @@ namespace Files.App.Views.LayoutModes
 			}
 
 			if (relativeIndex is -1)
+			{
 				// Get the index of the blade with the same path as the requested
-				while (relativeIndex < ColumnHost.ActiveBlades.Count &&
-					!column.NavPathParam.Equals(GetWorkingDirOfColumnAt(++relativeIndex))) ;
+				var blade = ColumnHost.ActiveBlades.FirstOrDefault(b => 
+					column.NavPathParam.Equals(((b.Content as Frame)?.Content as ColumnShellPage)?.FilesystemViewModel?.WorkingDirectory));
 
-			if (relativeIndex >= 0 && relativeIndex < ColumnHost.ActiveBlades.Count)
+				if (blade is not null)
+					relativeIndex = ColumnHost.ActiveBlades.IndexOf(blade);
+			}
+
+			if (relativeIndex >= 0)
 			{
 				ColumnHost.ActiveBlades[relativeIndex].FindDescendant<ColumnViewBase>()?.ClearOpenedFolderSelectionIndicator();
 				DismissOtherBlades(relativeIndex);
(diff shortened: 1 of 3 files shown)


yaira2:
@ferrariofilippo can you target the servicing branch?

hishitetsu:
> It may happen that the user creates a loop of columns

It is the same with Finder.
image

hishitetsu (commented):
It doesn't work well if looping shortcuts.

https://user-images.githubusercontent.com/66369541/233252819-12dcfa7e-b688-4c9c-85a3-d4a10c595028.mp4

ferrariofilippo:
Do you have double click to open? In that case I know that bug (is the one I mentioned in the description) and I'm working on a fix

hishitetsu:
> Do you have double click to open? In that case I know that bug (is the one I mentioned in the description) and I'm working on a fix

"Open items with a single click" is off, "Open folders with a single click in the Columns Layout" is on.
If the shortcut was not looped, the file could be opened without problems.

ferrariofilippo:
> "Open items with a single click" is off, "Open folders with a single click in the Columns Layout" is on. If the shortcut was not looped, the file could be opened without problems.

Ok, I just understand what you meant. I mentioned that in the *Notes*
> It may happen that the user creates a loop of columns (like: C:\Users\username\Desktop\shortcut_to_Users\username\Desktop\something_else). In this case, if the user clicks an item in Desktop, should we always close all columns after the first Desktop?

At the moment we can't know which column the user has clicked (that's an issue when we have three loops). Because of this, we have to choose between leaving open the first or the last occourence of the path

ferrariofilippo:
Everything should be fine now

src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
```diff
@@ -435,20 +435,29 @@ private void ColumnViewBase_ItemTapped(object? sender, EventArgs e)
...
+				{
+					var bladeColumn = ColumnHost.ActiveBlades[i].FindDescendant<ColumnViewBase>();
+					if (bladeColumn is not null && bladeColumn == column.Source)
+						relativeIndex = i;
+				}
+			}
+
+			if (relativeIndex is -1)
+				while (relativeIndex < ColumnHost.ActiveBlades.Count &&
+					(!column.NavPathParam?.Equals(GetWorkingDirOfColumnAt(++relativeIndex)) ?? false)) ;
```
hishitetsu, line 458: This code is confusing and should not be handled only by the conditional clause of the while loop.
Doesn't this cause an infinite loop when `NavPathParam` is `null`?

hishitetsu (commented):
File selection appears to work fine.
There is still a little weird behavior with folder selection when columns are looping, but it's a different issue and much better than not being able to open files.

src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
ferrariofilippo: I'll change that but it will not cause an infinite loop because of the first condition

ferrariofilippo: Actually, I'm not sure how to make it more readable. Should I use a `do-while`?

ferrariofilippo:
> There is still a little weird behavior with folder selection when columns are looping, but it's a different issue and much better than not being able to open files.

I can't reproduce it. Can you tell me more, please? (single or double click to open?)

hishitetsu:
> I can't reproduce it. Can you tell me more, please? (single or double click to open?)

Look at this video.

https://user-images.githubusercontent.com/66369541/233562759-e38e01cb-7915-4994-a217-a874936ed012.mp4

ferrariofilippo:
> Look at this video.

I can't reproduce it and, looking at the scrollbar on the bottom, it seems to work.
Do you have a loop like `Documents/Downloads/Documents/Downloads/Documents`? In that case clicking on the second `Documents` will close the following blades, but the view will also scroll and it will look like it's opening from the last column

src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
```diff
@@ -435,20 +435,33 @@ private void ColumnViewBase_ItemTapped(object? sender, EventArgs e)
...
+						relativeIndex = i;
+				}
+			}
+
+			if (relativeIndex is -1)
+				// Get the index of the blade with the same path as the requested
+				while (relativeIndex < ColumnHost.ActiveBlades.Count &&
+					!column.NavPathParam.Equals(GetWorkingDirOfColumnAt(++relativeIndex))) ;
+
+			if (relativeIndex >= 0 && relativeIndex < ColumnHost.ActiveBlades.Count)
```
hishitetsu, line 464:
```suggestion
			if (relativeIndex is -1)
			{
				// Get the index of the blade with the same path as the requested
				var bladeColumn = ColumnHost.ActiveBlades.FirstOrDefault(x
					=> column.NavPathParam.Equals(((x.Content as Frame)?.Content as ColumnShellPage)?.FilesystemViewModel?.WorkingDirectory));

				if (bladeColumn is not null)
					relativeIndex = ColumnHost.ActiveBlades.IndexOf(bladeColumn);
			}

			if (relativeIndex >= 0)
```
For example.

hishitetsu:
> I can't reproduce it and, looking at the scrollbar on the bottom, it seems to work.

Sorry, it is an oversight on my part. It works fine.

src/Files.App/Views/LayoutModes/ColumnViewBrowser.xaml.cs
ferrariofilippo: At first, I used a while because I thought that scanning through the blades twice was a waste of time, but as you pointed out that was not that readable, and so I'll use your code. Thank you

hishitetsu (approved):
LGTM

yaira2:
@ferrariofilippo @hishitetsu thank you for your work on this fix!

PR #12150, opened 2023-04-21
Feature: Disable warning when a new name is empty

### LINKED ISSUE #12140: Feature: Disable warning when a new name is empty

When the dialog box to enter a new item name is displayed, the user is warned that the name is invalid from the beginning. I think it is confusing to users to get a warning when they have not done anything yet and it is obvious that the name should not be empty, so there is no need to warn.
The same can be said when creating a new tag.

image

### Requirements

Disable warning for invalid names when the name is empty.

### Files Version

2.4.66.0

### Windows Version

Windows 11 22H2

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**
   Closes #12140

**Removed from**
- New Item Dialog
- New Archive Dialog
- New Shortcut Dialog

**Validation**
   1. Open app
   2. Open the menu to create a new file
   3. See that the warning is not displayed when the name is empty


yaira2:
Should we target the servicing branch?

ferrariofilippo:
I don't think this issue is that urgent

yaira2 (dismissed):
LGTM

ferrariofilippo:
> LGTM

What do you think about the two questions in the description?

yaira2:
> Should we disable the tip also when the shortcut path is invalid?
> I believe we shouldn't remove that from tags. What do you think?

Is there any reason to handle it differently?

ferrariofilippo:
> Is there any reason to handle it differently?

For the first one (shortcuts) I agree, we should not show the tip. It was not mentioned in the issue and I wanted to wait on some feedback to avoid wasting work
For tags, I don't think the warning is disturbing there, I mean it's not covering anything else (as the tip does)

yaira2:
We can try without it then 👍

hishitetsu:
I think this should be merged into the servicing brunch.

hishitetsu:
And I think the tag warning should be removed. The problem is not that something is hidden, but that the warnings are just unnecessary and offensive to the user.

ferrariofilippo:
> And I think the tag warning should be removed. The problem is not that something is hidden, but that the warnings are just unnecessary and offensive to the user.

I removed it for empty names but not for whitespaces

ferrariofilippo:
> I think this should be merged into the servicing brunch. It is certainly not a breaking bug, but I think it is making the UX much worse.

I swapped it to `servicing` but the changes are like 3k lines (instead of something like 20). I think I should change back to `main`

hishitetsu:
> I swapped it to `servicing` but the changes are like 3k lines (instead of something like 20). I think I should change back to `main`

Perhaps you have to create another branch based on the servicing branch and cherry-pick your commits into it.

PR #12159, opened 2023-04-22
Fix: Ensure GIF and TIFF are treated as images

### LINKED ISSUE #12133: Bug: GIF and TIFF aren't treated as images
### Description

GIF and TIFF aren't treated as images in Files, so they can't be set as the background or rotated.

### Steps To Reproduce

1. Select a GIF or TIFF image.
2. "Set as background" and "Rotate" option isn't appear.

### Requirements

Ensure GIF and TIFF are treated as images, as well as BMP, JPEG, PNG.

### Files Version

2.4.66.0

### Windows Version

Windows 11 22H2

### Log file

n/a

f-pereira:
**Resolved / Related Issues**
   Closes #12133

**Validation**
   1. Open app
   2. Select a .gif or .tiff file
   3. Should have image options (rotate left, rotate right, etc...)

**Screenshots (optional)**
Screenshot 2023-04-22 163710

### DIFF
--- a/src/Files.App/Helpers/BitmapHelper.cs
+++ b/src/Files.App/Helpers/BitmapHelper.cs
@@ -63,6 +63,12 @@ namespace Files.App.Helpers
 			using var memStream = new InMemoryRandomAccessStream();
 			BitmapEncoder encoder = await BitmapEncoder.CreateForTranscodingAsync(memStream, decoder);
 
+			for(int i = 0; i < decoder.FrameCount - 1; i++) 
+			{
+				encoder.BitmapTransform.Rotation = rotation;
+				await encoder.GoToNextFrameAsync();
+			}
+
 			encoder.BitmapTransform.Rotation = rotation;
 
 			await encoder.FlushAsync();
--- a/src/Files.App/Helpers/BitmapHelper.cs
+++ b/src/Files.App/Helpers/BitmapHelper.cs
@@ -63,7 +63,7 @@ namespace Files.App.Helpers
 			using var memStream = new InMemoryRandomAccessStream();
 			BitmapEncoder encoder = await BitmapEncoder.CreateForTranscodingAsync(memStream, decoder);
 
-			for(int i = 0; i < decoder.FrameCount - 1; i++) 
+			for (int i = 0; i < decoder.FrameCount - 1; i++) 
 			{
 				encoder.BitmapTransform.Rotation = rotation;
 				await encoder.GoToNextFrameAsync();
--- a/src/Files.Backend/Helpers/FileExtensionHelpers.cs
+++ b/src/Files.Backend/Helpers/FileExtensionHelpers.cs
@@ -26,7 +26,7 @@ namespace Files.Backend.Helpers
 		/// <returns><c>true</c> if the fileExtensionToCheck is an image;
 		/// otherwise, <c>false</c>.</returns>
 		public static bool IsImageFile(string? fileExtensionToCheck)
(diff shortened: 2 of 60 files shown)


src/Files.Backend/Helpers/FileExtensionHelpers.cs
```diff
@@ -26,7 +26,7 @@ public static bool HasExtension(string? filePathToCheck, params string[] extensi
 		/// <returns><c>true</c> if the fileExtensionToCheck is an image;
 		/// otherwise, <c>false</c>.</returns>
 		public static bool IsImageFile(string? fileExtensionToCheck)
-			=> HasExtension(fileExtensionToCheck, ".png", ".bmp", ".jpg", ".jpeg");
+			=> HasExtension(fileExtensionToCheck, ".png", ".bmp", ".jpg", ".jpeg", ".gif", ".tiff");
```
ferrariofilippo, line 29: You should also include `.tif`, since it is the same as `.tiff`

f-pereira: You're absolutely correct, my bad!

ferrariofilippo:
There's a bug when rotating GIFs, but I think it's a different issue

f-pereira:
Oh I see, It seems that only the first frame is rotated

hishitetsu:
Thank you for your PR! This page will help you to rotate all frames.
https://learn.microsoft.com/en-us/uwp/api/windows.graphics.imaging.bitmapencoder?view=winrt-22621

> If you want to encode an additional frame, call GoToNextFrameAsync.

f-pereira:
@hishitetsu Thank you for the documentation! I tried to use it and I found a problem that I am Looking for solution: the first frame rotates well, but if the next frame have a different width/height, it rotates well but It is misplaced (e.g It is placed on top right corner instead of center)

hishitetsu:
* Will there be any problem if all frames had the same width and height?
* Will rotating four times in a row or rotating left and then right restore all frames to there original state even if there are frames of different widths and heights?

If no to the former and yes to the latter, I think that is still fine.

f-pereira:
@hishitetsu Done! 
I tested with a gif where all the frames are equal in size and it worked well.
I also tested with other image formats and it is working

ferrariofilippo:
> @hishitetsu Done! I tested with a gif where all the frames are equal in size and it worked well. I also tested with other image formats and it is working

I'm getting a frame wrong (but if I rotate the gif 4 times everything works). I used <a href="https://upload.wikimedia.org/wikipedia/commons/thumb/2/2c/Rotating_earth_%28large%29.gif/260px-Rotating_earth_%28large%29.gif">this</a> to test

f-pereira:
I noticed that it happens when the frames are not equal in size (I checked your gif, the frames size are different)
To confirm this hypothesis I used this gif

 this

src/Files.App/Helpers/BitmapHelper.cs
```diff
@@ -63,6 +63,12 @@ public static async Task Rotate(string filePath, BitmapRotation rotation)
 			using var memStream = new InMemoryRandomAccessStream();
 			BitmapEncoder encoder = await BitmapEncoder.CreateForTranscodingAsync(memStream, decoder);
 
+			for(int i = 0; i < decoder.FrameCount - 1; i++)
```
hishitetsu, line 66:
```suggestion
			for (int i = 0; i < decoder.FrameCount - 1; i++) 
```

ferrariofilippo (approved):
LGTM and works for me!

yaira2 (approved):
Thank you

PR #12175, opened 2023-04-23
 Fix: Fixed possible System.NullReferenceException in ThemeHelper

yaira2:
**Resolved / Related Issues**
   Closes #issue...

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
image

### DIFF
--- a/src/Files.App/Helpers/ThemeHelper.cs
+++ b/src/Files.App/Helpers/ThemeHelper.cs
@@ -62,14 +62,11 @@ namespace Files.App.Helpers
 		private static async void UiSettings_ColorValuesChanged(UISettings sender, object args)
 		{
 			// Make sure we have a reference to our window so we dispatch a UI change
-			if (currentApplicationWindow is not null)
-			{
-				// Dispatch on UI thread so that we have a current appbar to access and change
-				await currentApplicationWindow.DispatcherQueue.EnqueueAsync(() =>
-				{
-					ApplyTheme();
-				});
-			}
+			if (currentApplicationWindow is null)
+				return;
+
+			// Dispatch on UI thread so that we have a current appbar to access and change
+			await currentApplicationWindow.DispatcherQueue.EnqueueAsync(ApplyTheme);
 		}
 
 		private static void ApplyTheme()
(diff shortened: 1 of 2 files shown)


hishitetsu:
https://github.com/files-community/Files/blob/c51f7a4f81f48a2cc930266828ed4bcf336ec131/src/Files.App/Helpers/ThemeHelper.cs#L46-L52

If `currentApplicationWindow` is `null`, a NullReferenceException may occur at line 52.

yaira2:
@hishitetsu thanks!

src/Files.App/Helpers/ThemeHelper.cs
```diff
@@ -62,24 +63,19 @@ public static void Initialize()
...
-			if (currentApplicationWindow is not null)
-			{
-				// Dispatch on UI thread so that we have a current appbar to access and change
-				await currentApplicationWindow.DispatcherQueue.EnqueueAsync(() =>
-				{
-					ApplyTheme();
-				});
-			}
+			if (currentApplicationWindow is null)
+				return;
```
hishitetsu, line 67: If `currentApplicationWindow` is `null` here, shouldn't we try initialize again?

yaira2: Wouldn't we get stuck in a loop?

hishitetsu: We should try one time every the event raised.

yaira2: I'm curious what you have in mind, would you add a Boolean to keep track?

hishitetsu: I have like this in mind.
```csharp
		private static async void UiSettings_ColorValuesChanged(UISettings sender, object args)
		{
			// Make sure we have a reference to our window so we dispatch a UI change
			if (currentApplicationWindow is null)
			{
				currentApplicationWindow = App.Window;

				if (currentApplicationWindow is not null)
					titleBar = App.GetAppWindow(currentApplicationWindow).TitleBar;
				else
					return;
			}

			// Dispatch on UI thread so that we have a current appbar to access and change
			await currentApplicationWindow.DispatcherQueue.EnqueueAsync(ApplyTheme);
		}
```

yaira2: @hishitetsu can you open a PR?

PR #11493, opened 2023-02-27
Feature: Implement more RichCommands

hecksmosis:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Related to #11481 (I'm trying to do these in little chunks to make them easier to review)

**Validation**
How did you test these changes?
- [x] Built and ran the app

PR #12193, opened 2023-04-24
Fix: Fixed issue where unpin from start didn't work

mafra99:
**Description**
The `UnpinFromStartAction` calls currently the `TryPinFolderAsync` instead of the `UnpinFromStartAsync` method.
The bug was introduced in #11493 where the code was migrated to a RichCommand.

With this change it is again possible to unpin items from the start menu.

**Validation**
   1. Unpin a single file
   2. Unpin multiple files
   3. Unpin single directory
   4. Unpin multiple directories

**Screenshots (optional)**
Add screenshots here.


yaira2 (approved):
LGTM

PR #12229, opened 2023-04-28
Fix: Fixed an issue where checkbox settings weren't applied in the grid layout

### LINKED ISSUE #12181: Bug: Show checkboxes when selecting items isn't applying in the grid layouts

It only work in details mode,this is bug?

### Steps to reproduce

- Turn off option to show checkboxes when selecting items
- Switch to grid layout
- Select an item
- Checkbox is visible 

### Steps to fix
Only show checkbox when explicitly hovering over it

### Files Version

2.4.67

### Windows Version

10.0.22621.1485

### Comments

_No response_

Krytan:
proof
This change resolves #12181 

**Validation**
1. Turn off option to show checkboxes when selecting items
2. Switch to grid layout
3. Select an item
4. Checkbox is not  visible


src/Files.App/Assets/FilesOpenDialog/FilesLauncher.exe
yaira2, line 1: This file shouldn't have been changed, must have been auto generated by Visual Studio.

Krytan: accidently commited the build version I gonna commit again without it

hishitetsu (changes requested):
This hides checkboxes even when the checkbox option is on and the item is hovered over. The behavior should not change when the option is on.

@yaira2 do you think checkboxes should not be displayed at all when the option is off? If so, the same modification should be made to the details layout.

Krytan:
> This hides checkboxes even when the checkbox option is on and the item is hovered over. The behavior should not change when the option is on.
> 
> @yair100 do you think checkboxes should not be displayed at all when the option is off? If so, the same modification should be made to the details layout.

In not grid layout mode when you turn off the "see checkboxes" settings you don't see the checkbox at all so I suggest the same applying for grid layout otherwise it can be a bit confusing.

I can see hovering in an item is not displaying the checkboxes when the settings are on  that's because I removed the mouse pointer that is also something you don't see in the not grid layout but I guess is something nice to have so I gonna add it back

hishitetsu:
In the detail layout, a checkbox appears when you hover over a thumbnail. Originally, the checkbox would appear when the cursor was hovered anywhere on the item, but this behavior was changed to the current one in response to user feedback that the thumbnail was frequently hidden.
The current checkbox option determines whether the checkbox is always displayed when an item is selected, and the checkbox feature itself is available whether this option is on or off.

Krytan:
> In the detail layout, a checkbox appears when you hover over a thumbnail. Originally, the checkbox would appear when the cursor was hovered anywhere on the item, but this behavior was changed to the current one in response to user feedback that the thumbnail was frequently hidden. The current checkbox option determines whether the checkbox is always displayed when an item is selected, and the checkbox feature itself is available whether this option is on or off.

I see I can make the same modifications on details as well so it only show the boxes when the settings to show check-boxes is set and then rename the setting to "show checkboxes" ? and when the setting is ON stick with checkbox is always displayed when an item is selected this seems like a good approach 

the best approach in my opinion would  be in the details layout having the checkbox on the left side of the thumbnail instead of on top of it so the thumbnail doesn't appear hidden when an item is selected it seems that the current option is only useful for detail layout

Krytan:
Screenshot 2023-04-29 183708

I made some modifications so I could show an example of my previous reply

yaira2:
@hishitetsu the expected behavior is to still show the checkboxes if the user hovers directly over the checkbox area. The details layout has the correct behavior already, this PR should help the grid layouts have the same behavior.

hishitetsu:
Okay, so this PR is trying to do something different.
This PR appears to completely disable the checkbox when the option is off.

In the details layout, there are thumbnails as markers for the location of checkboxes, but in the grid layout, there are no markers. If checkboxes are displayed only when the cursor is over the checkbox area, I think users will not know where the checkboxes are.

Krytan:
Exactly it is confusing when I first used the app I had no idea that in the details layout check boxes are displayed only when the cursor is over the thumbnail/icon I only found out that when explicit explained which makes me believe that  the majority of people won't know that either now for grid layout to have the same behavior it has to have a marker / a checkbox area somewhere.

yaira2:
> This PR appears to completely disable the checkbox when the option is off.

I'm not against this idea but I'm with @hishitetsu that we need to apply the same behavior to the details layout.

0x5bfa (commented):
Can you use XamlStyler? We are currently using a xaml styling extension named XamlStyler and we cannot approve until the annotations aren't fixed.

Depends on your visual studio version.
[XamlStyler for 2019](https://marketplace.visualstudio.com/items?itemName=TeamXavalon.XAMLStyler)
[XamlStyler for 2022](https://marketplace.visualstudio.com/items?itemName=TeamXavalon.XAMLStyler2022)

Krytan:
> 

Thanks I will do that

Krytan:
unknown_2023 04 30-17 53

To test:
   1. Switch to grid layout
   2. Turn off option to show checkboxes when selecting items
   3. Hover over an item with the cursor (shouldn't show any checkbox until your cursor is in the marker/the checkbox location)
   
Should work just like details layout

yaira2:
@Krytan thanks! This is what I was looking for 👍 

If we get feedback that it's still confusing users, we can hide the checkbox on hover as well but I'd like to try it this way first.

src/Files.App/Assets/FilesOpenDialog/FilesLauncher.exe
yaira2: Can you revert this?

yaira2:
@Krytan can you apply these changes to the tiles layout as well?

Krytan:
@yair100 Tiles should behave the same as well now. I absolutely agree on the idea of taking changes slowly and gathering feedback before making a step. let me know if everything works as intended

yair100 (dismissed):
LGTM

Will merge once the changes to the launcher are reverted.

Krytan:
Should be good now

src/Files.App/Views/LayoutModes/GridViewBrowser.xaml.cs
```diff
@@ -481,37 +483,35 @@ private void Grid_Loaded(object sender, RoutedEventArgs e)
...
 		{
 			if (sender is GridViewItem control && control.FindDescendant<UserControl>() is UserControl userControl)
 			{
-				// Save pointer over state accordingly
-				if (isPointerOver.HasValue)
-					control.SetValue(IsPointerOverProperty, isPointerOver);
 				// Handle visual states
 				// Show checkboxes when items are selected (as long as the setting is enabled)
-				// Show checkboxes when hovering of the item (regardless of the setting to hide them)
+				// Show checkboxes when hovering of the thumbnail (regardless of the setting to hide them)
```
yaira2, line 507:
```suggestion
				// Show checkboxes when hovering over the check box area (regardless of the setting to hide them)
```

yair100 (approved):

yaira2:
@Krytan thank you, congratulations on your first pull request!

Krytan:
@yair100 I have to thank you! going deep on the project and understanding how the code works really was a pleasure experience I can't describe the emotions involved and to be abbe to contribute at the end man I just wanna do more of this!

PR #12238, opened 2023-04-30
Code Quality: Update UWPToWinAppSDKUpgradeHelpers

0x5bfa:
### Motivation and Context

This file was for helping developers who wanted to migrate UWP to WASDK. But actual function is Interop PInvoke service.

### PR Checklist

### Screenshots

_None_


0x5bfa:
Ready for review.

src/Files.App/Helpers/InteropHelpers.cs
```diff
@@ -3,59 +3,60 @@
...
+
+			type.InvokeMember(
+				"ProtectedCursor",
+				BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.SetProperty | BindingFlags.Instance,
+				null,
+				uiElement,
+				new object[] { cursor }
+			);
 		}
 	}
```
hez2010, line 51: Should we get rid of this reflection call and change it to call ProtectedCursor setter from UIElement derived types directly instead?

0x5bfa: Might as well avoid unrelated discussion.

hez2010: Yes. That's why I didn't leave this as a code review.

src/Files.App/Helpers/InteropHelpers.cs
```diff
@@ -3,59 +3,60 @@
...
-		IntPtr GetForWindow([In] IntPtr appWindow, [In] ref Guid riid);
-
-		void ShowShareUIForWindow(IntPtr appWindow);
-	}
-
 	public static class InteropHelpers
 	{
 		public static readonly Guid DataTransferManagerInteropIID =
-			new Guid(0xa5caee9b, 0x8708, 0x49d1, 0x8d, 0x36, 0x67, 0xd2, 0x5a, 0x8d, 0xa0, 0x0c);
+			new(0xa5caee9b, 0x8708, 0x49d1, 0x8d, 0x36, 0x67, 0xd2, 0x5a, 0x8d, 0xa0, 0x0c);
```
yaira2, line 14: Let's keep `Guid`

0x5bfa: That can be simplified. however?

yaira2: We usually only simplify it if it's already clear what the type is. I just noticed that it's defined on line 13 so disregard what I said.

yaira2 (approved):
LGTM

PR #12228, opened 2023-04-28
Fix: Fixed a crash with ftp connections

ferrariofilippo:
**Resolved / Related Issues**
- [ ] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.

**Motivation**
When trying to access `ftp paths`, Files crashed because of a `stack overflow exception`. This PR prevents that 

**Validation**
How did you test these changes?
- [x] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [x] Did you implement any design changes to an existing feature?
   - [ ] Was this change approved?
- [x] Are there any other steps that were used to validate these changes?
   1. Open app
   2. Type and `ftp path` in the address bar

PR #12242, opened 2023-04-30
Feature: Display conflicts modal when pasting in FTP locations

### LINKED ISSUE #8609: Feature: Display conflicts modal when pasting on FTP locations
### What's the Problem?

When copy/pasting files the default behavior is to simply append " - copy" to the end of the file names, more often than not this is not the intended action on the users behalf. The frequency that a user wishes to copy over, or even cancel the entire operation is enough to warrant this being an issue.

### Requirements

- Display conflicts modal when there are conflicts on network shares

ferrariofilippo:
**Resolved / Related Issues**
   Closes #8609 
   Related #12228 (Files would still crash if the root `/` wasn't specified [e.g. `ftp://localhost:21`])

**Changes**
- There was a duplicated class that was causing issues (`FtpManager`), so I left only one
- I created `IFtpStorageService` which is essentially a copy of `IStorageService`. I might remove it and inject the service without its relative interface. @yair100 what are your thoughts on this?

**Limitations**
~Pasting from FTP to FTP isn't showing the conlicts dialog (I'm working on this)~ Fixed

**Validation**
   1. Open app and an FTP connection
   2. Paste an item with a conflicting name


yair100:
> I created IFtpStorageService which is essentially a copy of IStorageService. I might remove it and inject the service without its relative interface. @yair100 what are your thoughts on this?

Can you make `FtpStorageService` inherent from `IStorageService`?

ferrariofilippo:
I think it would cause issues once injected. We would have two different instances of the interface

yaira2:
What are the differences between the services?

ferrariofilippo:
> What are the differences between the services?

There are no differences, but we would inject two IStorageService instances, Native and Ftp. I think this might cause some issues when retrieving them. I'll try and see what happens

yaira2:
I meant, why did you create IFtpStorageService?

ferrariofilippo:
> I meant, why did you create IFtpStorageService?

Because I thought that if I have to retrieve it as IStorageService, but I have two instances of that (NativeStorageService and FtpStorageService), there would be issues since it may give me once Native and once Ftp
The two classes implement the interfaces but do not extend it

d2dyno1:
Maybe instead of duplicating `IStorageService` we could reuse it? I can think of two ways of doing it:
1. Use FtpStorageService in NativeStorageService and redirect calls to the former one when appropriate.
2. Manually inject (through the constructor) FtpStorageService instead of Ioc

I'm more into the second option because the first one would introduce more mess. The only difficulty with the second option is the injection itself -- the UI would need to inject appropriate implementation into the View Model

@ferrariofilippo

ferrariofilippo:
> Perhaps you have something else on your mind? @ferrariofilippo

Not really, I'm not really an expert on this, that's why I used this bad "hack"

ferrariofilippo:
> 2. Manually inject (through the constructor) FtpStorageService instead of Ioc

I need `FtpStorageService` in a static method, how should I manage it?

d2dyno1:
You could inject it by passing it as a parameter however, then we'd have no guarantee that appropriate service was passed. Unless a major refactor in this area is done, that's going to be an issue. We could perhaps continue with your approach, but to bit reduce code duplication and increase reusability you could inherit from `IStorageService`
```cs
public interface IFtpStorageService : IStorageService
```

d2dyno1 (approved):

Issue #12220, opened 2023-04-27
Bug: Files crashes when opening properties from search page


ferrariofilippo:
### Description

If you open a new tab on `Home` and you start searching anything (like tagged files), the properties button will be enabled on the landing page. Clicking on it will crash the app

### Steps To Reproduce

1. Open a new tab on `Home` (this won't happen if you navigate anywhere. The tab must open on `Home`)
2. Click on a tag on the sidebar (or search for anything)
3. See that the `Properties` button on the toolbar is enabled
4. Click on it
5. See that Files crashes

### Requirements

- The properties button should be disabled when in a search results

### Files Version

2.4.69.0

### Windows Version

Windows 11 Pro 22H2 10.0.22621.1555

### Log file

[debug.log](https://github.com/files-community/Files/files/11341856/debug.log)

yaira2:
@ferrariofilippo thank you for reporting this, we need to update the rich command to be disabled on the search page.

yaira2:
It might also be a good idea to add some error handling in the properties widow as well.

ferrariofilippo:
> @ferrariofilippo thank you for reporting this, we need to update the rich command to be disabled on the search page.

We don't have one at the moment

PR #12269, opened 2023-05-03
Rich Command: Open Properties

0x5bfa:
### Description 

As the title.

### Motivation and Context

Obvious reasons.

### PR Checklist
  - Related #11481


src/Files.App/Actions/Open/OpenPropertiesAction.cs
```diff
@@ -0,0 +1,47 @@
...
+
+		public string Label { get; } = "OpenProperties".GetLocalizedResource();
+
+		public string Description => "OpenPropertiesDescription".GetLocalizedResource();
+
+		public RichGlyph Glyph { get; } = new RichGlyph(opacityStyle: "ColorIconProperties");
+
+		public bool IsExecutable =>
+			context.ShellPage.InstanceViewModel.IsPageTypeNotHome &&
+			context.HasSelection;
```
QuaintMako, line 21: Shouldn't the command be usable on the Recent Files, which is on the Home Page?

src/Files.App/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -199,14 +199,13 @@
 					x:Name="PropertiesButton"
 					Width="Auto"
 					MinWidth="40"
-					AccessKey="O"
```
ferrariofilippo, line 202: I think we should leave the accessibility key

src/Files.App/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -199,14 +199,13 @@
...
-					Command="{x:Bind ViewModel.PropertiesCommand, Mode=OneWay}"
-					IsEnabled="{x:Bind converters:MultiBooleanConverter.AndConvert(ViewModel.CanViewProperties, ViewModel.InstanceViewModel.IsPageTypeNotHome), Mode=OneWay, FallbackValue=False}"
-					Label="{helpers:ResourceString Name=Properties}"
+					Command="{x:Bind Commands.OpenProperties}"
+					IsEnabled="{x:Bind Commands.OpenProperties.IsExecutable, Mode=OneWay}"
+					Label="{x:Bind Commands.OpenProperties.Label}"
 					LabelPosition="Collapsed"
-					ToolTipService.ToolTip="{helpers:ResourceString Name=Properties}">
-					<local:OpacityIcon Style="{StaticResource ColorIconProperties}" />
+					ToolTipService.ToolTip="{x:Bind Commands.OpenProperties.LabelWithHotKey}">
```
ferrariofilippo, line 207: I think you should add `Mode=OneWay`. At the moment nothing will change, but it will prevent future issues (hotkey not updating) once we introduce the remapping feature

ferrariofilippo (changes requested):
There are a couple of things I would change

src/Files.App/Actions/Open/OpenPropertiesAction.cs
ferrariofilippo: We should handle it very carefully, since it may cause bugs (see #12220)
Currently, if you have nothing selected you can click properties button to open parent folder ones. The condition above is blocking that behaviour

src/Files.App/UserControls/InnerNavigationToolbar.xaml
0x5bfa: Done

src/Files.App/Actions/Open/OpenPropertiesAction.cs
0x5bfa: Then always executable?

ferrariofilippo: I think we should check only if page is not home.
We should handle recent files differently, as we would do for sidebar items

src/Files.App/Actions/Open/OpenPropertiesAction.cs
```diff
@@ -0,0 +1,46 @@
...
+
+		public string Description => "OpenPropertiesDescription".GetLocalizedResource();
+
+		public RichGlyph Glyph { get; } = new RichGlyph(opacityStyle: "ColorIconProperties");
+
+		public bool IsExecutable =>
+			!context.ShellPage.InstanceViewModel.IsPageTypeSearchResults;
+
+		public HotKey HotKey { get; } = new(Keys.P, KeyModifiers.Ctrl);
+
```
ferrariofilippo, line 23: This shortcut is already assigned to `Open preview pane`

yaira2: Properties shouldn't have a shortcut by default

hecksmosis: In Explorer its `Alt+Enter`

yaira2: Oh that's right, we might actually use that already.

0x5bfa: >In Explorer it's Alt+Enter

How can I specify? I don't know which enums represent them.

hecksmosis: You should use `new HotKey(Keys.Enter, KeyModifiers.Menu)`

0x5bfa:
Abandon.

Issue #12037, opened 2023-04-10
Bug: Browser crash, preview, PDF, clicking the arrows


chezfo:
### Description

When a PDF is previewed, when you click the arrows to move from page to page, the browser crashes.
![1אנימציה](https://user-images.githubusercontent.com/85172877/230847891-a0bb08ac-22eb-43a2-bd40-13d253a4ddb6.gif)

### Steps To Reproduce

1. Stand on a PDF file
2. work for preview
3. Click the arrows to move between pages

### Requirements

Fix crash when clicking arrows.

### Files Version

Browser crash, preview, PDF, clicking the arrows

### Windows Version

11

### Log file

[debug.log](https://github.com/files-community/Files/files/11188494/debug.log)

Josh65-2201:
Can you see if this is an issue in 2.4.62.0 https://files.community/download (Select preview) and post the updated log if it happens.

yaira2:
@0x5bfa is this the same error you reported in #10280?

chezfo:
> Can you see if this is an issue in 2.4.62.0 https://files.community/download (Select preview) and post the updated log if it happens.

For several hours now, I have not been able to install the program in preview.
 Is there another way to install?

Josh65-2201:
You can try
1. Opening PowerShell as administrator
2. Run this command
```Add-AppxPackage -AppInstallerFile https://cdn.files.community/files/preview/Files.Package.appinstaller```

0x5bfa:
> @0x5bfa is this the same error you reported in https://github.com/files-community/Files/issues/10280?

Looks another one.

0x5bfa:
@chezfo Can you write 'Step To Reproduce' and 'Requirements' in English?

yaira2:
@chezfo Thank you for bringing this to our attention, we've added it to our project board for tracking.

chezfo:
> Can you see if this is an issue in 2.4.62.0 https://files.community/download (Select preview) and post the updated log if it happens.
The problem continues, at 66
 But the browser does not crash but reopens.
[debug.log](https://github.com/files-community/Files/files/11286342/debug.log)


By the way, I installed a browser that is not from the Windows store, and the browser is significantly faster.  Is it because of the version or because I installed without Windows Store?

Josh65-2201:
Files uses Webview2 which is built in to Widows for the browser.

yaira2:
@chezfo is this only with certain pdf documents?

[...]

Issue #8282, opened 2022-02-10
Feature: Add support for dragging recent files


TheBossMagnus:
### What's the Problem?

You can't drag a file from recent files

### Solution/Idea

Add support for dragging items from Recent Files widget into other programs

devovercome:
Hello!
May I take this issue?

I am working around the dragging stuff, gonna make a PR for the [Issue #8677](https://github.com/files-community/Files/issues/8677) and go for this feature afterwards.

yaira2:
Thank you for your feedback, I added this to the todo list.

Krytan:
Working on this task

yaira2:
We should be able to try this again now that #16628 has been merged.

PR #12293, opened 2023-05-06
Fix: Fixed crash that would occur when previewing PDFs in a RTL environment

Krytan:
image

In the picture you see the pdf priview in a RTL(Right To Left ) windows inviroment that's why the arrow start in the left side this is how people that speaks for example arabic, aramaic and hebrew see.
The bug was when viewing a PDF file using RTL and clicking in the left and right arrows the program would crash.

**Validation**
   A proper test would be using a windows environment with RightToLeft direction 
   You can simulate RTL  by going to pdfpreview.xaml And setting the FlowDirection to RightToLeft in FlipView and PipsPager 
   It should normally Crash if you don't use my fix version.

   1. Stand on a PDF file
   2. Have the preview panel open 
   3. Click the arrows to move between pages
   4. Should work fine in both LTR and RTL inviroments 

This should fix this bug #12037


Krytan:
Useful information about the issue and changes: : The crash seems to be related to the default arrows in the flip-view not functioning properly when using a right-to-left FlowDirection. By default, the flowdirection is set to auto to support both left-to-right and right-to-left systems. However, the default arrows cannot be accessed or customized, so I created custom arrows that function similarly to the default arrows but with added functionality for right-to-left systems.

To test My changes, I changed the flowdirection from left-to-right (which works properly) to right-to-left (which was previously causing crashes), and it now works without crashing.

Krytan:
Note:  1. arrows are shown when the preview tab is open by default, in a pdf file 
              before: they would only show if you hover the cursor over the pdf inside the preview tab
           
           2. Arrows layout is slight different from the default one.

Everything else should be the same

yaira2:
@Krytan can you confirm if this is a platform issue with the flip view? I think a cleaner solution would be to hardcode the flow direction for the flip view so that it's left to right.

Krytan:
oh aright yes that's correct since I could fix the issue by making a custom flipview rather than using the default flipview control buttons.

yaira2 (approved):
LGTM

Krytan:
@yaira2 Hello Yaira long time, can you assign me a new task I am looking to test my skills and continue contributing for files

yaira2:
@Krytan thank you for reaching out, do you want to work on #8282?

Krytan:
@yaira2 Sure,  it seems interesting thanks for quickly replying me back, I will try my best on the task

PR #12307, opened 2023-05-08
Fix: Fixed crash that would occur in git directories on WSL drives

### LINKED ISSUE #12292: Bug: Crash due to git directory on WSL drive
### Description

When trying to open directory in WSL that contains .git, Files crashes. This is a new issue with the new git integration.

### Steps To Reproduce

1. Go to WSL
2. Click to open a directory containing a .git directory
3. Crash

### Requirements

Fix crash that occurs when trying to open git directories on WSL.

### Files Version

2.4.80.0

### Windows Version

10.0.22621.1635

### Log file

[debug.log](https://github.com/files-community/Files/files/11412437/debug.log)

ferrariofilippo:
**Resolved / Related Issues**
   Closes #12292 

**IMPORTANT**
At the moment this PR only prevents the crash (branch name is still not displayed). The exception is raised because the repository belongs to a different user. We can't show the branch name/interact with the repository unless the user set it as "safe".
I think we should show a modal to warn the user about this. We might also provide the command to resolve the issue (`git config --global --add safe.directory = path`) or show a button to execute that

**Validation**
   1. Open a git repository in WSL

### DIFF
--- a/src/Files.App/Helpers/GitHelpers.cs
+++ b/src/Files.App/Helpers/GitHelpers.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See the LICENSE.
 
 using LibGit2Sharp;
-using System;
 using Files.App.Filesystem.StorageItems;
 
 namespace Files.App.Helpers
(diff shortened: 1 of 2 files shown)


yaira2:
> I think we should show a modal to warn the user about this. We might also provide the command to resolve the issue (git config --global --add safe.directory = path) or show a button to execute that

It would be less intrusive if we display a message on the status bar but either way would work.

yaira2 (approved):
LGTM

Issue #12340, opened 2023-05-12
Code Quality: Continued work to clean up entire project


0x5bfa:
### Description

This is a milestone to display what PR is related the cleaning-up plan. Rationale is that Files source codebase don't have consistency. All of us absolutely must be forced to follow the guideline to keep code clean. I will clean up the project with the strict order I defined so that I would not break any existing functionalities. 

### Concerned code

I have following stages to clean up:
1. Rename files with the appropriate name and put them into appropriate namespace for each.
2. Rename variables with preferred name to follow the guideline (The most likely to happen is that I rename property-backing-field from '_isHidden' to '_IsHidden')
3. Reorder variables & methods to follow the class structure guideline (Section 3 in the guildeline)
4. Add/Update/Improve C# XML Comments and double-slashed comments.
6. Refactor to improve performance. You can see what I will work on in Requirements section.

- Files.App.**
- Files.Shared
- Files.Backend

### Gains

- A better readability.
- A better maintainability.
- Faster build of Files.App

### Requirements

- Force to follow the guideline in anyway
- Add sealed accessor to entire Files.App except classes from which other classes have to be deriven.
- Add internal accessor to entire Files.App
- Append Async to the end of async method name
- Add init accessor to properties which have to be readonly
- Use Utils namespace for utilities, manangers, and watchers
- Use Lazy class for delay initialization
- Avoid boxing

yaira2:
> Append Async to the end of async method name

This is already being worked on

yaira2:
> Entire project

How do you want to split this up?

JaiganeshKumaran:
> The most likely to happen is that I rename property-backing-field from '_isHidden' to '_IsHidden'
I don't this is the recommended convention.

0x5bfa:
Please note that this issue's purpose is to force codebase to follow the guideline for great consistency.

> from '_isHidden' to '_IsHidden'

This is on the guideline. Can you open an issue for that?

yaira2:
Merging with #4180

PR #12342, opened 2023-05-12
Code Quality: Move some ViewModel classes into the appropriate namespaces

0x5bfa:
### Description

The first step to clean up ViewModels.

### PR Checklist
   Related #12340

### Screenshots

_None_

### DIFF
--- a/src/Files.App/ViewModels/UserControls/StatusCenterViewModel.cs
+++ b/src/Files.App/ViewModels/UserControls/StatusCenterViewModel.cs
@@ -4,7 +4,7 @@
 using Microsoft.UI.Xaml.Controls;
 using System.Windows.Input;
 
-namespace Files.App.ViewModels
+namespace Files.App.ViewModels.UserControls
 {
 	public class OngoingTasksViewModel : ObservableObject, IOngoingTasksActions
 	{
--- a/src/Files.App/ViewModels/UserControls/StatusCenterViewModel.cs
+++ b/src/Files.App/ViewModels/UserControls/StatusCenterViewModel.cs
@@ -1,15 +1,10 @@
 // Copyright (c) 2023 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using Microsoft.UI.Xaml.Controls;
-using System.Windows.Input;
-
 namespace Files.App.ViewModels.UserControls
 {
 	public class OngoingTasksViewModel : ObservableObject, IOngoingTasksActions
 	{
-		// Public Properties
-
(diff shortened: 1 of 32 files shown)


0x5bfa:
@d2dyno1 Can you review? I am sorry for pinging always I have codebase updates.

src/Files.App/ViewModels/UserControls/StatusCenterViewModel.cs
```diff
@@ -0,0 +1,140 @@
...
+			OnPropertyChanged(nameof(AnyOperationsOngoing));
+			OnPropertyChanged(nameof(InfoBadgeState));
+			OnPropertyChanged(nameof(InfoBadgeValue));
+		}
+
+		public void UpdateMedianProgress()
+		{
+			if (AnyOperationsOngoing)
+				MedianOperationProgressValue = (int)StatusBannersSource.Where((item) => item.IsProgressing).Average(x => x.Progress);
+		}
```
ferrariofilippo, line 138: I know this might be out of scope, but I think we should rename `MedianOperationProgressValue` to `AverageOperationProgressValue`. We are calculating the average value not the median one

ferrariofilippo (commented):
Nothing seems to miss, but still there are some style-related changes that you should consider

0x5bfa:
@ferrariofilippo I'm currently working on step 1 in #12340.

> 1. Rename files with the appropriate name and put them into appropriate namespace for each.
> 2. Rename variables with preferred name to follow the guideline (The most likely to happen is that I rename property-backing-field from '_isHidden' to '_IsHidden')
> 3. Reorder variables & methods to follow the class structure guideline (Section 3 in the guildeline)
>4. Refactor to improve performance. You can see what I will work on in Requirements section.

0x5bfa:
@ferrariofilippo But thank you always for reviewing my refactoring PR. I will refer to those reviews in the further PR.

@yair100 Ready for review.

ferrariofilippo (approved):
I checked it again and nothing is missing, so LGTM!
I found another couple of things that you may change in phase 2:
- At `src/Files.App/ViewModels/UserControls/StatusCenterViewModel.cs` `CloseBanner()`, it might be better to use `IndexOf` and then `RemoveAt` to avoid comparing all the elements twice
- At `src/Files.App/Data/Items/StatusBanner.cs StatusBanner()`, line 85, it might be better to use a swtch assing like: `Title = Operation switch ...`

yaira2:
> I checked it again and nothing is missing, so LGTM!

Thanks! Can I resolve the conversations?

Issue #12312, opened 2023-05-09
Mica


Jay-o-Way:
Now in 1.3
https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.media.micabackdrop

Jay-o-Way:
That was quick. Did I do something wrong?

yaira2:
Sorry for the confusion, in order to properly manage the issues, we have to close issues that don't follow the issue templates.

PR #12359, opened 2023-05-15
GitHub: Create config.yml

Jay-o-Way:
This will prevent any blank issues to be created. As seen in https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

**Resolved / Related Issues**
> in order to properly manage the issues, we have to close issues that don't follow the issue templates.

https://github.com/files-community/Files/issues/12312#issuecomment-1540577114

**Validation**
- Neither needed or possible :)

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,1 @@
+blank_issues_enabled: false


Jay-o-Way:
Having that said: What if there is an issue that really does not fit into one of the three [bug, code, feature] templates? For example: I would like to polish MD files, with help of MS [Learn Authoring Pack](https://learn.microsoft.com/contribute/how-to-write-docs-auth-pack) (for VS Code). What to do then?

yaira2 (approved):
I was looking for a way to do this but wasn't able to figure it out, thank you!

Jay-o-Way:
You're welcome. I _am_ curious what you would advise for my question above here, though.

yaira2:
> Having that said: What if there is an issue that really does not fit into one of the three [bug, code, feature] templates? For example: I would like to polish MD files, with help of MS [Learn Authoring Pack](https://learn.microsoft.com/contribute/how-to-write-docs-auth-pack) (for VS Code). What to do then?

Feature is best

PR #12365, opened 2023-05-16
Fix: Fixed Files crashes when opening properties from search page

ferrariofilippo:
**Resolved / Related Issues**
   Related #12220

This PR will only prevent the crash. I posted a fix to disable the button in #12269 

**Validation**
   1. Open a new tab on `Home` (this won't happen if you navigate anywhere. The tab must open on `Home`)
   2. Click on a tag on the sidebar (or search for anything)
   3. See that the `Properties` button on the toolbar is enabled
   4. Click on it
   5. See that Files doesn't crash anymore

### DIFF
--- a/src/Files.App/Helpers/FilePropertiesHelpers.cs
+++ b/src/Files.App/Helpers/FilePropertiesHelpers.cs
@@ -61,6 +61,9 @@ namespace Files.App.Helpers
 			{
 				// Instance's current folder
 				var folder = associatedInstance.FilesystemViewModel.CurrentFolder;
+				if (folder is null)
+					return;
+
 				item = folder;
 
 				var drivesViewModel = Ioc.Default.GetRequiredService<DrivesViewModel>();


yaira2:
Wouldn't it make more sense to disable the button if no files are selected?

ferrariofilippo:
Actually I like how it works now, if I'm in a folder it is faster if a click on that button, instead of navigating up and then right clicking on the folder

yaira2:
I mean for search results

ferrariofilippo:
@yaira2 is it ok if I create `Open Properties` rich command and I push it here?

src/Files.App/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -201,12 +201,13 @@
...
-					Label="{helpers:ResourceString Name=Properties}"
+					Command="{x:Bind Commands.OpenProperties, Mode=OneWay}"
+					IsEnabled="{x:Bind Commands.OpenProperties.IsExecutable, Mode=OneWay}"
+					KeyboardAcceleratorTextOverride="{x:Bind Commands.OpenProperties.HotKeyText, Mode=OneWay}"
+					Label="{x:Bind Commands.OpenProperties.Label}"
 					LabelPosition="Collapsed"
-					ToolTipService.ToolTip="{helpers:ResourceString Name=Properties}">
-					<local:OpacityIcon Style="{StaticResource ColorIconProperties}" />
+					ToolTipService.ToolTip="{x:Bind Commands.OpenProperties.LabelWithHotKey, Mode=OneWay}">
+					<local:OpacityIcon Style="{x:Bind Commands.OpenProperties.OpacityStyle}" />
```
yaira2, line 210: Can you set the mode? I think OneTime is okay but it's a good idea to test first.

yaira2 (dismissed):
LGTM

src/Files.App/UserControls/InnerNavigationToolbar.xaml
yaira2: I meant for `OpacityIcon` because it never changes.

ferrariofilippo: Oh ok 👍

ferrariofilippo: It works

PR #12444, opened 2023-05-24
Fix: Fixed an issue where the format option was incorrectly shown

### LINKED ISSUE #12401: Bug: "Format..." shouldn't be visible if items are selected

As title. The top option is "Format" instead of "Open" / "Open with". This could cause a mistaking operation.
P.S. This happens only in the root directory of a drive.
image

### Requirements

- "Format..." should not display in item context menu at all (since it only relates to the drive itself, not items in it), or at least stay at the second or the third part of context menu.

### Files Version

2.4.82.0 (Preview)

### Windows Version

10.0.22621.1702

### Comments

_No response_

tsurrdurr:
**Resolved / Related Issues**
   Closes #12401 

**Validation**
   1. Navigate to a **non-system** hard drive's root folder.
   2. Right click on any file or folder. Without the fix, it would contain a "Format..." option.
   3. Right click on an empty space. The option "Format..." should be still available there.


yaira2:
@tsurrdurr apologies for the delay on reviewing this, can you resolve the merge conflicts?

yaira2 (approved):
LGTM

Issue #12432, opened 2023-05-23
Feature: Add support for PowerToys Peek


cartifanwlr:
### What feature or improvement do you think would benefit Files?

PowerToys added the Peek feature in 0.70. It's essentially macOS QuickLook, allowing you to preview files with Ctrl+Space. This does not work in Files, and I'd like to see this feature supported in Files.

### Requirements

- PowerToys Peek should work inside Files

### Files Version

2.4.71.0

### Windows Version

Windows 11 version 22H2 build 22624.1755

yaira2:
Thank you for your feedback, Files already supports QuickLook so it shouldn't be too hard if PowerToys has an api for this.

hishitetsu:
I don't know if Peek has an API, but we need to think about which hotkey to assign if we implement this. We already assign Ctrl+Space to toggle selection. Should we add an option to select a tool to preview items?

yaira2:
> Should we add an option to select a tool to preview items

This is probably the best solution, this way we can have a single action for preview items.

yaira2:
I opened an issue on the PowerToys repo to start the discussion on creating an api for third party file managers https://github.com/microsoft/PowerToys/issues/2

yaira2:
Thank you for your feedback. We're closing this issue because PowerToys doesn't have an API for this, but we'll be happy to reevaluate if this changes in the future.

yair100:
This feature will be included in the next update. Please see #17954.

PR #12447, opened 2023-05-24
Code Quality: Created PreviewPopupService

yaira2:
**Resolved / Related Issues**
   Related to #8052 and #12432

**Validation**
   1. Install Quick Look
   2. Use spacebar to launch the preview

**Screenshots (optional)**
Add screenshots here.


src/Files.Backend/Services/IPreviewPopupService.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+	public interface IPreviewPopupService : INotifyPropertyChanged
+	{
+		/// <summary>
+		/// Opens preview popup
+		/// </summary>
+		/// <param name="path"></param>
+		/// <param name="switchPreview"></param>
+		/// <returns></returns>
+		Task OpenPreviewPopup(string path, bool switchPreview = false);
```
hishitetsu, line 18: Shouldn't SwitchPreview be a separate API?

yaira2: The only difference (at least with QuickLook) is the message so it can help reduce duplicate code. If you think it's worth it, I can split them up.

yaira2: It looks like SeerPro works differently so it'll make sense to split it.

hishitetsu: Yes I think they should be split. It is confusing that the popup doesn't actually open when switchPreview is true.

yaira2: Done

src/Files.App/ServicesImplementation/QuickLookPreviewPopupService.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+	/// <inheritdoc cref="IPreviewPopupService"/>
+	internal sealed class QuickLookPreviewPopupService : ObservableObject, IPreviewPopupService
+	{
+		private const int TIMEOUT = 500;
+		private static string pipeName = $"QuickLook.App.Pipe.{WindowsIdentity.GetCurrent().User?.Value}";
+		private static string pipeMessageSwitch = "QuickLook.App.PipeMessages.Switch";
+		private static string pipeMessageToggle = "QuickLook.App.PipeMessages.Toggle";
+
+		public async Task TogglePreviewPopup(string path)
```
hishitetsu, line 20: I think duplicate code can be reduced if there is a private method that has the message parameter and call it from the public API methods.

src/Files.App/ServicesImplementation/QuickLookPreviewPopupService.cs
```diff
@@ -0,0 +1,92 @@
...
+		{
+			bool isQuickLookAvailable = await DetectAvailability();
+			if (!isQuickLookAvailable)
+				return;
```
hishitetsu, line 24:
```suggestion
			if (!isQuickLookAvailable)
			{
				App.Logger.LogInformation("QuickLook not detected");
				return;
			}
```

hishitetsu: It certainly wouldn't be a problem without it.

hishitetsu (approved):
LGTM

hishitetsu (approved):
LGTM

PR #12448, opened 2023-05-24
Fix: Fixed TabView visual regression from recent WinUI update

chingucoding:
**Resolved / Related Issues**
- [ ] Fixes a regression in WinUI 3 with the TabView control.

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App (Package)/Files.Package.wapproj
+++ b/src/Files.App (Package)/Files.Package.wapproj
@@ -107,7 +107,7 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.3.230502000" IncludeAssets="build" />
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.4.230518007-experimental1" IncludeAssets="build" />
     <PackageReference Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.22621.756" IncludeAssets="build" />
   </ItemGroup>
   <Import Project="$(WapProjPath)\Microsoft.DesktopBridge.targets" />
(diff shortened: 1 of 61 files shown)


src/Files.App (Package)/Files.Package.wapproj
```diff
@@ -107,7 +107,7 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.3.230502000" IncludeAssets="build" />
+    <PackageReference Include="Microsoft.WindowsAppSDK" Version="1.4.230518007-experimental1" IncludeAssets="build" />
```
yaira2, line 110: Do we need experimental? We want to release to stable in a few weeks and this will hold us back.

chingucoding: The issue only repos in experimental for me, but I can also revert this.

yaira2: I can repro with the latest in `main`.

chingucoding: Reverted.

0x5bfa (changes requested):
Can youo create Generic.xaml in ResourceDictionaries/* for those styles?

chingucoding:
Sorry for the delayed response @0x5bfa. I'm afraid we cannot move the styles outside in a separate ResourceDictionary due to the resolving order when using merged dictionaries. When moving them into a separate ResourceDictionary, multiple resources inside the templates fail to resolve. 
If you/we like, we can also pull those resources inside the file but I would like to discuss this before following through with that plan.

PR #12475, opened 2023-05-28
Project: Rename 'Files.App (Package)' with 'Files.Package'

0x5bfa:
### Motivation and Context

This dir name was made by microsoft upgrade assistant. The path should not contain symbols and spaces. I did not include unrelated changes.

- Rename dir name with File Explorer
- Update azure pipeline yml files
- Update other reference path

> **Note**
> It takes about 3 minutes to load all diffs.

### PR Checklist

### Screenshots

_None_


hez2010:
We used to use Files.Package, but Files.App (Package) looks more like the package of Files.App, so we end up with the current name. 
BTW Files.App (Package) is also the default name created by WinUI 3 template. 
@yaira2 What do you think?

0x5bfa:
> so we end up with the current name.

This folder name was created by the .NET Upgrade Assistant. And this PR is based on the previous suggest of project refactoring plan. People avoided to change folder, but to avoid using spaces & symbols, we must get rid of them.

0x5bfa:
Failing Tests

PR #12477, opened 2023-05-29
Feature: Added SeerPro support

yaira2:
**Resolved / Related Issues**
   Related #8052... (this PR doesn't include switching between files)

**Validation**
   1. Install and open SeerPro
   2. Use space bar to toggle preview from Files

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/ServicesImplementation/PreviewPopupProviders/SeerProProvider.cs
+++ b/src/Files.App/ServicesImplementation/PreviewPopupProviders/SeerProProvider.cs
@@ -2,28 +2,35 @@
 // Licensed under the MIT License. See the LICENSE.
 
 using Files.Backend.Services;
-using Microsoft.Extensions.Logging;
-using System.IO;
-using System.IO.Pipes;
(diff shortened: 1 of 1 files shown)


hishitetsu:
Shouldn't there be an option to switch between QuickLook and Seer?

yaira2:
I'm undecided, we'll definitely want an option once we add a built-in preview feature but until then, I'm not sure if there is a valid use case where a user has both apps installed at the same time.

hishitetsu:
There may be situations where the user would like to select a preview tool if we support PowerToys Peek, but it certainly may not be necessary just yet.

hishitetsu (approved):
LGTM

hishitetsu (approved):
LGTM

yaira2:
> There may be situations where the user would like to select a preview tool if we support PowerToys Peek, but it certainly may not be necessary just yet.


We can work around this by putting PowerToys Peek as the last if statement, this way, if another provider is detected we know it's preferred.

PR #12488, opened 2023-05-30
Feature: Show a tooltip when a text in the details view is trimmed

### LINKED ISSUE #12481: Feature: Re-add tooltip to details layout

#11962 removed tooltip from details layout. However, it is more convenient to have the full file name displayed on hover when the file name does not fit in the column. So I think tooltip is necessary even for details layout.
image

### Requirements

Re-add tooltip to details layout (if posible, only display when text is cut off)

### Files Version

2.4.71.0

### Windows Version

Windows 11 22H2

### Comments

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #12481 

**Validation**
   1. When a text in the details view is trimmed, a tooltip will be displayed when hovering.
   2. When a text in the details view is not trimmed, a tooltip will not be displayed.

**Screenshots**
image

### DIFF
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -804,5 +804,10 @@ namespace Files.App.Views.LayoutModes
 					VisualStateManager.GoToState(userControl, "HideCheckbox", true);
 			}
 		}
+
+		private void TextBlock_IsTextTrimmedChanged(TextBlock sender, IsTextTrimmedChangedEventArgs e)
+		{
+			ToolTipService.SetToolTip(sender, sender.IsTextTrimmed ? sender.Text : null);
+		}
 	}
 }
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -805,6 +805,7 @@ namespace Files.App.Views.LayoutModes
 			}
 		}
 
+// Workaround for https://github.com/microsoft/microsoft-ui-xaml/issues/170
 		private void TextBlock_IsTextTrimmedChanged(TextBlock sender, IsTextTrimmedChangedEventArgs e)
 		{
 			ToolTipService.SetToolTip(sender, sender.IsTextTrimmed ? sender.Text : null);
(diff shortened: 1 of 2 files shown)


yaira2:
Thank you for the PR, do you know if this affects performance?

hishitetsu:
> Thank you for the PR, do you know if this affects performance?

I tested a folder contains 3,000 files. I did not notice any particular performance degradation when changing column widths, so I think it should be fine.

yaira2:
I'm more concerned about larger directories (40k items)

hishitetsu:
> I'm more concerned about larger directories (40k items)

At least in my environment, there seems to be no performance impact even with 40k files.

yaira2:
I found this feature request on the WinUI repo https://github.com/Microsoft/microsoft-ui-xaml/issues/170

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -804,5 +804,10 @@ private void UpdateCheckboxVisibility(object sender, bool isPointerOver)
 					VisualStateManager.GoToState(userControl, "HideCheckbox", true);
 			}
 		}
+
```
yaira2, line 807:
```suggestion

// Workaround for https://github.com/microsoft/microsoft-ui-xaml/issues/170
```

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -804,5 +804,10 @@ private void UpdateCheckboxVisibility(object sender, bool isPointerOver)
...
+		private void TextBlock_IsTextTrimmedChanged(TextBlock sender, IsTextTrimmedChangedEventArgs e)
```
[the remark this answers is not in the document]
chingucoding: If we rename this, we also need to rename it in XAML. Given that multiple controls use it, I would actually say it makes more sense to leave name as is.

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -804,5 +804,10 @@ private void UpdateCheckboxVisibility(object sender, bool isPointerOver)
...
+		{
+			ToolTipService.SetToolTip(sender, sender.IsTextTrimmed ? sender.Text : null);
```
chingucoding, line 810: I assume, we only do this because of UX, not because of performance reasons?

yaira2: That's right, we only want the tooltip to show if the text is trimmed.

yaira2 (approved):
LGTM

Issue #4635, opened 2021-04-16
Feature: Attach version info when submitting issues reports


kimlimjustin:
**Is your feature request related to a problem?**
Submitting issues on GitHub is somehow troublesome especially to create bug report, which requires many OS information.

**Describe the solution**
Being inspired by vscode, create an issue submission form in Files, and automatically attach OS information and Files information.

**Requirements**
- Automatically fill out the version information when clicking the report issue button on the about page.

**Additional comment**
Found this resource https://docs.github.com/en/rest/reference/issues#create-an-issue, hope it's useful :)

yaira2:
@kimlimjustin thank you for the feedback, I added this to the project board.

yaira2:
This feature will be included in the next release.

PR #12506, opened 2023-05-31
GitHub: Update issue templates to make users copy Windows version from settings

MarcAnt01:
**Validation**
My own fork with issue template on

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -37,7 +37,7 @@ body:
   - type: input
     attributes:
       label: Windows Version
-      description: Which version of Windows are you using? Press <kbd>Win</kbd> + <kbd>Pause/Break</kbd> and copy.
+      description: Which version of Windows are you using? The version number can be found by going to Files' Settings > About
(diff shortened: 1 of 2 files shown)


MarcAnt01:
> On second thought, the about page doesn't actaully show the Windows version. It's only after pressing the copy button.

What would you like to be displayed there instead?

yaira2:
We could theoretically explain that the copy button copies the Windows version as well, but I'm concerned that'll just confuse the issue authors. It might be best to just leave the template as is.

MarcAnt01:
> We could theoretically explain that the copy button copies the Windows version as well, but I'm concerned that'll just confuse the issue authors. It might be best to just leave the template as is.

Or also add a further button to copy the Windows version

yaira2:
It's not high priority but #4635 will make it a lot easier.

0x5bfa:
Do we really need specific windows version in the first place? Why not we make dropdown for the selection of Windows 10/11, and of Dev/Preview/Stable, and of 21H1/21H2/22H1…?

yaira2:
A dropdown is a good idea, can you open a PR for this?

PR #12520, opened 2023-06-01
Fix: Fixed issue where add item was disabled in column layout

yaira2:
**Resolved / Related Issues**
   1. Fixed issue where "add item" action was disabled in secondary columns.
   2. Fixed issue where "add item' was disabled when items were selected (it should still be hidden in the right click menu)

**Validation**
   1. Open a second column in the columns layout
   3. Confirm the new/add item action is enabled

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Actions/FileSystem/AddItemAction.cs
+++ b/src/Files.App/Actions/FileSystem/AddItemAction.cs
@@ -24,7 +24,7 @@ namespace Files.App.Actions
 
 		public RichGlyph Glyph { get; } = new(opacityStyle: "ColorIconNew");
 
-		public bool IsExecutable => context.CanCreateItem && !context.HasSelection;
+		public bool IsExecutable => context.CanCreateItem;
 
 		public AddItemAction()
 		{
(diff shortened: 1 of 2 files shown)


src/Files.App/Actions/FileSystem/AddItemAction.cs
```diff
@@ -24,7 +24,7 @@ internal class AddItemAction : ObservableObject, IAction
 
 		public RichGlyph Glyph { get; } = new(opacityStyle: "ColorIconNew");
 
-		public bool IsExecutable => context.CanCreateItem && !context.HasSelection;
```
hishitetsu, line 27: This change affects more than just column layout. Are there any regression concerns with removing this? I don't know why the action is disabled while items are selected though.

yaira2: That's right, it also fixes an issue where the option was disabled when items were selected.

hishitetsu: @ferrariofilippo why did you add `!context.HasSelection` to `IsExecutable`?
https://github.com/files-community/Files/commit/f52fda7b09b2f197431e406150e7571fb3fa7697

yaira2: I assume it was to hide the option when right clicking on multiple files. However, this is handled in the context menu helper so it's okay to remove it from the action.

hishitetsu (approved):
LGTM

PR #12534, opened 2023-06-03
Feature: Added support for changing backdrop material

heftymouse:
**Resolved / Related Issues**
   Closes #issue...

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.


src/Files.App/Helpers/SystemBackdropHelper.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Files.App.Helpers
+{
+	internal sealed class AppSystemBackdrop : SystemBackdrop
+	{
+		private bool isSecondaryWindow;
+		private IUserSettingsService userSettingsService;
```
d2dyno1, line 17: Inject IUserSettingsService here

src/Files.App/Helpers/SystemBackdropHelper.cs
```diff
@@ -0,0 +1,92 @@
...
+		protected override void OnTargetConnected(ICompositionSupportsSystemBackdrop connectedTarget, XamlRoot xamlRoot)
+		{
+			base.OnTargetConnected(connectedTarget, xamlRoot);
+			this.target = connectedTarget;
+			this.root = xamlRoot;
+			userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
+			controller = GetSystemBackdropController(userSettingsService.AppearanceSettingsService.AppThemeSystemBackdrop);
+			controller?.SetSystemBackdropConfiguration(GetDefaultSystemBackdropConfiguration(connectedTarget, xamlRoot));
+			controller?.AddSystemBackdropTarget(connectedTarget);
+			userSettingsService.OnSettingChangedEvent += OnSettingChanged;
```
d2dyno1, line 37: Attach in constructor

src/Files.App/Helpers/SystemBackdropHelper.cs
```diff
@@ -0,0 +1,92 @@
...
+			this.target = null!;
+			this.root = null!;
+			controller?.RemoveSystemBackdropTarget(disconnectedTarget);
+			controller?.Dispose();
+			userSettingsService.OnSettingChangedEvent -= OnSettingChanged;
+		}
+
+		private void OnSettingChanged(object? sender, Shared.EventArguments.SettingChangedEventArgs e)
+		{
+			switch (e.SettingName)
```
d2dyno1, line 52: First check if the target was successfully connected

d2dyno1 (changes requested):

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3341,4 +3341,16 @@
...
   </data>
+  <data name="Acrylic" xml:space="preserve">
+    <value>Acrylic</value>
+  </data>
+  <data name="Mica" xml:space="preserve">
+    <value>Mica</value>
+  </data>
+  <data name="MicaAlt" xml:space="preserve">
+    <value>Mica Alt</value>
+  </data>
```
yaira2, line 3352: Are these supposed to be localized?

heftymouse: It's localized in Microsoft Docs so I added it

src/Files.App/ServicesImplementation/Settings/AppearanceSettingsService.cs
```diff
@@ -70,6 +71,13 @@ public String AppThemeFontFamily
 			set => Set(value);
 		}
 
+		/// <inheritdoc/>
+		public SystemBackdropType AppThemeSystemBackdrop
+		{
+			get => Get(MicaController.IsSupported() ? SystemBackdropType.MicaAlt : SystemBackdropType.Solid);
```
yaira2, line 77: Wouldn't it be better to keep MicaAlt so that it uses the fallback colors?

src/Files.App/Helpers/SystemBackdropHelper.cs
heftymouse: Is there a good reason to? This isn't a service and it'll just be moving the same code elsewhere

src/Files.App/ServicesImplementation/Settings/AppearanceSettingsService.cs
heftymouse: That's something I missed, thanks for pointing it out

src/Files.App/Helpers/SystemBackdropHelper.cs
d2dyno1: Well, there are some problems with the current approach:
1. You haven't marked `userSettingsService` as nullable (anything that might access it will result in NRE)
2. The service will be injected every time the method `OnTargetConnected()` is called

heftymouse: 1. Good point, I'll fix that
2. AppSystemBackdrop doesn't support being used for more than one target so it'll only ever only be called once (I need to add a check for that, thanks for reminding me)

src/Files.Backend/Services/Settings/IAppearanceSettingsService.cs
```diff
@@ -51,5 +52,10 @@ public interface IAppearanceSettingsService : IBaseSettingsService, INotifyPrope
 		/// Gets or sets a value for the app theme font family.
 		/// </summary>
 		String AppThemeFontFamily { get; set; }
+
+		/// <summary>
+		/// Gets or sets a value for the app theme system backdrop.
```
yaira2, line 57:
```suggestion
		/// Gets or sets a value for the theme system backdrop.
```

src/Files.App/Views/Settings/AppearancePage.xaml
```diff
@@ -115,6 +115,18 @@
 					SelectedIndex="{x:Bind ViewModel.SelectedThemeIndex, Mode=TwoWay}" />
 			</local:SettingsBlockControl>
 
+			<!--  System Backdrop  -->
+			<local:SettingsBlockControl Title="{helpers:ResourceString Name=SettingsAppearanceSystemBackdrop}" HorizontalAlignment="Stretch">
```
yaira2, line 119: We don't need to specify "SettingsAppearance", especially since resources can be reused so we try to keep it generic.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3344,4 +3344,19 @@
...
+  <data name="SettingsAppearanceSystemBackdrop" xml:space="preserve">
+    <value>System Backdrop</value>
```
yaira2, line 3357: Maybe we can call this "Backdrop material" to be more user friendly, what do you think?

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3344,4 +3344,19 @@
...
+  <data name="Acrylic" xml:space="preserve">
+    <value>Acrylic</value>
+  </data>
+  <data name="Mica" xml:space="preserve">
+    <value>Mica</value>
+  </data>
+  <data name="MicaAlt" xml:space="preserve">
+    <value>Mica Alt</value>
+  </data>
+  <data name="SettingsAppearanceSystemBackdrop" xml:space="preserve">
```
yaira2, line 3356:
```suggestion
  <data name="BackdropMaterial" xml:space="preserve">
```

src/Files.App/Views/Settings/AppearancePage.xaml
yaira2, line 119:
```suggestion
			<!--  Backdrop Material  -->
			<local:SettingsBlockControl Title="{helpers:ResourceString Name=BackdropMaterial}" HorizontalAlignment="Stretch">
```

src/Files.App/Views/Settings/AppearancePage.xaml
```diff
@@ -115,6 +115,18 @@
...
+				<local:SettingsBlockControl.Icon>
+					<FontIcon Glyph="&#xEF1F;" />
+				</local:SettingsBlockControl.Icon>
+				<ComboBox
+					x:Name="SystemBackdropChooser"
+					AutomationProperties.Name="{helpers:ResourceString Name=SettingsAppearanceSystemBackdrop}"
```
yaira2, line 125:
```suggestion
					AutomationProperties.Name="{helpers:ResourceString Name=BackdropMaterial}"
```

src/Files.App/ViewModels/Settings/AppearanceViewModel.cs
```diff
@@ -129,5 +145,19 @@ public string AppThemeBackgroundColor
 				}
 			}
 		}
+
+		private string selectedSystemBackdrop;
+		public string SelectedSystemBackdrop
```
yaira2, line 150:
```suggestion
		public string SelectedBackdropMaterial
```
Can you do this for the other variable names as well?

src/Files.App/Helpers/SystemBackdropHelper.cs
yaira2: @d2dyno1

yaira2 (dismissed):

yaira2 (dismissed):

yaira2:
The app is crashing on the automated tests, it might be because mica isn't available.

src/Files.App/Helpers/AppSystemBackdrop.cs
```diff
@@ -0,0 +1,98 @@
...
+					newController?.AddSystemBackdropTarget(target);
+					controller = newController;
+					break;
+			}
+		}
+
+		private ISystemBackdropControllerWithTargets? GetSystemBackdropController(BackdropMaterialType backdropType)
+		{
+			if (isSecondaryWindow && backdropType == BackdropMaterialType.MicaAlt)
+				backdropType = BackdropMaterialType.Mica;
```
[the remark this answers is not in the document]
heftymouse: It's required so that the properties window uses mica instead of mica alt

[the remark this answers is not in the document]
yaira2: I tried changing the value in settings to `1` and it worked but it crashes when I open the appearance settings page. Is this related?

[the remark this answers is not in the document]
yaira2: I found the issue with the pipeline is that we're not displaying any options in the dropdown when the controllers aren't available. The workaround is to add the options and just rely on the fallback colors in this case.

yaira2:
Thank you

PR #12599, opened 2023-06-13
Fix: Fixed crash in Dual Pane mode when switching away from Git folder

### LINKED ISSUE #12589: Bug: Consistent crash in Dual Pane mode when getting out of Git folder
### Description

In dual pane mode, opening a folder containing a Git repository in one of the pane and trying to switch focus to the other pane seems to provoke a crash.
 
Crash

### Steps To Reproduce

1. Open dual pane mode.
2. In first pane, open any folder with element. Select one element.
3. In the second pane, open a folder containing a Git repository. Select one element.
4. Switch to the first pane, the application should crash.

### Requirements

- No crash when switching out of the Git pane.

### Files Version

V.2.81.0 Dev

### Windows Version

Windows 10 (21H2): 19044

### Log File

[debug.log](https://github.com/files-community/Files/files/11724642/debug.log)

gave92:
**Resolved / Related Issues**
   Closes #12589

There seem to be an issue with x:Bind so that you can't use a TwoWay x:Bind on an element that is inside an element with x:Load:
```cs
<StackPanel x:Load="{x:Bind ...}">
    <RadioButton IsChecked="{x:Bind ..., Mode=TwoWay}" />
</StackPanel>
```
Workaround is to use Binding.

**Validation**
   1. Open a non-git folder in left pane and set layout to Tiles
   2. Open a git folder in right pane and set layout to Details
   3. Switch focus between the two panes
   4. Verify that you can still switch between local and remote branches in the Git flyout

### DIFF
--- a/src/Files.App/UserControls/StatusBarControl.xaml
+++ b/src/Files.App/UserControls/StatusBarControl.xaml
@@ -232,10 +232,11 @@
 								Grid.Row="1"
 								Padding="4"
 								Background="{ThemeResource CardBackgroundFillColorDefaultBrush}"
+								DataContext="{x:Bind DirectoryPropertiesViewModel, Mode=OneWay}"
 								IsItemClickEnabled="True"
 								ItemClick="BranchesList_ItemClick"
 								ItemsSource="{x:Bind DirectoryPropertiesViewModel.BranchesNames, Mode=OneWay}"
-								SelectedIndex="{x:Bind DirectoryPropertiesViewModel.SelectedBranchIndex, Mode=TwoWay}"
+								SelectedIndex="{Binding SelectedBranchIndex, Mode=TwoWay}"
 								SelectionMode="Single" />
 						</Grid>
 					</Flyout>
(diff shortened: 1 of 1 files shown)


gave92:
Uops I've just tested again and it crashes with a different error. Re-checking.

QuaintMako:
Weird, I did not have any exception or crash while trying in the same set-up.

gave92:
Ok the second error I think it was caused by running "git status" in cmd while having that folder open in dual pane.

yaira2 (approved):

ferrariofilippo:
There are two problems related with this PR:
- You can't checkout branches (it will keep on checking out because `SelectedIndex` won't be reset to 0)
- Navigating to a git repository, going back and then reopening the repo will cause Files to create +80 branches (The `Locals` radio button will be unchecked)

ferrariofilippo:
> Navigating to a git repository, going back and then reopening the repo will cause Files to create +80 branches (The Locals radio button will be unchecked)

We could solve this by changing the binding to `OneWay` and manually updating `ShowLocals` when the user clicks one of the two radios.
What do you think?

yaira2:
Sounds good

QuaintMako:
It could work I believe

PR #12674, opened 2023-06-21
Feature: Added option to hide the Send To menu

### LINKED ISSUE #12546: Feature: Add option to hide "send to" menu

I would like being able to choose which entries show up in each context menu ; although, as this may not be compatible with the current way context menus have been implemented, or as it may require quite the effort, I would like to ask for an option to hide the "Send to" entry, as it looks out of place, and as I personally find no use in its presence (although this last bit is completely subjective).

### Requirements

- This suggestion will allow for further customisation
- This proposition would bring cohesion to the UI
- This feature could de-clutter each context menu, for those who find no use in the option.

### Files Version

2.4.71.0

### Windows Version

Windows 10 (22H2): 19045

### Comments

The fact that the 'Send to' entry does not feature an icon is what makes it stand out negatively. If none of the aforementioned options were to be implemented or implementable, I definitely think giving it an icon would add to the continuity and cohesion of the Files app's user interface.

Alternatively, one could follow the registry entries that determine whether 'Send to' should be enabled or not. This way, for those which have disabled it already (using third party software like WinaeroTweaker for instance), there would be no redundant setting to toggle.

yaira2:
**Resolved / Related Issues**
   Closes #12546...

**Validation**
   1. Confirm the Send To menu only shows when the setting is enabled
   2. Test in the widgets as well

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Helpers/ContextFlyoutItemHelper.cs
+++ b/src/Files.App/Helpers/ContextFlyoutItemHelper.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See the LICENSE.
 
 using Files.App.Commands;
+using Files.App.ServicesImplementation.Settings;
 using Files.Backend.Helpers;
 using Files.Backend.Services;
 using Microsoft.UI.Xaml.Media.Imaging;
@@ -545,7 +546,7 @@ namespace Files.App.Helpers
 					Tag = "SendTo",
 					CollapseLabel = true,
 					ShowInSearchPage = true,
-					ShowItem = itemsSelected
+					ShowItem = itemsSelected && userSettingsService.GeneralSettingsService.ShowEditTagsMenu
 				},
 				new ContextMenuFlyoutItemViewModel()
 				{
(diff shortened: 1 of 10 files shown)


hishitetsu:
Support for widgets isn't working well.
I think it would work to check the settings here instead.
https://github.com/files-community/Files/blob/d6cbb280425f26b49a5cd4fd04f7431b81387f84/src/Files.App/Helpers/ShellContextMenuHelper.cs#L330-L331

hishitetsu:
Both widgets. All widgets and the sidebar use this method to generate a context menu.

yaira2:
@hishitetsu can you check again?

hishitetsu:
> @hishitetsu can you check again?

It still doesn't work well. I'll look into it.

hishitetsu:
@yaira2 This PR should fix the widgets problem.
https://github.com/files-community/Files/pull/12684

hishitetsu (approved):
LGTM

PR #12683, opened 2023-06-21
Fix: Fixed issue where middle clicking selected folders would cause crash

### LINKED ISSUE #12613: Bug: Crash when opening highlighted folder in new tab
### Description

Majority of the time when opening a highlighted/selected folder via middle-mouse button or triple finger tap on a trackpad, Files crashes 

### Steps To Reproduce

1. Select/Highlight a folder
2. Press Middle mouse to open in new tab
3. 50% chance of it (Files) crashing 

https://github.com/files-community/Files/assets/51767230/f1464975-2971-47eb-8a41-02a35a8174d8
In the video I did the same exact thing twice, but only one of those times did it crash.

### Requirements

Fix the crash that happens whenever a highlighted folder is opened in a new tab via middle mouse (This does not seem to happen via the context menu)

### Files Version

Version 2.5.1.0 

### Windows Version

Windows 11 (22H2): 22621

### Log File

[debug.log](https://github.com/files-community/Files/files/11750849/debug.log)

yaira2:
**Resolved / Related Issues**
   Closes #12613...

**Validation**
   1. Select folder
   2. Middle click selected folder to open it in new tab
   3. Confirm it doesn't cause a crash
   4. Test in all the layouts

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
+++ b/src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
@@ -188,6 +188,13 @@ namespace Files.App.Views.LayoutModes
 			OldItemName = textBlock.Text;
 			textBlock.Visibility = Visibility.Collapsed;
 			textBox.Visibility = Visibility.Visible;
+
+			if (textBox.FindParent<Grid>() is null)
+			{
+				textBlock.Visibility = Visibility.Collapsed;
+				return;
+			}
+
 			Grid.SetColumnSpan(textBox.FindParent<Grid>(), 8);
 
 			textBox.Focus(FocusState.Pointer);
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
```diff
@@ -188,6 +188,13 @@ protected virtual void StartRenameItem(string itemNameTextBox)
 			OldItemName = textBlock.Text;
 			textBlock.Visibility = Visibility.Collapsed;
 			textBox.Visibility = Visibility.Visible;
+
+			if (textBox.FindParent<Grid>() is null)
+			{
+				textBlock.Visibility = Visibility.Collapsed;
```
hishitetsu, line 194:
```suggestion
				textBlock.Visibility = Visibility.Visible;
				textBox.Visibility = Visibility.Collapsed;
```

hishitetsu (commented):
The tiles layout doesn't crash, but there is a glitch that leaves the text box when you go back to the original tab, so we should do the same fix as below.

src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
yaira2: I can't believe I did this backwards!

PR #12716, opened 2023-06-24
Code Quality: Introduce App.Utils namespace

0x5bfa:
### Motivation & Context

- Improved readability

> **Warning**<br/>Please do not review code lines that I did not changed. I cannot make more diff. If there're more codes that are preferred to change, I will do that and mark that PR as the final one for that stage.

### PR Checklist

   Related #4180

### Screenshots

_None_


QuaintMako (commented):
Code looks OK, cannot test it or see why AZP did not work though.

src/Files.App/Utils/Archives/ArchiveCreator.cs
```diff
@@ -123,6 +124,9 @@ public ArchiveCreator()
 		/// <inheritdoc/>
 		public string GetArchivePath(string suffix = "")
 		{
+			if (string.IsNullOrEmpty(Directory))
+				throw new NullReferenceException($"{Directory} is null.");
```
yaira2, line 128: I don't think we should throw in release environments, returning early is better.

PR #12726, opened 2023-06-25
Fix: Fixed crash that would sometimes occur when navigating away from archives

hishitetsu:
image

**Resolved / Related Issues**
   Closes #issue...

**Validation**

### DIFF
--- a/src/Files.App/UserControls/InnerNavigationToolbar.xaml
+++ b/src/Files.App/UserControls/InnerNavigationToolbar.xaml
@@ -266,23 +266,23 @@
 						<MenuFlyout helpers:MenuFlyoutHelper.IsVisible="{x:Bind ViewModel.CanExtract, Mode=OneWay}" Placement="Bottom">
 							<MenuFlyoutItem
 								x:Name="ExtractSingle"
-								x:Load="{x:Bind ViewModel.IsMultipleArchivesSelected, Mode=OneWay, Converter={StaticResource BoolNegationConverter}}"
 								Command="{x:Bind Commands.DecompressArchive, Mode=OneWay}"
 								IsEnabled="{x:Bind ViewModel.CanExtract, Mode=OneWay, FallbackValue=False}"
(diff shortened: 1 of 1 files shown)


src/Files.App/UserControls/InnerNavigationToolbar.xaml
```diff
@@ -266,23 +266,23 @@
 						<MenuFlyout helpers:MenuFlyoutHelper.IsVisible="{x:Bind ViewModel.CanExtract, Mode=OneWay}" Placement="Bottom">
 							<MenuFlyoutItem
 								x:Name="ExtractSingle"
-								x:Load="{x:Bind ViewModel.IsMultipleArchivesSelected, Mode=OneWay, Converter={StaticResource BoolNegationConverter}}"
 								Command="{x:Bind Commands.DecompressArchive, Mode=OneWay}"
 								IsEnabled="{x:Bind ViewModel.CanExtract, Mode=OneWay, FallbackValue=False}"
 								KeyboardAcceleratorTextOverride="{x:Bind Commands.DecompressArchive.HotKeyText, Mode=OneWay}"
-								Text="{x:Bind Commands.DecompressArchive.Label}" />
+								Text="{x:Bind Commands.DecompressArchive.Label}"
+								Visibility="{x:Bind ViewModel.IsMultipleArchivesSelected, Mode=OneWay, Converter={StaticResource NegatedBoolToVisibilityConverter}}" />
```
yaira2, line 273: How does this fix the crash?

hishitetsu: The crash occurs when unloading this MenuFlyoutItem. I have used Visibility because I believe the crash is due to the timing overlap with the unloading of the parent AppBarButton.
Since there are only three items in the flyout and they are only loaded when manipulating the archive, I think there is little advantage to using x:Load to each MenuFlyoutItem.

yaira2 (approved):
LGTM

PR #12728, opened 2023-06-25
Feature: Support changing behavior of buttons with modifier keys

### LINKED ISSUE #12117: Feature: Change the behavior of toolbar and flyout buttons with modifier keys

The ability to change the behavior of toolbar and flyout buttons with modifier keys would allow for a greater variety of operations.
For example, In FIle Explorer the delete button acts as permanent deletion while pressing `Shift`, but it doesn't in Files.

### Requirements

Create a mechanism to change the behavior of toolbar and flyout buttons with modifier keys.

### Files Version

2.4.65.0

### Windows Version

Windows 11 22H2

### Comments

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #12117 

**Validation**
   1. Holding down Shift and pressing the delete button on the toolbar or flyout will result in permanent deletion.
   2. Holding down Shift and pressing the paste button on the toolbar will paste items into selected folder. If no items are selected, it will act as normal pasting.


src/Files.App/Commands/Manager/CommandManager.cs
```diff
@@ -168,6 +174,14 @@ public CommandManager()
...
 				.ToImmutableDictionary(command => command.Code);
 
+			_ModifiableDeleteItem = new ModifiableCommand(commands[CommandCodes.DeleteItem], new() {
+				{ KeyModifiers.Shift,  commands[CommandCodes.DeleteItemPermanently] }
+			});
+
+			_ModifiablePasteItem = new ModifiableCommand(commands[CommandCodes.PasteItem], new() {
+				{ KeyModifiers.Shift,  commands[CommandCodes.PasteItemToSelection] }
+			});
+
```
QuaintMako, line 184: I am not a fan at all of instanciating those methods here. They should at least be done inside of a dedicated method, possibly called by the constructor. I'd go as far as create a smaller class contained by the Command Manager to store the modifiable commands? If we get more modifiables here, it's gonna be really crowded soon.

src/Files.App/Commands/Manager/CommandManager.cs
```diff
@@ -563,5 +577,112 @@ private void Action_PropertyChanged(object? sender, PropertyChangedEventArgs e)
 				}
 			}
 		}
+
+		[DebuggerDisplay("Command {Code} (Modifiable)")]
+		private class ModifiableCommand : ObservableObject, IRichCommand
```
QuaintMako, line 582: To move to a new file. The concept is too important to be shove under another class.

src/Files.App/Commands/Manager/CommandManager.cs
```diff
@@ -563,5 +577,112 @@ private void Action_PropertyChanged(object? sender, PropertyChangedEventArgs e)
...
+			}
+
+			public bool IsExecutable => BaseCommand.IsExecutable;
+
+			public ModifiableCommand(IRichCommand baseCommand, Dictionary<KeyModifiers, IRichCommand> modifiedCommands)
+			{
+				BaseCommand = baseCommand;
+				ModifiedCommands = modifiedCommands.ToImmutableDictionary();
+
+				if (baseCommand is ActionCommand actionCommand)
```
QuaintMako, line 626: Invert if to reduce nesting.

QuaintMako (commented):
Some feedback about the code. I cannot test it though. Some interesting feature that might be very powerful on the long run.

To be honest fully, there is a little something feeling wrong to me but I cannot pinpoint what... I'll be sure to come back to elaborate when I have IDE to review it properly.

src/Files.App/Commands/Manager/CommandManager.cs
hishitetsu: The instantiating of commands has been carved out into a separate method. Also, the management of ModifiableCommand itself has been carved out into a new class.

src/Files.App/Commands/Manager/CommandManager.cs
hishitetsu: Since this code also serves as a cast, I left it as is.

QuaintMako (approved):
I cannot test the code, yet with the suggested refactoring done, I trust the concept way more!

LGTM, I'll need someone else to test the code.

PR #12750, opened 2023-06-26
Fix: Fixed issue where the System version of VS Code wasn't detected

ferrariofilippo:
**Resolved / Related Issues**

**Validation**
   1. Install VS Code (System)
   2. Navigate to a git repository


src/Files.App/Helpers/SoftwareHelpers.cs
```diff
@@ -11,16 +11,34 @@ public static bool IsVSCodeInstalled()
...
+
+			return true;
+		}
+
+		private static bool ContainsName(RegistryKey? key, string find)
+		{
+			if (key is null)
+				return false;
 
+			string? displayName;
```
mafra99, line 37: declaration of `displayName` could be moved into the foreach loop

mafra99 (commented):
I have tested the changes, works as expected, Open in VS code is now displayed

src/Files.App/Helpers/SoftwareHelpers.cs
ferrariofilippo: Won't that impact performance? It would declare a new variable at each cycle. I know it wouldn't make that much difference

mafra99: I am pretty sure that it doesn't

if you look at the lowered code you will see that there is no difference, example  [here](https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoBAGABAlAFgG5NMEBmfJXAYVwG9dnNnnLdgB7LgG1wAqAUwDOAFxQAKADJRxAHgLYAfLgDWQgJ4iANLgDKYgE6wA5rgBmsACYBKVmwaYAkK4tcjQiAGMAFrkkANwgjXBEAV2AAaS1cWHUtEXsMV2cnFNSlAH5cazkAB14ITQA5CDAhXABeMMiYzQA6aSEYUzFfBoEuQxNWyVtSDNcoCwCAQiUGgEkREvDeXgB5IwB1XygxIX18nyFJPJFC4rKK21wAMnPcgqLS8qEGwxCxERWN30krGDtk1LSXP4IADsuGM4SEgz+AF8Ac4YRhYcDLBBeCIIQ5mPC2LgMfgqNw+IJRGIkDI5GJFCgVAltHoemZLDZktj0pkqTkDkc7hVIW4PF4/AFgqEItFYvENNpfq5WalObcTpUaqL6k0Wm0Ol16X0BrDnCNxpMZnMFss1hstjtvHt5cd7mdLtdDgr7o8xM9Xu9PkzpTK9c4kWD0UM4bD4YiQRYUWjBtj4VCgA=)

src/Files.App/Helpers/SoftwareHelpers.cs
```diff
@@ -11,16 +11,33 @@ public static bool IsVSCodeInstalled()
 		{
 			string registryKey = @"Software\Microsoft\Windows\CurrentVersion\Uninstall";
 
-			var key = Registry.CurrentUser.OpenSubKey(registryKey);
+			return
+				ContainsName(Registry.CurrentUser.OpenSubKey(registryKey), "Microsoft Visual Studio Code") ||
+				ContainsName(Registry.LocalMachine.OpenSubKey(registryKey), "Microsoft Visual Studio Code");
```
QuaintMako, line 16: Put `Microsoft Visual Studio Code` in a variable since it's repeated.

QuaintMako (approved):
LGTM codewise, I cannot test it for now though.

PR #12810, opened 2023-06-30
Feature: Display Git details in the Details Pane

### LINKED ISSUE #12779: Feature: Include Git Repo name and current branch in Folder's Details pane

Can we display basic git info on the Details page?  Repo name, and current branch?
Of course any more detailed Git info would need to be in its own tab/page.

I think if the Status Bar Buttons are being collapsed or truncated, the details pane may be a useful place to have the branch name visible
the full git tab, will be quite busy, and it seems like a top level enough "detail" to display.

This should probably also be displayed in the Properties Dialog also.

image

image

### Requirements

- This proposal will ensure git repo folders are recognisable on the main details page.

### Files Version

2.5.10

### Windows Version

22621.1928

### Comments

_No response_

QuaintMako:
**Resolved / Related Issues**
   Closes #12779 

**Validation**
   1. Tried on a unsynchronized repository with no origin and no commit. No data is displayed.

**Screenshots**

Screenshot_1

### DIFF
--- a/src/Files.App/Resources/PreviewPanePropertiesInformation.json
+++ b/src/Files.App/Resources/PreviewPanePropertiesInformation.json
@@ -468,15 +468,14 @@
 		"ID": null
 	},
 	{
-		"NameResource": "FileTags",
+		"NameResource": "GitOriginRepositoryName",
 		"SectionResource": "PropertySectionCore",
 		"IsReadOnly": true,
-		"ID": "filetag"
+		"ID": null
 	},
 	{
-		"NameResource": "GitBranch",
+		"NameResource": "GitCurrentBranch",
 		"SectionResource": "PropertySectionCore",
-		"Property": null,
 		"IsReadOnly": true,
 		"ID": null
 	}
--- a/src/Files.App/Resources/PreviewPanePropertiesInformation.json
+++ b/src/Files.App/Resources/PreviewPanePropertiesInformation.json
@@ -478,5 +478,11 @@
 		"SectionResource": "PropertySectionCore",
 		"IsReadOnly": true,
 		"ID": null
+	},
+	{
+		"NameResource": "FileTags",
+		"SectionResource": "PropertySectionCore",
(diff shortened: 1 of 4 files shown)


yaira2:
Isn't the repo name the same as the folder?

src/Files.App/Resources/PreviewPanePropertiesInformation.json
```diff
@@ -467,6 +467,18 @@
...
+		"SectionResource": "PropertySectionCore",
+		"IsReadOnly": true,
+		"ID": null
+	},
+	{
+		"NameResource": "GitCurrentBranch",
+		"SectionResource": "PropertySectionCore",
+		"IsReadOnly": true,
+		"ID": null
+	},
```
yaira2, line 481: Because the tags field is often blank, it might look better to put these after tags. Alternatively we can hide the tags field when there aren't any tags.

mdtauk: Hiding tags when there are none is a good option.

Alternatively you could have an Add Tag button here, possibly with text saying "This file/folder currently has no tags attached"

mdtauk: As "tags" is pluralised, maybe this icon is more appropriate...
image

QuaintMako: The "Add Tag" is a pretty good idea @mdtauk, I'd suggest opening a feature request for that.

mdtauk: done #12821

QuaintMako: > Because the tags field is often blank, it might look better to put these after tags. Alternatively we can hide the tags field when there aren't any tags.

Tags and git info switched.

yaira2 (approved):
LGTM

PR #12839, opened 2023-07-03
Fix: Fixed issue were removing a tag after changing the layout would cause a crash

### LINKED ISSUE #12814: Bug: Deleting tags after changing layout may cause crash
### Description

As the title described. It is always reproduced by the following steps.

### Steps To Reproduce

1. Set "Home" tag and "Work" tag to a folder in the details layout.
2. Change the layout to the grid layout.
3. Change the layout back to the details layout.
4. Delete "Home" tag.
5. Delete "Work" tag.
6. The app will crash.

### Requirements

Fix crash.

### Files Version

2.5.11.0

### Windows Version

10.0.22621.1848

### Log File

[debug.log](https://github.com/files-community/Files/files/11923916/debug.log)

ferrariofilippo:
**Resolved / Related Issues**
   Closes #12814

**Validation**
   1. Set "Home" tag and "Work" tag to a folder in the details layout.
   2. Change the layout to the grid layout.
   3. Change the layout back to the details layout.
   4. Delete "Home" tag.
   5. Delete "Work" tag.
   6. The app will crash.

**Screenshots (optional)**
A part of the issue is caused by `Select` throwing exception with empty arrays, the other is related to this piece of code
Screenshot 2023-07-03 221547
Since `FileTagsUI` is empty `[0]` doesn't exist. I fixed setting it to null, but we could also create a `FirstTag` property

### DIFF
--- a/src/Files.App/Filesystem/ListedItem.cs
+++ b/src/Files.App/Filesystem/ListedItem.cs
@@ -133,7 +133,7 @@ namespace Files.App.Filesystem
 			}
 		}
 
-		public IList<TagViewModel> FileTagsUI
+		public IList<TagViewModel>? FileTagsUI
 		{
 			get => fileTagsSettingsService.GetTagsByIds(FileTags);
 		}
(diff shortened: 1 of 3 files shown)


yaira2 (approved):
LGTM

hishitetsu:
Certainly this avoids crashes, but the underlying problem is that the binding to the grid layout occurs despite the change to the details layout. This may cause other problems besides this crash.

yaira2:
We'll have to keep an eye out for this

PR #12853, opened 2023-07-05
Fix: Fixed issue where "Open in new window" followed the startup settings

### LINKED ISSUE #12812: Bug: "Open in new window" follows the startup settings
### Description

When I open a folder with "Open in new window", I expect only that folder to be opened in the new window. However, if the startup settings is "Continue where you left off", tabs in the original window are opened together.

### Steps To Reproduce

1. Set the startup settings other than "Open a new tab".
2. Open a folder with "Open in new window".
3. Tabs other than the folder you want to open are also opened according to the startup settings.

### Requirements

Only the selected folder should be opened in a new window.
However, if the user starts the app by specifying a folder from the command line, it should follow the startup settings as before.

### Files Version

2.5.11.0

### Windows Version

10.0.22621.1848

### Log File

n/a

hishitetsu:
**Resolved / Related Issues**
   Closes #12812 

**Validation**
   1. If you perform "Open in new window", the new window will contain a tab of selected item only.
   2. If you launch the app from command line with a path to open, it will still follow the startup settings.

### DIFF
--- a/src/Files.App/Data/EventArguments/MainPageNavigationArguments.cs
+++ b/src/Files.App/Data/EventArguments/MainPageNavigationArguments.cs
@@ -1,0 +1,12 @@
+﻿// Copyright (c) 2023 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.App.Data.EventArguments
+{
+	internal class MainPageNavigationArguments
+	{
+		public object? Parameter { get; set; }
+
+		public bool IgnoreStartupSettings { get; set; }
+	}
+}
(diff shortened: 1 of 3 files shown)


yaira2:
An alternative approach could be to check for an existing instance, besides fixing this issue, it'll help if the user opens a second instance from the Start Menu or Taskbar.

hishitetsu:
> An alternative approach could be to check for an existing instance, besides fixing this issue, it'll help if the user opens a second instance from the Start Menu or Taskbar.

Even now, when I open the second instance when the startup setting is "Continue where you left off", only the home page is displayed in the new window. Also, if I try to launch the app by specifying a folder while the app is already running, a new tab will be added to the existing window.
So I think it is fine as it is.

yaira2 (approved):
LGTM

PR #12885, opened 2023-07-07
Feature: Added path column to Details layout for search results

### LINKED ISSUE #7070: Feature: Add path column for search results/tags
### Description

When searching I would expect some way to identify the files found. There is no indication whatsoever (not even in the tooltip) where the item is from (folder), the only option is using "Open Parent" (or "Copy location") from the context menu of a file, which will throw away the search results because it opens in the same pane. Returning to the search results will repeat the search though.
image

### Expected behavior

1. Add a column with the file path in search results

### Files Version

Version: 2.0.34.0 

### Windows Version

OS Version: 10.0.19042.1348

### Relevant Assets

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #7070 

**Validation**

**Screenshots**
image

Grouping:
image

### DIFF
--- a/src/Files.App/Strings/en-US/Resources.resw
+++ b/src/Files.App/Strings/en-US/Resources.resw
@@ -133,7 +133,7 @@
     <value>Created:</value>
   </data>
   <data name="Path" xml:space="preserve">
-    <value>Path:</value>
+    <value>Path</value>
   </data>
   <data name="SizeLabel" xml:space="preserve">
     <value>Size:</value>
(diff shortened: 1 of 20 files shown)


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -133,7 +133,7 @@
     <value>Created:</value>
   </data>
   <data name="Path" xml:space="preserve">
```
yaira2, line 135: Is this being used in the properties window?

hishitetsu: image
"Location:" is used instead of "Path:". I can find no other place it is used.

yaira2 (approved):
LGTM

PR #12921, opened 2023-07-10
Code Quality: Refactor StatusCenter-related codes

0x5bfa:
### Motivation & Context

- Improved readability

> **Warning**<br/>Please do not review code lines that I did not changed. I cannot make more diff. If there're more codes that are preferred to change, I will do that and mark that PR as the final one for that stage.

### PR Checklist

   Related #4180

### Screenshots

image


src/Files.App/ViewModels/UserControls/StatusCenterViewModel.cs
```diff
@@ -1,17 +1,17 @@
...
 namespace Files.App.ViewModels.UserControls
 {
-	public class OngoingTasksViewModel : ObservableObject, IOngoingTasksActions
+	public class StatusCenterViewModel : ObservableObject
 	{
-		public ObservableCollection<StatusBanner> StatusBannersSource { get; private set; } = new ObservableCollection<StatusBanner>();
-
-		private int medianOperationProgressValue = 0;
+		private int _MedianOperationProgressValue = 0;
 		public int MedianOperationProgressValue
```
ferrariofilippo, line 11: Can you rename it to `AverageOperationProgressValue`? 
In `UpdateMedianProgress()` (which should be renamed as well) we are calculating the average value not the median

src/Files.App/Utils/StatusCenter/StatusCenterItem.cs
```diff
@@ -0,0 +1,163 @@
...
+	public class StatusCenterItem : ObservableObject
+	{
+		private readonly float _initialProgress = 0.0f;
+
+		private string? _FullTitle;
+		public string? FullTitle
+		{
+			get => _FullTitle;
+			set => SetProperty(ref _FullTitle, value ?? string.Empty);
+		}
```
ferrariofilippo, line 21: I don't get why you are assigning `string.Empty` when `value is null` even if you marked the property as nullable. I think you can initialize it to `string.Empty` and make it not nullable

ferrariofilippo (commented):
It works but some things might be refined

src/Files.App/Utils/StatusCenter/StatusCenterItem.cs
0x5bfa: Actually this is what this was. I will change it.

PR #12946, opened 2023-07-13
Feature: Add edit tags button to the preview pane

### LINKED ISSUE #12821: Feature: Tag section on Detials Pane, should include "Add Tags" button

As mentioned in #12810 when a file or folder has no tags, a tags section still shows in the detail pane.

image

Rather than hiding this when no tags have been attached to the item, we could put an "Add Tag" or "Add Tags" button.

image
image

### Requirements

- This will simplify adding tags to items, and emphasise the feature to people who are not familiar with it.

### Files Version

2.5.10

### Windows Version

22621.1928

### Comments

_No response_

### LINKED ISSUE #12947: Bug: Tags are not displayed in the Preview Pane if no item is selected
### Description

If no item is selected, the Preview Pane will show details of the current folder. The Preview Pane won't show its tags (even if it has some)

### Steps To Reproduce

1. Select a tagged folder
2. Open the preview pane
3. See that it displays selected folder's tags
4. Open that folder
5. See that the tags section won't be rendered (no tags are detected)

### Requirements

Tags (if any) should always be displayed in the preview pane

### Files Version

Files Preview 2.5.14.0

### Windows Version

Windows 11 Pro 22H2 22621.1992

### Log File

Selected item:
Screenshot 2023-07-14 135904

No item selected:
Screenshot 2023-07-14 135854

ferrariofilippo:
**Resolved / Related Issues**
   Closes #12821 
   Closes #12947 

**Validation**
   1. Open app and select a folder
   2. Open the preview pane
   3. Play with tags

**Screenshots (optional)**
Screenshot 2023-07-13 234208


yaira2:
We may want to consider hiding this on items without tags, this will be consistent with other properties that are hidden when empty.

mdtauk:
I am assuming this is using the same flyout as is used elsewhere in the app - so there is no code duplication...

ferrariofilippo:
> I am assuming this is using the same flyout as is used elsewhere in the app - so there is no code duplication...

Yes

ferrariofilippo:
> We may want to consider hiding this on items without tags, this will be consistent with other properties that are hidden when empty.

I made it always visible since that was the required behaviour in the issue
> Rather than hiding this when no tags have been attached to the item, we could put an "Add Tag" or "Add Tags" button.

Anyway, I think we should keep it even if there are no tags, it's quicker to edit tags using that button than using the context menu.
Still we could hide it if the user has turned off `Show tags in context menu`

ferrariofilippo:
image

yaira2:
@ferrariofilippo Can you display the tags horizontally (and wrap onto the next line if there isn't enough room)?

mdtauk:
> @ferrariofilippo Can you display the tags horizontally (and wrap onto the next line if there isn't enough room)?

Two per row I think will be enough, if they take up 50% of the space each, This way the tag icons below will align to the ones above, and it wont look staggered.

yaira2:
> Still we could hide it if the user has turned off

It's okay, we can always show it.

ferrariofilippo:
@yaira2 do you know any control better than `ItemsWrapGrid`? Using that you have to set manually items width

yaira2:
@ferrariofilippo We can use a GridView with a custom template, the Community Toolkit has a couple of templates with the desired behavior.

ferrariofilippo:
> @ferrariofilippo We can use a GridView with a custom template, the Community Toolkit has a couple of templates with the desired behavior.

Did you mean this?
https://learn.microsoft.com/en-us/windows/apps/design/controls/item-templates-gridview#icon-and-text

yaira2:
@ferrariofilippo https://learn.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.ui.controls.wraplayout?view=win-comm-toolkit-dotnet-7.0
image

ferrariofilippo:
> @ferrariofilippo https://learn.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.ui.controls.wraplayout?view=win-comm-toolkit-dotnet-7.0

Using that you can't display two tags/line, is that fine?
image

yaira2:
> Using that you can't display two tags/line, is that fine?

What do you mean? The screenshot looks like the desired behavior...

ferrariofilippo:
> Two per row I think will be enough, if they take up 50% of the space each, This way the tag icons below will align to the ones above, and it wont look staggered.

I thought we were going for this

yaira2 (approved):
LGTM

src/Files.App/UserControls/Pane/PreviewPane.xaml
```diff
@@ -269,6 +271,63 @@
 					</ItemsControl.ItemTemplate>
 				</ItemsControl>
 
+				<!--  Tags List  -->
+				<StackPanel
+					x:Name="DetailsTagsList"
+					Spacing="4"
```
yaira2, line 277: Can x:Load be false when there aren't any tags? Users will still be able to add tags from the menu flyout.

yaira2:
image
I think we can improve the design by removing the "edit tag" button and instead, adding a "plus" icon to the GridView.

ferrariofilippo:
> I think we can improve the design by removing the "edit tag" button and instead, adding a "plus" icon to the GridView.

Where should it be? Before/after the tags? Next to the header?

yaira2:
Ideally it'll be in the list of tags, is this possible?

ferrariofilippo:
I'm not sure we can do that but I'll see

ferrariofilippo:
@yaira2 I think we can do that using something like this, is it fine?
```
public class TagsListItem
{
	public object Item { get; set; }

	public bool IsButton => object is Button;

	public bool IsTag => object is Tag;
}

<DataTemplate>
	<StackPanel>
		<TagLayout x:Load="{x:Bind obj.IsTag, Mode=OneWay}" other_props />
		<Button x:Load="{x:Bind obj.IsButton, Mode=OneWay}" />
	</StackPanel>
</DataTemplate>

```

yaira2:
Seems like a bit of a hack but it's simple enough so we might as well.

yaira2:
@ferrariofilippo even better, what if we added a "plus" icon next to the "Tags" header?

ferrariofilippo:
Does it look better using `OpacityIcon` or using the `Glyph`?

Screenshot 2023-07-19 102839
Screenshot 2023-07-19 103318

Lukiluc29:
hello, sorry to embed myself in your discussion but it would not be better to put the button after the tags instead of putting it next to "tags"
254524240-81867ae9-fe5e-4dd7-ac09-e74a728721b2 (1)

ferrariofilippo:
> Hello, sorry to embed myself in your discussion but it would not be better to put the button after the tags instead of putting it next to "tags" 254524240-81867ae9-fe5e-4dd7-ac09-e74a728721b2 (1)

This was the original idea but it's quite a mess to implement, we would need some hacky stuff

Lukiluc29:
It's a shame because the original idea is much more beautiful than the one you have just proposed, as, for me, it would look out of place next to the 'tags'.

ferrariofilippo:
@yaira2 what do you think? Both the solutions work properly. I can push the one you prefer

yaira2:
> Is it going to be an edit button however, or an add tag button.

For now it's going to be the edit tag button.

mdtauk:
image
What about something like this.  You can quick remove tags, or press the button to do more and add new ones.

Lukiluc29:
> image What about something like this. You can quick remove tags, or press the button to do more and add new ones.

It's a great concept that will be beautiful once implemented.

PR #12972, opened 2023-07-18
Feature: Open all items in a tag

### LINKED ISSUE #7493: Feature: Option to open all items in a tag
### What's the Problem?

People usually use the same group of apps for common tasks. 

Right now Bundles are just a fancy quick-access bookmarks, but allowing to open all contained items at once (when the bundle is clicked) would make them much more useful

### Solution/Idea

Open all contained items at once when the Bundle is clicked

### Alternatives

Click+Shift+Click to select all items then Enter to open them

### Priorities

| Capability | Priority |
| :---------- | :------- |
| Allow users to still open items individually when the item name is clicked | Must |
| Allow users to open all items in 1 click, probably from a menu flyout on the top right of the tag container on the home page | Must |

### Files Version

.34.0

### Windows Version

21H2

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**
   Closes #7493

**Validation**
   1. Open app and navigate to `Home`
   2. Enable the `Tags` widget
   3. Right click on the title bar
   4. Click on `Open all`

**Screenshots (optional)**
Screenshot 2023-07-18 224854


src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace Files.App.Actions
+{
+    sealed class OpenAllTaggedActions: ObservableObject, IAction
+    {
+		private readonly IContentPageContext _pageContext;
+
+		private readonly ITagsContext _tagsContext;
+
+		public string Label
+			=> "OpenAllItems".GetLocalizedResource();
```
yaira2, line 13:
```suggestion
			=> "OpenAllTaggedItems".GetLocalizedResource();
```

yaira2:
@ferrariofilippo I pushed a small design tweak so that clicking the tag name does the search for more items. This will allow us to repurpose the "view more" button to "open all".

ferrariofilippo:
What about using our OpacityIcons? The Glyph looks strange
image

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
QuaintMako, line 13: The label should be `OpenAllTaggedItems`

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -0,0 +1,56 @@
...
+
+		public string Description
+			=> "OpenAllItemsDescription".GetLocalizedResource();
```
QuaintMako, line 16: `OpenAllTaggedItemsDescription`

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3398,4 +3398,10 @@
   <data name="OpenInNewWindowDescription" xml:space="preserve">
     <value>Open directory in new window</value>
   </data>
+  <data name="OpenAllItems" xml:space="preserve">
+    <value>Open all</value>
+  </data>
+  <data name="OpenAllItemsDescription" xml:space="preserve">
+    <value>Open all tagged items</value>
+  </data>
```
QuaintMako, line 3406: `OpenAllTaggedItems` and `OpenAllTaggedItemsDescription`

src/Files.App/UserControls/Widgets/FileTagsWidget.xaml
```diff
@@ -51,39 +52,44 @@
...
+							</StackPanel>
+						</HyperlinkButton>
+
+						<!--  Open All Items  -->
+						<HyperlinkButton
+							Grid.Column="2"
+							Margin="4"
+							AutomationProperties.Name="{helpers:ResourceString Name=OpenAllItems}"
+							Command="{x:Bind OpenAllCommand}"
+							ToolTipService.ToolTip="{helpers:ResourceString Name=OpenAllItems}">
```
QuaintMako, line 90: If the resources strings are indeed renamed as proposed, these entries will need to be renamed as well.

QuaintMako (commented):
Code looks good, I am not able to test it.
A small proposition of renaming to stay coherent with the current.

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
yaira2: I thought this as well but then I realized this string might be useful outside of tags.

ferrariofilippo: Should I revert it then?

ferrariofilippo:
> @ferrariofilippo I don't think the opacity icon is a good fit but to be honest, there isn't an intuitive icon for "open all". An alternative would be to use the `more` icon and display the menu item to "open all" when clicking the icon. 

Good call, we won't need to change the code if we will need any extra actions

ferrariofilippo:
image
 Should we remove the button border?

yaira2:
Can you set it to transparent?

yaira2:
Does this also work for the sidebar?

ferrariofilippo:
Not yet, I'll work on it

yaira2 (approved):
LGTM

PR #12977, opened 2023-07-19
Feature: Added a Command Palette

### LINKED ISSUE #2780: Feature: Add command palette

hishitetsu:
**Resolved / Related Issues**
   Closes #2780 

**Validation**
  1. Type ">" in the address bar or press Ctrl+Shift+P to open the command palette
  2. You can search commands by both description (translated) and command code.
  3. Select a suggestion or type a command code to invoke the command.

**Screenshots**
image


0x5bfa:
How about this?

- When user entered `>`, display custom glyph with background accent color filled.
  image

hishitetsu:
> * When user entered `>`, display custom glyph with background accent color filled.

Not implemented yet. In fact, I'm not sure how to implement it.

0x5bfa:
Then, this gonna be step 2.

mdtauk:
> > * When user entered `>`, display custom glyph with background accent color filled.
> 
> Not implemented yet. In fact, I'm not sure how to implement it.

I would suggest perhaps looking at the Tokenizing TextBox from the Community Toolkit
image

Or just modifying the template to shift the Text part 32px in from the left, and overlay an icon?

0x5bfa:
> Or just modifying the template to shift the Text part 32px in from the left, and overlay an icon?

I would rather do this.

- if you press backspace, that indicator would disappear. *vice versa*.

src/Files.App/Actions/Open/OpenCommandPaletteAction.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) 2023 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.App.Actions
+{
+	internal class OpenCommandPaletteAction : IAction
+	{
+		private readonly IContentPageContext context;
```
0x5bfa, line 8: _context is desired.

src/Files.App/Data/Items/NavigationBarSuggestionItem.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) 2023 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.App.Data.Items
+{
+	public class NavigationBarSuggestionItem : ObservableObject
```
0x5bfa, line 6: observing is really needed? theres no two way bindings though.

src/Files.App/Data/Items/NavigationBarSuggestionItem.cs
```diff
@@ -0,0 +1,43 @@
...
+	{
+		private string? _Text;
+		public string? Text
+		{
+			get => _Text;
+			set => SetProperty(ref _Text, value);
+		}
+
+		private string? _PrimaryDisplay;
+		public string? PrimaryDisplay
```
0x5bfa, line 16: can you describe more? like header, description, hotkey, path. or you concern about we might use this in different purpose? i dont think we need to think of it.

src/Files.App/ViewModels/UserControls/ToolbarViewModel.cs
```diff
@@ -751,56 +778,80 @@ public async Task SetAddressBarSuggestions(AutoSuggestBox sender, IShellPage she
...
+						PrimaryDisplay = "NavigationToolbarVisiblePathNoResults".GetLocalizedResource() } };
 					}
 
 					// NavigationBarSuggestions becoming empty causes flickering of the suggestion box
 					// Here we check whether at least an element is in common between old and new list
-					if (!NavigationBarSuggestions.IntersectBy(suggestions, x => x.Name).Any())
+					if (!NavigationBarSuggestions.IntersectBy(suggestions, x => x.PrimaryDisplay).Any())
 					{
 						// No elements in common, update the list in-place
 						for (int si = 0; si < suggestions.Count; si++)
```
0x5bfa, line 846: si to index.

0x5bfa (commented):
tiny things

src/Files.App/Data/Items/NavigationBarSuggestionItem.cs
hishitetsu: > you concern that we might use this in different purpose?

Yes, I also use this to suggest paths. Right now I only use PrimaryDisplay for path suggestions, but it can be extended to display both localized folder names and path, for example.

0x5bfa: Can you add xml commes describing the purpose(like, this is used for command palette item header) for each?

src/Files.App/Data/Items/NavigationBarSuggestionItem.cs
hishitetsu: Yes, we may reuse existing NavigationBarSuggestionItem instances instead of replacing them with new NavigationBarSuggestionItem instances when the item changes for smooth display.

0x5bfa (commented):
If you're not using XAML Styler, please use it to follow our code guidelines.

- VS2017/VS2019, download from [XAMLStyler](https://marketplace.visualstudio.com/items?itemName=TeamXavalon.XAMLStyler).
- VS2022, download from [XAMLStyler2022](https://marketplace.visualstudio.com/items?itemName=TeamXavalon.XAMLStyler2022).

src/Files.App/Data/Items/NavigationBarSuggestionItem.cs
0x5bfa: ok

0x5bfa (approved):
LGTM in terms of codebase quality

src/Files.App/UserControls/AddressToolbar.xaml
```diff
@@ -137,8 +138,45 @@
...
+								<!--  Primary Display  -->
+								<!--  This is used to display paths or command descriptions.  -->
+								<TextBlock
+									Grid.Column="0"
+									Opacity="{x:Bind DisplayOpacity, Mode=OneWay}"
+									Text="{x:Bind PrimaryDisplay, Mode=OneWay}" />
+
+								<!--  Supplementary Display  -->
+								<!--  This is used to display command hotkeys.  -->
+								<TextBlock
```
mdtauk, line 161: Can we use the standard theme brushes, instead of adding opacity to elements.

``TextFillColorSecondaryBrush`` or ``TextFillColorTertiaryBrush``

hishitetsu: Done.

yaira2:
@hishitetsu I tested the feature and it's working really well, the only feedback I have is that we should remove the command name, and that we should hide commands that aren't enabled.

hishitetsu:
> we should remove the command name

The command name is important for non-English users because they will want to search for commands both in their native language and in English. Would it be better to show the English description rather than the command name, which is what Visual Studio Code does?
image

> we should hide commands that aren't enabled

I don't think so because users may want to know what commands exist. But you want to hide them, I will do so.

yaira2:
@hishitetsu I meant only to hide the name, we can still include the name in the search query.

> I don't think so because users may want to know what commands exist. But you want to hide them, I will do so.

I personally found them confusing, but we can always reevaluate down the line and besides, I think this concern will be resolved when we add a settings page for remapping shortcuts.

hishitetsu:
@yaira2 updated. How about this?

yaira2 (approved):
LGTM

mdtauk:
image

Is there any value in only using the Semibold font for the characters you have typed into the Command Box?
image

hishitetsu:
> Is there any value in only using the Semibold font for the characters you have typed into the Command Box?

I tried to find it, but could not.
https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.autosuggestbox?view=windows-app-sdk-1.3

0x5bfa:
that must be implemented by themselves.

mdtauk:
Its not an automatic feature, we would have to parse the string, and provide probably Runs within TextBlocks with the characters styled.

mdtauk:
> > Its not an automatic feature, we would have to parse the string, and provide probably Runs within TextBlocks with the characters styled.
> 
> Implemented in #13095
>

Wow that was fast work, thank you.  I think it does make a difference in polish and adds a confirmation to the user

mdtauk:
If we can figure out how to add an Icon when the box is in the Command mode, should it match this?
image

0x5bfa:
yep. we can create custom control.

mdtauk:
Lets keep the icon simple and not too obnoxious.  It will probably require a ``> `` with a space to enter into the command mode, and switch the visuals of the AddressBar control.  Backspacing will exit out of the mode.

image
image

mdtauk:
@yaira2  What is the current status of the Command Palette?  I saw mention about adding a button to switch to the commanding mode as well as reading the ``>`` input char?

PR #12995, opened 2023-07-20
Fix: Folders in columns view cannot be deselected

### LINKED ISSUE #12969: Folders in columns view cannot be deselected & do not show checkboxes
### Description

When using columns view in a directory, once a subfolder in this view is selected, it's impossible to deselect it unless going inside another subfolder.

In parallel to this, we can see that there are no checkboxes for items in columns view, despite me having this option toggled on in the settings, as you can see it working before going in the columns view folder (let me know if I should split this in another issue).

[GIF](https://i.imgur.com/Gje2RZD.gif)

### Steps To Reproduce

1. Enable columns view for a specific folder
2. Click on a folder when in columns view to select it
3. Click on an empty space to deselect it (this is what fails)

### Requirements

The selected folder should be deselected when clicking on an empty space, just like it does in other viewing modes (details, cards, etc.)

### Files Version

2.5.10.0

### Windows Version

Windows 11 Pro 22H2 22621.1992

### Log File

No relevant logs to share

ferrariofilippo:
**Resolved / Related Issues**
   Closes #12969 

**Validation**
   1. Open a folder
   2. Set layout to `Columns`
   3. Select a folder
   4. Click on an empty space

### DIFF
--- a/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See the LICENSE.
 
 using CommunityToolkit.WinUI.UI;
-using Files.App.Data.Commands;
 using Files.App.UserControls.Selection;
 using Microsoft.UI.Input;
 using Microsoft.UI.Xaml;
@@ -88,7 +87,7 @@ namespace Files.App.Views.LayoutModes
 
 		protected override void ItemManipulationModel_FocusSelectedItemsInvoked(object? sender, EventArgs e)
 		{
-			if (SelectedItems.Any())
+			if (SelectedItems?.Any() ?? false)
 			{
 				FileList.ScrollIntoView(SelectedItems.Last());
 				(FileList.ContainerFromItem(SelectedItems.Last()) as ListViewItem)?.Focus(FocusState.Keyboard);
@@ -163,11 +162,11 @@ namespace Files.App.Views.LayoutModes
 			StartRenameItem("ListViewTextBoxItemName");
 		}
 
-		private void ItemNameTextBox_BeforeTextChanging(TextBox textBox, TextBoxBeforeTextChangingEventArgs args)
+		private async void ItemNameTextBox_BeforeTextChanging(TextBox textBox, TextBoxBeforeTextChangingEventArgs args)
 		{
 			if (IsRenamingItem)
 			{
-				ValidateItemNameInputText(textBox, args, (showError) =>
+				await ValidateItemNameInputText(textBox, args, (showError) =>
 				{
 					FileNameTeachingTip.Visibility = showError ? Visibility.Visible : Visibility.Collapsed;
 					FileNameTeachingTip.IsOpen = showError;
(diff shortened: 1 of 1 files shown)


hishitetsu:
This makes it impossible to focus on a parent folder column while keeping a subfolder column open. I don't think folders should be deselected when clicking on a column that is not in focus.

hishitetsu:
> We might poll the community on ds and see what they choose

That's a good idea. We don't have a solid concept for the column layout, so we should get feedback and firm up the specs.

hishitetsu:
> if you select a folder and then click it again the right blade will be closed.

I would like the folder to stay open and not close in this case. This behavior is not intuitive and makes little sense if the folder can be closed by clicking on the empty space.

yaira2:
I think the expected behavior is to close the folder, after all, the selection indicator is intended to identify the opened folders.

hishitetsu:
> > if you select a folder and then click it again the right blade will be closed.
> 
> I would like the folder to stay open and not close in this case. This behavior is not intuitive and makes little sense if the folder can be closed by clicking on the empty space.

@yaira2 do you think the folder should still be closed in this case? The selection indicator is there but the folder is not open.
image

yaira2:
I didn't realize we were talking about right click, I don't have a preference in this case.

hishitetsu:
> I didn't realize we were talking about right click, I don't have a preference in this case.

No, no, it's all about left click. Each time clicking on the selected folder, the folder opens or closes.
Also, clicking on an empty space while the folder is closed will deselect the folder.

https://github.com/files-community/Files/assets/66369541/0edc5ddb-52ca-4ec2-8efc-094a18b65e7f

yaira2:
> Each time clicking on the selected folder, the folder opens or closes

When the folder is closed, shouldn't we remove the selection indicator?

hishitetsu:
I agree. The question is whether a folder should be closed when clicking on an open folder.

yaira2:
I don't have a preference

ferrariofilippo:
I like that behaviour because with it, I can keep columns tidier when I need to interact with one element of the last column
IIRC we haven't receive any complaints since we introduced it

hishitetsu:
> IIRC we haven't receive any complaints since we introduced it

I think it's because currently we can't close a folder by clicking on an empty space. If we can close a folder by clicking on an empty space, I don't see the need for the ability to click on a folder to close it.
Personally, I don't want folders opening and closing when I click on a folder to just rename it.

ferrariofilippo:
@hishitetsu can you check if it works as expected?

src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
```diff
@@ -163,11 +162,11 @@ override public void StartRenameItem()
 			StartRenameItem("ListViewTextBoxItemName");
 		}
 
-		private void ItemNameTextBox_BeforeTextChanging(TextBox textBox, TextBoxBeforeTextChangingEventArgs args)
+		private async void ItemNameTextBox_BeforeTextChanging(TextBox textBox, TextBoxBeforeTextChangingEventArgs args)
 		{
 			if (IsRenamingItem)
 			{
-				ValidateItemNameInputText(textBox, args, (showError) =>
+				await ValidateItemNameInputText(textBox, args, (showError) =>
```
hishitetsu, line 169: Unnecessary async/await is removed in #12010. Is there a reason why async/await should be used?

hishitetsu (commented):
Great! It is working exactly as I expected.

src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
ferrariofilippo: I think something has changed since then, now that method returns an ` async Task`

hishitetsu: The warning can also be avoided by adding `_ = `. I think either is fine.

hishitetsu (approved):
LGTM

src/Files.App/Views/LayoutModes/ColumnViewBase.xaml.cs
ferrariofilippo: At the moment it uses ` await `  . Would it be better with ` _ ` ?

PR #13023, opened 2023-07-21
Fix: Fixed crash when launching from CMD with "Files"

### LINKED ISSUE #13019: Launching first instance of Files with command line arguments doesn't work anymore
### Description

In 2.5.20.0 launching Files (when no instance is currently open) with command line arguments specifying the folder to open causes the app to get stuck on the loading screen. This used to work in the previous update, and still works if Files is already open, it is only it is opened for the first time this way.

All following are using Powershell, interestingly CMD causes Files to crash entirely everytime which is another thing to investigate. Launching via the 'Run' dialog (Win + R) has the same results as PowerShell, as does the WSL.

- Launching using no arguments works fine

image

- However launching with arguments causes it to get stuck on the loading screen.

image

- Launching with Powershell's "Start-Process" also causes this.

image

- If Files is already open it works perfectly, opening a new tab at the specified directory, which is how it should work if it isnt open yet too.

image

This has been a big problem for me as I heavily use AutoHotkey keybinds to control my PC, as shown below I had setup and frequently used keybinds which let me instantly launch files at some frequently used directories, as shown below. This was working in the previous update but this new update has broken them (unless I already have Files open)
image

I assume there arent any other ways to launch from command line? I looked in ProcessExplorer at the strings of Files and couldn't see anything obvious. 
P.S. Would be amazing to get some command line arguments for Files. Stuff like: open in new tab, open in new window, add/remove Tag etc would be amazing to have as it would allow for integration into other systems such as automatically apply designated tags to files created by a program.

### Steps To Reproduce

1. Close all instances of Files
2. From a Powershell/WSL terminal, or via the Run dialog run 'files.exe C:\' (error happens when using the .exe extension and without, and with any path so you can use something else if you want)
3. You will observe that Files stays stuck on the loading screen 

Also:
1. Close all instances of files
2. Using CMD, run the command "files" or "files.exe" (The crash happens both with and without an argument applied)
3. You will see that files flashes up and then crashes 

### Requirements

1. Fix bug causing Files to get stuck on the loading screen when it is launched with command line arguments when no other instances are already open.
2. Fix immediate crash when launching Files from a command prompt window when no other instances are open 

### Files Version

2.5.20.0

### Windows Version

10.0.19045.3208

### Log File

[debug.log](https://github.com/files-community/Files/files/12123228/debug.log)

hishitetsu:
**Resolved / Related Issues**
   Closes #13019 

**Validation**
   Launch the app from CMD with "files C:\". You will see the app launches successfully.

### DIFF
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -73,7 +73,9 @@ namespace Files.App
 			switch (activatedEventArgs)
 			{
 				case ILaunchActivatedEventArgs launchArgs:
-					if (launchArgs.Arguments is not null && launchArgs.Arguments.Contains($"files.exe", StringComparison.OrdinalIgnoreCase))
+					if (launchArgs.Arguments is not null &&
+						(CommandLineParser.SplitArguments(launchArgs.Arguments, true)[0].EndsWith($"files.exe", StringComparison.OrdinalIgnoreCase)
+						|| CommandLineParser.SplitArguments(launchArgs.Arguments, true)[0].EndsWith($"files", StringComparison.OrdinalIgnoreCase)))
 					{
 						// WINUI3: When launching from commandline the argument is not ICommandLineActivatedEventArgs (#10370)
 						var ppm = CommandLineParser.ParseUntrustedCommands(launchArgs.Arguments);


pierrishm1987:
Great work

PR #13043, opened 2023-07-23
Feature: Added support for elevated file operations

### LINKED ISSUE #10422: Feature: Implement support for elevated operations

After we remove the full trust process, we need to implement a new model for elevated operations. 
Users can already open Files as administrator to work around this issue today, but it's still *better to have* from the perspective of user experience.

### Requirements

- For elevated operations, use an elevated COM component to execute them
- Possible blocker: https://github.com/microsoft/WindowsAppSDK/issues/3092

### Files Version

2.3.58

### Windows Version

Windows 11 25236

### Comments

_No response_

gave92:
**Resolved / Related Issues**
   Closes #10422

**Validation**
   1. Test copy, pasting, renaming, deleting, creating a file in C:

This PR (re)adds basic support for performing file operations as admin using a separate process. This is meant as a stopgap while we wait for a fix to https://github.com/microsoft/WindowsAppSDK/issues/3092.
To keep things simple: no progress, no cancellation, no undo/redo, ~no error feedback~.

**Screenshots (optional)**
image


src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
```diff
@@ -792,8 +806,34 @@ private void CancelOperation(object operationID)
...
+				using Process process = new Process();
+				process.StartInfo.UseShellExecute = true;
+				process.StartInfo.Verb = "RunAs";
+				process.StartInfo.FileName = Path.Combine(Windows.ApplicationModel.Package.Current.InstalledPath, "Files.App.Elevated", "Files.App.Elevated.exe");
+				process.StartInfo.Arguments = string.Join(" ", "FileOperation", request.ID);
+				await SafetyExtensions.IgnoreExceptions(() =>
+				{
+					process.Start();
+					return process.WaitForExitAsync();
+				});
```
ferrariofilippo, line 834: I think it might be better to handle the exception there, instead of ignoring it. We would be able to report `Operation failed` in the `Status center`, whenever a user rejects the second popup

gave92: Good idea. I can also try to return success of the operarion through the return value of the process.

yaira2:
> To keep things simple: no progress, no cancellation, no undo/redo, no error feedback

Should we notify the user beforehand that it won't be available? Alternatively, would it be easy to display File Explorers transfer UI in this case?

gave92:
Well you still see that the operation is ongoing in the status center.
image
Maybe we can show "Progress not available" instead of the progress line and disable the cancel button?

gave92:
Wait I'm feeling like an idiot.

gave92:
Ook realized there was a better way to do this. No need for a separate process and still get progress and error reporting :)

yaira2 (approved):
LGTM

yaira2:
@gave92 Thank you for working on this, it's has been a paint point since migrating to WinAppSdk and I know a lot of people are going to be very happy it's resolved.

gave92:
Is it just me or does this thing not work anymore? :/ Even checking out this PR admin operations fail.

yaira2:
It prompts for access but then fails to do the operation.

gave92:
Uff was a change in windows then? Could swear it was working grr.

yaira2:
I know it was working, I tested before merging 😄

gave92:
Confirmed: "KB5028185 Windows 11 Cumulative Update Build 22621.1992" breaks admin file operations for packages apps 😢. Just tried to go back to build 22621.1848 and tadaaa admin operations work again 😢. @yaira2 do you think we can notify anyone of this?

yaira2:
Is there an issue or post from Microsoft about this regression?

gave92:
None that I can find

riverar:
I can't reproduce this on `10.0.22621.2134`, can you see if this works for you? You should end up with a test.txt in `C:\Windows`.
[app.zip](https://github.com/files-community/Files/files/12413607/app.zip)

gave92:
Note that the issue is only with Packaged apps.

riverar:
Packaged the app and can reproduce failure but it never works. Working to try and get the repro project into a state that mimics the behavior we're seeing here with Files.

riverar:
Still can't reproduce this issue outside of Files, there's certainly no indication of an OS issue. Maybe there's a Files threading issue at play here?

gave92:
Will send repro app, thanks for checking.

riverar:
> Can you repro for Packaged apps then?

No, sorry for the confusion. When I posted that comment, my app was running in lowIL due to a manifest typo (that interestingly enough didn't cause an error). I since fixed that and now cannot reproduce.

Latest compiled sample: [packaged_app.zip](https://github.com/files-community/Files/files/12431067/packaged_app.zip)
Code: https://github.com/riverar/repro-1012c50820aa (`Run ./Package.ps1`)

hishitetsu:
When I run it in debug, it appears that the file operation has been processed before the UAC prompt appears and fails with an authorization error.

hishitetsu:
Attempting to perform a elevated file operation will result in the following exception. Would this be of any help in the analysis?
スクリーンショット 2023-10-03 173903

gave92:
Given that it used to work, this looks like an MS bug. I'm not sure we can fix it :(

hishitetsu:
I know it used to work too.
@yaira2 If this issue is not resolved by the time v3 is released, wouldn't it be better to revert this PR?

gave92:
It's not like without this admin ops would start to work so probably it does not matter. But it's annoying to miss out admin ops in v3 :(
I'm tempted to take commit 7ad5f78 which uses a different method and should work.

hishitetsu:
> It's annoying to miss out admin ops in v3 :(

Yeah but it's more annoying that admin ops fail even though clicking on the shield icon.

> I'm tempted to take commit [7ad5f78](https://github.com/files-community/Files/commit/7ad5f78db2579789cda795f06cbe399dfb5aa46e) up again. It uses a different method and should work.

I'd like to try that too.

hishitetsu:
Really? I have no success even if I have disabled UAC. Anyway, I think this is a blocker to release v3.

Issue #11326, opened 2023-02-16
Feature: Disable the display of Cloud Drive space information.


hishitetsu:
### What feature or improvement do you think would benefit Files?

As noted in #10001, the Cloud Drive space information is not displayed correctly.
But it may be a lot of work to get this right, so we should probably eliminate the display until we can fix it.
I think it is better not to display the wrong information than to display the wrong information.

### Requirements

Disable the display of Cloud Drive space information until we can display the right information.

### Files Version

Version: 2.4.33.0

### Windows Version

Windows 11 22H2 22621.1265

yaira2:
I think we should try and fix this, there might be an api we can use (similar to how we get the file status)

hishitetsu:
In Files, Getting the file status is already possible using the offline API. I am going to use this to fix #11318 . But there doesn't appear to be an offline API to retrieve the cloud drive capacity.

We can do it by using online APIs, but it requires the account authentication which is not currently required.
Using online APIs is a bit hard work and each cloud drive has different APIs, so it would be better to not display capacity until it can be implemented, and only display that have been implemented.

yaira2:
We can add custom logic for each cloud provider, the default should be to disable the capacity but with support for enabling it if possible.

hishitetsu:
Yes, that would be good.

PR #13052, opened 2023-07-24
Feature: Redesigned the left hand sidebar

### LINKED ISSUE #12930: Feature: Display sidebar options when right clicking headers

Hi,

I was recently struggling to show the libraries section on the sidebar. I tried to show options by clicking the parent menu items like "Drive", "Home" to get the options menu, because I couldn't find it anywhere in the settings. Then I was told that I need to right click on a small blank space to change the menu items options. I think that's a bad UX and many people wouldn't find this on their own.

I clicked where the red dot is:
obrazek

### Requirements

- Move the menu items options context menu to the settings.

### Files Version

2.5.10.0

### Windows Version

10.0.22631.1972

### Comments

_No response_

chingucoding:
This PR migrates the SideBar implemenation away from the NavigationView control to our own written control, making feature development and styling a lot easier.

Fixes #12930

ToDos:
- [x] ~Update minimal pane to cover entire side (if desired)?~
- [x] Animate expanding/collapsing pane
- [x] Animate expanding/collapsing items
- [x] Update styling of sidebar to match design
- [x] Keyboard handling
- [x] Proper drag and drop handling
- [x] Fix initial state being wrong
- [x] Fix cursor when hovering over the Grid Splitter
- [x] Add back double click (on the Grid Splitter) to collapse/expand the sidebar
- [x] Fix issue where right clicking item triggers a click
- [x] Add tooltips (big tooltips for drives are another issue)
- [x] Reordering favourites might crash and reset favourites
- [x] Double clicking resizer leaves it in weird visual state
- [x] Restore pane width between sessions
- [x] Restore compact mode between sessions
- [x] Fix icon in the network section
- [x] Fix the tooltip for OneDrive

**Screenshots (optional)**
Animation showing the different behaviors for the new sidebar


chingucoding:
@mdtauk and @0x5bfa, I would love your feedback on this!

mdtauk:
Is there a reason you have gone with a UserControl instead of a Custom Control?

chingucoding:
> Is there a reason you have gone with a UserControl instead of a Custom Control?

The performance benefit for a single use control is neglible in my opinion compared to the ease of development you get with a UserControl.

mdtauk:
Will this control not eventually be used for the Settings UI and for the Properties dialog?

chingucoding:
> Will this control not eventually be used for the Settings UI and for the Properties dialog?

This was not the initial plan but I don't see a reason why not. Of course, @yaira2 has the say on this one though.

0x5bfa:
Why not use SplitView?

I you'd like to have a look at the initial concreate example of mine: https://github.com/0x5bfa/SidebarSampleApp

chingucoding:
> Why not use SplitView? that has overlay mode view, to easily implement.

Mainly because that would make resizing the sidebar very difficult, either by having to retemplate or having to work around not being able to properly place the respective resizer control there.

yaira2:
Is there a state where the icons are hidden?
image

image

0x5bfa:
Then, is there any idea to implement overlay mode?

chingucoding:
Oh I see where those questions come from, I've used the wrong GIF to show what is present. Yes, there is a overlay pane/no icon mode @yaira2 and @0x5bfa

yaira2:
> Will this control not eventually be used for the Settings UI and for the Properties dialog?

@mdtauk that's the goal.

yaira2:
> Update minimal pane to cover entire side

I'm okay with the implementation as is and I think it'll be easier to support the control without having to cover the entire side.

mdtauk:
The toggle button for opening the pane in Overlay mode.  It's it possible to arbitrarily place that with some logic to show and hide it as needed.

I think placing it into the address bar, before the back and forward buttons makes the most sense, with the sidebar sliding in below.

@yaira2 has been wanting to add a tab management button into the top left before the first tab for a while now.  But having the toggle at the very top would cover that up

mdtauk:
As for resizing, have you considered or idea for the raise visual @yaira2 

Not sure if you've mentioned it to @chingucoding

yaira2:
> I think placing it into the address bar, before the back and forward buttons makes the most sense, with the sidebar sliding in below.

This is fine with me 👍

yaira2:
> As for resizing, have you considered or idea for the resize visual

It's out of scope for this PR.

mdtauk:
> > As for resizing, have you considered or idea for the resize visual
> 
> It's out of scope for this PR.

So the visual and how that would work, is not related to the Resizing of the sidebar?

yaira2:
> So the visual and how that would work, is not related to the Resizing of the sidebar?

It'll be the same as the current design

mdtauk:
> > So the visual and how that would work, is not related to the Resizing of the sidebar?
> 
> It'll be the same as the current design

Okie dokie.

0x5bfa:
> > Is there a reason you have gone with a UserControl instead of a Custom Control?
> 
> The performance benefit for a single use control is neglible in my opinion compared to the ease of development you get with a UserControl.

@chingucoding Can you consider to use ResourceDictionary instead of using UserControl? Someone definitely will create ViewModel of Host & Pane control to isolate commands. But I'd like them to be independent controls like MUXC.

0x5bfa:
SidebarItem should derive drom ButtonBase to use Click and Command parameters.

0x5bfa:
Shouldn't SidebarItem derive drom ButtonBase to use Click and Command parameters?

chingucoding:
I'm still inclined to say that there is little to no benefit to use a templated control instead of a user control. Regarding your points:

> We have to put a bunch of commands like Click (senario: invoking OpenNewTab on clicked item) into "the control classes".

Can you elaborate what you mean? If this is referring to the *.commands.cs file and similar scenarios, you are right that this should not be handled inside the SideBar control. Instead, a ViewModel should bind to the ItemInvoked/ItemContextRequested commands instead which would be cleaner regarding the separation of concerns.

> We are going to need ViewModels of the Pane and the Host. However, creating them for the Controls makes them less maintainable due to less isolation.

Why do we need a viewmodel for the pane? Ideally, you would only interact with the host and implement a viewmodel that provides:
* A selection
* A list of items
* Something handling item invocation
* Something handling item context requested (if supported)

> Also, shouldn't SidebarItem derive drom ButtonBase to use Click and Command parameters?

Since no one should use a SidebarItem on their own, no. Those properties should be handled on the host/pane level not the item level.

0x5bfa:
- SidebarView (template) - comprehensive control
  ```xaml
  <SidebarView>
    <Button x:Name="Toggle">
    <SidebarSplitView>
  </SidebarView>
  ```
- SidebarSplitView (template) - implement custom SplitView
- SidebarItem (template) - implement custom clickable button item
- SidebarHost (usercontrol) - handle all commands

is ideal imo.

> Why do we need a viewmodel for the pane?

Misunderstood. my bad.

> Can you elaborate what you mean?

The SidebarView (currently SidebarHost) should only trigger event handlers ideally. and handle them in the new SidebarHost which is UserControl, having ViewModel.

> Since no one should use a SidebarItem on their own, no. Those properties should be handled on the host/pane level not the item level.

I just now know NavigationViewItem doesn't derive from ButtonBase but Control. My bad also.

mdtauk:
If this control will be used for Settings and Properties, the Selected SideBarItem will need to pass it's Selection state to the Icon - so we can use the correct states and brushes.

Also could you tell me when the DragTargetIndicator shows?  Is it for Re-arranging, or dragging items off the Sidebar, or dragging files and folders, to the sidebar?

mdtauk:
Oh also, do we have some kind of differentiator for the header items, so we can hide the Icons and style the text on these differently to the other items?

We need to make sure they remain when the Sidebar is in its Collapsed state, where the other items are moved to the flyouts.

We also don't have pinned indicators for the Favourites/Pinned section 
image

chingucoding:
> The SidebarView (currently SidebarHost) should only trigger event handlers ideally. and handle them in the new SidebarHost which is UserControl, having ViewModel.

@0x5bfa So essentially just rename Sidebarhost and have the ViewModel just handle the listeners and move all the logic that is currently in those handlers in that ViewModel?

> If this control will be used for Settings and Properties, the Selected SideBarItem will need to pass it's Selection state to the Icon - so we can use the correct states and brushes.

We will find a way to handle this, maybe with a custom template for that case.

> Also could you tell me when the DragTargetIndicator shows? Is it for Re-arranging, or dragging items off the Sidebar, or dragging files and folders, to the sidebar?

Currently it is only used to indicate when an item is getting dropped onto it, however it is also capable of re-arranging the items.

> Oh also, do we have some kind of differentiator for the header items, so we can hide the Icons and style the text on these differently to the other items?

What exactly makes an item a header item? If it's just the ability to have children then we can add the necessary VisualStates on the SideBarItem.

> Are the child items grouped in some container? I would like to adjust the padding below these items, so each section has some visible separation - without having to add dummy items to move things down.

Yes they are.

> I also want the Home icon to be above the others with space below that, so maybe a StartItems section?

Yeah we can add this, I would call that "PaneHeader" and we could also handle the settings element by adding a "PaneFooter" element. @yaira2 is this something we want to add as part of this PR?

mdtauk:
> > The SidebarView (currently SidebarHost) should only trigger event handlers ideally. and handle them in the new SidebarHost which is UserControl, having ViewModel.
> 
> @0x5bfa So essentially just rename Sidebarhost and have the ViewModel just handle the listeners and move all the logic that is currently in those handlers in that ViewModel?
> 
> > If this control will be used for Settings and Properties, the Selected SideBarItem will need to pass it's Selection state to the Icon - so we can use the correct states and brushes.
> 
> We will find a way to handle this, maybe with a custom template for that case.
> 
> > We also don't have pinned indicators for the Favourites/Pinned section
> 
> Yeah we can make this work.
> 
> > Also could you tell me when the DragTargetIndicator shows? Is it for Re-arranging, or dragging items off the Sidebar, or dragging files and folders, to the sidebar?
> 
> Currently it is only used to indicate when an item is getting dropped onto it, however it is also capable of re-arranging the items.
> 
> > Oh also, do we have some kind of differentiator for the header items, so we can hide the Icons and style the text on these differently to the other items?
> 
> What exactly makes an item a header item? If it's just the ability to have children then we can add the necessary VisualStates on the SideBarItem.
> 
> > Are the child items grouped in some container? I would like to adjust the padding below these items, so each section has some visible separation - without having to add dummy items to move things down.
> 
> Yes they are.
> 
> > I also want the Home icon to be above the others with space below that, so maybe a StartItems section?
> 
> Yeah we can add this, I would call that "PaneHeader" and we could also handle the settings element by adding a "PaneFooter" element. @yaira2 is this something we want to add as part of this PR?
> 
> @mdtauk Is there a complete design file which shows all features we want to add? Adding feature after feature makes planning the control a bit hard 😅

I did send some spec notes to Yair.

image

image

image

---

```
Visual States -----------------------

SidebarNavigation ---

LAYOUTGROUP
Expanded
Compacted
Overlay

---

SidebarNavigationItem ---

STATEGROUP
Normal
PointerOver
Pressed
Disabled
Selected
PointerOverSelected
PressedSelected
DisabledSelected

LAYOUTGROUP
Expanded
Compacted

STATUSGROUP
Pinned
Unpinned

PRESENTERGROUP
Inline
Flyout

---

SidebarNavigationGroup ---

STATEGROUP
Normal
PointerOver
Pressed
Disabled

EXPANSIONGROUP
Opened
Closed

LAYOUTGROUP
Expanded
Compacted

---

SidebarNavigation.BottomItems ---

DIVIDERGROUP
DividerVisible
DividerCollapsed

LAYOUTGROUP
Expanded
Compacted

---

ResizeVisual ---

STATEGROUP
Normal          (Invisible at Rest)
PointerOver     (Fade In Transition)
Pressed

ORIENTATIONGROUP
Vertical
Horizontal





Template Parts + Properties ----------------------

SidebarNavigationItem ---

PART_Content (Label)
PART_Icon (FontIcon, Icon, SymbolIcon, PathIcon, ThemedIcon, WinShellIcon, BitmapIcon, etc)
PART_PinnedIndicatorIcon
PART_SelectionIndicator

PROP_IsEnabled
PROP_IsPinned
PROP_IsSelected
PROP_Background
PROP_Foreground
PROP_IsCompact

---

SidebarNavigationGroup ---
PART_Content (Label)
PART_Icon (FontIcon, Icon, SymbolIcon, PathIcon, ThemedIcon, WinShellIcon, BitmapIcon, etc)
PART_ExpansionIndicator (Chevron)

PROP_IsEnabled
PROP_IsOpened
PROP_Background
PROP_Foreground
PROP_IsCompact      (if IsCompact = True, then close and hide sub-items and use flyout instead)

---

SidebarCompactFlyout ---    (this will present child SidebarNavigationItems in a flyout)

ResizeVisual ---

PART_FlyoutBase

PROP_TrackMode      (TrackMode.Static - TrackMode.Vertical - TrackMode.Horizontal)
PROP_Orientation    (Vertical - Horizontal)
```

image

mdtauk:
Oh and for the sake of being complete, here is how I think the SideBar should look in High Contrast
image

yaira2:
> Yeah we can add this, I would call that "PaneHeader" and we could also handle the settings element by adding a "PaneFooter" element. @yaira2 is this something we want to add as part of this PR?

@chingucoding we want to try and match the design as close as possible, the exception is the moving the settings button to the sidebar which is out of scope for this PR.

mdtauk:
> > Yeah we can add this, I would call that "PaneHeader" and we could also handle the settings element by adding a "PaneFooter" element. @yaira2 is this something we want to add as part of this PR?
> 
> @chingucoding we want to try and match the design as close as possible, the exception is the moving the settings button to the sidebar which is out of scope for this PR.

Yea actually placing the button there is not for now, but the control will need to support it for the future.  He was asking for a spec of the features for the control.

yaira2:
I'm really looking forward to this, it's a lot easier to work with a custom control design for a specific scenario than it is to work with a generic control that's trying to do everything at once.

mdtauk:
> I'm really looking forward to this, it's a lot easier to work with a custom control design for a specific scenario than it is to work with a generic control that's trying to do everything at once.

Custom Controls do take more effort to make, but provide a simpler dev experience I think.  I don't have any details on which is more performant, or flexible, or robust/less error prone.

CustomControls or UserControls

0x5bfa:
https://github.com/files-community/Files/issues/13067 can we talk in this issue and hide all conversations here? @yaira2 @mdtauk

mdtauk:
When you reach the point you are working on Animations, Might I suggest we follow what the Expander control does, and use the same system provided AnimatedVisual and AnimatedIcon

```xml
<local:AnimatedIcon x:Name="ExpandCollapseChevron"
                    Width="{StaticResource ExpanderChevronGlyphSize}"
                    Height="{StaticResource ExpanderChevronGlyphSize}"
                    HorizontalAlignment="Center"
                    VerticalAlignment="Center"
                    local:AnimatedIcon.State="NormalOff"
                    AutomationProperties.AccessibilityView="Raw"
                    Foreground="{ThemeResource ExpanderChevronForeground}"
                    RenderTransformOrigin="0.5, 0.5">
    <animatedvisuals:AnimatedChevronUpDownSmallVisualSource />
    <local:AnimatedIcon.FallbackIconSource>
        <local:FontIconSource FontFamily="{StaticResource SymbolThemeFontFamily}"
                                FontSize="12"
                                Glyph="{StaticResource ExpanderChevronDownGlyph}"
                                IsTextScaleFactorEnabled="False" />
    </local:AnimatedIcon.FallbackIconSource>
    <local:AnimatedIcon.RenderTransform>
        <RotateTransform x:Name="ExpandCollapseChevronRotateTransform" Angle="0" />
    </local:AnimatedIcon.RenderTransform>
</local:AnimatedIcon>
 ```

and for the visual states

```xml
<Setter Target="ExpandCollapseChevron.(local:AnimatedIcon.State)" Value="PointerOverOff" />
```

NavigationView's animation when expanding seems to work nicely opening and closing - compared to the Expander control.

mdtauk:
@yaira2  you were working on the compact width to try to get it to be 56...

**SideBarView.xaml**
```xml
<VisualState x:Name="Compact">
    <VisualState.Setters>
        <Setter Target="DisplayColumn.Width" Value="56" />
        <Setter Target="MenuItemsHost.MaxWidth" Value="56" />
```

mdtauk:
> You may want smooth background transition like a Button.
> 
> Here's code.
> 
> ```
> <Grid.BackgroundTransition>
>     <BrushTransition Duration="0:0:0.083" />
> </Grid.BackgroundTransition>
> ```

I did experiment with a duration for the fade of a hover state, but it felt a bit laggy.

I did put a height animation on the Selection Indicator when moving to the Selected and Pressed States.

0x5bfa (commented):
If you're not using XAML Styler, please use it to follow our code guidelines.

- VS2017/VS2019, download from [XAMLStyler](https://marketplace.visualstudio.com/items?itemName=TeamXavalon.XAMLStyler).
- VS2022, download from [XAMLStyler2022](https://marketplace.visualstudio.com/items?itemName=TeamXavalon.XAMLStyler2022).

Can you figure out why the pipeline is keeping failed?

chingucoding:
I hope by adding the automation peers, the CI is finally unblocked. I'll also try to format the file more often, sorry about this @0x5bfa .

0x5bfa:
> I hope by adding the automation peers, the CI is finally unblocked. I'll also try to format the file more often, sorry about this @0x5bfa .

Finally! dont be(if downloaded it, you can apply every time when save with hotkey Ctrl+S), your work is really great!

mdtauk:
Is it just me getting an error when trying to build the last day or two?

```
'SidebarView' does not contain a definition for 'DisplayColumnGrid_ContextRequested' and no accessible extension method 'DisplayColumnGrid_ContextRequested' accepting a first argument of type 'SidebarView' could be found (are you missing a using directive or an assembly reference?)
```

chingucoding:
I'm inclined to say this is something on your end @mdtauk since the CI is also able to compile the project. Does `git status` show any changes?

mdtauk:
> I'm inclined to say this is something on your end @mdtauk since the CI is also able to compile the project. Does `git status` show any changes?

I did a Git Reset hard and its fine now.

mdtauk:
We have a clipping issue here...
image

Lukiluc29:
Is it normal that my removable disk icon does not appear?
Capture d'écran 2023-08-07 205820

yaira2:
@Lukiluc29 it's a known issue.

0x5bfa:
can you paste the new gif?

yaira2:
> can you paste the new gif?

It's still a work in progress but this is where it's at as of earlier today.
image

d2dyno1:
## The problem

I've managed to find the race condition that was causing the icons not to load (or rarely load) on the sidebar when plugging a USB drive. The race condition happens between `SideBarItem.UpdateIcon()` and `DriveItem.LoadThumbnailAsync()`.

When a USB is plugged in, the `UpdateIcon` method is called first which grabs the `IconSource` from its `Item` property:
image

Peeking into the `DriveItem.IconSource` property, the `Icon` used to set the source is null - this is because the method to load the icon (`DriveItem.LoadThumbnailAsync()`) did not complete before updating the icon:
image

After more investigation, I've discovered that setting the `Icon` property will also call `PropertyChanged` for the "IconSource". That event is supposed to be handled in the `SideBarItem.ItemPropertyChangedHandler()` method which calls the `UpdateIcon()`. This would in turn negate the effects of the race condition:
image
image

The event itself is not hooked properly and the method to do so (`SideBarItem.HookupItenChangeListener()`) is never called **after the sidebar item is added**.

(Also, `HookupItenChangeListener` contains a typo 💀)

## The solution
The easiest fix would be to call `HookupItenChangeListener()` when a new item is added.
The harder way would be to refactor the code behind 🤷‍♂️ 

@yaira2 @chingucoding @mdtauk

chingucoding:
@d2dyno1 Thank you for your help in debugging this but I'm afraid this is not the issue. When running the app with the latest commit (where the typo is also fixed), the handler does work properly. 

The issue is that the Drives have their icon loaded twice with varying results. You can run the app and look for the following in the output pane: "Exited load thumbnail for [your drive name]". The numbers after this indicate how many bytes were loaded and it seems sometimes, we load more or less than the previous time.

d2dyno1:
The race condition is still there. Putting a breakpoint below `Icon ??= await IconData.ToBitmapAsync()` will trigger only after the icon is updated and with no subsequent calls to `UpdateIcon()` method

chingucoding:
Well yes and no. There isn't really a race condition since LoadThumbnailAsync also sets the icon triggering the PropertyChangedListener of the Sidebar item.

The issue is that the icon gets set multiple times and the second time somehow results in an icon we cannot properly render in the sidebar. You can sometimes observe flickering when plugging in a USB drive which is also caused by this.

src/Files.App/Actions/Show/ToggleSidebarAction.cs
yaira2, line 1: Are there plans to add this back? It's not used very often but it can still be useful.

chingucoding: At least the code usage didn't say it was used. We can readd it but we need to do this without a circular reference to SidebarViewModel (which is the reason its being deleted in the first place)

yaira2:
> I've tested the new sidebar, and kudos for the great job! I just have one final point to address. Is it intentional that we can't transfer items from the favorites bar to a disk or cloud drive, or from a disk to a cloud drive, or from a cloud drive to a disk? Because it's a feature that I used very regularly and losing it would be a shame

Hi @Lukiluc29, is this related to the new design or is it just a general question?

yaira2:
We're aiming to have the same functionality as the current sidebar. What are the steps to reproduce this issue?

Lukiluc29:
> We're aiming to have the same functionality as the current sidebar. What are the steps to reproduce this issue?

Navigate to a folder (such as Documents) and, for instance, locate a folder you wish to move to a disk or a cloud drive. Files will notify you that you cannot do this, similarly when you attempt to move a file from a disk to a cloud drive or vice versa.

chingucoding:
Hi @Lukiluc29 , thank you for raising this issue. This is not intentional and I'm working on a fix for this!

chingucoding:
Update, the drag and drop behavior should be fixed now.

0x5bfa:
Are those intentional?

ScrollBar is visible when you cannot scroll
image

When I changed pane length to minimum, icon is not centered.
image

mdtauk:
> When I changed pane length to minimum, icon is not centered. image
> 
There is space reserved for the scrollbar and the resizer, so it won't be possible to centre the items in the collapsed sidebar.

> Pinned icon is not displayed image
> 
This is coming at some point, it was not a priority for the initial PR

> CornerRadius is not 0 when pane is overlayed. image
>
This is intentional, when it overlays it should have a rounded right edge

0x5bfa:
And overlay background shouldn't be solid?

image

> There is space reserved for the scrollbar and the resizer, so it won't be possible to centre the items in the collapsed sidebar.

We can fix it by changing min width

mdtauk:
> And overlay background shouldn't be solid?
> 
> image
> 

We went with an Acrylic overlay, to make it feel less heavy and imposing as it flies in.

chingucoding:
> ScrollBar is visible when you cannot scroll

This should be fixed now (@yaira2 FYI)

src/Files.App/Data/Items/DriveItem.cs
```diff
@@ -152,6 +177,28 @@ public bool ShowStorageSense
...
+		public object ToolTip
+		{
+			get => toolTip;
+			set
+			{
+				SetProperty(ref toolTip, value);
+			}
+		}
+
+		public bool IsExpanded { get => false; set { } }
```
yaira2, line 192: Is it worth adding a comment explaining what IsExpanded does?

src/Files.App/UserControls/Sidebar/SidebarItemDropPosition.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) 2023 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.App.UserControls.Sidebar
+{
+	public enum SidebarItemDropPosition
```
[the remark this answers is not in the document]
chingucoding: In the future, we can implement reorder using drag or implement pinning by dragging the item at the right position. I can remove this though for now if you feel like that's better for now.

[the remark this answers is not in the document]
yaira2: It's fine, I was just wondering what it was for.

src/Files.App/Data/Items/DriveItem.cs
chingucoding: Very good point, I went ahead and just wrote some more documentation in general (including for this property).

yaira2: Thank you

Lukiluc29:
I wanted to inquire if it's normal that the tooltip displaying the size of the Recycle Bin has disappeared.

chingucoding:
This is indeed not intentional (it only is for drives/cloud drives), I've fixed this now. Thank you for pointing that out @Lukiluc29 !

Lukiluc29:
The sidebar looks pretty good to me!

Lukiluc29:
Since you have disabled cloud disk tooltips because they are erroneous, this issue #11326 can be closed

chingucoding:
@yaira2 Would we like to reimplement the tooltips for disks/cloud drives to the way they were before or can #11326 be truly closed then?

yaira2:
@chingucoding I'm fine without reimplementing the rich tooltips, the plan is to display this info in the details pane instead.

mdtauk:
> @chingucoding I'm fine without reimplementing the rich tooltips, the plan is to display this info in the details pane instead.

What was the issue with displaying the tooltip?  Performance? Reliability of the data?

chingucoding:
> What was the issue with displaying the tooltip? Performance? Reliability of the data?

Mostly the issue how to separate the SidebarView from the Items and then somehow generating those tooltips based on the items.

Lukiluc29:
I'm conducting some tests to ensure there are no new sidebar-related bugs, and I've noticed that when we rearrange favorites and the "Favorites" section collapses, it displays as it used to, with a favorites icon instead of just using text. However, it reverts back to the correct current and rearranged display afterward.
Capture d'écran 2023-08-12 083519

chingucoding:
> I'm conducting some tests to ensure there are no new sidebar-related bugs, and I've noticed that when we rearrange favorites and the "Favorites" section collapses, it displays as it used to, with a favorites icon instead of just using text. However, it reverts back to the correct current and rearranged display afterward.

I'm inclined to say this is somewhat intentional and correct (at least from a SidebarView perspective). When we reorder the favorites, we essentially remove all favorites from the sidebar and than readd them. While there are no children, the favorite item acts as a top level element. @yaira2 @mdtauk What are your thoughts on this behavior?

mdtauk:
Ideally the item would know it's a group header, and will contain items when they are added to the group.

chingucoding:
But what if there are no child items, do we show the header and nothing else? Doesn't that invite users to click on it despite the section itself not making much sense?

mdtauk:
> But what if there are no child items, do we show the header and nothing else? Doesn't that invite users to click on it despite the section itself not making much sense?

- We could hide the chevron when there are no children
- Hide the group entirely
- Add a label with text like "Pin folders here for quick access" or "No network locations are available" etc

chingucoding:
> We could hide the chevron when there are no children

I went with this solution for now.

mdtauk:
> > We could hide the chevron when there are no children
> 
> I went with this solution for now.

What appears in the flyout, for an empty group when the sidebar is in the Collapsed mode?

chingucoding:
> What appears in the flyout, for an empty group when the sidebar is in the Collapsed mode?

I've updated the behavior to don't do anything, before that it would have shown an empty flyout. What do you think is better for users?

chingucoding:
And also, in my opinion the correct behavior for reordering would be to not actually remove all items and readd them but just insert the items in the correct position so we wouldn't walk into that situation at all.

mdtauk:
> > What appears in the flyout, for an empty group when the sidebar is in the Collapsed mode?
> 
> I've updated the behavior to don't do anything, before that it would have shown an empty flyout. What do you think is better for users?

If there is a label for the empty group state, this label can be displayed in the flyout.

yaira2 (approved):
LGTM

yaira2:
> Since you have disabled cloud disk tooltips because they are erroneous, this issue #11326 can be closed

@Lukiluc29 removing the tooltip doesn't fix the underlying issue because the data is still wrong.

PR #13058, opened 2023-07-24
Fix: Fixed an issue where an invalid Git Remote would cause a crash

### LINKED ISSUE #13048: Bug: Invalid Git Remote crashes app
### Description

When a Git repository contains a remote that no longer is valid (e.g. repo deleted), the app crashes. Worst case is that the app tries to open the same folder on startup and crashes again.

Exception:

Screenshot of debugger inspecting uncaught exception "... is not a valid remote"

### Steps To Reproduce

1. Open a repository with invalid Git Remotes

### Requirements

Should not crash

### Files Version

Commit: 75e41fafe7099e089a30f5b325b617b09aab9d0e

### Windows Version

22621.1992

### Log File

I can send them if requested.

ferrariofilippo:
**Resolved / Related Issues**
   Closes #13048 

**Possible Improvements**
At the moment this PRs only prevent the crash.
- Should we show a warning when there's an invalid remote?
- If the invalid remote is the current head, what should we do? Checkout to the first valid branch? Do not show any git info at all? (If the head is invalid we can't get its name and all the other infos [we do not know which branch is the head])

**Validation**
   1. Open app 
   2. Invalidate a git repo (I changed the last component of `ref: refs/...` to a `.` of the file `.git/HEAD`  )
   3. Open that git repo


ferrariofilippo:
> Possible Improvements
At the moment this PRs only prevent the crash.

> Should we show a warning when there's an invalid remote?
If the invalid remote is the current head, what should we do? Checkout to the first valid branch? Do not show any git info at all? (If the head is invalid we can't get its name and all the other infos [we do not know which branch is the head])

@yaira2 what do you think?

yaira2:
If the remote is invalid, do the local branches still work?

ferrariofilippo:
> If the remote is invalid, do the local branches still work?

At the moment I can only reproduce the bug when the `HEAD` branch is invalid. In this case you can't do anything: you have no active branch (if the branch is invalid you can't check if it's the `HEAD`) and so `git status bar` is not shown

ferrariofilippo:
@yaira2 I did some more testing:
- if the head is invalid we can't do anything (except showing a warning) [see above]
- if there are issues with a remote, after handling the exception, everything works well local-side (you can switch branch and the repo seems fine, but you can't get `ahead by/behind by`). I'll check what happens when you try to fetch/push/pull

ferrariofilippo:
> * if the head is invalid we can't do anything (except showing a warning) [see above]
> * if there are issues with a remote, after handling the exception, everything works well local-side (you can switch branch and the repo seems fine, but you can't get `ahead by/behind by`).
>   I'll check what happens when you try to fetch/push/pull ->It fails and shows `git pull/push timeout error`, we should add a specific warning

Do you have any better ideas?

yaira2:
Let's go with handling the exception without displaying a warning. If it becomes confusing in the future, we can add a warning.

yaira2:
@ferrariofilippo can you fix the merge conflict?

yaira2 (dismissed):

0x5bfa (approved):
LGTM, in terms of the codebase. Thank you!

PR #13111, opened 2023-07-31
Fix: Fixed an issue where Git columns were incorrectly shown in search results

### LINKED ISSUE #12927: Bug: Display Issue with Git Folder Search in Detail View
### Description

I have noticed an issue when performing a search within a Git folder in detail view. The Git columns remain visible, and the search columns are added after the Git columns instead of having only the search columns. I'm not sure if this is intentional, but I find it a bit inconvenient. I believe this might not be intentional because the file information is aligned as if it's not a search within a Git folder.

### Steps To Reproduce

1. Open a Git Folder
2. Launch a search

### Requirements

Identify the cause of the issue and determine if it is a bug or an intentional design choice.

### Files Version

2.5.14.0

### Windows Version

22H2 22621.1848

### Log File

Capture d'écran 2023-07-10 163405

QuaintMako:
**Resolved / Related Issues**
   - Closes #12927 
   - **I hid the status bar as well for consistency reasons.**

**Validation**
- [x] Did you implement any design changes to an existing feature? Yes, hidding the Git status bar as well.
   1. Open app ...
   2. Click settings button ...

**Screenshots**

GitInfoHidden

### DIFF
--- a/src/Files.App/Data/Models/DirectoryPropertiesViewModel.cs
+++ b/src/Files.App/Data/Models/DirectoryPropertiesViewModel.cs
@@ -7,6 +7,8 @@ namespace Files.App.Data.Models
 {
 	public class DirectoryPropertiesViewModel : ObservableObject
 	{
+		private IContentPageContext ContentPageContext { get; } = Ioc.Default.GetRequiredService<IContentPageContext>();
+
 		// The first branch will always be the active one.
 		public const int ACTIVE_BRANCH_INDEX = 0;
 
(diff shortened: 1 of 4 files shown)


src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -205,7 +205,6 @@ private void UpdateSortIndicator()
 
 		private void FilesystemViewModel_PageTypeUpdated(object? sender, PageTypeUpdatedEventArgs e)
 		{
-			// Show original path and date deleted columns in Recycle Bin
```
QuaintMako, line 208: Removed those comments that are just repeating the few lines below. They are not needed.

QuaintMako (commented):
Writing it thrice to be sure it won't be missed: the initial issue does not mention hiding the status bar as well. Yet if we want to be coherent, I believe we should mask it in the search result bar.

src/Files.App/Data/Models/DirectoryPropertiesViewModel.cs
```diff
@@ -86,13 +88,13 @@ public string ExtendedStatusInfo
...
 		{
-			NewBranchCommand = new AsyncRelayCommand(() 
+			NewBranchCommand = new AsyncRelayCommand(()
 				=> GitHelpers.CreateNewBranch(_gitRepositoryPath!, _localBranches[ACTIVE_BRANCH_INDEX]));
 		}
 
 		public void UpdateGitInfo(bool isGitRepository, string? repositoryPath, BranchItem[] branches)
 		{
-			GitBranchDisplayName = isGitRepository && branches.Any()
+			GitBranchDisplayName = isGitRepository && branches.Any() && !(ContentPageContext.ShellPage!.InstanceViewModel.IsPageTypeSearchResults)
```
yaira2, line 97: Should this be broken onto multiple lines

PR #13115, opened 2023-08-01
Code Quality: Pure logfiles to only keep the last 100 lines

### LINKED ISSUE #12182: Feature: Purging logs after a given size

While filling what we may expect of it, the logs saving management we have in Files is currently very limited. All the logs are save within the same file which is never cleared. It can be detrimental to debug when the logs are very populated.

For example, here is an extract of my log file, where four sessions are logged:

```
2023-04-15 11:47:54.9380|Information|Running on ARM: False
2023-04-23 21:44:33.6420|Information|App launched. Launch args type: LaunchActivatedEventArgs
2023-04-23 23:46:49.9762|Information|App launched. Launch args type: LaunchActivatedEventArgs
2023-04-24 10:05:09.8545|Information|App launched. Launch args type: LaunchActivatedEventArgs
```

We should implement a system that cleanse the log files once in a while.

### Requirements

- When reaching a given size or number of lines in the log file, it should be cleared.

### Files Version

Version 2.4.68 Dev

### Windows Version

W11 21H2 22000.1098

### Comments

_No response_

QuaintMako:
**Resolved / Related Issues**
   Closes #12182

**Validation**
   1. Made sure the process works when no logfile exists.


QuaintMako:
I'm opening this as a draft to have a bit of a discussion around the implementation. I've got several points: 

1. The current implementation is not the most efficient that there is. It's ready the collection of lines twice. It could done in one go, but I decided to go with this implementation for clarity's sake.

2. The purge is done currently at the start of the project, increasing ever so slightly the time of start up. And it's not helped the first point above. Moving the purge at the closing of the app might be an idea? Reactivity is not a big deal at that point.

3. The call to purge the logs is done at the moment during the constructor. I do not like it, but either I'm missing something big, or the actual implementation of the logger is really difficult to access. We could fix that by creating a new `IFilesLogger` that exposes the purge method, and implement the `ILogger` interface.

Any thought?

yaira2:
> The purge is done currently at the start of the project, increasing ever so slightly the time of start up. And it's not helped the first point above. Moving the purge at the closing of the app might be an idea? Reactivity is not a big deal at that point.

Regarding this point, I don't think it's a problem if the task is delayed.

QuaintMako:
> Regarding this point, I don't think it's a problem if the task is delayed.

This would require to have access to the `PurgeLog` method outside of the class, which is currently a problem. See point 3 above.

QuaintMako:
I ended up finding a way less complicated in the end to make the purge process `async`.

As a sidenote, for the culture, expanding ILogger would have been pretty hard because of the pretty complex process of generic logging we're using.

yaira2 (approved):
LGTM

PR #13119, opened 2023-08-02
Feature: Automatically update items when the date changes

### LINKED ISSUE #13109: Feature: Automatically update date differences

Currently, when a date is displayed by difference, its value is not updated unless the page is refreshed. Thus, creating a new file multiple times will all show "Now" or "1 second ago" even though they actually have different dates. If the date is displayed by difference, the value should be automatically updated.

### Requirements

If the date is displayed by difference, the value should be automatically updated.

### Files Version

2.5.21.0

### Windows Version

10.0.19045.3208

### Comments

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #13109 

**Validation**
   Following date displays will automatically update:
   - Date columns in the Details layout
   - Tooltips in the Tile, Grid and Column layout (each time a tooltip pops up)
   - Date displays in the General properties page
   - Date displays in the preview pane
   - Date displays in the Details properties page

### DIFF
--- a/src/Files.App/Views/Properties/GeneralPage.xaml.cs
+++ b/src/Files.App/Views/Properties/GeneralPage.xaml.cs
@@ -1,21 +1,13 @@
 // Copyright (c) 2023 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using CommunityToolkit.WinUI;
-using Files.App.Data.Items;
-using Files.App.Extensions;
-using Files.App.Utils;
-using Files.App.Helpers;
-using Files.App.Utils.Shell;
 using Files.App.ViewModels.Properties;
-using Files.Shared;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Input;
-using System.Collections.Generic;
 using System.IO;
 using System.Text.RegularExpressions;
-using System.Threading.Tasks;
 using Windows.Storage;
+using Microsoft.UI.Dispatching;
 
 namespace Files.App.Views.Properties
 {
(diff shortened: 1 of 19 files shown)


src/Files.App/Views/Properties/GeneralPage.xaml.cs
```diff
@@ -158,6 +165,7 @@ async Task<bool> SaveBaseAsync(ListedItem item)
 
 		public override void Dispose()
 		{
+			_updateDateDisplayTimer.Stop();
 		}
```
ferrariofilippo, line 169: I think we should also unsubscribe the `Tick` event here

ferrariofilippo (commented):
I get this exception when I try to run the code

image

src/Files.App/Views/Properties/GeneralPage.xaml.cs
hishitetsu: I believe unsubscribing is not needed because both the timer and the event are private members of the page and they have the same lifecycle.

hishitetsu:
> > I get this exception when I try to run the code
> 
> Oops, I don't have the environment to build now, I'll fix it later.

Fixed.

ferrariofilippo:
>   * Date displays in the preview pane
>   * Date displays in the Details properties page

Are these supposed to refresh automatically or should they update whenever the content is loaded?

hishitetsu:
> > * Date displays in the preview pane
> > * Date displays in the Details properties page
> 
> Are these supposed to refresh automatically or should they update whenever the content is loaded?

Now they should be refreshed automatically.

ferrariofilippo:
> Now they should be refreshed automatically.

It doesn't seem to work then

hishitetsu:
> It doesn't seem to work then

Did you pull the latest commit?

ferrariofilippo:
> Did you pull the latest commit?

My bad, I was on `Use dispatcherQueue`

ferrariofilippo (approved):
LGTM

PR #13120, opened 2023-08-02
Feature: Display keyboard shortcuts in the right click menu

### LINKED ISSUE #13055: Feature: Display keyboard shortcuts in the right click menu

Over the years I got used to do right-click>W>F to create a new folder for instance.

Win11 dropped those kb shortcuts (thanks MSFT), so the goal would be to equate Win10's.

### Requirements

- This proposal will accomplish having the same KB shortcuts in the cintextual menu as Win10 does

### Files Version

2.5.21.0

### Windows Version

10.0.22621.1926

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**
   Closes #13055 

When there's more than one shortcut, should the label display all of them?
There's a bug involving `Enter` and I'm working on it (if you hit `Enter`, which is the accelerator for `Open`, it will cut the item, since that's the first button in the menu)

**Validation**
   1. Open app
   2. Right click on a file / empty space

**Screenshots (optional)**
image

### DIFF
--- a/src/Files.App/Data/Models/ContextMenuFlyoutItemViewModelBuilder.cs
+++ b/src/Files.App/Data/Models/ContextMenuFlyoutItemViewModelBuilder.cs
@@ -1,7 +1,7 @@
 // Copyright (c) 2023 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using Files.App.Data.Commands;
+using Windows.System;
 
 namespace Files.App.Data.Models
 {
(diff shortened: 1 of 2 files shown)


ferrariofilippo:
> There's a bug involving Enter and I'm working on it (if you hit Enter, which is the accelerator for Open, it will cut the item, since that's the first button in the menu)

@yaira2  I thought that if we allow `Enter` accelerator in the context flyout, users won't be able to work with it using only the keyboard. You need to hit `Enter` to select an action, but doing so would trigger the shortcut. 
I think we should prevent this. We may remap `Enter` shortcuts to something else (like `Ctrl + Enter` or `Shift + Enter`) only in the context menu

ferrariofilippo:
> @ferrariofilippo is it possible to display the shortcut without adding the accelerators to the context menu? It might be confusing if we used custom shortcuts only in the context menu.

The label would be visible, but the shortcut would not work.
I don't want to use custom shortcuts in the context menu. I think we should change only `Enter` shortcut to something else, since it would make the context menu inaccessible to keyboard users (all the other shortcuts won't be affected [`Ctrl + C`, `Ctrl + V`, `Ctrl + Enter`, ...]).

yaira2:
> The label would be visible, but the shortcut would not work.

Wouldn't the shortcut work from the rich commands?

yaira2:
Regarding enter, we can leave that blank 👍

yaira2:
That's good to know, in that case we can just leave `enter` blank and fill out the other ones.

ferrariofilippo:
What about this?
image

Should we show all the available shortcuts or only the first one?

yaira2:
Just the first one

yaira2 (approved):
LGTM

PR #13134, opened 2023-08-05
Fix: Fixed crash that would occur when using drag & drop as admin

### LINKED ISSUE #12390: Bug: Disable drag & drop as admin to prevent crash
### Description

everytime i try to drag a file with the file running with admin rights it crashes

### Steps To Reproduce

1. open Files as admin
2. drag and drop a file
3. app crashes

### Requirements

- disable drag & drop as admin to prevent crash

### Files Version

2.4.70.0

### Windows Version

Microsoft Windows 11 Pro Version	10.0.22621 Build 22621

### Log file

[debug.log](https://github.com/files-community/Files/files/11508888/debug.log)

QuaintMako:
**Resolved / Related Issues**
   Closes #12390

**Validation**
   1. Open app ...
   2. Click settings button ...


src/Files.App/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
```diff
@@ -63,7 +63,7 @@ private void HorizontalTabView_TabItemsChanged(TabView sender, Windows.Foundatio
 		private async void TabViewItem_Drop(object sender, DragEventArgs e)
 		{
 			await ((sender as TabViewItem).DataContext as TabItem).Control.TabItemContent.TabItemDrop(sender, e);
-			HorizontalTabView.CanReorderTabs = true;
+			HorizontalTabView.CanReorderTabs = true && !ElevationHelpers.IsAppRunAsAdmin();
```
ferrariofilippo, line 66:
```suggestion
			HorizontalTabView.CanReorderTabs = !ElevationHelpers.IsAppRunAsAdmin();
```

I think we can remove `true &&`, it's redundant. But we might consider leaving it there for when we will re-enable drag & drop as admin

ferrariofilippo (approved):
LGTM

src/Files.App/UserControls/MultitaskingControl/HorizontalMultitaskingControl.xaml.cs
QuaintMako: You're quite right, it's redundant!

ferrariofilippo: There are other occurrences in the same file, can you fix them too

yaira2:
Should we display an modal to inform the user why drag and drop is disabled, or is that not posible?

QuaintMako:
> Should we display a modal to inform the user why drag and drop is disabled, or is that not possible for the same reason drag & drop doesn't work?

We could display a modal. Should it be displayed once at the start of the app as a reminder, or everytime a drag is attempted?

yaira2:
My suggestion is to show it every time but with an option to "don't show again".

yaira2 (approved):
LGTM

QuaintMako:
Sorry for the lack of answer, I am currently unable to dedicate time to Files. Thanks for having see it through, I'll handle the dialog box as soon as I can.

src/Files.App/Helpers/Environment/ElevationHelpers.cs
```diff
@@ -19,5 +18,11 @@ public static bool IsElevationRequired(string filePath)
 
 			return _IsElevationRequired(filePath);
 		}
+
+		public static bool IsAppRunAsAdmin()
+		{
+			using WindowsIdentity identity = WindowsIdentity.GetCurrent();
+			return new WindowsPrincipal(identity).IsInRole(WindowsBuiltInRole.Administrator);
```
yaira2, line 25: It turns out that it's not specific to admin, the crash can happen even without admin if the app is running with high trust #13394.

PR #13157, opened 2023-08-08
Fix: Fixed an issue where clicking an empty space would scroll to the top of the file list

ferrariofilippo:
**Resolved / Related Issues**
- [x] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
   Closes #12523 

**Validation**
How did you test these changes?
- [x] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [ ] Did you implement any design changes to an existing feature?
   - [ ] Was this change approved?
- [x] Are there any other steps that were used to validate these changes?
   1. Open a folder using `DetailsLayout`
   2. Scroll down
   3. Click on an empty space

PR #13158, opened 2023-08-08
Code Quality: Fixed issue with clicking search box

ferrariofilippo:
**Resolved / Related Issues**
   Related #13157

**Validation**
   1. Open a folder using `Details Layout`
   2. Click on the path bar or searchbox

### DIFF
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -880,7 +880,7 @@ namespace Files.App.Views.LayoutModes
 		private void FileList_LosingFocus(UIElement sender, LosingFocusEventArgs args)
 		{
 			// Fixes an issue where clicking an empty space would scroll to the top of the file list
-			if (args.NewFocusedElement is ListView)
+			if (args.NewFocusedElement == FileList)
 				args.TryCancel();
 		}
 	}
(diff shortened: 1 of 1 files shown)


ferrariofilippo:
@yaira2 @hishitetsu this should prevent the bug

hishitetsu:
When I single click once on an empty area, it works fine. But when I single click two or more times on an empty area, I can't type in the path bar.

ferrariofilippo:
> When I single click once on an empty area, it works fine. But when I single click two or more times on an empty area, I can't type in the path bar.

I see it, maybe it's sufficient to check if the next item to be focused is the listview, I'm trying this way

hishitetsu:
> (To be sure, if you click twice on the path bar you can use it, right?)

Yes, so this may be a minor issue.

ferrariofilippo:
> maybe it's sufficient to check if the next item to be focused is the listview, I'm trying this way

This works!
Still, when refocusing the listview, it will scroll to the top

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -427,7 +427,9 @@ private async void FileList_ItemTapped(object sender, TappedRoutedEventArgs e)
 			var shiftPressed = InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Shift).HasFlag(CoreVirtualKeyStates.Down);
 			var item = (e.OriginalSource as FrameworkElement)?.DataContext as ListedItem;
 			if (item is null)
+			{
 				return;
+			}
```
hishitetsu, line 432: Brackets are no more needed.

hishitetsu (approved):
LGTM. It certainly doesn't work well with refocusing, but it is still better than as is.

ferrariofilippo:
> It certainly doesn't work well with refocusing, but it is still better than as is.

Why might open an issue to track that

hishitetsu (approved):
LGTM

PR #13142, opened 2023-08-06
Feature: Show BitLocker options in main menu for drives

ferrariofilippo:
**Resolved / Related Issues**
- [x] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
   Closes #11303 

**Moved options**
- Turn on BitLocker
- Manage BitLocker

**Validation**
How did you test these changes?
- [x] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [x] Did you implement any design changes to an existing feature?
   - [x] Was this change approved?
- [x] Are there any other steps that were used to validate these changes?
   1. Open app
   2. Right click on a local drive (Home, Sidebar or root folder)
   3. See that BitLocker options are visible in the main menu
   4. Right click a cloud drive (e.g. Google Drive)
   5. See that BitLocker options are not visible

PR #13173, opened 2023-08-10
Feature: Added support for formatting drives from the properties window

### LINKED ISSUE #11607: Bug: Format drive option is missing

It is not possible to format a drive with Files.

### Requirements

 This option should be added to the drive context menus and in the properties popup.
Just open the Windows formatting interface. Possibly preceded by a warning.

### Files Version

2.4.49.0

### Windows Version

10.0.19045.2604

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**
   Closes #11607
   Related #13142 

**Details**
As pointed out in #11607, `Format...` option is already visible in the main context menu of drives.
I added a `Format` button to the properties window of drives.
After #13142, `Turn on BitLocker` was visible in the main menu only when the drive required admin privileges; this PR fixes that bug

**Validation**
   1. Open properties of a drive
   2. Click on `Format...`
   3. See that the format dialog pops up
   4. Right click on a USB stick and see that `Turn on BitLocker` is in the main menu

**Screenshots (optional)**
Screenshot 2023-08-10 110904


yaira2:
Do we really want to add this to properties? I'm not against this option, but it's worth considering how useful it'll be when it can already be accessed from the context menu.

ferrariofilippo:
Honestly, I think it will be a niche feature: I don't think people format their drives frequently. I'd use this button whenever I want to check something in the properties of a drive before formatting it. If I am using the properties window, it's easier to use that button instead of going back to Files, right-clicking on the drive and then choosing formatting. 
It might be useful for touch devices, where context menu is harder to use.
If you think we shouldn't add this button, I'll leave only the code to fix the bug with BitLocker

yaira2:
>  I'd use this button whenever I want to check something in the properties of a drive before formatting it. If I am using the properties window

Makes sense 👍

src/Files.App/Views/Properties/GeneralPage.xaml
```diff
@@ -564,6 +565,23 @@
 					Click="DiskCleanupButton_Click"
 					Content="{helpers:ResourceString Name=DiskCleanup}" />
 
+				<!--  Format Disk  -->
+				<TextBlock
```
yaira2, line 569: Can you wrap this in a settings control?

ferrariofilippo: image

What icons should we use?
One might be this one:
image

yaira2: It's probably best to not use any icon. As a side note, if you use the button style, the whole card will act as a button.

ferrariofilippo: image

yaira2: Nice! Can you move them out of the container so that they are by themselves?

yaira2: Can you also add an icon like this?
image

ferrariofilippo: image

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3438,6 +3438,9 @@
   <data name="CopyFailureFileToLarge" xml:space="preserve">
     <value>The following items are too large to be copied to this drive</value>
   </data>
+  <data name="FormatYourDrive" xml:space="preserve">
+    <value>Format your drive</value>
```
yaira2, line 3442:
```suggestion
    <value>Format drive</value>
```
What do you think?

src/Files.App/Views/Properties/GeneralPage.xaml
```diff
@@ -538,35 +537,38 @@
...
-					Text="{helpers:ResourceString Name=CleanupDrive}" />
-
-				<Button
-					x:Name="DiskCleanupButton"
-					Grid.Row="4"
-					Grid.Column="2"
-					Grid.ColumnSpan="2"
-					HorizontalAlignment="Right"
-					Click="DiskCleanupButton_Click"
-					Content="{helpers:ResourceString Name=DiskCleanup}" />
```
yaira2, line 566: Is this string still being used?

ferrariofilippo: It's not, I'll remove it

src/Files.App/Strings/en-US/Resources.resw
ferrariofilippo: I remember someone settled a debate about whether Files should be more friendly or neutral, but I forget how it ended. If we want the latter, we should remove `your`

yaira2: I think it's best to decide each case separately. In this scenario, it can be someone else's drive so "your" might be out of place.

src/Files.App/Views/Properties/GeneralPage.xaml
yaira2: Thanks

yaira2 (approved):
LGTM

PR #13207, opened 2023-08-16
Fix: Fixed taskbar behavior in full screen mode

### LINKED ISSUE #10767: Bug: Taskbar doesn't un-hide when Files is full screened/maximized
### Description

My taskbar autohides. When every other app is open, if I hover to the bottom of my screen, the taskbar will show up. But this is not the case when this app is open. This doesn't let the taskbar show up. I have tried restarting the app, restarting the PC and even reinstalling the app. Yet, always, this app prevents the taskbar from showing up.

### Steps To Reproduce

1. Enable the option `Automatically hide the taskbar`
2. Open files app in Maximized Window and let taskbar hide
3. Hover to the bottom of the screen and try to open the taskbar (it won't show up)
4. Open any other app as maximized window and hover to the bottom of the screen and now, the taskbar will show up

For now, the only ways to show the taskbar is to press the windows key or close/minimize/restore the Files app

### Steps to fix

Reveal the taskbar on hover (reaching the bottom of the screen) when taskbar is set to autohide

### Files Version

Version: 2.4.2.0 OS Version: 10.0.22621.963

### Windows Version

Edition	Windows 11 Pro 
Version	22H2 
OS build	22621.963 
Experience	Windows Feature Experience Pack 1000.22638.1000.0

### Relevant Assets/Logs

None

yaira2:
**Resolved / Related Issues**
   Closes #10767...

**Validation**
   1. Toggle full screen mode
   2. Confirmed taskbar is show when hovering the cursor at the bottom of the screen 

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -28,9 +28,6 @@ namespace Files.App
 
 		public IntPtr WindowHandle { get; }
 
-		[DllImport("user32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
-		static extern bool SetPropW(IntPtr hWnd, string lpString, IntPtr hData);
-
 		private MainWindow()
 		{
 			ApplicationService = new ApplicationService();
@@ -59,7 +56,7 @@ namespace Files.App
 
 			// Workaround for full screen window messing up the taskbar
 			// https://github.com/microsoft/microsoft-ui-xaml/issues/8431
-			SetPropW(WindowHandle, "NonRudeHWND", new IntPtr(1));
+			InteropHelpers.SetPropW(WindowHandle, "NonRudeHWND", new IntPtr(1));
 		}
 
 		public void ShowSplashScreen()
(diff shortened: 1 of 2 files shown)


src/Files.App/MainWindow.xaml.cs
```diff
@@ -27,6 +28,9 @@ public sealed partial class MainWindow : WindowEx
 
 		public IntPtr WindowHandle { get; }
 
+		[DllImport("user32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+		static extern bool SetPropW(IntPtr hWnd, string lpString, IntPtr hData);
```
d2dyno1, line 32: Could you use Vanara's User32 class instead?

yaira2: Is there a benefit to that? I though adding more references to Vanara would make it harder to enable trimming.

d2dyno1: You've got a point. Could you move it then to `InteropHelpers`?

yaira2: Good suggestion, I addressed it in the last commit.

d2dyno1 (approved):

PR #13211, opened 2023-08-18
Feature: Lazy load Git properties for better performance

### LINKED ISSUE #13182: Bug: Hidden Git columns are still being calculated
### Description

The data for the Git columns is calculated even when the columns are hidden. This leads to unnecessary resource consumption and should be an easy way to improve performance in Git repos.

### Steps To Reproduce

1. Hide all the Git columns in the Details layout
2. Observe that the memory isn't freed up

### Requirements

- only load Git data for visible columns
- free up data if a column is hidden

### Files Version

v2.5.21

### Windows Version

Windows 11

### Log File

NA

hishitetsu:
**Resolved / Related Issues**
   Closes #13182 

**Validation**
   Git properties will not be loaded in the following cases.
      * The layout is other than the details layout.
      * All Git columns are hidden in the details layout.
   When the above cases are no longer met, Git properties loading will begin.

### DIFF
--- a/src/Files.App/Data/Items/ListedItem.cs
+++ b/src/Files.App/Data/Items/ListedItem.cs
@@ -569,6 +569,13 @@ namespace Files.App.Utils
 
 	public class GitItem : ListedItem
 	{
+		private volatile int gitPropertiesInitialized = 0;
+		public bool GitPropertiesInitialized
+		{
+			get => gitPropertiesInitialized == 1;
+			set => Interlocked.Exchange(ref gitPropertiesInitialized, value ? 1 : 0);
+		}
+
 		private Style? _UnmergedGitStatusIcon;
 		public Style? UnmergedGitStatusIcon
 		{
--- a/src/Files.App/Data/Items/ListedItem.cs
+++ b/src/Files.App/Data/Items/ListedItem.cs
@@ -569,11 +569,18 @@ namespace Files.App.Utils
 
 	public class GitItem : ListedItem
 	{
-		private volatile int gitPropertiesInitialized = 0;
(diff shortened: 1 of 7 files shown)


hishitetsu:
I think it is more efficient to retrieve all Git properties at the point of showing one Git property rather than retrieving a value each time, since accessing the Git repository has great impact on performance.

yaira2:
By default we only show two of the Git columns, I guess with this change we should hide all of them for better performance.

hishitetsu:
I see, it may take some time, but I'll see if there is a better way.

hishitetsu:
I did a little research and found that Git properties are retrieved together as commit information except for status, so I'm afraid that not initializing properties that are not displayed will not improve performance.

image

yaira2:
How about hiding the columns by default (except for status)?

hishitetsu:
> How about hiding the columns by default (except for status)?

Performance can be improved if commit information is not retrieved. However, if status and commit information are retrieved at different times, the total processing time will increase because the repository is accessed twice.

hishitetsu:
@yaira2 status and commit information can now be loaded at different times. Also, the default git column to display is now status only.

yaira2 (approved):
LGTM

hishitetsu:
I don't know, but if you fix the layout to something other than the details layout and there are still performance problems, there might be other causes. Git properties shouldn't be loaded other than the details layout.

yaira2:
I'm having a tough time reproducing the issue consistently, interestingly it's mainly happening in the same folder but even that doesn't repro consistently.

PR #13230, opened 2023-08-21
Code Quality: Added decorators property to the sidebar

chingucoding:
**Resolved / Related Issues**
   Closes #issue...

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.


0x5bfa (commented):
Can you move ResourceDictionaries/SideBarResources.xaml to UserControls/SideBarControls.xaml?
Can you remove UserControls/Sidebar folder, which is not used anymore (that folder seems to be included in your initial PR incidentally)?

src/Files.App/Data/Items/LocationItem.cs
```diff
@@ -96,6 +97,23 @@ public virtual object ToolTip
...
 		}
 
+		private FrameworkElement? itemDecorator;
+		public FrameworkElement? ItemDecorator
+		{
+			get
+			{
+				if (Section == SectionType.Favorites)
+				{
+					itemDecorator ??= new OpacityIcon()
```
yaira2, line 107: I think we can use PathIcon here since it's only a single path (OpacityIcon is used when there are multiple paths and different colors for each path).

Lukiluc29:
I don't really like having the eject button on the sidebar because when you have a lot of removable drives, it can quickly clutter the sidebar. So, I suggest implementing a button in the settings to enable or disable this functionality.

src/Files.App/Data/Items/LocationItem.cs
yaira2: I had to use opacity icon for the eject icon in the end (it's made up of multiple paths) so let's keep it as is for consistency.

yaira2 (approved):
LGTM

Issue #8207, opened 2022-02-07
Add option to show popup progress bar doing file transfer


Thorium0:
### What's the Problem?

If you close the files window during a large file transfer, then it's impossible to see the progress if even if you reopen it.

### Solution/Idea

Add an option to show the default windows file transfer popup, instead of adding it to the queue window in the top right.

### Alternatives

Another solution would also be to make file transfer queues global between all files windows, instead of each window having their own queue.

### Files Version

2.1.5.0

### Windows Version

21H2

gave92:
I'll check if you can show the explorer progress dialog without stopping the operation. If not, would it be ok to show progress in the notification center?

Thorium0:
A notification would work, but it would be problematic if you dismissed the notification or simply don't use the notification center. A custom progress dialog would definitely be preferable.

ELowry:
Wouldn't it be possible to push Files to the system tray until the file transfer is finished? Possibly with some updating percentage bar under the icon, or a hover-title with the percentage as text?

Then clicking the taskbar icon would re-open the files UI with the file transfer popup open.

And right-click options could include `Open Files` and `Exit and Cancel Transfer`.

mdtauk:
Files should have background threads for tasks run when the app window was open, so background tasks can continue when the window is closed, but the app is continuing to run in the background.

For as long as the status bar icons still exist in Windows, the app icon should appear there while the app is open in the background.

As for what can be done with the app in that tray icon state, thats a different discussion.  But when tasks complete, Files should use the windows notifications to display any completed or error/problem messages the User will need to know about, so they can re-open the app window and take action.

yaira2:
This will still be needed in case the process is completly closed.

heftymouse:
I feel like it's reasonable to expect the user to keep the process open while doing an operation, now it can get entirely out of the way when you close the window

yaira2:
Displaying it in a flyout might work, but we have to keep in mind the user can hide the icon so we'd want to account for setups like this as well.

SilverGreen93:
I would prefer to have a system tray notification / popup of some kind, just like how KDE Dolphin shows the progress of file operations.
Of course, when reopening the Files main windows, the status center should show the progress there as well.

PR #13236, opened 2023-08-23
Feature: Added an option to keep Files running in the background for quicker startup time

### LINKED ISSUE #12614: Feature: Cache windows for quicker startup

The startup speed is largely limited by WinAppSdk limiations but we can cache windows for a workaround.

### Requirements

- Add experimental setting to enable caching

### Files Version

v2.5.1

### Windows Version

Windows 11 (22H2): 22621

### Comments

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #12614 

**Validation**
   1. Turn on "Leave app running in the background when the window is closed" in Advanced settings
   2. Close and restart the app.
   3. The app will launch quicker.

**Known Issues**
 - [X] ~~When only one tab is opened, the tab is not visually selected.~~
image
 - [X] ~~Closing the main window while a property window is still open freezes the property window until the next time the main window is opened.~~
 - [X] ~~The content flashes while it loads.~~

### DIFF
--- a/src/Files.App/App.xaml.cs
+++ b/src/Files.App/App.xaml.cs
@@ -178,18 +178,22 @@ namespace Files.App
 				FileTagsHelper.UpdateTagsDb();
 			});
 
-			// Check for required updates
-			var updateService = Ioc.Default.GetRequiredService<IUpdateService>();
-			await updateService.CheckForUpdates();
-			await updateService.DownloadMandatoryUpdates();
-			await updateService.CheckAndUpdateFilesLauncherAsync();
-			await updateService.CheckLatestReleaseNotesAsync();
+			await CheckForRequiredUpdates();
 
 			static async Task OptionalTask(Task task, bool condition)
 			{
 				if (condition)
 					await task;
 			}
+		}
+
+		private static async Task CheckForRequiredUpdates()
+		{
+			var updateService = Ioc.Default.GetRequiredService<IUpdateService>();
+			await updateService.CheckForUpdates();
+			await updateService.DownloadMandatoryUpdates();
+			await updateService.CheckAndUpdateFilesLauncherAsync();
+			await updateService.CheckLatestReleaseNotesAsync();
 		}
 
 		/// <summary>
@@ -331,6 +335,9 @@ namespace Files.App
 					Program.Pool.Dispose();
 					MainWindow.Instance.AppWindow.Show();
 					MainWindow.Instance.Activate();
+
+					_ = CheckForRequiredUpdates();
+
 					MainWindow.Instance.EnsureWindowIsInitialized().Navigate(typeof(MainPage), null, new SuppressNavigationTransitionInfo());
 				}
 
--- a/src/Files.App/App.xaml.cs
+++ b/src/Files.App/App.xaml.cs
@@ -306,7 +306,7 @@ namespace Files.App
 				!Process.GetProcessesByName("Files").Any(x => x.Id != Process.GetCurrentProcess().Id))
 			{
 				// Close open content dialogs
-				var contentDialogs = VisualTreeHelper.GetOpenPopups(Window.Current);
+				var contentDialogs = VisualTreeHelper.GetOpenPopups(MainWindow.Instance);
 				foreach (var popup in contentDialogs)
 					if (popup.Child is ContentDialog)
 						((ContentDialog)popup.Child).Hide();
--- a/src/Files.App/App.xaml.cs
+++ b/src/Files.App/App.xaml.cs
@@ -326,10 +326,12 @@ namespace Files.App
 				// Wait for all properties windows to close
 				await FilePropertiesHelpers.WaitClosingAll();
 
+				// Sleep current instance
 				Program.Pool = new(0, 1, "Files-Instance");
 				Thread.Yield();
 				if (Program.Pool.WaitOne())
 				{
+					// Resume the instance
 					Program.Pool.Dispose();
 					MainWindow.Instance.AppWindow.Show();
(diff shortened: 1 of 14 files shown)


hishitetsu:
> Closing the main window while a property window is still open freezes the property window until the next time the main window is opened.

One possible solution is to close all property windows together when closing the main window.

hishitetsu:
All known issues were resolved. Now ready for review.

Lukiluc29:
Hello, I just have one last little thing to ask. Is it normal that when I close "Files" and then reopen it, the status center appears? Because before, whether I checked "Open a new tab" or "Continue where you left off," it wouldn't appear when I had performed an action before closing it.

src/Files.App/Views/Settings/AdvancedPage.xaml
```diff
@@ -98,6 +98,17 @@
 				</ToggleSwitch>
 			</local:SettingsBlockControl>
 
+			<!--  Leave App Running  -->
+			<local:SettingsBlockControl Title="{helpers:ResourceString Name=SettingsLeaveAppRunning}" HorizontalAlignment="Stretch">
```
yaira2, line 102: Should this be under experimental? If you feel it's stable enough, we can keep it here.

src/Files.Core/Services/Settings/IGeneralSettingsService.cs
```diff
@@ -169,6 +169,11 @@ public interface IGeneralSettingsService : IBaseSettingsService, INotifyProperty
 		/// </summary>
 		bool ShowSendToMenu { get; set; }
 
+		/// <summary>
+		/// Gets or sets a value indicating whether or not to leave app running in the background.
+		/// </summary>
+		bool LeaveAppRunning { get; set; }
```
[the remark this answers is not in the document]
hishitetsu: There is no such service. Should I create a new one?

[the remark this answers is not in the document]
yaira2: Ah right, the other settings on this page aren't tracked by Files. I guess we can leave it as it is for now, we can always create a service in the future if we want to add more settings to this page.

hishitetsu:
> Hello, I just have one last little thing to ask. Is it normal that when I close "Files" and then reopen it, the status center appears? Because before, whether I checked "Open a new tab" or "Continue where you left off," it wouldn't appear when I had performed an action before closing it.

It is true that the behavior is different from when the process is terminated without caching, but I think it doesn't matter.

Lukiluc29:
> It is true that the behavior is different from when the process is terminated without caching, but I think it doesn't matter.

Yes, but it's a bit bothersome because when I close Files, I don't necessarily want to see the status center when I reopen Files, especially when I haven't taken any action upon its reopening.

yaira2:
@Lukiluc29 this is the expected behavior because you aren't actually closing Files, you're leaving it open in the background.

Lukiluc29:
> @Lukiluc29 this is the expected behavior because you aren't actually closing Files, you're leaving it open in the background.

Users tend to close applications after using them, and they might find it disruptive to encounter a status center that they didn't initiate themselves. By removing the status center in this scenario, we can enhance the user's sense of control and provide a smoother and more intuitive interaction with the application.

yaira2:
> Users tend to close applications after using them

Users enabling this setting are opting into this behavior, we might change this in the future but as of now the default setting is to not to leave the app running in the background.

src/Files.App/App.xaml.cs
```diff
@@ -300,6 +301,35 @@ private async void Window_Closed(object sender, WindowEventArgs args)
...
+				MainWindow.Instance.AppWindow.Hide();
+				args.Handled = true;
+
+				// Save and close all tabs
+				SaveSessionTabs();
+				MainPageViewModel.AppInstances.ForEach(tabItem => tabItem.Unload());
+				MainPageViewModel.AppInstances.Clear();
+				await Task.Delay(100);
+
+				// Wait for all properties windows to close
```
yaira2, line 317: We should be able to do something similar for content dialogs.

Lukiluc29:
> I rather think that notifications should remain even after restarting the app unless the user explicitly clears them. In any case, this is something we can discuss as a separate issue.

Where can we discuss this?

hishitetsu:
> > I rather think that notifications should remain even after restarting the app unless the user explicitly clears them. In any case, this is something we can discuss as a separate issue.
> 
> Where can we discuss this?

You can open a new issue once this PR is merged.

mdtauk:
When the app is in the background, could Files use the system notifications to alert on the completion of on-going tasks?

hishitetsu:
> When the app is in the background, could Files use the system notifications to alert on the completion of on-going tasks?

It stops the thread, so that may be difficult as it is now.
A related issue is already open. #8207

yaira2:
For what its worth, file operations already continue in the background even if the app is fully terminated.

src/Files.App/Views/Settings/AdvancedPage.xaml
```diff
@@ -98,6 +98,17 @@
...
+				<local:SettingsBlockControl.Icon>
+					<FontIcon Glyph="&#xE708;" />
```
yaira2, line 104:
```suggestion
					<FontIcon Glyph="&#xE8E6;" />
```

0x5bfa (approved):
Looks great to me in terms of the codebase quality!

yaira2 (approved):
This is a really nice improvement, it's a lot more enjoyable to use Files when you don't have to wait for it to load. LGTM!

PR #13264, opened 2023-08-28
Fix: Close all notification banners except in progress before sleep

### LINKED ISSUE #13243: Feature: Request to Disable Status Center Display in Background Mode for Files

I have a suggestion regarding the Files application. Could it be possible to prevent the status center from appearing when I open Files with the "Leave app running in the background when the window is closed" mode enabled? Sometimes, I don't want to see my previous actions. Typically, when we close an application, we don't necessarily want to review our previous activities.

### Requirements

Identify the user preference for the visibility of the status center in the background mode.

### Files Version

2.5.23.0

### Windows Version

22H2 22621.1992

### Comments

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #13243 

**Validation**

### DIFF
--- a/src/Files.App/UserControls/OngoingTasksFlyout.xaml.cs
+++ b/src/Files.App/UserControls/OngoingTasksFlyout.xaml.cs
@@ -1,11 +1,8 @@
 // Copyright (c) 2023 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using Files.App.ViewModels;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
-using System;
-using System.Threading.Tasks;
 
 // The User Control item template is documented at https://go.microsoft.com/fwlink/?LinkId=234236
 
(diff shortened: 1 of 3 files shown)


src/Files.App/UserControls/OngoingTasksFlyout.xaml.cs
```diff
@@ -37,14 +34,7 @@ private async void Button_Click_1(object sender, RoutedEventArgs e)
...
 		{
-			for (int i = OngoingTasksViewModel.StatusBannersSource.Count - 1; i >= 0; i--)
-			{
-				var itemToDismiss = OngoingTasksViewModel.StatusBannersSource[i];
-				if (!itemToDismiss.IsProgressing)
-				{
-					OngoingTasksViewModel.CloseBanner(itemToDismiss);
-				}
-			}
+			OngoingTasksViewModel.CloseAllBanner();
```
yaira2, line 37: While you're in here, would it be a good idea to rename it to "ViewModel"?

yaira2 (approved):
LGTM

src/Files.App/UserControls/OngoingTasksFlyout.xaml.cs
hishitetsu: OngoingTasksViewModel cannot be renamed to "ViewModel" because it is referenced from AddressToolbar which has a ToolbarViewModel named "ViewModel".

hishitetsu: I was a little mistaken. I have renamed it.

PR #13269, opened 2023-08-28
Feature: What's new popup

yaira2:
**Resolved / Related Issues**
- Replace the What's new teaching tip with a popup

**Validation**
   1. Hardcoded `IsReleaseNotesVisible` to true
   2. Hard coded the markdown content
   3. Opened the popup

**Screenshots (optional)**
image

### DIFF
--- a/src/Files.App/ViewModels/UserControls/ToolbarViewModel.cs
+++ b/src/Files.App/ViewModels/UserControls/ToolbarViewModel.cs
@@ -82,7 +82,7 @@ namespace Files.App.ViewModels.UserControls
 			set => SetProperty(ref releaseNotes, value);
 		}
 
-		private bool isReleaseNotesVisible = true;
+		private bool isReleaseNotesVisible;
 		public bool IsReleaseNotesVisible
 		{
 			get => isReleaseNotesVisible;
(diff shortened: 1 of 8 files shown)


src/Files.App/ViewModels/UserControls/ToolbarViewModel.cs
```diff
@@ -222,9 +217,15 @@ private async void UpdateService_OnPropertyChanged(object? sender, PropertyChang
 				await CheckForReleaseNotesAsync();
 		}
 
-		private void DoViewReleaseNotes()
+		private async void DoViewReleaseNotes()
```
hishitetsu, line 220: We should also change `ViewReleaseNotesCommand` to an AsyncRelayCommand.

src/Files.Core/ViewModels/Dialogs/ReleaseNotesDialogViewModel.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) 2023 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.Core.ViewModels.Dialogs
+{
+	public sealed class ReleaseNotesDialogViewModel : ObservableObject
+	{
+		public string _ReleaseNotesMadrkdown = string.Empty;
```
hishitetsu, line 8:
```suggestion
		private string _ReleaseNotesMadrkdown;
```
It should be private and specifying an initial value is meaningless because the constructor always overwrites the value.

hishitetsu (approved):
LGTM

PR #13331, opened 2023-09-05
Package: Localized application description

0x5bfa:
### Motivation & Context

- Localize English more

### PR Checklist

### Screenshots

_None_

### DIFF
--- a/src/Files.App (Package)/Package.appxmanifest
+++ b/src/Files.App (Package)/Package.appxmanifest
@@ -85,7 +85,7 @@
 
             <uap:VisualElements
                 DisplayName="Files - Dev"
-                Description="Organize your files and folders easily. Robust multitasking experiences, helpful tags, and deep integrations add to an intuitive design, made by the community."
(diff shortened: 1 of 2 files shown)


0x5bfa:
This description is shown in the Settings app on Windows 10 and 11.

yaira2:
Where?

0x5bfa:
I forgot, but I remember it is probably shown when you refer app info in the Settings app.

As you can see, this appears to users and can be localized https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-description

0x5bfa:
If you think it is not needed, I'm fine with that and then close.

yaira2:
I wonder if we can remove it all together, is it a required value in the manifest?

0x5bfa:
Yes. Required field.

https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-uap-visualelements

image

PR #13342, opened 2023-09-07
Fix: Fixed issue where the path bar would display wrong location at app startup

### LINKED ISSUE #13340: Bug: The path bar displays wrong location when restarting the app
### Description

When background running is off, the path bar displays wrong location when restarting the app. For example, when the C drive is opened, the path bar displays OneDrive as the location.
image

### Steps To Reproduce

1. Set leave app running to off.
2. Open the C drive.
3. Close the app.
4. Restart the app.
5. The app will open the C drive, but the path bar will display OneDrive as the location.

### Requirements

Display right location.

### Files Version

2.5.29.0

### Windows Version

10.0.19045.3324

### Log File

n/a

hishitetsu:
**Resolved / Related Issues**
   Closes #13340 

**Validation**

**Note**
The path bar will display the correct location, but will not display its volume label. This is because the drive information has not yet been loaded immediately after startup. I leave it as is because waiting for the drive information to be loaded slows down the page loading.


yaira2 (approved):
LGTM

Issue #10367, opened 2022-10-31
Refresh Favorite Bar


Djtpj:
### What feature or improvement do you think would benefit Files?

A way to refresh the icons on the favorites bar. I really like to use folder icons with my favorited folders so that it is easy to identify them. 

### Requirements

- A simple way to refresh or update the icons on the favorites bar.

### Files Version

2.3.12

### Windows Version

Windows 11 22000 1098

yaichenbaum:
What happens if you restart the app, does the correct icon load?

Djtpj:
Yes, it does.

0x5bfa:
FWIW, there's [a Windows Shell API](https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shchangenotifyregister) to detect shell folders and normal folders but I'm not sure we can do with WinUIEx because notifications are set to the registered window (in this case Files window) via WndProc. That's said, I believe we can create a hidden window object for us to be notified of changes.

```c#
LRESULT WndProc(HWND hWnd, unit message, WPARAM wParam, LPARAM lParam)
{
    switch (message)
    {
        case WM_SHCHANGENOTIFY:

            LPITEMIDLIST** ppidl;
            long* lEvent;
            var hLock = SHChangeNotification_Lock((nint)wParam, (uint)lParam, ppidl, lEvent);

            if (!hLock)
                break;

            if (*lEvent == SHCNE.SHCNE_UPDATEIMAGE)
                UpdateSidebarIcons();

            SHChangeNotification_Unlock(hLock);
            break;

        //...
    }

    return (LRESULT)0;
}
```

We have a method for this but we don't use this as of now.

https://github.com/files-community/Files/blob/db3491617d7e8d46fef01d798006ab96d3be88e7/src/Files.App/Data/Items/WindowEx.cs#L266-L293

---

FileSystemWatcher won't work for virtual shell folders such as Recycle Bin and Quick Access because it uses a raw Win32 API that requires a full Win32 namespace path.

PR #13344, opened 2023-09-07
Feature: Update the RecycleBin icon when emptying/filling/restoring it

### LINKED ISSUE #9946: Bug: Icon on the desktop does not update after emptying the bin
**Steps to reproduce**
- Empty recycle bin
- See that the icon on the desktop still looks full

**Steps to fix**
- Update the icon to be empty

QuaintMako:
**Resolved / Related Issues**
   Closes #9946 

**Validation**
   - Open the app with the Recycle Bin empty and filled to check if the icon is adequate.
   - Used the Empty Recycle Bin, Restore Recycle Bin and Delete File command.


src/Files.App/Actions/FileSystem/BaseDeleteAction.cs
```diff
@@ -35,8 +35,8 @@ protected async Task DeleteItems(bool permanently)
 							: FilesystemItemType.Directory));
 
 			await context.ShellPage!.FilesystemHelpers.DeleteItemsAsync(items, settings.DeleteConfirmationPolicy, permanently, true);
-
 			await context.ShellPage.FilesystemViewModel.ApplyFilesAndFoldersChangesAsync();
+			Ioc.Default.GetRequiredService<SidebarViewModel>().SidebarPinnedModel.UpdateRecycleBinIcon();
```
QuaintMako, line 39: I'm setting this PR as a draft for now because of those lines appearing inside the actions. This is dirty and this needs to be changed, the UI update shouldn't be done inside the command logic.
 
The problem is that no event is fired when executing a command as far as I can tell, so I cannot hook the update on anything. 

I was thinking about creating a system that broadcasts events when executing a command - for example, the command name for starters. 

It's not a huge implementation, but maybe somebody would have an idea to bypass this need? Or would this command execution broadcast event system be interesting to have?

yaira2: Do you think we can move forward without updating the sidebar at runtime? True, the sidebar won't refresh until the next session but at least the desktop icon will update immediately.

QuaintMako: I'm working on another implementation that should solve both issues.

yaira2: Just adding this for reference #10367

PR #13383, opened 2023-09-16
SevenZip exception placeholder

cekkr:
This is not a real pull request, but just a (wrong) way to allow the building of Files:
in fact SevenZipOpenFailedException is not existing, I tried to substitute it with SevenZipExtractionFailedException but I don't think is a good idea. Maybe a real resolution could be using just the generic SevenZipException class and comparing the HResult.


Josh65-2201:
If you're trying to build Files from source you don't need to change the code just follow https://github.com/files-community/Files?tab=readme-ov-file#building-from-source

PR #13397, opened 2023-09-19
Code Quality: Only execute actions once

### LINKED ISSUE #11737: Feature: Only open single tab if user holds down ctrl + t

HorseBandit:
**Resolved / Related Issues**
   Closes #11737 

**Validation**
   1. Open app ...
   2. hold ctrl + t and see that tabs don't continuously open
   3. press ctrl + t a few times
   4. hold ctrl +w and see that tabs don't continuously close
   5. press ctrl + w a few times
   6. refresh the same

**Screenshots (optional)**
Add screenshots here.


HorseBandit:
@yaira2 Ready for review. Fully implemented.

yaira2:
@QuaintMako @Ferrario-Filippo would it be worth extending this behavior to every action? If we do this, we can handle it across the board instead of duplicating the code for each action.

HorseBandit:
@yaira2 Thank you for the review. I'm waiting for further instruction. Is this on track to be merged to main?

yaira2:
I'd like to wait for the others to provide their feedback.

0x5bfa:
redo and undo actions shouldn’t be limited, but otherwise I think all actions should be.

yaira2:
> redo and undo actions shouldn’t be limited

I think they should, we wouldn't want someone undoing multiple operations by mistake.

ferrariofilippo:
> would it be worth extending this behavior to every action? If we do this, we can handle it across the board instead of duplicating the code for each action.

I think we should do that, I can't think of any use-case where a user needs to execute the same action over and over. It may also prevent some bugs we are not aware of.

yaira2:
@HorseBandit can you apply this logic to all actions? Instead of copying it to every action, I think we should build it into the underlying code.

HorseBandit:
@yaira2 @ferrariofilippo change fully implemented for all actions. Ready for you.

src/Files.App/Actions/DebounceAction.cs
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		public abstract string Label { get; }
+		public abstract string Description { get; }
+		public abstract RichGlyph Glyph { get; }
+		public abstract HotKey HotKey { get; }
+		public abstract HotKey SecondHotKey { get; }
+		public abstract HotKey ThirdHotKey { get; }
+		public abstract HotKey MediaHotKey { get; }
+		public abstract bool IsExecutable { get; }
```
QuaintMako, line 36: These should be move above the constructor to keep the structure consistency. (See Class Structure in [the coding style](https://files.community/docs/contributing/code-style )if you're curious 😄 )

src/Files.App/Actions/DebounceAction.cs
```diff
@@ -0,0 +1,40 @@
...
+
+		public virtual bool CanExecuteNow()
+		{
+			return DateTime.Now - lastExecuted > debounceTime;
+		}
+
+		public virtual void MarkAsExecuted()
+		{
+			lastExecuted = DateTime.Now;
+		}
```
QuaintMako, line 27: The name is not exactly telling of what it's doing. It should be something like `MarkLastExecutionTime`.

src/Files.App/Data/Commands/Manager/CommandManager.cs
```diff
@@ -199,149 +199,149 @@ public CommandManager()
...
-			[CommandCodes.ClosePane] = new ClosePaneAction(),
-			[CommandCodes.OpenFileLocation] = new OpenFileLocationAction(),
-			[CommandCodes.PlayAll] = new PlayAllAction(),
-			[CommandCodes.GitFetch] = new GitFetchAction(),
-			[CommandCodes.GitInit] = new GitInitAction(),
-			[CommandCodes.GitPull] = new GitPullAction(),
-			[CommandCodes.GitPush] = new GitPushAction(),
-			[CommandCodes.GitSync] = new GitSyncAction(),
-			[CommandCodes.OpenAllTaggedItems] = new OpenAllTaggedActions(),
+			[CommandCodes.OpenHelp] = new DebouncedActionDecorator(new OpenHelpAction(), TimeSpan.FromMilliseconds(800)),
```
QuaintMako, line 202: Seeing how often the `TimeSpan.FromMilliseconds(800)` is repeated, why not directly put it inside the constructor of `DebouncedActionDecorator` as a default value, and create a second constructor that'd take a timespan as second argument? 
It'd reduce the repetition, simplify the change of the default value and leaves the room for case to case customization.

src/Files.App/Actions/DebouncedActionDecorator.cs
```diff
@@ -0,0 +1,40 @@
...
+		public override HotKey ThirdHotKey => _innerAction.ThirdHotKey;
+		public override HotKey MediaHotKey => _innerAction.MediaHotKey;
+		public override bool IsExecutable => _innerAction.IsExecutable;
+
+		public override async Task ExecuteAsync()
+		{
+			if (!CanExecuteNow())
+			{
+				return;
+			}
```
QuaintMako, line 33: Curly braces are not needed for one liner `if`.

src/Files.App/Actions/DebouncedActionDecorator.cs
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		public override string Label => _innerAction.Label;
+		public override string Description => _innerAction.Description;
+		public override RichGlyph Glyph => _innerAction.Glyph;
+		public override HotKey HotKey => _innerAction.HotKey;
+		public override HotKey SecondHotKey => _innerAction.SecondHotKey;
+		public override HotKey ThirdHotKey => _innerAction.ThirdHotKey;
+		public override HotKey MediaHotKey => _innerAction.MediaHotKey;
+		public override bool IsExecutable => _innerAction.IsExecutable;
```
QuaintMako, line 26: Should be moved above the constructor

QuaintMako (commented):
A few feedbacks, mostly on the format.
The functionnality looks good to me, I cannot test it for now.

HorseBandit:
@QuaintMako @yaira2 Suggestions taken and implemented. I left one action with an optional value of 1000 milliseconds for demonstration. The others are defaulted at 800ms.

src/Files.App/Data/Commands/Manager/CommandManager.cs
```diff
@@ -199,149 +199,149 @@ public CommandManager()
...
+			[CommandCodes.OpenNewPane] = new DebouncedActionDecorator(new OpenNewPaneAction()),
+			[CommandCodes.ClosePane] = new DebouncedActionDecorator(new ClosePaneAction()),
+			[CommandCodes.OpenFileLocation] = new DebouncedActionDecorator(new OpenFileLocationAction()),
+			[CommandCodes.PlayAll] = new DebouncedActionDecorator(new PlayAllAction()),
+			[CommandCodes.GitFetch] = new DebouncedActionDecorator(new GitFetchAction()),
+			[CommandCodes.GitInit] = new DebouncedActionDecorator(new GitInitAction()),
+			[CommandCodes.GitPull] = new DebouncedActionDecorator(new GitPullAction()),
+			[CommandCodes.GitPush] = new DebouncedActionDecorator(new GitPushAction()),
+			[CommandCodes.GitSync] = new DebouncedActionDecorator(new GitSyncAction()),
+			[CommandCodes.OpenAllTaggedItems] = new DebouncedActionDecorator(new OpenAllTaggedActions(), TimeSpan.FromMilliseconds(1000))
```
0x5bfa, line 344: I believe we should leave as it (800 by default) is for now.

src/Files.App/Actions/DebounceAction.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Files.App.Actions
+{
+	public abstract class DebouncedAction : IAction
+	{
+		private DateTime lastExecuted = DateTime.MinValue;
+
+		private readonly TimeSpan debounceTime;
```
0x5bfa, line 13: debounceTimeSpan

0x5bfa (commented):
being particular but needed. LGTM otherwise, thank you!

src/Files.App/Actions/DebouncedActionDecorator.cs
```diff
@@ -0,0 +1,43 @@
...
+		/// <summary>
+		/// Initializes a new instance of the DebouncedActionDecorator class.
+		/// </summary>
+		/// <param name="innerAction">The IAction instance to be wrapped.</param>
+		/// <param name="debounceDuration">The debounce duration. Default is 800 milliseconds.</param>
+		public DebouncedActionDecorator(IAction innerAction, TimeSpan? debounceDuration = null)
+			: base(debounceDuration ?? TimeSpan.FromMilliseconds(800))
+		{
+			_innerAction = innerAction;
+		}
```
QuaintMako, line 31: How about this? It removes the possibility to have a null item, which is always a good thing.
```suggestion
		public DebouncedActionDecorator(IAction innerAction, TimeSpan debounceDuration = TimeSpan.FromMilliseconds(800))
			: base(debounceDuration)
		{
			_innerAction = innerAction;
		}
```

src/Files.App/Data/Commands/Manager/CommandManager.cs
QuaintMako, line 344: Not sure that adding an example is useful. After all, if someone needs one day to add a new command, their IDE should display the constructor with the possibility of changing the timespan. Here is just adds a "why is this command different from others?" moment when reading it.

HorseBandit:
@0x5bfa The above steps are completed. I had to comment out a line in CommandManager. Could you check that out? It was causing a problem.

HorseBandit:
@0x5bfa @yaira2 @QuaintMako 

Branch synced with main as of now. Ready to merge.

HorseBandit:
@yaira2 Could I get an update on this? I think it's ready to merge. Thank you.

QuaintMako (approved):
LGTM.

0x5bfa (approved):
LGTM

yaira2:
@HorseBandit can you take a look at the build error?

HorseBandit:
@yaira2 Fresh sync followed by a merge of 'main' into this branch. There are no build errors.

hecksmosis:
I think this is a pipeline bug iirc, don't remember what we used to do to fix it

yaira2:
@HorseBandit do the test projects run locally? It seems to be failing in the PR.

HorseBandit:
@yaira2 Yes they do.

yaira2:
@HorseBandit I think the best option is to duplicate the branch and create a new PR.

Issue #12545, opened 2023-06-05
Feature: Ctrl + shift to copy as path with quotes


lapaz17:
### What feature or improvement do you think would benefit Files?

Shift right click and selecting "copy as path" on Explorer.exe implements same function with quotes. This will provide bugs when users want to paste paths containing whitespaces to cmd or similar cli, so add copy feature with quotes

### Requirements

- This proposal will accomplish X
- This proposal will accomplish Y
- This proposal will accomplish Z

### Files Version

2.4.71.0

### Windows Version

Windows 11 (22H2): 22621

yaira2:
Thank you for the feedback

HorseBandit:
@yaira2 Any plan to approve this one?

yaira2:
We're waiting for more feedback before approving this.

PR #13398, opened 2023-09-19
Feature: Added `ctrl + alt + c` shortcut to copy path with quotes

niklasb22:
**Resolved / Related Issues**
Closes: #12545 

**Validation**
   1. Select item and use `ctrl + alt + c` shortcut
   2. Confirm the path is copied to the clipboard with quotes

**Screenshots (optional)**
Add screenshots here.


0x5bfa (commented):
LGTM in terms of codebasel except one thing.

> Are there any other steps that were used to validate these changes?
Please wwrite down.

yaira2:
@niklasb22 thank you for your contribution. We were still reviewing the details on #12545 and deciding if we should move forward with the feature proposal. I think that at this point, it doesn't hurt to include this option, but can you try changing the modifier key from `ctrl + shift` to `alt`?

yaira2:
It might also be worth creating a separate action for this so that both options are shown in the command palette.

yaira2:
@niklasb22 we're ready to move forward with this feature. The changes look good but one thing I would suggest doing differently, instead of adding a modifier key, it might be best to create a separate action with a different shortcut.

yaira2 (approved):
LGTM

PR #13415, opened 2023-09-23
Fix: Fixed scroll position after page transitions

### LINKED ISSUE #5825: Bug: Scroll position isn't always in the correct position
### Description

When in a folder that has folders named with _ at the start or group by options enabled the scroll will not be at the top.

### Steps To Reproduce

1) Create a folder with enough items for scrolling
2) Create a folder with an _ at the start or use group by options
3) Go to previous folder and back
4) See scroll position not at the top

### Expected behavior

Should start at top or directory always

### Files Version

1.9.23.0

### Windows Version

10.0.19043.1151

### Relevant Assets

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #5825

**Validation**
   1. Navigate to the root of Files' GitHub repository.
   2. Navigate to `artifacts` (next to `.github`).
   3. Navigate back.
   4. Scroll position will be top and `.github` will be seen.

### DIFF
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -66,6 +66,7 @@ namespace Files.App.Views.LayoutModes
 		protected override void ItemManipulationModel_ScrollIntoViewInvoked(object? sender, ListedItem e)
 		{
 			FileList.ScrollIntoView(e);
+			// The item will be seen at the bottom of the page
 			ContentScroller?.ChangeView(
 				null,
 				(FileList.Items.IndexOf(e) + 1.5) * Convert.ToInt32(Application.Current.Resources["ListItemHeight"]) - ContentScroller?.ActualHeight + HeaderGrid.Height,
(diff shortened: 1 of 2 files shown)


yaira2:
Does this take into account the group headers as well?

hishitetsu:
> Does this take into account the group headers as well?

Good point. We should consider them.

src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -65,17 +65,16 @@ public DetailsLayoutBrowser() : base()
 
 		protected override void ItemManipulationModel_ScrollIntoViewInvoked(object? sender, ListedItem e)
 		{
-			FileList.ScrollIntoView(e);
-			ContentScroller?.ChangeView(null, FileList.Items.IndexOf(e) * Convert.ToInt32(Application.Current.Resources["ListItemHeight"]), null, true); // Scroll to index * item height
+			(FileList.ContainerFromItem(e) as ListViewItem)?.StartBringIntoView();
```
hishitetsu, line 68: This should work with or without grouping, but sometimes it scrolls as expected and sometimes it does not scroll at all.
I have no idea how to make it work all the time.

hishitetsu:
It seems very difficult to scroll to the selected item immediately after the page transition while taking into account the group headers.

PR #13464, opened 2023-10-02
Fix: Fixed issue with appending "copy" instead of "#"

### LINKED ISSUE #11697: Bug: Generating new name for conflict is appending "copy" instead of a number "(#)"

When I move or copy a file from one folder to another folder that already contains a file with the same name, and I would like to keep both, I would like the option for it to append " (#)" to the file (like Windows) instead of "- Copy" when I select Generate New Name. 

If it were the same file, I would almost always choose Replace or Skip, not Keep Both (aka Generate New Name), so when I choose to keep both, it is almost always because it isn't the same file, and therefore is not a copy, but an additional item with the same name.

### Requirements

- [ ] When copying and pasting items within the same folder, "- copy" should be appended (as is File Explorer)
- [ ] When copying and pasting item from a different folder, selecting generate new name from the conflict modal should append a number (as in File Explorer)

### Files Version

2.4.40.0

### Windows Version

Windows 11 22H2 22623.1020

### Comments

_No response_

Mohammed-619:
**Resolved / Related Issues**
   Closes #11697

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**

### DIFF
--- a/src/Files.App/Utils/Storage/Helpers/FileOperationsHelpers.cs
+++ b/src/Files.App/Utils/Storage/Helpers/FileOperationsHelpers.cs
@@ -439,6 +439,34 @@ namespace Files.App.Utils.Storage
 				return (await moveTcs.Task, shellOperationResult);
 			});
 		}
+		
+		/*********************************************************************************************
+		  This function generates a unique new name for a file in the specified destination folder.
+		  If the destination folder is not accessible, an InvalidOperationException is thrown.
+		**********************************************************************************************/
+		private static string GetUniqueNewName(ShellFolder destination, string originalName)
+		{
+			if (destination == null)
+			{
+				throw new InvalidOperationException("Destination folder is not accessible.");
+			}
+
+			string newName = originalName;
+			int counter = 2;
+
+			// Iterate through the children of the destination folder and check if the new name already exists.
+			// If it does, append a counter to make it unique.
+			foreach (var existingItem in destination.EnumerateChildren(FolderItemFilter.Storage))
+			{
+				if (string.Equals(newName, existingItem.Name, StringComparison.OrdinalIgnoreCase))
+				{
+					newName = $"{originalName} ({counter})";
+					counter++;
+				}
+			}
+
+			return newName;
+		}
 
 		public static Task<(bool, ShellOperationResult)> CopyItemAsync(string[] fileToCopyPath, string[] copyDestination, bool overwriteOnCopy, long ownerHwnd, bool asAdmin, IProgress<StatusCenterItemProgressModel> progress, string operationID = "")
 		{
--- a/src/Files.App/Utils/Storage/Helpers/FileOperationsHelpers.cs
+++ b/src/Files.App/Utils/Storage/Helpers/FileOperationsHelpers.cs
@@ -387,7 +387,7 @@ namespace Files.App.Utils.Storage
 						using ShellItem shi = new ShellItem(fileToMovePath[i]);
 						using ShellFolder shd = new ShellFolder(Path.GetDirectoryName(moveDestination[i]));
 						string newName = GetUniqueNewName(shd, Path.GetFileName(fileToMovePath[i]));
-						op.QueueMoveOperation(shi, shd, Path.GetFileName(moveDestination[i]));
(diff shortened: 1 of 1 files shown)


Josh65-2201:
Thanks for contributing to Files, for future pull request
- In the title of the issue put `Feature:` or `Fix:` depending on what the PR is about. In this case it would be `Fix:`
- Please add the issue number were it says `Closes #issue...` it should look like this `Closes #11697`
- In the square brackets where you put yes replace the space with a `x` it should look like `[x]` and on the preview tab it should show a check mark.

src/Files.App/Utils/Storage/Helpers/FileOperationsHelpers.cs
```diff
@@ -439,6 +440,34 @@ public static Task<(bool, ShellOperationResult)> MoveItemAsync(string[] fileToMo
...
+			if (destination == null)
+			{
+				throw new InvalidOperationException("Destination folder is not accessible.");
+			}
+
+			string newName = originalName;
+			int counter = 2;
+
+			// Iterate through the children of the destination folder and check if the new name already exists.
+			// If it does, append a counter to make it unique.
```
QuaintMako, line 459: The for loop speaks for itself, no need to document its functionality.

src/Files.App/Utils/Storage/Helpers/FileOperationsHelpers.cs
```diff
@@ -439,6 +440,34 @@ public static Task<(bool, ShellOperationResult)> MoveItemAsync(string[] fileToMo
...
+			foreach (var existingItem in destination.EnumerateChildren(FolderItemFilter.Storage))
+			{
+				if (string.Equals(newName, existingItem.Name, StringComparison.OrdinalIgnoreCase))
+				{
+					newName = $"{originalName} ({counter})";
```
QuaintMako, line 464: I'd suggest building the newname only when the counter value is calculated. It's will prevent unnecessary string assignation. In the case of a lot of files sharing the same name, it'd reduce the cost by quite a margin.

src/Files.App/Utils/Storage/Helpers/FileOperationsHelpers.cs
```diff
@@ -439,6 +440,34 @@ public static Task<(bool, ShellOperationResult)> MoveItemAsync(string[] fileToMo
...
+					counter++;
+				}
+			}
+
+			return newName;
```
QuaintMako, line 469: Could return directly `return $"{originalName} ({counter})";`

src/Files.App/Utils/Storage/Helpers/FileOperationsHelpers.cs
QuaintMako: In the end, this foreach can be compacted into a linq request that returns the number of elements with name `originalName` inside the folder.
Might be worth to do the change for the readability gain.

QuaintMako (commented):
I cannot test the functionality, but a few feedback on the existing code already.

yaira2:
@Mohammed-619 thank you for the PR. 

Can you take a moment to fill out the section for "Are there any other steps that were used to validate these changes" so we can follow the step on our side?

Mohammed-619:
Move a file/folder into a folder that has file/folder with the same name.
for example move /Desktop/filename into /Desktop/folder, but /Desktop/folder has to have an item name filename. Please let me know if it didn’t work or if this different from what expected.

Sent from my iPhone

On 19 Oct 2023, at 5:43 PM, Yair ***@***.***> wrote:

﻿

[EXTERNAL EMAIL]

@Mohammed-619<https://urldefense.com/v3/__https://github.com/Mohammed-619__;!!JmPEgBY0HMszNaDT!oA-QZiO1_6k2Zd9A1VpZzuAp0n7rkSdRJOSMv1OUH64xSss8jJA0IZZfyWxoH0zEJPfiWj5kw9QSxS1g_QTq6xDktAqRNy8$> thank you for the PR.

Can you take a moment to fill out the section for "Are there any other steps that were used to validate these changes" so we can follow the step on our side?

yaira2:
@Mohammed-619 I'll test later but can you confirm that the conflicts dialog is still displayed?

PR #13498, opened 2023-10-10
Feature: Added support for opening Git root in VS Code

### LINKED ISSUE #13102: Feature: Open in VS/VSCode button should open the root of the repo

When browsing subfolders of a Git repo, the Open in VS(Code) button in the status bar should open the root of the repo instead of the current subfolders. Additionally, the button should show VS or VSCode depending on whether there is a .sln file in the root of the repo.

### Requirements

- This proposal will make the open in IDE button for Git repos take into account the root of the repo instead of the current subfolder

### Files Version

2.5.21.0

### Windows Version

10.0.25393.1

### Comments

_No response_

yaira2:
**Resolved / Related Issues**
   Closes #13102...

**Validation**
   1. Open sub folder in a Git epo
   2. Click the open button on the status bar
   3. Confirm that choosing "open repo" opened the root of the Git repo
   4. Confirm that choosing "open folder" opens the current directory

**Screenshots (optional)**
Add screenshots here.


src/Files.App/Actions/Open/OpenRepoInVSCodeAction.cs
```diff
@@ -0,0 +1,44 @@
...
+
+		public string Label
+			=> "OpenRepoInVSCode".GetLocalizedResource();
+
+		public string Description
+			=> "OpenRepoInVSCodeDescription".GetLocalizedResource();
+
+		public bool IsExecutable =>
+			_isVSCodeInstalled &&
+			_context.Folder is not null;
```
ferrariofilippo, line 22: I think we should disable this action when not in a git repo. At the moment a user can execute it in any folder from the command palette. In this case, the action may not work as expected

yaira2: This should be addressed in [55f41de](https://github.com/files-community/Files/pull/13498/commits/55f41dee3c1059cf2d29796685f9b550bdc5a948)

src/Files.App/UserControls/StatusBarControl.xaml
```diff
@@ -163,26 +163,47 @@
...
-			<!--  Open in VS Code Button  -->
+			<!--  Open in IDE Button  -->
 			<Button
-				x:Name="OpenInVSCodeButton"
+				x:Name="OpenInIDEButton"
 				Height="24"
 				Padding="8,0,8,0"
 				VerticalAlignment="Center"
-				x:Load="{x:Bind converters:MultiBooleanConverter.AndNotConvert(Commands.OpenInVSCode.IsExecutable, Commands.OpenInVS.IsExecutable), Mode=OneWay}"
+				x:Load="{x:Bind Commands.OpenInVS.IsExecutable, Mode=OneWay}"
```
yaira2, line 172: In the future, we might want to make this button available for non git folders as well.

ferrariofilippo: It doesn't seem to work properly...
At the moment the button is visbile only if there's a `.sln` file in the current folder. If you explore subfolders the button will be hidden

yaira2: I fixed this in the most recent commit

ferrariofilippo (approved):
LGTM

src/Files.App/UserControls/StatusBarControl.xaml
```diff
@@ -163,26 +163,47 @@
...
-						<TextBlock Text="{x:Bind Commands.OpenInVSCode.Label, Mode=OneWay}" />
+						<TextBlock Text="{helpers:ResourceString Name=Open}" />
 					</StackPanel>
 				</Button.Content>
+
+				<Button.Flyout>
+					<MenuFlyout>
+						<MenuFlyoutItem
+							x:Name="OpenInVSButton"
+							x:Load="{x:Bind Commands.OpenInVS.IsExecutable, Mode=OneWay}"
```
yaira2, line 189: This doesn't seem to be working properly

ferrariofilippo: Really? I'll check again

src/Files.App/Actions/Open/OpenRepoInVSCodeAction.cs
yaira2: What's the best way to check if the folder is a git repo?

src/Files.App/UserControls/StatusBarControl.xaml
yaira2: I fixed the issue, but on second thought I think we should remove the option. I'd rather let users choose the IDE instead of this.

src/Files.App/Actions/Open/OpenRepoInVSCodeAction.cs
ferrariofilippo: There should be a variable in the context

yaira2: Can you confirm it's working for both the root and sub folders?

yaira2: Never mind, we only need one option in the root folder.

yaira2: Hmm, seems to be working now.

PR #13533, opened 2023-10-19
Async all the way down

nephi-duff:
**Resolved / Related Issues**
- [x] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
   Closes #12310 

**Validation**
How did you test these changes?
- [x] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [ ] Did you implement any design changes to an existing feature?
   - [ ] Was this change approved?
- [x] Are there any other steps that were used to validate these changes?
   1. Open app ...
   2. Click around to make everything still works

PR #13554, opened 2023-10-22
Fix: Added tooltip to the eject button

0x5bfa:
## Summary

This is a small fix.

## Task Checklist

- [x] Added the tooltip for the eject button
- [x] Clean up
- [ ] Apply requested changes

## Known Issues

_N/A_

## PR Checklist

- [x] **Closes**: _N/A_
- [ ] **Co-authors**: _N/A_

## Steps To Validate Changes

_N/A_

## Screenshots

Before
image

After
image

[^2]: If you removed any en-US string resources, make sure that there are no references of those resources. When you add a new en-US string resources, do not make any changes on other languages' resources; [Crowdin](https://crowdin.com/project/files-app) will do, instead.


0x5bfa:
@yaira2 Ready for review.

src/Files.App/Data/Items/DriveItem.cs
```diff
@@ -214,12 +215,16 @@ public object ToolTip
...
+				ToolTipService.SetToolTip(itemDecorator, "SideBarEjectDevice/Text".GetLocalizedResource());
+
+				itemDecorator.Click += ItemDecorator_Click;
+
 				return itemDecorator;
 			}
 		}
 
-		private async void ItemDecorator_ClickAsync(object sender, RoutedEventArgs e)
+		private async void ItemDecorator_Click(object sender, RoutedEventArgs e)
```
hecksmosis, line 227: I think you should revert this as per #13533

hecksmosis (commented):
Other than that LGTM

src/Files.App/Data/Items/DriveItem.cs
yaira2: Let's keep as is. `Click` events are probably better like this.

0x5bfa: They did a bit confusing work. Practically, we should not append Async suffix to event methods.

hecksmosis: Ok then

src/Files.App/Data/Items/DriveItem.cs
```diff
@@ -214,12 +215,16 @@ public object ToolTip
 						Width = 16
 					}
 				};
-				itemDecorator.Click += ItemDecorator_ClickAsync;
+
+				ToolTipService.SetToolTip(itemDecorator, "SideBarEjectDevice/Text".GetLocalizedResource());
```
yaira2, line 219: Can you rename the existing string to "Eject"?

src/Files.App/Data/Items/DriveItem.cs
yaira2: Sure 👍

src/Files.App/Data/Items/DriveItem.cs
0x5bfa: Are you sure?
If I change it, the resource will be marked as un-translated and it will need to be reviewed again.

yaira2 (approved):
LGTM

PR #13570, opened 2023-10-24
Fix: Fixed issue where tab selection got stuck on the sidebar

### LINKED ISSUE #13548: Bug: Using Tab selection gets stuck resizing the sidebar
### Description

If I use tab to navigate in Files it will select the sidebar width divider and get stuck there. Tab, Shift + Tab will resize it to be bigger.

### Steps To Reproduce

1. Use Tab
2. Wait until sidebar is highlighted
3. See controls only make sidebar bigger

### Requirements

- Tab and shift tab should switch between elements and not resize.

### Files Version

2.5.30.0

### Windows Version

10.0.22621.2428

### Log File

[debug.log](https://github.com/files-community/Files/files/13059055/debug.log)

yaira2:
**Resolved / Related Issues**
   Closes #13548...

**Validation**
   1. Open app ...
   2. Use tab key until sidebar resizer is selected
   3. Confirm using the tab & tab + shift key combination moves the focus off the sidebar

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/UserControls/SideBar/SideBarView.xaml.cs
+++ b/src/Files.App/UserControls/SideBar/SideBarView.xaml.cs
@@ -142,7 +142,14 @@ namespace Files.App.UserControls.Sidebar
 
 		private void SidebarResizerControl_KeyDown(object sender, KeyRoutedEventArgs e)
 		{
-			if (e.Key == VirtualKey.Tab)
+			if
+			(
+				e.Key != VirtualKey.Space &&
+				e.Key != VirtualKey.Enter &&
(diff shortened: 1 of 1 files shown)


yaira2:
> Alternatively, we should filter for those keys at the start of the key handler and ignoring everything else.

I pushed a commit for this

0x5bfa (changes requested):
you shouldn't add line breaks like this.  this would confuse devs as if this is inner function or something.

Must be:
if (A &&
    B &&
    C)
{
}

chingucoding (approved):

Issue #13112, opened 2023-07-31
Feature: Add app execution aliases and URL schemes for each build channel


hishitetsu:
### What feature or improvement do you think would benefit Files?

Currently all Files builds have the same app execution alias (`Files.exe`) and URL scheme (`files-uwp:`). But it is inconvenient when multiple Files builds are installed at the same time.
* From the command line, only one build can be launched, as specified in the Windows settings.
* Auto restart feature doesn't work well.

### Requirements

* The app execution alias should have a build-specific one in addition to `Files.exe`.
(`Files-sta.exe`, `Files-pre.exe`, `Files-dev.exe`)
* URL scheme should be build-specific instead of generic `files-uwp:`.
(`files-app:`, `files-app-pre:`, `files-app-dev:`)

### Files Version

2.5.21.0

### Windows Version

10.0.22621.1992

yaira2:
We'll have to consider this and see if there are any downsides.

hishitetsu:
As far as I have been able to find out, here is.
```
Files.App (Package)\Package.appxmanifest(118):<uap:Protocol ReturnResults="none" Name="files-uwp" />
Files.App (Package)\Package.appxmanifest(123):<uap5:ExecutionAlias Alias="files.exe" />
Files.App.Launcher\FilesLauncher.cpp(175):std::wstring uriWithArgs = L"files-uwp:?cmd=" + str2wstr(wstring_to_utf8_hex(args));
Files.App.Launcher\FilesLauncher.cpp(190)://ShExecInfo.lpFile = L"files.exe";
Files.App.Launcher\FilesLauncher.cpp(205):ShExecInfo.lpFile = L"files-uwp:";
Files.App.Launcher\FilesLauncher.cpp(211)://ShExecInfo.lpFile = L"files.exe";
Files.App.OpenDialog\FilesOpenDialog.cpp(124):ShExecInfo.lpFile = L"files.exe";
Files.App\Actions\Navigation\OpenInNewWindowItemAction.cs(43):var folderUri = new Uri($"files-uwp:?folder={@selectedItemPath}");
Files.App\App.xaml.cs(312):!Process.GetProcessesByName("Files").Any(x => x.Id != Process.GetCurrentProcess().Id))
Files.App\App.xaml.cs(538):await Launcher.LaunchUriAsync(new Uri("files-uwp:"));
[...]

PR #13581, opened 2023-10-26
GitHub: Add action to bump version number

0x5bfa:
## Summary

I'd like to introduce the Actions bump up the package version. We're going to be able to use different package file and let the project file choose best package. In that way, we'll need a number of packaging file to distinguish their slight differences such as already shown in https://github.com/files-community/Files/issues/13112.

> [!IMPORTANT]
> @yair100 before merge or approve this PR, please add the token that have write access to the repository called repo:write, which is not included in the default token.

## Task Checklist

- [x] Add package version bump GitHub Actions
- [ ] Clean up
- [ ] Apply requested changes

## Known Issues

Following issues would be fixed in the upcoming PR.

_N/A_

## PR Checklist

- [x] **Closes**: _N/A_
- [ ] **Co-authors**: _N/A_

## Steps To Validate Changes

_N/A_

## Screenshots

_N/A_

[^2]: If you removed any en-US string resources, make sure that there are no references of those resources. When you add a new en-US string resources, do not make any changes on other languages' resources; [Crowdin](https://crowdin.com/project/files-app) will do, instead.


yaira2:
Is there any benefit over doing this manually? We need to type the version number regardless.

0x5bfa:
You don't need to open a PR and edit package manifest.

yaira2:
It's about the same amount of work.

PR #13594, opened 2023-10-28
Feature: Added app name to the splash screen

0x5bfa:
## Summary

During this few months tiral of initial splash screen, user have never complained about it. However, we noticed the blurry of the image and then @mdtauk suggested great idea to add the build branch name under the splash screen image.

## Task Checklist

- [x] Fixed image blurry
- [x] Added the label of the build branch
- [x] Clean up
- [ ] Apply requested changes

## Known Issues

_N/A_

## PR Checklist

- [x] **Closes**: _N/A_
- [x] **Co-authors**: @mdtauk

## Steps To Validate Changes

1. Launch the app
2. See the splash screen, which will be shown in second you launched and only for few moments

## Screenshots

image

### DIFF
--- a/src/Files.App/Views/SplashScreenPage.xaml.cs
+++ b/src/Files.App/Views/SplashScreenPage.xaml.cs
@@ -24,12 +24,12 @@ namespace Files.App.Views
 			InitializeComponent();
 		}
 
-		private void SplashScreenImage_ImageOpened(object sender, RoutedEventArgs e)
+		private void Image_ImageOpened(object sender, RoutedEventArgs e)
 		{
 			App.SplashScreenLoadingTCS?.TrySetResult();
 		}
 
-		private void SplashScreenImage_ImageFailed(object sender, RoutedEventArgs e)
+		private void Image_ImageFailed(object sender, RoutedEventArgs e)
 		{
 			App.SplashScreenLoadingTCS?.TrySetResult();
 		}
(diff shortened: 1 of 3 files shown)


src/Files.App/Views/SplashScreenPage.xaml.cs
```diff
@@ -11,17 +11,25 @@ namespace Files.App.Views
...
+				_ => string.Empty,
+			};
+
 		public SplashScreenPage()
 		{
 			InitializeComponent();
 		}
 
-		private void Image_ImageOpened(object sender, RoutedEventArgs e)
+		private void SplashScreenImage_ImageOpened(object sender, RoutedEventArgs e)
```
yaira2, line 27: Do we really need these events? I think @d2dyno1 suggested we remove them.

0x5bfa: I can remove. I don't know what is this TCS doing actually.

d2dyno1: I removed it in my PR

0x5bfa: Ok then, I’d like this to be merged.

d2dyno1: If you could rename it back to what it was, to avoid any future conflicts, if it's not needed 🙂

0x5bfa: Sure.

yaira2 (approved):
LGTM

Issue #12737, opened 2023-06-26
Bug: Git columns causing CPU to increase


soleon:
### Description

Performance of listing files in any folder in a git repository is abysmal for me to say the least. As soon as I open a git folder, CPU maxes out for a short period of time while the UI freezes/stutters intermittently, making the app unusable.

Sometimes, switching from one folder to another in the same git repository also freeze the app for a few seconds to a few 10 seconds before the content of the folder actually changes.

During these freezes, nothing seemed to have been logged to the debug.log file, I have attached it anyway.

I noticed repetitive `COMExceptions` in the log with the message "由于超时时间已过，该操作返回", they translate to "This operation returned because the timeout period expired".

See the following screenshot for the CPU utilisation over 60 seconds opening a folder containing 108 folders/files.
![image](https://github.com/files-community/Files/assets/780800/e6f0c115-71c9-4f62-9cab-b3fa714ae3d3)

### Steps To Reproduce

Open any folder in a git repository.

### Requirements

At least provide an option to disable native git integration, in case if this is a rare case that only happens to my git repo or on my PC.

### Files Version

2.5.10.0

### Windows Version

Windows 11 Pro 22H2 OS Build 22621.1848 Windows Feature Experience Pack 1000.22642.1000.0

### Log File

[debug.log](https://github.com/files-community/Files/files/11863905/debug.log)

yaira2:
Thank you for your feedback, we're looking into this.

yaira2:
The CPU seems to spike for a couple seconds when opening any type of folder, this is something we're definitely working on but it doesn't seem to be specific to Git.

soleon:
Thanks @yaira2. I have done a bit more testing on different computers. The UI freeze was consistent across all devices. However, I found that it happens a lot worse to the git folder in my screenshot. Other git folders had similar symptom, but they weren't freezing as bad.

This folder in a rather large repo. I've tried copying all files in the folder to a new repo and that folder was swift and responsive. Looks like it has something to do with the size of the repo.

yaira2:
How many branches does this repo have?

soleon:
850... I know, we haven't done a great job to keep it tidy 😅

yaira2:
I think that might be the cause, @ferrariofilippo any thoughts?

ferrariofilippo:
That might be the case, I remeber @hecksmosis reported something similar in one of the first PRs (I don't remember exactly the issue, but it was related to repos with +100 branches)
[...]

PR #13608, opened 2023-10-30
Code quality: Move Git operations to a different thread

ferrariofilippo:
**Resolved / Related Issues**
   Related #12737 

**Changes**
- Branches are enumerated and filtered on a different thread
- `Head branch` name is now "cached"
- `Fetch, Pull, Push, Sync, Checkout` are now executed on a different thread.

**Note**
You may need to log-in github from `Files Preview or Stable` to use some functionalities

**Validation**
   1. Check that Files doesn't freeze when opening large git repos
   2. Check that git actions (fetch, pull, push, checkout, sync) still work


yaira2:
Fyi @hishitetsu @hez2010

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -33,6 +33,8 @@ internal static class GitHelpers
 
 		private static readonly IApplicationService _applicationService = Ioc.Default.GetRequiredService<IApplicationService>();
 
+		private static readonly ThreadWithMessageQueue _owningThread = new();
```
hishitetsu, line 36: I don't think keep the thread static is a good approach. Could we make it so that we only keep the thread when we do Git operations?

ferrariofilippo: I changed the code, the thread is static but when git operations finish, it will be disposed. This way it won't create multiple threads if more actions are performed at the same time (like when you navigate to a repo {fetch from origin, get branches and get head name})

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -108,30 +121,54 @@ public static string GetOriginRepositoryName(string? path)
...
+		public static async Task<BranchItem[]> GetBranchesNames(string? path)
 		{
 			if (string.IsNullOrWhiteSpace(path) || !Repository.IsValid(path))
 				return Array.Empty<BranchItem>();
 
-			using var repository = new Repository(path);
+			_owningThread ??= new ThreadWithMessageQueue();
+			
+			lock (_operationsCountLock)
+				++_activeOperationsCount;
```
hishitetsu, line 132: It is better to use `Interlocked` class.
https://learn.microsoft.com/dotnet/api/system.threading.interlocked?view=net-7.0

ferrariofilippo: I used `Volatile.Read()` to check the value. Is it ok or should I use `Interlocked.Read()`?

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -759,11 +745,8 @@ private static bool IsAuthorizationException(Exception ex)
 
 		private static void DisposeIfFinished()
 		{
-			bool finished;
-			lock (_operationsCountLock)
-				finished = (--_activeOperationsCount) == 0;
-
-			if (finished)
+			Interlocked.Decrement(ref _activeOperationsCount);
+			if (Volatile.Read(ref _activeOperationsCount) == 0)
```
hishitetsu, line 749:
```suggestion
			if (Interlocked.Decrement(ref _activeOperationsCount)) == 0)
```

src/Files.App/Utils/Git/GitHelpers.cs
hishitetsu: See my suggestion.

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -62,6 +66,13 @@ private set
 
 		public static event EventHandler? GitFetchCompleted;
 
+		public static void TryDispose()
+		{
+			_owningThread?.Dispose();
+			_owningThread = null;
+			_activeOperationsCount = 0;
```
hishitetsu, line 73:
```suggestion
			var threadToDispose = _owningThread;
			_owningThread = null;
			Interlocked.Exchange(ref _activeOperationsCount, 0);
			threadToDispose?.Dispose();
```
`Dispose()` may take some time so we should null `_owningThread` first.

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -108,30 +119,50 @@ public static string GetOriginRepositoryName(string? path)
...
 
-		public static BranchItem[] GetBranchesNames(string? path)
+		public static async Task<BranchItem[]> GetBranchesNames(string? path)
 		{
 			if (string.IsNullOrWhiteSpace(path) || !Repository.IsValid(path))
 				return Array.Empty<BranchItem>();
 
-			using var repository = new Repository(path);
+			_owningThread ??= new ThreadWithMessageQueue();
+			Interlocked.Increment(ref _activeOperationsCount);
```
hishitetsu, line 128:
```suggestion
			Interlocked.Increment(ref _activeOperationsCount);
			_owningThread ??= new ThreadWithMessageQueue();
```

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -108,30 +119,50 @@ public static string GetOriginRepositoryName(string? path)
...
+		public static async Task<string> GetRepositoryHeadName(string? path)
 		{
 			if (string.IsNullOrWhiteSpace(path) || !Repository.IsValid(path))
 				return string.Empty;
 
-			using var repository = new Repository(path);
-			return GetValidBranches(repository.Branches)
-				.FirstOrDefault(b => b.IsCurrentRepositoryHead)?.FriendlyName ?? string.Empty;
+			_owningThread ??= new ThreadWithMessageQueue();
+			Interlocked.Increment(ref _activeOperationsCount);
```
hishitetsu, line 154:
```suggestion
			Interlocked.Increment(ref _activeOperationsCount);
			_owningThread ??= new ThreadWithMessageQueue();
```

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -180,16 +211,25 @@ public static async Task<bool> Checkout(string? repositoryPath, string? branch)
 
 			try
 			{
-				if (checkoutBranch.IsRemote)
-					CheckoutRemoteBranch(repository, checkoutBranch);
-				else
-					LibGit2Sharp.Commands.Checkout(repository, checkoutBranch, options);
+				_owningThread ??= new ThreadWithMessageQueue();
+				Interlocked.Increment(ref _activeOperationsCount);
```
hishitetsu, line 215:
```suggestion
				Interlocked.Increment(ref _activeOperationsCount);
				_owningThread ??= new ThreadWithMessageQueue();
```

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -266,20 +308,28 @@ public static void FetchOrigin(string? repositoryPath)
...
 			{
-				IsExecutingGitAction = false;
+				IsExecutingGitAction = true;
 			});
 
 			try
 			{
-				foreach (var remote in repository.Network.Remotes)
+				_owningThread ??= new ThreadWithMessageQueue();
+				Interlocked.Increment(ref _activeOperationsCount);
```
hishitetsu, line 317:
```suggestion
				Interlocked.Increment(ref _activeOperationsCount);
				_owningThread ??= new ThreadWithMessageQueue();
```

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -390,7 +457,15 @@ public static async Task PushToOriginAsync(string? repositoryPath, string? branc
 						b => b.UpstreamBranch = branch.CanonicalName);
 				}
 
-				repository.Network.Push(branch, options);
+				_owningThread ??= new ThreadWithMessageQueue();
+				Interlocked.Increment(ref _activeOperationsCount);
```
hishitetsu, line 461:
```suggestion
				Interlocked.Increment(ref _activeOperationsCount);
				_owningThread ??= new ThreadWithMessageQueue();
```

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -315,14 +365,25 @@ public static async Task PullOriginAsync(string? repositoryPath)
...
+			});
 
 			try
 			{
-				LibGit2Sharp.Commands.Pull(
-					repository,
-					signature,
-					_pullOptions);
+				_owningThread ??= new ThreadWithMessageQueue();
+				Interlocked.Increment(ref _activeOperationsCount);
```
hishitetsu, line 376:
```suggestion
				Interlocked.Increment(ref _activeOperationsCount);
				_owningThread ??= new ThreadWithMessageQueue();
```

hishitetsu (commented):
This is still not completely thread-safe theoretically, but I think it's actually okay.

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -315,14 +366,25 @@ public static async Task PullOriginAsync(string? repositoryPath)
...
+
+				await _owningThread.PostMethod(() =>
+				{
+					LibGit2Sharp.Commands.Pull(
+						repository,
+						signature,
+						_pullOptions);
+				});
+
+				DisposeIfFinished();
```
hishitetsu, line 387: Sorry to keep pointing out.
These processes can be combined in one place by creating a method that takes an Action as an argument like following.
```csharp
private static async Task PostMethodToThreadWithMessageQueue(Action payload)
{
	Interlocked.Increment(ref _activeOperationsCount);
	_owningThread ??= new ThreadWithMessageQueue();

	await _owningThread.PostMethod(payload);

	DisposeIfFinished();
}
```

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -664,5 +706,33 @@ private static bool IsAuthorizationException(Exception ex)
...
+		}
+
+		private static async Task PostMethodToThreadWithMessageQueueAsync(Action payload)
+		{
+			Interlocked.Increment(ref _activeOperationsCount);
+			_owningThread ??= new ThreadWithMessageQueue();
+
+			await _owningThread.PostMethod(payload);
+
+			DisposeIfFinished();
```
hishitetsu, line 723:
```suggestion
			try
			{
				_owningThread ??= new ThreadWithMessageQueue();

				await _owningThread.PostMethod(payload);
			}
			finally
			{
				DisposeIfFinished();
			}
```
`DisposeIfFinished()` should always be executed.

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -664,5 +706,33 @@ private static bool IsAuthorizationException(Exception ex)
...
+		private static async Task<T> PostMethodToThreadWithMessageQueueAsync<T>(Func<object> payload)
+		{
+			Interlocked.Increment(ref _activeOperationsCount);
+			_owningThread ??= new ThreadWithMessageQueue();
+
+			var result = await _owningThread.PostMethod<T>(payload);
+
+			DisposeIfFinished();
+
+			return result;
```
hishitetsu, line 735: As above.

yaira2:
Should we increase the number of branches that we display?

ferrariofilippo:
> Should we increase the number of branches that we display?

I think we can do it now that it runs on a separate thread. However, I think we should not display ALL the branches, I guess we may keep it below 50. We may double the threshold (from 15 to 30) and see if it causes any issues (I can't test by myself since I don't have any repo with that many branches)

src/Files.App/Utils/Git/GitHelpers.cs
ferrariofilippo: Different methods need different exception handling, I think we should call `DisposeIfFinished()` separately. The other option is to use some kind of error codes

hishitetsu: I think at least the decrementing of the count should be done in `finally` to maintain the integrity of the count.

yaira2:
Let's go with 30 👍

hishitetsu:
When I return to a folder displaying Git columns with the back button, the Git columns are empty.

ferrariofilippo:
> When I return to a folder displaying Git columns with the back button, the Git columns are empty.

@hishitetsu there's a PR (#13533) which may interfere with one of yours (#13211)
Because `BaseLayout.OnNavigatedTo()` is now async, this line of code may reset git columns:
```	
    // Git properties are not loaded by default
    ParentShellPageInstance.FilesystemViewModel.EnabledGitProperties = GitProperties.None;
``` 
How should we proceed?

hishitetsu:
> (edit. it seems to work)

I'm not quite sure why this is okay, but it certainly solved the issue.

hishitetsu (approved):
LGTM

yaira2:
Are there any good ways to stress test these changes?

ferrariofilippo:
> I'm not quite sure why this is okay, but it certainly solved the issue.

Making `BaseLayout.OnNavigatedTo()` async caused this line
https://github.com/files-community/Files/blob/main/src/Files.App/Views/LayoutModes/BaseLayout.cs#L480C11-L480C11
to be executed after this one
https://github.com/files-community/Files/blob/main/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs#L126
causing `EnabledGitProperties` to reset to `None`

soleon:
When browsing any folder in a large git repo, the UI was still unresponsive, froze everywhere, the app was unusable. This happened with or without VS or debugger attached.

ferrariofilippo:
I did some tests, here are the conclusions:
- Changing the priority of `GitOperations-Thread` does not consistently reduce CPU spikes (I tried with both `BelowNormal` and `Lowest`.
- The source of the freeze seems to be `GitHelpers.GetGitInformationForItem`, which used the CPU for 10-15% of the total time. That's a lot considering I opened many non-git folders. 
- `LibGit2Sharp.DiffCompare<T>` run for about 4-8% of total CPU time

I'll see if I can find a fix

yaira2:
@soleon can you try with the latest commit?

yaira2 (approved):
LGTM

soleon:
> @soleon can you try with the latest commit?

Sorry for the late response @yaira2, I can confirm this problem is fixed in the latest 'main' branch 🎉.

yaira2:
Thank you for testing

PR #13620, opened 2023-10-31
GitHub: Added build CI

0x5bfa:
## Summary

We are finally going to make build & test CI on the GitHub Actions to gain better integrity. This enables us to configure and manage the CI easily, and have better access to the package artifacts.

We had some difficulties below (wrote down to view quickly what we've overcome)
- Not being able to run the interaction tests on GitHub Actions
  - Fixed by adding proper app id, though it is still unclear that we were able to run that tests on Azure DevOps properly.
- Not being able to add a comment on PR that triggered CI
  - Will have fixed by using 'pull_request_target', which gives us more permissions nonethless makes the workflow more secure. If you'd like to know the differences between them, visit GitHub docs.

## Task Checklist

- [x] Added build job
- [x] Added test job
- [x] ~~Added creator of a comment on PR~~
- [x] Clean up
- [ ] Apply requested changes

## Known Issues

_N/A_

## PR Checklist

- [x] **Closes**: _N/A_
- [x] **Co-authors**: @chingucoding

## Steps To Validate Changes

1. Create a commit on your branch on your files-community fork.
2. Add a PR
3. See CI run building and testing

## Screenshots

_N/A_


.github/workflows/build-dev.yml
```diff
@@ -0,0 +1,69 @@
...
+    - name: Build ${{ env.SOLUTION_NAME }}
+      shell: pwsh
+      run: |
+        msbuild "$env:PACKAGE_PROJECT_PATH" `
+        -t:Build `
+        -p:Platform=$env:PLATFORM `
+        -p:Configuration=$env:CONFIGURATION `
+        -p:AppxBundle=Never
+
+# TODO: Add tests running
```
yaira2, line 69: I'll wait for these before reivewing

builds/azure-pipelines.yml
yaira2, line 1: Are we ready to remove this once the build-dev action is added?

0x5bfa: Yes totally.

.github/workflows/build-dev.yml
0x5bfa: Added

builds/azure-pipelines.yml
0x5bfa: Removed

.github/workflows/ci.yml
```diff
@@ -0,0 +1,171 @@
...
+  APPX_PACKAGE_DIR: ${{ github.workspace }}\artifacts\AppxPackages
+  APPX_SELFSIGNED_CERT_PATH: '${{ github.workspace }}\builds\Files_SelfSigned.pfx'
+
+jobs:
+  build:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        configuration: [Debug]
```
yaira2, line 42: Would it be worth having one for release as well?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,171 @@
...
+      name: Build Files
+      run: |
+        msbuild $env:PACKAGE_PROJECT_PATH `
+        -t:Build `
+        -p:Configuration=$env:CONFIGURATION `
+        -p:Platform=$env:ARCHITECTURE `
+        -p:AppxBundle=Never
+  
+    - if: env.ARCHITECTURE == env.PRIMARY_ARCHITECTURE
+      name: Build & Publish Files
```
yaira2, line 86:
```suggestion
      name: Build & Package Files
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,171 @@
...
+        path: ${{ env.ARTIFACTS_STAGING_DIR }}
+
+  test:
+    needs: [build]
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        configuration: [Debug]
+        platform: [x64]
```
yaira2, line 116: Can we do this for the ARM version as well?

builds/azure-pipelines.yml
yaira2, line 1: 🎉

tests/Files.InteractionTests/SessionManager.cs
```diff
@@ -16,7 +16,8 @@ public class SessionManager
 		private const string WindowsApplicationDriverUrl = "http://127.0.0.1:4723";
 		private static string[] FilesAppIDs = new string[]{
 			"49306atecsolution.FilesUWP_dwm5abbcs5pn0!App",
-			"FilesDev_ykqwq8d6ps0ag!App"
+			"FilesDev_ykqwq8d6ps0ag!App",
```
yaira2, line 19: @chingucoding can we remove the old id if it's not longer being used?

.github/workflows/ci.yml
0x5bfa: Sadly, GitHub-hosted runner doesn's support Windows ARM64 environment, however self-hosted runner supports it as a public beta.

.github/workflows/ci.yml
0x5bfa: I'm thinking that we can disable building&testing Release config on PRs to keep this fast and enable only on the main branch, like microsoft-ui-xaml team already does?

.github/workflows/ci.yml
yaira2: definitely not a blocker for now but might be nice to do this in the future.

builds/azure-pipelines.yml
0x5bfa: :partying_face:

.github/workflows/ci.yml
yaira2: Our existing pipeline is currently set to release. If timing is a concern, we can remove debug and just do release.

0x5bfa: Debug seems to be faster(build&test on Debug: 15 mins, Release: 20mins) and we can at least run Release only on the main branch. Are you sure?

yaira2: Yes, better to catch issues before merging the PR.

0x5bfa: > Our existing pipeline is currently set to release. If timing is a concern, we can remove debug and just do release.

Adding or replacing doesn't make any difference because they will be run parallelly and duration will be the same

0x5bfa: OK!

0x5bfa:
## Files Test Results

```
Time: 24/11/2023 11:23 pm JST
Duration: 3 mins 23 sec
All: 3, Passed: 2, Failed: 1
```

Test name|Result|Failed Location
---|---|---
FolderTest|✅|_None_
GeneralAppTest|✅|_None_
SettingsTests|❌|`Files.InteractionTests.Tests.SettingsTests.VerifySettingsAreAccessible`

<details>
<summary>Test log</summary>

```
Assert.Fail failed. Failed with the following accessibility errors 
Element Group(50026) violated rule 'The Name property of a focusable element must not be null.'.
Element Group(50026) violated rule 'The Name property of a focusable element must not be null.'.
Element Group(50026) violated rule 'The Name property of a focusable element must not be null.'.
```
</details>

tests/Files.InteractionTests/SessionManager.cs
chingucoding: Yeah i think we can remove it

.github/workflows/ci.yml
0x5bfa: Done

tests/Files.InteractionTests/SessionManager.cs
0x5bfa: Which one? I'm gonna remove and see test result

chingucoding: The one in line 19 and maybe even the one in line 18

.github/workflows/ci.yml
```diff
@@ -0,0 +1,228 @@
...
+    - name: Checkout the repository
+      uses: actions/checkout@v3
+
+    - name: Setup MSBuild
+      uses: microsoft/setup-msbuild@v1
+
+    - name: Setup NuGet
+      uses: NuGet/setup-nuget@v1.1.1
+
+    - name: Setup .NET 7
```
heftymouse, line 98: Setup .NET should be moved above installing XamlStyler

0x5bfa: But installing .NET is not check-xaml-formatting.yml
Do you know why this have kept failing even thought I did completely copied the same to here?

heftymouse: It's because the action checks out with a depth of 1, but the formatting part needs the last two commits to format only the XAML files changed in the last commit
I did it this way so it's faster, but it may be worth changing it to do the entire repo if it doesn't have a significant impact

0x5bfa: Thank you! I disagree with make it run entirely because we already must have ran this checking every time we merge a PR and I believe we don't need to ensure again every time someone create a commmit within their PR.

heftymouse:
It'd be good to cache NuGet packages by setting `<RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>` and a step like this:
```
- uses: actions/cache@v3
  id: nuget-cache
  with:
    path: ${{ env.NUGET_PACKAGES }}
    key: ${{ runner.os }}-${{ hashFiles('**/packages.lock.json') }}

- name: Restore NuGet
    if: steps.nuget-cache.outputs.cache-hit != 'true'
    run: ...
```

0x5bfa:
@heftymouse You mean this save us time about 3 minis in running Restore NuGet?

If so, that's nice and I'm going to do so!

heftymouse:
> I'm not sure I can run restoring when one of the cached dependencies updated, which is very important.

That's what the lockfile and the `key` in the action are for, it'll do a full restore when any of the dependencies are updated

I'm realising now that a better approach would be to just restore every time and cache the result, so if any of the dependencies change it just has to restore the changes

0x5bfa:
Is it better behavior that CI restores from nuget.org (takes 3 min) and cache result always on the main branch (on: push) and CI restores only from cached pool to make the CI on the PR (on: pull_request) as fast as and as reliable as possible?

heftymouse:
> Is it better behavior that CI restores from nuget.org (takes 3 min) and cache result always on the main branch (on: push) and CI restores only from cached pool to make the CI on the PR (on: pull_request) as fast as and as reliable as possible?

Sure, that works

tests/Files.InteractionTests/SessionManager.cs
yaira2: @0x5bfa

scripts/Convert-TrxToMarkdown.ps1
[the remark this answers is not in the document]
0x5bfa: The tests result file

You see this visually in VS2022:
image

yaira2 (approved):

0x5bfa:
Nah, testing is failing. Wait a moment

0x5bfa:
.NET test is failing now. I think removing app IDs is out of the scope and I cannot be sure if I cannot remove each for one by one. I can open a further PR including removal of this after this PR merged.

tests/Files.InteractionTests/SessionManager.cs
0x5bfa: Removal of line 18 was caused to fail tests. Removed line 19 and will see result.

0x5bfa:
Ready for merging

tests/Files.InteractionTests/SessionManager.cs
yaira2: What about the ones on 19 and 20?

0x5bfa: They were added in the PR in order to run properly.

yaira2: @chingucoding can you clarify what each one is for? It might help to also add comments in the code for this.

chingucoding: Only the last one FilesDev_dwm5abbcs5pn0 is needed for the CI, the other ones were probably needed at some point for the CI or local development. I think adding comments for them would be very helpful.

yaira2 (approved):

PR #13642, opened 2023-11-01
Feature: Added support for canceling tab tear off

### LINKED ISSUE #10234: Feature: Esc to cancel tab tear off

Sometimes when using a touchpad, I accidentally drag a tab out to a new window. While still holding down the left button, I tried to cancel it with Esc but didn't work. A cancel function would be helpful to quickly prevent unintentional dragging.

### Requirements

- Implement a cancel shortcut when tearing a tab out of the window <kbd>Esc</kbd>

### Files Version

2.3.49.0

### Windows Version

10.0.22623.746

### Comments

_No response_

yaira2:
**Resolved / Related Issues**
   Closes #10234...

**Validation**
   1. Open multiple tabs
   2. Drag tab outside the window and release mouse
   3. Confirm tab was teared off
   4. Drag tab outside window and press escape
   5. Confirm the drag operation is canceled

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -220,7 +220,7 @@ namespace Files.App.UserControls.TabBar
 
 		private void TabView_TabDragCompleted(TabView sender, TabViewTabDragCompletedEventArgs args)
 		{
-			// Unsubscribe from the key down event, it's only needed when a tag is actively being dragged
+			// Unsubscribe from the key down event, it's only needed when a tab is actively being dragged
 			PreviewKeyDown -= TabDragging_PreviewKeyDown;
 
 			if (ApplicationData.Current.LocalSettings.Values.ContainsKey(TabDropHandledIdentifier) &&
@@ -235,7 +235,7 @@ namespace Files.App.UserControls.TabBar
 
 		private async void TabView_TabDroppedOutside(TabView sender, TabViewTabDroppedOutsideEventArgs args)
 		{
-			// Unsubscribe from the key down event, it's only needed when a tag is actively being dragged
+			// Unsubscribe from the key down event, it's only needed when a tab is actively being dragged
 			PreviewKeyDown -= TabDragging_PreviewKeyDown;
 
 			if (isCancelingDragOperation)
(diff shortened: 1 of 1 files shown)


yaira2:
There seems to be an issue where this doesn't always work.

yaira2:
I'm wondering if there is a focus issue where the focus isn't always on the tag that's being dragged. If this is the case, it can cause the KeyDown event to not be fired.

yaira2:
@gave92 can you try this out?

gave92:
Sry can't test it rn :(

* if you go with this version remember to unsubscribe the event, currently in registers multiple times.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -139,12 +142,24 @@ private void TabHoverSelected(object sender, object e)
...
 			var tabViewItemArgs = (args.Item as TabBarItem).NavigationParameter;
 			args.Data.Properties.Add(TabPathIdentifier, tabViewItemArgs.Serialize());
 			args.Data.RequestedOperation = DataPackageOperation.Move;
 
 			InteropHelpers.GetCursorPos(out dragStartPoint);
 			dragStartTime = DateTimeOffset.UtcNow;
+
+			// Force focus in order that the keydown event works
+			Focus(FocusState.Programmatic);
+			PreviewKeyDown += (s, e) =>
```
yaira2, line 155: Is there a way to unsubscribe?

yaira2: Alternatively we can create a separate method, but we'd need a way to access `args.Cancel`.

gave92:
This mostly works, it fails consistently only when pressing ESC while tab is over a second instance of Files.

yaira2:
> This mostly works, it fails consistently only when pressing ESC while tab is over a second instance of Files

The other window is probably taking the focus, we could work around this by increasing the scope of the key press event but I think it's okay as is.

> What issues did you experience with https://github.com/files-community/Files/commit/b92feedb3c694e4b386d0d8bd532e9430b343c3a?

It would cancel the drag even without pressing esc.

PR #13645, opened 2023-11-02
Rev esc

gave92:
@yair100 this simple check works for me.


gave92:
Lol nevermind xD

yaira2:
@gave92 I think there is an issue where it sometimes loses the focus.

gave92:
Actually this pr works every time for me. But after some tries the tabs overlap each other.

yaira2:
What if you add this to `TabView_TabDragCompleted` as well?

PR #13662, opened 2023-11-05
Fix: Fixed NullReferenceException with DefaultDestinationFolderPath

yaira2:
**Resolved / Related Issues**
   Fixes https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/3137641778u/overview


hishitetsu (commented):
I think we need to consider the case where `SlimContentPage` is null.

yaira2:
In `DecompressArchiveDialogViewModel`?

hishitetsu:
Here, for example.
https://github.com/files-community/Files/blob/f5b10a84b42f92c2fd35a083bc3cb67f18cd7059/src/Files.App/Utils/Archives/DecompressHelper.cs#L211

yaira2:
Do you think that's enough, or should I keep my other changes as well?

hishitetsu:
I don't think the archive will ever be null, but it could be left for safety.

src/Files.App/Utils/Archives/DecompressHelper.cs
```diff
@@ -172,12 +172,18 @@ public static async Task DecompressArchiveHereAsync(IShellPage associatedInstanc
 			if (associatedInstance == null)
 				return;
 
+			if (associatedInstance.SlimContentPage == null)
+				return;
```
hishitetsu, line 176:
```suggestion
			if (associatedInstance?.SlimContentPage == null)
				return;
```

src/Files.App/Utils/Archives/DecompressHelper.cs
```diff
@@ -205,6 +211,9 @@ public static async Task DecompressArchiveToChildFolderAsync(IShellPage associat
...
```
hishitetsu, line 215:
```suggestion
			if (associatedInstance?.SlimContentPage == null)
				return;
```

src/Files.App/Utils/Archives/DecompressHelper.cs
yaira2: This is much better, thanks!

PR #13709, opened 2023-11-07
Fix: Fixed ArgumentNullException in AppUnhandledException

yaira2:
**Resolved / Related Issues**
https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/480119618u/overview


src/Files.App/App.xaml.cs
```diff
@@ -472,7 +472,7 @@ private static void AppUnhandledException(Exception? ex, bool shouldShowNotifica
 			Debugger.Break();
 
 			SaveSessionTabs();
-			App.Logger.LogError(ex, ex.Message);
+			App.Logger.LogError(ex, ex?.Message);
```
hishitetsu, line 475: The exception is `ArgumentNullException`, not `NullReferenceException`. Therefore, if the exception is null, the parameter should be set to something other than null.

yaira2: Would String.Empty work?

hishitetsu: I don't know because I don't have a development environment at hand right now, but I think it should log a message indicating that an error occurred instead of a stack trace.

hishitetsu: I did a little testing, but `LogError(null, null)` didn't throw an exception. So it is not clear if not passing null will prevent crashes.

yaira2: Do you want to go for this change just incase?

hishitetsu: I don't know if ArgumentNullException will no longer occur, but it would be a good idea to include this fix.

hishitetsu (approved):

PR #13748, opened 2023-11-08
Fix: Fixed NullReferenceException with OpenPropertiesWindow

yaira2:
**Resolved / Related Issues**
https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/1504308570u/overview

```
Files.App.Utils.Storage
FilePropertiesHelpers.OpenPropertiesWindow (IShellPage associatedInstance)
Files.App.Actions
OpenPropertiesAction.ExecuteAsync ()
```

### DIFF
--- a/src/Files.App/Utils/Storage/Helpers/FilePropertiesHelpers.cs
+++ b/src/Files.App/Utils/Storage/Helpers/FilePropertiesHelpers.cs
@@ -41,6 +41,9 @@ namespace Files.App.Utils.Storage
 		/// <param name="associatedInstance">Associated main window instance</param>
 		public static void OpenPropertiesWindow(IShellPage associatedInstance)
 		{
+			if (associatedInstance is null)
+				return;
+
 			object item;
 
 			var page = associatedInstance.SlimContentPage;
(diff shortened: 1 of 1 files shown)


hishitetsu (commented):
We should also consider the case where `SlimContentPage` or `FilesystemViewModel` is null.
The former should be treated as no items selected.

src/Files.App/Utils/Storage/Helpers/FilePropertiesHelpers.cs
```diff
@@ -41,6 +41,10 @@ public static nint GetWindowHandle(Window w)
 		/// <param name="associatedInstance">Associated main window instance</param>
 		public static void OpenPropertiesWindow(IShellPage associatedInstance)
 		{
+			if (associatedInstance?.SlimContentPage is null ||
+				associatedInstance?.FilesystemViewModel is null)
+				return;
```
hishitetsu, line 46: This will probably work, but it is better to continue the process if possible to reduce the risk of regression.
Here is my suggestion:
```csharp
			if (associatedInstance is null)
				return;

			object item;

			var page = associatedInstance.SlimContentPage;

			// Item(s) selected
			if (page is not null && page.IsItemSelected)
			{
				// Selected item(s)
				item = page.SelectedItems?.Count is 1
					? page.SelectedItem!
					: page.SelectedItems!;
			}
			// No items selected
			else
			{
				// Instance's current folder
				var folder = associatedInstance.FilesystemViewModel?.CurrentFolder;
				if (folder is null)
					return;
```

hishitetsu (approved):
LGTM

PR #13769, opened 2023-11-09
Fix: Fixed NullReferenceException in OpenDirectoryInNewTabAction

yaira2:
**Resolved / Related Issues**
https://appcenter.ms/orgs/FilesApp/apps/Files/crashes/errors/1495782651u/overview

```
Files.App.Actions
OpenDirectoryInNewTabAction.ExecuteAsync ()
Files.App.ViewModels.UserControls
ToolbarViewModel.CheckPathInputAsync (String currentInput, String currentSelectedPath, IShellPage shellPage)
Files.App.Views.Shells
BaseShellPage.NavigationToolbar_QuerySubmitted (Object sender, ToolbarQuerySubmittedEventArgs e)
System.Threading.Tasks.Task
```

### DIFF
--- a/src/Files.App/Actions/Navigation/OpenDirectoryInNewTabAction.cs
+++ b/src/Files.App/Actions/Navigation/OpenDirectoryInNewTabAction.cs
@@ -38,8 +38,7 @@ namespace Files.App.Actions
 
 		public async Task ExecuteAsync()
 		{
-			if (context.ShellPage?.SlimContentPage is null ||
-				context.ShellPage.SlimContentPage?.SelectedItems is null)
+			if (context.ShellPage?.SlimContentPage?.SelectedItems is null)
 				return;
 
 			foreach (ListedItem listedItem in context.ShellPage.SlimContentPage.SelectedItems)
--- a/src/Files.App/Actions/Navigation/OpenDirectoryInNewTabAction.cs
+++ b/src/Files.App/Actions/Navigation/OpenDirectoryInNewTabAction.cs
@@ -38,7 +38,11 @@ namespace Files.App.Actions
 
 		public async Task ExecuteAsync()
 		{
-			foreach (ListedItem listedItem in context.ShellPage?.SlimContentPage?.SelectedItems)
+			if (context.ShellPage?.SlimContentPage is null ||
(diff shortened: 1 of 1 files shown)


src/Files.App/Actions/Navigation/OpenDirectoryInNewTabAction.cs
```diff
@@ -36,7 +38,7 @@ public OpenDirectoryInNewTabAction()
 
 		public async Task ExecuteAsync()
 		{
-			foreach (ListedItem listedItem in context.ShellPage.SlimContentPage.SelectedItems)
+			foreach (ListedItem listedItem in context.ShellPage?.SlimContentPage?.SelectedItems)
```
hishitetsu, line 41: I don't think this avoids NullReferenceException. If `context.ShellPage?.SlimContentPage` is null, we should return before `foreach`.

yaira2: Can someone call ExecuteAsync if IsExecutable is false?

yaira2: I don't think this code is reachable if SlimContentPage is null.

hishitetsu: Normally your thoughts are right, but the possibility that `SlimContentPage` is nulled by another thread at the moment `ExecuteAsync` is called should be considered.

src/Files.App/Actions/Navigation/OpenDirectoryInNewTabAction.cs
```diff
@@ -36,6 +38,10 @@ public OpenDirectoryInNewTabAction()
 
 		public async Task ExecuteAsync()
 		{
+			if (context.ShellPage?.SlimContentPage is null ||
+				context.ShellPage.SlimContentPage?.SelectedItems is null)
```
hishitetsu, line 42:
```suggestion
			if (context.ShellPage?.SlimContentPage?.SelectedItems is null)
```

hishitetsu (approved):
LGTM

PR #13776, opened 2023-11-09
Fix: Fixed some null reference warnings

yaira2:
**Resolved / Related Issues**
Fixed some null reference warnings


src/Files.App/Actions/FileSystem/CreateShortcutAction.cs
```diff
@@ -17,6 +17,7 @@ public RichGlyph Glyph
 			=> new(opacityStyle: "ColorIconShortcut");
 
 		public override bool IsExecutable =>
+			context.ShellPage is not null &&
```
hishitetsu, line 20: This action seems to work even if `context.ShellPage` is null, so I don't think this check is necessary.

yaira2: 👍

PR #13811, opened 2023-11-11
Fix: Fixed issue where "Open Files on Windows startup" didn't work

### LINKED ISSUE #13720: Bug: Stuck on loading when starting on Windows logon
### Description

When Files is configured to launch on Windows startup, it fails. It stays stuck on the loading page.

### Steps To Reproduce

Activate launch on Windows startup.

### Requirements

Fix the bug.

### Files Version

3.0.1.0

### Windows Version

Windows 11 Pro 23H2 22635.2700

### Log File

[debug.log](https://github.com/files-community/Files/files/13287860/debug.log)

hishitetsu:
**Resolved / Related Issues**
   Closes #13720 

**Validation**

### DIFF
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -183,6 +183,12 @@ namespace Files.App
 					break;
 
 				case IStartupTaskActivatedEventArgs startupArgs:
+					// Just launch the app with no arguments
+					rootFrame.Navigate(typeof(MainPage), null, new SuppressNavigationTransitionInfo());
+					break;
+
+				default:
+					// Just launch the app with no arguments
 					rootFrame.Navigate(typeof(MainPage), null, new SuppressNavigationTransitionInfo());
 					break;
 			}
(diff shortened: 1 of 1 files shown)


src/Files.App/MainWindow.xaml.cs
```diff
@@ -181,6 +181,10 @@ public async Task InitializeApplicationAsync(object activatedEventArgs)
 						await mainPageViewModel.AddNewTabByPathAsync(typeof(PaneHolderPage), fileArgs.Files[index].Path);
 					}
 					break;
+
+				case IStartupTaskActivatedEventArgs startupArgs:
```
hishitetsu, line 185: Would it be better to use `default` instead of specifying the interface?

yaira2: We could use both, this way it's clear that they have the same behavior.

yaira2 (approved):
LGTM

src/Files.App/MainWindow.xaml.cs
yaira2: This might also be useful if/when we add an option to open the app in the background.

Yabino:
Sorry guys, it's still happening, the bug is still there.

[https://drive.google.com/file/d/1NAHuQ2D3KT1qu4l7U3CoPsO6y6TS_kyn/view?usp=sharing](url)

hishitetsu:
@Yabino It has only been merged into the dev branch right now. The bug should be fixed after the next update.

PR #13832, opened 2023-11-12
Code Quality: Fix async warnings

yaira2:
**Resolved / Related Issues**
Fix async warnings


0x5bfa (approved):
There're some methods shall not be async but all of them here look good

yaira2:
> There're some methods shall not be async

How can this be determined and should we suppress the warning in those cases?

0x5bfa:
> > There're some methods shall not be async
> 
> 
> 
> How can this be determined and should we suppress the warning in those cases?

No warnings there. IntelliSense suggests to make it async but no warning.

PR #13847, opened 2023-11-13
Fix: Fixed an issue where selection would get canceled after pressing upper arrow key

### LINKED ISSUE #13798: Bug: Highlight and selection gets cancel if upper arrow key pressed afterwards
### Description

Selection of the part of the file name gets canceled after reaching the start position and again keep pressing (once or more than once) the up arrow key.

This screenshot shows that I selected (shift+left arrow) the full file name from the last and reached the starting letter.
After this, if I press even for once the left arrow key, the whole selection cancels out and the arrow returns to the starting letter only.

### Steps To Reproduce

1. Select the file to rename.
2. Choose "rename' to either add or delete the texts as per your choice.
3. Keep the cursor anywhere except for the starting first letter.
4. Use shift+left arrow key to start the selection and select all the letters till you cover the starting letter as well.
5. Now press the shift+left key once more and the whole selection will cancel out. And your cursor will be at the beginning.

### Requirements

It should not cancel out the selection even after reaching and covering till first starting letter.

### Files Version

3.0.3.0

### Windows Version

10.0.22631.2506

### Log File

[debug.log](https://github.com/files-community/Files/files/13320658/debug.log)

RieBi:
**Resolved / Related Issues**
   Closes #13798 

**Validation**
   1. Open app
   2. Select an item
   3. Press F2 to rename
   4. Move pointer to middle of name
   5. Press Shift+UpArrow many times
   6. Confirm that selection does not get canceled
   7. Go to step 4
   8. Press Shift+DownArrow many times
   9. Confirm that selection does not get canceled

### DIFF
--- a/.github/workflows/format-xaml.yml
+++ b/.github/workflows/format-xaml.yml
@@ -1,0 +1,128 @@
+name: Format XAML
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  format-xaml:
+    if: github.event.issue.pull_request && github.event.comment.body == '/format'
+    runs-on: ubuntu-latest
+    environment: Pull Requests
+    defaults:
+      run:
+        shell: pwsh
+
+    steps:
+      - name: Generate GitHub Apps token
+        id: generate
+        uses: tibdex/github-app-token@v1
+        with:
+          app_id: ${{ secrets.BOT_APP_ID }}
+
+          private_key: ${{ secrets.BOT_PRIVATE_KEY }}
+
+
+      - name: Create run condition variable
+        run: |
+          # this is required for early termination in case any conditions aren't satisfied
+          # https://github.com/actions/runner/issues/662
+          
+          "CAN_RUN=1" | Out-File -FilePath $env:GITHUB_ENV -Append
+
+      - name: Login to gh cli
+        run: |
+          "GH_TOKEN=${{ steps.generate.outputs.token }}" | Out-File -FilePath $env:GITHUB_ENV -Append
+
+      - name: Check if PR can be committed to
+        run: |
+          if (!(gh pr -R ${{ github.repository }} view ${{ github.event.issue.number }} --json maintainerCanModify -q '.maintainerCanModify' | ConvertFrom-Json))
+          { 
+            gh pr comment ${{ github.event.issue.number }} -b "This PR cannot be committed to. Ensure that Allow edits from maintainers is enabled."
+            "CAN_RUN=0" | Out-File -FilePath $env:GITHUB_ENV -Append
+          }
+          
+          # all steps after this one must have the env.CAN_RUN == 1 condition
+
+      - name: Set git identity
+        if: env.CAN_RUN == 1
+        run: |
+          $data = gh api graphql -f query='query {
+            viewer {
+              login
+              databaseId
+            }
+          }'  --jq '.data.viewer' | ConvertFrom-Json
+
+          git config --global user.name "$($data.login)"
+          git config --global user.email "$($data.databaseId)+$($data.login)@users.noreply.github.com"
+
+
+      - uses: actions/checkout@v4
+        if: env.CAN_RUN == 1
+
+      - name: Checkout PR
+        if: env.CAN_RUN == 1
+        run: gh pr checkout ${{ github.event.issue.number }} -b pr
+
+      - name: Check if PR has modified XAML files
+        if: env.CAN_RUN == 1
+        run: |
+          $baseRef = gh pr view ${{ github.event.issue.number }} --json baseRefName --jq '.baseRefName'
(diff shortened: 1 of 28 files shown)


src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
```diff
@@ -248,11 +248,13 @@ protected async void RenameTextBox_KeyDown(object sender, KeyRoutedEventArgs e)
 					e.Handled = true;
 					break;
 				case VirtualKey.Up:
-					textBox.SelectionStart = 0;
+					if (textBox.SelectionStart != 0)
```
yaira2, line 251: I think it might be cleaner to check if the `shift` key is pressed. Let me know if you have any questions on how to do this.

RieBi: Yeah, I'm now checking for the shift key

RieBi:
So there were some inconsistent behaviors with how selection and up arrow key and down arrow key worked, for example pressing down arrow key moves pointer to the end, but if pointer was at the start, then it did nothing, which I suppose shouldn't have been the case.

I fixed those things along with the main issue where selection would get canceled after pressing shift+up arrow key multiple times. I hope my changes did not introduce any behaviors that were not meant to be

yaira2:
@RieBi the original issue author reported that the issue is already fixed on their side so it might have just been a transient issue.

RieBi:
> @RieBi the original issue author reported that the issue is already fixed on their side so it might have just been a transient issue.

Well it's kind of interesting because the original issue that the author was talking about is still present in the latest release

RieBi:
@yaira2 maybe I could open my own issue addressing the problem? Because the bug still exists in the latest release and it would have been nice to have it fixed

src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
```diff
@@ -248,23 +250,47 @@ protected async void RenameTextBox_KeyDown(object sender, KeyRoutedEventArgs e)
 					e.Handled = true;
 					break;
 				case VirtualKey.Up:
-					textBox.SelectionStart = 0;
+					if (isShiftPressed)
+					{
```
yaira2, line 254:
```suggestion
					if (!isShiftPressed)
					     textBox.SelectionStart = 0;
```
I think this should be enough

RieBi: Nope it isn't

RieBi:
OK I updated the branch to the latest commit of main. Regarding the logic, I don't think anything can be shortened there without breaking anything (that's concerning @yaira2 's last review)

src/Files.App/Views/LayoutModes/StandardLayoutMode.cs
yaira2: I can do some more testing on my side, but it should be enough to just return early is shift is pressed.

RieBi: Yep you might try. But I have already tried doing such things and it mostly didn't work so I had written full behavior for every case.

yaira2:
@RieBi can you take a look at commit I pushed?

yaira2 (approved):

RieBi:
> @RieBi can you take a look at the commit I pushed?

Yeah. I tested it, and it doesn't work quite as expected. It fixes like 50% of stuff, but the other 50% is still bugged and inconsistent.

Basically it's the same thing that I did first when I was fixing this issue, and it fixes some things, but there is more to it because the base textbox selection behavior is very buggy and weird per se.

RieBi:
@yaira2

yaira2:
What types of issue are still present?

RieBi:
For example, I suppose that intended behavior of pressing down arrow should move the cursor to the end. And it does, except when the cursor is at the very beginning - in that case it does nothing.

Also, shift+down arrow should expand the selection until the end, and sometimes it does, but sometimes not - specifically, when there's already any selection present. 

So as I see it is repeatedly pressing up and down arrows should change the cursor position between beginning and end. But in reality, you can now go to start with up button, you wouldn't be able to go to the end by pressing down arrow, although pressing down arrow in other scenarios would bring the cursor to the end.

So I find it kind of frustrating to work with selection using keyboard as of now due to the mentioned issues.

RieBi:
@yaira2

yaira2:
> For example, I suppose that intended behavior of pressing down arrow should move the cursor to the end. And it does, except when the cursor is at the very beginning - in that case it does nothing.

This appears to be the default behavior for the WinUI text box so I think we can leave it as is.

yaira2:
> Also, shift+down arrow should expand the selection until the end, and sometimes it does, but sometimes not - specifically, when there's already any selection present.

I'll take a look at this

PR #13899, opened 2023-11-15
Fix: Fixed NullReferenceException in UpdateInstancePropertiesAsync

yaira2:
Fixed NullReferenceException in UpdateInstancePropertiesAsync


src/Files.App/ViewModels/MainPageViewModel.cs
```diff
@@ -144,10 +144,8 @@ public async Task UpdateInstancePropertiesAsync(object navigationArg)
 					var rightTabInfo = await GetSelectedTabInfoAsync(paneArgs.RightPaneNavPathParam);
 					windowTitle = $"{leftTabInfo.tabLocationHeader} | {rightTabInfo.tabLocationHeader}";
 				}
-				else
-				{
+				else if (paneArgs.LeftPaneNavPathParam is not null)
```
hishitetsu, line 147: Sorry, I missed your PR and created another one. But I don't think that a null argument to GetSelectedTabInfoAsync will cause an exception.

yaira2: Let's go with your PR

PR #13902, opened 2023-11-16
Fix: Fixed potential crash when restoring from Recycle Bin

yaira2:
**Resolved / Related Issues**
Fixes `System.Runtime.InteropServices.COMException: Error HRESULT E_FAIL has been returned from a call to a COM component`

### DIFF
--- a/src/Files.App/Utils/RecycleBin/RecycleBinHelpers.cs
+++ b/src/Files.App/Utils/RecycleBin/RecycleBinHelpers.cs
@@ -79,7 +79,7 @@ namespace Files.App.Utils.RecycleBin
 
 		public static async Task RestoreRecycleBinAsync()
 		{
-			var ConfirmEmptyBinDialog = new ContentDialog()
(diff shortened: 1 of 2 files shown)


src/Files.App/Utils/RecycleBin/RecycleBinHelpers.cs
```diff
@@ -92,7 +92,10 @@ public static async Task RestoreRecycleBinAsync()
 
 			if (result == ContentDialogResult.Primary)
 			{
-				Vanara.Windows.Shell.RecycleBin.RestoreAll();
+				SafetyExtensions.IgnoreExceptions(() =>
```
hishitetsu, line 95: This should work, but wouldn't it be more helpful to have a dialog if an exception occurs?

yaira2: Added 👍

hishitetsu (approved):
LGTM

PR #13920, opened 2023-11-16
Fix: Fixed IO Error in RunCompatibilityTroubleshooterAsync

yaira2:
**Resolved / Related Issues**
Fixed `System.IO.IOException: The process cannot access the file 'C:\Users\USER\AppData\Local\Temp\CompatibilityTroubleshooterAnswerFile.xml' because it is being used by another process.`

### DIFF
--- a/src/Files.App/Utils/Shell/LaunchHelper.cs
+++ b/src/Files.App/Utils/Shell/LaunchHelper.cs
@@ -34,7 +34,11 @@ namespace Files.App.Utils.Shell
 		public static Task<bool> RunCompatibilityTroubleshooterAsync(string filePath)
 		{
 			var compatibilityTroubleshooterAnswerFile = Path.Combine(Path.GetTempPath(), "CompatibilityTroubleshooterAnswerFile.xml");
-			File.WriteAllText(compatibilityTroubleshooterAnswerFile, string.Format("<?xml version=\"1.0\" encoding=\"UTF-8\"?><Answers Version=\"1.0\"><Interaction ID=\"IT_LaunchMethod\"><Value>CompatTab</Value></Interaction><Interaction ID=\"IT_BrowseForFile\"><Value>{0}</Value></Interaction></Answers>", filePath));
+
+			SafetyExtensions.IgnoreExceptions(() =>
+			{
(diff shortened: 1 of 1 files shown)


src/Files.App/Utils/Shell/LaunchHelper.cs
```diff
@@ -34,7 +34,11 @@ public static Task<bool> LaunchAppAsync(string application, string arguments, st
 		public static Task<bool> RunCompatibilityTroubleshooterAsync(string filePath)
 		{
 			var compatibilityTroubleshooterAnswerFile = Path.Combine(Path.GetTempPath(), "CompatibilityTroubleshooterAnswerFile.xml");
-			File.WriteAllText(compatibilityTroubleshooterAnswerFile, string.Format("<?xml version=\"1.0\" encoding=\"UTF-8\"?><Answers Version=\"1.0\"><Interaction ID=\"IT_LaunchMethod\"><Value>CompatTab</Value></Interaction><Interaction ID=\"IT_BrowseForFile\"><Value>{0}</Value></Interaction></Answers>", filePath));
+
+			SafetyExtensions.IgnoreExceptions(() =>
+			{
+				File.WriteAllText(compatibilityTroubleshooterAnswerFile, string.Format("<?xml version=\"1.0\" encoding=\"UTF-8\"?><Answers Version=\"1.0\"><Interaction ID=\"IT_LaunchMethod\"><Value>CompatTab</Value></Interaction><Interaction ID=\"IT_BrowseForFile\"><Value>{0}</Value></Interaction></Answers>", filePath));
+			});
```
hishitetsu, line 41: Ignoring the exception here would not make the next process work.
I think we should continue the process by trying another file name instead of ignoring the exception.

yaira2: Is this a file that we're generating from Files?

hishitetsu: I believe so and any file name is acceptable.

yaira2: I added a try/catch to try once more with a different file name.

hishitetsu (approved):
LGTM

PR #13921, opened 2023-11-16
Fix: Fixed potential COMException when setting the wallpaper

yaira2:
**Resolved / Related Issues**
- Fixed com exception when setting wallpaper

`System.Runtime.InteropServices.COMException: Retrieving the COM class factory for component with CLSID {C2CF3110-460E-4FC1-B9D0-8A1C0C9CC4BD} failed`

### DIFF
--- a/src/Files.App/Strings/en-US/Resources.resw
+++ b/src/Files.App/Strings/en-US/Resources.resw
@@ -3608,4 +3608,7 @@
   <data name="FailedToRestore" xml:space="preserve">
     <value>Failed to restore items from Recycle Bin</value>
   </data>
+  <data name="FailedToSetBackground" xml:space="preserve">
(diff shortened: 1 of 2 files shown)


hishitetsu (commented):
Wouldn't it be better to modify `SetSlideshow` in the same way?

hishitetsu (commented):
You might want to make the same correction to `LockScreen.SetImageFileAsync` just to be safe.

yaira2:
> You might want to make the same correction to `LockScreen.SetImageFileAsync` just to be safe.

I thought about it, but it's a different API. It might be worth leaving this alone in case there is an issue that needs to be handled differently.

hishitetsu (approved):
LGTM

PR #13959, opened 2023-11-19
Code Quality: Introduced PathBreadcrumbViewModel

0x5bfa:
## Summary

I could not find the reason why PathBreadcrumb had never exist and I found it'd better add it. However, looking into the way of referencing to AddressToolbarViewModel and PathBreadcrumbView is too far from MVVM concept. I think we need to change it in the future.

## Task Checklist

- [x] Introduced PathBreadcrumbViewModel
- [x] De-clutter AddressToolbarViewModel

## Steps To Validate Changes

1. Launch the app
2. Go to Desktop
3. See the path breadcrumb items are changed accordingly and proeprly
4. Add a new tab
5. Change tabs
6. See the path breadcrumb items are changed accordingly and properly

## Known Issues

_N/A_

## PR Checklist

- [ ] **Development**: _N/A_
- [ ] **Co-authors**: _N/A_

## Screenshots

_N/A_


0x5bfa:
> The path items won't be updated after tab item is changed

@hishitetsu Do you have any idea why? I think I cutted a certain trigger.

hishitetsu:
> @hishitetsu Do you have any idea why? I think I cut a certain trigger.

Shouldn't PathBreadcrumbViewModel be ObservableObject?

0x5bfa:
OMG

PR #13961, opened 2023-11-19
Fix: Fixed null reference warnings in navigation helpers

yaira2:
**Resolved / Related Issues**
- Fixed null reference warnings in navigation helpers

### DIFF
--- a/src/Files.App/Helpers/Navigation/NavigationHelpers.cs
+++ b/src/Files.App/Helpers/Navigation/NavigationHelpers.cs
@@ -293,7 +293,7 @@ namespace Files.App.Helpers
 
 			if (isShortcut)
 			{
-				if (string.IsNullOrEmpty(shortcutInfo.TargetPath))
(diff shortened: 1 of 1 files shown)


src/Files.App/Helpers/Navigation/NavigationHelpers.cs
```diff
@@ -293,7 +293,7 @@ private static async Task<FilesystemResult> OpenFile(string path, IShellPage ass
 
 			if (isShortcut)
 			{
-				if (string.IsNullOrEmpty(shortcutInfo.TargetPath))
+				if (string.IsNullOrEmpty(shortcutInfo.TargetPath) && args is not null)
 				{
 					await Win32Helpers.InvokeWin32ComponentAsync(path, associatedInstance, args);
```
hishitetsu, line 298: `InvokeWin32ComponentAsync` actually accepts null, so if you want to get rid of the warning, you should make the args of `InvokeWin32ComponentAsync` null acceptable.

PR #13963, opened 2023-11-19
Fix: Fixed null warnings in the details layout

yaira2:
**Resolved / Related Issues**
- Fixed null warnings in the details layout

### DIFF
--- a/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
+++ b/src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See the LICENSE.
 
 using CommunityToolkit.WinUI.UI;
-using Files.App.Actions;
 using Files.App.UserControls.Selection;
 using Microsoft.UI.Input;
 using Microsoft.UI.Xaml;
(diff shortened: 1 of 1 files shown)


src/Files.App/Views/LayoutModes/DetailsLayoutBrowser.xaml.cs
```diff
@@ -136,7 +135,7 @@ protected override void OnNavigatedTo(NavigationEventArgs eventArgs)
 
 			var parameters = (NavigationArguments)eventArgs.Parameter;
 			if (parameters.IsLayoutSwitch)
-				ReloadItemIconsAsync();
+				await ReloadItemIconsAsync();
```
hishitetsu, line 138: Shouldn't we leave it as fire and forget?

yaira2: Is there any downside to using `await`? If we remove it, should we suppress the warning and add a comment?

hishitetsu: Await will cause the process to pause there, and no further processing will take place until icons are reloaded. If it is working fine now, this modification may lead to performance degradation.
If you just want to get rid of the warning, you can write the following:
```csharp
_ = ReloadItemIconsAsync();
```

yaira2: 👍

hishitetsu (approved):
LGTM

PR #13964, opened 2023-11-19
Fix: Fixed null ref in sidebar pinned model

yaira2:
**Resolved / Related Issues**
- Fixed null ref in sidebar pinned model


src/Files.App/Data/Models/SidebarPinnedModel.cs
```diff
@@ -100,7 +100,7 @@ public async Task<LocationItem> CreateLocationItemFromPathAsync(string path)
 			if (res || (FilesystemResult)FolderHelpers.CheckFolderAccessWithWin32(path))
 			{
 				locationItem.IsInvalid = false;
-				if (res)
+				if (res.Result is not null)
```
hishitetsu, line 103:
```suggestion
				if (res && res.Result is not null)
```

yaira2: It doesn't hurt to have the check but according to VS `res` isn't null here.

PR #13967, opened 2023-11-19
Fix: Fixed crash when calculating window bounds

yaira2:
**Resolved / Related Issues**
- Fixes `System.Runtime.InteropServices.COMException: The operation identifier is not valid. The WinUI Desktop Window object has already been closed`

### DIFF
--- a/src/Files.App/Views/PaneHolderPage.xaml.cs
+++ b/src/Files.App/Views/PaneHolderPage.xaml.cs
@@ -1,7 +1,6 @@
 // Copyright (c) 2023 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using Files.App.UserControls.TabBar;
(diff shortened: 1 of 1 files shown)


src/Files.App/Views/PaneHolderPage.xaml.cs
```diff
@@ -80,7 +79,15 @@ public bool IsMultiPaneActive
 			=> IsRightPaneVisible;
 
 		public bool IsMultiPaneEnabled
-			=> MainWindow.Instance.Bounds.Width > DualPaneWidthThreshold;
+		{
+			get
+			{
+				if (App.AppModel.IsMainWindowClosed == true)
```
hishitetsu, line 85:
```suggestion
				if (App.AppModel.IsMainWindowClosed)
```

PR #13971, opened 2023-11-19
Fix: Fixed check for admin

### LINKED ISSUE #13394: Bug: App crashes when dragging a file while app is running with a high integrity level
### Description

Crashed after opening app and draging a file. No other actions. App is not running as admin, and is not set as the default file manager. Sometimes it happened to work well with no crash for drag & drop but I am not sure about the reason.

- [ ] 

### Steps To Reproduce

1. open Files with a high integrity level
2. drag a file or tab
3. the app crashes

### Requirements

- Detect and disable drag & drop on all trust levels that don't support it

### Files Version

2.5.29.0

### Windows Version

10.0.19045.3448

### Log File

[debug.log](https://github.com/files-community/Files/files/12653254/debug.log)

yaira2:
**Resolved / Related Issues**
   For #13394...

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Helpers/Environment/ElevationHelpers.cs
+++ b/src/Files.App/Helpers/Environment/ElevationHelpers.cs
@@ -21,8 +21,8 @@ namespace Files.App.Helpers
 
 		public static bool IsAppRunAsAdmin()
 		{
-			using WindowsIdentity identity = WindowsIdentity.GetCurrent();
-			return new WindowsPrincipal(identity).IsInRole(WindowsBuiltInRole.Administrator);
+			WindowsIdentity identity = WindowsIdentity.GetCurrent();
+			return new WindowsPrincipal(identity).IsInRole(new SecurityIdentifier(WellKnownSidType.BuiltinAdministratorsSid, null));
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


src/Files.App/Helpers/Environment/ElevationHelpers.cs
```diff
@@ -21,8 +21,8 @@ public static bool IsElevationRequired(string filePath)
 
 		public static bool IsAppRunAsAdmin()
 		{
-			using WindowsIdentity identity = WindowsIdentity.GetCurrent();
-			return new WindowsPrincipal(identity).IsInRole(WindowsBuiltInRole.Administrator);
+			WindowsIdentity identity = WindowsIdentity.GetCurrent();
```
hishitetsu, line 24: Is there no need for `using`?

yaira2: It works without it but I'm not sure if it makes a difference.

yaira2: Anyways, I think we need some more testing to see if this change unblocks users that have UAC turned off.

hishitetsu: I believe it is safer to leave the `using`. But I can't determine whether the modification on line 25 is correct or not.

PR #13973, opened 2023-11-19
Fix: Fixed null ref warnings in BaseShellPage

yaira2:
**Resolved / Related Issues**
- Fixed null ref warnings in BaseShellPage


hishitetsu (approved):
LGTM

PR #13975, opened 2023-11-20
Fix: Fixed missing awaits

yaira2:
**Resolved / Related Issues**
- Fixed missing awaits


hishitetsu:
It is necessary to examine carefully whether `await` should be added or it should remain fire and forget.

yaira2:
Very true, moving forward we should add a comment for future reference as it can be hard to determine at a glance.

src/Files.App/ViewModels/Properties/Items/FolderProperties.cs
yaira2, line 1: Should I use discard for `GetOtherPropertiesAsync`?

Issue #9895, opened 2022-09-04
Feature: Nullable reference types annotations


hez2010:
We are migrating to WASDK and adopt .NET 6, so we get build-in nullable reference types support.

**Requirements**
- We should annotate our code to eliminate thousands of warnings about NRT. 

This is a large one, but we can do it progressively.

yaichenbaum:
Good idea, I believe we can do this one by one (e.g. update things as you notice them).

yaira2:
> We should annotate our code to eliminate thousands of warnings about NRT.

What's the proper way to resolve these warnings? Making properties nullable solves the warnings but doesn't offer any protection.

hecksmosis:
According to stackoverflow and the Microsoft docs, the way to eliminate the warnings with "possible" is to mark the properties as nullable or to check for null values

chingucoding:
My 2 cents: If something can be null, we should make it nullable and add null checks where they are used. If the property can never be null (for reasons that the compiler cannot know about), then disabling the warning for that property with an explanation why is the best way.

yaira2:
I think that's a fair solution. That being said, these warnings should be fixed as needed because otherwise, opening a pull request just to fix the warnings takes away from other review efforts.

chingucoding:
So opening a PR just to fix these warnings is not encouraged?

hecksmosis:
I think not

yaira2:
> So opening a PR just to fix these warnings is not encouraged?

It can also impact stability to make large changes like this without understanding the context.

yaira2:
Just to clarify my previous comments, opening PRs for this that are limited in scope (eg one or two classes) are fine.

0x5bfa:
Removed null-ref warnings in all Widgets classes in the widget view models introduction PR.

0x5bfa:
As this might have to be done gradually and not in Separated PRs, can’t we merge this to #4180?

PR #13985, opened 2023-11-20
Fix: Display prompt when trying to delete folder that's too big for recycle bin 

### LINKED ISSUE #8358: Feature: Display prompt when trying to delete folder that's too big for recycle bin
### Description

When you want to delete a file too big for the trash, even if you say yes to delete forever there is an error and the file isn't deleted. (sry for the bad english I'm french)

### Steps To Reproduce

1. Delete a file too big for trash
2. Say yes to delete forever
3. An error occured and the files is still alive

### Expected behavior
- Auto select permanently delete check box when there isn't enough space in recycle bin
- Disable the check box

### Files Version

Version : 2.1.13.0 OS Version: 10.0.19044.1526

### Windows Version

Windows Feature Experience Pack 120.2212.4170.0

### Relevant Assets/Logs

image
44/153939382-290454dc-bc57-432e-bc9c-82366f385975.png)

denisdekh:
**Resolved / Related Issues**
-  Closes #8358 
Implemented the siggestion of @BanbanProduction to check the size of the folder being deleted (issue only seemed to occur with folders). The solution involves forcing a permament deletion if the folder might be too big for the bin and may cause the error.
This solution is not very pretty for a couple reasons:
- Some overhead to compute the size of the folder (although it seems low - see Validation)
- Uses the registry to get the size of the bin, which may hit or miss on different machines
- Only guesses if the folder is too big, going by the smallest of the recycle bins on the drives (I know of no way to discern them in the registry)
- As a consequence of the previous point can force permanent deletion for big folders unnecessarily to recycle bin size differences

As alternate fixes I can think of adding a prompt for the user to permanently delete files manually in the error message, checking the box to permanently delete for the user, but allowing them to uncheck to account for incorrect guesses, or keeping the same implementation with a better way to query the maximum recycle bin size (which I could not find anywhere).

- Adresses #9895
Also cleared some nullable reference type warnings in the related files. Those are in separate commits so can be merged regardless of the rest of pull request

**Validation**
- Confirmed that the issue exists on main 
- Verified that the new behavior only affects folders
- Doesn't slow the program down noticably even when computing the size of a folder with 100 files

### DIFF
--- a/src/Files.App/Views/HomePage.xaml.cs
+++ b/src/Files.App/Views/HomePage.xaml.cs
@@ -1,12 +1,12 @@
 // Copyright (c) 2023 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using Files.App.Dialogs;
 using Files.App.UserControls.Widgets;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Navigation;
 using System.IO;
 using System.Runtime.InteropServices;
+using Windows.Foundation.Metadata;
 using Windows.Storage;
 
 namespace Files.App.Views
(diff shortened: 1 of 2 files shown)


yaira2:
> Some overhead to compute the size of the folder

I think this is fine as long the check happens after failing to delete the items.

yaira2:
> As a consequence of the previous point can force permanent deletion for big folders unnecessarily to recycle bin size differences

We should avoid doing this without confirmation from the user, the main thing is to display a proper error and allow the user to move forward from there.

denisdekh:
> > As a consequence of the previous point can force permanent deletion for big folders unnecessarily to recycle bin size differences
> 
> We should avoid doing this without confirmation from the user, the main thing is to display a proper error and allow the user to move forward from there.

Sounds good, I will look into having a proper message for the user. This issue is tricky because the failure seems to happen in OS code, so resolving the issue directly is challenging. I will try to identify the error somehow and add a specific error message similar to "folder could not be moved to the recycling bin".

yaira2:
It might be a bit tricky to intercept the error.

PR #14014, opened 2023-11-21
Feature: Display error when failing to calculate the hash of open files

### LINKED ISSUE #13564: Bug: Hash computation fails when office documents are opened
### Description

image

If a Microsoft Office document (e.g. docx or pptx) is opened, Files will not be able to compute its hash.

### Steps To Reproduce

1. Create a MS Word document (empty, 0 KB is fine)
2. Open the document
3. Calculate the Hash

### Requirements

There should not be an error in debug.log and a more user-friendly message should be shown in the Hash UI, instead of a cryptic error.

### Files Version

2.5.21.0

### Windows Version

10.0.19045.3570

### Log File

[debug.log](https://github.com/files-community/Files/files/13110939/debug.log)

yaira2:
**Resolved / Related Issues**
   Closes #13564...

**Validation**
   1. Open office document
   2. Right click and open properties for that file
   3. Switch to the hashes page and confirm it explains why the hashes can't be calculated

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/ViewModels/Properties/HashesViewModel.cs
+++ b/src/Files.App/ViewModels/Properties/HashesViewModel.cs
@@ -103,6 +103,11 @@ namespace Files.App.ViewModels.Properties
 					{
 						// not an error
 					}
+					catch (IOException)
+					{
+						// File is currently open
+						hashInfoItem.HashValue = "CalculationErrorFileIsOpen".GetLocalizedResource();
+					}
 					catch (Exception)
 					{
 						hashInfoItem.HashValue = "CalculationError".GetLocalizedResource();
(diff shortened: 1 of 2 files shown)


src/Files.App/ViewModels/Properties/HashesViewModel.cs
```diff
@@ -103,6 +103,11 @@ private void ToggleIsEnabled(string? algorithm)
 					{
 						// not an error
 					}
+					catch (IOException)
+					{
+						// File is currently open
+						hashInfoItem.HashValue = "CalculationErrorFileIsOpen".GetLocalizedResource();
+					}
```
hishitetsu, line 110: This would also catch the exception if the file had already been deleted.

yaira2: I think that's an edge case, but there are ways to handle it if users start opening issues regarding this.

hishitetsu (approved):
LGTM

PR #14067, opened 2023-11-25
Fix: Fixed an issue with dropping multiple items onto executable files

### LINKED ISSUE #14033: Bug: Drag-and-Drop Multiple Files Opens Multiple Instances Instead of Passing as Arguments for .py and .bat Files
### Description

When multiple files are dragged and dropped onto Python (.py) or batch (.bat) files in the app, the current behavior incorrectly opens multiple instances of the respective script, one for each file. Instead, it should open a single instance of the script and pass the paths of all selected files as arguments. This issue affects the usability and efficiency of script execution via drag-and-drop.

### Steps To Reproduce

1. Create a Python (.py) or batch (.bat) file that accepts file paths as arguments.
2. Select multiple files in the Files application.
3. Drag and drop these files onto the Python or batch file.

The application opens a new instance of the Python or batch script for each file dragged onto it, resulting in multiple script instances running separately.

### Requirements

The application should open only one instance of the script (Python or batch).
All dragged file paths should be passed to this instance.
The script should be able to process multiple arguments, with each argument representing the path of a dragged file.

This functionality is expected to align with typical command-line behavior where scripts can receive multiple arguments.

### Files Version

3.0.10.0

### Windows Version

Windows 11 Pro Version	23H2 22631.2715

### Log File

[debug.log](https://github.com/files-community/Files/files/13448740/debug.log)

antonioseet:
…also checks to see if python is installed on system before allowing the drop like in File Explorer.
*removed redundant code*

**Resolved / Related Issues**
   Closes #14033 

**Validation**
   1. Open app ...
   2. Create 4 types of files: test.py, test.bat, test.cmd, test.exe. Each created to test incoming arguments list.
image
image
image
image

Tested dragging and dropping multiple files to each of them to verify changes:
image

Also validated that when python is not installed, the user cannot drop files onto the python file.

### DIFF
--- a/src/Files.App/Helpers/Environment/SoftwareHelpers.cs
+++ b/src/Files.App/Helpers/Environment/SoftwareHelpers.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See the LICENSE.
 
 using Microsoft.Win32;
+using System.IO;
 
 namespace Files.App.Helpers
 {
@@ -28,6 +29,32 @@ namespace Files.App.Helpers
 			key.Close();
 
 			return true;
+		}
+
+		public static bool IsPythonInstalled()
+		{
+			try
+			{
+				ProcessStartInfo psi = new ProcessStartInfo();
+				psi.FileName = "python";
+				psi.Arguments = "--version";
+				psi.RedirectStandardOutput = true;
+				psi.UseShellExecute = false;
+				psi.CreateNoWindow = true;
+
+				using (Process process = Process.Start(psi))
+				{
+					using (StreamReader reader = process.StandardOutput)
+					{
+						string result = reader.ReadToEnd();
+						return result.Contains("Python");
+					}
+				}
+			}
+			catch
+			{
+				return false;
+			}
 		}
 
 		private static bool ContainsName(RegistryKey? key, string find)
--- a/src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
+++ b/src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
@@ -255,16 +255,10 @@ namespace Files.App.Utils.Storage
 				else if (operation.HasFlag(DataPackageOperation.Link))
 				{
 					// Open with piggybacks off of the link operation, since there isn't one for it
-					if (isTargetExecutable)
+					if (isTargetExecutable || isTargetPythonFile)
 					{
 						var items = await GetDraggedStorageItems(packageView);
 						NavigationHelpers.OpenItemsWithExecutableAsync(associatedInstance, items, destination);
-						return ReturnResult.Success;
-					}
-					else if (isTargetPythonFile)
-					{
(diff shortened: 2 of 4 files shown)


src/Files.App/Helpers/Environment/SoftwareHelpers.cs
```diff
@@ -30,6 +31,32 @@ public static bool IsVSInstalled()
 			return true;
 		}
 
+		public static bool IsPythonInstalled()
```
yaira2, line 34: What happens when dragging a file if we don't have this check?

antonioseet: The main scenario I considered is the following:

User **does not** have Python installed and this check **does not** exist.

This check exists to prevent a user without Python installed on their machine from seeing this prompt:
image

WITHOUT checking that python is installed, a user will still see that prompt, and may cause confusion.
If this check is removed, differnt safeguards will need to be implemented to disallow the ```OpenItemsWithExecutableAsync(...)``` from being called with python files.

That being said, people with python files, will most likely already have python installed, so it probably is not an issue in the grand scheme of things.

yaira2: The question is if it's worth checking every time the user drags a file, it might be better to just show the tooltip and have nothing happen.

antonioseet: Yes, that is understandable.
The other option is to add the check after the drop happens to prevent the call to Win32Helpers.InvokeWin32ComponentAsync(...), that way we avoid the performance hit for every drag and it only happens after the drop onto a python file.

yaira2: That would be good, we could also cache the result but I don't think we need to go that far.

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -255,18 +255,14 @@ public async Task<ReturnResult> RestoreItemsFromTrashAsync(IEnumerable<IStorageI
 				else if (operation.HasFlag(DataPackageOperation.Link))
 				{
 					// Open with piggybacks off of the link operation, since there isn't one for it
-					if (isTargetExecutable)
+					if (isTargetExecutable || isTargetPythonFile)
 					{
 						var items = await GetDraggedStorageItems(packageView);
+						if (isTargetPythonFile && SoftwareHelpers.IsPythonInstalled())
```
yaira2, line 261:
```suggestion
						if (isTargetPythonFile && SoftwareHelpers.IsPythonInstalled())
```
Shouldn't this be `!SoftwareHelpers.IsPythonInstalled`?

antonioseet: Yes, fixing now

yaira2 (approved):
LGTM

PR #13490, opened 2023-10-08
Code Quality: Fixed issues with the properties window components

0x5bfa:
## Summary

During removing SettingsViewModel, I had to alter ThemeHelper since that ViewModel contained theme change event handler. And while I changing ThemeHelper, I noticed there were several visual issues therefore I fixed in this PR.

## Task Checklist

- [x] Removed the duplicated SettingsViewModel
- [x] Fixed an issue where the action buttons of the properties window were invisible
- [x] Fixed an issuw where the window icon was not set properly
- [x] Clean up
- [ ] Apply requested changes

## Known Issues

Following issues would be fixed in the upcoming PR.

_N/A_

## PR Checklist

- [x] **Related**: #4180
- [ ] **Approval**: Have discussed with and got approval from the team[^1]
- [x] **Tests**: Tested accessibility on the local end
- [x] **Deployment**: Deployed the app on the local end
- [ ] **Localization**: Modified en-US string resources[^2]
- [ ] **Co-authors**: _N/A_

## Steps To Validate Changes

  - Open the app
  - Open properties window
  - See x button visibility on the active window
  - And hover the Files icon on the TaskBar
  - See the icon is set properly

## Screenshots

image

image

[^1]: If the request hasn't been agreed by the team, this work might be rejected. Make sure that you get approval from the team before opening any PR related the request.
[^2]: If you removed any en-US string resources, make sure that there are no references of those resources. When you add a new en-US string resources, do not make any changes on other languages' resources; [Crowdin](https://crowdin.com/project/files-app) will do, instead.

PR #14068, opened 2023-11-25
Code Quality: Introduced AppThemeModeService

0x5bfa:
## Summary

This is one of the efforts to migrate static Helpers to singleton Services.

## Task Checklist

- [x] Introduced `AppThemeModeService` : `IAppThemeModeService`
- [x] Removed SettingsViewModel, which had theme changed event handler and has been marked as obslete.

## Steps To Validate Changes

1. Launch the app
2. See the sections you have expanded are expanded expectedly
3. Go to Settings dialog
4. Change theme mode
5. See theme will be changed both in main app window and dialog

## Known Issues

_N/A_

## PR Checklist

- [x] **Development**: Related: #13490
- [ ] **Co-authors**: _N/A_

## Screenshots

_N/A_


0x5bfa:
I executed tests on my end and it reported as 3 succeeded, 0 failed.
I couldn't figure out why the tests have being failed.

src/Files.Core/Services/IAppThemeModeService.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) 2023 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.Core.Services
+{
+	public interface IAppThemeModeService
```
yaira2, line 6: Do you think we can simplify this to `IThemeService`?

src/Files.App/ViewModels/SettingsViewModel.cs
yaira2, line 1: This is a nice milestone!

0x5bfa: I searched references of LocalSettings namespace, which is deprecated on Files codebase, and I think all of them can be moved to our own json settings scheme. That's out of scope of this PR though.

src/Files.Core/Services/IAppThemeModeService.cs
0x5bfa: There's already AppThemeBackground and App would be necessary for such application, which have a lot of features to differentiate others.

image

0x5bfa: This is still on planning stage but I'm planning rename ResourcesService, JumpListService, UpdateService to have App prefix, move into Files.App.Services.Application.*.

image

This is also on planning stage but I'm planning to rename those 2 helpers (DragZoneHelper & JumpListHelper) to have App prefix to be consistent
image

I have some outlook for the future of improvements on Files codebase, so I'd like to keep as it is.

src/Files.App/Dialogs/PropertiesDialog.xaml
yaira2, line 1: Why was this removed?

src/Files.Core/Data/Enums/AppThemeMode.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) 2023 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.Core.Data.Enums
+{
+	public enum AppThemeMode
+	{
+		Default,
```
yaira2, line 8: Should this be "System"?

src/Files.Core/Services/IAppThemeModeService.cs
yaira2, line 1: Please add comments

src/Files.App/Dialogs/PropertiesDialog.xaml
0x5bfa: This was used from nowhere and was a leftover from UWP. Should I remove in another PR?

src/Files.Core/Data/Enums/AppThemeMode.cs
0x5bfa: I don't have preferences. Would like to rename to System?

yaira2: As long as we keep the existing user settings

PR #14144, opened 2023-12-03
Feature: Open folder in new tab when dragging folder to "add tab" button

denisdekh:
**Resolved / Related Issues**
- [x] Were these changes approved in an issue or discussion with the project maintainers? 
   Closes #4710 

Also added support for drag opening archives, but commented it out for now as the best solution I was able to find was too ugly in my opinion (parsing the extention from the filename and checking it agaist a list of zip extentions)

**Validation**
How did you test these changes?
- [x] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [ ] Did you implement any design changes to an existing feature?
   - [ ] Was this change approved?
- [x] Are there any other steps that were used to validate these changes?
1. Ran the app
2. Tested dragging a single folder / multiple folders
3. Works even if non-folder items are selected and dragged (those are ignored)
4. Stress tested to some extend with dragging 50+ folders at once

**Screenshots (optional)**
Too hard to take a screenshot while dragging lol

PR #14145, opened 2023-12-03
Code Quality: Fixed null warnings in MainPage.xaml.cs

denisdekh:
**Resolved / Related Issues**
- [ ] Were these changes approved in an issue or discussion with the project maintainers? -sort of
   Addresses #9895 

Fixed some NRT warnings in MainPage.xaml.cs while working on #14144 

**Validation**
   

**Screenshots (optional)**
Add screenshots here.


src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -198,23 +203,23 @@ private void PaneHolder_PropertyChanged(object? sender, PropertyChangedEventArgs
...
 		{
 			if (StatusBarControl is not null)
 			{
 				StatusBarControl.DirectoryPropertiesViewModel = SidebarAdaptiveViewModel.PaneHolder?.ActivePaneOrColumn.SlimContentPage?.DirectoryPropertiesViewModel;
-				StatusBarControl.SelectedItemsPropertiesViewModel = SidebarAdaptiveViewModel.PaneHolder?.ActivePaneOrColumn.SlimContentPage?.SelectedItemsPropertiesViewModel;
+				StatusBarControl.SelectedItemsPropertiesViewModel = SidebarAdaptiveViewModel.PaneHolder!.ActivePaneOrColumn.SlimContentPage.SelectedItemsPropertiesViewModel; 
 			}
 		}
-
+		// SidebarAdaptiveViewModel.PaneHolder must not be null
```
0x5bfa, line 215: Add a new line before this line.

According to the guideline Section 3, Item 4.
> Insert a blank line after closing curly bracket.

0x5bfa (commented):
Overall lgtm.

denisdekh:
Fixed the styling issues, thanks for feedback!

denisdekh:
It failed again on the tests with this message:
"Windows Application Driver listening for requests at: http://127.0.0.1:4723/
Address 'http://127.0.0.1:4723/' is already in use
Failed to initialize: 0x80004005"
Is it an issue with the CI pipeline?

0x5bfa:
Did you launched with this branch on your end?

denisdekh:
Should be all good now, made a mistake with one of the fixes. Compiler thought it was null-safe, but, well, it wasn't. I instead made the field nullable since it was being assigned null already

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -175,9 +177,15 @@ public void MultitaskingControl_CurrentInstanceChanged(object? sender, CurrentIn
 				SidebarAdaptiveViewModel.PaneHolder.PropertyChanged -= PaneHolder_PropertyChanged;
 
 			var navArgs = e.CurrentInstance.TabItemParameter?.NavigationParameter;
-			SidebarAdaptiveViewModel.PaneHolder = e.CurrentInstance as IPaneHolder;
+			var currentInstance = e.CurrentInstance as IPaneHolder;
+			if (currentInstance is null || navArgs is null) 
+   				return;
```
0x5bfa, line 182: You can use pattern matching.

```suggestion
			if (e.CurrentInstance is not IPaneHolder currentInstance || navArgs is null) 
   				return;
```

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -175,9 +177,15 @@ public void MultitaskingControl_CurrentInstanceChanged(object? sender, CurrentIn
...
+			SidebarAdaptiveViewModel.PaneHolder = currentInstance;
 			SidebarAdaptiveViewModel.PaneHolder.PropertyChanged += PaneHolder_PropertyChanged;
-			SidebarAdaptiveViewModel.NotifyInstanceRelatedPropertiesChanged((navArgs as PaneNavigationArguments).LeftPaneNavPathParam);
+			var pageNav = navArgs as PaneNavigationArguments;
+			if (pageNav is null) 
+   				return;
```
0x5bfa, line 187: As above

```suggestion
			if (navArgs is not PaneNavigationArguments pageNav)
   				return;
```

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -198,16 +206,16 @@ private void PaneHolder_PropertyChanged(object? sender, PropertyChangedEventArgs
 			UpdateNavToolbarProperties();
 			LoadPaneChanged();
 		}
-
+
```
0x5bfa, line 209: Please remove.

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -198,16 +206,16 @@ private void PaneHolder_PropertyChanged(object? sender, PropertyChangedEventArgs
...
 		{
 			if (StatusBarControl is not null)
 			{
 				StatusBarControl.DirectoryPropertiesViewModel = SidebarAdaptiveViewModel.PaneHolder?.ActivePaneOrColumn.SlimContentPage?.DirectoryPropertiesViewModel;
-				StatusBarControl.SelectedItemsPropertiesViewModel = SidebarAdaptiveViewModel.PaneHolder?.ActivePaneOrColumn.SlimContentPage?.SelectedItemsPropertiesViewModel;
+				StatusBarControl.SelectedItemsPropertiesViewModel = SidebarAdaptiveViewModel.PaneHolder?.ActivePaneOrColumn.SlimContentPage?.SelectedItemsPropertiesViewModel; 
 			}
 		}
-
+
```
0x5bfa, line 218: As above.

0x5bfa (commented):
Other parts LTGM. Thank you so much!

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -204,7 +210,7 @@ private void UpdateStatusBarProperties()
 			if (StatusBarControl is not null)
 			{
 				StatusBarControl.DirectoryPropertiesViewModel = SidebarAdaptiveViewModel.PaneHolder?.ActivePaneOrColumn.SlimContentPage?.DirectoryPropertiesViewModel;
-				StatusBarControl.SelectedItemsPropertiesViewModel = SidebarAdaptiveViewModel.PaneHolder?.ActivePaneOrColumn.SlimContentPage?.SelectedItemsPropertiesViewModel;
+				StatusBarControl.SelectedItemsPropertiesViewModel = SidebarAdaptiveViewModel.PaneHolder?.ActivePaneOrColumn.SlimContentPage?.SelectedItemsPropertiesViewModel;
```
hishitetsu, line 213: This doesn't make sense, just an extra space at the end of the line.

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -175,9 +177,13 @@ public void MultitaskingControl_CurrentInstanceChanged(object? sender, CurrentIn
 				SidebarAdaptiveViewModel.PaneHolder.PropertyChanged -= PaneHolder_PropertyChanged;
 
 			var navArgs = e.CurrentInstance.TabItemParameter?.NavigationParameter;
-			SidebarAdaptiveViewModel.PaneHolder = e.CurrentInstance as IPaneHolder;
+			if (e.CurrentInstance is not IPaneHolder currentInstance || navArgs is null) 
+   				return;
+			SidebarAdaptiveViewModel.PaneHolder = currentInstance;
 			SidebarAdaptiveViewModel.PaneHolder.PropertyChanged += PaneHolder_PropertyChanged;
```
hishitetsu, line 183: Shouldn't we just skip subscribing the event instead of exiting the method?

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -175,9 +177,13 @@ public void MultitaskingControl_CurrentInstanceChanged(object? sender, CurrentIn
...
-			SidebarAdaptiveViewModel.NotifyInstanceRelatedPropertiesChanged((navArgs as PaneNavigationArguments).LeftPaneNavPathParam);
+			if (navArgs is not PaneNavigationArguments pageNav)
+   				return;
+			SidebarAdaptiveViewModel.NotifyInstanceRelatedPropertiesChanged(pageNav.LeftPaneNavPathParam);
```
hishitetsu, line 186:
```suggestion
			SidebarAdaptiveViewModel.NotifyInstanceRelatedPropertiesChanged((navArgs as PaneNavigationArguments)?.LeftPaneNavPathParam);
```
Here, too, it is better to let the method continue as much as possible than to exit it.

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -869,7 +870,9 @@ public static bool ContainsRestrictedFileName(string input)
 		public void Dispose()
 		{
 			filesystemOperations?.Dispose();
-
+   
+	 // SUPPRESS: Cannot convert null literal to non-nullable reference type.
+	#pragma warning disable CS8625
 			associatedInstance = null;
 			filesystemOperations = null;
```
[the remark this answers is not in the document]
denisdekh: They are assigned values in the constructor and shouldn't ever be null, and are only assigned null when the object is Disposed, so it won't be in use anymore from what I understand. So the values are only null when the object is no longer used, so in this case it made sense to me to supress the warning since the compiler doesn't have that context. If the two are marked as nullable then that will probably cascade into many more warning in the file, even though those fields shouldn't be null. If you think that this isn't a good solution, we can come up with something better

src/Files.App/Views/MainPage.xaml.cs
denisdekh: Good point, the argument is nullable

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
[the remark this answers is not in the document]
hishitetsu: Okay, then I don't think the null assignment is necessary, since references to the disposed object should be immediately lost, so references to these objects should also be lost without explicitly nulling them here.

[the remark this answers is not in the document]
denisdekh: The documentation on how Dispose works in C# is a bit weighty, so I am not sure if this is the most ideal implementation. However, from what I understand, the assignment of null signals to the Garbage Collector that the object is safe to remove from memory and seems to be standard practice in the application. This may not be the case but I haven't done enought research to be certain enough be make that change.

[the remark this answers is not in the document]
hishitetsu: I still think the null assignment is unnecessary, but the purpose of this PR is to remove the warnings, so if you think it's better to leave it as is, then leave it as is.

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -159,9 +159,9 @@ public async void TabItemContent_ContentChanged(object? sender, CustomTabViewIte
 			if (SidebarAdaptiveViewModel.PaneHolder is null)
 				return;
 
-			var paneArgs = e.NavigationParameter as PaneNavigationArguments;
+			PaneNavigationArguments? paneArgs = e.NavigationParameter as PaneNavigationArguments;
```
hishitetsu, line 162:
```suggestion
			var paneArgs = e.NavigationParameter as PaneNavigationArguments;
```
There is no need to specify the type explicitly here.

hishitetsu (commented):
One last point.

hishitetsu (approved):
LGTM

PR #14171, opened 2023-12-04
Feature: Recognize Tilde as %USERPROFILE%

### LINKED ISSUE #14166: Feature: Recognize Tilde as %USERPROFILE%

The `~` (tilde) character is interpreted by *NIX shells as the current logged in user's home directory (HOME env on *NIX). On Windows, `HOME` (afaik) is `USERPROFILE`. Given the popularity of UNIX/LINUX, this feature may be highly desirable from a user perspective.

### Requirements

- This proposal will accomplish *NIX like tilde/`~` expansion, as defined:
  - The feature only applies to the **first** path component in a sequence of path components
  - The feature is functionally defined as a substitution of `~` to %USERPROFILE%
  - The feature shall then forward this substitution to the underlying API

### Files Version

3.0.15.0

### Windows Version

10.0.22631 Build 22631

### Comments

_No response_

awderh:
**Resolved / Related Issues**
   Closes #14166 

**Validation**
   1. Open app
   2. Type the following cases into the toolbar: `~`, `~/`, `~\`
   3. They should all reference the same user profile folder. The forward/backslash cases should show files inside

**Screenshots (optional)**

Standalone-tilde support:
image

Forward-slash with tilde support:
image

Example with sub-directories:
image

(Technically, this PR does two things, but it felt awkward without also supporting forward-slash.)

### DIFF
--- a/src/Files.App/Utils/Storage/Helpers/StorageFileExtensions.cs
+++ b/src/Files.App/Utils/Storage/Helpers/StorageFileExtensions.cs
@@ -302,7 +302,7 @@ namespace Files.App.Utils.Storage
 
 		private static string GetPathWithoutEnvironmentVariable(string path)
 		{
-			if (path.StartsWith("~\\", StringComparison.Ordinal))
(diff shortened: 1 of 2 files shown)


yaira2 (approved):
LGTM

PR #14186, opened 2023-12-06
Fix: Fixed crash when navigating up in the column layout

### LINKED ISSUE #14146: Bug: Hitting Up arrow in the navigation bar causes the application to crash
### Description

When the user is in the **column** view,  if you navigate up a folder by hitting the arrow up icon in the navigation bar, when it reaches the last folder and if the user hit the arrow one more time, the application will crash.

image

### Steps To Reproduce

1 - Switch the view style to the column view
2 - Start navigating into the folders
3 - Hit the up arrow in the navigation bar to move back one level up
4 - Keep clicking on the up arrow button until you reach the highest level of navigation
5 - application will crash

### Requirements

The expected behavior is, that when the user reaches the highest level, it should move to the Home directory and the up arrow button should be grayed out or disabled.

### Files Version

3.0.8.0

### Windows Version

10.0.22631 Build 22631

### Log File

[debug.log](https://github.com/files-community/Files/files/13539540/debug.log)

hishitetsu:
**Resolved / Related Issues**
   Closes #14146 

**Validation**

### DIFF
--- a/src/Files.App/Views/Layouts/ColumnsLayoutPage.xaml.cs
+++ b/src/Files.App/Views/Layouts/ColumnsLayoutPage.xaml.cs
@@ -289,12 +289,18 @@ namespace Files.App.Views.Layouts
 			(ParentShellPageInstance as ModernShellPage)?.Forward_Click();
 		}
 
-		public async void NavigateUp()
+		public void NavigateUp()
 		{
 			if (ColumnHost.ActiveBlades?.Count > 1)
 				DismissOtherBlades(ColumnHost.ActiveBlades[ColumnHost.ActiveBlades.Count - 2]);
 			else
-				await Commands.NavigateUp.ExecuteAsync();
+			{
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/Layouts/ColumnsLayoutPage.xaml.cs
```diff
@@ -289,12 +289,18 @@ public void NavigateForward()
 			(ParentShellPageInstance as ModernShellPage)?.Forward_Click();
 		}
 
-		public async void NavigateUp()
+		public void NavigateUp()
 		{
 			if (ColumnHost.ActiveBlades?.Count > 1)
 				DismissOtherBlades(ColumnHost.ActiveBlades[ColumnHost.ActiveBlades.Count - 2]);
 			else
-				await Commands.NavigateUp.ExecuteAsync();
```
yaira2, line 297: Does the command need to be updated as well?

hishitetsu: I don't think so. The command eventually calls this `NavigateUp()` method. In other words, there would be an infinite recursive calls without this fix.

yaira2 (approved):
LGTM

PR #14190, opened 2023-12-06
Fix: Fixed crash when creating taskbar progress

yaira2:
**Resolved / Related Issues**
```
Win32API.CreateTaskbarObject ()
System.IO.FileNotFoundException: Retrieving the COM class factory for component with CLSID {56FDF344-FD6D-11D0-958A-006097C9A090} failed due to the following error: 8007007e
```

### DIFF
--- a/src/Files.App/Utils/Shell/Win32API.cs
+++ b/src/Files.App/Utils/Shell/Win32API.cs
@@ -574,7 +574,7 @@ namespace Files.App.Utils.Shell
 				taskbar2.HrInit();
 				return taskbar2 as Shell32.ITaskbarList4;
 			}
-			catch (NotImplementedException)
+			catch (Exception)
 			{
 				// explorer.exe is not running as a shell
 				return null;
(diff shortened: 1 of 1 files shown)


src/Files.App/Utils/Shell/Win32API.cs
```diff
@@ -568,18 +568,17 @@ public static Task<bool> MountVhdDisk(string vhdPath)
...
 		public static Shell32.ITaskbarList4? CreateTaskbarObject()
 		{
-			var taskbar2 = new Shell32.ITaskbarList2();
 			try
 			{
+				var taskbar2 = new Shell32.ITaskbarList2();
 				taskbar2.HrInit();
+				return taskbar2 as Shell32.ITaskbarList4;
 			}
 			catch (NotImplementedException)
```
hishitetsu, line 577:
```suggestion
			catch (Exception)
```
FileNotFoundException should also be able to be caught.

hishitetsu (approved):
LGTM

PR #14201, opened 2023-12-07
Feature: Added `Ctrl` + `Alt` + `V` shortcut to duplicate selected items

### LINKED ISSUE #14135: Add Files Duplicate command

Please consider adding a Files **Duplicate** command to duplicate a file in one step.

For instance, in macOS FInder, you can right-click a file and choose **Duplicate** to quickly create a duplicate "filename copy.ext" file. Or you can choose **File > Duplicate** with one or more files selected to duplicate them.

This is faster and easier than having to copy and paste the file(s).

### Requirements

As a user, I would liketo be able to duplicate selected file(s) with one quick action.

### Files Version

Preview 3.0.13.0

### Windows Version

10.0.19045.3324

### Comments

_No response_

yaira2:
**Resolved / Related Issues**
   Closes #14135...

**Validation**
   1. Confirm the duplicate items action is only enabled when items are selected
   2. Confirm the action copies and pastes the selected items

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Actions/FileSystem/DuplicateItemAction.cs
+++ b/src/Files.App/Actions/FileSystem/DuplicateItemAction.cs
@@ -1,0 +1,60 @@
+﻿// Copyright (c) 2023 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.App.Actions
+{
+	internal class DuplicateItemAction : ObservableObject, IAction
+	{
+		private readonly IContentPageContext context;
+
+		public string Label
+			=> "Duplicate".GetLocalizedResource();
+
+		public string Description
+			=> "DuplicateItemDescription".GetLocalizedResource();
+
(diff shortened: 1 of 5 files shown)


hishitetsu (commented):
This would cause the original clipboard contents to be lost. Shouldn't the items be copied without going through the clipboard?
I don't think it makes much sense as it is, since it's the same thing as pressing C and V in sequence with holding Ctrl, but it doesn't reduce the number of key presses.

yaira2:
> This would cause the original clipboard contents to be lost. Shouldn't the items be copied without going through the clipboard?
I don't think it makes much sense as it is, since it's the same thing as pressing C and V in sequence with holding Ctrl, but it doesn't reduce the number of key presses.

I agree with both points, I'll wait to see if there is more interest before continuing with this.

Issue #14240, opened 2023-12-15
Feature: Minimalize the app running in the background to the system tray


0x5bfa:
### What feature or improvement do you think would benefit Files?

Currently, we don't have our own way to kill the app process running in the background except doing so from Task Manager. This enables us to kill the process quickly.

For more about the implementation, visit #13524

### Requirements

- Add SysTray manager.
- Manage the app lifecycle.

### Files Version

3.0.0

### Windows Version

10.0.22000

yaira2:
You can also kill the process by turning off the option in settings

PR #14254, opened 2023-12-18
Feature: Added systray icon of Files shown when running on the background

0x5bfa:
## Summary

Users doesn't have a proper way to kill the app when running on the background. Added system tray icon to handle killing app by the app itself.

## Task Checklist

- [x] Added utility classes for the Taskbar
- [ ] Added functionalities to kill/resume

## Steps To Validate Changes

1. Launch the app.
2. Enable background process running.
3. Close window.
4. See notification area on the taskbar.

## Known Issues

_N/A_

## PR Checklist

- [x] **Development**: #14240
- [x] **Localization**: Modified en-US string resources[^2]
- [ ] **Co-authors**: _N/A_

## Screenshots

image


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3659,4 +3659,10 @@
   <data name="SortPriority" xml:space="preserve">
     <value>Sort priority</value>
   </data>
+  <data name="Quit" xml:space="preserve">
+    <value>Quit</value>
```
yaira2, line 3663:
```suggestion
    <value>Close Files</value>
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3659,6 +3659,12 @@
   <data name="SortPriority" xml:space="preserve">
     <value>Sort priority</value>
   </data>
+  <data name="Quit" xml:space="preserve">
```
yaira2, line 3662:
```suggestion
  <data name="CloseFiles" xml:space="preserve">
```

0x5bfa: I mistakenly merged your suggestion. IIRC we've decided what to call [before](https://discord.com/channels/725513575971684472/1164222926749442059/1164224058670448753). But I think we can decide again lol.

yaira2: Either one works

0x5bfa: User would be confused why have to close again. Idk but I often see quit but never close.

PR #14300, opened 2023-12-27
Fix: Fixed taskbar behavior in full screen mode

### LINKED ISSUE #14292: Bug: Taskbar Overlaps App in Full Screen Mode
### Description

When entering full screen mode, part of the app is displayed beneath the taskbar.
Switching to the dev version 3.0.16.0 did not resolve the issue.
image

### Steps To Reproduce

Press F11 or use command palette to enter full screen mode

### Requirements

The app should display above the taskbar, or the taskbar should hide

### Files Version

3.0.15.0

### Windows Version

10.0.19045.3693

### Log File

[debug.log](https://github.com/files-community/Files/files/13768550/debug.log)

Qin-shihuang:
**Resolved / Related Issues**
   Closes #14292

**Validation**
   1. Open app ...
   2. Click settings button ...

### DIFF
--- a/src/Files.App/Helpers/Application/AppLifecycleHelper.cs
+++ b/src/Files.App/Helpers/Application/AppLifecycleHelper.cs
@@ -331,11 +331,12 @@ namespace Files.App.Helpers
 			const string registryKey = @"Software\Microsoft\Windows\CurrentVersion\Explorer\StuckRects3";
 			const string valueName = "Settings";
 
-			using (var key = Microsoft.Win32.Registry.CurrentUser.OpenSubKey(registryKey))
-			{
-				var value = key?.GetValue(valueName) as byte[];
-				return value != null && ((value[8] & 0x01) == 1); // The least significant bit of the 9th byte controls the auto-hide setting
-			}
+			using var key = Microsoft.Win32.Registry.CurrentUser.OpenSubKey(registryKey);
+
+			var value = key?.GetValue(valueName) as byte[];
+
+			// The least significant bit of the 9th byte controls the auto-hide setting																		
+			return value != null && ((value[8] & 0x01) == 1);
 		}
 	}
 }
--- a/src/Files.App/Helpers/Application/AppLifecycleHelper.cs
+++ b/src/Files.App/Helpers/Application/AppLifecycleHelper.cs
@@ -322,5 +322,17 @@ namespace Files.App.Helpers
 			}
 			Process.GetCurrentProcess().Kill();
 		}
+
+		public static bool IsAutoHideTaskbarEnabled()
+		{
+			const string registryKey = @"Software\Microsoft\Windows\CurrentVersion\Explorer\StuckRects3";
+			const string valueName = "Settings";
+
+			using (var key = Microsoft.Win32.Registry.CurrentUser.OpenSubKey(registryKey))
+			{
+				var value = key?.GetValue(valueName) as byte[];
+				return value != null && ((value[8] & 0x01) == 1); // The least significant bit of the 9th byte controls the auto-hide setting
+			}
+		}
 	}
 }
--- a/src/Files.App/Helpers/Application/AppLifecycleHelper.cs
+++ b/src/Files.App/Helpers/Application/AppLifecycleHelper.cs
@@ -323,6 +323,9 @@ namespace Files.App.Helpers
 			Process.GetCurrentProcess().Kill();
 		}
 
+		/// <summary>
+		///	Checks if the taskbar is set to auto-hide.
+		/// </summary>
 		public static bool IsAutoHideTaskbarEnabled()
 		{
 			const string registryKey = @"Software\Microsoft\Windows\CurrentVersion\Explorer\StuckRects3";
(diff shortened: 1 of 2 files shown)


src/Files.App/MainWindow.xaml.cs
```diff
@@ -56,9 +56,26 @@ private void EnsureEarlyWindow()
...
 			// Workaround for full screen window messing up the taskbar
 			// https://github.com/microsoft/microsoft-ui-xaml/issues/8431
-			InteropHelpers.SetPropW(WindowHandle, "NonRudeHWND", new IntPtr(1));
+			// This property should only be set if the "Automatically hide the taskbar in desktop mode" setting is enabled.
+			// Setting this property when the setting is disabled will result in the taskbar overlapping the application
+			if (IsAutoHideTaskbarEnabled()) 
+				InteropHelpers.SetPropW(WindowHandle, "NonRudeHWND", new IntPtr(1));
 		}
 
+		private static bool IsAutoHideTaskbarEnabled()
```
yaira2, line 65: This should probably go in a helper
Fyi @0x5bfa

yaira2:
@Qin-shihuang can you confirm if the behavior is working correctly when auto-hide is enabled? I tried testing but it wasn't working as expected.

Qin-shihuang:
This fix relies on reading the registry to work, but it seems that the registry doesn't update immediately after system settings are changed. I suspect it mey only update on occasions like a restart. When testing, I manually changed the registry value to 0x03/0x02 and then restarted explorer.exe and Files, in which case it behaved as expected.

Stay young<https://youtu.be/VI9cMDhcw0Y?t=94>, stay simple<https://youtu.be/VI9cMDhcw0Y?t=138>. Always naïve<https://youtu.be/VI9cMDhcw0Y?t=141>.
________________________________

yaira2:
Restarting Windows after changing the setting doesn't seem to help. I wonder if it's because I'm on Windows 11.

src/Files.App/MainWindow.xaml.cs
0x5bfa: Yes. However, we still don't have a proper class for something like this, so `AppLifecycleHelper` class.

Qin-shihuang:
Basing on my test, I think the new version also works on Windows 11.

src/Files.App/MainWindow.xaml.cs
0x5bfa: @Qin-shihuang Please move `IsAutoHideTaskbarEnabled` class to that helper class to make `MainWindow` class slim.

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
```diff
@@ -322,5 +322,20 @@ public static void HandleAppUnhandledException(Exception? ex, bool showToastNoti
...
+		public static bool IsAutoHideTaskbarEnabled()
+		{
+			const string registryKey = @"Software\Microsoft\Windows\CurrentVersion\Explorer\StuckRects3";
+			const string valueName = "Settings";
+
+			using (var key = Microsoft.Win32.Registry.CurrentUser.OpenSubKey(registryKey))
+			{
+				var value = key?.GetValue(valueName) as byte[];
+				return value != null && ((value[8] & 0x01) == 1); // The least significant bit of the 9th byte controls the auto-hide setting
+			}
```
0x5bfa, line 338: Decouple using statement and put comment in a new line.

```suggestion
			using var key = Microsoft.Win32.Registry.CurrentUser.OpenSubKey(registryKey);

			var value = key?.GetValue(valueName) as byte[];

			// The least significant bit of the 9th byte controls the auto-hide setting																		
			return value != null && ((value[8] & 0x01) == 1);
```

0x5bfa (commented):
Last thing.

Otherwise, LGTM.

0x5bfa (approved):

yaira2:
@0x5bfa is the behavior working for you?

0x5bfa:
Yes.

Steps to validate:

1. Launch the app
2. Enter the full screen mode (F11)

Before(Stable)|After(Dev)
---|---
image|image

0x5bfa (approved):

yaira2:
@0x5bfa did you confirm that the existing before is still used when turning on the option to auto hide the taskbar?

0x5bfa:
When the auto hiding is enabled, you can hover on the taskbar area to show back.
It works well.

Hovering:
image

0x5bfa:
No, it's not. I tested again. Taskbar won't show when hovering on the taskbar area.

Qin-shihuang:
@0x5bfa In my local testing, both cases worked normally, but it required restarting Files twice for the changes to take effect. This might again be an issue with registry updates. 
I believe users won't be constantly changing the taskbar behavior, especially while a Files process is running, so this modification should be functional most of the time. To maintain consistent normal behavior, consider moving the related logic to the point where full screen/maximization is triggered.

https://github.com/files-community/Files/assets/68013339/60cff1ee-5e9a-442e-8fe9-2f28db1732ef

yaira2 (approved):
LGTM

PR #14309, opened 2023-12-28
Feature: Open minimized window on Windows startup

hishitetsu:
**Resolved / Related Issues**
   Closes #issue...

**Validation**
- [X] Did you implement any design changes to an existing feature?
   - [X] Was this change approved?
   1. If both "Open Files at startup" and "Leave app running" is on, Files will launch in the background at Windows startup.
   2. If "Open Files at startup" is on but "Leave app running" is off, Files will launch in the foreground at Windows startup.

### DIFF
--- a/src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
+++ b/src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
@@ -35,7 +35,7 @@ namespace Files.App.Utils.Taskbar
 
 		private bool _notifyIconCreated;
 
-		private DispatcherQueueTimer _timer;
+		private DateTime _lastLaunchDate;
 
 		// Properties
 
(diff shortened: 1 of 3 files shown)


src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
```diff
@@ -266,18 +266,17 @@ private void ShowContextMenu()
...
-				_timer ??= DispatcherQueue.GetForCurrentThread().CreateTimer();
-				_timer.Interval = TimeSpan.FromSeconds(1);
-				_timer.IsRepeating = false;
-				_timer.Start();
+				_lastLaunchDate = DateTime.Now;
 
 				_ = Launcher.LaunchUriAsync(new Uri("files-uwp:"));
 			}
+			else
+				MainWindow.Instance.Activate();
```
hishitetsu, line 279: This will bring the already open window to the foreground.

yaira2: Thanks!

PR #14330, opened 2023-12-30
Feature: Added support for hiding compression options from the context menu

### LINKED ISSUE #14305: Feature: Option to hide compression options from the context menu

Hello,
I came across this Explorer alternative in mid-November. Because Explorer still causes far too many problems in Windows 11.

And I have to say I like it very much so far.
Except that "Files" still crashes a bit too often and I don't dare to do larger copy operations.

But I think this is more likely to happen with the integrated "7zip".
So far I've only noticed that it doesn't run reliably, crashes, doesn't unzip properly.  

My separate "7-zip" runs much more reliably and especially when it crashes the explorer/files are not taken away.

What I would like to see in Files in the future:
- More freedom in the individual menu
- Remove the menu item "Unzip" because I prefer to use 7-zip separately and don't want to see two entries in it! 

- Further improvements to stability and reliability

I don't need any more bells and whistles, I'm already very satisfied. You're already doing a lot better than fucking Microsoft. :D

I want to reinstall Windows 11 in February or March.
Since the upgrade did not go so well, because there are still one or two problems, especially massive problems with the WIN 11 23H2 update.

Perhaps some stability problems are related to this.
But the fact is that I don't like the fact that 7-zip was integrated directly. This simply allows far too many problems.

Better remove it again and install 7-zip separately and implement a better integration.

Thank you.

### Requirements

- Add option to context menu settings to hide compression options

### Files Version

3.0.15.0

### Windows Version

10.0.22621.2506

### Comments

_No response_

mafra99:
**Resolved / Related Issues**
   Closes #14305 

**Validation**
   1. Compress/Extract get displayed in the context menu, when option is turned on (default behavior)
   2. Compress/Extract is hidden in the context menu if the option has been deactivated in the settings

**Screenshots (optional)**
image

### DIFF
--- a/src/Files.App/Helpers/MenuFlyout/ContextFlyoutItemHelper.cs
+++ b/src/Files.App/Helpers/MenuFlyout/ContextFlyoutItemHelper.cs
@@ -515,7 +515,7 @@ namespace Files.App.Helpers
 						new ContextMenuFlyoutItemViewModelBuilder(commands.CompressIntoZip).Build(),
 						new ContextMenuFlyoutItemViewModelBuilder(commands.CompressIntoSevenZip).Build(),
 					},
-					ShowItem = itemsSelected && CompressHelper.CanCompress(selectedItems)
+					ShowItem = userSettingsService.GeneralSettingsService.ShowCompressionOptions && itemsSelected && CompressHelper.CanCompress(selectedItems)
 				},
 				new ContextMenuFlyoutItemViewModel
 				{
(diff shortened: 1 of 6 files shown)


mafra99:
Currently it is always displayed in the InnerNavigationToolbar, should it also be hidden there, or only from the context menu?

yaira2:
> Currently it is always displayed in the InnerNavigationToolbar, should it also be hidden there, or only from the context menu?

I was wondering the same but I think it's best to limit it to the context menu.

yaira2 (approved):
LGTM

Issue #13632, opened 2023-11-01
OOP WinRT component for operations


hez2010:
### What feature or improvement do you think would benefit Files?

We initially chose Win32 Shell APIs for file operations because we wanted the file operations to continue even after Files was closed, but Win32 Shell APIs are designed for automation, not for the general use by a file explorer. And it comes with its own limitations (e.g. slow, no built-in transfer speed calculation and etc.).

We should consider implementing file operations as an out-of-process WinRT component, so that we no longer need to rely on the Win32 Shell APIs, neither facing the performance and maintainability issues due to serialization. By using WinRT we can expose the APIs with WinRT types directly (such as `IAsyncOperation<T>` and `Stream`), and call them from another process.

### Requirements

It was possible only with C++/WinRT, and CsWinRT still doesn't have built-in support for authoring OOP WinRT exe server for now.
But recently a project [Shmuelie.WinRTServer](https://github.com/shmuelie/Shmuelie.WinRTServer) has this implemented, so I think we can start to investigate it. 

### Files Version

Planning for post 3.0

### Windows Version

10.0.22631.2506

gave92:
You mean implementing file ops using Win32 API? E.g CopyFileEx, MoveFileWithProgress, etc?

hez2010:
Yes. Implementing file ops using Win32 API in an external process and expose those APIs through WinRT ABI in an external process. This is similar with what we did before in fullTrustedProcess but not the same. By authoring a WinRT exe server, we don't need to serialize and dispath messages by ourselves, instead we cancan avoid the overhead of json serialization

gave92:
Can a WinRT exe server run as administrator?

hez2010:
> Can a WinRT exe server run as administrator? (without the main app running as admin)

We may not be able to activate an elevated WinRT type from a non-elevated process.

yaira2:
Will this help us support third party handlers? #7518

hez2010:
The investigation and experiment have done. Now I'm going to commit the implementation.
Note this is not a Code Quality work but a Feature work (which allows us to have a standalone background server for long-running tasks)

JaiganeshKumaran:
Why not simply just run the regular executable but without a window open? It will be much simpler. There are not that many resources about creating an out-of-process WinRT server.

hez2010:
> Why not simply just run the regular executable but without a window open? It will be much simpler. There are not that many resources about creating an out-of-process WinRT server.

[...]

PR #13096, opened 2023-07-28
Feature: Added support for previewing Office files

gave92:
**Resolved / Related Issues**
- [x] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
- Closes #4041
- Closes #13044
- Closes #5582
- Closes #5117

**Status**
- [x] The preview renders in the correct z-order
- [x] The preview accepts mouse/keyboard input (hackish)
- [ ] The preview follows system theme, not Files setting (light/dark)
- [x] ContentExternalOutputLink requires experimental version of WinAppSdk

**Validation**
How did you test these changes?
- [x] Did you build the app and test your changes?

**Screenshots (optional)**
Screenshot 2023-07-28 202017

PR #14335, opened 2023-12-31
Feature: Added an out-of-proc WinRT server for long-running background tasks

hez2010:
**Changes**
- Upgraded to .NET 8
- Applied optimizations to all projects
- Added an out-of-proc WinRT server for long-running background tasks
- Added an `AppInstanceMonitor` to exit the server automatically if no Files app instance exists

Classes and delegates are supported, to add a class or delegate in the WinRT server:
1. It must be sealed if it's a class. 
2. Make sure you also add the type as an `ActivatableClass` in Package.appxmanifest under `OutOfProcessServer`. 
3. Methods, properties and events for both static and non-static are supported, but the type must have a projection (primitive types, types that have a [.NET/WinRT mapping](https://learn.microsoft.com/en-us/windows/apps/develop/platform/csharp-winrt/net-mappings-of-winrt-types), WinRT types and types defined in the server project) before it can be used in the signature. 

Notes:
The WinRT server is an RPC server and it **doesn't** own the object being passed to it from the client, instead it only holds a reference to it. Although you can pass an object from client to the server and use it in the server, you cannot close the client in this case as the object reference will be invalid if the client gone. If you want to use an object even after closing the client or want to share it between clients, the best approach is to instantiate the object on the server and pass it to the client so that the object will live as long as the server lives (the server won't be closed once launched, unless it was terminated manually). 

Contributes to #13632 

**Validation**

**Screenshots (optional)**
Add screenshots here.


src/Files.App (Package)/Package.appxmanifest
```diff
@@ -160,6 +160,19 @@
                 <ActivatableClass ActivatableClassId="Files.App.BackgroundTasks.UpdateTask" ThreadingModel="both" />
             </InProcessServer>
         </Extension>
+        <Extension Category="windows.activatableClass.inProcessServer">
+            <InProcessServer>
+                <Path>Files.App\Files.App.Server.dll</Path>
+                <ActivatableClass ActivatableClassId="Files.App.Server.CsWinRTWorkaround" ThreadingModel="both" />
```
hez2010, line 166: This line is necessary to workaround a bug in the appx manifest.

src/Files.App.Server/Files.App.Server.csproj
```diff
@@ -32,9 +37,16 @@
...
     <ItemGroup>
         <Manifest Include="app.manifest" />
-        <TrimmerRootAssembly Include="Files.App.Server"></TrimmerRootAssembly>
+        <TrimmerRootAssembly Include="Files.App.Server" />
         <PackageReference Include="Microsoft.Windows.CsWinRT" Version="2.0.4" />
         <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.1.647-beta" PrivateAssets="all" />
     </ItemGroup>
 
+    <ItemGroup>
+        <None Include="..\Files.App.Server\bin\$(Platform)\$(Configuration)\$(TargetFramework)\win-$(Platform)\Files.App.Server.winmd">
```
hez2010, line 46: A workaround for appx toolchains. Without this the implementation of the server (Files.App.Server.dll) will be removed from the package layout.

src/Files.App/nupkgs/Microsoft.UI.Content.Private/WinRT.cs
yaira2, line 1: Why was this removed?

yaira2: We still need this for shell previews https://github.com/files-community/Files/pull/13096

hez2010: This file will be auto-generated by CsWinRT introduced by this PR.

yaira2: I believe it had some slight modifications so it might need to be restored.  You can confirm by trying to preview a word document and seeing if the app crashes.

hez2010: I pushed a new commit that resolved this issue.

yaira2 (approved):
LGTM

PR #14350, opened 2024-01-03
Fix: Fixed NullReferenceException when renaming files

yaira2:
## Fixed NullReferenceException when renaming files
```
Files.App.Views.Layouts
GridLayoutPage.EndRename (TextBox textBox)
Files.App.Views.Layouts
BaseGroupableLayoutPage.CommitRenameAsync (TextBox textBox)
Files.App.Views.Layouts
GridLayoutPage.FileList_ItemTapped (Object sender, TappedRoutedEventArgs e)
```

### DIFF
--- a/src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
+++ b/src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
@@ -253,20 +253,27 @@ namespace Files.App.Views.Layouts
 			{
 				Popup? popup = gridViewItem.FindDescendant("EditPopup") as Popup;
 				TextBlock? textBlock = gridViewItem.FindDescendant("ItemName") as TextBlock;
-				popup!.IsOpen = false;
-				textBlock!.Opacity = (textBlock.DataContext as ListedItem)!.Opacity;
+
+				if (popup is not null)
+					popup.IsOpen = false;
+
+				if (textBlock is not null)
+					textBlock.Opacity = (textBlock.DataContext as ListedItem)!.Opacity;
 			}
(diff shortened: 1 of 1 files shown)


src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
```diff
@@ -430,15 +430,17 @@ private async void FileList_ItemTapped(object sender, TappedRoutedEventArgs e)
 						if (FolderSettings.LayoutMode == FolderLayoutModes.GridView)
 						{
 							Popup popup = gridViewItem.FindDescendant("EditPopup") as Popup;
-							var textBox = popup.Child as TextBox;
+							var textBox = popup?.Child as TextBox;
 
-							await CommitRenameAsync(textBox);
+							if (textBox is not null)
+								await CommitRenameAsync(textBox);
```
hishitetsu, line 436: I don't think this makes sense because `GridLayoutPage.EndRename` doesn't appear to throw a NullReferenceException even if `textBox` is null.

yaira2: What do you think is causing the exception?

hishitetsu: Maybe we need to consider the possibility of `popup` and `textBlock` being null.
https://github.com/files-community/Files/blob/fa2caab676b86447dbec45fc0f85b4a2106466b0/src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs#L252-L264

hishitetsu (approved):
LGTM

PR #14356, opened 2024-01-04
Fix: Fixed COMException when removing focus from the path bar

yaira2:
## Fixed COMException when removing focus from the path bar

`System.Runtime.InteropServices.COMException: The operation identifier is not valid. The WinUI Desktop Window object has already been closed`


hishitetsu (approved):
LGTM

Issue #14224, opened 2023-12-11
Automatically downloads contents of onedrive folder 


aaronsuydam:
### Description

I store my downloads folder in OneDrive. When I navigate to the folder using Files, it opens and then begins to download all the files in my downloads that have been removed from my device to save space (the "files-on-demand" functionality of OneDrive). This poses a major problem because I have very large files that I download, to keep them all on my device is not possible. I'd be happy to provide logs or something if someone wants to tell me where to look and what sections would be helpful. 

### Steps To Reproduce

1. Change download folder's location to a path in OneDrive.
2. Populate that folder and then allow OneDrive to automatically free up space.
3. Navigate to the folder using Files. 

### Requirements

The folder should open and be fully interactable (sort by date/name, scroll, etc) without downloading any files.

### Files Version

3.0.15.0

### Windows Version

10.0.22621.2715

### Log File

[debug.log](https://github.com/files-community/Files/files/13637744/debug.log)

yaira2:
Can you also clarify
1. Is files on demand turned on or off?
2. Are there any exe files in this location?
3. Does this happen for every location, or only some locations?
4. Are the files fully online, or are they partially synced?

NightmareZero:
> Can you also clarify
> 
> 1. Is files on demand turned on or off?
> 2. Are there any exe files in this location?
> 3. Does this happen for every location, or only some locations?
> 4. Are the files fully online, or are they partially synced?

I encountered the same problem.
1. on demand is on
2. no exe in this location
3. in some localtion
4.  partially synced, and there is always a re-download some files when I perform 'free up space'.

yaira2:
@NightmareZero which version of Files are you using? Can you confirm if this happens in all OneDrive locations or is it only some of them?

NightmareZero:
> @NightmareZero which version of Files are you using? Can you confirm if this happens in all OneDrive locations or is it only some of them? 

If that directory has an exe file, it will occur at any location
If there is no exe file, it will occur in certain specific directories, but I have not found any pattern

The latest version I am using is from the Windows App Store, and the version is 3.0.15.0， windows version 10.0.22631.2861

yaira2:
Are there any folders in those locations?

aaronsuydam:
> Can you also clarify
> 
>     1. Is files on demand turned on or off?
> 
>     2. Are there any exe files in this location?
> 
>     3. Does this happen for every location, or only some locations?
> 
[...]

PR #14398, opened 2024-01-09
Fix: Use the Stream API to list streams instead of opening the file

mbartlett21:
It now uses FindFirstStreamW and FindNextStreamW rather than actually opening the file. Opening a file forces a file download which is a cause of #14224.

**Resolved / Related Issues**
   Issue #14224

**Validation**
   1. Open app
   2. Make sure alternate data streams are showing in settings
   3. Open a OneDrive folder and verify that it does not download files
   4. Open a folder with a file containing an alternate data stream and verify that it appears
   4. Open a folder with a subfolder containing an alternate data stream and verify that it appears


yaira2 (approved):
LGTM

yaira2:
@mbartlett21 thank you for taking the time to debug and fix this issue!

PR #14470, opened 2024-01-16
Fix: Fixed an issue where the Properties window had the wrong icon

0x5bfa:
## Summary

As the title.

## Task Checklist

_N/A_

## Steps To Validate Changes

1. Launch the app
2. Open a properties window
3. See window icon from the taskbar

## Known Issues

_N/A_

## PR Checklist

- [ ] **Development**: _N/A_
- [ ] **Co-authors**: _N/A_

## Screenshots

Before:
image


src/Files.App/Utils/Storage/Helpers/FilePropertiesHelpers.cs
```diff
@@ -120,7 +121,14 @@ public static void OpenPropertiesWindow(object item, IShellPage associatedInstan
 			appWindow.TitleBar.ButtonBackgroundColor = Colors.Transparent;
 			appWindow.TitleBar.ButtonInactiveBackgroundColor = Colors.Transparent;
 
-			appWindow.SetIcon(applicationService.AppIcoPath);
+			string appIcoPath = ApplicationService.AppEnvironment switch
```
yaira2, line 124: Doesn't the application service already contain this logic?

0x5bfa: That was marked as obsolete.

yaira2: The service or just that method?

0x5bfa: Just that.

yaira2: I think it's worth having it in a single place instead of duplicating it.

0x5bfa: Should I do that in this PR?

yaira2: Might as well 👍

0x5bfa: I decided to use AppLifecycleHelper to put those values. Thanks for your idea.

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
```diff
@@ -26,6 +27,39 @@ namespace Files.App.Helpers
...
+		public static Version AppVersion { get; } = new(
+			Package.Current.Id.Version.Major,
+			Package.Current.Id.Version.Minor,
+			Package.Current.Id.Version.Build,
+			Package.Current.Id.Version.Revision);
+
+		/// <summary>
+		/// Gets application icon relative path.
+		/// </summary>
+		public static string AppIconPath { get; } = AppEnvironment switch
```
yaira2, line 56: What do you think about combining this with `Package.Current.InstalledLocation.Path`?

0x5bfa: Fair enough, thanks!

0x5bfa:
@yaira2 I think we better use MicaAlt and background mask to be consistent to the main window coloring. How about?
image

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
0x5bfa: Done

yaira2:
MicaAlt is best used when there are more than two layers

yaira2 (approved):

PR #14479, opened 2024-01-16
Code Quality: Added more null ref protections when renaming items

yaira2:
Added more null ref protections when renaming items

### DIFF
--- a/src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
+++ b/src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
@@ -178,7 +178,7 @@ namespace Files.App.Views.Layouts
 		override public void StartRenameItem()
 		{
 			RenamingItem = SelectedItem;
-			if (RenamingItem is null)
+			if (RenamingItem is null || FolderSettings is null)
 				return;
 
 			int extensionLength = RenamingItem.FileExtension?.Length ?? 0;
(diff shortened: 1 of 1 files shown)


src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
```diff
@@ -183,27 +183,35 @@ override public void StartRenameItem()
...
+			if (FolderSettings?.LayoutMode == FolderLayoutModes.GridView)
 			{
-				Popup popup = gridViewItem.FindDescendant("EditPopup") as Popup;
-				TextBlock textBlock = gridViewItem.FindDescendant("ItemName") as TextBlock;
+				if (gridViewItem.FindDescendant("EditPopup") is not Popup popup)
+					return;
+
 				textBox = popup.Child as TextBox;
+				if (textBox is null)
+					return;
```
0x5bfa, line 202:
```suggestion
				if (popup.Child is not TextBox textBox)
					return;
```

src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
```diff
@@ -183,27 +183,35 @@ override public void StartRenameItem()
...
 				textBlock.Opacity = 0;
 				popup.IsOpen = true;
 				OldItemName = textBlock.Text;
 			}
 			else
 			{
-				TextBlock textBlock = gridViewItem.FindDescendant("ItemName") as TextBlock;
 				textBox = gridViewItem.FindDescendant("TileViewTextBoxItemName") as TextBox;
+				if (textBox is null)
+					return;
```
0x5bfa, line 213:
```suggestion
				if (gridViewItem.FindDescendant("TileViewTextBoxItemName") is not TextBox textBox)
					return;
```

0x5bfa (commented):
Simplified type checks

src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
yaira2: `textBox` is already defined on 192 so I don't think this will work

hishitetsu (approved):
LGTM

PR #14481, opened 2024-01-16
Fix: Handle InvalidOperationException in RunPowershellCommandAsync

yaira2:
### Handle InvalidOperationException in RunPowershellCommandAsync

```
System.InvalidOperationException: No process is associated with this object
System.Diagnostics
Process.WaitForExitAsync (CancellationToken cancellationToken)
Files.App.Utils.Shell
Win32API.RunPowershellCommandAsync (String command, Boolean runAsAdmin)
```

### DIFF
--- a/src/Files.App/Utils/Shell/Win32API.cs
+++ b/src/Files.App/Utils/Shell/Win32API.cs
@@ -387,6 +387,10 @@ namespace Files.App.Utils.Shell
 			{
 				return false;
 			}
+			catch (InvalidOperationException)
+			{
+				return false;
+			}
 			catch (Win32Exception)
 			{
 				// If user cancels UAC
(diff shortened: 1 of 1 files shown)


src/Files.App/Utils/Shell/Win32API.cs
```diff
@@ -387,6 +387,10 @@ public static async Task<bool> RunPowershellCommandAsync(string command, bool ru
 			{
 				return false;
 			}
+			catch (InvalidOperationException)
+			{
+				return false;
+			}
```
hishitetsu, line 393: Since this exception should not normally occur, it is better to log the stack trace for debugging purposes.

yaira2: Should we also throw?

hishitetsu: I don't think so. Just logging is enough.

hishitetsu (approved):
LGTM

PR #14509, opened 2024-01-21
Fix: Fixed IOException in QuickLookProvider

yaira2:
**Resolved / Related Issues**
- Handle IOException in QuickLookProvider

### DIFF
--- a/src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
+++ b/src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
@@ -45,6 +45,10 @@ namespace Files.App.Services.PreviewPopupProviders
 			{
 				client.Close();
 			}
+			catch (IOException)
+			{
+				client.Close();
+			}
 		}
 
 		public async Task<bool> DetectAvailability()
@@ -64,6 +68,11 @@ namespace Files.App.Services.PreviewPopupProviders
 					return serverInstances;
 				}
 				catch (TimeoutException)
+				{
+					client.Close();
+					return 0;
+				}
+				catch (IOException)
 				{
 					client.Close();
 					return 0;
--- a/src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
+++ b/src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
@@ -42,7 +42,7 @@ namespace Files.App.Services.PreviewPopupProviders
 			}
 			catch (Exception ex) when (ex is TimeoutException or IOException)
 			{
-				client.Close();
+				// ignore
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
```diff
@@ -45,6 +45,10 @@ private async Task DoPreviewAsync(string path, string message)
 			{
 				client.Close();
 			}
+			catch (IOException)
```
hishitetsu, line 48: The following would allow for a single `catch` clause.
```csharp
catch (Exception ex) when (ex is TimeoutException or IOException)
```

src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
```diff
@@ -45,6 +45,10 @@ private async Task DoPreviewAsync(string path, string message)
...
+			{
+				client.Close();
```
hishitetsu, line 50: Although not directly related to this fix, `client` is declared with `using`, so there is no need to explicitly call `Close()`.

src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
yaira2: Thanks!

src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
yaira2: That's good to know

src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
```diff
@@ -41,7 +40,7 @@ private async Task DoPreviewAsync(string path, string message)
 				await writer.WriteLineAsync($"{message}|{path}");
 				await writer.FlushAsync();
 			}
-			catch (TimeoutException)
+			catch (Exception ex) when (ex is TimeoutException or IOException)
 			{
 				client.Close();
```
hishitetsu, line 45:
```suggestion
				// ignore
```

src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
```diff
@@ -63,7 +62,7 @@ async Task<int> QuickLookServerAvailable()
 
 					return serverInstances;
 				}
-				catch (TimeoutException)
+				catch (Exception ex) when (ex is TimeoutException or IOException)
 				{
 					client.Close();
 					return 0;
```
hishitetsu, line 68:
```suggestion
					return 0;
```

hishitetsu (commented):
I'd rather not call `Close()`, as the call to it may raise another exception.

hishitetsu (approved):
LGTM

PR #14539, opened 2024-01-23
Fix: Fixed issue where search results would sometimes use the Columns View

### LINKED ISSUE #13439: Bug: Issue with Search Page Layout Synchronization
### Description

I'm experiencing an issue with the search functionality on my computer. When I perform an initial search, the page displaying the items containing the desired keyword appears in detailed view. However, when I initiate a second search while still on the search page, the results of the second search are displayed correctly, but the layout is synchronized with the settings. In my case, for the second search, it displays in column view since I have synchronized this layout.

### Steps To Reproduce

1. Navigate to the settings, check "Sync layout and sorting preferences across directories", and make sure the column layout is selected in "Layout"
2. Perform a search from anywhere as long as you are in column mode.

### Requirements

Investigate why the search page layout is being synchronized with your settings.

### Files Version

2.5.29.0

### Windows Version

22H2 22621.1992

### Log File

...

yaira2:
**Resolved / Related Issues**
   Closes #13439...

**Validation**
   1. Set folder location to use Columns View
   2. Submit a search and confirm it uses the Details View to display the results
   3. Submit another search and confirm it's still using the Details View
   4. Go back to the original location and confirm it uses the Columns View

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Views/Shells/BaseShellPage.cs
+++ b/src/Files.App/Views/Shells/BaseShellPage.cs
@@ -498,7 +498,7 @@ namespace Files.App.Views.Shells
 			if (layout == typeof(ColumnsLayoutPage))
 				NavigateToPath(FilesystemViewModel.WorkingDirectory, typeof(DetailsLayoutPage), args);
 			else
-				ItemDisplay.Navigate(layout, args);
+				NavigateToPath(FilesystemViewModel.WorkingDirectory, layout, args);
 		}
 
 		public void NavigateWithArguments(Type sourcePageType, NavigationArguments navArgs)
--- a/src/Files.App/Views/Shells/BaseShellPage.cs
+++ b/src/Files.App/Views/Shells/BaseShellPage.cs
@@ -493,10 +493,12 @@ namespace Files.App.Views.Shells
 				SearchQuery = query,
 			};
 
-			if (this is ColumnShellPage)
+			var layout = InstanceViewModel.FolderSettings.GetLayoutType(FilesystemViewModel.WorkingDirectory);
+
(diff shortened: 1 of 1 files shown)


src/Files.App/Views/Shells/BaseShellPage.cs
```diff
@@ -493,10 +493,12 @@ public void SubmitSearch(string query)
...
 			};
 
-			if (this is ColumnShellPage)
+			var layout = InstanceViewModel.FolderSettings.GetLayoutType(FilesystemViewModel.WorkingDirectory);
+
+			if (layout == typeof(ColumnsLayoutPage))
 				NavigateToPath(FilesystemViewModel.WorkingDirectory, typeof(DetailsLayoutPage), args);
 			else
-				ItemDisplay.Navigate(InstanceViewModel.FolderSettings.GetLayoutType(FilesystemViewModel.WorkingDirectory), args);
+				ItemDisplay.Navigate(layout, args);
```
hishitetsu, line 501:
```suggestion
				NavigateToPath(FilesystemViewModel.WorkingDirectory, layout, args);
```
We should always call `NavigateToPath` at page transitions.

yaira2: What's the difference?

hishitetsu (approved):
LGTM

src/Files.App/Views/Shells/BaseShellPage.cs
hishitetsu: The processing is specific to the implementation class of IShellPage.
Especially in CloumnShellPage, calling Navigate directly may cause unexpected behavior.

PR #14627, opened 2024-02-02
Fix: Fixed null ref in OpenProperties

yaira2:
Fixed System.NullReferenceException in RecentFilesWidget.OpenProperties (RecentItem item)

### DIFF
--- a/src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
+++ b/src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
@@ -243,10 +243,11 @@ namespace Files.App.UserControls.Widgets
 
 		private void OpenProperties(RecentItem item)
 		{
-			if (item is null)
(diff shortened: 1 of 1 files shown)


src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -243,6 +243,9 @@ private void OpenFileLocation(RecentItem item)
 
 		private void OpenProperties(RecentItem item)
 		{
+			if (item is null)
+				return;
+
```
hishitetsu, line 248: This check is also good, but it is more likely that the `flyout` is null.

hishitetsu (approved):
LGTM

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
yaira2: Thanks

PR #14634, opened 2024-02-03
Fix: Fixed ArgumentOutOfRangeException in ItemViewModel

### LINKED ISSUE #14169: Bug: Adding a file to an open Files directory causes a crash
### Description

Often, I will have an imge file open in Photoshop 2023/2024 and when I "Save As" to create a new copy of the image file, Files will crash in the background.

This happens intermittently - not for every file I save as.

### Steps To Reproduce

1. Open a folder with images in Files.
2. Open an image in this folder with Adobe Photoshop 2023 or 2024.
3. In Photoshop, choose File > Save As, and give the new file a name similar to the original (for instance add " (large)" to the name.
4. Sometimes Files crashes as a result of the new file being added to the directory. Sometimes it does not crash.

### Requirements

As a user, I expect Files not to crash when files are added to an open directory.

### Files Version

3.0.14.0

### Windows Version

10.0.19045.3324

hishitetsu:
`BeginBulkOperation` suppresses `CollectionChanged` and allows changing the collection in non-UI threads. But changing the collection still needs to be done in UI threads, because changes in non-UI threads may not be seen from UI threads soon due to the memory cache.
This PR puts all `FilesAndFolders` operations into UI threads. This may affect UI performance, but `ArgumentOutOfRangeException` should no longer be thrown.

**Resolved / Related Issues**
   Closes #14169 

**Validation**

### DIFF
--- a/src/Files.App/Data/Models/ItemViewModel.cs
+++ b/src/Files.App/Data/Models/ItemViewModel.cs
@@ -698,51 +698,11 @@ namespace Files.App.Data.Models
 						{
 							FilesAndFolders.BeginBulkOperation();
 
-							var startIndex = -1;
-							var tempList = new List<ListedItem>();
+							if (addFilesCTS.IsCancellationRequested)
+								return;
 
-							void ApplyBulkInsertEntries()
-							{
-								if (startIndex != -1)
-								{
-									FilesAndFolders.ReplaceRange(startIndex, tempList);
-									startIndex = -1;
-									tempList.Clear();
-								}
-							}
-
-							for (var i = 0; i < filesAndFoldersLocal.Count; i++)
-							{
-								if (addFilesCTS.IsCancellationRequested)
-									return;
-
-								if (i < FilesAndFolders.Count)
-								{
-									if (FilesAndFolders[i] != filesAndFoldersLocal[i])
-									{
-										if (startIndex == -1)
-											startIndex = i;
-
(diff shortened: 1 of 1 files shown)


hez2010:
> But changing the collection still needs to be done in UI threads

I think only signaling the event is necessary to happen on UI threads.

hishitetsu:
> I think only signaling the event is necessary to happen on UI threads.

I thought so, too. But I found that putting a breakpoint between the collection change in the non-UI thread and the triggering of the event in the UI thread prevents the exception that could have occurred without the breakpoint. This implies that collection changes are not synchronized between threads. In fact, this PR prevents exceptions. I guess ObservableCollection doesn't assume that collection changes and event triggering take place in different threads.

src/Files.App/Data/Models/ItemViewModel.cs
```diff
@@ -692,67 +692,21 @@ void ClearDisplay()
...
 								return;
 
-							if (i < FilesAndFolders.Count)
-							{
-								if (FilesAndFolders[i] != filesAndFoldersLocal[i])
-								{
-									if (startIndex == -1)
-										startIndex = i;
+							FilesAndFolders.Clear();
+							FilesAndFolders.AddRange(filesAndFoldersLocal);
```
hishitetsu, line 705: I believe this time there is no problem and this simplifies the process.

yaira2 (approved):

PR #14643, opened 2024-02-04
Fix: Fixed crash that would occur when Git path contained an emoji

yaira2:
**Resolved / Related Issues**
- Fixed crash that would occur when Git path contained an emoji
```
StrictFilePathMarshaler.MarshalManagedToNative (Object managedObj)
/_/LibGit2Sharp/Core/FilePathMarshaler.cs, line 76
System.Text.EncoderFallbackException: Unable to translate Unicode character \\uDC83 at index 24 to specified code page
```


hishitetsu (approved):
LGTM (but I think this is rather an issue with LibGit2Sharp)

PR #14651, opened 2024-02-05
Fix: Fixed layout doesn't update on open tabs when the default layout changes

### LINKED ISSUE #14580: Bug: Layout doesn't update on open tabs when the default layout changes
### Description

If I change the layout on one tab then switch to another tab the layout wont be using the same as the previous tab. Refreshing doesn't fix it but navigating to another folder in that tab shows that it is applied. 

There is also sometimes a crash from this when navigating around or clicking one of the flyouts in the tool bar

### Steps To Reproduce

1. Make sure `Files settings > Folders > Sync layout and sorting preferences across directories` is enabled
2. Open two tab
3. Change the layout in one tab
4. Switch to the other
5. See layout not synced

### Requirements

- Make sure all tab update the current layout when selected again

### Files Version

3.1.8.0

### Windows Version

10.0.22621.3007

### Log File

[debug.log](https://github.com/files-community/Files/files/14090494/debug.log)

ferrariofilippo:
**Resolved / Related Issues**
   Closes #14580 

**Notes**
The `if` at line 432 is not easily readable. Should I add a comment/split it, or is it fine?

**Validation**
   1. Make sure Files settings > Folders > Sync layout and sorting preferences across directories is enabled
   2. Open two tab
   3. Change the layout in one tab
   4. Switch to the other


src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -188,7 +188,8 @@ public void MultitaskingControl_CurrentInstanceChanged(object? sender, CurrentIn
 			UpdateStatusBarProperties();
 			UpdateNavToolbarProperties();
 			LoadPaneChanged();
-			NavigationHelpers.UpdateInstancePropertiesAsync(navArgs);
+			UpdateLayoutIfRequired(e.CurrentInstance);
+			await NavigationHelpers.UpdateInstancePropertiesAsync(navArgs);
```
yaira2, line 192: @hishitetsu does this need to be awaited?

yaira2:
Does this also work if the default layout is changed in Settings?

yaira2:
> The if at line 432 is not easily readable. Should I add a comment/split it, or is it fine?

Can you split it?

src/Files.App/Views/MainPage.xaml.cs
hishitetsu: It is not necessary, but it is easier to debug. But if you are going to await, this should be done at the end of the method to prevent subsequent processing delays.

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -188,7 +188,8 @@ public void MultitaskingControl_CurrentInstanceChanged(object? sender, CurrentIn
...
 
 			e.CurrentInstance.ContentChanged -= TabItemContent_ContentChanged;
 			e.CurrentInstance.ContentChanged += TabItemContent_ContentChanged;
```
hishitetsu, line 195:
```suggestion

			e.CurrentInstance.ContentChanged -= TabItemContent_ContentChanged;
			e.CurrentInstance.ContentChanged += TabItemContent_ContentChanged;

			await NavigationHelpers.UpdateInstancePropertiesAsync(navArgs);
```

ferrariofilippo:
> Does this also work if the default layout is changed in Settings?

Not yet but I'll work on that and on dual pane mode

src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
```diff
@@ -642,6 +646,46 @@ public static void SetLayoutPreferencesForPath(string path, LayoutPreferencesIte
 			}
 		}
 
+		private void UpdateLayoutIfRequired(bool isChangingGridSize = false, bool force = false)
+		{
+			var isUsingGridViewWithDifferentSize =
```
yaira2, line 651: What's this check for?

src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
```diff
@@ -664,6 +668,46 @@ public static void SetLayoutPreferencesForPath(string path, LayoutPreferencesIte
...
+					);
+					break;
+				case FolderLayoutModes.ColumnView:
+					ToggleLayoutModeColumnView(true);
+					break;
+				case FolderLayoutModes.Adaptive:
+					ToggleLayoutModeAdaptive();
+					break;
+			}
+		}
```
yaira2, line 709:
```suggestion
private void UpdateLayoutIfRequired(bool isChangingGridSize = false, bool force = false)
{
	if (!force &&
		LayoutMode == UserSettingsService.FoldersSettingsService.DefaultLayoutMode)
	{
		return;
	}

	switch (UserSettingsService.FoldersSettingsService.DefaultLayoutMode)
	{
		case FolderLayoutModes.DetailsView:
			if (LayoutMode is FolderLayoutModes.DetailsView && GridViewSize != UserSettingsService.LayoutSettingsService.DefaultIconSizeDetailsView)
				ToggleLayoutModeDetailsView(true);
			break;
		case FolderLayoutModes.ListView:
			if (LayoutMode is FolderLayoutModes.ListView && GridViewSize != UserSettingsService.LayoutSettingsService.DefaultIconSizeListView)
				ToggleLayoutModeList(true);
			break;
		case FolderLayoutModes.TilesView:
			if (LayoutMode is FolderLayoutModes.TilesView && GridViewSize != UserSettingsService.LayoutSettingsService.DefaulIconSizeTilesView)
				ToggleLayoutModeTiles(true);
			break;
		case FolderLayoutModes.GridView:
			IsAdaptiveLayoutEnabled = false;

			if (LayoutMode is FolderLayoutModes.GridView && GridViewSize != UserSettingsService.LayoutSettingsService.DefaulIconSizeGridView)
				ToggleLayoutModeGridView(UserSettingsService.LayoutSettingsService.DefaulIconSizeGridView);
			break;
		case FolderLayoutModes.ColumnView:
			if (LayoutMode is FolderLayoutModes.ColumnView && GridViewSize != UserSettingsService.LayoutSettingsService.DefaultIconSizeColumnsView)
				ToggleLayoutModeColumnView(true);
			break;
		case FolderLayoutModes.Adaptive:
			ToggleLayoutModeAdaptive();
			break;
	}
}
```
We're in the process of supporting multiple icon sizes in all the layouts, this is just a rough concept but do you think something like this will work?

ferrariofilippo: If we always use `GridViewSize` regardless of the layout type, it seems fine.
We may need to add `force || (other conditions)` to all the `ifs` in the `switch`, but I need some testing to be sure about that. (Layouts may not be updated when changing the default one from settings)

yaira2: What does `force` do?

ferrariofilippo: > I would get rid of the first if statement

Still I think we should use the condition I wrote above, otherwise the layout would update only when increasing/decreasing icons' size, and not when the default layout is changed

ferrariofilippo: > What does `force` do?

When a user changes the default layout from settings, `LayoutMode` is updated in all the existing pages, but the layout itself is not applied. Because of that, `LayoutMode` would be equal to `DefaultLayoutMode` and so the layout wouldn't be updated. `force` is used to skip all the checks and force a layout refresh

yaira2: `LayoutMode is not FolderLayoutModes.LAYOUT_TYPE || GridViewSize != UserSettingsService.LayoutSettingsService.ICON_SIZE`
This should work but it'll need to be added to each case

ferrariofilippo: Should I implement it now?

yaira2: 👍

ferrariofilippo: > > What does `force` do?
> 
> When a user changes the default layout from settings, `LayoutMode` is updated in all the existing pages, but the layout itself is not applied. Because of that, `LayoutMode` would be equal to `DefaultLayoutMode` and so the layout wouldn't be updated. `force` is used to skip all the checks and force a layout refresh

Nevermind, it's not needed using the new code

yaira2 (approved):

hishitetsu (approved):

Issue #14362, opened 2024-01-06
Tags reset everytime you restart app


pj815000:
### Description

just how the title states it, all tags reset whenever you restart the app. Im sure this is a bug not intentional

### Steps To Reproduce

1. edit tags to something personal
2. restart the app manually with either taskbar or task manager
3. watch as all your personalized tags disappear

### Requirements

- whenever the files app restarts, Make sure all tags stay the same with their respected color and name

### Files Version

3.1.0.0

### Windows Version

10.0.22631.2861

### Log File

[debug.log](https://github.com/files-community/Files/files/13848291/debug.log)

Josh65-2201:
Thanks for the report, Are you talking about added tags to file/folders being removed or what the tag text is in Files settings?

arrhat:
I believe I am having the same issue, and I have more details to add. Edits to existing tags (i.e. changes to the tag name and/or color) are being dropped on program exit, except _sometimes_ the first edit made is saved. Here's what happened:

- I started with a list of nine tags: Tag1, Tag2,... Tag9. (I don't assume that the exact number of tags does not matter, but I had nine when I encountered this.)
- Rename all tags: Tag1a, Tag2a,... Tag9a. 
- - Note that I closed the settings dialogue (but not Files) between each tag edit. This might explain some behavior encountered later on.
- Exit and reopen Files. Only the _first_ edit was saved! So I now have a list of tags: Tag1a, Tag2, Tag3,... Tag9. 
- - Note that it's the first _edit_ that's saved, not the first tag in the list. If I started renaming tags from the bottom of the list -- so that Tag9a was the first edit -- then I end up with a tag list like this: Tag1, Tag2,... Tag8, Tag9a.
- I then tried to edit just 2 tags at a time, with the same result. (That is, if I change Tag2 to Tag2a and Tag3 to Tag3a, then exit and reopen, then only that first edit is saved -- I end up with Tag2a and Tag3.) The only way to edit all the tags in the list was to rename one at a time and close/reopen Files between each edit.
- Next I tried to edit all the tag colors at once. I picked new colors for all nine tags, closed and reopen Files. _None_ of the edits were saved.
- Changing tag colors one at a time (closing and reopening Files between each edit), worked. 
- I don't know why it saved the first edit when I changed all the tag names at once and none of the edits when I was changing all the tag colors, but possibly it has to do with the fact that I closed the settings dialogue between each tag rename, but did the color changes all at once?

[...]

PR #14662, opened 2024-02-07
Fix: Fixed issue where renaming a tag wouldn't save the new name

### LINKED ISSUE #12535: Bug: Renaming a tag doesn't save the changes
### Description

When editing the name of the existing tags, It doesn't save the changes in full. Only saves 1 of the changes. Sometimes not even that. 
I have tried editing one then closing the window, then editing the next. I've tried changing one or more then closing the full app and only some changes stick until a later windows restart. 
I have only tried editing the existing colour names without adding more tags.
The edited names seem to save one by one but in random intervals and tries.

### Steps To Reproduce

Open settings
Tags
Edit
Rename
Save Changes
At a later time some of the names turn to the default name

### Requirements

Have tag names saved until edited again.

### Files Version

2.4.71.0

### Windows Version

10.0.19045.2965

### Log file

[debug.log](https://github.com/files-community/Files/files/11643386/debug.log)

ferrariofilippo:
**Resolved / Related Issues**
   Closes #12535 
   Related #14362 

**Notes**
When editing a tag, we need to instantiate a new `TagViewModel` so that `CachingJsonSettingsDatabase.SetValue() -> UpdateValueInCache()` will find a difference between the old and the new list, hence saving the changes.
Before we were making a shallow-copy of the tag, causing changes to affect the object in both the lists

**Validation**
   1. Rename/Change the color of two or more tags
   2. Save Changes
   3. Restart the app 
   4. Check that changes were applied

### DIFF
--- a/src/Files.Core/ViewModels/FileTags/TagViewModel.cs
+++ b/src/Files.Core/ViewModels/FileTags/TagViewModel.cs
@@ -30,7 +30,7 @@ namespace Files.Core.ViewModels.FileTags
 			if (obj is not TagViewModel vm)
 				return false;
 
-			return Uid == vm.Uid && Name == vm.Color && Color == vm.Color;
+			return Uid == vm.Uid && Name == vm.Name && Color == vm.Color;
 		}
 	}
 }
--- a/src/Files.Core/ViewModels/FileTags/TagViewModel.cs
+++ b/src/Files.Core/ViewModels/FileTags/TagViewModel.cs
@@ -24,13 +24,5 @@ namespace Files.Core.ViewModels.FileTags
 			Color = color;
 			Uid = uid;
 		}
-
-		public override bool Equals(object? obj)
-		{
-			if (obj is not TagViewModel vm)
-				return false;
-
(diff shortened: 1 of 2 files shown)


src/Files.Core/ViewModels/FileTags/TagViewModel.cs
```diff
@@ -24,5 +24,13 @@ public TagViewModel(string name, string color, string uid)
 			Color = color;
 			Uid = uid;
 		}
+
+		public override bool Equals(object? obj)
+		{
+			if (obj is not TagViewModel vm)
+				return false;
+
+			return Uid == vm.Uid && Name == vm.Color && Color == vm.Color;
```
hishitetsu, line 33:
```suggestion
			return Uid == vm.Uid && Name == vm.Name && Color == vm.Color;
```

ferrariofilippo: Thank you, I missed it

yaira2 (approved):
LGTM

PR #14669, opened 2024-02-08
Feature: Added a prompt when failing to rename items requiring additional permissions

### LINKED ISSUE #8326: No under permission warning when renaming some specific files
### Description

No under permission warnings are shown when renaming Utilman.exe to Utilman2.exe under system32

### Steps To Reproduce

1. Go to system32
2. find Utilman.exe
2. Rename it to Utilman2.exe
3. Select "Yes" for UAC
4. Nothing changed.

### Expected behavior

There is no rename failure message. Normally it should tell us to get permission from TrustedInstaller

### Files Version

Version: 2.1.13.0

### Windows Version

Windows 11 Dev 22543.1000

### Relevant Assets/Logs

[debug_fulltrust.log](https://github.com/files-community/Files/files/8054901/debug_fulltrust.log)
[debug.log](https://github.com/files-community/Files/files/8054902/debug.log)

2022-02-13

ferrariofilippo:
**Resolved / Related Issues**
   Closes #8326 

**Known bugs**
- There are two cases where `Edit permissions` doesn't work:
  1. You start to rename an item on one page, and then (somehow) navigate to another before clicking `Edit permissions`
  2. You open the properties window for that item, navigate to a different page and then rename it

The issue behind these bugs is that the `Properties` window depends on `IShellPage` and `ListedItem`, while `RenameAsync` doesn't. To get those values, I'm using `ContentPageContext`, but this doesn't work if you change the working directory while renaming. Do you have any ideas on how I can fix this without messing up the code too much?

**Validation**
   1. Navigate to `C:\Windows\Ssytem32`
   4. Try to rename any file/folder (such as `Utilman.exe`)
   5. When the warning is shown try clicking both `Ok` and `Edit permissions`

**Screenshots**
Screenshot 2024-02-07 210409


yaira2:
Trusted Installer is just an example, I imagine there can be others cases as well.

ferrariofilippo:
> Trusted Installer is just an example, I imagine there can be others cases as well.

I guess it can only be either `TrustedInstaller` or `SYSTEM`. Is there any other account with greater privileges than admin?
Anyway, I need to find a way to distinguish the two cases since they both return code `0x80270021` (`COPYENGINE_E_ACCESS_DENIED_SRC`)

ferrariofilippo:
I think we should also extended this to other operations such as `Delete`

yaira2:
> Is there any other account with greater privileges than admin?

It could also be a non admin account

ferrariofilippo:
> It could also be a non admin account that needs access from admin.

In that case there's the UAC modal

ferrariofilippo:
I tried to find a way to distinct between `TrustedInstaller` and `SYSTEM`, but got nowhere.
Using `FileSecurityHelpers.GetAccessControlList()` and `FileSecurityHelpers.GetOwner()`, I can't get anything but generic errors or exceptions (e.g. "Access denied", "NullReference" caused by `source.ItemType` or by `GetOwner()`).
I can either use a more general string (e.g. "requires higher permissions") or try again, but in this case I may need some help

yaira2:
> I can either use a more general string (e.g. "requires higher permissions")

I think that's fine

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
```diff
@@ -33,6 +33,16 @@ public MainPropertiesPage()
 				FlowDirection = FlowDirection.RightToLeft;
 		}
 
+		public bool TryNavigateToPage(PropertiesNavigationViewItemType pageType)
```
yaira2, line 36:
```suggestion
		
		// Navigates to specified properties page
		public bool TryNavigateToPage(PropertiesNavigationViewItemType pageType)
```

yaira2 (approved):
LGTM

0x5bfa (commented):
LGTM regarding codebase quaility

PR #14673, opened 2024-02-08
Feature: Added a notification when the app goes to the background for the first time

### LINKED ISSUE #14668: Feature: Show a notification when the app goes to the background for the first time

By default, Files continues to run in the background even if the window is closed for faster startup. However, new users of the app may not be aware of this and may wonder when they find the icon in the system tray.
So, when the app goes to the background for the first time, it would be helpful to notify the user that the app continues to run in the background.

### Requirements

Show a notification when the app goes to the background for the first time.

### Files Version

3.1.1.0

### Windows Version

10.0.19045.3930

### Comments

_No response_

hishitetsu:
**Resolved / Related Issues**
   Closes #14668

**Validation**

### DIFF
--- a/src/Files.App/Strings/en-US/Resources.resw
+++ b/src/Files.App/Strings/en-US/Resources.resw
@@ -3686,7 +3686,7 @@
     <value>Change album cover</value>
   </data>
   <data name="BackgroundRunningNotificationBody" xml:space="preserve">
-    <value>Running in the background improves the startup performance.</value>
+    <value>Files is still running in the background to improve startup performance.</value>
   </data>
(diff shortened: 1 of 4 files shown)


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3685,4 +3685,10 @@
   <data name="ChangeAlbumCover" xml:space="preserve">
     <value>Change album cover</value>
   </data>
+  <data name="BackgroundRunningNotificationBody" xml:space="preserve">
+    <value>Running in the background improves the startup performance.</value>
```
yaira2, line 3689:
```suggestion
    <value>Files is still running in the background to improve startup performance.</value>
```

PR #14699, opened 2024-02-11
Fix: Fixed NullReferenceException in Item_Opening

yaira2:
Fixed `System.NullReferenceException: Object reference not set to an instance of an object`

**Stacktrace**
```csharp
Files.App.UserControls.Menus
FileTagsContextMenu.Item_Opening (Object sender, Object e)
```

### DIFF
--- a/src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
+++ b/src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
@@ -62,7 +62,7 @@ namespace Files.App.UserControls.Menus
 			}
 		}
 
-		private void Item_Opening(object sender, object e)
+		private void Item_Opening(object? sender, object e)
 		{
 			Opening -= Item_Opening;
 
--- a/src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
+++ b/src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
@@ -1,20 +1,11 @@
 ﻿// Copyright (c) 2023 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using CommunityToolkit.Mvvm.DependencyInjection;
-using Files.App.Utils;
-using Files.App.Helpers;
-using Files.Core.Services.Settings;
-using Files.Core.ViewModels.FileTags;
-using Files.Shared.Extensions;
-using Microsoft.UI.Xaml.Controls;
-using Microsoft.UI.Xaml.Media;
-using System;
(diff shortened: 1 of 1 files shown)


src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
```diff
@@ -62,7 +62,7 @@ private void TagItem_Click(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
 			}
 		}
 
-		private void Item_Opening(object sender, object e)
+		private void Item_Opening(object? sender, object e)
```
hishitetsu, line 65: This doesn't help to prevent exceptions. I think we need to consider the case where `SelectedItems` is null.

src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
```diff
@@ -46,7 +37,8 @@ public FileTagsContextMenu(IEnumerable<ListedItem> selectedItems)
 
 			SelectedItems = selectedItems;
 
-			Opening += Item_Opening;
+			if (SelectedItems is not null)
```
hishitetsu, line 40: Since the event has a time lag before it occurs, it would be better to do a null check in the event.

src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
```diff
@@ -62,15 +53,18 @@ private void TagItem_Click(object sender, Microsoft.UI.Xaml.RoutedEventArgs e)
...
+				return;
+
 			// go through each tag and find the common one for all files
 			var commonFileTags = SelectedItems
-				.Select(x => x.FileTags ?? Enumerable.Empty<string>())
-				.Aggregate((x, y) => x.Intersect(y))
-				.Select(x => Items.FirstOrDefault(y => x == ((TagViewModel)y.Tag)?.Uid));
+			.Select(x => x.FileTags ?? Enumerable.Empty<string>())
+			.Aggregate((x, y) => x.Intersect(y))
+			.Select(x => Items.FirstOrDefault(y => x == ((TagViewModel)y.Tag)?.Uid));
```
hishitetsu, line 67: Should the indentation be changed?

hishitetsu (approved):
LGTM

src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
yaira2: Fixed

PR #14715, opened 2024-02-13
Fix: Fixed NullReferenceException in DirectoryInfoUpdated

yaira2:
```csharp
BaseShellPage.FilesystemViewModel_DirectoryInfoUpdated (Object sender, EventArgs e)
```

### DIFF
--- a/src/Files.App/Views/Shells/BaseShellPage.cs
+++ b/src/Files.App/Views/Shells/BaseShellPage.cs
@@ -248,34 +248,37 @@ namespace Files.App.Views.Shells
 					? await GitHelpers.GetRepositoryHead(InstanceViewModel.GitRepositoryPath)
 					: null;
 
-			if (InstanceViewModel.GitRepositoryPath != FilesystemViewModel.GitDirectory)
+			if (FilesystemViewModel.GitDirectory is not null)
 			{
-				InstanceViewModel.GitRepositoryPath = FilesystemViewModel.GitDirectory;
-
-				InstanceViewModel.GitBranchName = headBranch is not null
-					? headBranch.Name
-					: string.Empty;
-
-				if (!_gitFetch.IsCompleted)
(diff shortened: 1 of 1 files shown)


src/Files.App/Views/Shells/BaseShellPage.cs
```diff
@@ -248,36 +248,39 @@ protected async void FilesystemViewModel_DirectoryInfoUpdated(object sender, Eve
 					? await GitHelpers.GetRepositoryHead(InstanceViewModel.GitRepositoryPath)
 					: null;
 
-			if (InstanceViewModel.GitRepositoryPath != FilesystemViewModel.GitDirectory)
+			if (FilesystemViewModel.GitDirectory is not null)
```
hishitetsu, line 251: I think this process should be executed even if `FilesystemViewModel.GitDirectory` is null.

src/Files.App/Views/Shells/BaseShellPage.cs
```diff
@@ -248,36 +248,39 @@ protected async void FilesystemViewModel_DirectoryInfoUpdated(object sender, Eve
...
-			if (!GitHelpers.IsExecutingGitAction)
-			{
-				ContentPage.DirectoryPropertiesViewModel.UpdateGitInfo(
-					InstanceViewModel.IsGitRepository,
-					InstanceViewModel.GitRepositoryPath,
-					headBranch);
-			}
-
 			ContentPage.DirectoryPropertiesViewModel.DirectoryItemCount = $"{FilesystemViewModel.FilesAndFolders.Count} {directoryItemCountLocalization}";
 			ContentPage.UpdateSelectionSize();
```
hishitetsu, line 285: Even if a null check of ContentPage is performed at the beginning, there is a possibility that it will be nulled by other threads during processing. I think that is the cause of the exception and should be changed to `ContentPage?`.

PR #14779, opened 2024-02-19
Fix: Use correct DPI settings when requesting icons

### LINKED ISSUE #14499: Bug: Some file/drive icons are displayed as a small icon
### Description

Screenshot 2024-01-20 112232

On the left how it is displayed, on the right how it should be displayed (Properties screen from Files app itself) 

The icon is set via autorun.inf

`[autorun]
icon = KODAK.ico`

### Steps To Reproduce

1. Change a drives icon using an AutoRun file and provide an icon that only contains a 128*128 icon size
2. Open Files and the icon has large margins

### Requirements

1. Solution shouldn't reduce the quality of the other icons
2. Solution must remove margin from the icon

### Files Version

3.1.1.0

### Windows Version

22621.2861

### Log File

no log for that.

### LINKED ISSUE #14772: Code Quality: Request 16x16 image for sidebar
### Description

We switched the sidebar & tabs away from Storage Thumbnail in #14701. This introduced a regression in the icon quality so we started requesting a larger size as a workaround, but this isn't a satisfactory solution for the long term.

### Concerned code

`SidebarPinnedModel.cs`
`NavigationHelpers.cs`

### Gains

- Higher quality thumbnail

### Requirements

- Sidebar and tabs should use `16` (except Recycle Bin which should use `48`)
- Image should be sharp

### Comments

image

yaira2:
**Resolved / Related Issues**
   Closes #14772
   Closes #14499

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Helpers/Interop/InteropHelpers.cs
+++ b/src/Files.App/Helpers/Interop/InteropHelpers.cs
@@ -32,6 +32,9 @@ namespace Files.App.Helpers
 		[DllImport("user32.dll", SetLastError = true)]
 		public static extern void SwitchToThisWindow(IntPtr hWnd, bool altTab);
 
+		[DllImport("User32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+		public static extern int GetDpiForWindow(IntPtr hwnd);
+
 		[DllImport("ole32.dll")]
 		public static extern uint CoWaitForMultipleObjects(uint dwFlags, uint dwMilliseconds, ulong nHandles, IntPtr[] pHandles, out uint dwIndex);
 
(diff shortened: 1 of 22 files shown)


src/Files.App/Helpers/Interop/InteropHelpers.cs
```diff
@@ -32,6 +32,9 @@ public static class InteropHelpers
 		[DllImport("user32.dll", SetLastError = true)]
 		public static extern void SwitchToThisWindow(IntPtr hWnd, bool altTab);
 
+		[DllImport("User32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+		public static extern int GetDpiForWindow(IntPtr hwnd);
```
yaira2, line 36: This is returning the wrong value, `168` instead of `175` in my case.

PR #14795, opened 2024-02-20
Feature: Restore tabs when Files process is closed incorrectly

### LINKED ISSUE #12624: Feature: Save tabs on Windows shutdown or app crash

See #12619 - Startup Setting Ignored if Files Not Closed Before PC Shutdown.

### Requirements

- Honour the startup state if Files is open when the PC is shutdown.

### Files Version

2.5.1.0

### Windows Version

Windows 11 (22H2): 22621

yaira2:
**Resolved / Related Issues**
   Closes #12624...

**Validation**
   1. Open multiple tabs
   2. Force close app from task manager
   3. Confirm tabs are restored during the next session

**Notes***
This doesn't support multiple windows, the list of saved tabs will be from whichever window was last active.

### DIFF
--- a/src/Files.App/Data/Contexts/Multitasking/MultitaskingContext.cs
+++ b/src/Files.App/Data/Contexts/Multitasking/MultitaskingContext.cs
@@ -39,6 +39,7 @@ namespace Files.App.Data.Contexts
 		private void AppInstances_CollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
 		{
 			UpdateTabCount();
+			AppLifecycleHelper.SaveSessionTabs();
 		}
 		private void AppModel_PropertyChanged(object? sender, PropertyChangedEventArgs e)
 		{


src/Files.App/Data/Contexts/Multitasking/MultitaskingContext.cs
```diff
@@ -39,6 +39,7 @@ public MultitaskingContext()
 		private void AppInstances_CollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
 		{
 			UpdateTabCount();
+			AppLifecycleHelper.SaveSessionTabs();
```
hishitetsu, line 42: Can we delete any other code that calls `SaveSessionTabs()`?

yaira2: Probably

yaira2: They might be worth keeping because `AppInstances_CollectionChanged` is only called when a tab is added/removed. This is fine for a single instance, but it can cause issues with multiple windows. That said, if we add support for saving the tab list for multiple windows, I think we can rely on this one call.

hishitetsu (approved):
LGTM

PR #14871, opened 2024-02-29
Code Quality: Remove thumbnail cache to simplify logic and improve performance

yaira2:
**Resolved / Related Issues**
   Closes #issue...

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Data/Models/ItemViewModel.cs
+++ b/src/Files.App/Data/Models/ItemViewModel.cs
@@ -937,16 +937,18 @@ namespace Files.App.Data.Models
 			return shieldIcon;
 		}
 
+		/// <summary>
+		/// Loads basic icon using ReturnIconOnly flag
+		/// </summary>
+		/// <param name="item"></param>
+		/// <returns></returns>
 		private async Task LoadBasicIconAsync(ListedItem item)
 		{
-			var thumbnailSize = folderSettings.GetRoundedIconSize();
-			var isFolder = item.IsFolder;
-
 			// Get icon
 			var icon = await FileThumbnailHelper.GetIconAsync(
 					item.ItemPath,
-					thumbnailSize,
-					isFolder,
+					folderSettings.GetRoundedIconSize(),
+					item.IsFolder,
 					false,
 					IconOptions.ReturnIconOnly);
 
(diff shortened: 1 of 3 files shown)


src/Files.App/Data/Models/ItemViewModel.cs
```diff
@@ -937,117 +937,142 @@ private async Task<BitmapImage> GetShieldIcon()
...
+			var icon = await FileThumbnailHelper.GetIconAsync(
+					item.ItemPath,
+					thumbnailSize,
+					item.IsFolder,
+					true,
+					IconOptions.None);
+
+			// Loop until cached thumbnail is loaded or timeout is reached
+			var cancellationTokenSource = new CancellationTokenSource(3000);
+			while (!icon.isIconCached)
```
hishitetsu, line 1052:
```suggestion
			while (true)
```

src/Files.App/Data/Models/ItemViewModel.cs
```diff
@@ -937,117 +937,142 @@ private async Task<BitmapImage> GetShieldIcon()
...
+					IconOptions.None);
 
-				if (iconInfo.IconData is not null)
-				{
-					await dispatcherQueue.EnqueueOrInvokeAsync(async () =>
-					{
-						item.FileImage = await iconInfo.IconData.ToBitmapAsync();
-					}, Microsoft.UI.Dispatching.DispatcherQueuePriority.Low);
-				}
+				if (cancellationTokenSource.Token.IsCancellationRequested)
```
hishitetsu, line 1061:
```suggestion
				if (icon.isIconCached || cancellationTokenSource.Token.IsCancellationRequested)
```
This prevents an extra 500 msec wait.

src/Files.App/Data/Models/ItemViewModel.cs
```diff
@@ -1198,6 +1223,14 @@ public async Task LoadExtendedItemPropertiesAsync(ListedItem item)
...
 							});
 						}
+						else
+						{
+							// Load cached thumbnail for cloud files
+							if (item.SyncStatusUI.SyncStatus != CloudDriveSyncStatus.NotSynced && item.SyncStatusUI.SyncStatus != CloudDriveSyncStatus.Unknown)
+								_ = LoadCachedThumbnailAsync(item);
+							else
+								_ = LoadThumbnailAsync(item);
+						}
```
hishitetsu, line 1233: Should this not be executed if `wasSyncStatusLoaded` is false? More to the point, I think this doesn't need to be executed after the properties are loaded, but could be executed immediately after the basic icon has been loaded.

yaira2: > but could be executed immediately after the basic icon has been loaded.

We can execute `LoadThumbnailAsync` at that point, but then we'd also be executing `LoadCachedThumbnailAsync` for cloud files once sync status is loaded.

hishitetsu: I see.

src/Files.App/Data/Models/ItemViewModel.cs
```diff
@@ -1092,7 +1117,7 @@ public async Task LoadExtendedItemPropertiesAsync(ListedItem item)
 						}
 
 						cts.Token.ThrowIfCancellationRequested();
-						_ = LoadItemThumbnailAsync(item);
+						_ = LoadBasicIconAsync(item);
```
hishitetsu, line 1120:
```suggestion
						await LoadBasicIconAsync(item);
```
It is better to await if the process is to be completed quickly,

src/Files.App/Data/Models/ItemViewModel.cs
yaira2: If you think that's fine, then we can go ahead and do make this change.

hishitetsu (approved):
LGTM in terms of code quality

src/Files.App/Data/Models/ItemViewModel.cs
yaira2: I made additional changes that I think resolve this thread

src/Files.App/Data/Models/ItemViewModel.cs
```diff
@@ -1198,6 +1098,15 @@ public async Task LoadExtendedItemPropertiesAsync(ListedItem item)
 								SetFileTag(item);
 							});
 						}
+						else
+						{
+							// Try loading thumbnail for cloud files in case they weren't cached the first time
+							if (item.SyncStatusUI.SyncStatus != CloudDriveSyncStatus.NotSynced && item.SyncStatusUI.SyncStatus != CloudDriveSyncStatus.Unknown)
+							{
+								await Task.Delay(500);
+								_ = LoadThumbnailAsync(item);
```
hishitetsu, line 1107:
```suggestion
								_ = Task.Run(async () => {
									await Task.Delay(500);
									await LoadThumbnailAsync(item);
								});
```

PR #14857, opened 2024-02-27
Fix: Fixed issue where tabs weren't restored when resuming from the background

yaira2:
**Resolved / Related Issues**
- [x] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
   Closes #14856...

**Validation**
How did you test these changes?
- [ ] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [ ] Did you implement any design changes to an existing feature?
   - [ ] Was this change approved?
- [ ] Are there any other steps that were used to validate these changes?
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.

PR #14874, opened 2024-03-01
Fix: Save tab list on path change

yaira2:
Follow up for #14857

### DIFF
--- a/src/Files.App/Views/MainPage.xaml.cs
+++ b/src/Files.App/Views/MainPage.xaml.cs
@@ -167,6 +167,9 @@ namespace Files.App.Views
 			LoadPaneChanged();
 			UpdateNavToolbarProperties();
 			await NavigationHelpers.UpdateInstancePropertiesAsync(paneArgs);
+
+			// Save the updated tab list
+			AppLifecycleHelper.SaveSessionTabs();
 		}
 
 		public async void MultitaskingControl_CurrentInstanceChanged(object? sender, CurrentInstanceChangedEventArgs e)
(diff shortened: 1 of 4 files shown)


src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -167,6 +167,9 @@ public async void TabItemContent_ContentChanged(object? sender, CustomTabViewIte
 			LoadPaneChanged();
 			UpdateNavToolbarProperties();
 			await NavigationHelpers.UpdateInstancePropertiesAsync(paneArgs);
+
+			// Save the updated tab list
+			AppLifecycleHelper.SaveSessionTabs();
```
yaira2, line 172: For some time this code runs twice when a second pane is added

hishitetsu: This can be avoided by modifying `PangeHolderPage.IsRightPaneVisible` setter as follows:
```csharp
			set
			{
				if (value != _IsRightPaneVisible)
				{
					_IsRightPaneVisible = value;
					if (!_IsRightPaneVisible)
					{
						ActivePane = PaneLeft;
						Pane_ContentChanged(null, null);
					}

					NotifyPropertyChanged(nameof(IsRightPaneVisible));
					NotifyPropertyChanged(nameof(IsMultiPaneActive));
				}
			}
```

yaira2: Thanks! I pushed a commit with this change

PR #14983, opened 2024-03-18
Feature: Add dimensions to hover on image file

### LINKED ISSUE #14975: Feature: Show image resolution on hover information

Would love it if you added the feature in the original file explorer where when you hover over photos it shows the photo's dimensions. That could really come in handy. instead of going into the properties menu.

### Requirements

### Files Version

3.2.4.0

### Windows Version

Windows 11 Pro 22631.3296

### Comments

_No response_

btomblinson:
**Resolved / Related Issues**
   Closes #14975 

**Validation**
- [x] Did you remove any strings from the en-us resource file?
   - [x] Did you search the solution to see if the string is still being used? 
   1. Tested the changes and made sure it matched what the native files app shows

**Screenshots (optional)**
Screenshot 2024-03-18 125521


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -141,6 +141,9 @@
   <data name="SizeLabel" xml:space="preserve">
     <value>Size:</value>
   </data>
+  <data name="DimensionsLabel" xml:space="preserve">
+    <value>Dimensions:</value>
```
yaira2, line 145: There is already a string for dimensions (see the details pane). It doesn't have a colon at the end but I think we can reuse the string and just insert the colon from the code.

src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
```diff
@@ -261,6 +262,15 @@ CancellationToken cancellationToken
 				itemType = itemFileExtension.Trim('.') + " " + itemType;
 			}
 
+			int imageHeight = 0, imageWidth = 0;
+			if (FileExtensionHelpers.IsImageFile(itemFileExtension))
```
yaira2, line 266: I'm wondering if we need to initialize this data when enumerating items. It might be better for performance if we retrieve it on demand from the getter in ListedItem. @hishitetsu

yaira2:
@btomblinson did you test to make sure the tooltip for non-images works as expected? It might also be worth checking each of the image formats to make sure it works correctly.

btomblinson:
Yes I tested hovering over non image files and folders, since it’s relying on the existing helper to detect if it’s an image file I figured that would be good, I can test a .jpeg too

src/Files.App/Strings/en-US/Resources.resw
yaira2: Thank you

src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
btomblinson: My thoughts are that at this stage in the pipeline its already doing a lot of metadata gathering anyway, after doing some research I went with the FileStream approach as well as Image.FromFile gets a bad rep

btomblinson:
Tested with a .jpg as well

Screenshot 2024-03-18 143011

src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
hishitetsu: I think we can get dimensions from `System.Image.Dimensions` property. I don't know which one performs better though.

hishitetsu:
With the current implementation, wouldn't it display properly in a zip file or on an FTP server?

src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
yaira2: @btomblinson

btomblinson:
> With the current implementation, wouldn't it display properly in a zip file or on an FTP server?

It doesn't

Screenshot 2024-03-21 225941

src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
btomblinson: That is how `PropertiesInformation.json` gets it, which I looked into but couldn't follow it so found it too complex for usage in this scenario, the approach I took uses standard .NET APIs that are more understandable, I also found `System.Drawing.Bitmap` which might be worth a look, do we have any suggestions on how to benchmark the 3 approaches?

btomblinson:
> > With the current implementation, wouldn't it display properly in a zip file or on an FTP server?
> 
> It doesn't
> 
> Screenshot 2024-03-21 225941

I just noticed too in the above screenshot there’s a .png below and you can faintly see the dimensions 😉

src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
btomblinson: I tested with a 21MB image for performance. Screenshots of code and results, a copy of the .cs and a copy of the image are attached. Consistently, using `System.Drawing.Image` has clocked in at 5ms or less and the `System.Drawing.Bitmap` approach took 20+ ms. I do not know how to test `System.Image.Dimensions` since thats win32 api but I believe I've shown my usage of  `System.Drawing.Image` is more than sufficient for this use case and might even confirm its the best approach period.

[Performance.zip](https://github.com/files-community/Files/files/14717849/Performance.zip)

btomblinson: > My thoughts are that at this stage in the pipeline its already doing a lot of metadata gathering anyway, after doing some research I went with the FileStream approach as well as `Image.FromFile` gets a bad rep

The below ZIP confirms my research that FileStream is definitely the fastest option

[Performance.zip](https://github.com/files-community/Files/files/14717937/Performance.zip)

yaira2: In that case let's stick with FileStream, but we can optimize further by only fetching the data when needed. True, most other properties are loaded right away, but it reduces performance and it's something we want to optimize moving forward.

btomblinson: What if anything do you need me to do?

hishitetsu:
> > > With the current implementation, wouldn't it display properly in a zip file or on an FTP server?
> > 
> > 
> > It doesn't
> > Screenshot 2024-03-21 225941
> 
> I just noticed too in the above screenshot there’s a .png below and you can faintly see the dimensions 😉

I mean image files inside an archive. Dimensions shouldn't be displayed if they can't be retrieved.

image

hishitetsu:
If there are corrupt image files in a folder, the folder cannot be opened. The procedure is as follows:
1. Create a text file.
2. Change the extension of the file to `png`.
3. Try to open the folder which contains that file.

yaira2:
@btomblinson can you take a look at that issue?

Jay-o-Way (changes requested):
Small thing

src/Files.App/Data/Items/ListedItem.cs
```diff
@@ -268,6 +270,31 @@ public string FileSize
...
+		{
+			get => imageHeight;
+			set
+			{
+				SetProperty(ref imageHeight, value);
+				OnPropertyChanged(nameof(DimensionsDisplay));
+			}
+		}
+
+		public string DimensionsDisplay => IsImage ? $"{ImageWidth} x {ImageHeight}" : string.Empty;
```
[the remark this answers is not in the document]
0x5bfa: I see many apps use x instead of symbols.
Is there a reason why you suggested?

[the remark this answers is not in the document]
Jay-o-Way: > I see many apps use x instead of symbols.
> Is there a reason why you suggested?

Just because somebody else does it, doesn't mean it's right 😉

Please use
image
image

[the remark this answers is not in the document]
0x5bfa: It sounds right.

https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/m/multiplication-sign

0x5bfa:
> If there are corrupt image files in a folder, the folder cannot be opened. The procedure is as follows:
> 
> 1. Create a text file.
> 
> 2. Change the extension of the file to `png`.
> 
> 3. Try to open the folder which contains that file.

@btomblinson Can you work on this?

btomblinson:
I am out of the loop, what is meant by corrupt image? Can I have an example?

0x5bfa:
He means an invalid image file; for example, it doesn't have data (0 byte) or it has been modified with notepad or something, resulting in corruption.

src/Files.App/Data/Items/ListedItem.cs
```diff
@@ -277,6 +279,31 @@ public string FileSize
...
+		{
+			get => imageHeight;
+			set
+			{
+				SetProperty(ref imageHeight, value);
+				OnPropertyChanged(nameof(DimensionsDisplay));
+			}
+		}
+
+		public string DimensionsDisplay => IsImage ? $"{ImageWidth} × {ImageHeight}" : string.Empty;
```
0x5bfa, line 305:
```suggestion
		public string DimensionsDisplay => IsImage ? $"{ImageWidth} \u00D7 {ImageHeight}" : string.Empty;
```

0x5bfa (changes requested):

0x5bfa:
> Dimensions shouldn't be displayed if they can't be retrieved.

> If there are corrupt image files in a folder, the folder cannot be opened. The procedure is as follows:
> 
> 1. Create a text file.
> 2. Change the extension of the file to `png`.
> 3. Try to open the folder which contains that file.

@btomblinson Hi! How's going on those issues above?

yaira2:
> He means an invalid image file; for example, it doesn't have data (0 byte) or it has been modified with notepad or something, resulting in corruption.

@0x5bfa do you have any example files to upload?

0x5bfa:
Create an empty txt file
Rename extension to PNG or smth.

yaira2:
@0x5bfa did you close this by mistake?

0x5bfa:
We'll take over this feature implementation in #15543

PR #14990, opened 2024-03-19
Code quality: Seal all non-inherited classes to improve performance

### LINKED ISSUE #12304: Code Quality: Seal classes when possible
### Description

Related to #4180.

Unsealed classes introduce extra overhead for running each virtual method/type checks because the compiler/JIT has to do extra inheritance checks.

For example, there are three non-sealed classes called `ClassA`, `ClassB` and `ClassC`, and `ClassB` inherits from `ClassA`. If `ClassC` doesn't inherit from `ClassB` (in other forms, if nothing inherits from `ClassB`) then we don't need the extra dispatch operations or checks. Thus making `ClassB` sealed removes that overhead when calling virtual methods or doing type checks.

See https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/#peanut-butter for more in-depth details.

### Concerned code

Most C# source files in the entire app solution.

### Gains

- Slightly better performance.

### Requirements

- Make classes that don't need to be inherited sealed.

### Comments

_No response_

gumbarros:
**Resolved / Related Issues**
Closes #12304

**Validation**

### DIFF
--- a/src/Files.App/Helpers/WMI/ManagementEventWatcher.cs
+++ b/src/Files.App/Helpers/WMI/ManagementEventWatcher.cs
@@ -11,7 +11,7 @@ namespace Files.App.Helpers
 	public delegate void EventArrivedEventHandler(object sender, EventArrivedEventArgs e);
 
 	/// <summary>
-	/// A public sealed class used to start/stop the subscription to specific indication source,
+	/// A public class used to start/stop the subscription to specific indication source,
 	/// and listen to the incoming indications, event <see cref="EventArrived" />
 	/// will be raised for each cimindication.
 	/// Original Sourced from: https://codereview.stackexchange.com/questions/255055/trying-to-replace-managementeventwatcher-class-in-system-management-to-switch-to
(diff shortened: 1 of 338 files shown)


yaira2 (dismissed):

src/Files.App/Helpers/WMI/ManagementEventWatcher.cs
```diff
@@ -11,7 +11,7 @@ namespace Files.App.Helpers
 	public delegate void EventArrivedEventHandler(object sender, EventArrivedEventArgs e);
 
 	/// <summary>
-	/// A public class used to start/stop the subscription to specific indication source,
+	/// A public sealed class used to start/stop the subscription to specific indication source,
```
0x5bfa, line 14: Why is this changed?

0x5bfa (commented):
lgtm.

src/Files.App/Helpers/WMI/ManagementEventWatcher.cs
gumbarros: Good eye, I changed the classes using Replace All, compiling and seeing what can be `sealed` or not.

0x5bfa: I'd do that too but is that necessary change or does it make sense to be changed?

gumbarros: It's not a necessary change, ManagementEventWatcher is not sealed.

0x5bfa (approved):

yaira2:
I think we should merge this regardless, but it would be interesting to see if there is a performance difference.

PR #14991, opened 2024-03-19
Fix: Fixed ComboBox position shift

### LINKED ISSUE #14965: Bug: ComboBox list shift to the right when clicked on
### Description

When clicking on a ComboBox list (list of selectable items) in Settings, the list moves away from under the cursor.

https://github.com/files-community/Files/assets/87632612/f9ab3841-c470-4430-a10c-33a00bfc5cb2

### Steps To Reproduce

1. Open Settings
2. Click on any ComboBox 
3. List moves to the right and is not at the same location as it was before.

### Requirements

Would be better if the list stays under the mouse cursor on being clicked.

### Files Version

Latest dev build (3.2.7.0)

### Windows Version

22631.3296

### Log File

[debug.log](https://github.com/files-community/Files/files/14611969/debug.log)

TheChilledBuffalo:
**Resolved / Related Issues**
   Closes #14965 

**Validation**
   1. Open Files Dev build
   2. Click settings button
   3. Go to pages where `ComboBox` -es are present (General, Appearance, Folders, Layout)
   4. Check the `ComboBox` -es to see if the changes have been applied

**Screenshots (optional)**

https://github.com/files-community/Files/assets/87632612/705f4a71-70a5-40d1-a307-52df82d6b8b0


src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
```diff
@@ -0,0 +1,9 @@
+<!--  Copyright (c) 2024 Files Community. Licensed under the MIT License. See the LICENSE.  -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+
+	<Style x:Key="ComboBoxStyle" TargetType="ComboBox">
+		<Setter Property="CornerRadius" Value="4" />
+		<Setter Property="MinWidth" Value="240" />
```
0x5bfa, line 6: What is this based on? I don’t like this kind of hard coding.

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
```diff
@@ -0,0 +1,9 @@
+<!--  Copyright (c) 2024 Files Community. Licensed under the MIT License. See the LICENSE.  -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+
+	<Style x:Key="ComboBoxStyle" TargetType="ComboBox">
+		<Setter Property="CornerRadius" Value="4" />
```
0x5bfa, line 5: Use StaticResource at least.
And I think this isn’t needed because this is the default value.

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
```diff
@@ -0,0 +1,9 @@
+<!--  Copyright (c) 2024 Files Community. Licensed under the MIT License. See the LICENSE.  -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+
+	<Style x:Key="ComboBoxStyle" TargetType="ComboBox">
```
0x5bfa, line 4: Make more unique name such as FixedWidthComboboxStyle.

0x5bfa (changes requested):
I’m really not a fan of this.

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
TheChilledBuffalo: I tried to dynamically set a width depending on the width of the largest `ComboBoxItem` but I couldn't. In the end I settled with 240, which seemed reasonable during my testing.

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
TheChilledBuffalo: I thought it was default too, but it didn't show up when I tested. So I manually set it.

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
TheChilledBuffalo: fixed

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
yaira2: 240 is a bit large, I would aim for 100.

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
0x5bfa: If you'd like this to be default, you can name it as 'DefaultComboboxStyle'

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
0x5bfa: Can't we make a dynamic extension to measure the largest item size and set that width as min width as other context menus do?

0x5bfa: If not, 100 sounds good to me

yaira2: It's better to be dynamic based on the current selected item

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
0x5bfa: Ah you didn't set a style to be based on. You should add BasedOn="static resource DefaultComboboxStyle"(not correct grammar tho).

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
0x5bfa: Could make an extension using dependency object.
100 doesn't have rationale.

yaira2: I understood that as long as a minimum value was provided, the combo box wouldn't shift, is this not the case?

0x5bfa: Yeah as a workaround this is good enough.

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
TheChilledBuffalo: Fixed!

src/Files.App/ResourceDictionaries/ComboBoxStyle.xaml
TheChilledBuffalo: @yaira2 @0x5bfa Hey thank you for your suggestions, but 100 seems to be not adequate. Here's a video:

https://github.com/files-community/Files/assets/87632612/5e2e7eb3-e852-47c1-91c1-a31972c9363a

How about a `160`? But to be honest even `160` isn't enough to some long strings

0x5bfa: I can’t say anything and approve this PR because there’s a much better approach towards this issue; we should set MinWidth set based on the max width of the binding elements using xaml extension.

yaira2: I had originally thought this was a WinUI issue and from this [comment](https://github.com/files-community/Files/issues/14965#issuecomment-1999062570), I thought it meant that setting a value (even if it's `1`), would serve as a workaround. After looking into this further, it seems that this behavior is by design.

It might be best to bring this up on the WinUI repo to learn more about this behavior and what the recommended action is in this case.

TheChilledBuffalo: @yaira2 I found this

https://github.com/microsoft/microsoft-ui-xaml/issues/9009

yaira2: Creating an extension like @0x5bfa suggested sounds like a reasonable approach but it would need to update whenever the selection changes.

0x5bfa: Yeah but this should've been done by MUX team. Can I work on it?

TheChilledBuffalo: I'm ok with it ig. Thank you for taking it up.

0x5bfa: Off you can work on it if you want.
Are you gonna work on it with Yair's approval?

TheChilledBuffalo: I think I will leave this to you. I'm still only getting familiar with the codebase and the current solution seems too complex for me atm.

0x5bfa: What do you think?
I've made a branch and added an extension for that though I don't know if it works yet.

I'll need to add a dependency property inside of DependencyObject-derived class(the extension class) and add `extensions:ComboBoxExtensions.IsKeepWidthEnabled="True"`.

0x5bfa: @yaira2 here’s the implementation. That easy but seems not working when loaded because ActualWidth is 0 then.

https://github.com/files-community/Files/compare/main...0x5bfa:5bfa/Feature-KeepComboBoxWidth

yaira2: That looks like a neat solution, I think we should move forward with it if that last issue is resolved.

yaira2:
It's all worth it if it leads to a better solution 🙂

Issue #10268, opened 2022-10-21
Code quality: Await is being used inside foreach loop


QuaintMako:
### Description

Through the code, we can find some pieces of code looking like :

```
foreach (obj o in objectCollection)
{
    await AsyncMethod(o);
}
```

It is not the most efficient way of performing the tasks, since it means we are waiting for each task to be done before performing the next one. It would be much faster to perform all the task using the `Task.WhenAll `process.

As an exemple, using this process inside the `AddAllItemsToSidebar` method allowed for a gain of 15 % in speed.

### Concerned code

- Whole solution

### Gains

- A gain in performances, depending on how often the code is used.

### Requirements

Replacing 

```
foreach (obj o in objectCollection)
{
    await AsyncMethod(o);
}
```

By

```
var tasksCollections = objectCollection.Select(o=> AsyncMethod(o));
await Task.WhenAll(tasksCollections );
```

yaichenbaum:
Marking this as ready to build

gave92:
Consider that doing this in AddAllItemsToSidebar  would change the order of the items. Some more changes are required.

yaichenbaum:
I think every scenario needs to be evaluated.

QuaintMako:
I did the tests on `AddAllItemsToSidebar ` since it was a pretty easy step to test. Each and every modification will be evaluated for sure, it works only in scenarios where the order of task execution does not matter.

yaira2:
> Consider that doing this in AddAllItemsToSidebar would change the order of the displayed items. Some more changes are required.

@gave92 do you have anything specific in mind?

gave92:
It's been a while sorry I do not recall :)

yaira2:
How about
```csharp
await Task.WhenAll(FavoriteItems.AsParallel().AsOrdered().Select(path => AddItemToSidebarAsync(path)));
```

PR #15005, opened 2024-03-20
Code Quality: Reduced usage of Await inside foreach loops

gumbarros:
**Resolved / Related Issues**
   Related to #10268 

**Validation**
   1. Open app ...
   2. Click settings button ...

Fixed every `.ForEach` async void, except two edge cases at `ItemViewModel.cs`,in the method `UpdateDateDisplay` and the property `EnabledGitProperties`.


src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -38,8 +38,13 @@ public async Task ExecuteAsync()
 
 			await Task.WhenAll(files.Select(file 
 				=> NavigationHelpers.OpenPath(file.path, _pageContext.ShellPage!)));
-
-			folders.ForEach(async folder => await NavigationHelpers.OpenPathInNewTab(folder.path, false));
+			
+			var openTasks = folders.Select(async folder =>
+			{
+				await NavigationHelpers.OpenPathInNewTab(folder.path, false);
```
yaira2, line 44: I think this messes with the order, the ideal behavior is for them to open in the same order they are displayed in the list.

src/Files.App/Data/Factories/ShellContextFlyoutHelper.cs
```diff
@@ -359,20 +359,30 @@ async Task InvokeShellMenuItemAsync(ContextMenu contextMenu, object? tag)
 					itemContextMenuFlyout.SecondaryCommands.Insert(1, sendToItems.FirstOrDefault());
 				}
 
-				// Add items to shell submenu
-				shellMenuItems.Where(x => x.LoadSubMenuAction is not null).ForEach(async x =>
-				{
-					await x.LoadSubMenuAction();
 
-					if (!UserSettingsService.GeneralSettingsService.MoveShellExtensionsToSubMenu)
+				var itemsWithSubMenu = shellMenuItems.Where(x => x.LoadSubMenuAction is not null);
```
yaira2, line 363: See my earlier comment

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -901,21 +901,25 @@ private async Task AddShellMenuItemsAsync(List<ContextMenuFlyoutItemViewModel> s
 				}
 			}
 
-			// Add items to main shell submenu
-			mainShellMenuItems.Where(x => x.LoadSubMenuAction is not null).ForEach(async x =>
+			// Filter mainShellMenuItems that have a non-null LoadSubMenuAction
+			var mainItemsWithSubMenu = mainShellMenuItems.Where(x => x.LoadSubMenuAction is not null).ToList();
```
yaira2, line 905: See my earlier comment

jmarrujo91-max (approved):

gumbarros:
@yaira2  fixed the places that you asked, the other places can run concurrently?

yaira2:
> Good morning @yaira2! fixed the places that you asked, the other places can run concurrently?

The code for clearing the jumplist can, but I'm pretty sure the other ones need to run in a specific order. It might be best to revert those changes and just add a comment for future reference.

gumbarros:
@yaira2 I think now it's ready, the only places that have `Task.WhenAll` now are for clearing the jumplist can, where the order does not matter.  Just clarifying the other changes, places with `.ForEach` with `async ` are changed to C# `foreach` so the tasks can be correctly handled.

yaira2:
Thank you for staying on top of this! I'll try to finish my review in the next few days.

Just a quick tip regarding "conversations", instead of marking them as resolved, it's best to add a comment so the reviewer can see it and complete the conversation if they are happy with the solution.

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -39,7 +39,15 @@ public async Task ExecuteAsync()
...
-			folders.ForEach(async folder => await NavigationHelpers.OpenPathInNewTab(folder.path, false));
+			var newTabsTasks = folders.Select(async folder =>
+			{
+				await NavigationHelpers.OpenPathInNewTab(folder.path, false);
+			});
+			
+			foreach (var task in newTabsTasks)
+			{
+				await task;
+			}
```
hez2010, line 50: Can you use `await Task.WhenAll(newTabsTasks)` instead?

hez2010 (changes requested):

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
gumbarros: If it's ok to run out of order this, I can make the change.

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -39,7 +39,15 @@ public async Task ExecuteAsync()
 			await Task.WhenAll(files.Select(file 
 				=> NavigationHelpers.OpenPath(file.path, _pageContext.ShellPage!)));
 
-			folders.ForEach(async folder => await NavigationHelpers.OpenPathInNewTab(folder.path, false));
+			var newTabsTasks = folders.Select(async folder =>
+			{
+				await NavigationHelpers.OpenPathInNewTab(folder.path, false);
```
BY-DevInnovation, line 44: You should not put `await` here. What you want to do is returning the task then you make `await Task.WhenAll(newTabsTasks);`

gumbarros: Done, you're 100% right, I removed `async void` from ForEach and forgot that `Select` also don't support `Task`.

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
gumbarros: I think the order here does not matter, see 1137d1945ea8734c425cb44b6c5e3003b7b41dff

yaira2: Opening new tabs, I think should be based on the order of the items in the list, otherwise it can appear as a bug to the user when the same action sometimes has different results.

gumbarros: Agree, changed to `foreach`

BY-DevInnovation (approved):

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -33,13 +33,18 @@ public OpenAllTaggedActions()
 
 		public async Task ExecuteAsync()
 		{
-			var files = _tagsContext.TaggedItems.Where(item => !item.isFolder);
-			var folders = _tagsContext.TaggedItems.Where(item => item.isFolder);
+			var files = _tagsContext.TaggedItems.Where(item => !item.isFolder).Select(f => f.path);
+			var folders = _tagsContext.TaggedItems.Where(item => item.isFolder).Select(f => f.path);
```
yaira2, line 37:
```suggestion
			var filePaths = _tagsContext.TaggedItems.Where(item => !item.isFolder).Select(f => f.path);
			var folderPaths = _tagsContext.TaggedItems.Where(item => item.isFolder).Select(f => f.path);
```

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -33,13 +33,18 @@ public OpenAllTaggedActions()
...
-
-			folders.ForEach(async folder => await NavigationHelpers.OpenPathInNewTab(folder.path, false));
+			foreach (var file in files)
+			{
+			    await NavigationHelpers.OpenPath(file, _pageContext.ShellPage!);
+			}
+			
+			foreach (var folder in folders)
+			{
+			    await NavigationHelpers.OpenPath(folder, _pageContext.ShellPage!);
```
yaira2, line 46: Isn't this supposed to be `OpenPathInNewTab`?

gumbarros: You're right

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -33,13 +33,18 @@ public OpenAllTaggedActions()
...
-
-			folders.ForEach(async folder => await NavigationHelpers.OpenPathInNewTab(folder.path, false));
+			foreach (var file in filePaths)
+			{
+			    await NavigationHelpers.OpenPath(file, _pageContext.ShellPage!);
+			}
+			
+			foreach (var folder in folderPaths)
+			{
+			    await NavigationHelpers.OpenPathInNewTab(folder,false);
```
yaira2, line 46:
```suggestion
			    await NavigationHelpers.OpenPathInNewTab(folder, false);
```

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -33,13 +33,18 @@ public OpenAllTaggedActions()
...
-			var files = _tagsContext.TaggedItems.Where(item => !item.isFolder);
-			var folders = _tagsContext.TaggedItems.Where(item => item.isFolder);
+			var filePaths = _tagsContext.TaggedItems.Where(item => !item.isFolder).Select(f => f.path);
+			var folderPaths = _tagsContext.TaggedItems.Where(item => item.isFolder).Select(f => f.path);
 
-			await Task.WhenAll(files.Select(file 
-				=> NavigationHelpers.OpenPath(file.path, _pageContext.ShellPage!)));
-
-			folders.ForEach(async folder => await NavigationHelpers.OpenPathInNewTab(folder.path, false));
+			foreach (var file in filePaths)
```
yaira2, line 39:
```suggestion
			foreach (var path in filePaths)
```

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -33,13 +33,18 @@ public OpenAllTaggedActions()
...
+			{
+			    await NavigationHelpers.OpenPath(file, _pageContext.ShellPage!);
+			}
+			
+			foreach (var folder in folderPaths)
```
yaira2, line 44:
```suggestion
			foreach (var path in folderPaths)
```
I think we can also remove the curly brackets

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
yaira2, line 1: The changes here look good but I'd like some more time to go over the other files.

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -161,11 +161,13 @@ public async Task<ReturnResult> DeleteItemsAsync(IEnumerable<IStorageItemWithPat
...
+			
+			// Create a list of tasks to remove items from the jump list asynchronously
+			var removeTasks = source.Select(x => jumpListService.RemoveFolderAsync(x.Path));
 
-			var itemsDeleted = history?.Source.Count ?? 0;
-
-			// Remove items from jump list
-			source.ForEach(async x => await jumpListService.RemoveFolderAsync(x.Path));
+			// Await all removal tasks to complete concurrently
+			await Task.WhenAll(removeTasks);
```
yaira2, line 170: I think these can be merged into a single line

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -476,8 +478,11 @@ public async Task<ReturnResult> MoveItemsAsync(IEnumerable<IStorageItemWithPath>
 				App.HistoryWrapper.AddHistory(history);
 			}
 
-			// Remove items from jump list
-			source.ForEach(async x => await jumpListService.RemoveFolderAsync(x.Path));
+			// Create a list of tasks to remove items from the jump list asynchronously
+			var removeTasks = source.Select(x => jumpListService.RemoveFolderAsync(x.Path));
+
+			// Await all removal tasks to complete
+			await Task.WhenAll(removeTasks);
```
yaira2, line 485: As above

src/Files.App/Views/Layouts/BaseLayoutPage.cs
yaira2, line 1: Is this just formatting?

gumbarros: No, it's `.ForEach(async x =>` converted to a C# `foreach`, because LINQ foreach does not support async.

gumbarros: We can also migrate to `Parallel.ForEachAsync`

yaira2: The order makes a difference here

gumbarros: It's fine with the OG `foreach` with `async`

yaira2 (approved):

hishitetsu (commented):
Why are you changing LINQ ForEach to foreach loop? Since lambda expressions in ForEach are not awaited, awaiting them in foreach loop is not good for performance.

gumbarros:
> Why are you changing LINQ ForEach to foreach loop? Since lambda expressions in ForEach are not awaited, awaiting them in foreach loop is not good for performance.

https://stackoverflow.com//questions/59518243/what-is-the-proper-way-to-await-for-async-inside-of-a-foreach#answer-59518587

This answer explains well the difference.

Then we can migrate the changed .ForEach with `async void` to `Task.WhenAll`, as they were already out of order.

hishitetsu:
Does this mean that this PR will not only improve performance, but will also include modifications to guarantee order where order is important, even at the expense of performance?

> Then we can migrate the changed .ForEach with async void to Task.WhenAll

Awaiting `Task.WhenAll` also affects performance compared to .ForEach with `async void`. Is that also factored in?

gumbarros:
> Does this mean that this PR will not only improve performance, but will also include modifications to guarantee order where order is important, even at the expense of performance?
> 
> > Then we can migrate the changed .ForEach with async void to Task.WhenAll
> 
> Awaiting `Task.WhenAll` also affects performance compared to .ForEach with `async void`. Is that also factored in?

What about `Parallel.ForEachAsync` instead of ForEach with `async void`?

hishitetsu:
The issue is whether or not to await `Task.WhenAll`. If not awaiting, the performance is the same as .ForEach with `async void`.
If there is no performance problem with await, then that's fine.

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -33,13 +33,14 @@ public OpenAllTaggedActions()
...
-
-			await Task.WhenAll(files.Select(file 
-				=> NavigationHelpers.OpenPath(file.path, _pageContext.ShellPage!)));
-
-			folders.ForEach(async folder => await NavigationHelpers.OpenPathInNewTab(folder.path, false));
+			var filePaths = _tagsContext.TaggedItems.Where(item => !item.isFolder).Select(f => f.path);
+			var folderPaths = _tagsContext.TaggedItems.Where(item => item.isFolder).Select(f => f.path);
+
+			foreach (var path in filePaths) 
+				await NavigationHelpers.OpenPath(path, _pageContext.ShellPage!);
```
hishitetsu, line 40: I don't think the order is important here and `Task.WhenAll` should be used.

src/Files.App/Data/Factories/ShellContextFlyoutHelper.cs
```diff
@@ -359,20 +359,27 @@ async Task InvokeShellMenuItemAsync(ContextMenu contextMenu, object? tag)
...
-				shellMenuItems.Where(x => x.LoadSubMenuAction is not null).ForEach(async x =>
-				{
-					await x.LoadSubMenuAction();
 
+				var itemsWithSubMenu = shellMenuItems.Where(x => x.LoadSubMenuAction is not null);
+				
+				//The order here may be important
+				foreach (var item in itemsWithSubMenu)
+				{
+					await item.LoadSubMenuAction();
```
hishitetsu, line 368: The order is important for `AddItemsToMainMenu` and `AddItemsToOverflowMenu`, but not for `LoadSubMenuAction`.

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -903,22 +903,25 @@ private async Task AddShellMenuItemsAsync(List<ContextMenuFlyoutItemViewModel> s
...
+			// Filter mainShellMenuItems that have a non-null LoadSubMenuAction
+			var mainItemsWithSubMenu = mainShellMenuItems.Where(x => x.LoadSubMenuAction is not null);
 
-			// Add items to overflow shell submenu
-			overflowShellMenuItems.Where(x => x.LoadSubMenuAction is not null).ForEach(async x =>
+			// Load main submenus asynchronously
+			foreach (var item in mainItemsWithSubMenu)
 			{
-				await x.LoadSubMenuAction();
+				await item.LoadSubMenuAction();
```
hishitetsu, line 912: As above.

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -903,22 +903,25 @@ private async Task AddShellMenuItemsAsync(List<ContextMenuFlyoutItemViewModel> s
...
+			}
 
-				ShellContextFlyoutFactory.AddItemsToOverflowMenu(overflowItem, x);
-			});
+			// Filter overflowShellMenuItems that have a non-null LoadSubMenuAction
+			var overflowItemsWithSubMenu = overflowShellMenuItems.Where(x => x.LoadSubMenuAction is not null);
 
+			foreach (var item in overflowItemsWithSubMenu)
+			{
+				await item.LoadSubMenuAction();
```
hishitetsu, line 921: As above.

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
yaira2: https://github.com/files-community/Files/pull/15005#discussion_r1545862037

hishitetsu: I think yes for folders, but no for files, because they are opened by other applications.

yaira2: Good point

src/Files.App/Actions/Content/Tags/OpenAllTaggedActions.cs
```diff
@@ -33,13 +33,21 @@ public OpenAllTaggedActions()
...
+				.Select(f => f.path)
+				.ToList();
+			
+			var folderPaths = _tagsContext
+				.TaggedItems
+				.Where(item => item.isFolder)
+				.Select(f => f.path)
+				.ToList();
+			
+			await Task.WhenAll(filePaths.Select(path=> NavigationHelpers.OpenPath(path, _pageContext.ShellPage!)));
```
hishitetsu, line 47:
```suggestion
			await Task.WhenAll(filePaths.Select(path => NavigationHelpers.OpenPath(path, _pageContext.ShellPage!)));
```

src/Files.App/Data/Factories/ShellContextFlyoutHelper.cs
```diff
@@ -359,20 +359,27 @@ async Task InvokeShellMenuItemAsync(ContextMenu contextMenu, object? tag)
 					itemContextMenuFlyout.SecondaryCommands.Insert(1, sendToItems.FirstOrDefault());
 				}
 
-				// Add items to shell submenu
-				shellMenuItems.Where(x => x.LoadSubMenuAction is not null).ForEach(async x =>
-				{
-					await x.LoadSubMenuAction();
 
+				var itemsWithSubMenu = shellMenuItems.Where(x => x.LoadSubMenuAction is not null).ToList();
```
hishitetsu, line 363:
```suggestion
				// Add items to shell submenu
				var itemsWithSubMenu = shellMenuItems.Where(x => x.LoadSubMenuAction is not null);
```

src/Files.App/Data/Factories/ShellContextFlyoutHelper.cs
```diff
@@ -359,20 +359,27 @@ async Task InvokeShellMenuItemAsync(ContextMenu contextMenu, object? tag)
...
+					else
 					{
-						AddItemsToOverflowMenu(overflowItem, x);
+						var overflowItem = itemContextMenuFlyout.SecondaryCommands.FirstOrDefault(x => x is AppBarButton appBarButton && (appBarButton.Tag as string) == "ItemOverflow") as AppBarButton;
+
+						if (overflowItem != null)
+						{
+							AddItemsToOverflowMenu(overflowItem, item);
+						}
 					}
```
hishitetsu, line 379: I don't think this change is necessary.

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -161,11 +161,9 @@ public async Task<ReturnResult> DeleteItemsAsync(IEnumerable<IStorageItemWithPat
...
 			if (!permanently && registerHistory)
 				App.HistoryWrapper.AddHistory(history);
-
-			var itemsDeleted = history?.Source.Count ?? 0;
-
-			// Remove items from jump list
-			source.ForEach(async x => await jumpListService.RemoveFolderAsync(x.Path));
+			
+			// Await all removal tasks to complete concurrently
+			await Task.WhenAll(source.Select(x => jumpListService.RemoveFolderAsync(x.Path)));
```
hishitetsu, line 166: I'm not sure if we should await this. This is deleting items from the jump list, which is different from deleting the items itself, so it may be performed asynchronously.

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -476,8 +474,8 @@ public async Task<ReturnResult> MoveItemsAsync(IEnumerable<IStorageItemWithPath>
 				App.HistoryWrapper.AddHistory(history);
 			}
 
-			// Remove items from jump list
-			source.ForEach(async x => await jumpListService.RemoveFolderAsync(x.Path));
+			// Await all removal tasks to complete
+			await Task.WhenAll(source.Select(x => jumpListService.RemoveFolderAsync(x.Path)));
```
hishitetsu, line 478: As above.

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -903,22 +903,27 @@ private async Task AddShellMenuItemsAsync(List<ContextMenuFlyoutItemViewModel> s
...
-
-				ShellContextFlyoutFactory.AddItemsToMainMenu(mainItems, x);
-			});
+			// Filter mainShellMenuItems that have a non-null LoadSubMenuAction
+			var mainItemsWithSubMenu = mainShellMenuItems.Where(x => x.LoadSubMenuAction is not null);
 
-			// Add items to overflow shell submenu
-			overflowShellMenuItems.Where(x => x.LoadSubMenuAction is not null).ForEach(async x =>
+			// Load main submenus asynchronously
+			foreach (var item in mainItemsWithSubMenu)
```
hishitetsu, line 910: This should also be done by `Task.WhenAll`.

src/Files.App/Data/Factories/ShellContextFlyoutHelper.cs
gumbarros: It's the same thing, the old line is too long to read at the IDE

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -903,22 +903,28 @@ private async Task AddShellMenuItemsAsync(List<ContextMenuFlyoutItemViewModel> s
...
-			mainShellMenuItems.Where(x => x.LoadSubMenuAction is not null).ForEach(async x =>
-			{
-				await x.LoadSubMenuAction();
+			// Filter mainShellMenuItems that have a non-null LoadSubMenuAction
+			var mainItemsWithSubMenu = mainShellMenuItems.Where(x => x.LoadSubMenuAction is not null);
 
-				ShellContextFlyoutFactory.AddItemsToMainMenu(mainItems, x);
-			});
+			// Load main submenus asynchronously
+			await Task.WhenAll(mainItemsWithSubMenu.Select(item => item.LoadSubMenuAction()));
```
hishitetsu, line 910: `ShellContextFlyoutFactory.AddItemsToMainMenu(mainItems, item);` should also be called in `Task.WhenAll` like overflow items.

src/Files.App/Data/Factories/ShellContextFlyoutHelper.cs
hishitetsu: Then you should divide new L373.

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
gumbarros: Agree

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -161,11 +161,9 @@ public async Task<ReturnResult> DeleteItemsAsync(IEnumerable<IStorageItemWithPat
...
 			if (!permanently && registerHistory)
 				App.HistoryWrapper.AddHistory(history);
-
-			var itemsDeleted = history?.Source.Count ?? 0;
-
-			// Remove items from jump list
-			source.ForEach(async x => await jumpListService.RemoveFolderAsync(x.Path));
+			
+			// Await all removal tasks to complete concurrently
+			_ = Task.WhenAll(source.Select(x => jumpListService.RemoveFolderAsync(x.Path)));
```
hishitetsu, line 166: This comment does not match the actual process.

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -476,8 +474,8 @@ public async Task<ReturnResult> MoveItemsAsync(IEnumerable<IStorageItemWithPath>
 				App.HistoryWrapper.AddHistory(history);
 			}
 
-			// Remove items from jump list
-			source.ForEach(async x => await jumpListService.RemoveFolderAsync(x.Path));
+			// Await all removal tasks to complete
+			_ = Task.WhenAll(source.Select(x => jumpListService.RemoveFolderAsync(x.Path)));
```
hishitetsu, line 478: As above.

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -903,22 +903,27 @@ private async Task AddShellMenuItemsAsync(List<ContextMenuFlyoutItemViewModel> s
...
-
-				ShellContextFlyoutFactory.AddItemsToMainMenu(mainItems, x);
-			});
+			// Filter mainShellMenuItems that have a non-null LoadSubMenuAction
+			var mainItemsWithSubMenu = mainShellMenuItems.Where(x => x.LoadSubMenuAction is not null);
 
-			// Add items to overflow shell submenu
-			overflowShellMenuItems.Where(x => x.LoadSubMenuAction is not null).ForEach(async x =>
+			// Load main submenus asynchronously
+			await Task.WhenAll(mainItemsWithSubMenu.Select(async item =>
```
hishitetsu, line 910: I think `mainItemsWithSubMenu.Select` tasks and `overflowItemsWithSubMenu.Select` tasks can be grouped together and await in one `Task.WhenAll`.

hishitetsu (commented):
Probably my last comments.

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -903,22 +903,26 @@ private async Task AddShellMenuItemsAsync(List<ContextMenuFlyoutItemViewModel> s
...
+			var overflowItemsWithSubMenu = overflowShellMenuItems.Where(x => x.LoadSubMenuAction is not null);
 
-				ShellContextFlyoutFactory.AddItemsToOverflowMenu(overflowItem, x);
+			var overflowSubMenuTasks = overflowItemsWithSubMenu.Select(async item =>
+			{
+				await item.LoadSubMenuAction();
+				ShellContextFlyoutFactory.AddItemsToOverflowMenu(overflowItem, item);
 			});
 
+			await Task.WhenAll(mainSubMenuTasks.Concat(overflowSubMenuTasks));
```
hishitetsu, line 924: Sorry, one more thing.
This awaiting should be put in the end of the method, because the following process doesn't need to wait loading submenus.

hishitetsu (approved):
LGTM 🎉

PR #15033, opened 2024-03-23
Fix: Fixed issue where ftp addresses were not displayed in the path bar

### LINKED ISSUE #15031: Bug: Address bar is empty when opening an ftp server
### Description

image

### Steps To Reproduce

1. Open an ftp server.
2. See the address bar.

### Requirements

Show the ftp server address.

### Files Version

3.3.0.0

### Windows Version

10.0.22631.3296

### Log File

n/a

hishitetsu:
**Resolved / Related Issues**
   Closes #15031 

**Validation**

### DIFF
--- a/src/Files.App/Utils/Storage/Helpers/StorageFileExtensions.cs
+++ b/src/Files.App/Utils/Storage/Helpers/StorageFileExtensions.cs
@@ -1,7 +1,6 @@
 // Copyright (c) 2024 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using FluentFTP.Helpers;
 using System.Collections.Immutable;
 using System.IO;
 using System.Text;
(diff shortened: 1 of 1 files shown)


src/Files.App/Utils/Storage/Helpers/StorageFileExtensions.cs
```diff
@@ -297,7 +297,7 @@ private static PathBoxItem GetPathItem(string component, string path)
 			{
 				title = "SidebarNetworkDrives".GetLocalizedResource();
 			}
-			else if (component.Contains(':', StringComparison.Ordinal))
+			else if (component.EndsWith(":", StringComparison.Ordinal))
```
gumbarros, line 300: I think it's better to use the `char` overload.

`component.EndsWith(':')`

hishitetsu:
```suggestion
			else if (component.EndsWith(':'))
```

gumbarros: Forgot about the `StringComparison.Ordinal` at my example 💀, you can keep it if you want, but I think it make no difference.

hishitetsu: Comparison by char is done in `Ordinal`, so I've modified the code to be simpler as you said👍

PR #15044, opened 2024-03-25
Fix: Fixed an issue where there wasn't space to right click folder in…

### LINKED ISSUE #15012: Feature: Add extra space in column layout for right clicking

Currently when using column layout there is no whitespace to right click in the folder itself especially if it's a parent fodler. I propose that some more whitespace got added to the left or preferably to the bottom of the list. MacOS appears to be having half the list item height for their implementation of the column tree, but I think 2/3 of the line item height would be better.

### Requirements

- This proposal will allow for better accessibility as it gives more space to right click
- It will also make it easier to share folders if usig OneDrive

### Files Version

3.3.0.0

### Windows Version

10.0.22631.3296

### Comments

_No response_

yaira2:
… Column View

**Resolved / Related Issues**
   Closes #15012...

**Validation**
   1. Confirmed there is now enough room to right click folders in the Column View

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Views/Layouts/ColumnLayoutPage.xaml
+++ b/src/Files.App/Views/Layouts/ColumnLayoutPage.xaml
@@ -175,6 +175,7 @@
 			<SemanticZoom.ZoomedInView>
 				<ListView
 					x:Name="FileList"
+					Padding="8,0,8,0"
 					wct:ScrollViewerExtensions.EnableMiddleClickScrolling="{x:Bind IsMiddleClickToScrollEnabled, Mode=OneWay}"
 					x:FieldModifier="public"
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/Layouts/ColumnLayoutPage.xaml
```diff
@@ -199,12 +201,6 @@
 					SelectionMode="Extended"
 					Tapped="FileList_ItemTapped">
 
-					<ListView.ItemContainerTransitions>
```
yaira2, line 202: This is a bit unrelated but it makes the Column View consistent with the Grid View in this regard.

PR #15050, opened 2024-03-26
Feature: Keep MinWidth in ComboBox

### LINKED ISSUE #14965: Bug: ComboBox list shift to the right when clicked on

0x5bfa:
## Summary

As the title.

## PR Checklist

   Closes #14965
   1. Go to Settings dialog
   2. Go to General or Appearance page.
   3. Select another item in ComboBox
   4. See the width changed.

## Screenshots

_None_

### DIFF
--- a/src/Files.App/Extensions/ComboBoxExtensions.cs
+++ b/src/Files.App/Extensions/ComboBoxExtensions.cs
@@ -40,20 +40,18 @@ namespace Files.App.Extensions
 		private static void ComboBox_DropDownClosed(object? sender, object e)
 		{
 			if (sender is ComboBox comboBox &&
-				comboBox.FindDescendant<ContentPresenter>() is ContentPresenter contentPresenter &&
-				contentPresenter.FindDescendant<TextBlock>() is TextBlock textBlock)
+				comboBox.FindDescendant<ContentPresenter>() is ContentPresenter contentPresenter)
 			{
-				comboBox.MinWidth = 12 + textBlock.ActualWidth + 38;
+				comboBox.MinWidth = 12 + contentPresenter.ActualWidth + 38;
 			}
 		}
 
 		private static void ComboBox_SizeChanged(object sender, SizeChangedEventArgs e)
 		{
 			if (sender is ComboBox comboBox &&
-				comboBox.FindDescendant<ContentPresenter>() is ContentPresenter contentPresenter &&
-				contentPresenter.FindDescendant<TextBlock>() is TextBlock textBlock)
+				comboBox.FindDescendant<ContentPresenter>() is ContentPresenter contentPresenter)
 			{
-				comboBox.MinWidth = 12 + textBlock.ActualWidth + 38;
+				comboBox.MinWidth = 12 + contentPresenter.ActualWidth + 38;
 			}
 		}
 	}
--- a/src/Files.App/Extensions/ComboBoxExtensions.cs
+++ b/src/Files.App/Extensions/ComboBoxExtensions.cs
@@ -7,6 +7,12 @@ using Microsoft.UI.Xaml.Controls;
 
 namespace Files.App.Extensions
 {
+	/// <summary>
+	/// Provides extension for ComboBox.
+	/// </summary>
+	/// <remarks>
+	/// - IsKeepWidthEnabled: Prevents from opening popup at wrong position.
+	/// </remarks>
 	public sealed class ComboBoxExtensions : DependencyObject
 	{
 		public static readonly DependencyProperty IsKeepWidthEnabledProperty =
@@ -30,14 +36,14 @@ namespace Files.App.Extensions
 		{
 			if (d is ComboBox comboBox)
 			{
-				comboBox.DropDownClosed -= ComboBox_DropDownClosed;
-				comboBox.DropDownClosed += ComboBox_DropDownClosed;
+				comboBox.SelectionChanged -= ComboBox_SelectionChanged;
+				comboBox.SelectionChanged += ComboBox_SelectionChanged;
 				comboBox.SizeChanged -= ComboBox_SizeChanged;
 				comboBox.SizeChanged += ComboBox_SizeChanged;
 			}
 		}
 
-		private static void ComboBox_DropDownClosed(object? sender, object e)
+		private static void ComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e)
 		{
 			if (sender is ComboBox comboBox &&
 				comboBox.FindDescendant<ContentPresenter>() is ContentPresenter contentPresenter)
--- a/src/Files.App/Extensions/ComboBoxExtensions.cs
+++ b/src/Files.App/Extensions/ComboBoxExtensions.cs
@@ -46,9 +46,10 @@ namespace Files.App.Extensions
 		private static void ComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e)
 		{
 			if (sender is ComboBox comboBox &&
-				comboBox.FindDescendant<ContentPresenter>() is ContentPresenter contentPresenter)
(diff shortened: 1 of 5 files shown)


0x5bfa:
Don't review yet please.
I'm still investigating a way to set width seamlessly.
You can test the behavior but if you re-select smaller item the width won't be changed. This is the last bug I have right now.

0x5bfa:
Ready for review.

yaira2:
@TheChilledBuffalo fyi

yaira2:
> You can test the behavior but if you re-select smaller item the width won't be changed. This is the last bug I have right now.

This seems to still be an issue

src/Files.App/Extensions/ComboBoxExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+				comboBox.SizeChanged += ComboBox_SizeChanged;
+			}
+		}
+
+		private static void ComboBox_SelectionChanged(object sender, SelectionChangedEventArgs e)
+		{
+			if (sender is ComboBox comboBox &&
+				comboBox.FindDescendant<ContentPresenter>() is ContentPresenter contentPresenter)
+			{
+				comboBox.MinWidth = 12 + contentPresenter.ActualWidth + 38;
```
yaira2, line 51: Where are these numbers from and what are they doing?

0x5bfa: Left padding and right padding(including chevron and both sides padding).

0x5bfa:
I'm not sure if this is caused sometimes or in some items but I'm still not able to find a way to retrieve width right after item changed or width changed, using events.

TheChilledBuffalo:
> I'm still not able to find a way to retrieve width right after item changed or width changed, using events.

Yup, this was the exact same issue I was facing when trying to dynamically set a width.

Maybe you guys can go through this thread to get a better idea? https://github.com/microsoft/microsoft-ui-xaml/issues/9009

0x5bfa:
Fixed the issue.

0x5bfa:
@marcelwgn Would you mind if I ask if my implementation is good or there's a rooms to improve? Currently this impelementation have a timing issue where the ComboBox size change and select change is fired before actually XAML is loaded.

marcelwgn:
Sure thing @0x5bfa, I'll take a look!

marcelwgn:
The implementation looks okay (hardcoding the margin isn't ideal though), what surprises me though is why we need to do this in the first place. Something is causing weird layout shifts and I wasn't able to reproduce the weird moving aside of the flyout using the WinUI Gallery:

Video of opening and closing a flyout of a combobox

0x5bfa:
It sometimes goes outside of window bound.

image

image

src/Files.App/Extensions/ComboBoxExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		private static void ComboBox_SizeChanged(object sender, SizeChangedEventArgs e)
+		{
+			if (sender is ComboBox comboBox &&
+				comboBox.FindDescendant<ContentPresenter>() is ContentPresenter contentPresenter)
+			{
+				comboBox.MinWidth = 12 + contentPresenter.ActualWidth + 38;
+			}
+		}
```
marcelwgn, line 63: Why do these two methods not use the exact same code to set min width?

src/Files.App/Extensions/ComboBoxExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+				comboBox.MinWidth = 12 + frameworkElement.ActualWidth + 38;
+			}
+		}
+
+		private static void ComboBox_SizeChanged(object sender, SizeChangedEventArgs e)
+		{
+			if (sender is ComboBox comboBox &&
+				comboBox.FindDescendant<ContentPresenter>() is ContentPresenter contentPresenter)
+			{
+				comboBox.MinWidth = 12 + contentPresenter.ActualWidth + 38;
```
marcelwgn, line 61: We should not be using magic numbers here but a) not have them at all or b) look them up using resources

src/Files.App/Extensions/ComboBoxExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+
+namespace Files.App.Extensions
+{
+	/// <summary>
+	/// Provides extension for ComboBox.
+	/// </summary>
+	/// <remarks>
+	/// - IsKeepWidthEnabled: Prevents from opening popup at wrong position.
```
marcelwgn, line 14: I would say this comment is not entirely correct. There are definitely cases where one wouldn't need this extension. The comment should probably indicate that in fact this extension is setting min width

marcelwgn (commented):
Yeah the issue is, this extension is probably not fixing the issue why this is happening in the first place but rather adding a somewhat expensive workaround. I guess it's fine to do it like that but ideally we should know why ComboBox behaves like that and why it does not in the WinUI Gallery

src/Files.App/Extensions/ComboBoxExtensions.cs
0x5bfa: I think I can remove one of them. One of which didn't get fired on loaded.

src/Files.App/Extensions/ComboBoxExtensions.cs
0x5bfa: I thought the same thing but that's complicated because some of them use Thickness not Double.
We could do like this(at least should add comments) I guess.

src/Files.App/Extensions/ComboBoxExtensions.cs
0x5bfa: We use this for ComboBoxes in Settings, which are aligned Right position. When there's no MinWidth popup opens at ugly position, swapping to the right.

0x5bfa:
> why it does not in the WinUI Gallery
I think the set broarder fixed width?

src/Files.App/Extensions/ComboBoxExtensions.cs
marcelwgn: What you describe is what bug this fixes, the symptom you are adressing. The extension however only sets the min width and the comment should explain exactly that in my opinion.

src/Files.App/Extensions/ComboBoxExtensions.cs
marcelwgn: If we add comments I guess it's fine, however if we restyle the controls one would also need to think about these numbers.

marcelwgn:
> > why it does not in the WinUI Gallery
> 
> I think the set broarder fixed width?

The strange thing is, in the Gallery the min width etc. are not set from what I can see nor did I do that in my testing.

0x5bfa:
@yaira2 This doesn't work only for "Startup settings" ConboBox?

yaira2:
It's not consistent, sometimes it works and other times it doesn't.

0x5bfa:
I'll ask in MUXC team. In the meantime, I'd like to close for now.

0x5bfa:
@marcelwgn Thanks for the tipping. I couldn't find a solution to this... closing.

Issue #12577, opened 2023-06-09
Trace to visualize the path in column view only working in dark mode


py-Alexis:
### Description

I can only see the trace when in dark mode 
![image](https://github.com/files-community/Files/assets/85698747/5a57260d-ebf7-4e94-905f-2ed302f8f9b1)
![image](https://github.com/files-community/Files/assets/85698747/0d3de2b6-31fb-47bf-a48d-c998b5b64774)

It does appear in light mode only very bright colours as background so it's probably only a small coloring issue
![image](https://github.com/files-community/Files/assets/85698747/788910ae-ab8b-4ac2-b33f-e1b76d69f623)
 
and looking at it with a color picker, it seems like there is actually a small difference between the colours
normal background
![image](https://github.com/files-community/Files/assets/85698747/a33eb84f-fc48-416d-9ce2-306d5a427261)
the selected folder
![image](https://github.com/files-community/Files/assets/85698747/2bba33ff-d5ef-41b2-9fc2-359c09432327)

### Steps To Reproduce

turn the theme to light (even closing and reopening Files preview doesn't solve the issue

### Requirements

Change the colour for the previous folders in column view

### Files Version

2.4.85.0

### Windows Version

Windows 10 (22H2): 19045

### Log File

The content however appears when I open the settings and does not seem to be related to changing the theme
[debug.log](https://github.com/files-community/Files/files/11707194/debug.log)

yaira2:
What happens if you set the app to the default theme, and set the Windows theme to light mode?

py-Alexis:
OH I had tried that but without relaunching the app
after putting windows theme to light and relaunching the app it works well
![image](https://github.com/files-community/Files/assets/85698747/c0c5cd0a-e27b-422b-af52-cea9914a44ac)

yaira2:
@chingucoding fyi

yaira2:
@py-Alexis is this still an issue in v3.2?

marcelwgn:
I think the easiest way to solve this issue would be to add an event listener for the `ActualThemeChanged` event so that the `ColumnsLayoutPage` gets notified of theme changes and can apply them. When the theme changes, we then need to reapply the background using the `SetFolderBackground` method of that class.

Josh65-2201:
@py-Alexis Can you see if this is still an issue in the current preview version. https://files.community/download

yaira2:
This is still an issue

R0h1tAnand:
[...]

PR #15096, opened 2024-04-01
Fix: Fixed issue of overlapping brushes leaving gap in the Columns View

marcelwgn:
**Resolved / Related Issues**
   Closes #issue...
Discussed with @yair100 , this PR aims to fix the small gap visible on the items that are part of the path but are not the last item when they are hovered as shown on the screenshot below:
Screenshot of small gap on hovered item

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Views/Layouts/ColumnLayoutPage.xaml.cs
+++ b/src/Files.App/Views/Layouts/ColumnLayoutPage.xaml.cs
@@ -92,8 +92,7 @@ namespace Files.App.Views.Layouts
 				return;
 
 			openedFolderPresenter.Background = new SolidColorBrush(Microsoft.UI.Colors.Transparent);
-			var presenter = openedFolderPresenter.FindDescendant<Grid>()!;
-			presenter!.Background = new SolidColorBrush(Microsoft.UI.Colors.Transparent);
+			SetFolderBackground(openedFolderPresenter, new SolidColorBrush(Microsoft.UI.Colors.Transparent));
 			openedFolderPresenter = null;
 		}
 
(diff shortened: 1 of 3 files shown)


src/Files.App/Views/Layouts/ColumnLayoutPage.xaml.cs
```diff
@@ -157,8 +156,8 @@ private void HighlightPathDirectory(ListViewBase sender, ContainerContentChangin
 			if (args.Item is ListedItem item && columnsOwner?.OwnerPath is string ownerPath
 				&& (ownerPath == item.ItemPath || ownerPath.StartsWith(item.ItemPath) && ownerPath[item.ItemPath.Length] is '/' or '\\'))
 			{
-				var presenter = args.ItemContainer.FindDescendant<Grid>()!;
-				presenter!.Background = this.Resources["ListViewItemBackgroundSelected"] as SolidColorBrush;
+				SetFolderBackground(args.ItemContainer as ListViewItem, this.Resources["ListViewItemBackgroundSelected"] as SolidColorBrush);
```
yaira2, line 159: Is there a way to make this theme aware to fix #12577?

yair100 (approved):
LGTM

src/Files.App/Views/Layouts/ColumnLayoutPage.xaml.cs
marcelwgn: I think, the only option would be to listen to theme changes and based on that reapply the background. Unfortunately, there is no code behind way to create a ThemeResouce binding (at none that I am aware of).

yaira2: Is it worth adding logic for that or should we leave as is since restarting the app will use the correct theme?

yaira2: We've only had one bug report so it's not the highest priority, if it's a simple fix I guess we should move forward but otherwise it's not worth it if it requires a complex workaround.

marcelwgn: I think it's not too hard, we would probably only need an event listener for the actual theme changed event. I'll add a comment to the issue in case anyone else wants to pick it up (and make it easier for newcomers to contribute)

yaira2: Thanks!

PR #15111, opened 2024-04-03
Feature: Switch settings dialog to tab

0x5bfa:
## Summary

As the title.

- [x] test what happens if settings is opened in multiple tabs
  - Working very well!!
- [ ] disable dual pane mode
- [ ] hide the status bar & preview pane
- [ ] create a colored icon to use for the tab
- [ ] settings is currently disabled in compact overlay mode, we'll need to figure out how to handle this
- [ ] decide if we should move the settings icon from the address bar to the sidebar
- [ ] add access keys back to settings menu

## PR Checklist


0x5bfa:
This is still just a proof of concept to test if we can dive into tabs. We plan to use our own custom icons called OpacityIcon, which can be set multiple colors like accent color and secondary/tertiary colors.

PR #15119, opened 2024-04-05
Feature: Added support for filtering current directory when typing into search box

### LINKED ISSUE #14391: Feature: Filter current directory when typing into search box

This feature is an absolute must-have for me: imagine a list of thousands of files, typing just "test" in the filter zone will display all the files or folders containing "test" in their name, it soooo useful, I'm using this search zone dozen times a day!

I'm still using [QTTabBar](http://qttabbar.wikidot.com/), but it's no more maintained, and bugs appear in Win11 23H2 after using ViVeTool for disabling the default Windows tabs.

### Requirements

1. Typing query into the search box should filter the items in the current directory
2. The dropdown menu should keep the existing behavior and continue to display suggestions
3. Pressing enter should start a full search
4. Removing the focus from the search box currently resets the search text, this behavior should be changed so that users can keep the current filter without keeping the focus on the search box.

### Files Version

v3.1

### Windows Version

10.0.22631.2861

### Comments

file-filter

hishitetsu:
**Resolved / Related Issues**
   Closes #14391 

**Validation**

### DIFF
--- a/src/Files.App/Data/Models/ItemViewModel.cs
+++ b/src/Files.App/Data/Models/ItemViewModel.cs
@@ -663,8 +663,10 @@ namespace Files.App.Data.Models
 			EmptyTextType = FilesAndFolders.Count == 0 ? (IsSearchResults ? EmptyTextType.NoSearchResultsFound : EmptyTextType.FolderEmpty) : EmptyTextType.None;
 		}
 
+		public string? FilesAndFoldersFilter { get; set; }
+
 		// Apply changes immediately after manipulating on filesAndFolders completed
-		public async Task ApplyFilesAndFoldersChangesAsync(string? filter = null)
+		public async Task ApplyFilesAndFoldersChangesAsync()
 		{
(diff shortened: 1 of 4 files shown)


yaira2:
This is working well 🎉 

Would it make sense to supporting filtering search results as well?

yaira2:
There are some weird behaviors in the columns view but I don't think it's a blocking issue.

hishitetsu:
I found that sorting after filtering will remove the filtering. I'll fix it.

hishitetsu:
> I found that sorting after filtering will remove the filtering. I'll fix it.

Fixed.

0x5bfa (approved):
Looks good codewise.

yaira2:
@hishitetsu what about grouping?

hishitetsu:
> @hishitetsu what about grouping?

There should be no problem.

yaira2:
Nice work!

PR #15146, opened 2024-04-10
Fix: Fixed issue where open tabs were duplicated when opening a folder from the outside of Files

### LINKED ISSUE #15107: Bug: Opening Files from explorer.exe causes tabs to duplicate
### Description

WIN+E causes tabs to duplicate. If you have 1 tab, now you have 2. If you have 2, now you have 4, and so goes on.

### Steps To Reproduce

1. Opening 3 tabs
2. Using `Win`+`E`
3. The tabs are duplicated

~~Probably~~ this needs the option "Replace file explorer".

### Requirements

When Files is already opened tabs should remain the same.

### Files Version

3.3.0

### Windows Version

Version	10.0.22631 Build 22631

### Log File

[debug.log](https://github.com/files-community/Files/files/14845629/debug.log)

hishitetsu:
**Resolved / Related Issues**
   Closes #15107 

**Validation**

### DIFF
--- a/src/Files.App/ViewModels/MainPageViewModel.cs
+++ b/src/Files.App/ViewModels/MainPageViewModel.cs
@@ -125,15 +125,14 @@ namespace Files.App.ViewModels
 					else if (UserSettingsService.GeneralSettingsService.ContinueLastSessionOnStartUp &&
 						UserSettingsService.GeneralSettingsService.LastSessionTabList is not null)
 					{
-						foreach (string tabArgsString in UserSettingsService.GeneralSettingsService.LastSessionTabList)
+						if (AppInstances.Count == 0)
 						{
-							var tabArgs = CustomTabViewItemParameter.Deserialize(tabArgsString);
-							await NavigationHelpers.AddNewTabByParamAsync(tabArgs.InitialPageType, tabArgs.NavigationParameter);
+							foreach (string tabArgsString in UserSettingsService.GeneralSettingsService.LastSessionTabList)
+							{
+								var tabArgs = CustomTabViewItemParameter.Deserialize(tabArgsString);
+								await NavigationHelpers.AddNewTabByParamAsync(tabArgs.InitialPageType, tabArgs.NavigationParameter);
+							}
 						}
-
-						var defaultArg = new CustomTabViewItemParameter() { InitialPageType = typeof(PaneHolderPage), NavigationParameter = "Home" };
-
-						UserSettingsService.GeneralSettingsService.LastSessionTabList = [defaultArg.Serialize()];
 					}
 					else
 					{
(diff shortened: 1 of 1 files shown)


src/Files.App/ViewModels/MainPageViewModel.cs
```diff
@@ -125,15 +125,14 @@ public async Task OnNavigatedToAsync(NavigationEventArgs e)
...
+							foreach (string tabArgsString in UserSettingsService.GeneralSettingsService.LastSessionTabList)
+							{
+								var tabArgs = CustomTabViewItemParameter.Deserialize(tabArgsString);
+								await NavigationHelpers.AddNewTabByParamAsync(tabArgs.InitialPageType, tabArgs.NavigationParameter);
+							}
 						}
-
-						var defaultArg = new CustomTabViewItemParameter() { InitialPageType = typeof(PaneHolderPage), NavigationParameter = "Home" };
-
-						UserSettingsService.GeneralSettingsService.LastSessionTabList = [defaultArg.Serialize()];
```
gumbarros, line 136: Why this is removed?

src/Files.App/ViewModels/MainPageViewModel.cs
```diff
@@ -158,17 +157,14 @@ public async Task OnNavigatedToAsync(NavigationEventArgs e)
...
 						{
 							foreach (string tabArgsString in UserSettingsService.GeneralSettingsService.LastSessionTabList)
 							{
 								var tabArgs = CustomTabViewItemParameter.Deserialize(tabArgsString);
 								await NavigationHelpers.AddNewTabByParamAsync(tabArgs.InitialPageType, tabArgs.NavigationParameter);
 							}
-
-							var defaultArg = new CustomTabViewItemParameter() { InitialPageType = typeof(PaneHolderPage), NavigationParameter = "Home" };
-
-							UserSettingsService.GeneralSettingsService.LastSessionTabList = [defaultArg.Serialize()];
```
gumbarros, line 171: And also here

gumbarros:
The `AppInstances.Count == 0` if makes sense, thank you very much for the fix btw. But why is the `LastSessionTabList `no longer stored?

src/Files.App/ViewModels/MainPageViewModel.cs
hishitetsu: Because it is no more needed. It used to save the current tabs only when the app exited, but now it saves them every time a tab is opened.

hishitetsu:
> But why is the `LastSessionTabList `no longer stored?

It is no longer needed when #14795 is merged.

gumbarros (approved):

src/Files.App/ViewModels/MainPageViewModel.cs
yaira2: I think it's worth keeping for the following scenario

1. Open two windows.
2. Add a few tabs to window _a_.
3. Close window _a_ first, then close window _b_.
4. Upon restarting Files, it will restore the tabs from window _a_, even though window _b_ had the most recent set of tabs.

hishitetsu: Deleting this code doesn't change that behavior. This code was originally intended to open the home page upon restart after a Files crash.

Issue #14572, opened 2024-01-29
Code Quality: Duplicated functionality classes


0x5bfa:
### Description

There're so many duplicated classes that can be merged into each single class.

#### Duplicated classes

Following classes are duplicates and also can be converted into each service class.

- [ ] Recycle bin (**ITrashService**)
  - (helper) RecycleBinHelpers
  - (watcher) RecycleBinManager
- [ ] File tags (**IFileTagsService**)
  - (helper) FileTagsHelper
  - (watcher) FileTagsManager
  - (service) FileTagsService
- [ ] Network drives (**INetworkDrivesService**)
  - (helper) NetworkDrivesAPI
  - (service) NetworkDrivesService
  - (watcher) NetworkDrivesViewModel
- [ ] Removable drives (**IRemovableDrivesService**)
  - (service) RemovableDrivesService
  - (watcher) DrivesViewModel
- [ ] Quick access (**IQuickAccessService**)
  - (service) QuickAccessService
  - (watcher) QuickAccessManager
- [ ] Jump list (**IJumpListDrivesService**)
  - (service) JumpListService
  - (helper) JumpListHelper
- [ ] Widgets (**WidgetService**)
  - (helper) WidgetsHelpers
- [ ] Library (**IWindowsLibraryService**)
  - (watcher) LibraryManager
- [ ] Git (**IGitService**)
  - (helper) GitHelpers

#### Implement watchers

- [ ] IWatcher
- [ ] IFolderWatcher
- [ ] IDeviceWatcher
- [ ] ITrashWatcher

### Gains

- Better maintainability
- Reduced code

### Requirements

- Use single service per a functionality

yaira2:
This looks like a good breakdown, let's split each task into a separate PR for quality and code review purposes.

yaira2:
If someone plans to work on a specific service, please add a comment so we can prevent duplicate PRs.

0x5bfa:
Working on No.12 - **AppThemeModeService**

0x5bfa:
This looks very promising but it seems that the amount of work (even each) is really huge. It turned out that I have to work on base watchers first.

@yaira2 Do you think it is ok if I add a reference of WinUI to Files.App.Storage (Files.WinUI.Storage or Files.Windows.Storage should be more understandable though)?

0x5bfa:
Working on `IAppThemeModeService`.

0x5bfa:
Working on `INetworkDrivesService`.

[...]

PR #15150, opened 2024-04-10
Code Quality: Replace ThemeHelper with AppThemeModeService

0x5bfa:
## Summary

Replacing with singleton pattern service means that we can use interface to conver up the implementation.

## PR Checklist

   Related #14572
   _None_

## Screenshots

_None_


yaira2:
I took a quick glance and this looks good, but we'll need to test these changes before merging.

gumbarros (changes requested):
LGTM

src/Files.App/Dialogs/AddItemDialog.xaml
yaira2, line 1: I used the following steps to test this change
1. Set Files theme to system default
2. Opened the add item dialog (ctrl + shift + i)
3. Change the system theme and confirmed the modal updated in Files
4. Changed the Files theme to the opposite of the system theme
5. Confirmed the dialog was displayed correctly

src/Files.App/ViewModels/SettingsViewModel.cs
yaira2, line 1: Can we delete this ViewModel?

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
```diff
@@ -97,36 +96,20 @@ private void UpdatePageLayout()
...
+		private async void AppThemeModeService_AppThemeModeChanged(object? sender, EventArgs e)
 		{
 			if (Parent is null)
 				return;
 
 			await DispatcherQueue.EnqueueOrInvokeAsync(() =>
 			{
-				((Frame)Parent).RequestedTheme = ThemeHelper.RootTheme;
-
-				switch (ThemeHelper.RootTheme)
```
yaira2, line 109: Are the title bar buttons still displayed correctly even if the theme is changed at runtime?

src/Files.App/Services/ResourcesService.cs
```diff
@@ -45,7 +47,7 @@ public void SetAppThemeFontFamily(string contentControlThemeFontFamily)
 		/// <inheritdoc/>
 		public void ApplyResources()
 		{
-			ThemeHelper.ApplyResources();
+			AppThemeModeService.RefreshAppThemeMode();
```
yaira2, line 50: Is there a reason the method was renamed?

src/Files.App/ViewModels/Settings/AppearanceViewModel.cs
```diff
@@ -20,6 +21,7 @@ public AppearanceViewModel(IUserSettingsService userSettingsService, IResourcesS
 		{
 			UserSettingsService = userSettingsService;
 			ResourcesService = resourcesService;
+			selectedThemeIndex = (int)Enum.Parse(typeof(ElementTheme), AppThemeModeService.AppThemeMode.ToString());
```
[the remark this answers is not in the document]
0x5bfa: Done,

src/Files.App/ViewModels/SettingsViewModel.cs
0x5bfa: Yeah, but me might need it for the Settings tabs.

src/Files.App/Views/Properties/MainPropertiesPage.xaml.cs
0x5bfa: Yes, absolutely.
When the Combo box invokes theme changed event, properties window also apply the theme through the same method without invoking theme changed event.

src/Files.App/Dialogs/AddItemDialog.xaml
0x5bfa: Yes, tested locally and works very well.
Binding a helper from XAML is not a good idea if we use service because it should be static. So I changed to the way used in SettingsDialog.xaml

src/Files.App/Services/ResourcesService.cs
0x5bfa: Yes, this method does toggle themes between dark and light and then change to the right one. It means to refresh. It's never been apply 'resource'.

yaira2: It's not really refreshing anything, it's a workaround to force WinUI to apply the updated resource (see https://github.com/microsoft/microsoft-ui-xaml/issues/4651). `ApplyTheme` is fine with me but I prefer resources since its a little more generic.

src/Files.App/ViewModels/SettingsViewModel.cs
yaira2: Do you have something in mind?

src/Files.App/Services/ResourcesService.cs
0x5bfa: > any bugs
I knew that...

Is it ok if I add xml comments that mention this issue since method should not exist at the first place?

yaira2: Definitely, I think the interface would be the best location. In the meantime, please revert to `ApplyResources` or `ApplyTheme`.

src/Files.App/ViewModels/SettingsViewModel.cs
0x5bfa: I would convert selection change event into SelectedItem or something and moreover we might be going to use SidebarView instead, which requires more logic in view model such as generating items instead of hard coding in xaml.

yaira2: Isn’t there a different ViewModel for that? SettingsPageViewModel?

0x5bfa: I'll convert SettingsDialog into MainSettingsPage(just change base class) so SettingsViewModell will be MainSettingsViewModel.

src/Files.App/Services/ResourcesService.cs
0x5bfa: What about this one?

```cs
/// Summary:
/// Forces to load the current application theme mode.
/// Remarks:
/// This is a workaround for a WinUI bug #000.
public void ForceToLoadAppThemeMode();
```

yaira2: It's not specific to themes, we used to use this for compact spacing as well.

0x5bfa: I mean for IAppThemeModeService.ForceToLoadAppThemeMode();
For ResourcesService, it's good as is(ApplyResources).

yaira2: I prefer keeping the original naming

0x5bfa: image

yaira2 (approved):
LGTM

PR #15155, opened 2024-04-11
Code Quality: Removed NetworkDrivesViewModel and NetworkDrivesAPI

0x5bfa:
## Summary

Now that evry code related network share point is in the service class

## PR Checklist

   Related #14572
   _None_

## Screenshots

_None_


yaira2 (dismissed):

src/Files.App/Data/Models/ItemViewModel.cs
```diff
@@ -50,6 +50,7 @@ public sealed class ItemViewModel : ObservableObject, IDisposable
 		private readonly IJumpListService jumpListService = Ioc.Default.GetRequiredService<IJumpListService>();
 		private readonly IDialogService dialogService = Ioc.Default.GetRequiredService<IDialogService>();
 		private IUserSettingsService UserSettingsService { get; } = Ioc.Default.GetRequiredService<IUserSettingsService>();
+		private INetworkDrivesService NetworkDrivesService = Ioc.Default.GetRequiredService<INetworkDrivesService>();
```
hishitetsu, line 53:
```suggestion
		private readonly INetworkDrivesService NetworkDrivesService = Ioc.Default.GetRequiredService<INetworkDrivesService>();
```

src/Files.App/UserControls/Widgets/DrivesWidget.xaml.cs
```diff
@@ -23,7 +23,7 @@ public sealed partial class DrivesWidget : BaseWidgetViewModel, IWidgetViewModel
 		public IUserSettingsService userSettingsService { get; } = Ioc.Default.GetRequiredService<IUserSettingsService>();
 		private IHomePageContext HomePageContext { get; } = Ioc.Default.GetRequiredService<IHomePageContext>();
 		private DrivesViewModel drivesViewModel = Ioc.Default.GetRequiredService<DrivesViewModel>();
```
hishitetsu, line 25:
```suggestion
		private readonly DrivesViewModel drivesViewModel = Ioc.Default.GetRequiredService<DrivesViewModel>();
```

src/Files.App/UserControls/Widgets/DrivesWidget.xaml.cs
```diff
@@ -23,7 +23,7 @@ public sealed partial class DrivesWidget : BaseWidgetViewModel, IWidgetViewModel
...
-		private NetworkDrivesViewModel networkDrivesViewModel = Ioc.Default.GetRequiredService<NetworkDrivesViewModel>();
+		private INetworkDrivesService NetworkDrivesService = Ioc.Default.GetRequiredService<INetworkDrivesService>();
```
hishitetsu, line 26:
```suggestion
		private readonly INetworkDrivesService NetworkDrivesService = Ioc.Default.GetRequiredService<INetworkDrivesService>();
```

src/Files.App/Services/NetworkDrivesService.cs
```diff
@@ -1,73 +1,208 @@
...
+using Vanara.InteropServices;
 using Vanara.PInvoke;
 using Vanara.Windows.Shell;
+using static Vanara.PInvoke.AdvApi32;
+using static Vanara.PInvoke.Mpr;
 
 namespace Files.App.Services
 {
-	public sealed class NetworkDrivesService : INetworkDrivesService
+	public sealed class NetworkDrivesService : ObservableObject, INetworkDrivesService
```
hishitetsu, line 14: Is it ok for a service class to be an ObservableObject? I think service classes should be completely separated from views and observable properties should be held by view models.

0x5bfa: We have some properties like this such as Removable drives, Quick access folders,  recent files/folders, wsl drives, network drives.

What do you think an idea to create a comprehensive view model that holds such drives and folders?

hishitetsu: Ideally, a view model should exist tied to a view, and one should avoid creating a view model that is inconsistent with the view's lifecycle.

0x5bfa: Those drives and folders are used only for sidebars and widgets. Sidebar has the longest lifecycle, which ends when user kills the process, so making that view model singleton pattern is totally fine. while recent files are used only for widgets, the view model of the recent widget user control is singleton pattern.

hishitetsu: I still don't think the ViewModel should be registered as a singleton because the lifecycle of SideBar may change in the future.
Anyway, I don't think it is a good idea to make NetworkDrivesService an ObservableObject.

src/Files.App/Services/NetworkDrivesService.cs
```diff
@@ -1,73 +1,208 @@
...
+					MenuOptions = new()
+					{
+						IsLocationItem = true,
+						ShowShellItems = true,
+						ShowProperties = true,
+					},
 				};
 
+				networkItem.MenuOptions.ShowEjectDevice = networkItem.IsRemovable;
+
```
hishitetsu, line 102: I don't think this change makes much sense if `ShowEjectDevice` needs to be set separately.

hishitetsu:
Aside from the comments, there doesn't seem to be any regression from this change.

src/Files.App/Services/NetworkDrivesService.cs
hishitetsu: Ok, I'm fine with that.

src/Files.App/Services/NetworkDrivesService.cs
0x5bfa: Done!

hishitetsu (approved):
LGTM

Issue #15141, opened 2024-04-08
Feature: Support custom args for actions


yaira2:
### What feature or improvement do you think would benefit Files?

With the introduction of custom key-bindings in version 3.4, it's a good time to start considering how we want to implement support for custom arguments in actions.

### Requirements

- Adjust actions to support custom args
- Support adding args via the actions page
- Add support for renaming actions to help users differentiate between multiple actions of the same type

### Files Version

v3.3

### Windows Version

Windows 11

yaira2:
A good use case for this is applying custom tags (see #11352)

yaira2:
One thing to consider is how to handle validation. Should there just be a TextBox for inputting args or should it be more specific (eg. a dropdown with a list of tags)?

0x5bfa:
The latter should be more secure but needs to store a collection(it won' be a big deal I assume).

yaira2:
Secure as in making sure it's a valid arg?

0x5bfa:
Yes

yaira2:
This feature would be targeting advanced users but it's still something to consider.

yaira2:
We've discussed this internally and we're leaning towards including validation, we will probably need the following types, but they can be added as needed.
- path
- file extension
- tag name
- username (for RunAsUser)
- archive format (for compressing files)

PR #15167, opened 2024-04-13
Code Quality: Introduced IActionsSettingsService

0x5bfa:
## Summary

We're likely to expect users to customize key binding, such as changing default key binding or adding args for a specific action; therefore, we have to create another setting file for it and get them put customized actions as json.

## PR Checklist

   Related: #15141
   _None_

## Screenshots

_None_

### DIFF
--- a/src/Files.App/Constants.cs
+++ b/src/Files.App/Constants.cs
@@ -147,6 +147,8 @@ namespace Files.App
 			public const string UserSettingsFileName = "user_settings.json";
 
 			public const string FileTagSettingsFileName = "filetags.json";
+
+			public const string ActionsSettingsFileName = "user_settings.actions.json";
 		}
 
 		public static class PreviewPane
--- a/src/Files.App/Views/Settings/ActionsPage.xaml.cs
+++ b/src/Files.App/Views/Settings/ActionsPage.xaml.cs
@@ -12,7 +12,7 @@ namespace Files.App.Views.Settings
 {
 	public sealed partial class ActionsPage : Page
 	{
-		private IGeneralSettingsService GeneralSettingsService { get; } = Ioc.Default.GetRequiredService<IGeneralSettingsService>();
+		private IActionsSettingsService ActionsSettingsService { get; } = Ioc.Default.GetRequiredService<IActionsSettingsService>();
 
 		private readonly string PART_EditButton = "EditButton";
 		private readonly string NormalState = "Normal";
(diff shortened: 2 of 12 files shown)


src/Files.App/Constants.cs
```diff
@@ -147,6 +147,8 @@ public static class LocalSettings
 			public const string UserSettingsFileName = "user_settings.json";
 
 			public const string FileTagSettingsFileName = "filetags.json";
+
+			public const string ActionsSettingsFileName = "user_settings.actions.json";
```
yaira2, line 151: Do we need a separate file? I can see a use case for keeping it in user_settings.json to make it easier for users who want to manually modify the file.

src/Files.App/Data/Items/ModifiableActionItem.cs
```diff
@@ -0,0 +1,59 @@
+﻿// Copyright (c) 2024 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.App.Data.Items
+{
+	/// <summary>
+	/// Represents item for a hotkey that is registered for a <see cref="IRichCommand"/>.
+	/// </summary>
+	public class ModifiableActionItem : ObservableObject
```
yaira2, line 9: Should there be a property to set if customs args are supported?

yaira2:
Editing appears to be broken

src/Files.App/Constants.cs
0x5bfa: I thought it will be a large list. If user edits a different property and breaks somewhat, the app is likely to reset all settings and start over from the beginning. The list won't be such a place I guess.


But I will use the same file.

src/Files.App/Data/Items/ModifiableActionItem.cs
0x5bfa: Added.

0x5bfa:
I’ll test again later.

src/Files.App/Constants.cs
yaira2: Thank you

yaira2:
I've done extensive testing and aside from the issue listed in the description, this is good to go behavior wise.

yaira2 (dismissed):

0x5bfa:
## Issues

- The command palette isn't displaying shortcuts for a number of actions (this isn't consistent but I can usually repro by changing the query a couple times)
- When adding a new command, the save button is enabled even before a keybinding is set.

0x5bfa:
I used VS Source Control AI for generating commit message as below:

#### Refactor key bindings and remove unused properties
This commit includes several changes. In `CommandManager.cs`, the `continue` statement was replaced with `command.RestoreKeyBindings()`, ensuring default key bindings are restored when there are no custom ones. The properties `SecondaryDisplay` and `SupplementaryDisplay` were removed from `NavigationBarSuggestionItem.cs` and consequently from `AddressToolbar.xaml` and `AddressToolbarViewModel.cs`. This suggests these properties are no longer needed. In `AddressToolbarViewModel.cs`, the `HotKeys` property is now set with a new instance of the `HotKeys` from the command and suggestions, ensuring it has its own copy and is not referencing the same object.

src/Files.App/Views/Settings/ActionsPage.xaml.cs
yaira2, line 1: There seems to be a regression where adding a new command doesn't focus the TextBox.

yaira2: This is because we only switch the focus when `KeyBindingEditorTextBox` is enabled. But with this PR, an item is preselected so it's always enabled.

yaira2: I pushed a fix in 9b10646e8e7e78b20bd24b6d846ea7a899f831dc

yaira2 (approved):

PR #15131, opened 2024-04-07
Code Quality: Migrated CopyFileFromApp and MoveFileFromApp to CsWin32

gumbarros:
**Resolved / Related Issues**
- [X] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
   Part of #14505, #15000 

**Validation**
How did you test these changes?
- [X] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [ ] Did you implement any design changes to an existing feature?
   - [ ] Was this change approved?
- [ ] Are there any other steps that were used to validate these changes?

Issue #15000, opened 2024-03-19
Code Quality: SYSLIB1054 - Use LibraryImportAttribute instead of DllImportAttribute to generate p/invoke marshalling code at compile time.


gumbarros:
### Description

Related to https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib1050-1069.

### Concerned code

```
Found issues (67 issues)
	Found usages (67 issues)
		<src>\<core> (4 issues)
			<src>\<core>\<Files.Core> (4 issues)
				<src>\<core>\<Files.Core>\Helpers (4 issues)
					<src>\<core>\<Files.Core>\Helpers\NativeFindStorageItemHelper.cs (4 issues)
						Mark the method 'FindFirstFileExFromApp' with 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time (0 issues)
						Mark the method 'FindNextFile' with 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time (0 issues)
						Mark the method 'FindClose' with 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time (0 issues)
						Mark the method 'FileTimeToSystemTime' with 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time (0 issues)
		<src>\<platforms> (63 issues)
			<src>\<platforms>\<Files.App> (63 issues)
				<src>\<platforms>\<Files.App>\Utils\Shell (3 issues)
					<src>\<platforms>\<Files.App>\Utils\Shell\PreviewHandler.cs (1 issue)
						Mark the method 'CoCreateInstance' with 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time (0 issues)
					<src>\<platforms>\<Files.App>\Utils\Shell\ItemStreamHelper.cs (2 issues)
						Mark the method 'SHCreateStreamOnFileEx' with 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time (0 issues)
						Mark the method 'SHCreateItemFromParsingName' with 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time (0 issues)
				<src>\<platforms>\<Files.App>\Services (1 issue)
[...]

Issue #14505, opened 2024-01-21
Code Quality: Remove Vanara and DllImport & use CsWin32


0x5bfa:
### Description

Since I introduced CsWin32 into Files app because of incapability of Vanara being trimmed, we have not been able to reduce Vanara referenced code.

### Concerned code

- Vanara references
- Win32PInvoke class

### Gains

- Being able to trim Win32 accessing parts

### Requirements

- Remove Vanara reference is the ultimate goal
- Removing DllImport is as important as above

PR #15175, opened 2024-04-14
Code Quality: Migrated `CreateDirectoryFromApp` to CsWin32 

gumbarros:
**Resolved / Related Issues**
   Part of #14505, #15000 

**Validation**

**Additional comments**
Also removed unused methods from a previous PR #15131

### DIFF
--- a/src/Files.App/Utils/Serialization/Implementation/DefaultSettingsSerializer.cs
+++ b/src/Files.App/Utils/Serialization/Implementation/DefaultSettingsSerializer.cs
@@ -1,9 +1,9 @@
 // Copyright (c) 2024 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using Files.Shared.Extensions;
-using System;
 using System.IO;
+using Windows.Win32;
+using Windows.Win32.Security;
 using static Files.App.Helpers.NativeFileOperationsHelper;
 
 namespace Files.App.Utils.Serialization.Implementation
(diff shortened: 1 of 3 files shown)


src/Files.App/Utils/Serialization/Implementation/DefaultSettingsSerializer.cs
```diff
@@ -14,7 +14,7 @@ internal sealed class DefaultSettingsSerializer : ISettingsSerializer
 
 		public bool CreateFile(string path)
 		{
-			CreateDirectoryFromApp(Path.GetDirectoryName(path), IntPtr.Zero);
+			PInvoke.CreateDirectoryFromApp(Path.GetDirectoryName(path), new SECURITY_ATTRIBUTES());
 
 			var hFile = CreateFileFromApp(path, GENERIC_READ, FILE_SHARE_READ, IntPtr.Zero, OPEN_ALWAYS, (uint)File_Attributes.BackupSemantics, IntPtr.Zero);
 			if (hFile.IsHandleInvalid())
```
yaira2, line 20: What does `SECURITY_ATTRIBUTES` represent?

0x5bfa: Shouldn't that be nint.Zero or something?

Also, in fileapifromapp.h there should be more params.

Can you send me a screenshot of the definition of the function?

gumbarros: > What does `SECURITY_ATTRIBUTES` represent?

From the docs:
The **lpSecurityDescriptor** member of the structure specifies a security descriptor for the new directory. If *lpSecurityAttributes* is **NULL**, the directory gets a default security descriptor. The ACLs in the default security descriptor for a directory are inherited from its parent directory. The target file system must support security on files and directories for this parameter to have an effect.

Checking the generated source, if null will do the same thing inside the method.

image


We can just pass null or create the default value like I'm doing , what you think?

0x5bfa: Passing null, default or nint.Zero is valid instead.

gumbarros: `nint.Zero` is not valid, passing null instead.

0x5bfa: Thanks!

0x5bfa (approved):

PR #15180, opened 2024-04-15
Fix: Fixed issue where adaptive layout is not disabled when the sync layout setting is on

### LINKED ISSUE #14854: Bug: Adaptive layout is broken
### Description

Starting in v3.2.4, adaptive layout is missing from the right click context menu, and doesn't work when toggled from the layout picker.

### Steps To Reproduce

1. Open layout picker
2. Toggle the option for adaptive
3. Nothing happens

### Requirements

- Correctly apply adaptive layout

### Files Version

v3.2.4

### Windows Version

Windows 11

### Log File

NA

hishitetsu:
**Resolved / Related Issues**
   Closes #14854 

**Validation**
- [X] Did you implement any design changes to an existing feature?
   - [X] Was this change approved?


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3740,4 +3740,7 @@
   <data name="Customized" xml:space="preserve">
     <value>Customized</value>
   </data>
+  <data name="AdaptiveLayoutDisabledNotification" xml:space="preserve">
+    <value>Adaptive layout is not available when the sync option is on.</value>
```
yaira2, line 3744:
```suggestion
    <value>The option for adaptive layout cannot be used while the sync option is enabled.</value>
```

src/Files.App/Views/Settings/LayoutPage.xaml
```diff
@@ -200,5 +210,13 @@
...
 
 		</StackPanel>
+
+		<!--  Adaptive Layout Disabled Teaching Tip  -->
+		<TeachingTip
+			CloseButtonContent="{helpers:ResourceString Name=OK}"
+			IsLightDismissEnabled="True"
+			IsOpen="{x:Bind ViewModel.ShowAdaptiveDisabledTeachingTip, Mode=TwoWay}"
+			Subtitle="{helpers:ResourceString Name=AdaptiveLayoutDisabledNotification}"
+			Target="{x:Bind LayoutTypeComboBox}" />
```
yaira2, line 220: Do you think it would be better to target the sync preferences setting?

hishitetsu: Yes if the teaching tip message is replaced as you suggest, so I have changed the target.

yaira2:
LGTM

PR #15220, opened 2024-04-18
Code Quality: Use registry for preferences and tags storage

### LINKED ISSUE #15203: Bug: Increased crashes in preview version
### Description

It just crashed when I opened a folder, but I was using normally for a time..

Worth stacktrace:
```cs
WinRT
ABI.Files.App.Server.Database
ILayoutPreferencesDatabaseClassMethods.GetPreferences (IObjectReference _obj, String filePath, Nullable`1 frn)
Files.App.Data.Models
LayoutPreferencesManager.GetLayoutPreferencesForPath (String path)
Files.App.Data.Models
LayoutPreferencesManager.GetLayoutType (String path, Boolean changeLayoutMode)
Files.App.Views
HomePage.DrivesWidget_DrivesWidgetInvoked (Object sender, DrivesWidgetInvokedEventArgs e)
Files.App.UserControls.Widgets
DrivesWidget.Button_Click (Object sender, RoutedEventArgs e)
System.Threading.Tasks.Task
WinRT
Microsoft.AppCenter.Utils
```

### Steps To Reproduce

1. Use normally for a time
2. Open a folder
3. Crash

### Requirements

Fix crash that occurs when I open a folder.

### Files Version

3.3.2.0

### Windows Version

10.0.22631.3447

### Log File

[debug.log](https://github.com/files-community/Files/files/15003460/debug.log)

hez2010:
**Resolved / Related Issues**
   Closes #15203

This also brings us a performance gain as we no longer need to scan the whole database to filter tags under a directory. Registry is a tree-based database so we can simply iterate from our desired directory node without unnecessary scan. 

Can confirm that Files is no longer crashing on navigation with this change. 

**Validation**

**Screenshots (optional)**
How the things look like in the registry (virtualized):

image


0x5bfa:
Is this reviewable?

hez2010:
> Is this reviewable?

Yes. It's up for review.

src/Files.App.Server/Utils/RegistryUtils.cs
```diff
@@ -0,0 +1,204 @@
...
+				return;
+			}
+
+			foreach (var p in target.GetType().GetProperties(BindingFlags.Public | BindingFlags.Instance))
+			{
+				if ((p.PropertyType == typeof(string)
+					|| p.PropertyType == typeof(string[])
+					|| p.PropertyType.IsEnum
+					|| p.PropertyType.IsPrimitive
+					|| (p.PropertyType.IsGenericType && p.PropertyType.GetGenericTypeDefinition() == typeof(Nullable<>)))
```
gumbarros, line 37: Even with reflection are you sure it's faster? Can we cache the Reflection somewhere like a `FrozenDictionary`?

hez2010: The engine of LiteDB uses reflection and expression tree to build the query, so we are just moving from reflection to another reflection here.

hez2010: Ideally I would like to use source generator to generate code so that we can avoid reflection at all.

yaira2:
> Users need to export their settings from old version and import settings to the new version to migrate their tags and preferences.

Can we do this automatically?

src/Files.App.Server/Utils/RegistryUtils.cs
```diff
@@ -0,0 +1,204 @@
...
+using System.Runtime.CompilerServices;
+
+namespace Files.App.Server.Utils
+{
+	internal sealed class RegistryUtils
+	{
+		internal static string CombineKeys(string key1, string key2)
+		{
+			key1 = key1.Replace('/', '\\').TrimStart('\\').TrimEnd('\\');
+			key2 = key2.Replace('/', '\\').TrimStart('\\').TrimEnd('\\');
```
0x5bfa, line 12: I think you can just use Trim.

src/Files.App/ViewModels/Settings/AdvancedViewModel.cs
```diff
@@ -182,13 +182,13 @@ private async Task ImportSettingsAsync()
...
 					fileTagsSettingsService.ImportSettings(importTags);
-					var fileTagsDB = await zipFolder.GetFileAsync(Path.GetFileName(Server.Database.FileTagsDatabase.GetFileTagsDbPath()));
+					var fileTagsDB = await zipFolder.GetFileAsync("filetags.db");
 					string importTagsDB = await fileTagsDB.ReadTextAsync();
 					var tagDbInstance = FileTagsHelper.GetDbInstance();
 					tagDbInstance.Import(importTagsDB);
 
 					// Import layout preferences and DB
-					var layoutPrefsDB = await zipFolder.GetFileAsync(Path.GetFileName(LayoutPreferencesManager.LayoutSettingsDbPath));
+					var layoutPrefsDB = await zipFolder.GetFileAsync("user_settings.db");
```
0x5bfa, line 191: Can you define in Constants class?

src/Files.App/ViewModels/Settings/AdvancedViewModel.cs
```diff
@@ -230,12 +230,12 @@ private async Task ExportSettingsAsync()
 					await zipFolder.CreateFileAsync(new MemoryStream(exportTags), Constants.LocalSettings.FileTagSettingsFileName, CreationCollisionOption.ReplaceExisting);
 					var tagDbInstance = FileTagsHelper.GetDbInstance();
 					byte[] exportTagsDB = UTF8Encoding.UTF8.GetBytes(tagDbInstance.Export());
-					await zipFolder.CreateFileAsync(new MemoryStream(exportTagsDB), Path.GetFileName(Server.Database.FileTagsDatabase.GetFileTagsDbPath()), CreationCollisionOption.ReplaceExisting);
+					await zipFolder.CreateFileAsync(new MemoryStream(exportTagsDB), "filetags.db", CreationCollisionOption.ReplaceExisting);
```
0x5bfa, line 233: As above.

0x5bfa (commented):
Cool!

src/Files.App.Server/Utils/RegistryUtils.cs
gumbarros: Fair, I think we can do this is another scope, solving the crash is more important :)

0x5bfa:
Can we do this automatically? Resetting the tags list is not an option.


We currently have some needs to support back ward compatibility as below.
Ca we make a class to keep that compatibility something like Files.App.Service.BackwardCompatibilityService?

- Actions
- Tags
- Layout preferences

src/Files.App.Server/Utils/RegistryUtils.cs
hez2010: I have switched to use a source generator for this.

src/Files.App.Server/Utils/RegistryUtils.cs
hez2010: You are right :D

hez2010:
> Can we do this automatically? Resetting the tags list is not an option.

By doing this we cannot get rid of dependencies on LiteDB though.

yaira2:
> > Can we do this automatically? Resetting the tags list is not an option.
> 
> By doing this we cannot get rid of the dependency on LiteDB though.

True, but I think it's a necessity to avoid causing any friction. It shouldn't be for more than a couple of months, after a while we can assume that most users are already updated.

hez2010:
> True, but I think it's a necessity to avoid causing any friction. It shouldn't be for more than a couple of months, after a while we can assume that most users are already updated.

Done. I added support for automatically configuration migration from old versions.

src/Files.App/ViewModels/Settings/AdvancedViewModel.cs
hez2010: Done.

src/Files.App.Server/Utils/RegistryUtils.cs
gumbarros: Congrats, looks awesome!

yaira2 (approved):

hez2010:
Added a schema version in the key so that we will be able to migrate from old versions in case we need to update the structure of registry in the future.

src/Files.App/ViewModels/Settings/AdvancedViewModel.cs
hishitetsu, line 233:
```suggestion
					await zipFolder.CreateFileAsync(new MemoryStream(exportTagsDB), Constants.LocalSettings.FileTagSettingsDatabaseFileName, CreationCollisionOption.ReplaceExisting);
```

src/Files.App/ViewModels/Settings/AdvancedViewModel.cs
```diff
@@ -230,12 +230,12 @@ private async Task ExportSettingsAsync()
...
 
 					// Export layout preferences DB
 					var layoutDbInstance = LayoutPreferencesManager.GetDatabaseManagerInstance();
 					byte[] exportPrefsDB = UTF8Encoding.UTF8.GetBytes(layoutDbInstance.Export());
-					await zipFolder.CreateFileAsync(new MemoryStream(exportPrefsDB), Path.GetFileName(LayoutPreferencesManager.LayoutSettingsDbPath), CreationCollisionOption.ReplaceExisting);
+					await zipFolder.CreateFileAsync(new MemoryStream(exportPrefsDB), "user_settings.db", CreationCollisionOption.ReplaceExisting);
```
hishitetsu, line 238:
```suggestion
					await zipFolder.CreateFileAsync(new MemoryStream(exportPrefsDB), Constants.LocalSettings.UserSettingsDatabaseFileName, CreationCollisionOption.ReplaceExisting);
```

src/Files.App.Server/Utils/RegistryUtils.cs
```diff
@@ -0,0 +1,22 @@
...
+		key1 = key1.Replace('/', '\\').TrimStart('\\').TrimEnd('\\');
+		key2 = key2.Replace('/', '\\').TrimStart('\\').TrimEnd('\\');
+		return $"{key1}\\{key2}";
+	}
+
+	internal static string CombineKeys(string key1, string key2, string key3)
+	{
+		key1 = key1.Replace('/', '\\').TrimStart('\\').TrimEnd('\\');
+		key2 = key2.Replace('/', '\\').TrimStart('\\').TrimEnd('\\');
+		key3 = key3.Replace('/', '\\').TrimStart('\\').TrimEnd('\\');
```
hishitetsu, line 19:
```suggestion
		key1 = key1.Replace('/', '\\').Trim('\\');
		key2 = key2.Replace('/', '\\').Trim('\\');
		return $"{key1}\\{key2}";
	}

	internal static string CombineKeys(string key1, string key2, string key3)
	{
		key1 = key1.Replace('/', '\\').Trim('\\');
		key2 = key2.Replace('/', '\\').Trim('\\');
		key3 = key3.Replace('/', '\\').Trim('\\');
```

hishitetsu (approved):
It works very well! LGTM

hez2010:
Small tweaks around styles, also added an error if the type is not supported by the source generator.
PTAL. @yaira2 

It's ready for merging.

yaira2:
Thank you @hez2010!

PR #15225, opened 2024-04-19
Fix: Fixed an issue where the home key wouldn't move focus when renaming items in List View

### LINKED ISSUE #15048: Bug: the Home key doesn't move to the 1st character when pasting long text from clipboard into file to be renamed
### Description

When renaming a file with a long name in Windows Explorer, the whole name is displayed, whereas Files only displays part of the name. In Files, when the name is in edit mode (F2) the Home & End keys only move to the start & end of the visual/ displayed text rather than the 1st or last character. This issue seems to stem from Files difference in how it displays long file names; generally they are shown in a truncated form whereas Explorer generally shows the whole name.

### Steps To Reproduce

1. say clipboard contains "this is a long file name, this is a long file name" but the layout is List, showing 28 characters.
2. move focus to any file
3. press F2 to rename the file
4. press ctrl+V to paste the new long filename
5. observe only a truncated part of the new long filename is displayed, and pressing Home only moves the cursor to the 1st displayed character rather than (as with Explorer) the 1st actual character

### Requirements

Not quite sure, but its currently confusing. Initially I thought a truncated name was being pasted, but you can use the cursor-left key to see all the other characters are actually pasted, just not displayed.

### Files Version

3.3

### Windows Version

11 23H2

### Log File

this should be optional?

nunescostr:
**Summary**
The Home and end key weren't being acknowledged by the textBox.Keydown on the List View, Grid View and Tile View Layouts. To fix this I replaced it with textBox.PreviewKeyDown, this way the handling is not being consumed before it reaches the controller. I had to implement specific handling for these keys on the switch case already implemented on the GridLayoutPage.xaml.cs.

**Resolved / Related Issues**
   Closes #15048

**Validation**
   None.

**Screenshots (optional)**
None.


src/Files.App/Views/Layouts/BaseGroupableLayoutPage.cs
```diff
@@ -252,7 +252,7 @@ protected virtual void StartRenameItem(string itemNameTextBox)
 
 			textBox.Focus(FocusState.Pointer);
 			textBox.LostFocus += RenameTextBox_LostFocus;
-			textBox.KeyDown += RenameTextBox_KeyDown;
+			textBox.PreviewKeyDown += RenameTextBox_KeyDown;
```
hishitetsu, line 255: Since the event is unsubscribed when the renaming is finished, it is necessary to change that to `PreviewKeyDown` as well.

src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
```diff
@@ -345,7 +345,7 @@ override public void StartRenameItem()
 
 			textBox.Focus(FocusState.Pointer);
 			textBox.LostFocus += RenameTextBox_LostFocus;
-			textBox.KeyDown += RenameTextBox_KeyDown;
+			textBox.PreviewKeyDown += RenameTextBox_KeyDown;
```
hishitetsu, line 348: As above.

hishitetsu (commented):
Pressing the Home key several times now returns the cursor to the top of the textbox, but pressing it once still only returns the cursor to the top of the visible area.

src/Files.App/Views/Layouts/BaseGroupableLayoutPage.cs
yaira2: In theory, we should be able to check the logic for the Details View and copy from there.

yaira2:
@nunescostr please see the requested changes

src/Files.App/Views/Layouts/BaseGroupableLayoutPage.cs
```diff
@@ -321,6 +321,14 @@ protected async void RenameTextBox_KeyDown(object sender, KeyRoutedEventArgs e)
 				case VirtualKey.Right:
 					e.Handled = (textBox.SelectionStart + textBox.SelectionLength) == textBox.Text.Length;
 					break;
+				case VirtualKey.Home:
+					textBox.SelectionStart = 0;
```
yaira2, line 325:
```suggestion
		if (!isShiftPressed)
					textBox.SelectionStart = 0;
```

src/Files.App/Views/Layouts/BaseGroupableLayoutPage.cs
```diff
@@ -321,6 +321,14 @@ protected async void RenameTextBox_KeyDown(object sender, KeyRoutedEventArgs e)
...
+					e.Handled = true;
+					break;
+				case VirtualKey.End:
+					textBox.SelectionStart = textBox.Text.Length;
```
yaira2, line 329:
```suggestion
		if (!isShiftPressed)
					textBox.SelectionStart = textBox.Text.Length;
```

src/Files.App/Views/Layouts/BaseGroupableLayoutPage.cs
yaira2: @hishitetsu do you agree with this suggestion? (see #13847 for reference)

src/Files.App/Views/Layouts/BaseGroupableLayoutPage.cs
yaira2: I pushed a commit with this change

src/Files.App/Views/Layouts/BaseGroupableLayoutPage.cs
hishitetsu: Yes, but indentation should be changed.

yaira2 (approved):

hishitetsu (commented):
I have found this PR makes it impossible to select text with Shift+Up or Shift+Down.

yaira2:
Does changing back to `KeyDown` resolve that issue?

hishitetsu:
KeyDown will make the selection, but only to the middle of the text box, not to the beginning or end of the text box in case of Grid/Tiles/List view. Also, the original issue will no longer be fixed.

0x5bfa:
@nunescostr Could you address those issues above? Let me know if you hardly have time to tackle with them.

nunescostr:
Right now I am struggling to find the time, but I can look into this closer to next month's time

yaira2:
@nunescostr are you still planning to work on this?

PR #15226, opened 2024-04-19
Code Quality: Disable composite ReadyToRun for server

hez2010:
Found some random crashes only happens with composite R2R. Disable it for now.


hez2010:
Seems not related. Closing.

PR #15229, opened 2024-04-19
Code Quality: Various fixes to the WinRT server call

hez2010:
**Resolved / Related Issues**

- Do not call `RoUninitialize` as this will be automatically done on the process exit
- Correct the path for `tag:*` to `Home` while getting the layout preferences
- Kill hanging server at startup

**Validation**
   1. Open app ...
   2. Click settings button ...

**Screenshots (optional)**
Add screenshots here.

### DIFF
--- a/src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
+++ b/src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
@@ -506,9 +506,7 @@ namespace Files.App.Data.Models
 				path = path.TrimPath() ?? string.Empty;
 
 				if (path.StartsWith("tag:", StringComparison.Ordinal))
-				{
 					return GetLayoutPreferencesFromDatabase("Home", null);
-				}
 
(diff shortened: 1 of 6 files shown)


0x5bfa (approved):
As to codebase quality, lgtm.

src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
```diff
@@ -505,6 +505,11 @@ public static void SetLayoutPreferencesForPath(string path, LayoutPreferencesIte
 			{
 				path = path.TrimPath() ?? string.Empty;
 
+				if (path.StartsWith("tag:", StringComparison.Ordinal))
+				{
+					return GetLayoutPreferencesFromDatabase("Home", null);
+				}
```
hishitetsu, line 511:
```suggestion
				if (path.StartsWith("tag:", StringComparison.Ordinal))
					return GetLayoutPreferencesFromDatabase("Home", null);
```

hishitetsu (approved):
LGTM except for one small coding convention violation.

PR #15231, opened 2024-04-19
Code Quality: Introduced Files.App.Win32 to avoid duplication

0x5bfa:
## Summary

To avoid duplicating the same generated code by CsWin32.

## PR Checklist

   _None_

## Screenshots

_None_


src/Files.App.Win32/NativeMethods.json
```diff
@@ -0,0 +1,4 @@
+{
+	"$schema": "https://aka.ms/CsWin32.schema.json",
+	"public": true
+}
```
hez2010, line 4: Can you add `"allowMarshaling": false` as well? We want to make sure we don't use any runtime marshalling, especially in the server.

src/Files.App.Win32/NativeMethods.txt
```diff
@@ -44,4 +38,33 @@ MoveFileFromApp
...
 GetKeyState
-CreateDirectoryFromApp
\ No newline at end of file
+CreateDirectoryFromApp
+RoRegisterActivationFactories
+RoRevokeActivationFactories
+WindowsCreateString
+WindowsDeleteString
+RoInitialize
+RoUninitialize
```
hez2010, line 47: This is no longer needed after #15229

hez2010 (changes requested):
Otherwise LGTM as long as the project builds.

src/Files.App.Win32/NativeMethods.json
0x5bfa: I added it in the file in Server project.
Do we want to override the setting for every project?

src/Files.App.Win32/NativeMethods.txt
0x5bfa: Alr!

src/Files.App.Win32/NativeMethods.json
hez2010: I don't think you need to add this configuration in other projects. The generated types will be public in Files.App.Win32, and can be consumed directly by other projects.

0x5bfa: Sure

0x5bfa: public: true
doesn't work...

0x5bfa:
I’m not sure

Issue #13116, opened 2023-08-01
Bug: Pinned section is empty


smokelore:
> [!important]
> A potential fix is available in the preview builds https://files.community/download. Please comment if it worked or not for you.
> If not you can try clear recent folders in Windows File Explorer settings > ... > Options > Clear and disabling `show recent folders` in > that same menu.

### Description

![image](https://github.com/files-community/Files/assets/1188899/ec104937-3cfa-4241-b774-600f32b5754a)

I installed Filed Preview by accident and it was working quite well until I realized that the Favorites dropdown in the Sidebar is always empty. I uninstalled Files Preview and installed Files Stable and the problem remains. If I add more items to Favorites they do not show up. They're also not present in the Quick Access section of the Home page. I've tried hiding the Favorites section and unhiding it but that didn't resolve it either.

![image](https://github.com/files-community/Files/assets/1188899/767f42e1-3380-47ad-8d02-bb49c26614e4)

Really like this software so far and would love to continue using it if I can get this working. Thanks everyone!

### Steps To Reproduce

Use Files Preview as normal, moving files, renaming folders, running applications, adding and removing Favorites. Favorites then became empty. I really don't know what else I might have done.

### Requirements

Expected behavior is that I'm able to see my Favorites in the sidebar, and remove them, add more of them, and reorder them as normal.

### Files Version

2.5.21.0

### Windows Version

Edition: Windows 11 Pro, Version: 21H22, OS Build: 2000.2245

### Log File

[debug.log](https://github.com/files-community/Files/files/12233858/debug.log)

Josh65-2201:
Thanks for the report, Can you send what in your favorites from Windows File Explorer.

smokelore:
Here you go, @Josh65-2201. Thanks for the prompt reply.

![image](https://github.com/files-community/Files/assets/1188899/c4a0d4f6-1827-4b28-b219-c02280872c64)

smokelore:
This install of Windows is only a couple days old so my Favorites menu hasn't gotten cluttered yet, haha

yaira2:
@smokelore can you see what happens if you remove Google Drive from Quick Access in File Explorer?

smokelore:
![image](https://github.com/files-community/Files/assets/1188899/ff8dd998-c016-440a-9b8f-1711af9f869a)

Good idea, but it didn't fix it even when I removed everything except for Downloads and Desktop from Quick Access.

smokelore:
[...]

Issue #13087, opened 2023-07-27
Bug: Favorites take a long time to load if there are offline network drives


333csx:
### Description

After opening a new Files window it takes about 2 minutes to show the Favourites and the Quick acces items.
This issue is not due to a slow CPU or insufficient RAM.
But I should mention, that I deactivated search indexing on my machine - maybe this could be the reason.

### Steps To Reproduce

1. Pin offline network drives to the sidebar
2. Open Files

### Requirements

Show Favourites and Quick access items imediately

### Files Version

2.5.21.0

### Windows Version

10.0.19045.3208

### Log File

[debug.log](https://github.com/files-community/Files/files/12183454/debug.log)

Josh65-2201:
Thanks for the report. What is on your favorites list?

333csx:
In my favorites list there are 17 local folders and 4 network shares that are currently offline.

yaira2:
Can you see what happens if you remove the network shares? I'm curious if it's a factor.

333csx:
When I remove the network shares favourites and quick acces show imediately.
Imho the network shares should be displayed directly and the connection should only be checked when they are needed or the PC is idle.

yaira2:
Thanks for the feedback

Josh65-2201:
@333csx Is this still an issue on v4.0.11 https://files.community/download

Josh65-2201:
@333csx Can you answer the above question

PR #15259, opened 2024-04-27
Code Quality: Improved loading of pinned folders and recent files

hishitetsu:
Currently we load pinned folders and recent files from the quick access shell folder. It contains both pinned folders and recent files, but it causes both loads to be affected when one of them goes wrong.
So I changed it to load pinned folders and recent files separately. This should allow pinned folders to be displayed without having to do workarounds such as disabling recent files. Also, recent files should be displayed before pinned folders have finished loading.

**Resolved / Related Issues**
   Related to #13116 
   Related to #13087 

**Validation**

### DIFF
--- a/src/Files.App/Utils/RecentItem/RecentItems.cs
+++ b/src/Files.App/Utils/RecentItem/RecentItems.cs
@@ -109,7 +109,8 @@ namespace Files.App.Utils.RecentItem
 		/// </summary>
 		public async Task<List<RecentItem>> ListRecentFilesAsync()
 		{
-			return (await Win32Helper.GetShellFolderAsync(QuickAccessGuid, false, true, 0, int.MaxValue)).Enumerate
+			// Since the maximum number of recent files is 20, we set the count to 20 to avoid loading of unnecessary shell items.
+			return (await Win32Helper.GetShellFolderAsync(QuickAccessGuid, false, true, 0, 20)).Enumerate
 				.Where(link => !link.IsFolder)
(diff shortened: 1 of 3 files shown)


src/Files.App/Utils/RecentItem/RecentItems.cs
```diff
@@ -109,7 +109,8 @@ public async Task UpdateRecentFoldersAsync()
 		/// </summary>
 		public async Task<List<RecentItem>> ListRecentFilesAsync()
 		{
-			return (await Win32Helper.GetShellFolderAsync(QuickAccessGuid, false, true, 0, int.MaxValue)).Enumerate
+			// Since the maximum number of recent files is 20, we set the count to 20 to avoid loading of unnecessary shell items.
+			return (await Win32Helper.GetShellFolderAsync(QuickAccessGuid, false, true, 0, 20)).Enumerate
```
hishitetsu, line 113: This change is not perfect because there are not always 20 recent files. But I don't know how to find out the number of recent files without affecting performance, so I fixed the number of items to 20.

0x5bfa: I think the number 20 is pretty small, while we might have ways to get count of items in that folder using DirectoryInfo or FindFirstFile/FindNextFile functions.

yaira2: Fair enough

yaira2 (approved):
LGTM

PR #15280, opened 2024-04-30
Code Quality: Use protocol launch for open and save file dialogs

gave92:
Use protocol launch for open and save file dialogs. Reason: on Windows Preview command line launch is currently broken. Files Launcher also uses protocol instead of command line.

@0x5bfa could you check if this works for you?


0x5bfa (approved):
LGTM both for Save and Open, thank you!
I'm working on UI parts for now.

Actually I rolled my Windows back to the latest stable and alias is also working.

gave92:
Sorry also pushed a small change so file open/save dialog always spins up a separate Files instance.

0x5bfa:
It seems that those dialogs are opened as Files app process, but they probably should be opened in the host window process, such as Notepad or Edge? Is that possible to open as a child process?

gave92:
Maybe with CreateProcess instead of ShellExecute but then I can't use protocol launch.

0x5bfa:
It must be an issue with Windows and after having gone back from Insider to Stable, it just fixed. Maybe that version was malfunctioned. In that case, can we use alias launch as the primary way of launching and use protocol launch as another way of launching in case of failing.

I mean grouped in task bar. It currently opens in Files task bar icon.

gave92:
Aah will try with https://stackoverflow.com/questions/13924264/grouping-separate-processes-in-the-windows-taskbar

gave92:
Yayy sort of:
image

0x5bfa:
Wow, it's amazing!

I believe yes!

PR #15284, opened 2024-05-01
Code Quality: Remove context flyout builder

0x5bfa:
## Summary

This is also an prepatation PR for the IRichCommand parameter introduction.

In the effort to introduction of IRichCommand parameter, it'd be better to remove duplication and provide single way of generating ContextFlyoutItems.

- `ContextMenuFlyoutItemType` to `ContextFlyoutItemType`
- Removed `ContextMenuFlyoutItemViewModelBuilder`
- `ContextMenuFlyoutItemViewModel` to `ContextFlyoutItemModel`
- Added XML comments in `ContextFlyoutItemModel`
- Added some constructors to adopt with IRichCommand in `ContextFlyoutItemModel`
- Converted `OpacityIconModel` from struct to class to use constructor
- ~~Use `IsVisible` instead of `Show` to be more boolean-tic~~
- ~~Removed `ID` and use `Tag` instead.~~

## PR Checklist

   1. Go to layout page
   2. Right click (check if shown correctly)
   3. Select item and right click (check again)
   4. Right click Sidebar item (again)
   5. Go to Home
   6. Right click Widget item (again)

## Screenshots

_None_


src/Files.App/Data/Models/ContextFlyoutItemModel.cs
```diff
@@ -0,0 +1,117 @@
...
+
+			bool isExecutable = richCommand.IsExecutable;
+			if (!isExecutable)
+				return;
+
+			Text = richCommand.Label;
+			IsEnabled = isExecutable;
+			IsChecked = richCommand.IsOn;
+			ItemType = type;
+			IsAvailable = true;
```
yaira2, line 86: What does this property do?

src/Files.App/Data/Models/ContextFlyoutItemModel.cs
yaira2, line 1: Can you clarify which changes were made here?

yaira2:
> Are there any other steps that were used to validate these changes

src/Files.App/Data/Models/ContextFlyoutItemModel.cs
0x5bfa: IsAvailable is the value that indicates whether item should be displayed. If true, it will be converted to an UIElement; otherwise, even never included the return value of the generation method. IsVisible is a temporary solution to hide, once it's hidden and is ready to be displayed again, it will be set true.

src/Files.App/Data/Models/ContextFlyoutItemModel.cs
0x5bfa: Sure.
- I moved Build() method content to this class's constructor.
- Renamed ShowItem with IsAvailable
- Renamed IsHidden with IsVisible (inverted values too)

src/Files.App/Data/Models/ContextFlyoutItemModel.cs
yaira2: We need some comments in the code but that's out of scope for now, in the meantime, I find "ShowItem" to be a little more intuitive.

0x5bfa: I'm probably not gonna touch around here again so adding comments would be a better idea.

0x5bfa:
We have two unused boolean properties, can I remove?

ID and SingleItemOnly.

src/Files.App/Data/Models/ContextFlyoutItemModel.cs
0x5bfa: Done

0x5bfa:
Causing many issues....

PR #15298, opened 2024-05-04
Code Quality: Use `LibraryImportAttribute` at undocumented EntryPoint

gumbarros:
**Resolved / Related Issues**
   Part of #15000 

**Validation**
   1. Open app ...
   2. Click settings button ...

**Additional Context**
[#865 is undocumented](https://github.com/microsoft/CsWin32/issues/1170), so it's impossible to use CsWin32.

### DIFF
--- a/src/Files.App/Helpers/Environment/ElevationHelpers.cs
+++ b/src/Files.App/Helpers/Environment/ElevationHelpers.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) 2024 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using System.Runtime.InteropServices;
 using System.Security.Principal;
 
 namespace Files.App.Helpers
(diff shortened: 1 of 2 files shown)


0x5bfa (approved):

src/Files.App/Helpers/Environment/ElevationHelpers.cs
yaira2, line 1: Are we actively using this helper? I think it can be removed.

0x5bfa: On board

gumbarros: Agree, but just tomorrow at night bros

gumbarros: @yaira2 good night!

The `class is used in these places`

image

The Win32 method in just these places:
image

Can I remove the class and update the callers with the `if` or leave as is?

yaira2: I mixed it up with a different method, we can keep this as in then.

PR #15300, opened 2024-05-04
Code Quality: Use SidebarContext and HomePageContext in OpenIn... RichCommands

0x5bfa:
## PR Checklist

   Closes #issue
   _None_

## Screenshots

_None_

### DIFF
--- a/src/Files.App/Actions/ExecutableContextType.cs
+++ b/src/Files.App/Actions/ExecutableContextType.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) 2024 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+namespace Files.App.Actions
+{
+	public enum ExecutableContextType
+	{
+		None,
+
+		ContentPageContext,
+
+		HomePageContext,
+
+		SidebarContext,
+	}
+}
--- a/src/Files.App/Actions/ExecutableContextType.cs
+++ b/src/Files.App/Actions/ExecutableContextType.cs
@@ -3,14 +3,29 @@
 
 namespace Files.App.Actions
 {
+	/// <summary>
+	/// Defines constant that specifies the <see cref="IRichCommand"/> executable type.
+	/// </summary>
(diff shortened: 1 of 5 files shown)


0x5bfa:
@hishitetsu Does tho look good to you and do you happen to have another idea to concern?

0x5bfa:
Should can IsExecutable in the first line of ExecuteAsync?

Idk but it looks odd to me somewhat though it's my idea lol.

hishitetsu:
It would be better to do the same conditional branching in `ExecuteAsync` as in `GetIsExecutable`.
In the first place, is there a need to use SidebarContext or HomePageContext here?

0x5bfa:
I'm gonna use RochCommand for sidebar and widget card right click items.

0x5bfa:
> It would be better to do the same conditional branching in `ExecuteAsync` as in `GetIsExecutable`.
> 
> But in the first place, is there a need to use SidebarContext or HomePageContext here?

In that case, calling GetIsExecutable beforehand wouldn't work?

hishitetsu:
Actions should be stateless. In other words, they shouldn't have any fields which change the process depending on the value. Control of processing should always be by reference to external properties.

0x5bfa:
As you may know we have discussion about Actions with multi-state. Why the actions should be stateless? ContentPageContext is already state and if user is viewing widgets page, ContentPageContext values are null. This means we determine the state already.

In this case, calling GetIsExecutable and return the context type (also remove the field) at the first line of ExecuteAsync is the best idea. Besides, this ExecutableContextType is just getting executable context type literally determining whether values of each context are null or not.

0x5bfa:
You are telling that there’s no guarantee of GetIsExecutable being called at all right?

0x5bfa:
I understand the point, then calling GetIsExecutable in ExecuteAsync and returning executable context might be a solution to this.

0x5bfa:
This is ready to be merged.

PR #15324, opened 2024-05-07
Code Quality: Pin WinRT server object to prevent object reference being invalid

hez2010:
**Resolved / Related Issues**

The potential AV coming from `FileTagsDatabase` and `LayoutPreferencesDatabase` can be caused by CsWinRT who may not pin the server object, so that the object reference holding by the client can become invalid which can lead to access violation and crash the process.

This is a workaround to CsWinRT. Let's see if this resolves the `COMException` crashing issue.

/cc: @yair100 @hishitetsu


yaira2:
LGTM

PR #15339, opened 2024-05-08
Feature: Added support for SeerPro's "Track selected file" setting

### LINKED ISSUE #14966: Bug: Seer preview should close when selecting a different file via the mouse
### Description

Seer has a quick preview feature that I often use in my work. Its function is as follows: when any file/folder in the file manager is selected and the spacebar is pressed, it allows browsing the contents without directly opening it. This is particularly convenient for someone like me who works with numerous Word and Excel documents.

Files is my favorite file manager. However, when combining the two, there's a rather troublesome issue: once the quick preview feature is activated for the first time by pressing the spacebar, selecting any file thereafter or returning to the previous level folder will automatically invoke the quick preview feature, unless Seer is closed in the background to stop it. It's as if once the spacebar is pressed for the first time, every subsequent selection of any file/folder with the mouse automatically adds the quick preview feature corresponding to the spacebar. This is quite frustrating. Both applications are very useful, and I am reluctant to give up either.

Through testing, this issue does not occur with Seer and the default file manager of Windows. Currently, I am not sure whether the issue lies with Files or Seer, so I have submitted feedback for both applications, hoping to get help from the developers. I am very grateful for the excellent software you have developed and hope it can be further improved!

### Steps To Reproduce

1. When Seer and Files are running, but Seer has not yet been used for the first time (before the first press of the spacebar), clicking to select any file with the mouse does not present any issues.

[2024-03-15 17-22-13.mkv.zip](https://github.com/files-community/Files/files/14613585/2024-03-15.17-22-13.mkv.zip)

2. When Seer is running, pressing the spacebar on any file to use Seer's quick preview feature will lead to the automatic invocation of the quick preview feature upon any subsequent mouse selection of any file. It's as though every mouse click to select a file automatically presses the spacebar.

[2024-03-15 17-23-34.mkv.zip](https://github.com/files-community/Files/files/14613586/2024-03-15.17-23-34.mkv.zip)

### Requirements

 hope to prevent the automatic invocation of Seer's quick preview feature when using Files and Seer together, so that the feature is only used when I actively press the spacebar according to my needs.

### Files Version

3.2.7.0

### Windows Version

10.0.22631.3296

### Log File

[debug.log](https://github.com/files-community/Files/files/14613628/debug.log)

yaira2:
**Resolved / Related Issues**

- Closes #14966

1. Install SeerPro
2. Confirm that the follow selection setting in SeerPro is reflected in Files
   - This setting is only detected once, Files requires a restart after updating the setting in SeerPro

### DIFF
--- a/src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
+++ b/src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
@@ -72,7 +72,7 @@ namespace Files.App.Services.PreviewPopupProviders
 			var keyName = @"HKEY_CURRENT_USER\Software\Corey\Seer";
 			var value = Registry.GetValue(keyName, "tracking_file", null);
 
-			if (value is not null)
+			if (value?.ToString() == "true")
 				return Task.FromResult(true);
 
 			// List of possible paths for the Seer Pro settings file
--- a/src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
+++ b/src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
@@ -1,6 +1,7 @@
 // Copyright (c) 2024 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
+using Microsoft.Win32;
 using System.IO;
 using System.Runtime.InteropServices;
 using Vanara.PInvoke;
@@ -67,6 +68,12 @@ namespace Files.App.Services.PreviewPopupProviders
 		private Task<bool> DetectTrackSelectionSetting()
 		{
 			bool trackSelectedFile = true;
+
+			var keyName = @"\\HKEY_CURRENT_USER\\Software\\Corey\\Seer";
+			var value = Registry.GetValue(keyName, "General", null);
+
+			if (value is not null)
+				return Task.FromResult(true);
 
 			// List of possible paths for the Seer Pro settings file
 			string[] paths =
--- a/src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
+++ b/src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
@@ -1,7 +1,7 @@
 // Copyright (c) 2024 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using Files.App.Services;
+using System.IO;
 using System.Runtime.InteropServices;
 using Vanara.PInvoke;
 
@@ -34,6 +34,18 @@ namespace Files.App.Services.PreviewPopupProviders
 
 		public async Task SwitchPreviewAsync(string path)
 		{
+			// Close preview window is track selection setting is disabled
+			if (!isTrackSelectionSettingEnabled)
+			{
+				HWND Window = User32.FindWindow("SeerWindowClass", null);
+				if (User32.IsWindowVisible(Window))
+				{
+					// Hide window if it's already visible
+					return;
+				}
+			}
+
+			// Update the preview window if the path changed
 			if (CurrentPath is not null && path != CurrentPath)
 				await TogglePreviewPopupAsync(path);
 		}
@@ -42,6 +54,58 @@ namespace Files.App.Services.PreviewPopupProviders
 		{
 			var handle = User32.FindWindow("SeerWindowClass", null).DangerousGetHandle();
 			return handle != IntPtr.Zero && handle.ToInt64() != -1;
+		}
+
+		private bool? _isTrackSelectionSettingEnabledCache;
+		private bool isTrackSelectionSettingEnabled
+		{
+			get
+			{
+				if (_isTrackSelectionSettingEnabledCache is null)
+					_isTrackSelectionSettingEnabledCache = DetectTrackSelectionSetting().Result;
+
+				return _isTrackSelectionSettingEnabledCache.Value;
+			}
+		}
+
+		private Task<bool> DetectTrackSelectionSetting()
+		{
+			bool trackSelectedFile = true;
+
+			// List of possible paths for the Seer Pro settings file
+			string[] paths =
+			{
+				Environment.ExpandEnvironmentVariables("%USERPROFILE%\\Documents\\Seer\\uwp.ini"),
+				Environment.ExpandEnvironmentVariables("%USERPROFILE%\\appdata\\Local\\Corey\\Seer\\uwp.ini"),
+				Environment.ExpandEnvironmentVariables("%USERPROFILE%\\Documents\\Seer\\config.ini")
+			};
+
+			// Find the first existing path
+			foreach (var path in paths)
+			{
+				if (File.Exists(path))
+				{
+					// Read the settings file and look for the tracking_file setting
+					string[] lines = File.ReadAllLines(path);
+
+					foreach (var line in lines)
+					{
+						if (line.StartsWith("tracking_file", StringComparison.OrdinalIgnoreCase))
+						{
+							string[] keyValue = line.Split('=');
+							if (keyValue.Length == 2 && bool.TryParse(keyValue[1].Trim(), out bool isTrackingFile))
+							{
+								trackSelectedFile = isTrackingFile;
+								break;
+							}
+						}
+					}
+
+					break;
+				}
+			}
+
+			return Task.FromResult(trackSelectedFile);
 		}
 	}
 }
--- a/src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
+++ b/src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
@@ -69,7 +69,7 @@ namespace Files.App.Services.PreviewPopupProviders
 		{
 			bool trackSelectedFile = true;
 
-			var keyName = @"\HKEY_CURRENT_USER\Software\Corey\Seer";
+			var keyName = @"HKEY_CURRENT_USER\Software\Corey\Seer";
 			var value = Registry.GetValue(keyName, "tracking_file", null);
 
 			if (value is not null)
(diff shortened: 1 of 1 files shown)


yaira2:
Fyi @ccseer

src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
```diff
@@ -43,5 +51,58 @@ public async Task<bool> DetectAvailability()
...
+
+		private Task<bool> DetectTrackSelectionSetting()
+		{
+			bool trackSelectedFile = true;
+
+			// List of possible paths for the Seer Pro settings file
+			string[] paths =
+			{
+				Environment.ExpandEnvironmentVariables("%USERPROFILE%\\Documents\\Seer\\uwp.ini"),
+				Environment.ExpandEnvironmentVariables("%USERPROFILE%\\appdata\\Local\\Packages\\CNABA5E861-AC2A-4523-B3C1.Seer-AWindowsQuickLookTo_p7t0z30wh4868\\LocalCache\\Local\\Corey\\Seer\\uwp.ini"),
```
yaira2, line 75: @ccseer does this work for you?

ccseer: hi, why not include the `win32` path?

yaira2: That's included on the next line

ccseer:
cool. thx.

yaira2:
I tested the store version and it works correct, @ccseer can you confirm this works for the win32 version?

0x5bfa (commented):
Code wise, looks good but some parts.

src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
```diff
@@ -43,5 +51,58 @@ public async Task<bool> DetectAvailability()
 			var handle = User32.FindWindow("SeerWindowClass", null).DangerousGetHandle();
 			return handle != IntPtr.Zero && handle.ToInt64() != -1;
 		}
+
+		private bool? _isTrackSelectionSettingEnabledCache;
+		private bool isTrackSelectionSettingEnabled
```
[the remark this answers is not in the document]
yaira2: Private is fine because this is the only provider that has this option.

ccseer:
> I tested the store version and it works correct, @ccseer can you confirm this works for the win32 version?

do i need to build from the source to test?
I just read the code, the default path for win32 is from REG, but it's not in your code.
Am I missing something?

src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
```diff
@@ -43,5 +51,58 @@ public async Task<bool> DetectAvailability()
...
+		}
+
+		private Task<bool> DetectTrackSelectionSetting()
+		{
+			bool trackSelectedFile = true;
+
+			// List of possible paths for the Seer Pro settings file
+			string[] paths =
+			{
+				Environment.ExpandEnvironmentVariables("%USERPROFILE%\\Documents\\Seer\\uwp.ini"),
```
0x5bfa, line 74: It might help us with other providers if we create a constant class 'Constants.CloudDriveSettings' or something similar one.

0x5bfa (approved):
One thing I concern, what do you think?
Putting aside it, LGTM, code wise.

yaira2:
> I just read the code, the default path for win32 is from REG, but it's not in your code.

I didn't implement anything for the registry, I'll modify this PR to account for that possibility.

src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
yaira2: This is only for SeerPro, other providers don't have this option.

0x5bfa: Spreading hard coded strings doesn't look great however this issue is out of scope and so I'm fine for now.

yaira2: Okay

ccseer (commented):
image


hi, in the REG, you can just read `tracking_file` from `Computer\HKEY_CURRENT_USER\SOFTWARE\Corey\Seer` 
`General` group is only for ini files.

ccseer (commented):
Hi, I just got some code from Copilot, hope it helps.
```c#
using System;
using Microsoft.Win32;

class Program
{
    static void Main()
    {
        // Open the subkey
        using (RegistryKey key = Registry.CurrentUser.OpenSubKey(@"SOFTWARE\Corey\Seer"))
        {
            if (key != null)
            {
                // Attempt to read the value
                object value = key.GetValue("tracking_file");

                if (value != null)
                {
                    // Convert the value to a boolean
                    bool trackingFile = Convert.ToBoolean(value);

                    Console.WriteLine($"tracking_file value: {trackingFile}");
                }
                else
                {
                    Console.WriteLine("The 'tracking_file' value does not exist.");
                }
            }
            else
            {
                Console.WriteLine("The specified registry key does not exist.");
            }
        }
    }
}
```

ccseer:
1.	Double Backslashes in Key Name: In the keyName variable, you've used double backslashes (\\) at the beginning and within the path. When specifying registry paths in C#, you should use a single backslash (\) as the separator, and there's no need for leading backslashes. The correct format for the key name would be @"HKEY_CURRENT_USER\Software\Corey\Seer".
2.	Registry.GetValue Usage: The Registry.GetValue method is correctly used to read a value from the registry. However, the third parameter you've passed is null, which is the default value returned if the specified valueName does not exist. If you're trying to read a boolean value named General, ensure that General is the correct name of the value you're interested in. If you're trying to read a value named tracking_file (as mentioned in your previous question), you should replace "General" with "tracking_file".
3.	Return Type: The snippet ends with return Task.FromResult(true);, which suggests you want this code to be part of an asynchronous method returning a Task<bool>. However, the code snippet as provided does not include the method signature, so it's unclear how this fits into the larger context of your application.


```C#
using System;
using System.Threading.Tasks;
using Microsoft.Win32;

public class RegistryReader
{
    public Task<bool> ReadTrackingFileValueAsync()
    {
        var keyName = @"HKEY_CURRENT_USER\Software\Corey\Seer";
        var value = Registry.GetValue(keyName, "tracking_file", null);

        if (value is not null)
        {
            // Assuming the value is stored as an integer (0 for false, non-zero for true)
            bool trackSelectedFile = Convert.ToBoolean(value);
            return Task.FromResult(trackSelectedFile);
        }
        else
        {
            // Return false or handle the case where the value does not exist
            return Task.FromResult(false);
        }
    }
}

```

hishitetsu:
When I tried to test it, it crashed with the following exception:
```
2024-05-12 22:17:34.4955|Error|System.ArgumentException: Registry key name must start with a valid base key name. (Parameter 'keyName')
   at Microsoft.Win32.Registry.GetBaseKeyFromKeyName(String keyName, String& subKeyName)
   at Microsoft.Win32.Registry.GetValue(String keyName, String valueName, Object defaultValue)
   at Files.App.Services.PreviewPopupProviders.SeerProProvider.DetectTrackSelectionSetting()
   at Files.App.Services.PreviewPopupProviders.SeerProProvider.get_IsTrackSelectionSettingEnabled()
   at Files.App.Services.PreviewPopupProviders.SeerProProvider.SwitchPreviewAsync(String path)
   at Files.App.Actions.LaunchPreviewPopupAction.SwitchPopupPreviewAsync()
   at Files.App.Actions.LaunchPreviewPopupAction.Context_PropertyChanged(Object sender, PropertyChangedEventArgs e)
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at Microsoft.UI.Dispatching.DispatcherQueueSynchronizationContext.<>c__DisplayClass2_0.<Post>b__0()
```

yaira2:
@hishitetsu can you try with the latest commit?

src/Files.App/Services/PreviewPopupProviders/SeerProProvider.cs
```diff
@@ -43,5 +52,64 @@ public async Task<bool> DetectAvailability()
...
+
+				return _IsTrackSelectionSettingEnabledCache.Value;
+			}
+		}
+
+		private Task<bool> DetectTrackSelectionSetting()
+		{
+			bool trackSelectedFile = true;
+
+			var keyName = @"\HKEY_CURRENT_USER\Software\Corey\Seer";
```
hishitetsu, line 72:
```suggestion
			var keyName = @"HKEY_CURRENT_USER\Software\Corey\Seer";
```
Fixing it this way will prevent crashes, but it will always track the file regardless of the settings.

yaira2: Do you have a link to the Win32 version? I can't seem to locate it.

hishitetsu: Is this what you mean?
http://1218.io/

yaira2:
Everything looks good from my end

0x5bfa (approved):
LGTM, code wise.

hishitetsu:
It still tracks the file regardless of the settings in my environment.

ccseer:
the default value is true.
if Seer is running, and the Key is not found, you can take it as true.

yaira2:
> It still tracks the file regardless of the settings in my environment.

Did you restart Files in between changing the setting?

yaira2:
I tested with the Win32 version and confirm it's working as expected. The one thing to note is that a restart of Files is required after changing the setting in Seer.

PR #15357, opened 2024-05-10
Code Quality: Moving registry database to Files.App

hez2010:
**Resolved / Related Issues**

Fixes all `COMException` crashing issues caused by CsWinRT. Particularly, `LayoutPreferenceDatabase` and `FileTagsDatabase`.


hez2010:
> @hez2010 can you add steps we can use to test these changes?

No behavior changes as this is basically a copy-paste from Files.App.Server to Files.App

0x5bfa (approved):
Can't we have both to easily revert when the package's issue fixed instead of deleting classes?

Code changes are looking good.

hez2010:
> Can't we have both to easily revert when the package's issue fixed instead of deleting classes?

I made this change in about 20 minutes. It's almost a copy-paste and fixing namespace thing. 
We are using registry now, so we don't need to move it to the server project as registry is thread safe. Thus, I don't think we want to revert this PR in the future.

hishitetsu (approved):
I confirmed that it works fine. LGTM

PR #15362, opened 2024-05-10
Fix: Updated Files running as admin prompt for when UAC is off

Josh65-2201:
**Resolved / Related Issues**

- Closes https://discord.com/channels/725513575971684472/725514946112127057/1238497829811519600

1. Open Files as admin
2. Check text is updated and formatted correctly


src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3271,7 +3271,7 @@
     <value>Files is running as administrator</value>
   </data>
   <data name="FilesRunningAsAdminContent" xml:space="preserve">
-    <value>Due to platform limitations, drag and drop isn't available when running Files as administrator.</value>
+    <value>Due to an issue with Windows, drag and drop isn't available when running Files as admin. If you want to use drag and drop and have opened Files normally, Open UAC (User Account Control) from the start menu then move the slider to 3rd level from the bottom then restart Windows. Otherwise, you can keep using Files without drag & drop.</value>
```
yaira2, line 3274:
```suggestion
    <value>Due to an issue with Windows, drag and drop isn't available when running Files as admin. If you want to use drag and drop and have opened Files normally, Open UAC (User Account Control) from the start menu then move the slider to 3rd level from the bottom then restart Windows. Otherwise, you can keep using Files without drag &amp; drop.</value>
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3271,7 +3271,7 @@
     <value>Files is running as administrator</value>
   </data>
   <data name="FilesRunningAsAdminContent" xml:space="preserve">
-    <value>Due to platform limitations, drag and drop isn't available when running Files as administrator.</value>
+    <value>Due to a limiation with Windows and WinAppSdk, drag and drop isn't available when running Files as admin. If you wish to use drag and drop, you can work around this limitation by opening UAC (User Account Control) from the Start Menu and selecting the third level, and restarting Windows. Otherwise, you can keep using Files without drag &amp; drop.</value>
```
0x5bfa, line 3274: Most users don't know what WinAppSdk is.

SDK is better I believe.

0x5bfa: Or SDK is also part of Windows so even might not needed.

yaira2: We might want to add some line breaks as well, I'll have a better idea once I compile this PR on my device.

Josh65-2201: I agree that WinAppSDK part is not needed and would be less to read. I would also suggest that the instructions are formatted as to make it more clear
1.
2.
3.

0x5bfa: MarkdownTextBlock might help you

PR #15371, opened 2024-05-11
Fix: Fixed an issue where execution via URI or alias could run different version of Files

### LINKED ISSUE #13112: Feature: Add app execution aliases and URL schemes for each build channel

0x5bfa:
### Resolved / Related Issues

- Closes #13112

1. Open stable and preview (and dev as well)
2. Enable open in background
3. Click task tray icon
4. Check if correct version of Files opens

### DIFF
--- a/.github/workflows/deploy-preview-legacy.yml
+++ b/.github/workflows/deploy-preview-legacy.yml
@@ -48,6 +48,12 @@ jobs:
         $xmlDoc.Package.Identity.Publisher="$env:SIDELOAD_PUBLISHER_SECRET"
         $xmlDoc.Package.Properties.DisplayName="Files - Preview"
         $xmlDoc.Package.Applications.Application.VisualElements.DisplayName="Files - Preview"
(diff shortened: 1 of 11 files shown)


gave92:
Iirc "Files.exe" might also be encoded in the .reg file to set Files as default.

PR #15419, opened 2024-05-16
Feature: Updated the Status Bar design

yaira2:
**Resolved / Related Issues**

- Updated the Status Bar design 
- This is prep work for completely floating panes

1. Confirmed the corner radius is set to `0` when the sidebar is collapsed
2. Confirmed the sidebar resizer height matches the border height
3. Confirmed the border thickness is correct when Dual Pane mode is enabled
4. Confirmed the border thickness is correct when the Info Pane is opened

**Screenshots**
image

### DIFF
--- a/src/Files.App/Views/MainPage.xaml
+++ b/src/Files.App/Views/MainPage.xaml
@@ -215,78 +215,81 @@
 			SelectedItem="{x:Bind SidebarAdaptiveViewModel.SidebarSelectedItem, Mode=TwoWay}"
 			ViewModel="{x:Bind SidebarAdaptiveViewModel}">
 
-			<Grid
-				x:Name="RootGrid"
-				PreviewKeyDown="RootGrid_PreviewKeyDown"
-				SizeChanged="RootGrid_SizeChanged">
-				<Grid.RowDefinitions>
-					<RowDefinition Height="Auto" />
-					<RowDefinition
-						x:Name="ContentRow"
-						Height="*"
-						MinHeight="100" />
-					<RowDefinition Height="Auto" />
-					<RowDefinition x:Name="PaneRow" Height="Auto" />
-					<RowDefinition Height="32" />
-				</Grid.RowDefinitions>
-				<Grid.ColumnDefinitions>
-					<ColumnDefinition
-						x:Name="ContentColumn"
-						Width="*"
-						MinWidth="100" />
-					<ColumnDefinition Width="Auto" />
-					<ColumnDefinition x:Name="PaneColumn" Width="Auto" />
-				</Grid.ColumnDefinitions>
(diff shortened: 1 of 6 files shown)


src/Files.App/Views/PaneHolderPage.xaml
```diff
@@ -126,7 +126,7 @@
 				<icore:ChangePropertyAction
 					PropertyName="CurrentInstanceBorderThickness"
 					TargetObject="{Binding ElementName=PaneLeft}"
-					Value="0,0,0,1" />
+					Value="0,0,0,0" />
```
marcelwgn, line 129: Is this related and if so, how does this affect the Statusbar?

src/Files.App/Views/MainPage.xaml
```diff
@@ -215,78 +215,81 @@
...
+				</Grid>
+			</sidebar:SidebarView.InnerContent>
 
+			<!--  Footer Content  -->
+			<sidebar:SidebarView.FooterContent>
+				<Border>
+					<!--  Status Bar  -->
+					<uc:StatusBar
+						x:Name="StatusBar"
+						Height="32"
```
marcelwgn, line 288: Could this be a problem with larger text sizes and text getting clipped? (Not like this wouldn't have been an issue without this PR)

marcelwgn (approved):
LGTM. I'm a bit concerned that if all items are expanded, there is no visual separator between the footer content and the sidebar items, but I think that is quite hard to solve. Overall, I think it looks quite well though.

Besides that, I think its good. I'm not too happy with the name FooterContent (since Footer is also something one could expect in the left pane) but I also can't think of anything more fitting.

0x5bfa:
Yair is going to add Settings button as a pane footer.

image

As I agree with @marcelwgn on that, it should be renamed to 'ContentFooter' and pane one to be 'PaneFooter'.

src/Files.App/Views/PaneHolderPage.xaml
yaira2: It’s an unrelated fix, I can try to send the before and after…

src/Files.App/Views/MainPage.xaml
yaira2: I haven’t seen any reports of this nature, but like you said, it wouldn’t be a new issue if that’s the case…

src/Files.App/Views/PaneHolderPage.xaml
marcelwgn: Just wanted to make sure, I understand this change correctly and that I didnt miss anything.

yaira2: Actually, this change is related.

The border is handled on a different control now so we no longer apply it here.

PR #15441, opened 2024-05-21
Fix: Correct load os language for `MessageFormatExtensions`

### LINKED ISSUE #15440: Bug: Not correct Default system language load for `MessageFormatExtensions`
### Description

The system language is incorrectly defined, a problem occurs when Formater takes the default 'en' for parsing

### Screenshot problem
1. If I don't choose the exact language
image
2. Language not set to system language
image
3. This causes the plural localization to function incorrectly
image

### Steps To Reproduce

1. If not specified default language load OS default correctly

XTorLukas:
**Resolved / Related Issues**

- Closes #15440

1. The default language for `MessageFormatExtensions` is loaded as the default OS language

### DIFF
--- a/src/Files.App/Extensions/MessageFormatExtensions.cs
+++ b/src/Files.App/Extensions/MessageFormatExtensions.cs
@@ -15,7 +15,7 @@ namespace Files.App.Extensions
 		private static readonly ResourceMap resourcesTree = new ResourceManager().MainResourceMap.TryGetSubtree("Resources");
 
 		// CultureInfo based on the application's primary language override
-		private static readonly CultureInfo locale = new(ApplicationLanguages.PrimaryLanguageOverride);
+		private static readonly CultureInfo locale = ApplicationLanguages.PrimaryLanguageOverride == string.Empty ? CultureInfo.InstalledUICulture : new(ApplicationLanguages.PrimaryLanguageOverride);
 
 		// Message formatter with caching enabled, using the current UI culture's two-letter ISO language name
 		private static readonly MessageFormatter formatter = new(useCache: true, locale: locale.TwoLetterISOLanguageName);
--- a/src/Files.App/Extensions/MessageFormatExtensions.cs
+++ b/src/Files.App/Extensions/MessageFormatExtensions.cs
@@ -15,7 +15,10 @@ namespace Files.App.Extensions
 		private static readonly ResourceMap resourcesTree = new ResourceManager().MainResourceMap.TryGetSubtree("Resources");
 
 		// CultureInfo based on the application's primary language override
-		private static readonly CultureInfo locale = ApplicationLanguages.PrimaryLanguageOverride == string.Empty ? CultureInfo.InstalledUICulture : new(ApplicationLanguages.PrimaryLanguageOverride);
+		private static readonly CultureInfo locale =
+			ApplicationLanguages.PrimaryLanguageOverride == string.Empty
+				? CultureInfo.InstalledUICulture
+				: new(ApplicationLanguages.PrimaryLanguageOverride);
 
 		// Message formatter with caching enabled, using the current UI culture's two-letter ISO language name
 		private static readonly MessageFormatter formatter = new(useCache: true, locale: locale.TwoLetterISOLanguageName);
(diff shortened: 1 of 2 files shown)


XTorLukas:
This is a quick solution, but it would be good to do this when starting the application, because when I delete `Local > Packages`, the default language is not set to os, but here you would need to test if the language can be used within the application, or set the default 'en'

src/Files.App/ViewModels/Settings/GeneralViewModel.cs
```diff
@@ -152,6 +152,11 @@ private void AddSupportedAppLanguages()
 				.ThenBy(language => language.LanguageName);
 			AppLanguages = new ObservableCollection<AppLanguageItem>(appLanguages);
 
+			// Add default language
+			var osDefaultLanguage = new AppLanguageItem(CultureInfo.InstalledUICulture.Name);
+			if (AppLanguages.Select(language => language.LanguageName.Contains(osDefaultLanguage.LanguageName)).Any())
+				AppLanguages[0].LanguagID = osDefaultLanguage.LanguagID;
```
yaira2, line 158: I don't think we need to make any changes here, it should be possible to handle everything from the format extension class.

src/Files.App/Extensions/MessageFormatExtensions.cs
```diff
@@ -15,7 +15,7 @@ public static class MessageFormatExtensions
 		private static readonly ResourceMap resourcesTree = new ResourceManager().MainResourceMap.TryGetSubtree("Resources");
 
 		// CultureInfo based on the application's primary language override
-		private static readonly CultureInfo locale = new(ApplicationLanguages.PrimaryLanguageOverride);
+		private static readonly CultureInfo locale = ApplicationLanguages.PrimaryLanguageOverride == string.Empty ? CultureInfo.InstalledUICulture : new(ApplicationLanguages.PrimaryLanguageOverride);
```
0x5bfa, line 18:
```suggestion
		private static readonly CultureInfo locale =
			ApplicationLanguages.PrimaryLanguageOverride == string.Empty
				? CultureInfo.InstalledUICulture
				: new(ApplicationLanguages.PrimaryLanguageOverride);
```

src/Files.App/ViewModels/Settings/GeneralViewModel.cs
```diff
@@ -152,6 +152,11 @@ private void AddSupportedAppLanguages()
 				.ThenBy(language => language.LanguageName);
 			AppLanguages = new ObservableCollection<AppLanguageItem>(appLanguages);
 
+			// Add default language
+			var osDefaultLanguage = new AppLanguageItem(CultureInfo.InstalledUICulture.Name);
+			if (AppLanguages.Select(language => language.LanguageName.Contains(osDefaultLanguage.LanguageName)).Any())
```
0x5bfa, line 157:
```suggestion
			if (AppLanguages.Select(language =>
				language.LanguageName.Contains(osDefaultLanguage.LanguageName)).Any())
```

0x5bfa (approved):
Codewise

src/Files.App/Extensions/MessageFormatExtensions.cs
yaira2, line 18: If `ApplicationLanguages.PrimaryLanguageOverride` is empty, we can check if the system language is in the list of supported languages. Otherwise, we can fall back to en--us.

I don't think we need to modify `GeneralViewModel.cs`.

XTorLukas:
although I hadn't planned to finish it today I have an auxiliary class ready, but then this fix will be unnecessary

src/Files.App/Extensions/MessageFormatExtensions.cs
0x5bfa: Agree

yaira2: @XTorLukas what are your thoughts? In theory, this should be the cleanest solution.

XTorLukas: It looks good, it's just very primitive, but a future helper will add better control over application languages and also the ability to see if I'm compatible with the system language, otherwise use en-US.

XTorLukas: The reason for the change to `GeneralViewModel.cs` was to store the language name, thus avoiding string.Empty, but this is also addressed in a future helper class

src/Files.App/ViewModels/Settings/GeneralViewModel.cs
XTorLukas: The reason for the change to GeneralViewModel.cs was to store the language name, thus avoiding string.Empty, but this is also addressed in a future helper class

PR #10890, opened 2022-12-30
Feature: Added detection for Nutstore cloud

hecksmosis:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #9832 

**Validation**
How did you test these changes?
- [x] Built and ran the app

**Screenshots (optional)**
Add screenshots here.

PR #15445, opened 2024-05-22
Fix: Fixed an issue with Nutstore cloud drive detection

wangfu91:
Hi, I'm a developer at the company [Nutstore](https://www.jianguoyun.com/), I've found out that the cloud drive detection logic for Nutstore in PR #10890 is incorrect. 

Searching all the sub-directories named "Nutstore" in `%userprofile%` can be extremely time-consuming and will likely cause the Files app the stop responding, it may also result in multiple incorrect Nutstore cloud drive entries. 

Querying the registry entries in `SyncRootManager` is a more efficient way to detect the root dir of the Nutstore cloud drive.


yaira2:
Fyi @hecksmosis

yaira2 (approved):

yaira2:
@wangfu91 thank you for the fix! While you're here, are there any other integrations that we can do with Nutestore to improve the experience (eg. context menu options etc)?

wangfu91:
Thanks for merging my PR @yaira2, kudos for building such a beautiful app.

> While you're here, are there any other integrations that we can do with Nutestore to improve the experience (eg. context menu options etc)?

Files have already properly integrated the context menu, I haven't found anything else to improve at the moment, I will raise an issue or create a PR if I do.

PR #15448, opened 2024-05-22
Fix: Fixed formatting bug when entering commands into the Address Bar

### LINKED ISSUE #15447: Bug: Address bar formatting causes issues when running commands through address bar
### Description

Commands run through the address bar are being formatted, which breaks some commands. For example, `//` and `/` are converted to `\` in the address bar. `https://github.com/files-community/Files` becomes `https:\github.com\files-community\Files`. This behavior is inconsistent with how windows explorer or run dialog handles commands.

This can cause issues with many commands that take links/paths/strings as parameters and expect a certain format, like `git`, `curl`. For example, running `git clone https://github.com/files-community/Files` does not work in Files but works in Windows explorer. `curl https://httpbin.org/get` does not work in Files for the same reason.

### Steps To Reproduce

1. run `ping https://github.com/files-community/Files` from the address bar. (we are only interested in the output, not if it pings)
2. Actual output: 
     ```Ping request could not find host https:\github.com\files-community\Files. Please check the name and try again.```

### Requirements

Expected output: 
```Ping request could not find host https://github.com/files-community/Files. Please check the name and try again.```

This is the behavior in Windows Explorer. More specifically, `https://github.com/files-community/Files`  remains unchanged when passed as a parameter to the command.

### Files Version

3.4.1.0

### Windows Version

10.0.22631.3593

### Log File

(not necessary)

### PR DESCRIPTION
**Pull request by yaira2**
**Resolved / Related Issues**

- Closes #15447

1. Enter `git clone https://github.com/files-community/Files` into address bar
2. Confirm the repo is cloned correctly

### DIFF
--- a/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
+++ b/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
@@ -688,6 +688,10 @@ namespace Files.App.ViewModels.UserControls
 
 		private static string NormalizePathInput(string currentInput, bool isFtp)
 		{
+			// Check if input is a command and not a path
+			if (currentInput.Length > 1 && currentInput.ElementAt(1) != ':')
+				return currentInput;
+
 			if (currentInput.Contains('/') && !isFtp)
 				currentInput = currentInput.Replace("/", "\\", StringComparison.Ordinal);
 
--- a/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
+++ b/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
@@ -688,10 +688,6 @@ namespace Files.App.ViewModels.UserControls
 
 		private static string NormalizePathInput(string currentInput, bool isFtp)
 		{
-			// Check if input is a command and not a path
-			if (currentInput.Length > 1 && currentInput.ElementAt(1) != ':')
-				return currentInput;
-
 			if (currentInput.Contains('/') && !isFtp)
 				currentInput = currentInput.Replace("/", "\\", StringComparison.Ordinal);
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs [hunk 1]
```diff
@@ -688,6 +688,10 @@ public async Task SetPathBoxDropDownFlyoutAsync(MenuFlyout flyout, PathBoxItem p
 
 		private static string NormalizePathInput(string currentInput, bool isFtp)
 		{
+			// Check if input is a command and not a path
+			if (currentInput.Length > 1 && currentInput.ElementAt(1) != ':')
```
yaira2, line 692: Is `:` always the second character in a valid path?

FieryRMS: Couldn't the paths be relative? `./Child` `../Sibling` etc? It won't have a drive letter then

FieryRMS:
I think it will be a big hassle to individually blacklist/whitelist each case. My approach would be to pass the original `currentInput` to `LaunchApplicationFromPath`  instead of the normalized one.

https://github.com/files-community/Files/blob/4b77a5a1c74ffaa5c2bd129fa97bc79009b23590/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs#L782

Then after finding the `filename`, before `LaunchAppAsync` (line  827) we can `NormalizePathInput` if we require.

https://github.com/files-community/Files/blob/4b77a5a1c74ffaa5c2bd129fa97bc79009b23590/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs#L821-L828

FieryRMS:
https://github.com/files-community/Files/blob/4b77a5a1c74ffaa5c2bd129fa97bc79009b23590/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs#L821-L828

After looking a bit more into this part of the code and doing some testing, it seems like if there is spaces in the path of the application (and there is parameters to the program), the program fails to run 
eg: `C:\Users\FieryRMS\Desktop\Folder 2\test 1.exe test`

I propose to additionally check if `currentInput` starts with `"`/`'`, then it should get the `fileName` until the next `"`/`'` (or end if it doesn't exist).

src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
yaira2: I'm not concerned since we don't have plans on supporting that, but worst case we can come up with a new fix if things change.

yaira2:
> I think it will be a big hassle to individually blacklist/whitelist each case. My approach would be to pass the original currentInput to LaunchApplicationFromPath instead of the normalized one.

That's a good approach. Should I disregard https://github.com/files-community/Files/pull/15448#issuecomment-2125469824?

FieryRMS (approved):
lgtm. built and tested it

FieryRMS:
> Should I disregard https://github.com/files-community/Files/pull/15448#issuecomment-2125469824?

That bug I mentioned should persist even after the current fix. This issue comes from 
`fileName = trimmedInput.Substring(0, positionOfBlank);`. If file path is `C:/Folder 1/test 1.exe param` then `fileName="C:/Folder"` instead of `"C:/Folder 1/test 1.exe"`. That's why I proposed that we can check for quotes to calculate the delimiting position, so if the user inputs `"C:/Folder 1/test 1.exe" param` we are able to correctly execute it.

I suppose I should make a separate issue for this. It seems unrelated to the issue at hand.

yaira2:
Thanks!

hishitetsu (approved):

Issue #9348, opened 2022-06-05
Feature: Add support for hiding the Toolbar


knuxyl:
### What's the Problem?

[ FEATURE # 1 ]
I would like an option to hide the tool bar that has the New button. I will never use this because all of those features are available in either the context menu or hotkeys so the option to hide this would make more sense and give more screen space.

### Solution/Idea

[ Feature # 1 ]
Settings toggle to hide the directory toolbar that contains basic context menu entries such as New, Cut, and Paste

### Alternatives

Continue using Windows explorer

### Priorities

| MUST |
All of these features are required for me to efficiently use a file manager.
Screen space is extremely important.

### Files Version

2.3.0.0

### Windows Version

Windows 11 Pro 21H2

yaichenbaum:
Merging with #9370

yaira2:
It's been a while since this was requested, but I'd like to resume this discussion and see if we can incorporate this moving foward.

0x5bfa:
I think Inner Toolbar should not be automatically hidden in Home page because Copy and even Eject, Format RichCommands if we make in the future should be shown.
What do you think?

yaira2:
I don't think this should be the default, agreed.

knuxyl:
> @knuxyl I'm excited to share that support for hiding the toolbar will be included in the v3.5 release.

That is great to hear, thank you!

PR #15451, opened 2024-05-23
Code Quality: Moved toolbar background to MainPage

yaira2:
**Resolved / Related Issues**

- In case we decide to support #9348

### DIFF
--- a/src/Files.App/Views/MainPage.xaml
+++ b/src/Files.App/Views/MainPage.xaml
@@ -217,7 +217,13 @@
 
 			<!--  Content Header  -->
 			<sidebar:SidebarView.ContentHeader>
-				<Border>
+				<Border
+					Margin="0,0,0,4"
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/MainPage.xaml
```diff
@@ -217,7 +217,13 @@
 
 			<!--  Content Header  -->
 			<sidebar:SidebarView.ContentHeader>
-				<Border>
+				<Border
```
0x5bfa, line 220: Changes look good but I think wrapping with Border increases amount of xaml tree.

0x5bfa (commented):
Let’s format XAML

src/Files.App/Views/MainPage.xaml
yaira2: The border was already there before (it seems to be a requirement).

0x5bfa (approved):

PR #15458, opened 2024-05-23
Feature: Moved Settings button to Sidebar

0x5bfa:
### Summary

TBA in the next release.

- Merged ISidebarItemModel into INavigationControlItem
- Change type of `IconSource` from `IconElement` to `FrameworkElement`
- Added an user setting `ShowSettingsButtonOnAddressToolbar`
- Added settings button on the Sidebar pane footer
- Added `AnimatedIcon.State` to certain VisualStates
- Added `PaneFooterStates` state to hide/show pane footer
- Removed `SideBarResources`

1. Open Files app

### Screenshots

image


src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
```diff
@@ -271,6 +272,23 @@ private Task<LocationItem> CreateItemHomeAsync()
 			return CreateSectionAsync(SectionType.Home);
 		}
 
+		private void CreateItemSettingsAsync()
+		{
+			var item = new LocationItem()
+			{
+				Text = "Settings".GetLocalizedResource(),
+				Path = "Settings",
+				Icon = new BitmapImage(new Uri(Constants.FluentIconsPaths.HomeIcon)),
```
yaira2, line 281: Is there a way to use an animated icon?

yaira2:
> I tested on my local but interaction test seems to be failing. @yaira2 how did you fixed the last time you encountered?

It seems to be related to this issue https://github.com/microsoft/microsoft-ui-xaml/issues/1971

yaira2:
> I tested on my local but interaction test seems to be failing. @yaira2 how did you fixed the last time you encountered?

It seems to be related to this issue https://github.com/microsoft/microsoft-ui-xaml/issues/1971.

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
0x5bfa: Yeah.
I’m gonna add a FrameworkElement instead of a IconElement to render PathIcons and OpacityIcons. Is there a new type of icons?

yaira2: AnimatedIcon is what we use for the existing settings button.

0x5bfa: Ah true, sure.

yaira2:
@0x5bfa the changes in [c5b0cfd](https://github.com/files-community/Files/pull/15458/commits/c5b0cfdbd6441dd25c5c050bf76adb5ddadacf83) are still needed for the tests.

src/Files.App/UserControls/SideBar/SideBarControls.xaml
```diff
@@ -42,8 +39,12 @@
 	<DataTemplate x:Key="DefaultSidebarItemTemplate">
 		<local:SidebarItem
 			AutomationProperties.AutomationId="{Binding Text, Mode=OneWay}"
+			Decorator="{Binding ItemDecorator, Mode=OneWay}"
+			Icon="{Binding IconSource, Mode=OneWay}"
 			IsExpanded="{Binding IsExpanded, Mode=TwoWay}"
-			Item="{Binding}" />
+			Item="{Binding}"
+			Text="{Binding Text, Mode=OneWay}"
+			TooltipText="{Binding ToolTip, Mode=TwoWay}" />
```
0x5bfa, line 47: There was an issue with display of Text and Tooltip.
Using `RelativeSource` instead of implicit one.

src/Files.App/Views/MainPage.xaml
```diff
@@ -215,6 +215,25 @@
 			SelectedItem="{x:Bind SidebarAdaptiveViewModel.SidebarSelectedItem, Mode=TwoWay}"
 			ViewModel="{x:Bind SidebarAdaptiveViewModel}">
 
+			<sidebar:SidebarView.PaneFooter>
+				<sidebar:SidebarItem
```
yaira2, line 219: Can you take the AutomationID from the toolbar button and move it here instead?

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
```diff
@@ -271,6 +273,19 @@ private Task<LocationItem> CreateItemHomeAsync()
 			return CreateSectionAsync(SectionType.Home);
 		}
 
+		private void CreateItemSettingsAsync()
```
yaira2, line 276: Is this still needed?

src/Files.App/Services/Settings/GeneralSettingsService.cs
```diff
@@ -263,6 +263,12 @@ public FileNameConflictResolveOptionType ConflictsResolveOption
 			set => Set(value);
 		}
 
+		public bool ShowSettingsButtonOnAddressToolbar
```
yaira2, line 266: I had an idea to invert the setting to "Display settings button on the sidebar" and to set the value to true.

src/Files.App/UserControls/SideBar/SideBarControls.xaml
yaira2: What was the issue?

src/Files.App/Services/Settings/GeneralSettingsService.cs
0x5bfa: Sure

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
0x5bfa: Yes as long as I finish refactoring SidebarView it uses core item "INavigationControlItem" in SidebarItem.Item property though this is odd.

src/Files.App/Views/MainPage.xaml
0x5bfa: AddressToolbar one named "SettingsButton" tho.
Is that ok?

src/Files.App/UserControls/SideBar/SideBarControls.xaml
0x5bfa: Didn't update on changed because those dont implement that logic but dependency prop does.

src/Files.App/Views/MainPage.xaml
yaira2: That's fine

src/Files.App/UserControls/AddressToolbar.xaml
```diff
@@ -532,7 +532,7 @@
 					x:Name="SettingsButton"
 					AccessKey="I"
 					AccessKeyInvoked="Button_AccessKeyInvoked"
-					AutomationProperties.AutomationId="SettingsButton"
+					AutomationProperties.AutomationId="Settings"
```
yaira2, line 535: This should still be "SettingsButton".

src/Files.App/Services/Settings/GeneralSettingsService.cs
```diff
@@ -263,6 +263,12 @@ public FileNameConflictResolveOptionType ConflictsResolveOption
 			set => Set(value);
 		}
 
+		public bool DisplaySettingsButtonOnSidebar
+		{
+			get => Get(false);
```
yaira2, line 268:
```suggestion
			get => Get(true);
```
The UI for this can be displayed under experimental.

src/Files.App/Views/MainPage.xaml
0x5bfa: Sure!

src/Files.App/UserControls/SideBar/SideBarItem.cs
```diff
@@ -20,13 +20,14 @@ public sealed partial class SidebarItem : Control
...
 		public bool CollapseEnabled => DisplayMode != SidebarDisplayMode.Compact;
+		public bool IsFooterItem => Item?.Section == SectionType.Footer;
 
 		private bool hasChildSelection => selectedChildItem != null;
 		private bool isPointerOver = false;
 		private bool isClicking = false;
 		private object? selectedChildItem = null;
 		private ItemsRepeater? childrenRepeater;
-		private ISidebarItemModel? lastSubscriber;
+		private INavigationControlItem? lastSubscriber;
```
marcelwgn, line 30: This seems like quite a strong mixing of domains. INavigationControlItem contains a lot of stuff that the SideBarItem will never need (e.g. SectionType) and I would be quite concerned if we use SideBar for settings and the entries there having such properties. 

Maybe I'm missing something here but ISidebarItemModel was there to decouple it from the other domains since it was designed to be reusable and not contain much "business logic" of file system related stuff.

0x5bfa: INavigationControlItem previously inherited from ISidebarItemModel.
And ISidebarViewModel was never used else where.
This means separation is useless.

0x5bfa:
> I'm quite sure that the test errors stem from the fact that the settings item is unable to resolve its position inside the list of items.
> 
> When implementing UIA patterns for lists (like the SideBar is), the elements should/need to determine where they sit in that list (SizeOfSet and PositionInSet). The settings item however is an outlier for this and the logic we currently use does not work for the settings item resulting in it not knowing its position in set (method throws an exception),

As Windows Terminal does, I tried to implement IList<object> PaneFooterItems in SidebarView, but there was casting error and null ref at runtime as a result it caused xaml unhandled exception.
What’s the best way to overcome this issue?

0x5bfa:
Fixed

src/Files.App/Views/MainPage.xaml
```diff
@@ -3,6 +3,7 @@
 	x:Class="Files.App.Views.MainPage"
 	xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
 	xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+	xmlns:animatedvisuals="using:Microsoft.UI.Xaml.Controls.AnimatedVisuals"
```
yaira2, line 6: Can this be removed?

src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
```diff
@@ -225,6 +232,16 @@ private async void UpdateService_OnPropertyChanged(object? sender, PropertyChang
 				await CheckForReleaseNotesAsync();
 		}
 
+		private async void WindowContext_OnPropertyChanged(object? sender, PropertyChangedEventArgs e)
```
yaira2, line 235: Do we also need to watch for changes to settings?

src/Files.App/UserControls/SideBar/SidebarItemAutomationPeer.cs
```diff
@@ -105,14 +105,23 @@ protected override int GetPositionInSetCore()
...
 			{
-				return items;
+				if (Owner.Owner.PaneFooterItemSource is IList<INavigationControlItem> footerItems)
+				{
+					return menuItems.Concat(footerItems).ToList();
+				}
+				else
+				{
+					return menuItems.ToList();
+				}
```
yaira2, line 122:
```suggestion
					return menuItems.Concat(footerItems).ToList();
				else
					return menuItems.ToList();
```

src/Files.App/UserControls/SideBar/SidebarItemAutomationPeer.cs
yaira2, line 1: @marcelwgn are you okay with these changes?

src/Files.App/UserControls/SideBar/SideBarView.xaml.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Input;
 using Microsoft.UI.Xaml.Markup;
+using System.Diagnostics.Eventing.Reader;
```
yaira2, line 9: Can this be removed?

src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
0x5bfa: It is not in Settings dialog and it will never be changed runtime.
We don’t have a way to watch file change from outside yet but once done we should watch.

src/Files.App/Services/Settings/IGeneralSettingsService.cs
```diff
@@ -209,5 +209,10 @@ public interface IGeneralSettingsService : IBaseSettingsService, INotifyProperty
 		/// A dictionary to determine which hashes should be shown.
 		/// </summary>
 		Dictionary<string, bool> ShowHashesDictionary { get; set; }
+
+		/// <summary>
+		/// Gets or sets a value that indicates whether the Settings button is visible in sidebar. DisplaySettingsButtonOnSidebar
```
yaira2, line 214:
```suggestion
		/// Gets or sets a value that indicates whether the Settings button should be visible in the sidebar.
```

src/Files.App/Services/Settings/GeneralSettingsService.cs
```diff
@@ -263,6 +263,12 @@ public FileNameConflictResolveOptionType ConflictsResolveOption
 			set => Set(value);
 		}
 
+		public bool DisplaySettingsButtonOnSidebar
```
yaira2, line 266: Can you add to `RaiseOnSettingChangedEvent`?

0x5bfa: https://github.com/files-community/Files/pull/15458#discussion_r1617830808

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
```diff
@@ -271,6 +272,22 @@ private Task<LocationItem> CreateItemHomeAsync()
...
+		{
+			var settingsItem = new LocationItem()
```
yaira2, line 277: Can you set the tooltip to include the keyboard shortcut?

yaira2:
I think we need to fix the icon alignment on the footer item
image

src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
yaira2: In that case it's fine, I thought we were adding to Settings.

src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
```diff
@@ -225,6 +232,16 @@ private async void UpdateService_OnPropertyChanged(object? sender, PropertyChang
...
+		{
+			switch (e.PropertyName)
+			{
+				case nameof(IWindowContext.IsCompactOverlay):
```
yaira2, line 239: We need a similar check for the sidebar item.

0x5bfa:
Footer uses the same xaml. I think icon element has 40 width and centered, and leader has 8 padding from the left, in a result it keeping a correct alignment seemingly. But we should set column width to Auto.

src/Files.App/UserControls/SideBar/SideBarView.xaml
```diff
@@ -84,13 +95,27 @@
...
+				x:Name="PaneFooterItemsRepeater"
+				Grid.Row="2"
+				Margin="0,4"
+				AutomationProperties.AccessibilityView="Content"
+				AutomationProperties.Name="SidebarPaneFooterItems"
+				ElementPrepared="MenuItemsHost_ElementPrepared"
+				IsTabStop="True"
+				ItemTemplate="{StaticResource DefaultSidebarItemTemplate}"
+				ItemsSource="{x:Bind PaneFooterItemSource, Mode=OneWay}"
+				VerticalCacheLength="5"
```
yaira2, line 116: I don't think cache length makes a difference here since there is only one item.

src/Files.App/UserControls/SideBar/SideBarView.xaml.cs
0x5bfa: Done

src/Files.App/Views/MainPage.xaml
0x5bfa: Done

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
0x5bfa: Done

src/Files.App/UserControls/SideBar/SideBarView.xaml
0x5bfa: Removed

0x5bfa:
- [x] icon alignment DONE
- [ ] hide the button in compact overlay FAILED
- [x] tooltip should display keyboard shortcut DONE

src/Files.App/UserControls/SideBar/SideBarItem.cs
marcelwgn: So do we revert this once we use Sidebar in other areas? I would argue, only because right now it is not used otherwise doesn't mean that mixing stuff is fine. After all SideBar could have been built as part of the Mainwindow but was not (similar things for the SidebarViewModel).

src/Files.App/UserControls/SideBar/SidebarItemAutomationPeer.cs
marcelwgn: Seems like a good approach to me.

src/Files.App/UserControls/SideBar/SideBarItem.cs
0x5bfa: Yeah this is not needed imo. And SidebarItem.Item reference too.
Generalizing would solve this issue.

yaira2:
>  hide the button in compact overlay FAILED

I think we can remove this requirement (especially if we end up switching Settings to a tab).

src/Files.App/UserControls/SideBar/SideBarControls.xaml
```diff
@@ -106,20 +107,25 @@
 								x:Name="IconPresenter"
 								Width="16"
 								Height="16"
+								Margin="11,0,0,0"
```
yaira2, line 110: Is an odd margin required?

0x5bfa: I'll revert because this is out of scope.
11 is a mixed up margin of 3px of selection indicator and 8px spacing.

0x5bfa: I should've use ColumnDefinition in the first place tho

0x5bfa: Reverted

0x5bfa:
@yaira2 I’d like to close and re-implement after refactoring because this is most likely to be altered again when refactoring.

PR #15459, opened 2024-05-23
Fix: Fixed an issue with launching programs from the Address Bar

### LINKED ISSUE #15455: Bug: Address bar does not launch programs when path contains spaces and has parameters
### Description

The Files app fails to find and launch programs if the path contains spaces and has parameters. Eg: running 
`D:\Folder With Spaces\program.exe param` from the address bar, Files tries to "execute" `D:\Folder` with params ` With Spaces\program.exe param`. The program still fails to run even if we explicitly specify the filename using quotes like, 
`"D:\Folder With Spaces\program.exe" param` .

### Steps To Reproduce

1. create a batch file named `test script.bat` with the follow script:
    ```cmd
    @echo off
    echo %*
    ```
2. execute the script with some parameters through the address bar, e.g.: `test script.bat param`. The script fails to run.
3. execute the script with quotes around the file name, e.g.: `"test script.bat" param`. The script fails to run.

### Requirements

Expected outcome was that it should've been able to run when filename was explicitly specified with quotes. (`"test script.bat" param`)

This bug is caused by this part of the code,
https://github.com/files-community/Files/blob/4b77a5a1c74ffaa5c2bd129fa97bc79009b23590/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs#L821-L828

I propose to check if the user provided `"` to avoid ambiguity. If the first character is a `"` we look for the next `"` and use that as the delimiter instead. If we cannot find the second `"` we use the whole input as `fileName`. This behavior should be consistent with how `cmd` handles spaces.
Here is a possible implementation.

```cs
private static async Task<bool> LaunchApplicationFromPath(string currentInput, string workingDir)
{
	currentInput = currentInput.TrimStart();

	var positionOfDelimiter = currentInput.IndexOf(' ');
	if (currentInput.ElementAt(0) == '"')
		positionOfDelimiter = currentInput.IndexOf('"', 1);

	// if no spaces or doesn't start with a '"'
	if (positionOfDelimiter == -1)
		positionOfDelimiter = currentInput.Length - 1;

	positionOfDelimiter += 1; // include delimiter in fileName
	var fileName = currentInput.Substring(0, positionOfDelimiter);
	var arguments = currentInput.Substring(positionOfDelimiter);

	return await LaunchHelper.LaunchAppAsync(fileName, arguments, workingDir);
}
```

### Files Version

3.4.1.0 - Dev

### Windows Version

10.0.22631.3593

### Log File

(not required)

FieryRMS:
**Resolved / Related Issues**
- Closes #15455 

1. Create `test1.bat` and `test 1.bat` both with the same contents as below and place it in `C:\Folder`,
    ```bat
    @echo off
    echo %*
    ```
2. Run each line from below through the address bar,
    ```
    // should fail
    C:\Folder\test 1
    C:\Folder\test 1 param1
    C:\Folder\test 1"
    "C:\Folder\test 1 param1
    "C:\Folder\test 1 param1"
    
    // should pass
    C:\Folder\test 1.bat
    "C:\Folder\test 1"
    "C:\Folder\test 1" param1
    "C:\Folder\test 1" param1 param2 "param 3"
    "C:\Folder\test 1
    C:\Folder\test1
    C:\Folder\test1 param1
    C:\Folder\test1 param1 param2 "param 3"
    "C:\Folder\test 1.bat"     param1           param2
    ```

**Undefined behavior**
The below work as well, although it's not clear how exactly these should be handled.
```
C:\Folder\test1""
"C:\Folder\test1"""" param1 param2
"C:\Folder\test 1""""" param1 param2 "param 3"
```


hishitetsu:
Shouldn't we use `CommandLineParser`?
https://github.com/files-community/Files/blob/main/src/Files.App/Utils/CommandLine/CommandLineParser.cs

hishitetsu (approved):
LGTM

0x5bfa (approved):
Lets use CreateProcess in Win32API in the future to avoid this thing!

But, code wise, LGTM.

yaira2 (approved):

PR #15469, opened 2024-05-25
Feature: Updated the design for Dual Pane mode

0x5bfa:
TBA in the next release officially.

- Remove accent color on the bottom for the active pane to display 
- Use shadow and focus background only for the active pane

1. Open Files app
2. Open the one more pane
3. Open InfoPane
4. Make window smaller and see InfoPane at the bottom

**Screenshots**
image

### DIFF
--- a/src/Files.App/UserControls/SideBar/SideBarView.xaml
+++ b/src/Files.App/UserControls/SideBar/SideBarView.xaml
@@ -49,10 +49,12 @@
 			Canvas.ZIndex="10"
 			ContextRequested="PaneColumnGrid_ContextRequested"
 			XYFocusKeyboardNavigation="Enabled">
+
 			<Grid.RowDefinitions>
 				<RowDefinition Height="*" />
 				<RowDefinition Height="Auto" />
 			</Grid.RowDefinitions>
+
 			<Grid.RenderTransform>
 				<CompositeTransform x:Name="PaneColumnGridTransform" />
 			</Grid.RenderTransform>
@@ -122,15 +124,21 @@
 				UseSystemFocusVisuals="True" />
 		</Border>
 
-		<ContentPresenter
+		<Grid
 			x:Name="ContentPresenter"
 			Grid.Row="1"
 			Grid.Column="2"
 			Margin="2,0,8,0"
-			HorizontalAlignment="Stretch"
-			Content="{x:Bind InnerContent, Mode=OneWay}"
-			TabFocusNavigation="Local" />
+			HorizontalAlignment="Stretch">
 
+			<!--  Content  -->
+			<Border
+				x:Name="Content"
+				VerticalAlignment="Stretch"
+				Child="{x:Bind InnerContent, Mode=OneWay}"
+				TabFocusNavigation="Local" />
+
+		</Grid>
 		<VisualStateManager.VisualStateGroups>
 			<VisualStateGroup x:Name="DisplayModes">
 				<VisualState x:Name="Expanded" />
--- a/src/Files.App/UserControls/SideBar/SideBarView.xaml
+++ b/src/Files.App/UserControls/SideBar/SideBarView.xaml
@@ -49,12 +49,10 @@
 			Canvas.ZIndex="10"
 			ContextRequested="PaneColumnGrid_ContextRequested"
 			XYFocusKeyboardNavigation="Enabled">
-
 			<Grid.RowDefinitions>
 				<RowDefinition Height="*" />
 				<RowDefinition Height="Auto" />
 			</Grid.RowDefinitions>
-
 			<Grid.RenderTransform>
 				<CompositeTransform x:Name="PaneColumnGridTransform" />
 			</Grid.RenderTransform>
@@ -124,21 +122,15 @@
 				UseSystemFocusVisuals="True" />
 		</Border>
 
-		<Grid
+		<ContentPresenter
 			x:Name="ContentPresenter"
 			Grid.Row="1"
 			Grid.Column="2"
 			Margin="2,0,8,0"
-			HorizontalAlignment="Stretch">
+			HorizontalAlignment="Stretch"
+			Content="{x:Bind InnerContent, Mode=OneWay}"
+			TabFocusNavigation="Local" />
 
-			<!--  Content  -->
-			<Border
-				x:Name="Content"
-				VerticalAlignment="Stretch"
-				Child="{x:Bind InnerContent, Mode=OneWay}"
-				TabFocusNavigation="Local" />
-
-		</Grid>
 		<VisualStateManager.VisualStateGroups>
 			<VisualStateGroup x:Name="DisplayModes">
 				<VisualState x:Name="Expanded" />
--- a/src/Files.App/Views/MainPage.xaml
+++ b/src/Files.App/Views/MainPage.xaml
@@ -235,7 +235,7 @@
 						<ColumnDefinition
 							x:Name="ContentColumn"
 							Width="*"
-							MinWidth="100" />
+							MinWidth="208" />
 						<ColumnDefinition Width="Auto" />
 						<ColumnDefinition x:Name="PaneColumn" Width="Auto" />
 					</Grid.ColumnDefinitions>
--- a/src/Files.App/Views/MainPage.xaml
+++ b/src/Files.App/Views/MainPage.xaml
@@ -273,7 +273,7 @@
 						Content="{x:Bind ((viewmodels:MainPageViewModel)DataContext).SelectedTabItem.ContentFrame, Mode=OneWay}"
 						CornerRadius="8" />
 
-					<!--  Sizer  -->
+					<!--  Preview Pane Splitter  -->
 					<toolkit:GridSplitter
 						x:Name="PaneSplitter"
 						Grid.Row="1"
--- a/src/Files.App/Views/MainPage.xaml
+++ b/src/Files.App/Views/MainPage.xaml
@@ -266,8 +266,7 @@
 						Grid.Column="0"
 						HorizontalAlignment="Stretch"
 						HorizontalContentAlignment="Stretch"
-						Content="{x:Bind ((viewmodels:MainPageViewModel)DataContext).SelectedTabItem.ContentFrame, Mode=OneWay}"
-						CornerRadius="8" />
+						Content="{x:Bind ((viewmodels:MainPageViewModel)DataContext).SelectedTabItem.ContentFrame, Mode=OneWay}" />
 
 					<!--  Preview Pane Splitter  -->
 					<toolkit:GridSplitter
--- a/src/Files.App/Views/MainPage.xaml
+++ b/src/Files.App/Views/MainPage.xaml
@@ -266,10 +266,6 @@
 						Grid.Column="0"
 						HorizontalAlignment="Stretch"
 						HorizontalContentAlignment="Stretch"
-						Background="{ThemeResource App.Theme.FileArea.BackgroundBrush}"
-						BackgroundSizing="InnerBorderEdge"
-						BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
-						BorderThickness="1"
 						Content="{x:Bind ((viewmodels:MainPageViewModel)DataContext).SelectedTabItem.ContentFrame, Mode=OneWay}"
 						CornerRadius="8" />
 
--- a/src/Files.App/Views/MainPage.xaml
+++ b/src/Files.App/Views/MainPage.xaml
@@ -296,7 +296,7 @@
 		</sidebar:SidebarView>
 
 		<VisualStateManager.VisualStateGroups>
-			<VisualStateGroup>
+			<VisualStateGroup x:Name="WindowAdaptiveStates">
 				<VisualState>
 					<VisualState.StateTriggers>
 						<AdaptiveTrigger x:Name="CollapseSearchBoxAdaptiveTrigger" MinWindowWidth="750" />
@@ -311,7 +311,7 @@
 					</VisualState.Setters>
 				</VisualState>
 			</VisualStateGroup>
-			<VisualStateGroup>
+			<VisualStateGroup x:Name="SidebarStates">
 				<VisualState x:Name="NormalSidebarState">
 					<VisualState.StateTriggers>
 						<AdaptiveTrigger MinWindowWidth="0" />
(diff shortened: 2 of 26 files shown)


src/Files.App/UserControls/SideBar/SideBarView.xaml
```diff
@@ -124,21 +122,15 @@
...
+		<ContentPresenter
 			x:Name="ContentPresenter"
 			Grid.Row="1"
 			Grid.Column="2"
 			Margin="2,0,8,0"
-			HorizontalAlignment="Stretch">
-
-			<!--  Content  -->
-			<Border
-				x:Name="Content"
```
yaira2, line 136: Are there any visual states connected to this?

0x5bfa: This isn't defined as x:Modifier="public" so I searched in this view but no occurrences found.
And others as well.

0x5bfa:
I moved background/borderbrush to each user control's Grid to remove unneeded xaml tree (like you did in #7824). Should I separate changes?

yaira2:
> I moved background/borderbrush to each user control's Grid to remove unneeded xaml tree (like you did in #7824). Should I separate changes?

I think it'll be easier for future reference if it's seperate.

yaira2:
Nice work on the shadows!

src/Files.App/Views/Shells/ModernShellPage.xaml
```diff
@@ -28,7 +21,18 @@
...
-	<Grid HorizontalAlignment="Stretch" wct:UIElementExtensions.ClipToBounds="True">
+	<Grid
+		x:Name="RootGrid"
+		wct:UIElementExtensions.ClipToBounds="True"
+		x:FieldModifier="public"
+		Background="{ThemeResource App.Theme.FileArea.BackgroundBrush}"
+		BackgroundSizing="InnerBorderEdge"
+		BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
+		BorderThickness="1"
+		CornerRadius="8,8,8,8">
```
yaira2, line 32:
```suggestion
		CornerRadius="8">
```
If all values are the same I think we can use a single number.

src/Files.App/Views/PaneHolderPage.xaml
```diff
@@ -70,84 +57,34 @@
...
+			Grid.Column="2"
+			x:Load="{x:Bind IsRightPaneVisible, Mode=OneWay}"
+			ContentChanged="Pane_ContentChanged"
+			Loaded="Pane_Loaded"
+			NavParams="{x:Bind NavParamsRight, Mode=OneWay}"
+			PaneHolder="{x:Bind}" />
 
 		<i:Interaction.Behaviors>
-			<icore:DataTriggerBehavior Binding="{x:Bind converters:MultiBooleanConverter.AndConvert(IsLeftPaneActive, IsRightPaneVisible), Mode=OneWay}" Value="True">
-				<icore:ChangePropertyAction
```
yaira2, line 112: Glad we can remove all these visual states! They were delicate to work with and had certain issues with theming.

src/Files.App/Views/PaneHolderPage.xaml
```diff
@@ -70,84 +57,34 @@
...
-
-		<!--  Pane Resizing Splitter  -->
+		<!--  Sizer  -->
 		<toolkit:GridSplitter
 			x:Name="PaneResizer"
 			Grid.Column="1"
-			Width="5"
-			MinWidth="5"
-			HorizontalAlignment="Left"
+			Width="4"
```
yaira2, line 64: I think width is already 4 by default, can you check?

src/Files.App/Views/PaneHolderPage.xaml
```diff
@@ -70,84 +57,34 @@
...
-			ResizeDirection="Auto"
-			Style="{StaticResource DefaultGridSplitterStyle}">
-			<toolkit:GridSplitter.RenderTransform>
-				<TranslateTransform X="0" />
-			</toolkit:GridSplitter.RenderTransform>
-		</toolkit:GridSplitter>
+			Style="{StaticResource DefaultGridSplitterStyle}" />
+
+		<!--  Pane Right  -->
+		<shellpages:ModernShellPage
```
yaira2, line 78: I thought this needed to be wrapped in a border for something. Perhaps @gave92 will know.

src/Files.App/Views/PaneHolderPage.xaml
0x5bfa: By the way those ones left can also be simplified by using Visual States.
You and I would say it is out of scope tho.
image

yaira2: Correct

yaira2: We'll work on this more when we implement item templates.

src/Files.App/Views/PaneHolderPage.xaml
0x5bfa: MinWidth yes. When I didn't set this prop explicity, it took 1.0* width.
image

yaira2: Okay

yaira2:
We should make sure to test this with Column View

src/Files.App/Views/PaneHolderPage.xaml
0x5bfa: For now reverted.
I thought itd be ok because when I saw your change on this this left in your single commit without discussion.
https://github.com/files-community/Files/pull/7824/commits/26957954d6ce4a99ea6d93916c06cdbad068b8f8#diff-532bef17cf7fcb9657770727df869172f428d84c811de9524fe79dccedda6fdcL74-R80

src/Files.App/Views/MainPage.xaml
```diff
@@ -272,12 +271,7 @@
 						BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
```
yaira2, line 271: I think the background and border can be removed from here.

src/Files.App/Views/PaneHolderPage.xaml
yaira2: I tried removing it once before but we had to add it back for some reason.

src/Files.App/Views/PaneHolderPage.xaml
```diff
@@ -56,6 +42,7 @@
 				x:Name="LeftColumn"
 				Width="*"
 				MinWidth="300" />
+			<ColumnDefinition x:Name="SizerColumnDefinition" Width="Auto" />
```
yaira2, line 45: I'm not sure if it's related to this exact change, but there seems to be some issue with the visual states.
image

0x5bfa: Wdym? Overwrapping?

yaira2: Yeah, it seems to overlap a couple pixels on the right pane.

0x5bfa: This is existing issue. You can see it in stable too

yaira2: I can't repro this in stable

yaira2: I guess we can solve in another PR.

yaira2: In fact, I can repro this with a single pane as well.

src/Files.App/Views/MainPage.xaml
0x5bfa: Done

yaira2: Thanks

src/Files.App/Views/PaneHolderPage.xaml
yaira2: > For now reverted.

Did you push this yet?

0x5bfa: I thought I did, thank you that you did it for me!

yaira2: I removed the background but I didn't do anything for this.

yaira2: I'm not sure if we need to revert it, everything seems to be working properly.

src/Files.App/Views/PaneHolderPage.xaml
yaira2: You're correct! However, it seems to be more noticable with these changes so I'll push a fix.

src/Files.App/Views/MainPage.xaml
```diff
@@ -236,7 +235,7 @@
 						<ColumnDefinition
 							x:Name="ContentColumn"
 							Width="*"
-							MinWidth="100" />
+							MinWidth="208" />
```
yaira2, line 238: This provides enough space for two panes (each pane has a min width of `100`).

src/Files.App/Views/PaneHolderPage.xaml.cs
```diff
@@ -389,7 +418,7 @@ private void PaneResizer_ManipulationStarted(object sender, ManipulationStartedR
 
 		private void PaneResizer_ManipulationCompleted(object sender, ManipulationCompletedRoutedEventArgs e)
 		{
-			if (PaneRight is not null && PaneRight.ActualWidth <= 300)
+			if (PaneRight is not null && PaneRight.ActualWidth <= 100)
```
yaira2, line 421: This change is to match the new min width for panes

src/Files.App/Views/PaneHolderPage.xaml.cs
```diff
@@ -342,6 +346,8 @@ private void Pane_Loaded(object sender, RoutedEventArgs e)
 		{
 			((UIElement)sender).GotFocus += Pane_GotFocus;
 			((UIElement)sender).RightTapped += Pane_RightTapped;
+
+			PaneLeft.RootGrid.Translation = new System.Numerics.Vector3(0, 0, 8);
```
yaira2, line 350: I moved this to the loaded event, otherwise it would have triggered whenever the user navigated to a different page.

src/Files.App/Views/MainPage.xaml
```diff
@@ -267,17 +266,8 @@
...
-						BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
-						BorderThickness="1"
 						Content="{x:Bind ((viewmodels:MainPageViewModel)DataContext).SelectedTabItem.ContentFrame, Mode=OneWay}"
-						CornerRadius="8"
-						Translation="0,0,8">
-						<ContentPresenter.Shadow>
-							<ThemeShadow x:Name="PageContentThemeShadow" />
-						</ContentPresenter.Shadow>
-					</ContentPresenter>
+						CornerRadius="8" />
```
yaira2, line 270: Adding a padding of `4` fixes the shadow issue, so my theory is that the shadow is there and it's getting clipped.

XTorLukas:
There's an issue where the preview doesn't load when switching between panels

https://github.com/files-community/Files/assets/84145589/17d02cf5-0011-4d24-91f2-f9332c5a154f

XTorLukas:
How about instead of changing the shadow, which isn't as pronounced, you use the pane selector.

https://github.com/files-community/Files/assets/84145589/0a4335e4-594f-4af2-b69b-7fb3f570b42f

yaira2:
Additional design work will take place in a second PR.

src/Files.App/Views/PaneHolderPage.xaml.cs
yaira2, line 350: @0x5bfa is this needed, or does GotFocus always get called on startup?

src/Files.App/Views/PaneHolderPage.xaml.cs
```diff
@@ -363,6 +369,28 @@ private void Pane_GotFocus(object sender, RoutedEventArgs e)
 			var activePane = isLeftPane ? PaneLeft : PaneRight;
 			if (ActivePane != activePane)
 				ActivePane = activePane;
+
+			// Add theme shadow to the active pane
+			if (isLeftPane)
+			{
+				PaneRight?.ShellContentThemeShadow.Receivers.Remove(StatusBar);
+				PaneLeft?.ShellContentThemeShadow.Receivers.Add(StatusBar);
```
yaira2, line 377: Can we add the receivers to both panes on the initial load instead of on GotFocus?

src/Files.App/Views/PaneHolderPage.xaml.cs
0x5bfa: No, it looks like not called on the startup, therefore, added.
Do you think it doesn’t have to be focused on the startup?

src/Files.App/Views/PaneHolderPage.xaml.cs
yaira2: We'll still need to adjust the translation on GotFocus, but this way we can at least optimize the receivers.

src/Files.App/Views/PaneHolderPage.xaml.cs
yaira2: What if we assign the default translation from XAML?

yaira2: We'll still have to adjust it when the focus changes, but at least we'll be able to remove this line and clean up the code behind a bit.

src/Files.App/Views/PaneHolderPage.xaml.cs
0x5bfa: I suppose yeah.

0x5bfa: And I’m not sure but should receiver be StatusBar?
I’m wondering if it should be RootContentGrid.
Or is this intended to make sure shadow will be only on the sidebar and be clipped on the left/right/top border?

src/Files.App/Views/PaneHolderPage.xaml.cs
0x5bfa: > What if we assign the default translation from XAML?

Thought the same. Code will be simplified but have to call VisualState.GoToState(this, “LeftShadowOn”, true);
However this makes me one more hurdle to refactor the pane.

yaira2: Let's keep as is for now

src/Files.App/Views/PaneHolderPage.xaml.cs
yaira2: > I’m wondering if it should be [MainPage.RootGrid](https://github.com/files-community/Files/blob/main/src/Files.App/Views/MainPage.xaml#L221).

Let's try RootGrid and see what it looks like.

mdtauk: I made the same suggestion, the shadow needs to fall over anywhere in the app window.

yaira2: The truth is, we're not going to see a difference on Windows 11. Let's apply to RootGrid and test on Windows 10.

src/Files.App/Views/Shells/BaseShellPage.cs
```diff
@@ -142,6 +142,18 @@ public bool IsCurrentInstance
 						_IsCurrentInstanceTCS = new();
 
 					NotifyPropertyChanged(nameof(IsCurrentInstance));
+
+					if (!IsColumnView)
```
yaira2, line 146: What about Column View?

src/Files.App/Views/Shells/BaseShellPage.cs
```diff
@@ -142,6 +142,18 @@ public bool IsCurrentInstance
...
+					{
+						if (value)
+						{
+							VisualStateManager.GoToState(this, "ShellBackgroundFocusOnState", true);
+						}
```
yaira2, line 151: Can you remove the brackets?

yaira2 (approved):
LGTM

src/Files.App/Views/Shells/BaseShellPage.cs
0x5bfa: Avoided to getting notified from column focus change.

I strongly suppose this shell concept is instantaniously away from the earth, the codebase makes us to implement extensibly. I'll keep stick around this.

src/Files.App/Views/Shells/BaseShellPage.cs
0x5bfa: Sure.

src/Files.App/Views/Shells/BaseShellPage.cs
yaira2: What's the plan to change the background color in that case?

0x5bfa: Wdym?

yaira2: The active pane background color doesn't update if one of the panes is Columns View.

yaira2: Never mind

Issue #8974, opened 2022-04-21
Refactor - Storage Migration


cinqmilleans:
### What's the Problem?

This issue discusses the migration from Storage to the backend. This step is preliminary to the ListedItem refactor.
With this discussion, we can parallelize the changes to save time. @lu

### Solution/Idea

Items in the Filesystem folder should be removed from unwanted dependencies.
- Concerning localized texts. These dependencies can remain because the backend uses them. There are some cases where localized text is used to identify types. It would be better to use a enum or non-localized string but that can be done after the migration.
- Regarding dependencies on ListedItem and its derivatives. They prevent migration but are easy to remove by reversing the dependencies.
- The ViewModel and App dependencies are very annoying and you have to find the right way to remove them.
- Settings dependencies should be replaced by properties. This is the case with enumerators.

We take the opportunity to rework the style to clarify the files. The order of the elements is completely random. PRs are already pending. We can also check what is should be public or not. It will also be necessary to manage nullables because the backend manages them differently.

### Alternatives

None

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish W | Must |
| This proposal will allow end users to accomplish X | Should |
| This proposal will allow developers to accomplish Y | Could |
| This proposal will allow end users to accomplish Z | Won't |

gave92:
Could you point me a couple of examples where ViewModel and App are referenced?

cinqmilleans:
FilesystemOperations depends on IShellPage to access the ItemViewModel, for example here:
`FilesystemResult<BaseStorageFolder> destinationResult = await associatedInstance.FilesystemViewModel.GetFolderFromPathAsync(PathNormalization.GetParentDir(destination));`
GetFolderFromPathAsync depends on the current page. This dependency must therefore be circumvented.

In FilesystemHelpers, we find for example:
`App.JumpList.RemoveFolder(source.Path); // Remove items from jump list`

gave92:
Ah ok, I assumed this was just about the StorageFile/Folder implementations. The ugliest thing is probably FilesystemOperations accessing/editing the UI FilesAndFolders collection directly.

lukeblevins:
Edited to include my ideas for #8405 which was somewhat informed by @d2dyno1's code review suggestions a couple months ago.

![RefactoringEnum](https://user-images.githubusercontent.com/20365014/164537240-e1632215-306b-40c8-88c7-665afc4b7c9c.png)

[...]

PR #15477, opened 2024-05-26
Added initial SFTP implementation using SSH.NET

### LINKED ISSUE #2407: SFTP support
Please add SFTP support for Files-Beta.
I use KDE-Connect to access files on my android phone, adding SFTP client support would make life easier for many

**Additional context**
Add any other context or screenshots about the feature request here. Ex: What device type/UI layout and app version?

itsWindows11:
**Resolved / Related Issues**

- For #2407

Currently waiting for #8974 to be implemented (since the current way of hooking filesystems to the UI is a total mess I cannot figure out) along with #15473, this PR is to lay the groundwork for resolving #2407 and will be marked as ready when everything is finished.


0x5bfa:
This is around Storage Abstraction and it'll take a huge amount of time to get this abstraction ready for actual use.
@yaira2  we might as well merge like we have done as leave until finish.
If you agree, I'll take a look and leave a review.

yaira2:
Sounds good

src/Files.Core.Storage/IFtpStorageService.cs
```diff
@@ -6,7 +6,7 @@ namespace Files.Core.Storage
 	/// <summary>
 	/// Provides an abstract layer for accessing an ftp file system
 	/// </summary>
-	public interface IFtpStorageService : IStorageService
+	public interface ISftpStorageService : IStorageService
```
yaira2, line 9: Should Sftp and Ftp be kept seperate?

itsWindows11: Yup that's intended but I was copy pasting certain things to save time and I accidentally changed it here. Will change 👍

0x5bfa (approved):
LGTM otherwise!

You seem to have implemented for BaseStorageItem.
Is that working well?
If so, how did you validate/use SFTP storage?

itsWindows11:
~~It's a draft for a reason 🙃~~ I'll do the testing outside of Files in a separate project, until I can figure out the UI part.

I've seen what's implemented for FTP support and tried to go for a similar implementation since it's pretty much the only non-shell-based storage implementation here with different protocols (AFAIK)

yaira2:
@itsWindows11 what the status of this PR?

0x5bfa:
Storage abstraction won't be implemented until the next year unless someone take over in Files and given OwlCore's progress. I think we should wrap up in order to merge.

tmds:
Hi all, I wonder if you may be open to use [Tmds.Ssh](https://github.com/tmds/Tmds.Ssh) to implement this feature?

The `SftpClient` is fully async and it has methods that match well with what you need. Additionally, the upload/download methods are optimized to parallelize the transfer while keeping the overall .NET memory allocations low.

I'd be interested in any feedback you have while using the library.

itsWindows11:
Sounds like a better option, but my SSH.NET issue regarding async IO support seems to be addressed so I don't think there's a need to switch.

The main blocker is the storage abstraction though, we either need to get this merged with switching the methods to be async and someone completes the integration, or wait until we can use OwlCore.Storage w/ extensions on the Files end.

tmds:
@itsWindows11 the reason to consider switching could be performance. From the limited comparisons I did, Tmds.Ssh was considerably faster.

itsWindows11:
I have just played around a bit with your library while working on an [OwlCore.Storage](https://github.com/Arlodotexe/OwlCore.Storage) wrapper.

Besides tmds/Tmds.Ssh#239, there are a few issues I've encountered with the library, especially when enumerating contents in directories:

- In mysterious cases that I can't seem to investigate, an `InvalidOperationException` is randomly thrown when trying to enumerate items in a directory.
- Authentication requires a lot of work than required when compared to SSH.NET, specifically around known hosts & certificate management.

tmds:
> I have no clue what this might be about. Do you have an exception message and stacktrace?

@itsWindows11 if you have some time for it, this would be very useful to help me understand the issue.

> It's similar to accepting an untrusted certificate in your browser.

You may want to address this regardless of the SSH library you are using. You could for example also store the sha256fingerprint of the ssh server. To help the user, you can use the SSH library to retrieve the fingerprint.

For reference, this is the `ssh` UX:

```
❯ ssh github.com
The authenticity of host 'github.com (140.82.121.4)' can't be established.
ED25519 key fingerprint is SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU.
This key is not known by any other names.
Are you sure you want to continue connecting (yes/no/[fingerprint])?
```

(Note that Tmds.Ssh integrates with the OpenSSH `known_hosts` file so it shares the trust information with other ssh tools used on the system.)

lakani:
how can i test it ?

Lamparter:
@lakani you'll need to build from source from the [`itswindows11:feature/sftp`](https://github.com/itsWindows11/Files/tree/feature/sftp) branch.
I'm not sure whether it will work though.

PR #15482, opened 2024-05-27
Code Quality: Change active pane border thickness in high contrast

yaira2:
**Resolved / Related Issues**

- Increased border thickness for active pane in high contrast mode

image


src/Files.App/Views/Shells/ModernShellPage.xaml
```diff
@@ -84,11 +100,13 @@
 				<VisualState x:Name="ShellBackgroundFocusOnState">
 					<VisualState.Setters>
 						<Setter Target="RootGrid.Background" Value="{ThemeResource App.Theme.FileArea.BackgroundBrush}" />
+						<Setter Target="RootGrid.BorderThickness" Value="{ThemeResource Local.ActivePane.BorderThickness}" />
```
yaira2, line 103: We need a way to only apply this resource when dual pane is enabled.

yaira2: It seems that we'll need another Visual State for when Dual Pane is turned off.

0x5bfa: Yeah this is for background so we can create ShellBorderThicknessFocusOnState and OffState.

0x5bfa (commented):
LGTM, code wise.

PR #15490, opened 2024-05-28
Code Quality: Fixed an issue where the forward icon was wrongly clipped

0x5bfa:
### Summary

This issue has been there but we’d never noticed because there’s splitter that had the same color and it’d overwritten.

### Resolved / Related Issues

- https://discord.com/channels/725513575971684472/725514946112127057/1245057778788208691

image

### DIFF
--- a/src/Files.App/Views/Shells/ModernShellPage.xaml
+++ b/src/Files.App/Views/Shells/ModernShellPage.xaml
@@ -50,11 +50,12 @@
 			<ThemeShadow x:Name="ShellContentThemeShadow" x:FieldModifier="public" />
 		</Grid.Shadow>
 
+		<!--  Swipe Backward Icon  -->
 		<Border
 			x:Name="BackIcon"
 			Width="48"
 			Height="48"
-			Margin="-1,0,0,0"
+			Margin="-4,0,0,0"
 			HorizontalAlignment="Left"
 			VerticalAlignment="Center"
 			wct:VisualExtensions.NormalizedCenterPoint="0.5,0.5"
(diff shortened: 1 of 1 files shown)


0x5bfa:
cc @XTorLukas

XTorLukas:
Tested and bug fixed 😊

XTorLukas (approved):
LGTM

0x5bfa:
Thank you!

yaira2:
What are the steps to repro this issue?

heftymouse (changes requested):
the icons are now hidden for most of the swipe gesture, they only appear at the very end after 64 pixels. this is a regression that makes it feel a lot less responsive to use

instead, a simpler fix would be to only increase the offset in the expression animation from 48 to something like 52, this would push it off the side of the pane by a few pixels. I've tested this and can confirm it works and keeps all of the existing behaviour.

XTorLukas:
All changes can be deleted, the solution is just to set the margin here:

https://github.com/files-community/Files/blob/f3e2d2a895ca244c22197e8d4a16285bac6372a9/src/Files.App/Views/Shells/ModernShellPage.xaml#L57

```
Margin="-1,0,4,0"
```

https://github.com/files-community/Files/blob/f3e2d2a895ca244c22197e8d4a16285bac6372a9/src/Files.App/Views/Shells/ModernShellPage.xaml#L78

```
Margin="1,0,-4,0"
```

heftymouse:
amazing, even better! the 1 and -1 can probably be removed too

XTorLukas:
No -1 and 1 must remain here, the problem would then occur on the other side
Prevents the problem on this side
image

heftymouse:
sorry, i mean like `-4,0,0,0` and `0,0,-4,0`

XTorLukas:
> sorry, i mean like `-4,0,0,0` and `0,0,-4,0`

Yes, that's enough. I think it's more elegant.

XTorLukas (approved):
LGTM, It's simple, but it's work.

heftymouse (approved):

PR #15493, opened 2024-05-28
Fix: Fixed an issue where panes couldn't be switched via touch

XTorLukas:
**Resolved / Related Issues**

https://discord.com/channels/725513575971684472/958783217941487616/1245094972043296848
- An issue where there is no switching between panels using the touch screen.

1. Switch between panels in Dual Pane mode via touch on touch screen

### DIFF
--- a/src/Files.App/Views/PaneHolderPage.xaml.cs
+++ b/src/Files.App/Views/PaneHolderPage.xaml.cs
@@ -339,9 +339,7 @@ namespace Files.App.Views
 		}
 
 		private void Pane_PointerPressed(object sender, PointerRoutedEventArgs e)
-		{
-			((UIElement)sender).Focus(FocusState.Pointer);
-		}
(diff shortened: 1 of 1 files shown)


yaira2 (dismissed):
This looks good to me but I'd like to wait until @hishitetsu can review.

hishitetsu (approved):
This also slightly changes the behavior when using a mouse. The pane will switch when the button is pressed instead of when the button is released, but it would rather do so.
LGTM

XTorLukas:
@hishitetsu If we want to specify more precisely the finger tap, I can use `Tapped` instead of `PointerPressed`

hishitetsu:
> @hishitetsu If we want to specify more precisely the finger tap, I can use `Tapped` instead of `PointerPressed`

I'd rather use `PointerPressed`.

PR #15532, opened 2024-06-02
Fix: Fixed an issue where text on the Actions Settings page didn't wrap

XTorLukas:
**Fix when the description text is not displayed in full**

From:
image
To:
image


yaira2 (approved):
LGTM

PR #15535, opened 2024-06-02
Code Quality: Replace AppCenter with Sentry

yaira2:
**Resolved / Related Issues**

- Closes #

1. Confirmed custom events display in the Exceptionless dashboard


0x5bfa (approved):
Let's change PRIVACY.md too
Code wise, LGTM.

hez2010:
Is unhandled exception being reported too?

yaira2:
What's a good way to force a crash?

hishitetsu:
You can embed code that throws an Exception in an action and call that action.
I'll review this tomorrow.

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
```diff
@@ -109,23 +111,25 @@ public static async Task CheckAppUpdate()
...
 			try
 			{
-				if (!Microsoft.AppCenter.AppCenter.Configured)
+				if (!ExceptionlessClient.Default.Configuration.IsValid)
 				{
-					Microsoft.AppCenter.AppCenter.Start(
-						Constants.AutomatedWorkflowInjectionKeys.AppCenterSecret,
-						typeof(Microsoft.AppCenter.Analytics.Analytics),
-						typeof(Microsoft.AppCenter.Crashes.Crashes));
+					var fakeID = Guid.NewGuid().ToString();
```
yaira2, line 122: We might need to store the fake id across sessions but I think that's fine.

0x5bfa: If this is going to be used we should show it in About to let user know that's session ID.
And output into the debug.log

yaira2: This is actually a user ID, there is another thing we can do which is to generate reference IDs for specific crashes but it's out of scope for this PR.

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
```diff
@@ -259,6 +259,8 @@ public static void HandleAppUnhandledException(Exception? ex, bool showToastNoti
 
 			if (ex is not null)
 			{
+				ex.ToExceptionless().Submit();
```
yaira2, line 262: I wonder if the app is closing before the client has time to submit the log.

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
```diff
@@ -259,6 +259,12 @@ public static void HandleAppUnhandledException(Exception? ex, bool showToastNoti
 
 			if (ex is not null)
 			{
+				ex.ToExceptionless()
+					.SetReferenceId(Guid.NewGuid().ToString())
+					.Submit();
+
+				await ExceptionlessClient.Default.ProcessQueueAsync();
```
hez2010, line 269: The method returns when it hit the first `await` and causes the app closing.
You need to use `ExceptionlessClient.Default.ProcessQueueAsync().ConfigureAwait(false).GetAwaiter().GetResult()` instead, also make this method non-async.

yaira2:
> Is unhandled exception being reported too?

Good call, I had to make some changes to get this working.

hez2010 (changes requested):

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
yaira2: Should I set this here or when I first configure the client?

yaira2:
I've been playing around with Exceptionless and it doesn't seem like it's going to fit our needs.

.github/PRIVACY.md
yaira2, line 1: This file needs to be updated

src/Files.App/App.xaml.cs
```diff
@@ -90,14 +90,14 @@ async Task ActivateAsync()
...
-#endif
-
 				var userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
 				var isLeaveAppRunning = userSettingsService.GeneralSettingsService.LeaveAppRunning;
 
+#if STORE || STABLE || PREVIEW
+				// Configure Sentry
+				AppLifecycleHelper.ConfigureSentry();
+#endif
+
```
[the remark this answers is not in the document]
yaira2: It's probably not applicable with Sentry, but when we were testing Exceptionless, we needed access to settings to retrieve the user id.

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
```diff
@@ -259,6 +255,8 @@ public static void HandleAppUnhandledException(Exception? ex, bool showToastNoti
 
 			if (ex is not null)
 			{
+				SentrySdk.CaptureException(ex);
```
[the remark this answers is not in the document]
yaira2: It might be worth moving this to the method that writes logs.

src/Files.App/App.xaml.cs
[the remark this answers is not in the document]
hishitetsu: Then the original location is better.

src/Files.App/Services/Settings/ActionsSettingsService.cs
```diff
@@ -22,13 +20,6 @@ public ActionsSettingsService(ISettingsSharingContext settingsSharingContext)
 
 		protected override void RaiseOnSettingChangedEvent(object sender, SettingChangedEventArgs e)
 		{
-			switch (e.SettingName)
-			{
-				case nameof(ActionsV2):
-					Analytics.TrackEvent($"Set {e.SettingName} to {e.NewValue}");
```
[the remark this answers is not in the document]
yaira2: I'd like to rethink our approach to monitoring settings. The current system in AppCenter falls short as it doesn’t accurately reflect active user engagement with specific settings. For example, if a user switches a setting on, off, and back on—this sequence skews the data, rendering the metrics less meaningful.

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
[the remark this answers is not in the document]
hishitetsu: That's a good idea.

src/Files.App/Services/Settings/ActionsSettingsService.cs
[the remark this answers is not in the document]
hishitetsu: Surely, we can start small and gradually enhance the information we monitor.

src/Files.App/App.xaml.cs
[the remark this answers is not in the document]
yaira2: Agreed

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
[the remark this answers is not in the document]
hishitetsu: Ah, but it would be better to be able to distinguish whether that exception caused the crash or not.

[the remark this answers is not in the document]
yaira2: That's a good point, we'll probably want to do this as needed.

src/Files.App/App.xaml.cs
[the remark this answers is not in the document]
yaira2: I added back the user id for Sentry so I think we need to leave this as is.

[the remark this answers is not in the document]
yaira2: This thread is resolved

src/Files.App (Package)/Package.appxmanifest
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
hishitetsu, line 1: I don't think it's a problem, but the byte order mark has been gone.

src/Files.App/Data/Commands/ActionCommand.cs
```diff
@@ -1,8 +1,8 @@
-﻿// Copyright (c) 2024 Files Community
+// Copyright (c) 2024 Files Community
```
hishitetsu, line 1: As above.

hishitetsu (approved):
LGTM

src/Files.App/Data/Commands/ActionCommand.cs
yaira2: Probably because I used the online editor.

0x5bfa: It woun't be problem because VS adds back.

src/Files.App (Package)/Package.appxmanifest
0x5bfa: It woun't be problem because VS adds back.

src/Files.App/Data/Commands/ActionCommand.cs
0x5bfa: As above.

PR #15537, opened 2024-06-02
Feature: Added support for bulk renaming items

### LINKED ISSUE #8191: Feature: Support renaming multiple files at the same time
### What's the Problem?

Only one file can be renamed at the same time.

### Solution/Idea

1. Using the rename action when multiple items are selected should open a "Rename items" dialog
2. The rename dialog should consist of a simple title and text box for entering the new name
3. Items containing different extensions should get the new name
4. Items with the same extension should have a number attached to the end

marianalcar:
**Resolved / Related Issues**

Previously, only one file could be renamed at the same time. We implemented the functionality of renaming multiple at the same time giving Files more flexibility and increased functionalities.

- Closes #8191 

1. Opened Files
2. Went to each layout view
3. Created different files and folders
4. Selected them all and renamed them at the same time.

### DIFF
--- a/src/Files.App/Actions/FileSystem/RenameAction.cs
+++ b/src/Files.App/Actions/FileSystem/RenameAction.cs
@@ -34,15 +34,8 @@ namespace Files.App.Actions
 
 		public Task ExecuteAsync(object? parameter = null)
 		{	
-			if (context.SelectedItems.Count == 1)
-			{
-				context.ShellPage?.SlimContentPage?.ItemManipulationModel.StartRenameItem();
-			}
-			else
-			{
-				context.ShellPage?.SlimContentPage?.ItemManipulationModel.StartRenameItems();
-			}
-
+			
+			context.ShellPage?.SlimContentPage?.ItemManipulationModel.StartRenameItem();
 			return Task.CompletedTask;
 		}
 
--- a/src/Files.App/Actions/FileSystem/RenameAction.cs
+++ b/src/Files.App/Actions/FileSystem/RenameAction.cs
@@ -33,15 +33,22 @@ namespace Files.App.Actions
 		}
 
 		public Task ExecuteAsync(object? parameter = null)
-		{
-			context.ShellPage?.SlimContentPage?.ItemManipulationModel.StartRenameItem();
+		{	
+			if (context.SelectedItems.Count == 1)
+			{
+				context.ShellPage?.SlimContentPage?.ItemManipulationModel.StartRenameItem();
+			}
+			else
+			{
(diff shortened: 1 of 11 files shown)


src/Files.App/Actions/FileSystem/RenameAction.cs
```diff
@@ -33,15 +33,22 @@ public RenameAction()
 		}
 
 		public Task ExecuteAsync(object? parameter = null)
-		{
-			context.ShellPage?.SlimContentPage?.ItemManipulationModel.StartRenameItem();
+		{	
+			if (context.SelectedItems.Count == 1)
```
yaira2, line 37: Is this if statement needed? It looks like the else statement has the same logic.

src/Files.App/Actions/FileSystem/RenameAction.cs
```diff
@@ -33,15 +33,22 @@ public RenameAction()
...
+			{
+				context.ShellPage?.SlimContentPage?.ItemManipulationModel.StartRenameItem();
```
yaira2, line 39: What are your thoughts on moving the logic to within `StartRenameItem`?

0x5bfa:
Could you provide a screen recording?

yaira2:
> Using the rename action when multiple items are selected should open a "Rename items" dialog

I compiled this branch, but it seems to be missing the dialog when bulk renaming.

marianalcar:
> Could you provide a screen recording?

https://github.com/files-community/Files/assets/114589228/01b8d4f8-7f8b-4724-b706-bc0067c49113

src/Files.App/Actions/FileSystem/RenameAction.cs
marianalcar: We commited that change now. Thank you for the feedback

yaira2:
@marianalcar the behavior seems to closely mimic File Explorer, while it technically works, we were hoping to add more advanced options in the future and therefore this feature requires a content dialog.

src/Files.App/Actions/FileSystem/RenameAction.cs
yaira2: Thanks!

yaira2:
Sorry about the confusion! For now, the rename dialog should consist of a simple title and text box for entering the new name. In the future, we may want to add more options.

yaira2:
@marianalcar are you still planning to work on this?

yaira2:
@marianalcar I've made the requested changes in #16228. Can you take a look and comment if you have any feedback?

PR #15539, opened 2024-06-02
Fixed an issue where foreground of some FontIcons was not set

0x5bfa:
Comes from #15501

### DIFF
--- a/src/Files.App/App.xaml
+++ b/src/Files.App/App.xaml
@@ -31,6 +31,10 @@
 				<Setter Target="HighContrastAdjustment" Value="None" />
 			</Style>
 
+			<Style TargetType="FontIcon">
+				<Setter Target="Foreground" Value="{ThemeResource App.Theme.IconBaseBrush}" />
+			</Style>
+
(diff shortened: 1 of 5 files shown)


src/Files.App/App.xaml
```diff
@@ -31,6 +31,10 @@
 				<Setter Target="HighContrastAdjustment" Value="None" />
 			</Style>
 
+			<Style TargetType="FontIcon">
```
yaira2, line 34: I don't think we want to target all font icons. It's best to do it as needed.

0x5bfa: But icons that use glyph are not rendered with correct foreground in context flyout with your idea.
If there’ll be a need to change color for specific items, we could set TextFillColorPrimaryBrush.
Even we don’t want to do this app-wide, could we add it to each layout page, sidebar, inner nav toolbar, address toolbar instead of setting to each?

yaira2: > But icons that use glyph are not rendered with correct foreground in context flyout with your idea.

Ideally we'd have a custom style for the menu items. This will also solve an issue with the font.

0x5bfa: All of them?
Then I should close this PR

Issue #15202, opened 2024-04-16
Bug: OneDrive items are auto downloaded


yaira2:
### Description

In recent preview builds, opening a OneDrive folder auto downloads the contents.

### Steps To Reproduce

1. Open OneDrive folder

### Requirements

- Don't auto download items

### Files Version

v3.3.2

### Windows Version

Windows 11

### Log File

NA

yaira2:
I can confirm that this issue isn't present in bb8031c0aa8ec3632f74d61271c5c73cd48ff715

yaira2:
Fyi @hez2010

yaira2:
It appears that this issue was introduced with a0acba3e8d5201bd67d63c770e5f339e4d257a76.

hez2010:
It turns out to be the tag initialization. Will be fixed in #15216.

yaira2:
This issue is back

yaira2:
I pushed a fix to `main`.

PR #15543, opened 2024-06-03
Feature: Added dimensions to tooltip when hovering over image files

### LINKED ISSUE #14975: Feature: Show image resolution on hover information

0x5bfa:
### Resolved / Related Issues

- Closes #14975

1. Open Files app
2. Hover over an image file
3. See if it is corrupt image or dimention property is unavailable it wont be shown

### DIFF
--- a/src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
+++ b/src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
@@ -405,7 +405,7 @@ namespace Files.App.Data.Factories
 				new ContextMenuFlyoutItemViewModel()
 				{
 					Text = "BaseLayoutItemContextFlyoutSetAs/Text".GetLocalizedResource(),
-					ShowItem = itemsSelected && (selectedItemsPropertiesViewModel?.IsSelectedItemImage ?? false),
+					ShowItem = itemsSelected && (selectedItemsPropertiesViewModel?.CanSelectedItemBeManipulated ?? false),
 					ShowInSearchPage = true,
 					Items =
 					[
@@ -419,13 +419,13 @@ namespace Files.App.Data.Factories
 				{
 					IsVisible = !currentInstanceViewModel.IsPageTypeRecycleBin
 								&& !currentInstanceViewModel.IsPageTypeZipFolder
-								&& (selectedItemsPropertiesViewModel?.IsSelectedItemImage ?? false)
+								&& (selectedItemsPropertiesViewModel?.CanSelectedItemBeManipulated ?? false)
 				}.Build(),
 				new ContextMenuFlyoutItemViewModelBuilder(Commands.RotateRight)
 				{
 					IsVisible = !currentInstanceViewModel.IsPageTypeRecycleBin
 								&& !currentInstanceViewModel.IsPageTypeZipFolder
-								&& (selectedItemsPropertiesViewModel?.IsSelectedItemImage ?? false)
+								&& (selectedItemsPropertiesViewModel?.CanSelectedItemBeManipulated ?? false)
 				}.Build(),
 				new ContextMenuFlyoutItemViewModelBuilder(Commands.RunAsAdmin).Build(),
 				new ContextMenuFlyoutItemViewModelBuilder(Commands.RunAsAnotherUser).Build(),
--- a/src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
+++ b/src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
@@ -405,7 +405,7 @@ namespace Files.App.Data.Factories
 				new ContextMenuFlyoutItemViewModel()
 				{
 					Text = "BaseLayoutItemContextFlyoutSetAs/Text".GetLocalizedResource(),
-					ShowItem = itemsSelected && (selectedItemsPropertiesViewModel?.CanSelectedItemBeManipulated ?? false),
+					ShowItem = itemsSelected && (selectedItemsPropertiesViewModel?.IsCompatibleToSetAsWindowsWallpaper ?? false),
 					ShowInSearchPage = true,
 					Items =
 					[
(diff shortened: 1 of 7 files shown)


hishitetsu (commented):
The condition of dimensions display availability can be improved, but it works fine.

hishitetsu (approved):
LGTM

src/Files.App/Data/Items/ListedItem.cs
```diff
@@ -326,6 +328,30 @@ public ObservableCollection<FileProperty> ItemProperties
...
+		{
+			get => imageHeight;
+			set
+			{
+				SetProperty(ref imageHeight, value);
+				OnPropertyChanged(nameof(DimensionsDisplay));
+			}
+		}
+
+		public string DimensionsDisplay => IsImage && ImageWidth > 0 && ImageHeight > 0 ? $"{ImageWidth} \u00D7 {ImageHeight}" : string.Empty;
```
yaira2, line 353: Is there a way to lazy load this property so that it only loads the first time we try retrieving it?

0x5bfa: 1. That’s what I’m actually thinking about in my Storage Layer concept for properties. But is this loaded more than once?
2. If we should we should make more flexible extension methods. Such as IsCompatibleToSetAsWallpaper and IsCompatibleToGetDimentions.
3. I thought the same but I saw this used in MSDN. https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/m/multiplication-sign

yaira2: 1. It's about only loading it when it's needed, otherwise we're spending compute time on a property that'll only be displayed on hover.
2. I think that we can keep `IsImage` as a general extension method for image files, and we can create a more specific one for `IsCompatibleToSetAsWallpaper`.
3. Character Map UWP has a feature to filter out deprecated icons, it might be worth checking there.

0x5bfa: All done.

0x5bfa: Testing rn.

0x5bfa: > I think the icon u00D7 is deprecated, can you try uE711 instead?

It is garbling.

image

0x5bfa:
Works fine on my end.
Ready for review again.

src/Files.Shared/Helpers/FileExtensionHelpers.cs
```diff
@@ -31,10 +31,20 @@ public static bool HasExtension(string? filePathToCheck, params string[] extensi
...
+		{
+			return HasExtension(fileExtensionToCheck, ".png", ".bmp", ".jpg", ".jpeg", ".jfif", ".gif", ".tiff", ".tif", ".webp");
+		}
+
+		/// <summary>
+		/// Checks if the file can be set as wallpaper.
+		/// </summary>
+		/// <param name="fileExtensionToCheck">The file extension to check.</param>
+		/// <returns><c>true</c> if the fileExtensionToCheck is an image; otherwise, <c>false</c>.</returns>
+		public static bool IsManipulateableImageFile(string? fileExtensionToCheck)
```
yaira2, line 43:
```suggestion
		public static bool IsCompatibleToSetAsWindowsWallpaper(string? fileExtensionToCheck)
```

src/Files.App/Data/Items/ListedItem.cs
yaira2: I'll check into the icon, as far as the first point, do you have any ideas on if this can be done?

src/Files.Shared/Helpers/FileExtensionHelpers.cs
0x5bfa: Sure.
FYI, I think when I refactor WallpaperHelper to be a servive called `WindowsWallpaperService`, I would add a method called `IsValidFileExtension(string extension)` to check if being a valid image file or not.

src/Files.App/Data/Items/ListedItem.cs
0x5bfa: First point means *Is there a way to lazy load this property so that it only loads the first time we try retrieving it?*?

I've implemented this using a property. Property doesn't prepare its value until actually requested to obtain.

src/Files.Shared/Helpers/FileExtensionHelpers.cs
0x5bfa: Done

src/Files.App/Data/Items/ListedItem.cs
mdtauk: It shouldn't be an icon, but the Multiply glyph
image

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
```diff
@@ -405,7 +405,7 @@ private static bool Check(ContextMenuFlyoutItemViewModel item, CurrentInstanceVi
 				new ContextMenuFlyoutItemViewModel()
 				{
 					Text = "BaseLayoutItemContextFlyoutSetAs/Text".GetLocalizedResource(),
-					ShowItem = itemsSelected && (selectedItemsPropertiesViewModel?.IsSelectedItemImage ?? false),
+					ShowItem = itemsSelected && (selectedItemsPropertiesViewModel?.CanSelectedItemBeManipulated ?? false),
```
yaira2, line 408:
```suggestion
					ShowItem = itemsSelected && (selectedItemsPropertiesViewModel?.IsCompatibleToSetAsWindowsWallpaper?? false),
```

Can you update the other locations as well?

0x5bfa: Done

yaira2 (approved):
LGTM

yaira2:
I noticed that it's not returning values for webp files, it would be nice to resolve this but it's not a blocker.

src/Files.App/Data/Items/ListedItem.cs
```diff
@@ -326,6 +329,41 @@ public ObservableCollection<FileProperty> ItemProperties
 			set => SetProperty(ref itemProperties, value);
 		}
 
+		public string DimensionsDisplay
```
yaira2, line 332: I discovered this is fetching as soon as the location is opened

yaira2: This also causes #15202

yaira2: I'll be commenting this out because the feature is low priority, but it'll be nice to add back once these issues are resolved.

0x5bfa: @hishitetsu The first implementer said this way is the fastest however speaking of this should we use WinRT API to retrieve the dimentions of images?

hishitetsu: I think the tooltip value can be empty at first, since the tooltip value is updated each time the pointer is entered.
https://github.com/files-community/Files/blob/0923cf0c03d17f48c97804f5fbe609a97185b576/src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs#L675-L680

> @hishitetsu The first implementer said this way is the fastest however speaking of this should we use WinRT API to retrieve the dimentions of images?

I don't know if using the WinRT API will solve any problems.

Issue #11735, opened 2023-03-15
Bug: Drag and drop to some apps still fails


hishitetsu:
### Description

#11728 improved drag and drop for several apps, including WhatsApp, but we know that drag and drop still does not work for some apps (Notepad, XShell, etc.). 
Also, #11728 doesn't improved it for virtual files that do not exist on the file system, such as files in archives.

### Steps To Reproduce

(Same as #10171)
1. open xshell and files app
2. drag a file to xshell

### Requirements

Drag and drop should work for all apps and all files, including virtual files.

### Files Version

2.4.51.0

### Windows Version

10.0.22621.1413

### Log file

n/a

Laiteux:
Can confirm it won't work with Notepad++ either.

chhus:
Dragging file to Microsoft Visio window also fails

0x5bfa:
Clipboard WinRT class probably generates incompatible format with some apps or we're doing in a wrong way.
Like the author of #12784, we probably might want to use [Win32 drag APIs](https://learn.microsoft.com/windows/win32/shell/clipboard#formats-for-transferring-file-system-objects) for this as almost all of them listed are Win32 apps probably in IWindowsClipboardService unless we found we were doing in a wrong way. Here's [the overview MSDN](https://learn.microsoft.com/windows/win32/dataxchg/clipboard).

yaira2:
A fix will be included in the next update.

yaira2:
@hishitetsu are any of the applications listed here resolved with #16628?

hishitetsu:
I confirmed that XShell was resolved and Notepad wasn't resolved. I haven't checked other applications, but I expect many of them to be resolved.

SemperFratres:
I can confirm that VMware has not been resolved.

Cinabutts:
came to this issue because i was trying to drag and drop some scripts into notepad++. doesn't work. notepad also does not work. this is absurd for a file explorer. lol please fix.

Cinabutts:
actually. need to comment again on this, how is this TWO YEARS OLD? lmfao, I love the features this app provides but godamn, priorities? lol

yair100:
@Cinabutts our top priorities for Files are currently stability and performance. That said, Files is open-source and we're happy to review pull requests from contributors. This particular issue has proven difficult to solve, every solution we've implemented in the past ended up introducing new issues.

Cinabutts:
[...]

PR #15555, opened 2024-06-05
Feature: Added support for dragging recent files

### LINKED ISSUE #8282: Feature: Add support for dragging recent files

Krytan:
### **Resolved / Related Issues**

To prevent extra work, all changes to the Files codebase must link to an approved issue marked as Ready to build. Please insert the issue number following the hashtag with the issue number that this Pull Request resolves.

Closes #8282

### **Steps used to test these changes**

1 - Open Files app
2 - Go to Home
3 -  Scroll down to Recent files
4 -  Dragging files should be possible

### **How does dragging work?**
Dragging recent files to other programs or outside Files will trigger "copy" similar to  windows File explorer, drags files to the desired location without changing the main path.

Dragging recent files inside the app Files will trigger "Move to", drags files to the desired location changing the main path to the new desired path,  similar to when dragging files from Documents or Downloads. 
file explorer uses "Copy" instead of "Move to" for this.

I can change to copy as well if it makes more sense.


src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -1,13 +1,13 @@
-// Copyright (c) 2024 Files Community
-// Licensed under the MIT License. See the LICENSE.
-
+using Microsoft.Extensions.Logging;
```
XTorLukas, line 1:
```suggestion

```
Is there a need?

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +31,54 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
 		{
 			ViewModel.BuildItemContextMenu(e.OriginalSource, e);
 		}
```
XTorLukas, line 33:
```suggestion
		=>	ViewModel.BuildItemContextMenu(e.OriginalSource, e);
```

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +31,54 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+				foreach (var item in items)
+				{
+					try
+					{
+						// Attempt to get the file from its path
+						var file = await StorageFile.GetFileFromPathAsync(item.RecentPath);
+						if (file != null)
+						{
+							storageItems.Add(file);
+						}
```
XTorLukas, line 58:
```suggestion
						if (file != null)
							storageItems.Add(file);
```

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +31,54 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+
+				if (storageItems.Count > 0)
+				{
+					// Create a new data package and set the storage items
+					DataPackage dataPackage = new DataPackage();
+					dataPackage.SetStorageItems(storageItems);
+					e.Data.SetDataProvider(StandardDataFormats.StorageItems, async request =>
+					{
+						request.SetData(storageItems);
+					});
```
XTorLukas, line 75:
```suggestion
					e.Data.SetDataProvider(StandardDataFormats.StorageItems, request => request.SetData(storageItems));
```

XTorLukas (changes requested):

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
Krytan: No need for logging thanks for spotting that

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +31,54 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
 		{
 			ViewModel.BuildItemContextMenu(e.OriginalSource, e);
 		}
+
+		private void AddDragAndDropSupport()
+		{
+			RecentFilesListView.CanDragItems = true;
+			// Registering event handlers for drag
+			RecentFilesListView.DragItemsStarting += RecentFilesListView_DragItemsStarting;
```
yaira2, line 39: It looks like this is also registered in the XAML, can this line be removed?

Krytan: That's correct RecentFilesListView.DragItemsStarting += RecentFilesListView_DragItemsStarting; can be removed

Krytan: also CanDragItems can be added to XAML so no need of a function called AddDragAndDropSupport, Thanks for pointing it out

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -27,8 +26,43 @@ private void RecentFilesListView_ItemClick(object sender, ItemClickEventArgs e)
 		}
 
 		private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEventArgs e)
+		=>	ViewModel.BuildItemContextMenu(e.OriginalSource, e);
```
XTorLukas, line 29:
```suggestion
		    =>	ViewModel.BuildItemContextMenu(e.OriginalSource, e);
```

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -27,8 +26,43 @@ private void RecentFilesListView_ItemClick(object sender, ItemClickEventArgs e)
...
+					DataPackage dataPackage = new DataPackage();
+					dataPackage.SetStorageItems(storageItems);
+					e.Data.SetDataProvider(StandardDataFormats.StorageItems, request => request.SetData(storageItems));
+
+					// Set the requested operation to Copy if dragging outside the application
+					e.Data.RequestedOperation = DataPackageOperation.Copy;
+				}
+			}
 		}
+
```
XTorLukas, line 66:
```suggestion
```

XTorLukas (approved):
LGTM

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -1,13 +1,12 @@
-// Copyright (c) 2024 Files Community
-// Licensed under the MIT License. See the LICENSE.
```
yaira2, line 2: Can you restore the license header?

XTorLukas: Yes, I missed that.

yaira2:
I'm trying to drag some text files into Notepad and while they can be dragged, dropping them into Notepad isn't opening the file.

Krytan:
Interesting, I will try to simulate that, I know that it works with google drive the web version

hishitetsu (commented):
Shouldn't it be possible to perform operations other than copying? Ideally, the process should be common to this.
https://github.com/files-community/Files/blob/fd18e70dfcc677978fdb1f46ba2cfd2a45045379/src/Files.App/ViewModels/Layouts/BaseLayoutViewModel.cs#L88

hishitetsu:
> I'm trying to drag some text files into Notepad and while they can be dragged, dropping them into Notepad isn't opening the file.

That is a known issue. It would be great if we could fix it. https://github.com/files-community/Files/issues/11735

Krytan:
@hishitetsu  Correct I checked with downloads and documents and they don't work when copying to notepad or text editor applications but I found I way to fix this

yaira2:
I used to not have any issues but I guess it's been a while since I tried dragging into Notepad.

Krytan:
@yaira2 @hishitetsu  dragging text files into Notepad or any text editor should work now for recent files

Krytan:
> Shouldn't it be possible to perform operations other than copying? Ideally, the process should be common to this.
> 
> https://github.com/files-community/Files/blob/fd18e70dfcc677978fdb1f46ba2cfd2a45045379/src/Files.App/ViewModels/Layouts/BaseLayoutViewModel.cs#L88

What other operations you got in mind? I believe the reason windows file explorer only uses copy for recent files is because recent files contain multiple files or folders from different paths/directories, so in case an accident happen the user is just copying and never messing with the original path/directory.

hishitetsu:
I tried that fix before, but iirc it didn't work when dragging multiple text files. However, the widget doesn't allow multiple files to be dragged, so this may be fine.

hishitetsu:
> What other operations you got in mind? I believe the reason windows file explorer only uses copy for recent files is because recent files contain multiple files or folders from different paths/directories, so in case an accident happen the user is just copying and never messing with the original path/directory.

What about the Link operation?

Krytan:
I see that could be okay to have I don't see any potential issues on it, I don't know why file explorer don't have that actually if is desirable I can add it too

yaira2:
> I tried that fix before, but iirc it didn't work when dragging multiple text files. However, the widget doesn't allow multiple files to be dragged, so it may be fine.

Would it work to apply this fix globally when single items are being dragged?

hishitetsu:
> I see. that could be okay to have I don't see any potential issues on it, I don't know why file explorer don't have that actually, if is desirable I can add it too

If Explorer only supports the copy operation, then leave it as it is.

Krytan:
@hishitetsu Explorer also supports Link operation and I am being struggling trying to implement it I was looking at your code and pretty much followed everything but still getting the circle with a cross, is not allowing me to do the link when I press shift

hishitetsu:
> @hishitetsu Explorer also supports Link operation and I am being struggling trying to implement it I was looking at your code and pretty much followed everything but still getting the circle with a cross, is not allowing me to do the link when I press shift

To do the link, don't you press the Alt key? Link operation worked when I simply replaced Copy in your PR with Link.

Krytan:
@hishitetsu I was trying to add dragover but found out I didn't needed to was simpler than I thought so now link operator works as well.

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,58 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+					try
+					{
+						// Attempt to get the file from its path
+						var file = await StorageFile.GetFileFromPathAsync(item.RecentPath);
+						if (file != null)
+							storageItems.Add(file);
+
+						// Read the file text content
+						var fileContent = await FileIO.ReadTextAsync(file);
+						fileContents.Add(fileContent);
```
hishitetsu, line 56: Shouldn't it only be executed if the file is not null and is a text file?

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,58 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+		private async void RecentFilesListView_DragItemsStarting(object sender, DragItemsStartingEventArgs e)
+		{
+			var items = e.Items.OfType<RecentItem>().ToList();
+			if (items.Count > 0)
+			{
+				var storageItems = new List<IStorageItem>();
+				var fileContents = new List<string>();
+
+
+				foreach (var item in items)
```
hishitetsu, line 45:
```suggestion
				var fileContents = new List<string>();

				foreach (var item in items)
```

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,58 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+						// Attempt to get the file from its path
+						var file = await StorageFile.GetFileFromPathAsync(item.RecentPath);
+						if (file != null)
+							storageItems.Add(file);
+
+						// Read the file text content
+						var fileContent = await FileIO.ReadTextAsync(file);
+						fileContents.Add(fileContent);
+
+					}
```
hishitetsu, line 58:
```suggestion
						fileContents.Add(fileContent);
					}
```

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,58 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+					dataPackage.SetStorageItems(storageItems);
+					e.Data.SetDataProvider(StandardDataFormats.StorageItems, request => request.SetData(storageItems));
+
+					if (fileContents.Count > 0)
+					{
+						// Create a new data package and set the file contents as text
+						dataPackage.SetText(string.Join("\n", fileContents));
+						e.Data.SetDataProvider(StandardDataFormats.Text, request => request.SetData(string.Join("\n", fileContents)));
+
+					}
```
hishitetsu, line 79:
```suggestion
						e.Data.SetDataProvider(StandardDataFormats.Text, request => request.SetData(string.Join("\n", fileContents)));
					}
```

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,58 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+					// Create a new data package and set the storage items
+					DataPackage dataPackage = new DataPackage();
+					dataPackage.SetStorageItems(storageItems);
+					e.Data.SetDataProvider(StandardDataFormats.StorageItems, request => request.SetData(storageItems));
+
+					if (fileContents.Count > 0)
+					{
+						// Create a new data package and set the file contents as text
+						dataPackage.SetText(string.Join("\n", fileContents));
+						e.Data.SetDataProvider(StandardDataFormats.Text, request => request.SetData(string.Join("\n", fileContents)));
```
hishitetsu, line 77:
```suggestion
						var text = string.Join("\n", fileContents)
						dataPackage.SetText(text);
						e.Data.SetDataProvider(StandardDataFormats.Text, request => request.SetData(text));
```

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,58 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+					}
+
+					// Set the requested operation to Copy if dragging outside the application
+					e.Data.RequestedOperation = DataPackageOperation.Copy | DataPackageOperation.Link;
+				}
+			}
+		}
+
+
 	}
```
hishitetsu, line 88:
```suggestion
		}
	}
```

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,58 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+						var file = await StorageFile.GetFileFromPathAsync(item.RecentPath);
+						if (file != null)
+							storageItems.Add(file);
+
+						// Read the file text content
+						var fileContent = await FileIO.ReadTextAsync(file);
+						fileContents.Add(fileContent);
+
+					}
+					catch (Exception)
```
hishitetsu, line 59:
```suggestion
					catch
```

hishitetsu (commented):
Feedback about code quality.

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -41,19 +41,23 @@ private async void RecentFilesListView_DragItemsStarting(object sender, DragItem
...
 						var file = await StorageFile.GetFileFromPathAsync(item.RecentPath);
 						if (file != null)
+						{
 							storageItems.Add(file);
 
-						// Read the file text content
-						var fileContent = await FileIO.ReadTextAsync(file);
-						fileContents.Add(fileContent);
+							//Attempt to read the file text content
+							var fileContent = await FileIO.ReadTextAsync(file);
```
hishitetsu, line 55: Sorry to keep reviewing, but wouldn't this cause a memory squeeze in the case of a large text file? It would be better to set an upper limit on the file size to be read.

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,60 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+						if (file != null)
+						{
+							storageItems.Add(file);
+
+							//Attempt to read the file text content
+							var fileContent = await FileIO.ReadTextAsync(file);
+							if (!string.IsNullOrEmpty(fileContent))
+							{
+								fileContents.Add(fileContent);
+							}
```
hishitetsu, line 59:
```suggestion
							if (!string.IsNullOrEmpty(fileContent))
								fileContents.Add(fileContent);
```

0x5bfa (approved):
LGTM (other than things @hishitetsu pointed out). Thank you! 🎉

Krytan:
@hishitetsu  added limit for reading text file to 50 mb,  and added a new method for ReadFileContent, felt that there was many ifs in one section.

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,77 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+						if (file != null)
+						{
+							storageItems.Add(file);
+							
+							// Attempt to check if the file have text content
+							var (hasContent, content) = await TryReadFileContentAsync(file);
+							if(hasContent)
+								fileContents.Add(content); 
+							
+						}
```
hishitetsu, line 59:
```suggestion
							if (hasContent)
								fileContents.Add(content); 
						}
```

hishitetsu (commented):
I have found that it inserts text into the existing file rather than opening the dragged file when dragging a file into Notepad where another file is already open. I'll wait for @yaira2 to decide if this is fine.

hishitetsu:
I have also found that I can't drag from recent files on the home page in Explorer to Notepad, even though I can drag from normal folders. So, I guess WinUI 3 doesn't support dragging files to Notepad.

Krytan:
> I have found that it inserts text into the already open tab rather than newly opening the dragged file when dragging a file into Notepad. I'll wait for @yaira2 to decide if this is fine.

That's correct and is not the expected behavior. the expected behavior is to open the file into notepad with the text on the tab. righter than creating a new file.

Krytan:
> I have also found that I can't drag from recent files on the home page in Explorer to Notepad, even though I can drag from normal folders. So, I guess WinUI 3 doesn't support dragging files to Notepad.

Correct. explorer doesn't let dragging from recent files to text editor applications. whenever WinUI 3 dosen't support it I am not so sure.

hishitetsu:
The only problem now with this PR is drag-and-drop to Notepad. I think it would be better to delete those codes so that this PR can be merged.

Krytan:
I agree since dragging to notepad doesn't work anywhere else in the program and no helpful source was found. so one step at time.

Krytan:
Deleted text to be dropped into text editor for example notepad because of an unexpected issue causing the texteditor  to open a new file instead of editing in the same file. moreover FileExplorer also doesn't allow Recent files to drop text into notepad or texteditor apps.

src/Files.App/UserControls/Widgets/RecentFilesWidget.xaml.cs
```diff
@@ -30,5 +32,42 @@ private void RecentFilesListView_RightTapped(object sender, RightTappedRoutedEve
...
+				foreach (var item in items)
+				{
+					try
+					{
+						// Attempt to get the file from its path 
+						var file = await StorageFile.GetFileFromPathAsync(item.RecentPath);
+						if (file != null)
+						{
+							storageItems.Add(file);
+						}
```
hishitetsu, line 51:
```suggestion
						if (file != null)
							storageItems.Add(file);
```

hishitetsu (commented):
I found a new problem. When dropping on the desktop I can only copy or link, but when dropping on the sidebar I can move the item.

Krytan:
@hishitetsu I see it is because the SideBar also have a drag code that changes the operation to move hmmm to fix this a solution would be in the sidebar.cs add a if estatment that if the file is coming from recent changes the operation to copy

hishitetsu:
> @hishitetsu I see it is because the SideBar also have a drag code that changes the operation to move hmmm to fix this a solution would be in the sidebar.cs add a if estatment that if the file is coming from recent, change the operation to copy

I agree.

hishitetsu:
This issue hasn't been fixed. https://github.com/files-community/Files/pull/15555#pullrequestreview-2129136824

yaira2:
> I found a new problem. When dropping on the desktop I can only copy or link, but when dropping on the sidebar I can move the item.

@Krytan are you still planning to work on this?

PR #15560, opened 2024-06-06
Code Quality: Added ThemedIcon

mdtauk:
**Resolved / Related Issues**
This is an ongoing project to try to work on a new ThemedIcon control

- Closes #

1. Opened Files ...
2. ...


src/Files.App/App.xaml
```diff
@@ -37,7 +37,9 @@
 				<XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
 
 				<!--  Styles for the custom icons  -->
+				<ResourceDictionary Source="ms-appx:///UserControls/ThemedIcon/ThemedIcon.xaml" />
 				<ResourceDictionary Source="ms-appx:///Styles/PathIcons.xaml" />
+				<ResourceDictionary Source="ms-appx:///Styles/ThemedIconSet.xaml" />
```
yaira2, line 42: What does "Set" mean in this context?

src/Files.App/Data/Contracts/IAppThemeModeService.cs
```diff
@@ -14,6 +14,11 @@ public interface IAppThemeModeService
 		/// </summary>
 		public event EventHandler? AppThemeModeChanged;
 
+		/// <summary>
+		/// Gets invoked when whether High Contrast mode is enabled.
```
yaira2, line 18:
```suggestion
		/// Invoked when High Contrast mode is turned on or off.
```

src/Files.App/App.xaml
0x5bfa: I think Styles/ThemedIcon should make more sense.
'Set' means 'a set of ThemedIcon styles' I assume.

yaira2: Are you okay with `ThemedIcons`?

src/Files.App/Styles/ThemedIconSet.xaml
```diff
@@ -0,0 +1,4568 @@
...
+							<local:ThemedIconLayer LayerType="Alt" PathData="" />
+							<local:ThemedIconLayer LayerType="Base" PathData="" />
+							<local:ThemedIconLayer LayerType="Accent" PathData="" />
+							<local:ThemedIconLayer LayerType="AccentContrast" PathData="" />
+						</local:ThemedIconLayers>
+					</Setter.Value>
+				</Setter>
+			</Style>
+
+			<!--#region  Common Icons-->
```
yaira2, line 36: In general we stay away from regions but I agree that it might be useful here.

0x5bfa: ThemedIcons.Common.xaml makes you guys hard to follow?

src/Files.App/App.xaml
mdtauk: ThemedIcons works, as long as it doesn't cause any confusion with ThemedIcon.xaml which is the control's template

src/Files.App/Styles/ThemedIconSet.xaml
```diff
@@ -0,0 +1,4568 @@
...
+			<Style
+				x:Name="TIcon_NavForward"
```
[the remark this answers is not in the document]
mdtauk: I would suggest it be ThemedIcon**s**.NavForward so as not to be confused by any of the ThemedIcon properties

src/Files.App/Styles/ThemedIconSet.xaml
mdtauk: You would end up with several Xaml files containing icon styles, which may or may not add confusion with autocomplete dot separated style names.

src/Files.App/App.xaml
0x5bfa: We only use this in App.xaml.

src/Files.App/Styles/ThemedIconSet.xaml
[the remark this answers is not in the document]
0x5bfa: `App.Icons.NavForward` to keep consistency with `App.Themes.*`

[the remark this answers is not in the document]
yaira2: `ThemedIcons.NavForward` & `App.Icons.NavForward` are fine with me.

src/Files.App/Styles/ThemedIconSet.xaml
0x5bfa: I dont think so?

yaira2: I have to agree with Martin here

src/Files.App/Styles/ThemedIconSet.xaml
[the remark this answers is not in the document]
mdtauk: App.ThemedIcons.NavForward is a good way to do it.

It's clear which type of icon control is used, and the App, follows other conventions.

src/Files.App/Styles/ThemedIconSet.xaml
mdtauk: The naming of the icons, as well as the regions for logical separation, is enough to keep on top of the amount of xaml there is, and also adding new icons, only means finding the right region to add to it.  And if the icon xaml is put in the "wrong" xaml region, its in the right file, and the naming will still work as expected.

src/Files.App/App.xaml
mdtauk: > We only use this in App.xaml. If we want consistency with other controls such as `SidebarStyles` and `GridSplitterStyles`, we might want to rename it to `ThemedIconStyles.xaml`.

Styles are technically what they are, but in terms of UI logic, they are different ThemedIcons

src/Files.App/Styles/ThemedIconSet.xaml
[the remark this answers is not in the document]
0x5bfa: You mean, like PathIcons? Shouldn't they have all different names even among other types?

src/Files.App/App.xaml
0x5bfa: file name isn't matter when it comes to developing once decided thus I'd like to keep consisntency here.

0x5bfa: File name doesn't matter when it comes to developing once decided thus I'd like to keep consisntency here.

src/Files.App/Styles/ThemedIconSet.xaml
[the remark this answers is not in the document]
mdtauk: There may be a mix of Icons as you start to replace older ones with ThemedIcon versions.

PathIcons, OpacityIcons, and there may be FontIcons and SymbolIcons there alongside ThemeIcons.

src/Files.App/App.xaml
mdtauk: > File name doesn't matter when it comes to developing once decided thus I'd like to keep consisntency here.

Sure, but as long as people know where to go when adding icons in the future.

src/Files.App/Styles/ThemedIconSet.xaml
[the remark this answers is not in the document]
0x5bfa: ThemedIcons can be `App.Icons.*` since others already have other names.

image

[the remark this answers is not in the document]
mdtauk: I would still opt for App.ThemedIcons.LayoutTiles

and we only append a size when there are larger icons than the 16x16 ones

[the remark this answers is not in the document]
yaira2: @0x5bfa I pushed a commit. Make sure to pull the commit before making any other changes.

src/Files.App/App.xaml
yaira2: I renamed to `ThemedIconStyles.xaml`.

src/Files.App/Data/Enums/ThemedIconStyles.cs
yaira2, line 1: @0x5bfa we're doing the final review on this PR. Can you please remove the enum to reduce confusion?

src/Files.App/Services/App/AppThemeModeService.cs
```diff
@@ -61,6 +67,10 @@ public AppThemeModeService()
 
 			// Registering to color changes, so that we can notice when changed system theme mode
 			UISettings.ColorValuesChanged += UISettings_ColorValuesChanged;
+
+			var windowId = MainWindow.Instance.AppWindow.Id;
+			ThemeSettings = ThemeSettings.CreateForWindowId(windowId);
+			ThemeSettings.Changed += (s, e) => { IsHighContrastChanged?.Invoke(this, s.HighContrast); };
```
yaira2, line 73: Should this be split onto separate lines? I'm fine with either way but I wanted to point this out just in case.

src/Files.App/Styles/MenuFlyoutItemStyles.xaml
yaira2, line 1: I understand this was just added for testing, let's keep a copy but save it for the PR where we actually remove OpacityIcon and replace all instances with ThemedIcon.

src/Files.App/UserControls/AddressToolbar.xaml
```diff
@@ -552,6 +552,16 @@
...
+				<!--  Show The Files App Settings Dialog  -->
+				<Button
+					x:Name="OpenUITestsWindowButton"
+					AutomationProperties.AutomationId="OpenUITestsWindowButton"
+					AutomationProperties.Name="Open UITests"
+					Click="OpenUITestsWindowButton_Click"
+					Style="{StaticResource AddressToolbarButtonStyle}">
+					<uc:ThemedIcon />
+				</Button>
+
```
yaira2, line 564:
```suggestion
```
If it helps, we can add a keyboard shortcut for displaying the test window.

src/Files.App/UserControls/AddressToolbar.xaml.cs
```diff
@@ -150,5 +150,11 @@ private void Button_AccessKeyInvoked(UIElement sender, AccessKeyInvokedEventArgs
 			if (VisualTreeHelper.GetOpenPopupsForXamlRoot(MainWindow.Instance.Content.XamlRoot).Any())
 				args.Handled = true;
 		}
+
+		private void OpenUITestsWindowButton_Click(object sender, RoutedEventArgs e)
+		{
+			var window = new UITestsWindow();
+			window.Activate();
+		}
```
yaira2, line 158:
```suggestion
```

src/Files.App/UserControls/Menus/MenuFlyoutItemWithImage.xaml.cs
```diff
@@ -5,10 +5,9 @@
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media.Imaging;
 
-// The User Control element template is documented at https://go.microsoft.com/fwlink/?LinkId=234236
-
 namespace Files.App.UserControls
 {
+	[Obsolete("Use MenuFlyoutItemEx")]
```
yaira2, line 10:
```suggestion
```

src/Files.App/UserControls/Menus/MenuFlyoutItemWithOpacityIcon.xaml.cs
```diff
@@ -4,10 +4,9 @@
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 
-// The User Control element template is documented at https://go.microsoft.com/fwlink/?LinkId=234236
-
 namespace Files.App.UserControls
 {
+	[Obsolete("Use MenuFlyoutItemEx")]
```
yaira2, line 9:
```suggestion
```

src/Files.App/Views/Settings/AboutPage.xaml
```diff
@@ -9,6 +9,7 @@
 	xmlns:helpers="using:Files.App.Helpers"
 	xmlns:local="using:Files.App.UserControls.Settings"
 	xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+	xmlns:uc="using:Files.App.UserControls"
```
yaira2, line 12:
```suggestion
```

src/Files.App/Styles/ThemedIconStyles.xaml
```diff
@@ -0,0 +1,4568 @@
+<!--  Copyright (c) 2024 Files Community. Licensed under the MIT License. See the LICENSE.  -->
```
marcelwgn, line 1: From a maintenance perspective, a XAML file with 4500 lines is not ideal 😅

src/Files.App/UserControls/ThemedIcon/ThemedIcon.xaml
```diff
@@ -0,0 +1,343 @@
...
+								<VisualState x:Name="Toggle">
+									<VisualState.Setters>
+										<Setter Target="PART_OutlinePath.Fill" Value="{ThemeResource ThemedIconAccentContrastBrush}" />
+										<Setter Target="PART_FilledPath.Fill" Value="{ThemeResource ThemedIconAccentContrastBrush}" />
+									</VisualState.Setters>
+								</VisualState>
+								<VisualState x:Name="HighContrast">
+									<VisualState.Setters>
+										<Setter Target="PART_OutlinePath.Fill" Value="{ThemeResource ThemedIconHighContrastBrush}" />
+										<Setter Target="PART_FilledPath.Fill" Value="{ThemeResource ThemedIconHighContrastBrush}" />
```
marcelwgn, line 222: Why are we doing this manually instead of all brushes just pointing to these two values in high contrast? That would allow us to not having to listening to the high contrast event.

mdtauk: We have the option to always use High Contrast icons, for those users who don't like the use of the accent colour.

src/Files.App/Styles/ThemedIconStyles.xaml
mdtauk: Splitting it can be done, but adding new icons would involve finding the right place to add it to.

We could use a merged dictionary for this.

yaira2: I would prefer a single file, but I'll follow your lead on this if you suggest a different approach.

src/Files.App/UserControls/ThemedIcon/ThemedIcon.xaml
yaira2: @mdtauk Is this a use case we plan to support? 
@marcelwgn does this affect performance or is there no harm in keeping it just in case?

mdtauk: > @mdtauk Is this a use case we plan to support? 
I don't know, you will have to decide on that

> @marcelwgn does this affect performance or is there no harm in keeping it just in case?
It is just a property ``UseContrast`` which works alongside the internal ``_isHighContrast`` bool which is set by the Theme Changed service

0x5bfa: Those brushes are set when user set Windows high contrast on AND/OR when user make a preference on using high contrast icons (if we decide to support this). If we drop the second one’s support, we should define high contrast brush only in the HighContrast XAML key.

mdtauk: I don't know of another way to handle the choice of high contrast, as well as light and dark normal support, without having a high contrast brush and state.

If we drop support for the `UseContrast` property, and only support it with the system theme - then we could then remove the extra brushes and visual states

yaira2: I don't foresee this being an option, but if keeping the code doesn't affect performance or make it harder to maintain, it can't hurt to keep it around.

src/Files.App/Styles/ThemedIconStyles.xaml
mdtauk: > I would prefer a single file, but I'll follow your lead on this if you suggest a different approach.

I am happy to split it up, if there is a need for smaller Xaml files - but I would like you to decide on how you would like to group them.  My Xaml regions are just how I see them being grouped, but as I wont always be the one adding new icons in the future, the grouping  should reflect your sorting @yaira2

src/Files.App/UserControls/ThemedIcon/ThemedIcon.xaml
mdtauk: I could differentiate between `UseContrast` and `_isHighContrast` so the HighContrast resources kick in, and the specific High Contrast brush we use is reserved only for the optional use.

marcelwgn: My 0.02€: If we currently don't support the option to toggle high contrast on for Files regardless of the system settings, I would not go the length to support it now. I agree with @mdtauk that it's not much more to support, however event listeners aren't free either. And the ThemeIcon control might get rendered dozens of times so small costs might add up.

src/Files.App/Styles/ThemedIconStyles.xaml
marcelwgn: Would it make sense to split out the icons into a separate file and have the base template be its own file? That ways it's easier to spot where to find icons or the actual ThemeIcon control's code.

src/Files.App/UserControls/ThemedIcon/ThemedIcon.xaml
mdtauk: They way I see it, HighContrast based on the system rather than the control's "UseContrast" property.  Still requires an event listener, so we can replace the Layered visual state with the outline.

The UseContrast setting basically also replaces the layered icon state with the Outlined icon - however we also have a separate brush resource, which makes the icons use the same brush as the text around it, ensuring it stands out clearer. (our base brush is closer to the TextSecondary brush so they don't feel too intrusive.)

Even if we were to remove the alternate brush choice, and the UseContrast property - and decide never to support it, or custom icon colours for the custom themes in Files - the HighContrast system event is still essential.

mdtauk: I am happy to remove the UseContrast additional setting, if no purpose is seen in it.

I know some people use a grey Accent colour or may want to change the icon colour via a custom theme - but if there is no plans to implement that, then so be it.

src/Files.App/Styles/ThemedIconStyles.xaml
mdtauk: I can split the icons into separate Xaml, the base default control template is already separate.

src/Files.App/UserControls/ThemedIcon/ThemedIcon.xaml
mdtauk: > Those brushes are set when user set Windows high contrast on AND/OR when user make a preference on using high contrast icons (if we decide to support this). If we drop the second one’s support, we should define high contrast brush only in the HighContrast XAML key.

That is the case, in the other repo, I have removed the UseContrast property and states, as well as removed the brushes they used to use.

yaira2: Olay, let's remove and if the need arises, we can refer to this PR and add it back.

src/Files.App/Data/Enums/ThemedIconStyles.cs
0x5bfa: For the future reference, we’ll make a test project and it’ll need this but we don’t want to make things complicated thus we might be going to create a source generator for it.

src/Files.App/Services/App/AppThemeModeService.cs
0x5bfa: I’m fine either but I prefer this.

src/Files.App/Styles/MenuFlyoutItemStyles.xaml
0x5bfa: Sure.

0x5bfa:
@mdtauk has a separate repo for us to be able to test the control. But the problem is WinUI test project cannot reference WinUI project (#15762) (we definitely don’t want to include test features inside Files.App project). Therefore, we should wrap up everything there and make a single commit and then merge asap. I’ll try to figure out how to reference the project from an unit test project but this shouldn’t be a blocker against merging.

0x5bfa:
@marcelwgn as you suggested we decided to create a separate class library for controls (Files.App.Controls) and an unit test project (Files.App.UITests) for that project. To ease PR’s review work, we’ve splitter up work into three: adding template project for both, adding ThemedIcon implementation and its test and lastly replacing the old icons with this icon control. Thank you for reviewing!

PR #15579, opened 2024-06-10
Code Quality: Reverting changes for DateTime formatting without ICU

### LINKED ISSUE #15554: Bug: Files slows to a crawl if opening a folder with lots of files (preview build)
### Description

Effectively if I open a folder that has lots of files with varying sizes (between 1kb-9GB), and the folder contains around 2000+ Files will just slow to a crawl until I close the tab with said folder.

### Steps To Reproduce

1. Open files
2. Navigate to the folder with 2000+ items
3. It'll slow to a crawl after a few seconds

### Requirements

To open the folder with minimum delay between reading/loading all the files in a folder with large amount of files.

### Files Version

3.4.5

### Windows Version

10.0.19045.4170

### Log File

[debug.log](https://github.com/user-attachments/files/15571927/debug.log)

XTorLukas:
**Resolved / Related Issues**

- Closes #15554

1. Try loading bigger folder e.g. `C:/Windows/System32`
2. It is now faster to load strings with formatting using `GetLocalizedFormatResource`

### DIFF
--- a/src/Files.App/Services/DateTimeFormatter/ApplicationDateTimeFormatter.cs
+++ b/src/Files.App/Services/DateTimeFormatter/ApplicationDateTimeFormatter.cs
@@ -1,48 +1,80 @@
 // Copyright (c) 2024 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using Files.App.Extensions;
-using System;
+using System.Text;
 
 namespace Files.App.Services.DateTimeFormatter
 {
+	/// <summary>
+	/// Application-specific implementation of the DateTimeFormatter.
+	/// </summary>
 	internal sealed class ApplicationDateTimeFormatter : AbstractDateTimeFormatter
 	{
+		/// <summary>
+		/// Gets the name of the formatter.
+		/// </summary>
 		public override string Name
 			=> "Application".GetLocalizedResource();
 
+		/// <summary>
+		/// Converts the provided <see cref="DateTimeOffset"/> to a short label.
+		/// </summary>
+		/// <param name="offset">The <see cref="DateTimeOffset"/> to convert.</param>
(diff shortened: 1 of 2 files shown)


XTorLukas:
@yaira2 fyi

src/Files.App/Services/DateTimeFormatter/ApplicationDateTimeFormatter.cs
```diff
@@ -1,48 +1,80 @@
...
 // Licensed under the MIT License. See the LICENSE.
 
-using Files.App.Extensions;
-using System;
+using System.Text;
 
 namespace Files.App.Services.DateTimeFormatter
 {
+	/// <summary>
+	/// Application-specific implementation of the DateTimeFormatter.
```
0x5bfa, line 9:
```suggestion
	/// Represents service for application-specific <see cref=“DateTimeOffset”/> formatter.
```

0x5bfa (approved):
Those are used in bulk enumeration so speed is important imo.
Do you think having our own message formatter *would* improve performance?

Codewise, looks good to me.

XTorLukas:
> Those are used in bulk enumeration so speed is important imo. Do you think having our own message formatter _would_ improve performance?

According to the benchmarks I am doing while developing my own UnicodeFormatter, the results are promising so far, but I am still working on optimisation and development, here I am already running into the limits of C# and it is necessary to redirect some operations to the C++ part or use pointers in the unsafe part of the code. I will continue the development. For now, however, this MessageFormat cannot be used with a large number of elements.

yaira2 (approved):
Hopefully we can figure this out so that we can bring back the code improvements, but the main thing is that performance is back to how it was before.

PR #15582, opened 2024-06-10
Fix: Fixed ArgumentException at CommandManager.OverwriteKeyBindings

### LINKED ISSUE #15331: Bug: CommandManager.OverwriteKeyBindings System.ArgumentException
### Description

```csharp
System.ArgumentException: An element with the same key but a different value already exists. Key: 'Alt+Ctrl+Haut'
```

### Steps To Reproduce

NA

### Requirements

- Fix crash

### Files Version

v3.4

### Windows Version

Windows 11

### Log File

```csharp
System.Collections.Immutable.ImmutableDictionary`2
HashBucket.Add (TKey key, TValue value, IEqualityComparer`1 keyOnlyComparer, IEqualityComparer`1 valueComparer, KeyCollisionBehavior behavior, OperationResult& result)
System.Collections.Immutable
ImmutableDictionary`2.AddRange (IEnumerable`1 items, MutationInput origin, KeyCollisionBehavior collisionBehavior)
System.Collections.Immutable
ImmutableDictionary`2.AddRange (IEnumerable`1 pairs, Boolean avoidToHashMap)
Files.App.Data.Commands
CommandManager.OverwriteKeyBindings ()
Files.App.Utils.Serialization
BaseObservableJsonSettings.Set[TValue] (TValue value, String propertyName)
Files.App.Services.Settings
ActionsSettingsService.set_ActionsV2 (List`1 value)
Files.App.ViewModels.Settings
ActionsViewModel.ExecuteSaveCommand (ModifiableActionItem item)
CommunityToolkit.Mvvm.Input
RelayCommand`1.Execute (Object parameter)
ABI.System.Windows.Input.ICommand
Vftbl.Do_Abi_Execute_3 (IntPtr thisPtr, IntPtr parameter)
WinRT
Microsoft.AppCenter.Utils
```

0x5bfa:
### Summary

The fix is basically to catch exceptions and use default keys for commands that have the same key(if other exception than ArgumentException the app uses the default keys for all commands).

### Resolved / Related Issues

- Closes #15331

1. Set the same key *1
2. Open Files app
3. See no crashes

Used this to test
```
  "ActionsV2": [
    {
      "CommandCode": "OpenHelp",
      "CommandParameter": "",
      "KeyBinding": "F1"
    },
    {
      "CommandCode": "OpenHelp",
      "CommandParameter": "",
      "KeyBinding": "Ctrl+C"
    }
  ],
```


src/Files.App/Data/Commands/Manager/CommandManager.cs
```diff
@@ -420,9 +423,51 @@ private void OverwriteKeyBindings()
...
+				_allKeyBindings = commands.Values
+					.SelectMany(command => command.HotKeys, (command, hotKey) => (Command: command, HotKey: hotKey))
+					.ToImmutableDictionary(item => item.HotKey, item => item.Command);
+
+				App.Logger.LogError(ex, "The app is temporarily using default key for some commands bindings because of a serious error of assigning custom keys.");
+			}
+			catch (Exception ex)
+			{
+				foreach (var command in commands.Values.OfType<ActionCommand>())
+					command.RestoreKeyBindings();
```
hishitetsu, line 463: This can also be written in LINQ. It may be a matter of preference though.

src/Files.App/Data/Commands/Manager/CommandManager.cs
```diff
@@ -420,9 +423,51 @@ private void OverwriteKeyBindings()
...
+							occurrences.ElementAt(index).RestoreKeyBindings();
+						}
+					}
+				}
+
+				_allKeyBindings = commands.Values
+					.SelectMany(command => command.HotKeys, (command, hotKey) => (Command: command, HotKey: hotKey))
+					.ToImmutableDictionary(item => item.HotKey, item => item.Command);
+
+				App.Logger.LogError(ex, "The app is temporarily using default key for some commands bindings because of a serious error of assigning custom keys.");
```
hishitetsu, line 458: * Couldn't you write that the reason is that there are duplicate key assigns?
* I think the log level for exceptions with known causes should be Warning, not Error.

src/Files.App/Data/Commands/Manager/CommandManager.cs
```diff
@@ -420,9 +423,51 @@ private void OverwriteKeyBindings()
...
+				{
+					if (!string.IsNullOrEmpty(item))
+					{
+						var occurrences = allCommands.Where(x => x.HotKeys.Select(x => x.LocalizedLabel).Contains(item));
+
+						// Restore the defaults for all occurrences
+						for (int index = 0; index < occurrences.Count(); index++)
+						{
+							occurrences.ElementAt(index).RestoreKeyBindings();
+						}
```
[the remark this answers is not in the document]
0x5bfa: Done

src/Files.App/Data/Commands/Manager/CommandManager.cs
0x5bfa: Done

src/Files.App/Data/Commands/Manager/CommandManager.cs
0x5bfa: Done

hishitetsu (commented):
Crashes can be avoided, but changing duplicate key assignments from the settings doesn't seem to be saved.

0x5bfa:
I want to reset them in json too then
@yaira2 fyi

hishitetsu:
This can only happen by editing the configuration file directly, so I think it is fine to leave it as it is if avoiding crashes is enough.

0x5bfa:
I mean, instead of using default ones, I'd like to reset in json first and reload. What do you think? Losing user data without notice would be bad but this reload happens before MainPage loaded, meaning dialog cannot be shown.

hishitetsu:
> I mean, instead of using default ones, I'd like to reset in json first and reload. What do you think? Losing user data without notice would be bad but this reload happens before MainPage loaded, meaning dialog cannot be shown.

I don't think it is a good idea, unless structure of the json is broken.

0x5bfa:
Okay, I’ll just fix

hishitetsu:
It seems that the app no longer starts when there is a duplicate key.

0x5bfa:
Proven I’m bad at bug fix.
Let me take a look tonight.

hishitetsu (approved):
I confirmed it works. LGTM

yaira2:
What solution did we end up going with?

hishitetsu:
If there are any custom hotkeys that duplicate other hotkeys, they are removed from the json and restored to default.

yaira2:
Thank you @0x5bfa and @hishitetsu for your work on this fix!

PR #15584, opened 2024-06-10
Code Quality: Assign background/border properties in respective controls

0x5bfa:
### Summary

This makes the app content area look more appropriate when the window a bit stuck or hang on startup because background and border will be rendered actually loaded. Previously Inner toolbar background are rendered without inner controls are loaded, before it actually loaded.

- Assign background/border properties in respective controls
- Unwrap from `Border` element
- Removed unnecessary assignments to `d:DesignHeight`, `d:DesignWidth`, `TabStripVisibility (always true)` and so on.

### Resolved / Related Issues

- Closes #

1. Open Files app
2. See no crash and visual change

### DIFF
--- a/src/Files.App/Views/MainPage.xaml
+++ b/src/Files.App/Views/MainPage.xaml
@@ -132,7 +132,8 @@
 			Source="{x:Bind ViewModel.AppThemeBackgroundImageSource, Mode=OneWay}"
 			Stretch="{x:Bind ViewModel.AppThemeBackgroundImageFit, Mode=OneWay}" />
 
-		<Grid x:Name="RightMarginGrid" VerticalAlignment="Top">
+		<!--  Header Area  -->
+		<Grid>
 			<Grid.RowDefinitions>
 				<RowDefinition Height="Auto" />
 				<RowDefinition Height="Auto" />
@@ -141,15 +142,10 @@
 			<!--  Tab Control  -->
 			<tabbar:TabBar
 				x:Name="TabControl"
-				HorizontalAlignment="Stretch"
-				VerticalAlignment="Stretch"
 				HorizontalContentAlignment="Stretch"
-				x:FieldModifier="public"
 				x:Load="False"
 				Loaded="HorizontalMultitaskingControl_Loaded"
-				TabIndex="0"
-				TabStripVisibility="Visible"
-				Visibility="Visible">
+				TabIndex="0">
 				<tabbar:TabBar.FooterElement>
 					<!--  Height is not divisble by four in order to properly align the button  -->
 					<Button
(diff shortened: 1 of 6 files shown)


0x5bfa:
I wonder if this wrapping by Grid is necessary and we should create one more row definition at the root Grid.
Removing visual tree element should help improve loading on startup.

image

src/Files.App/Views/MainPage.xaml
```diff
@@ -132,24 +132,20 @@
...
-		<Grid x:Name="RightMarginGrid" VerticalAlignment="Top">
+		<!--  Header Area  -->
+		<Grid>
 			<Grid.RowDefinitions>
 				<RowDefinition Height="Auto" />
 				<RowDefinition Height="Auto" />
 			</Grid.RowDefinitions>
 
-			<!--  Tab Control  -->
+			<!--  Tab Control (Lazy Load)  -->
```
yaira2, line 142: What do you mean by "Lazy Load"?

src/Files.App/Views/MainPage.xaml
```diff
@@ -280,32 +268,25 @@
 						ResizeBehavior="BasedOnAlignment"
 						Style="{StaticResource DefaultGridSplitterStyle}" />
 
-					<!--  Preview Pane  -->
-					<Border
-						x:Name="InfoPaneContainer"
+					<!--  Info Pane (Lazy Load)  -->
+					<uc:InfoPane
+						x:Name="InfoPane"
```
yaira2, line 273: Are there any visual states that need updating to reflect this change?

src/Files.App/Views/MainPage.xaml
0x5bfa: To boost startup loading, d2dyno get almost of controls in MainPage to be loaded after a bit passed. You can see x:Load is False by default. The code behind shows them by finding these Xaml element.

src/Files.App/Views/MainPage.xaml
0x5bfa: I don't think so.
I already updated in Main page.xaml.cs

src/Files.App/Views/MainPage.xaml
yaira2: I see

yaira2: I'm not sure how helpful the comment is in the current form, do you think it needs to be mentioned at all?

0x5bfa: For me, felt in that way because it is unloaded at first and load afterwards and thought some people would wonder why.
But I also get to see the future of many people getting wondered what the Lazy Load means.

I think it's clearer to write all this down into the code behind instead. I'll revert.

0x5bfa: Done

PR #15586, opened 2024-06-10
Fix: Fixed crash that could occur if the Libraries folder wasn't located

yaira2:
**Resolved / Related Issues**

- Closes https://files-org.sentry.io/issues/5472102996/events/dd9fe91b32134e1eaaae5fe6a1375e0d/

NA

### DIFF
--- a/src/Files.App/Data/Items/ShellLibraryItem.cs
+++ b/src/Files.App/Data/Items/ShellLibraryItem.cs
@@ -9,7 +9,7 @@ namespace Files.App.Data.Items
 	{
 		public const string EXTENSION = ".library-ms";
 
-		public static readonly string LibrariesPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Microsoft", "Windows", "Libraries");
+		public static readonly string LibrariesPath = Win32PInvoke.GetFolderFromKnownFolderGUID(new Guid("1B3EA5DC-B587-4786-B4EF-BD1DC332AEAE"));
 
 		/// <summary>
 		/// Full path of library file.<br/>
--- a/src/Files.App/Helpers/Win32/Win32PInvoke.Methods.cs
+++ b/src/Files.App/Helpers/Win32/Win32PInvoke.Methods.cs
@@ -502,5 +502,22 @@ namespace Files.App.Helpers
 			string pszRootPath,
 			ref SHQUERYRBINFO pSHQueryRBInfo
 		);
+
+		[DllImport("shell32.dll")]
+		static extern int SHGetKnownFolderPath(
+			[MarshalAs(UnmanagedType.LPStruct)] Guid rfid,
+			uint dwFlags,
+			IntPtr hToken,
+			out IntPtr pszPath
+		);
+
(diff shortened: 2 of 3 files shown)


hishitetsu (commented):
I have found that if I move the original library folder to a different location and create a new library, it will be created in the original library folder that was moved, not in the new library folder that is created by this PR. So, it would be better to try to find out where the library folder is located in the first place.

yaira2:
Is there a registry key or way to locate the path?

hishitetsu:
There is a registry key, but it says not to use this.
image

yaira2:
That help point me in the right direciton! I ended up using PInvoke.

src/Files.App/Data/Items/ShellLibraryItem.cs
```diff
@@ -9,7 +9,7 @@ public sealed class ShellLibraryItem
 	{
 		public const string EXTENSION = ".library-ms";
 
-		public static readonly string LibrariesPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Microsoft", "Windows", "Libraries");
+		public static readonly string LibrariesPath = Win32PInvoke.GetFolderFromKnownFolderGUID(new Guid("1B3EA5DC-B587-4786-B4EF-BD1DC332AEAE"));
```
0x5bfa, line 12: Please create a new class named ‘CLSID’ (I was going to do so).
It enables us to add also other than special folders such as Com Interfaces and more.

src/Files.App/Helpers/Win32/Win32PInvoke.Methods.cs
```diff
@@ -502,5 +502,22 @@ int dwFlags
 			string pszRootPath,
 			ref SHQUERYRBINFO pSHQueryRBInfo
 		);
+
+		[DllImport("shell32.dll")]
+		static extern int SHGetKnownFolderPath(
```
0x5bfa, line 507: Please use CsWin32…
If we in rush, no problem, I’ll do that instead.

0x5bfa (changes requested):

src/Files.App/Data/Items/ShellLibraryItem.cs
yaira2: Under Win32PInovke?

src/Files.App/Helpers/Win32/Win32PInvoke.Methods.cs
yaira2: Sounds good.

src/Files.App/Data/Items/ShellLibraryItem.cs
0x5bfa: Under Files.App.Constants

yaira2: I'll leave this for you in a follow up PR.

0x5bfa: Ok...

hishitetsu (approved):
Aside from the code, I have confirmed that this change will solve the issue.

PR #15600, opened 2024-06-16
Fix: Prevent issue "WinUI Desktop Window object has already been closed"

### LINKED ISSUE #15599: Bug: System.Runtime.InteropServices.COMException: The operation identifier is not valid.
Sentry Issue: [FILES-APP-E](https://files-org.sentry.io/issues/5456839101/?referrer=github_integration)

```
System.Runtime.InteropServices.COMException: The operation identifier is not valid.

The WinUI Desktop Window object has already been closed.
  ?, in async void InfoPaneViewModel.ContentPageContext_PropertyChanged(object sender, PropertyChangedEventArgs e)
```

yaira2:
**Resolved / Related Issues**

- Closes #15599

I'm not sure if this is the best fix, but it should at least prevent the crash.

### DIFF
--- a/src/Files.App/ViewModels/UserControls/InfoPaneViewModel.cs
+++ b/src/Files.App/ViewModels/UserControls/InfoPaneViewModel.cs
@@ -6,6 +6,7 @@ using Files.App.ViewModels.Previews;
 using Files.Shared.Helpers;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
+using Sentry;
 using Windows.Storage;
 
 namespace Files.App.ViewModels.UserControls
@@ -13,6 +14,7 @@ namespace Files.App.ViewModels.UserControls
 	public sealed class InfoPaneViewModel : ObservableObject, IDisposable
 	{
 		private IInfoPaneSettingsService infoPaneSettingsService { get; } = Ioc.Default.GetRequiredService<IInfoPaneSettingsService>();
+		private IGeneralSettingsService generalSettingsService { get; } = Ioc.Default.GetRequiredService<IGeneralSettingsService>();
 		private IContentPageContext contentPageContext { get; } = Ioc.Default.GetRequiredService<IContentPageContext>();
 
 		private CancellationTokenSource loadCancellationTokenSource;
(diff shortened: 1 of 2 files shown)


yaira2:
I'd prefer to come up with a better fix long term, but I think we should do this as a temporary solution.

src/Files.App/ViewModels/UserControls/InfoPaneViewModel.cs
```diff
@@ -136,12 +138,27 @@ private async void ContentPageContext_PropertyChanged(object? sender, PropertyCh
...
+					{
+						// Handle exception in case WinUI Windows is closed
+						// (see https://github.com/files-community/Files/issues/15599)
+
+						SentrySdk.CaptureException(ex, scope =>
+						{
+							scope.User.Id = generalSettingsService.UserId;
+							scope.Level = SentryLevel.Warning;
+						});
+					}
```
hishitetsu, line 160: Shouldn't it also log to a file?
I'd like to incorporate sending reports to Sentry into the logger soon.

yaira2: I suppose

src/Files.App/ViewModels/UserControls/InfoPaneViewModel.cs
```diff
@@ -136,12 +139,29 @@ private async void ContentPageContext_PropertyChanged(object? sender, PropertyCh
...
+						// Handle exception in case WinUI Windows is closed
+						// (see https://github.com/files-community/Files/issues/15599)
+
+						SentrySdk.CaptureException(ex, scope =>
+						{
+							scope.User.Id = generalSettingsService.UserId;
+							scope.Level = SentryLevel.Warning;
+						});
+
+						App.Logger.LogError(ex, ex.Message);
```
hishitetsu, line 162:
```suggestion
						App.Logger.LogWarning(ex, ex.Message);
```

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -442,14 +444,30 @@ private void PaneSplitter_ManipulationCompleted(object sender, ManipulationCompl
...
+				// Handle exception in case WinUI Windows is closed
+				// (see https://github.com/files-community/Files/issues/15599)
+
+				SentrySdk.CaptureException(ex, scope =>
+				{
+					scope.User.Id = generalSettingsService.UserId;
+					scope.Level = SentryLevel.Warning;
+				});
+
+				App.Logger.LogError(ex, ex.Message);
```
hishitetsu, line 469:
```suggestion
				App.Logger.LogWarning(ex, ex.Message);
```

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -442,14 +444,30 @@ private void PaneSplitter_ManipulationCompleted(object sender, ManipulationCompl
...
+			}
 
 			UpdatePositioning();
```
hishitetsu, line 472: Shouldn't this be included in `try`?

0x5bfa:
Feel free to add Sentry to GlobalUsings．

src/Files.App/Views/MainPage.xaml.cs
yaira2: It shouldn't cause a crash but I guess we should only update if the window is visible.

Issue #12285, opened 2023-05-05
Bug: Renaming a file in dual pane causes focus to keep changing


Androl404:
### Description

After entering panel view, and working between the files a bit, Files starts switching very fast between the two panels. You can see it in the right bar if you're working on two different disks.

### Steps To Reproduce

1. Open Files
2. Enter pannel mode
3. Start working
4. After about 2 minutes, the issue occurs

### Requirements

- Make Files to stay in the last used pannel

### Files Version

2.4.70.0

### Windows Version

Windows 10 Family 22H2 19045.2913 Windows Feature Experience Pack 1000.19041.1000.0

### Log file
[debug.log](https://github.com/files-community/Files/files/11410170/debug.log)

Josh65-2201:
Can you please upload (Do not copy and paste content) the debug.log file. You can get it in Files settings > About > Open log folder.

Androl404:
It says it's too long because its 10 707 lines

Josh65-2201:
It should work if you upload the file. I assume you trying to copy and paste the content of it into the box.

Androl404:
Sorry, I've modified the first message, it should contain the log (the whole file)

Josh65-2201:
Thanks

yaira2:
I was able to reproduce this issue

ferrariofilippo:
I was able to reproduce this in 2.4.70 but I can't in 2.4.81

yaira2:
I don't think we changed anything that would have fixed the issue but it's hard to test since it's on and off.

yaira2:
@Androl404 this issue will be resolved in v2.5.

Androl404:
> @Androl404 this issue will be resolved in v2.5.

Sounds great ! Thank you very much !

PR #15614, opened 2024-06-17
Feature: Added support for closing the left pane

### LINKED ISSUE #3342: Feature: Add the ability to close the primary/left panel
**Requirements**
- Add support for closing the left pane

0x5bfa:
### Resolved / Related Issues

*Internally planned.*

For the record, this implementation avoids major refactoring of removing left/right pane to be able to create panes more than two panes.

Closes #3342

1. Open Files app
2. Open right pane
3. Focus left pane
4. Close active pane from the menu


src/Files.App/Views/ShellPanesPage.xaml.cs
```diff
@@ -176,23 +176,23 @@ public bool IsRightPaneVisible
 				if (value != _IsRightPaneVisible)
 				{
 					_IsRightPaneVisible = value;
```
yaira2, line 178: I suppose we still want to remove this property in a future PR.

0x5bfa: Definitely yes. But we have to do it slowly not in the same PR. I realized that in my previous trial 😅

yaira2:
More testing is needed but I haven't found any issues on my side.

0x5bfa:
It's weird that when I tested on local it worked...?

Now works

yaira2:
I'm still having this issue

0x5bfa:
🥹

yaira2:
The previously mentioned issue has been resolved. But don't worry, I found another one 😂 

1. Open two panes and scroll down in the right pane
2. Closing the left pane scrolls to the top of the list

0x5bfa:
Single underscore 😭😭😭😭

yaira2:
> Single underscore 😭😭😭😭

Is this related to the issue I mentioned?

0x5bfa:
Yes, fixed

yaira2:
`AAAAAAAAAAAAAAA` is the right name 😂 

I just wanted to confirm.

yaira2 (approved):
LGTM

yaira2:
I found a bug,
1. Open two panes
2. Resize the window to be small enough that the right pane hides
3. Make the window larger again, and notice that the right pane navigated to the home page

hishitetsu:
> I found a bug,
> 
> 1. Open two panes
> 2. Resize the window to be small enough that the right pane hides
> 3. Make the window larger again, and notice that the right pane navigated to the home page

This is an existing issue, so it doesn't have to be fixed in this PR.

src/Files.App/Actions/Navigation/CloseActivePaneAction.cs
```diff
@@ -20,23 +20,20 @@ public RichGlyph Glyph
...
-
-			context.PropertyChanged += Context_PropertyChanged;
+			ContentPageContext.PropertyChanged += ContentPageContext_PropertyChanged;
 		}
 
 		public Task ExecuteAsync(object? parameter = null)
 		{
-			context.ShellPage!.PaneHolder.CloseSecondaryPane();
-
+			ContentPageContext.ShellPage!.PaneHolder.CloseActivePane();
```
hishitetsu, line 32:
```suggestion
			ContentPageContext.ShellPage?.PaneHolder.CloseActivePane();
```
It should be null safe to avoid crashes, even if IsExecutable has a null check.

src/Files.App/Views/ShellPanesPage.xaml.cs
```diff
@@ -343,14 +344,35 @@ private void RemovePane(int index = -1)
 			// Get proper position of sizer that resides with the pane that is wanted to be removed
 			var childIndex = index * 2 - 1;
 			childIndex = childIndex >= 0 ? childIndex : 0;
+			if (childIndex + 1 > RootGrid.Children.Count)
+				return;
+
+			if (childIndex is 0)
+			{
```
hishitetsu, line 351:
```suggestion
			if (childIndex >= RootGrid.Children.Count)
				return;

			if (childIndex == 0)
			{
```
It may be a matter of taste, but this is easier for me to understand.

src/Files.App/Views/ShellPanesPage.xaml.cs
```diff
@@ -491,7 +513,7 @@ private void Pane_GotFocus(object sender, RoutedEventArgs e)
 
 		private void Pane_RightTapped(object sender, RoutedEventArgs e)
 		{
-			if (sender != ActivePane && sender is IShellPage shellPage && shellPage.SlimContentPage is not ColumnsLayoutPage)
+			if (sender != ActivePane)
```
hishitetsu, line 516: Why did you delete these conditions? It doesn't appear to be directly related to this PR.

yaira2: Are they needed?

hishitetsu: I think they are for focusing not a pane but a column when in the column layout.

0x5bfa: When it's column view, even if you focus background wouldn't changed. You can try in preview build.

0x5bfa:
> > I found a bug,
> > 
> > 1. Open two panes
> > 2. Resize the window to be small enough that the right pane hides
> > 3. Make the window larger again, and notice that the right pane navigated to the home page
> 
> This is an existing issue, so it doesn't have to be fixed in this PR.

My guess is here:

https://github.com/files-community/Files/blob/bfba7d4833a59d32814fa8a6e9b115548189b68b/src/Files.App/Views/ShellPanesPage.xaml.cs#L527-L533

src/Files.App/Views/ShellPanesPage.xaml.cs
hishitetsu: That is a different issue, and this change causes the focus to flicker when right-clicking a column. I think this change should be reverted.
We have had similar issues before. https://github.com/files-community/Files/issues/12285

hishitetsu (approved):
LGTM

PR #15628, opened 2024-06-19
Fix: Fixed an issue where arguments weren’t passed when running shortcuts as admin

### LINKED ISSUE #15595: Bug: Shortcut set to run as Admin doesn't work if there are arguments
### Description

I have a shortcut in my Desktop to connect to a local VM (using Windows vmconnect). It is set to run as administrator in the file properties and it has arguments to connect to the VM.
If I simply double click the shortcut, I get the elevation prompt, but vmconnect opens as if there were no arguments.
If I right click -> Run as administrator, it actually runs as it's supposed to be.

### Steps To Reproduce

1. Create a text file on your desktop called new1.txt
2. Create a shortcut to notepad.exe and pass the new1.txt path as argument
3. Set the shortcut to run as admin
4. If you double click the shortcut, you get the elevation prompt, but Notepad opens blank
5. If you right click -> run as administrator, it actually opens the file correctly 

image

### Requirements

Shortcuts should work as in default Windows Explorer, so if there's a shortcut to a program and it passes arguments, it should pass those arguments to the programs when you start it.

### Files Version

3.4.1

### Windows Version

23H2 Build 22631.3737

### Log File

[debug.log](https://github.com/user-attachments/files/15844592/debug.log)

yaira2:
**Resolved / Related Issues**

- Closes #15595

1. Followed the repro steps outlined in issue #15595.
2. Verified that arguments are correctly passed in both scenarios: when the shortcut is configured to run as an administrator and when it is not.


yaira2:
In the long run, we should modify `HandleApplicationLaunch` to handle this correctly, but I think this is a good workaround to avoid a refactor.

0x5bfa:
> In the long run, we should modify `HandleApplicationLaunch` to handle this correctly. It the meantime, I think this is a good workaround to avoid a refactor.

Could you add a NOTE comment saying this, so I (or someone) can notice?

gave92:
I'd probably go with the refactor route, it's not that bad, see here:
2f835a619cb2491a86056b73a1f9fc62282c9fb8

If you prefer the current solution, make sure to remove "process.WaitForExit();" and to wrap in a try-catch in case user says no to UAC, like:
cfb1af004addfd82a5677655483daaaee6c3b7a7

yaira2:
> I'd probably go with the refactor route, it's not that bad, see here: https://github.com/gave92/files-uwp/tree/rev_adminargs
> 
> If you prefer the current solution, make sure to remove "process.WaitForExit();" and to wrap in a try-catch in case user says no to UAC, like: [cfb1af0](https://github.com/files-community/Files/commit/cfb1af004addfd82a5677655483daaaee6c3b7a7)

Your way is fine with me, can you open a PR?

gave92 (approved):

PR #11433, opened 2023-02-23
Fix: Fixed issue where the properties window opened from widgets was hidden behind

hishitetsu:
**Resolved / Related Issues**
Items resolved / related issues by this PR.
- Closes #11432 

**Validation**
How did you test these changes?
- [X] Built and ran the app
- [X] Tested the changes for accessibility

PR #15636, opened 2024-06-20
Code Quality: Bring properties window to foreground on loaded

### LINKED ISSUE #13697: When opening empty folder in Files, property window for a drive (Right-click>Properties) moves to background & stays there
### Description

When using the current Store version (3.0.1.0), when you open a folder in Files that is empty, then right-clicking a drive and opening the drive properties puts the info window into the background, so the main Files window always stays on top. Even clicking the info window lets it only flash up for a moment, but doesn't bring it to the foreground. 

### Steps To Reproduce

1. Open Files.
2. Go to a folder on your system that has no files or sub-folders inside. Removing files doesn't work, it has to be empty when you enter the folder. 
3. Right-click a drive in the sidebar. 
4. Click 'Properties' ("Eigenschaften" in German).
5. The popup window opens, but immediately moves behind the main Files window.
6. Clicking the info window in the background doesn't move it to the foreground.

### Requirements

- Let Files open the Properties window in the foreground and don't steal focus off it. 

### Files Version

3.0.1.0

### Windows Version

23H2 (22635.2700)

### Log File

Not needed

0x5bfa:
### Resolved / Related Issues

- Closes #13697 

1. Open Files app
2. Open file properties window
3. See brought to foreground

### DIFF
--- a/src/Files.App/Data/Contexts/SideBar/ISideBarContext.cs
+++ b/src/Files.App/Data/Contexts/SideBar/ISideBarContext.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) 2024 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
+using Microsoft.UI.Xaml.Controls;
+
 namespace Files.App.Data.Contexts
 {
 	/// <summary>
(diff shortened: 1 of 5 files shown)


0x5bfa:
@yaira2 Found easy one 😄

yaira2:
Did this fix the issue for you? On my end it's still an issue.

- Open an empty folder
- Right click a drive from the sidebar and open properties
- The main window takes the focus

yaira2:
We had a similar issue with widgets (see #11433).

0x5bfa:
When I first try to repro, I removed existing files and I thought it fixed (it sure poped up to the foreground).
@yaira2 I did the same thing but no progress here. I seemed to have been wrong...

PR #15647, opened 2024-06-23
Fix: Fixed System.TypeInitializationException in FileTagsDatabase

### LINKED ISSUE #15601: Bug: System.TypeInitializationException: The type initializer for 'Files.App.Utils.FileTags.FileTagsDatabase' threw an exception.
Sentry Issue: [FILES-APP-2B](https://files-org.sentry.io/issues/5478011341/?referrer=github_integration)

```
System.Runtime.InteropServices.COMException: None
  ?, in Package IPackageStaticsMethods.get_Current(IObjectReference _obj)
  ?, in static FileTagsDatabase()

System.TypeInitializationException: The type initializer for 'Files.App.Utils.FileTags.FileTagsDatabase' threw an exception.
  ?, in static FileTagsHelper()+() => { }
  ?, in void ItemViewModel.SetFileTag(ListedItem item)
  ?, in async Task ItemViewModel.LoadExtendedItemPropertiesAsync(ListedItem item)+(?) => { }
  ?, in async Task<FilesystemResult> FilesystemTasks.Wrap(Func<Task> wrapped)
...
(1 additional frame(s) were not displayed)

System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread.
```

hishitetsu:
This prevents crashes due to the exception, but the tags feature doesn't work properly in that case.
It may be better to define the registry key statically instead of using `Package.Current`, though migration is required.

**Resolved / Related Issues**

- Closes #15601 

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Utils/FileTags/FileTagsDatabase.cs
+++ b/src/Files.App/Utils/FileTags/FileTagsDatabase.cs
@@ -12,10 +12,14 @@ namespace Files.App.Utils.FileTags
 {
 	public sealed class FileTagsDatabase
 	{
-		private readonly static string FileTagsKey = @$"Software\Files Community\{Package.Current.Id.FullName}\v1\FileTags";
+		private static string? _FileTagsKey;
+		private string? FileTagsKey => _FileTagsKey ??= SafetyExtensions.IgnoreExceptions(() => @$"Software\Files Community\{Package.Current.Id.FullName}\v1\FileTags");
 
 		public void SetTags(string filePath, ulong? frn, string[] tags)
 		{
+			if (FileTagsKey is null)
+				return;
+
 			using var filePathKey = Registry.CurrentUser.CreateSubKey(CombineKeys(FileTagsKey, filePath));
 
 			if (tags is [])
@@ -47,6 +51,9 @@ namespace Files.App.Utils.FileTags
 
 		private TaggedFile? FindTag(string? filePath, ulong? frn)
 		{
+			if (FileTagsKey is null)
+				return null;
+
 			if (filePath is not null)
 			{
 				using var filePathKey = Registry.CurrentUser.CreateSubKey(CombineKeys(FileTagsKey, filePath));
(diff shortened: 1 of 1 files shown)


hez2010:
I'm wondering under what circumstances can this throw an exception?
Is someone trying to run Files as an unpackage app?

hishitetsu:
You may be right, or it may be a bug of Windows. But as long as the exception is occurring, I think we should take care of it.

hez2010:
I think the issue is not caused by end users but instead by developers. 
Someone tried to build and launch Files from VS without selecting `Files.Package` and encountered this exception. 
I don't think it's worth adding complexity to Files to address this subtle issue. Even if we fix this issue, they will immediately encounter into another exception because umpackaged deployment is not supported at all.

0x5bfa:
If we can detect if unpackaged, we could close main window and show only classic message dialog saying *Unpackaged launching is unsupported, build from Files.App (Package) project* at least.

hishitetsu:
> I think the issue is not caused by end users but instead by developers. Someone tried to build and launch Files from VS without selecting `Files.Package` and encountered this exception. I don't think it's worth adding complexity to Files to address this subtle issue. Even if we fix this issue, they will immediately encounter another exception (which is hidden by this one) because unpackaged deployment is not supported at all.

In that case, wouldn't an exception report be sent to Sentry?

0x5bfa:
Same if you click exe directly in installed path

yaira2:
> I think the issue is not caused by end users but instead by developers. Someone tried to build and launch Files from VS without selecting `Files.Package` and encountered this exception. I don't think it's worth adding complexity to Files to address this subtle issue. Even if we fix this issue, they will immediately encounter another exception (which is hidden by this one) because unpackaged deployment is not supported at all.

@hez2010 Sentry is only on the release version of Files so this has to be coming from a packaged app.

yaira2 (approved):

yaira2:
If it helps, we can try logging additional info about that state of the app to try figuring out how this occurs.

Issue #15504, opened 2024-05-30
Code Quality: Generative resource management using Source Generator


0x5bfa:
### Description

Instead of pasting raw string code of resources let’s have a source generator to use properties. The source generator will update the class right after resw file changed.

### Concerned code

None. Making new robust code.

### Gains

- Robustness

### Requirements

App.Strings.Strings partial class.

If comments start with `GENERIC` or something we can apart the class into Strings.Generic, Strings.Infecrectable, Strings.Settings, etc. this is just in my mind and not official proposal tho.

Here’s code:
```c#
// Strings.Inflectsble.cs
/// <auto-generated/>
public partial class Strings
{
    /// <summary>
    /// (generated comments from resw here)
    /// </summary>
    public string AddNewItem
        => “Add new items”;
}
``` 

```c#
Strings.AddNewItems.GetLocalized(itemCount);
```

If we want to make localization methods simplify, let's make Strings.Extension class file as partial.

```cs
// Strings.Extension.cs
[MarkupExtensionReturnType(ReturnType = typeof(string))]
public partial class Strings : MarkupExtension
{
    public string Name { get; set; }

    protected override object ProvideValue()
        => GetType().GetProperty(Name)?.GetValue(this);
}
```

0x5bfa:
No one reacts…

@XTorLukas fyi

0x5bfa:
If there’s slash the SG avoids to generate it, since we’re in a long term to reduce number of strings that contains it.

XTorLukas:
Yes I agree that a plain RAW string is not quite suitable for finding errors and editing later. I would also omit the slash. But splitting would be useful e.g. Strings.KeyDelDescription.GetLocalized() --> Key: "Key.Del.Description" Value: "Del". Possible splitting according to a predefined parser that replaces 'Key' with 'Key.' ... This would not use a slash in user space but in the background of the code

XTorLukas:
Generating a class using resx would certainly be useful, but creating it would not be trivial and the subsequent implementation would have to be on the git side. If you dare, I can possibly help you with creating a parser and implementing it in the project

0x5bfa:
Then replace it with underscore.

0x5bfa:
> Generating a class using resx would certainly be useful, but creating it would not be trivial and the subsequent implementation would have to be on the git side. If you dare, I can possibly help you with creating a parser and implementing it in the project

I guess Source Generator has features specifically for this kind of use case. CsWin32 uses it.

[...]

PR #15662, opened 2024-06-24
Code Quality: Generative resource management and refactoring

XTorLukas:
**Resolved / Related Issues**

- From #15504

1. Generated Strings properties from keys in primary Resources file
2. Compatible import from .resw and .json file


0x5bfa:
Is this still on-development?
If you are ok, I’d like to review.

0x5bfa:
Could you duplicate [ResourceHelpers](https://github.com/files-community/Files/blob/main/src/Files.App/Helpers/ResourceHelpers.cs) in `Strings.MarkupExtension.cs` as partial class of `Strings`?

Sample impl:
```cs
[MarkupExtensionReturnType(ReturnType = typeof(string))]
public partial class Strings : MarkupExtension
{
    public string Name { get; set; }

    protected override object ProvideValue()
        => GetType().GetProperty(Name)?.GetValue(this);
}
```

Sample usage (not for now):
```xml
<TextBlock Text="{helpers:Strings Name=Allow}" />
```

0x5bfa (commented):
- Make use of `SyntaxKind` for C# keywords
- Make use of `Constants` class for raw strings in generated file such as `<auto-generated/>` and license header.

XTorLukas:
> ```
> <TextBlock Text="{helpers:Strings Name=Allow}" />
> ```

I would use instead of Name=, Key=
```
  <TextBlock Text="{helpers:Strings Key=Allow}" />
```

XTorLukas:
> ```cs
> [MarkupExtensionReturnType(ReturnType = typeof(string))]
> public partial class Strings : MarkupExtension
> {
>     public string Name { get; set; }
> 
>     protected override object ProvideValue()
>         => GetType().GetProperty(Name)?.GetValue(this);
> }
> ```

This way it would be usable, you would just have to introduce a rule into the generator that key names e.g. 'Key' or 'Name' cannot be used because they are used internally. Evaluating with an analyzer would be good.

src/Files.App/Files.App.csproj
```diff
@@ -134,5 +134,9 @@
         <TrimmerRootAssembly Include="CommunityToolkit.WinUI.UI.Controls.Media" />
         <TrimmerRootAssembly Include="CommunityToolkit.WinUI.UI.Controls.Primitives" />
     </ItemGroup>
+
+	<ItemGroup>
+		<AdditionalFiles Include="Strings\en-US\Resources.resw" />
+	</ItemGroup>
```
yaira2, line 140: Files are included even without manually listing them here.
```suggestion
```

XTorLukas: If I don't specify a file for the source generator, it won't find it. The file must be added as AdditionalFiles. Do you have another way to do this without listing the file in AdditionalFiles?

0x5bfa: CsWin32 doesn’t seem to require this for NativeMethods.txt while the .NET banned API analyzer requires.

XTorLukas:
> * Make use of `SyntaxKind` for C# keywords

- [x] Finished
> * Make use of `Constants` class for raw strings in generated file such as `<auto-generated/>` and license header.

- [x] Finished

### Now need testing

src/Files.Core.SourceGenerator/Files.Core.SourceGenerator.csproj
```diff
@@ -19,8 +19,10 @@
     </PropertyGroup>
 
     <ItemGroup>
+        <PackageReference Include="MarcosLopezC.LightJson" Version="0.5.3" PrivateAssets="all" />
```
yaira2, line 22: We're trying hard to avoid additional references especially from third parties. Fyi @0x5bfa

0x5bfa: Source generator affects package size?

src/Files.Core.SourceGenerator/Data/AdditionalTextWithHash.cs
```diff
@@ -0,0 +1,29 @@
...
+	internal readonly record struct AdditionalTextWithHash(AdditionalText File, Guid Hash)
+	{
+		public bool Equals(AdditionalTextWithHash other)
+			=> File.Path.Equals(other.File.Path) && Hash.Equals(other.Hash);
+
+		public override int GetHashCode()
+		{
+			unchecked
+			{
+				return (File.GetHashCode() * 397) ^ Hash.GetHashCode();
```
0x5bfa, line 22: This can return a value larger than int.MaxValue?

src/Files.Core.SourceGenerator/Generators/DependencyPropertyGenerator.cs
0x5bfa, line 1: Could you revert L74?
I have a PR I worked on improving here.

0x5bfa (commented):
This might collide with my PR but looks good to me.

src/Files.Core.SourceGenerator/Data/AdditionalTextWithHash.cs
XTorLukas: In the end, I didn't use this class

src/Files.Core.SourceGenerator/Generators/DependencyPropertyGenerator.cs
XTorLukas: Revert all code changes in file or only L74?

XTorLukas: Ok

src/Files.Core.SourceGenerator/Data/AdditionalTextWithHash.cs
0x5bfa: Let's remove then
I used Guid.New() to avoid name collision.

XTorLukas: Ok

src/Files.Core.SourceGenerator/Utilities/SourceGeneratorHelper.cs
```diff
@@ -1,19 +1,18 @@
...
 
-namespace Files.Core.SourceGenerator
+namespace Files.Core.SourceGenerator.Utilities
 {
+	/// <summary>
+	/// Helper methods and constants for source code generation tasks.
+	/// </summary>
 	internal static class SourceGeneratorHelper
 	{
+		internal const string AttributeNamespace = $"{nameof(Files)}.App.Data.Attributes.";
```
0x5bfa, line 13: Now we use Files.Shared.Attributes

0x5bfa (approved):
Every methods are well-documented! LGTM and thanks for the conflict resolving.
Could you share the latest out out as a screenshot?

yaira2:
Thank you

PR #15670, opened 2024-06-24
Fix: Fixed crash that could occur when adding items to the jumplist

yaira2:
**Resolved / Related Issues**

- https://files-org.sentry.io/issues/5461933781/events/298838d246a84ccda2706468c3fba438

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Services/Windows/WindowsJumpListService.cs
+++ b/src/Files.App/Services/Windows/WindowsJumpListService.cs
@@ -46,7 +46,10 @@ namespace Files.App.Services
 						await instance.SaveAsync();
 					}
 				}
-				catch { }
+				catch (Exception ex)
+				{
+					App.Logger.LogWarning(ex, ex.Message);
+				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


src/Files.App/Services/Windows/WindowsJumpListService.cs
```diff
@@ -46,7 +46,10 @@ public async Task AddFolderAsync(string path)
 						await instance.SaveAsync();
 					}
 				}
-				catch { }
+				catch (Exception ex)
+				{
+					App.Logger.LogWarning(ex, ex.Message);
+				}
```
hishitetsu, line 52: This is simply adding log output, which doesn't help avoid crashes.
I think we need to expand the try block.

yaira2: I'll do that, but it will be nice to fix these errors.

There are some reports about jumplists not populating.

hishitetsu (approved):
LGTM

Issue #14875, opened 2024-03-01
Code Quality: Use seperate json file for tabs list


yaira2:
### Description

We currently update the settings file whenever there is a change to the tab list, this requires serializing the settings into json and isn't nesasacry.

### Concerned code

`AppLifecycleHelper.SaveSessionTabs();`

### Gains

- Better performance

### Requirements

- save the tabs list to its own json file

d2dyno1:
The only challenging bit of this issue is migrating existing tabs to the new format

yaira2:
> The only challenging bit of this issue is migrating existing tabs to the new format

How about checking if there is an existing list and copying them to the new format, once the existing list is empty we can ignore the check moving forward.

saikotek:
Hello, I think I could do it.

From what I can see you share singleton instance of `UserSettingsService` across all classes of settings by passing ISettingsSharingContext to them. This means that property `userSettingsService.GeneralSettingsService.LastSessionTabList` uses the same `UserSettingsService` singleton which saves data through `CachingJsonSettingsDatabase`. I could register a new service just like `FileTagsSettingsService` to save to its own file, right?

If my understanding is correct then I can attempt to do it :)

yaira2:
@d2dyno1 is that right?

0x5bfa:
Yes, correct.

0x5bfa:
In order to avoid using left and right wording, we might as well use a list instead.
```json
[
  {
    "TabList":
    [
      "C:\\Users",
      "C:\\Windows",
    ]
    "ShellPaneArrangement": 0
  },
  {
    "TabList":
    [
      "C:\\Windows\\System32",
      "C:\\Program Files (x86)",
    ]
    "ShellPaneArrangement": 1
  },
  {
    "TabList":
    [
      "C:\\Users\\0x5BFA\\Desktop",
      "C:\\Users\\0x5BFA\\AppData\\Local",
    ]
    "ShellPaneArrangement": 0
  }
]
```

yaira2:
Something like that 👍

PR #15680, opened 2024-06-25
Feature: Added support for choosing the dual pane orientation

### LINKED ISSUE #12427: Feature: Support choosing the dual pane orientation

The current and vertical panels (one next to the other) are very cool, but they have a problem of poor use of the screen on large screens, especially when using the column layout!
It could still exist, but it would be great to have another option for horizontal panels (one on top of the other), this would be perfect for being able to visualize a whole hierarchy of columns on both sides!

[actual problem]
image

[concept solution]
image

### Requirements

- keep current dashboards for those who prefer
- add a new option to view panels one above the other instead of next to each other

### Files Version

2.4.71.0

### Windows Version

Windows 11 22h2 22621.1702

### Comments

_No response_

0x5bfa:
### Resolved / Related Issues

- Closes #12427

- Check for splitting
  1. Click the Tab Actions button
  2. Click 'Split pane'
  3. Add pane in both ways
- Check for arranging
  1. Click the Tab Actions button
  2. Click 'Arrange pane'
  3. Arrange panes in both ways
- Check for the saved arrangement of panes
  1. Arrange panes vertically
  2. Close the window
  3. Re-open
- Check for the setting to open dual pane always
  1. Select Horizontal/Vertical
  2. Open a new tab
  3. Open a new pane


src/Files.App/Data/Commands/Manager/CommandCodes.cs
```diff
@@ -186,10 +186,13 @@ public enum CommandCodes
 		PreviousTab,
 		NextTab,
 		CloseSelectedTab,
+
+		// Shell Panes
 		OpenNewPane,
```
yaira2, line 191: Will this action be removed?

src/Files.App/Data/Commands/Manager/CommandCodes.cs
```diff
@@ -186,10 +186,13 @@ public enum CommandCodes
...
 		CloseActivePane,
 		FocusLeftPane,
 		FocusRightPane,
+		SplitPane,
```
yaira2, line 195: I think we'll need an action for horizontal and vertical.

src/Files.App/Data/Commands/Manager/CommandCodes.cs
0x5bfa: No.

src/Files.App/Data/Commands/Manager/CommandCodes.cs
0x5bfa: Done.

src/Files.App/Data/Commands/Manager/CommandCodes.cs
yaira2: I thought it's replaced by the vertical and horizontal actions.

0x5bfa: @mdtauk Since I haven't seen Open new pane and have seen Close active pane in your concepts, I thought there's still Open new pane. What should I do?

yaira2: The "Split pane" menu is replacing "Open new pane"

image

0x5bfa: Ah, then yes.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2672,23 +2672,14 @@
...
   <data name="CloseActivePaneDescription" xml:space="preserve">
     <value>Close active pane</value>
   </data>
-  <data name="FocusLeftPane" xml:space="preserve">
-    <value>Focus left pane</value>
-  </data>
-  <data name="FocusLeftPaneDescription" xml:space="preserve">
-    <value>Switch focus to the left pane</value>
+  <data name="FocusOtherPane" xml:space="preserve">
+    <value>Focus other pane</value>
```
0x5bfa, line 2679:
```suggestion
    <value>Focus the other pane</value>
```

yaira2: The title is supposed to be short so I think we can leave as is.

src/Files.App/Services/Settings/GeneralSettingsService.cs
```diff
@@ -267,6 +267,12 @@ public string UserId
 			set => Set(value);
 		}
 
+		public ShellPaneAlignmentDirection ShellPaneAlignmentDirection
```
yaira2, line 270: Wouldn't this force all tabs to have the same orientation?

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -39,6 +42,15 @@ public UIElement FooterElement
 			set => SetValue(FooterElementProperty, value);
 		}
 
+		public bool ShowTabActionsButton
+		{
+			get => AppearanceSettingsService.ShowTabActions;
+			set
```
yaira2, line 48: It's a one way binding so I think we can remove `set`.

0x5bfa:
@yaira2 because the text `ShowTabActions` has `button` and the control type of `ToggleSwitch` is `button` thus the text includes the control type name, the AXE test is failing. Could you consider text that doesn't include `button`?

image

src/Files.App/Services/Settings/GeneralSettingsService.cs
0x5bfa: True, it overrides.
I think we should include this in `LastOpnedTabsList`?

yaira2: As an arg?

0x5bfa: Yeah probably adding the enum to the parameter class. I’ll try.

yaira2: Once we implement #14875, we'll be able to add properties for each tab in the list. In the meantime, we can just add it as a param.

0x5bfa: Done.

src/Files.App/Data/Contexts/ContentPage/ContentPageContext.cs
```diff
@@ -43,7 +43,7 @@ internal sealed class ContentPageContext : ObservableObject, IContentPageContext
 
 		public bool CanCreateItem => GetCanCreateItem();
 
-		public bool IsMultiPaneEnabled => ShellPage is not null && ShellPage.PaneHolder is not null && ShellPage.PaneHolder.IsMultiPaneEnabled;
+		public bool IsMultiPaneEnabled => ShellPage is not null && ShellPage.PaneHolder is not null && ShellPage.PaneHolder.CanBeDualPane;
```
yaira2, line 46:
```suggestion
		public bool IsMultiPaneAvailable => ShellPage is not null && ShellPage.PaneHolder is not null && ShellPage.PaneHolder.CanBeDualPane;
```

src/Files.App/Data/Contexts/MultiPanes/IMultiPanesContext.cs
```diff
@@ -18,6 +18,11 @@ public interface IMultiPanesContext
 		/// </summary>
 		event EventHandler? ActivePaneChanged;
 
+		/// <summary>
+		/// Gets invoked when shell pane arrangement is changed.
```
yaira2, line 22:
```suggestion
		/// Invoked when shell pane arrangement is changed.
```

yaira2:
1. The "Open in secondary pane" action opens an additional pane even when two panes area already opened. The desired behavior is to replace the content in the other pane (we can display a prompt to confirm that the user wants to do this).
2. Continue where you left off doesn't restore with the correct arrangement.
3. The cursor for the gripper doesn't update after changing the arrangement.

src/Files.App/Data/Contexts/MultiPanes/IMultiPanesContext.cs
0x5bfa: It's out of scope for now but I'd like tospecify our guideline for event comments in order to keep consistency.

WinAppSdk API starts with `Occurs during ...`, while WCT API starts with `Fires when ...`
image

Reference: https://github.com/search?q=repo%3Afiles-community%2FFiles+%22Gets+invoked%22&type=code

Which one would you like or would you like to get it started with `Invoked ...`?

mdtauk:
> 4. If one tab is arranged vertically and the other horizontally, the arrangement menu doesn’t update when switching between tabs.

Can you ever be in a situation where two tabs are open in different arrangements?

src/Files.App/Data/Contexts/MultiPanes/IMultiPanesContext.cs
yaira2: `Occurs` or `Invoked` is fine with me.

src/Files.App/Data/Contexts/ContentPage/ContentPageContext.cs
0x5bfa: I used to want it to be CanBeDualPane but this makes more sense.
Renamed.

0x5bfa:
1. Suppressed by preventing opening more than 2 panes.
2. I cannot repro.
3. Fixed.
4. Fixed.

yaira2:
All the issues I reported (including the second one) are resolved.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -49,17 +50,11 @@ public bool AllowTabsDrag
 		public Rectangle DragArea
 			=> DragAreaRectangle;
 
-		/// <summary> Starting position when dragging a tab.</summary>
```
yaira2, line 52: Please restore the comments

0x5bfa: Do you really think this is helpful?

yaira2: Yeah, otherwise it's not so clear what it's referring to.

yaira2 (dismissed):
LGTM

yaira2 (approved):

hishitetsu (commented):
It's working almost perfectly!
The only problem I have found is that the Arrange panes checkmark disappears when the same arrangement is selected.

yaira2:
We might need to remove the command and making the IsChecked binding two way.

PR #15724, opened 2024-06-28
Code Quality: Simplify UI operation layer for Copy and Cut

0x5bfa:
### Summary

This is a part of Storage Abstraction.
The short-term goals are:
- Remove UIFilesystemHelpers (doing this)
- Refactor ShellViewModel
- Create StorageQueryService
- Create StorageEnumerationService
- Create StorageOperationService

### Resolved / Related Issues

- `https://github.com/files-community/Files/issues/8974`

1. Open Files app
2. Do cut, copy


yaira2:
This looks good to me on the surface, but further review is required before merging.

src/Files.App/Actions/FileSystem/Transfer/BaseTransferItemAction.cs
```diff
@@ -0,0 +1,129 @@
...
+					string[] filePaths = ContentPageContext.SelectedItems.Select(x => x.ItemPath).ToArray();
+					await FileOperationsHelpers.SetClipboard(filePaths, type);
+
+					return;
+				}
+
+				// Reset cut mode
+				ContentPageContext.ShellPage.SlimContentPage.ItemManipulationModel.RefreshItemsOpacity();
+
+				return;
```
0x5bfa, line 114: @hishitetsu I’m wondering if the app should report the produced exception since this try-catch may throw IOException (and others in WinRT APIs) before returning.

src/Files.App/Actions/FileSystem/Transfer/BaseTransferItemAction.cs
```diff
@@ -0,0 +1,129 @@
...
+								: await ContentPageContext.ShellPage.ShellViewModel.GetFolderFromPathAsync(listedItem.ItemPath).OnSuccess(t => items.Add(t));
+
+						if (!result)
+							throw new IOException($"Failed to process {listedItem.ItemPath} in cutting/copying to the clipboard.", (int)result.ErrorCode);
+					}
+				},
+				10,
+				statusCenterItem?.CancellationToken ?? default);
+
+				var standardObjectsOnly = items.All(x => x is StorageFile or StorageFolder or SystemStorageFile or SystemStorageFolder);
```
0x5bfa, line 87: Just a note: using StorageFile and/or StorageFolder for `DataPackage.SetStorageItems` method is fair enough since what this way produces is the exactly same as what File Explorer produces. However, as we’re in a move of avoiding WinRT APIs in an aspect of slowness, I may want to create our own service using Win32 API or store path strings instead of objects of WinRT. With that’s said we could create converter to WinRT objects in order to avoid such hard work.

src/Files.App/Actions/FileSystem/Transfer/BaseTransferItemAction.cs
hishitetsu: If the exception occurs for unexpected reasons, it should be reported. But if you know why the exception occurs, for example, it occurs by a known user operation, it would be better not to report it.

0x5bfa: Done.

src/Files.App/Actions/FileSystem/Transfer/BaseTransferItemAction.cs
```diff
@@ -0,0 +1,135 @@
...
+				Clipboard.SetContent(dataPackage);
+			}
+			catch (Exception ex)
+			{
+				dataPackage = default;
+
+				if (ex is not IOException)
+				{
+					App.Logger.LogWarning(ex, "Failed to process cutting/copying due to an unknown error.");
+				}
```
hishitetsu, line 107:
```suggestion
				if (ex is not IOException)
					App.Logger.LogWarning(ex, "Failed to process cutting/copying due to an unknown error.");
```

hishitetsu (approved):
LGTM

Issue #15698, opened 2024-06-26
Bug: Google Drive doesn't show in Cloud Drives when mounts as a folder


kohyung:
### Description

Google Drive didn't show up in Cloud Drives Sidebar, Dropbox & Onedrive got no problem.
![screenshot-1719384747145](https://github.com/files-community/Files/assets/65665469/7c6ffcb2-cfd9-4baa-a0c3-046732f6ca10)

### Steps To Reproduce

Just new install already like that.
![screenshot-1719384747145](https://github.com/files-community/Files/assets/65665469/8f821506-752d-4a8a-ad15-6c872e3c9df3)

### Requirements

New installation already like that.
![screenshot-1719384747145](https://github.com/files-community/Files/assets/65665469/7acd70ae-3307-4727-a305-40b48c442195)

### Files Version

3.5.0.0

### Windows Version

10.0.19045.4529

### Log File

[debug.log](https://github.com/user-attachments/files/15983692/debug.log)

Josh65-2201:
Thanks for the report. Have you set up the desktop client, opened it and restarted Files from the system tray?

kohyung:
Hi, pls refer to the attachment, I have tried a few times.

Josh65-2201:
Email attachments don't work. You will have to click the `View on GitHub` link at the bottom of the email > Copy and paste or click the add attachment button to select the file > Wait until you don't see `uploading` in the text > Click the green comment button.

kohyung:
Hi, pls refer to the attachment, I have tried a few times.
![screenshot-1719404309390](https://github.com/files-community/Files/assets/65665469/f6cccedf-21bd-40f5-88ee-ea2d7ae1b2db)

Josh65-2201:
Can you send a screenshot of the full Windows File Explorer window on `This PC`

kohyung:
![Screenshot 2024-06-26 22 33 00](https://github.com/files-community/Files/assets/65665469/72f0fa90-2d14-49fe-8738-3f206a837919)

Josh65-2201:
Need a screenshot from Windows File Explorer, not Files.

kohyung:
![Screenshot 2024-06-26 22 37 44](https://github.com/files-community/Files/assets/65665469/423aadfe-bf9a-46cd-98eb-02b28faed95f)

hez2010:
We don't support Google Drive mounting as a folder instead of a dedicate drive currently.

[...]

PR #15771, opened 2024-07-07
Fix: Fixed an issue where Google Drive wasn't displayed when mounted as a folder

wharvex:
### Summary

* Registry-artifact-based Google Drive mounting point detection

  * [x] Add method for reading Google Drive registry artifacts and parsing them as JSON.

  * [x] Add method for safely traversing the resulting JSON and creating/returning new CloudProvider objects based on the data inside.

  * [x] Call these methods after the old approach that uses the "root_preference_sqlite" file. Google Drive now appears under Cloud Drives" in the sidebar regardless of whether it is mounted as a drive or a folder.

* GoogleDriveCloudDetector private fields

  * [x] Add private fields to GoogleDriveCloudDetector (GDCD) for the strings used to access/traverse Google Drive registry artifacts.

  * [x] Add field to GDCD for the logger service so we can log warnings when artifact access/traversal attempts fail.

* Icon-file-getting logic

  * [x] Create method for getting the icon file so we can reuse this logic.

  * [x] Improve null safety of icon file getting logic.

### Resolved / Related Issues

* https://github.com/files-community/Files/issues/15698

1. Under Google Drive Preferences -> Settings -> Google Drive Streaming Location, choose "Folder" or "Drive Letter".

    1. Folder

        1. I opened Files and looked at the sidebar and "Google Drive" was there under "Cloud Drives".

        2. I clicked on "Google Drive" and it opened my drive with all its contents viewable and accessible, as expected. I opened a few of my files (images and pdfs) and they opened as expected.

    2. Drive Letter

        1. I opened Files and looked at the sidebar and "Google Drive" was there under "Cloud Drives".

        2. I clicked on "Google Drive" and it opened my drive with all its contents viewable and accessible, as expected. I opened a few of my files (images and pdfs) and they opened as expected.

2. I also found that everything worked exactly the same if I had the entire old approach that uses the "root_preference_sqlite" file commented out. However, I did not want to commit it in this state because my testing was not especially thorough, but it might be worth looking into whether this will work now entirely without the old approach.

### DIFF
--- a/src/Files.App/Files.App.csproj
+++ b/src/Files.App/Files.App.csproj
@@ -86,8 +86,8 @@
         <PackageReference Include="TagLibSharp" Version="2.3.0" />
         <PackageReference Include="Tulpep.ActiveDirectoryObjectPicker" Version="3.0.11" />
         <PackageReference Include="WinUIEx" Version="2.3.4" />
-        <PackageReference Include="Vanara.Windows.Extensions" Version="4.0.1" />
-        <PackageReference Include="Vanara.Windows.Shell" Version="4.0.1" />
+        <PackageReference Include="Vanara.Windows.Extensions" Version="4.0.2" />
+        <PackageReference Include="Vanara.Windows.Shell" Version="4.0.2" />
         <PackageReference Include="Microsoft.Management.Infrastructure" Version="3.0.0" />
         <PackageReference Include="Microsoft.Management.Infrastructure.Runtime.Win" Version="3.0.0" />
         <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.3.106" PrivateAssets="all" />
--- a/src/Files.App/Files.App.csproj
+++ b/src/Files.App/Files.App.csproj
@@ -86,8 +86,8 @@
         <PackageReference Include="TagLibSharp" Version="2.3.0" />
         <PackageReference Include="Tulpep.ActiveDirectoryObjectPicker" Version="3.0.11" />
         <PackageReference Include="WinUIEx" Version="2.3.4" />
-        <PackageReference Include="Vanara.Windows.Extensions" Version="4.0.2" />
-        <PackageReference Include="Vanara.Windows.Shell" Version="4.0.2" />
+        <PackageReference Include="Vanara.Windows.Extensions" Version="4.0.1" />
+        <PackageReference Include="Vanara.Windows.Shell" Version="4.0.1" />
         <PackageReference Include="Microsoft.Management.Infrastructure" Version="3.0.0" />
         <PackageReference Include="Microsoft.Management.Infrastructure.Runtime.Win" Version="3.0.0" />
         <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.3.106" PrivateAssets="all" />
--- a/src/Files.App/Services/Storage/StorageDevicesService.cs
+++ b/src/Files.App/Services/Storage/StorageDevicesService.cs
@@ -19,30 +19,20 @@ namespace Files.App.Services
 		public async IAsyncEnumerable<ILocatableFolder> GetDrivesAsync()
 		{
 			var list = DriveInfo.GetDrives();
-			var googleDrivePath = App.AppModel.GoogleDrivePath;
 			var pCloudDrivePath = App.AppModel.PCloudDrivePath;
 
+			var sw = Stopwatch.StartNew();
+			var googleDrivePath = GoogleDriveCloudDetector.GetRegistryBasePath();
+			sw.Stop();
 #if DEBUG
-			Debug.WriteLine($"In RDS.GDA: googleDrivePath: {googleDrivePath}");
+			Debug.WriteLine($"In RemovableDrivesService: Time elapsed for registry check: {sw.Elapsed}");
 #endif
+			App.AppModel.GoogleDrivePath = googleDrivePath ?? string.Empty;
 
 			foreach (var drive in list)
 			{
-				var shouldSkip = false;
-				var sw = Stopwatch.StartNew();
-				await foreach (var cloudProvider in GoogleDriveCloudDetector.GetGoogleDriveProvidersFromRegistryAsync(false))
-				{
-					if (cloudProvider.SyncFolder.Equals(drive.Name))
-						shouldSkip = true;
-				}
-				sw.Stop();
-
-#if DEBUG
-				Debug.WriteLine($"In RDS.GDA after registry check: Time elapsed for filter: {sw.Elapsed}");
-				Debug.WriteLine($"In RDS.GDA: drive.Name: {drive.Name}");
-#endif
-
-				if (shouldSkip)
+				// We don't want cloud drives to appear in a plain "Drives" section.
+				if (drive.Name.Equals(googleDrivePath) || drive.Name.Equals(pCloudDrivePath))
 					continue;
 
 				var res = await FilesystemTasks.Wrap(() => StorageFolder.GetFolderFromPathAsync(drive.Name).AsTask());
@@ -63,10 +53,6 @@ namespace Files.App.Services
 				var type = DriveHelpers.GetDriveType(drive);
 				var label = DriveHelpers.GetExtendedDriveLabel(drive);
 				var driveItem = await DriveItem.CreateFromPropertiesAsync(res.Result, drive.Name.TrimEnd('\\'), label, type, thumbnail);
-
-				// Don't add here because Google Drive is already displayed under cloud drives
-				if (drive.Name == googleDrivePath || drive.Name == pCloudDrivePath)
-					continue;
 
 				App.Logger.LogInformation($"Drive added: {driveItem.Path}, {driveItem.Type}");
 
--- a/src/Files.App/Services/Storage/StorageDevicesService.cs
+++ b/src/Files.App/Services/Storage/StorageDevicesService.cs
@@ -22,8 +22,20 @@ namespace Files.App.Services
 			var googleDrivePath = App.AppModel.GoogleDrivePath;
 			var pCloudDrivePath = App.AppModel.PCloudDrivePath;
 
+#if DEBUG
+			Debug.WriteLine($"In RDS.GDA: googleDrivePath: {googleDrivePath}");
+#endif
+
 			foreach (var drive in list)
 			{
+				var shouldSkip = false;
+				await foreach (var cloudProvider in GoogleDriveCloudDetector.GetGoogleDriveProvidersFromRegistryAsync(false))
+				{
+					if (cloudProvider.SyncFolder.Equals(drive.Name))
+						shouldSkip = true;
+				}
+				if (shouldSkip)
+					continue;
 				var res = await FilesystemTasks.Wrap(() => StorageFolder.GetFolderFromPathAsync(drive.Name).AsTask());
 				if (res.ErrorCode is FileSystemStatusCode.Unauthorized)
 				{
@@ -43,6 +55,9 @@ namespace Files.App.Services
 				var label = DriveHelpers.GetExtendedDriveLabel(drive);
 				var driveItem = await DriveItem.CreateFromPropertiesAsync(res.Result, drive.Name.TrimEnd('\\'), label, type, thumbnail);
 
+#if DEBUG
+				Debug.WriteLine($"In RDS.GDA: drive.Name: {drive.Name}");
+#endif
 				// Don't add here because Google Drive is already displayed under cloud drives
 				if (drive.Name == googleDrivePath || drive.Name == pCloudDrivePath)
 					continue;
--- a/src/Files.App/Services/Storage/StorageDevicesService.cs
+++ b/src/Files.App/Services/Storage/StorageDevicesService.cs
@@ -28,23 +28,22 @@ namespace Files.App.Services
 
 			foreach (var drive in list)
 			{
-				var sw = Stopwatch.StartNew();
 				var shouldSkip = false;
+				var sw = Stopwatch.StartNew();
 				await foreach (var cloudProvider in GoogleDriveCloudDetector.GetGoogleDriveProvidersFromRegistryAsync(false))
 				{
 					if (cloudProvider.SyncFolder.Equals(drive.Name))
 						shouldSkip = true;
 				}
+				sw.Stop();
+
+#if DEBUG
+				Debug.WriteLine($"In RDS.GDA in await foreach: drive.Name: {drive.Name}");
+				Debug.WriteLine($"In RDS.GDA in await foreach: Time elapsed for filter: {sw.Elapsed}");
+#endif
 
 				if (shouldSkip)
-				{
-					sw.Stop();
-#if DEBUG
-					Debug.WriteLine($"In RDS.GDA in await foreach: drive.Name: {drive.Name}");
-					Debug.WriteLine($"In RDS.GDA in await foreach: Time elapsed for filter: {sw.Elapsed}");
-#endif
 					continue;
-				}
 
 				var res = await FilesystemTasks.Wrap(() => StorageFolder.GetFolderFromPathAsync(drive.Name).AsTask());
 				if (res.ErrorCode is FileSystemStatusCode.Unauthorized)
--- a/src/Files.App/Services/Storage/StorageDevicesService.cs
+++ b/src/Files.App/Services/Storage/StorageDevicesService.cs
@@ -24,9 +24,7 @@ namespace Files.App.Services
 			var sw = Stopwatch.StartNew();
 			var googleDrivePath = GoogleDriveCloudDetector.GetRegistryBasePath();
 			sw.Stop();
-#if DEBUG
 			Debug.WriteLine($"In RemovableDrivesService: Time elapsed for registry check: {sw.Elapsed}");
-#endif
 			App.AppModel.GoogleDrivePath = googleDrivePath ?? string.Empty;
 
 			foreach (var drive in list)
--- a/src/Files.App/Services/Storage/StorageDevicesService.cs
+++ b/src/Files.App/Services/Storage/StorageDevicesService.cs
@@ -28,14 +28,24 @@ namespace Files.App.Services
 
 			foreach (var drive in list)
 			{
+				var sw = Stopwatch.StartNew();
 				var shouldSkip = false;
 				await foreach (var cloudProvider in GoogleDriveCloudDetector.GetGoogleDriveProvidersFromRegistryAsync(false))
 				{
 					if (cloudProvider.SyncFolder.Equals(drive.Name))
 						shouldSkip = true;
 				}
+
 				if (shouldSkip)
+				{
+					sw.Stop();
+#if DEBUG
+					Debug.WriteLine($"In RDS.GDA in await foreach: drive.Name: {drive.Name}");
+					Debug.WriteLine($"In RDS.GDA in await foreach: Time elapsed for filter: {sw.Elapsed}");
+#endif
 					continue;
+				}
+
 				var res = await FilesystemTasks.Wrap(() => StorageFolder.GetFolderFromPathAsync(drive.Name).AsTask());
 				if (res.ErrorCode is FileSystemStatusCode.Unauthorized)
 				{
--- a/src/Files.App/Services/Storage/StorageDevicesService.cs
+++ b/src/Files.App/Services/Storage/StorageDevicesService.cs
@@ -38,8 +38,8 @@ namespace Files.App.Services
 				sw.Stop();
 
 #if DEBUG
-				Debug.WriteLine($"In RDS.GDA in await foreach: drive.Name: {drive.Name}");
-				Debug.WriteLine($"In RDS.GDA in await foreach: Time elapsed for filter: {sw.Elapsed}");
+				Debug.WriteLine($"In RDS.GDA after registry check: Time elapsed for filter: {sw.Elapsed}");
+				Debug.WriteLine($"In RDS.GDA: drive.Name: {drive.Name}");
 #endif
 
 				if (shouldSkip)
@@ -64,9 +64,6 @@ namespace Files.App.Services
 				var label = DriveHelpers.GetExtendedDriveLabel(drive);
 				var driveItem = await DriveItem.CreateFromPropertiesAsync(res.Result, drive.Name.TrimEnd('\\'), label, type, thumbnail);
 
-#if DEBUG
-				Debug.WriteLine($"In RDS.GDA: drive.Name: {drive.Name}");
-#endif
 				// Don't add here because Google Drive is already displayed under cloud drives
 				if (drive.Name == googleDrivePath || drive.Name == pCloudDrivePath)
 					continue;
--- a/src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
+++ b/src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
@@ -6,6 +6,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Win32;
 using System.IO;
 using Windows.Storage;
+// TODO: Avoid to use Vanara (#15000)
 using Vanara.Windows.Shell;
 
 namespace Files.App.Utils.Cloud
--- a/src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
+++ b/src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
@@ -72,6 +72,7 @@ namespace Files.App.Utils.Cloud
 				App.AppModel.GoogleDrivePath = path;
 
 #if DEBUG
+				Debug.WriteLine($"In GDCD in roots table: App.AppModel.GoogleDrivePath being set to: {path}");
 				Debug.WriteLine("YIELD RETURNING from GoogleDriveCloudDetector#GetProviders (roots): ");
 				Debug.WriteLine($"name=Google Drive ({title}); path={path}");
 #endif
@@ -106,6 +107,7 @@ namespace Files.App.Utils.Cloud
 				var iconFile = await GetGoogleDriveIconFileAsync();
 
 #if DEBUG
+				Debug.WriteLine($"In GDCD in media table: App.AppModel.GoogleDrivePath being set to: {path}");
 				Debug.WriteLine("YIELD RETURNING from GoogleDriveCloudDetector#GetProviders (media): ");
 				Debug.WriteLine($"name={title}; path={path}");
 #endif
@@ -160,7 +162,7 @@ namespace Files.App.Utils.Cloud
 #endif
 		}
 
-		private JsonDocument? GetGoogleDriveRegValJson()
+		private static JsonDocument? GetGoogleDriveRegValJson()
 		{
 			// This will be null if the key name is not found.
 			using var googleDriveRegKey = Registry.CurrentUser.OpenSubKey(_googleDriveRegKeyName);
@@ -192,7 +194,7 @@ namespace Files.App.Utils.Cloud
 			return googleDriveRegValueJson;
 		}
 
-		private async IAsyncEnumerable<ICloudProvider> GetGoogleDriveProvidersFromRegistryAsync()
+		public static async IAsyncEnumerable<ICloudProvider> GetGoogleDriveProvidersFromRegistryAsync(bool addMyDriveToPath = true)
 		{
 			var googleDriveRegValJson = GetGoogleDriveRegValJson();
 
@@ -228,13 +230,16 @@ namespace Files.App.Utils.Cloud
 				if (path is null)
 					continue;
 
-				if (!AddMyDriveToPathAndValidate(ref path))
+				if (!AddMyDriveToPathAndValidate(ref path, addMyDriveToPath))
 				{
 					_logger.LogWarning($"Validation failed for {path} (media)");
 					continue;
 				}
 
 				App.AppModel.GoogleDrivePath = path;
+#if DEBUG
+				Debug.WriteLine($"In GDCD in registry: App.AppModel.GoogleDrivePath being set to: {path}");
+#endif
 
 				var iconFile = await GetGoogleDriveIconFileAsync();
 
@@ -247,7 +252,7 @@ namespace Files.App.Utils.Cloud
 			}
 		}
 
-		private async Task<StorageFile?> GetGoogleDriveIconFileAsync()
+		private static async Task<StorageFile?> GetGoogleDriveIconFileAsync()
 		{
 			var programFilesEnvVar = Environment.GetEnvironmentVariable("ProgramFiles");
 
@@ -259,7 +264,7 @@ namespace Files.App.Utils.Cloud
 			return await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(iconPath).AsTask());
 		}
 
-		private bool AddMyDriveToPathAndValidate(ref string path)
+		private static bool AddMyDriveToPathAndValidate(ref string path, bool addMyDrive = true)
 		{
 			// If Google Drive is mounted as a drive, then the path found in the registry will be
 			// *just* the drive letter (e.g. just "G" as opposed to "G:\"), and therefore must be
@@ -280,27 +285,31 @@ namespace Files.App.Utils.Cloud
 				path = temp.RootDirectory.Name;
 			}
 
-			// If `path` contains a shortcut named "My Drive", store its target in `shellFolderBaseFirst`.
-			// This happens when "My Drive syncing options" is set to "Mirror files".
-			// TODO: Avoid to use Vanara (#15000)
-			using var shellFolderBase = ShellFolderExtensions.GetShellItemFromPathOrPIDL(path) as ShellFolder;
-			var shellFolderBaseFirst = Environment.ExpandEnvironmentVariables((
-				shellFolderBase?.FirstOrDefault(si =>
-					si.Name?.Equals("My Drive") ?? false) as ShellLink)?.TargetPath
-			    ?? string.Empty);
+			if (addMyDrive)
+			{
+				// If `path` contains a shortcut named "My Drive", store its target in `shellFolderBaseFirst`.
+				// This happens when "My Drive syncing options" is set to "Mirror files".
+				// TODO: Avoid to use Vanara (#15000)
+				using var shellFolderBase = ShellFolderExtensions.GetShellItemFromPathOrPIDL(path) as ShellFolder;
+				var shellFolderBaseFirst = Environment.ExpandEnvironmentVariables((
+						shellFolderBase?.FirstOrDefault(si =>
+							si.Name?.Equals("My Drive") ?? false) as ShellLink)?.TargetPath
+					?? string.Empty);
 
 #if DEBUG
-			Debug.WriteLine("INVALID PATHS LOGGER");
-			shellFolderBase?.ForEach(si => Debug.WriteLine(si.Name));
+				Debug.WriteLine("INVALID PATHS LOGGER");
+				shellFolderBase?.ForEach(si => Debug.WriteLine(si.Name));
 #endif
 
-			if (!string.IsNullOrEmpty(shellFolderBaseFirst))
-			{
-				path = shellFolderBaseFirst;
-				return true;
+				if (!string.IsNullOrEmpty(shellFolderBaseFirst))
+				{
+					path = shellFolderBaseFirst;
+					return true;
+				}
+
+				path = Path.Combine(path, "My Drive");
 			}
 
-			path = Path.Combine(path, "My Drive");
 			if (Directory.Exists(path))
 				return true;
 			_logger.LogWarning($"Invalid Google Drive mount point path: {path}");
--- a/src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
+++ b/src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
@@ -6,6 +6,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Win32;
 using System.IO;
 using Windows.Storage;
+using Vanara.Windows.Shell;
 
 namespace Files.App.Utils.Cloud
 {
@@ -101,7 +102,7 @@ namespace Files.App.Utils.Cloud
 				if (string.IsNullOrWhiteSpace(path))
 					continue;
 
-				if (!AddMyDriveToPathAndValidate(ref path))
+				if (!AddMyDriveToPathAndValidate(ref path, invalidPathsLogger))
 				{ 
 					invalidPathsLogger.LogInformation("Validation failed for " + path + " (media)");
 					continue;
@@ -219,6 +220,7 @@ namespace Files.App.Utils.Cloud
 
 		private async IAsyncEnumerable<ICloudProvider> GetGoogleDriveProvidersFromRegistryAsync(ILogger invalidPathsLogger)
 		{
+			var registryLogger = GetAltLogger("debugRegistry.log");
             var googleDriveRegValJson = GetGoogleDriveRegValJson();
 
             if (googleDriveRegValJson is null)
@@ -240,6 +242,9 @@ namespace Files.App.Utils.Cloud
 				yield break;
 			}
 
+			// TESTING
+			registryLogger.LogInformation(googleDriveRegValJsonProperty.ToString());
+
 			foreach (var item in googleDriveRegValJsonProperty.Value.EnumerateArray())
 			{
 				if (!item.TryGetProperty(_googleDriveRegValPropName, out var googleDriveRegValProp))
@@ -253,7 +258,7 @@ namespace Files.App.Utils.Cloud
 				if (path is null)
 					continue;
 
-				if (!AddMyDriveToPathAndValidate(ref path))
+				if (!AddMyDriveToPathAndValidate(ref path, invalidPathsLogger))
 				{
 					invalidPathsLogger.LogInformation("Validation failed for " + path + " (media)");
 					continue;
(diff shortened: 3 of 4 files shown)


0x5bfa:
Thank you for the contribution.
Is this ready?

wharvex:
@0x5bfa No problem. I would say it's ready to be merged in the sense that it seems to be working based on the steps I followed in the "Steps used to test these changes" section above.

However, I just wanted to get a second opinion on the code I wrote on lines 191-196.

When Google Drive is mounted as a folder, the value stored in `mount_point_path` will be a proper path that can be passed to a method like `StorageFolder.GetFolderFromPathAsync` without issue.

But when Google Drive is mounted as a drive, the value stored in `mount_point_path` is JUST the drive letter (e.g. "G"), and as such causes an error when passed to a method like `StorageFolder.GetFolderFromPathAsync`.

The solution I came up with for this issue was to check if the value stored in `mount_point_path` is exactly one character in length, and if it is, get a `DriveInfo` object for that character and get the proper useable path from that object.

The effectiveness/safety of this solution depends on two assumptions:

* If the value stored in mount_point_path is one character in length, it must be a drive letter.
* This drive letter will be a valid drive (if it's not, the DriveInfo constructor will throw an ArgumentException).

My particular solution is what allows us to get a useable path from the registry-artifact method regardless of Google Drive's mount point, and therefore what would allow us to potentially remove the old "root_preference_sqlite" mount point detection method if we wanted.

So I was just curious what your team's thoughts might be on my approach and on this issue in general.

wharvex:
I just made a new commit that makes the logic that gets the valid path for a single-character `mount_point_path` value a little safer.

In my opinion, it is now ready to merge. But please let me know if there is anything you want me to change.

Josh65-2201:
I found two minor issues, Google Drive set to mirror to a user selected folder will show both the redirect folder/drive (This show one folder that redirects to the store location) and another that goes directly to the store location. Preferably it would only show the direct folder in mirror mode.

Also the Google drive is now always shown in the drives section instead of being hidden like 3.5.6

wharvex:
@Josh65-2201 , I looked into the first issue today. I am still working on it, but I just wanted to share what I've found.

I took screenshots of the sidebar as it appears on my computer under the following conditions (all with my Google Drive set to mirror files):

1. With my code commented-out and with Google Drive mounted as a drive
sidebar_mirror_drive_letter_no_registry
2. With my code active and with Google Drive mounted as a drive
sidebar_mirror_drive_letter_registry
3. With my code commented-out and with Google Drive mounted as a folder
sidebar_mirror_folder_no_registry
4. With my code active and with Google Drive mounted as a folder
sidebar_mirror_folder_registry

I am thinking of changing my code so that it appends "My Drive" to the path I get from the registry, so it will be weeded out as a duplicate DriveItem in CloudDrivesManager.UpdateDrivesAsync() if a DriveItem with Path set to the "My Drive" path was already added by the Sqlite detection approach. Or please let me know if you would prefer a different solution.

As for the second issue, under what conditions is Google Drive hidden in 3.5.6?
Are you saying it's an issue that it's no longer hidden under these conditions?
My understanding was that my pull request should result in Google Drive no longer being hidden from the drives section when Google Drive is mounted as a folder.

Josh65-2201:
Google drive should always be hidden from the drives section and only show in the cloud drives section.

wharvex:
> Google drive should always be hidden from the drives section and only show in the cloud drives section.

Ah okay, I will look into that as well, thank you.

wharvex:
Sorry for the delay, I'm still investigating how to tell from the Sqlite database tables ("roots" and "media") that Google Drive is mirrored as opposed to just streaming. I'm hoping to have some real time to work on it this weekend.

In the meantime, I just wanted to ask, did it cause any issues that I updated the branch the other day? I just clicked the "Update branch" button on GitHub on this PR and then pulled the changes down to my local branch. But I just noticed there is a notification now that @yaira2 "force-pushed the main branch." Was this to fix something I did? And is the way I updated the branch (with a merge commit as opposed to rebase) the preferred way?

yaira2:
> In the meantime, I just wanted to ask, did it cause any issues that I updated the branch the other day? I just clicked the "Update branch" button on GitHub on this PR and then pulled the changes down to my local branch. But I just noticed there is a notification now that @yaira2 "force-pushed the main branch." Was this to fix something I did? And is the way I updated the branch (with a merge commit as opposed to rebase) the preferred way?

It's unrelated, we have protections enabled so you don't have to worry about accidently pushing to `main`.

wharvex:
Just to give an update, today I figured out how to create logs in separate files of the database inspections I've been doing. 

I have made a commit to my GoogleDriveDetection branch with this logging in place.

I plan to remove this logging once I determine and can demonstrate:

A) what this implementation of GetProviders "yield returns", 

B) what the contents of various Google Drive artifacts are (e.g. registry keys and Sqlite databases), and

C) whether any of these artifacts contain useful information for determining user preference configurations that tell us when we should or shouldn't yield return certain paths.

I plan to use the logging to gather data related to A, B, and C under these conditions:

My code active; Streaming; folder-mounted
My code active; Streaming; drive-mounted
My code active; Mirrored; folder-mounted
My code active; Mirrored; drive-mounted
My code inactive; Streaming; folder-mounted
My code inactive; Streaming; drive-mounted
My code inactive; Mirrored; folder-mounted
My code inactive; Mirrored; drive-mounted

I also learned about the existence of a "mirror_metadata_sqlite.db" file from [this](https://amgedwageh.medium.com/drivefs-sleuth-investigating-google-drive-file-streams-disk-artifacts-0b5ea637c980) article, it might be useful to try to lookup this database as well.

yaira2:
Thanks for the update.

wharvex:
With the two commits that I just made, I think that the first issue Josh 
mentioned has been fixed (see my copied commit message below).

I have also confirmed that when running the main branch, that Google Drive still
appears under Drives in the sidebar. As such, I think this issue is not caused
by getting Google Drive paths from the registry in 
`GoogleDriveCloudDetector#GetProviders`. I would happy to investigate this 
issue, but maybe in a separate PR.

I still have logging in place in case you want to use it to confirm the current
behavior of everything. Let me know if you think it looks good and I'll make a
new commit removing all the logging.

Begin copied commit messages:

AddMyDriveToPathAndValidate; is_my_drive

Renamed `ValidatePath` to `AddMyDriveToPathAndValidate`, and made it also append
`My Drive' to the path it is given.

The only path we ever get from the media table is the drive letter 
(if `Google Drive streaming location` is set to `Drive letter`, otherwise we get nothing 
from the media table) followed by `:\`. As such, I added a call to 
`AddMyDriveToPathAndValidate' on this path.

We only find paths in the roots table whose `is_my_drive` is true when `My Drive
syncing options` is set to Mirror files. I believe these paths are the ones @Josh65-2201 
was referring to when he said 

> Google Drive set to mirror to a user selected folder will show both the 
> redirect drive/folder (This show one My Drive folder) and another that goes 
> directly to the store folder location. 

The path we get from roots is the one that "goes directly to the store folder 
location" (hence its `is_my_drive` is true).

With this PR, we now get the "redirect drive/folder" as well, from the registry.
As such, since Josh said

> Preferably it would only show the redirect drive/folder and add /My Drive to 
> the path, then hide all other locations

I added a call to `AddMyDriveToPathAndValidate` on every path we get from the
registry, and I made the roots table read loop `continue` if it finds a path 
whose `is_my_drive` is true.

my drive shortcut fix; registry logging

Handled the case where My Drive is a shortcut (when "My Drive syncing options" 
is set to "Mirror files").

Added logging for registry values.

src/Files.App/Files.App.csproj
yaira2, line 1: @wharvex can you revert these changes?

wharvex: Sorry about that, I wasn't sure how that commit ended up in the PR, I didn't make it. When I view it, it says "hishitetsu authored and wharvex committed". Do you want me to undo the commit that @hishitetsu authored? I did not edit any file other than `GoogleDriveCloudDetector.cs` in the commits that I authored.

0x5bfa: This committish was included when you force pushed. However, this is a reverted change so it's totally fine.

wharvex: Okay, and do you want me to revert the whole commit, or just the change to the csproj file?

0x5bfa: The whole commit, otherwise build gonna fail.

wharvex: Gotcha, thanks. I'll take care of it in a couple hours.

wharvex: Okay I just pushed the revert.

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -2,8 +2,11 @@
 // Licensed under the MIT License. See the LICENSE.
 
 using Microsoft.Data.Sqlite;
+using Microsoft.Extensions.Logging;
+using Microsoft.Win32;
 using System.IO;
 using Windows.Storage;
+using Vanara.Windows.Shell;
```
yaira2, line 9: @wharvex is this required here? We've been working to remove the Vanara package so hopefully we can find an alternative API.

wharvex: Ah okay, thanks for letting me know. I used that because I needed to get the target of a shortcut. When syncing is set to mirror, the "My Drive" folder in the streaming location is a shortcut, so I needed to get its target path for the CloudProvider SyncFolder. I tried to find how shortcut targets usually get resolved in the repo, and I could only find examples that used Vanara. I used `Files.App.Utils.Shell.ShellFolderExtensions.GetShellItemFromPathOrPIDL` which returns a Vanara `ShellItem`.

I'm working on an alternate method for getting the target (without dependencies), based on the approach I found [here](https://stackoverflow.com/a/77142277/16458003). I got a prototype working in a separate repo, I'm just cleaning it up and then I'll test it out in my Files fork. I'm hoping to make a push in the next couple days. Or let me know if you already have another preferred method of getting shortcut targets that does not use Vanara.

0x5bfa: > Or let me know if you already have another preferred method of getting shortcut targets that does not use Vanara.

You're trying to enumerate inner items in a IShellItem (a folder).

I'm working on removing Vanara, to support NativeAOT when WinAppSdk 1.6 available (Vanara contains ComWrappers which isn't compatible with IL Trimming and NativeAOT)
The replacement is CsWin32, generator of P/Invokes (this is compatible with it). 

However there's an issue with CsWin32, where it emits COM interfaces with methods that return void instead of HRESULT. But this can be resolved by not allowing marshaling in CsWin32's config file. We should be able to do that but currently blocked by #15834 (I'm removing it because it collides with our CsWin32 config and bc of package size reduction).

---

The conclusion is that you can use it for now. I'll replace it later (with IWindowsShellService.Enumerate() I guess).
I appreciate if you append a comment in the right above as:
`// TODO: Avoid to use Vanara (#15000)`

Thanks!

0x5bfa:
Is this ready for code quality assurance review?

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
wharvex: Okay no problem, I'll add that comment. And yes, just to clarify, in `AddMyDriveToPathAndValidate` I get an IShellItem for the streaming location and cast it to a ShellFolder, which allows me to get its contents as ShellItems. I then get "My Drive" from these items as a ShellLink and get its `TargetPath` property, which is what I ultimately needed. Unfortunately it's a little hard to follow in the code because I had to do it all in one line or else it gave me an exception that said something like "COM object that has been separated from its underlying RCW cannot be used."

When I add the comment, would you also like me to remove all the debug logging I added? I used this logging when I was testing things out, but left it in so whoever tries out the code can confirm that things are working as they should. I marked all these pieces of logging with a `// TESTING` comment.

wharvex:
@0x5bfa The only things I can think of that you might want me to change first are bringing the branch up-to-date with the parent main, and removing the `// TESTING` logging mentioned above. Please let me know if you'd like me to do these before the QA review.

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
0x5bfa: You can do this like

#ifdef DEBUG
    Debug.WriteLine("testing A");
    Debug.WriteLine("confirming B is...);
#endif

0x5bfa: And sorry, I meant to add the comment right above the code instead of the using.

wharvex: No worries. I just pushed these changes.

wharvex:
I just brought my branch up-to-date with the main Files branch. I also moved the Vanara comment and changed my file logging to `Debug.WriteLine` calls inside preprocessor checks for Debug mode. Let me know if you want me to change anything else.

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+#endif
+
+				yield return provider;
+			}
+		}
+
+		private async Task Inspect(SqliteConnection database, string sqlCommand, string contentsOf)
+		{
+			await using var cmdTablesAll =
+				new SqliteCommand(sqlCommand, database);
```
0x5bfa, line 146:
```suggestion
			await using var cmdTablesAll = new SqliteCommand(sqlCommand, database);
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+			Debug.WriteLine("BEGIN LOGGING of " + contentsOf + " contents");
+			while (reader.Read())
+			{
+				var colVals = new object[reader.FieldCount];
+				reader.GetValues(colVals);
+				colVals.Select((val, j) => $"row {i}: column {j}: {colNamesList[j]}: {val}").ToList()
+					.ForEach(s => Debug.WriteLine(s));
+				i++;
+			}
+			Debug.WriteLine("END LOGGING of " + contentsOf + " contents");
```
0x5bfa, line 159:
```suggestion

#if DEBUG
			Debug.WriteLine("END LOGGING of " + contentsOf + " contents");
#endif
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+
+            var googleDriveRegVal = googleDriveRegKey.GetValue(_googleDriveRegValName);
+
+            if (googleDriveRegVal is null)
+            {
+                _logger.LogWarning(
+                    "Google Drive registry value for value name `"
+                        + _googleDriveRegValName
+                        + "' not found."
+                );
```
0x5bfa, line 193:
```suggestion
                _logger.LogWarning($"Google Drive registry value for value name '{_googleDriveRegValName}' not found.");
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+            // This will be null if the key name is not found.
+            using var googleDriveRegKey = Registry.CurrentUser.OpenSubKey(_googleDriveRegKeyName);
+
+            if (googleDriveRegKey is null)
+            {
+                _logger.LogWarning(
+                    "Google Drive registry key for key name `"
+                        + _googleDriveRegKeyName
+                        + "' not found."
+                );
```
0x5bfa, line 181:
```suggestion
                _logger.LogWarning($"Google Drive registry value for value name '{_googleDriveRegValName}' not found.");
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+                googleDriveRegValueJson = JsonDocument.Parse(googleDriveRegVal.ToString() ?? "");
+            }
+            catch (JsonException je)
+            {
+                _logger.LogWarning(
+                    je,
+                    "Google Drive registry value for value name `"
+                        + _googleDriveRegValName
+                        + "' could not be parsed as a JsonDocument."
+                );
```
0x5bfa, line 209:
```suggestion
                _logger.LogWarning(je, $"Google Drive registry value for value name {_googleDriveRegValName} could not be parsed as a JsonDocument.");
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+					    out var googleDriveRegValPropProp))
+					continue;
+
+				var path = googleDriveRegValPropProp.GetString();
+				if (path is null)
+					continue;
+
+				if (!AddMyDriveToPathAndValidate(ref path))
+				{
+					_logger.LogWarning("Validation failed for " + path + " (media)");
```
0x5bfa, line 258:
```suggestion
					_logger.LogWarning($"Validation failed for {path} (media)");
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+			if (path.Length == 1)
+			{
+				DriveInfo temp;
+				try
+				{
+					temp = new DriveInfo(path);
+				}
+				catch (ArgumentException e)
+				{
+					_logger.LogWarning(e, "Could not resolve drive letter `" + path + "' to a valid drive.");
```
0x5bfa, line 301:
```suggestion
					_logger.LogWarning(e, $"Could not resolve drive letter '{path}' to a valid drive.");
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+			if (!string.IsNullOrEmpty(shellFolderBaseFirst))
+			{
+				path = shellFolderBaseFirst;
+				return true;
+			}
+
+			path = Path.Combine(path, "My Drive");
+			if (Directory.Exists(path))
+				return true;
+			_logger.LogWarning("Invalid Google Drive mount point path: " + path);
```
0x5bfa, line 333:
```suggestion

			_logger.LogWarning($"Invalid Google Drive mount point path: {path}");

```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+			// If `path` contains a shortcut named "My Drive", store its target in `shellFolderBaseFirst`.
+			// This happens when "My Drive syncing options" is set to "Mirror files".
+			// TODO: Avoid to use Vanara (#15000)
+			using var shellFolderBase = ShellFolderExtensions.GetShellItemFromPathOrPIDL(path) as ShellFolder;
+			var shellFolderBaseFirst = Environment.ExpandEnvironmentVariables((
+				                                                                  shellFolderBase?.FirstOrDefault(si =>
+					                                                                  si.Name?.Equals("My Drive") ??
+					                                                                  false) as ShellLink)
+			                                                                  ?.TargetPath ??
+			                                                                  "");
```
0x5bfa, line 317:
```suggestion
			var shellFolderBaseFirst = Environment.ExpandEnvironmentVariables((
				shellFolderBase?.FirstOrDefault(si =>
					si.Name?.Equals("My Drive") ?? false)?.TargetPath
				?? string.Empty);
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+			Debug.WriteLine(googleDriveRegValJsonProperty.ToString());
+#endif
+
+			foreach (var item in googleDriveRegValJsonProperty.Value.EnumerateArray())
+			{
+				if (!item.TryGetProperty(_googleDriveRegValPropName, out var googleDriveRegValProp))
+					continue;
+
+				if (!googleDriveRegValProp.TryGetProperty(_googleDriveRegValPropPropName,
+					    out var googleDriveRegValPropProp))
```
0x5bfa, line 249:
```suggestion
				if (!googleDriveRegValProp.TryGetProperty(_googleDriveRegValPropPropName, out var googleDriveRegValPropProp))
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+
+			// A default JsonProperty struct has an "Undefined" Value#ValueKind and throws an
+			// error if you try to call EnumerateArray on its Value.
+			if (googleDriveRegValJsonProperty.Value.ValueKind == JsonValueKind.Undefined)
+			{
+				_logger.LogWarning(
+					"Root element of Google Drive registry value for value name `"
+						+ _googleDriveRegValName
+						+ "' was empty."
+				);
```
0x5bfa, line 234:
```suggestion
				_logger.LogWarning($"Root element of Google Drive registry value for value name '{_googleDriveRegValName}' was empty.");

```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
 					IconData = iconFile is not null ? await iconFile.ToByteArrayAsync() : null,
 				};
 			}
+
+#if DEBUG
+			await Inspect(database, "SELECT * FROM roots", "root_preferences db, roots table");
+			await Inspect(database, "SELECT * FROM media", "root_preferences db, media table");
+			await Inspect(database, "SELECT name FROM sqlite_master WHERE type = 'table' ORDER BY 1",
+				"root_preferences db, all tables");
```
0x5bfa, line 128:
```suggestion
			await Inspect(database, "SELECT name FROM sqlite_master WHERE type = 'table' ORDER BY 1", "root_preferences db, all tables");
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -74,13 +95,23 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
-				string iconPath = Path.Combine(Environment.GetEnvironmentVariable("ProgramFiles"), "Google", "Drive File Stream", "drive_fs.ico");
 
 				App.AppModel.GoogleDrivePath = path;
 
-				StorageFile iconFile = await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(iconPath).AsTask());
+				StorageFile? iconFile = await GetGoogleDriveIconFileAsync();
+
+#if DEBUG
+				Debug.WriteLine("YIELD RETURNING from GoogleDriveCloudDetector#GetProviders (media): ");
+				Debug.WriteLine("name=" + title + "; path=" + path);
```
0x5bfa, line 113:
```suggestion
				Debug.WriteLine($"name={title}; path={path}");
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -74,13 +95,23 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
 				if (string.IsNullOrWhiteSpace(path))
 					continue;
 
+				if (!AddMyDriveToPathAndValidate(ref path))
+				{ 
+					_logger.LogWarning("Validation failed for " + path + " (media)");
```
0x5bfa, line 100:
```suggestion
					_logger.LogWarning($"Validation failed for {path} (media)");

```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -58,6 +74,11 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
 
 				App.AppModel.GoogleDrivePath = path;
 
+#if DEBUG
+				Debug.WriteLine("YIELD RETURNING from GoogleDriveCloudDetector#GetProviders (roots): ");
+				Debug.WriteLine("name=" + $"Google Drive ({title}); path=" + path);
```
0x5bfa, line 79:
```suggestion
				Debug.WriteLine("name=" + $"Google Drive ({title}); path={path}");
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -12,6 +15,16 @@ namespace Files.App.Utils.Cloud
...
 	{
+		private static readonly ILogger _logger = Ioc.Default.GetRequiredService<ILogger<App>>();
+
+        private const string _googleDriveRegKeyName = @"Software\Google\DriveFS";
+
+        private const string _googleDriveRegValName = "PerAccountPreferences";
+
+        private const string _googleDriveRegValPropName = "value";
+
+        private const string _googleDriveRegValPropPropName = "mount_point_path";
```
0x5bfa, line 26:
```suggestion
        private const string _googleDriveRegKeyName = @"Software\Google\DriveFS";
        private const string _googleDriveRegValName = "PerAccountPreferences";
        private const string _googleDriveRegValPropName = "value";
        private const string _googleDriveRegValPropPropName = "mount_point_path";
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -74,13 +95,23 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+				}
+
 				var folder = await StorageFolder.GetFolderFromPathAsync(path);
 				string title = reader["name"]?.ToString() ?? folder.Name;
-				string iconPath = Path.Combine(Environment.GetEnvironmentVariable("ProgramFiles"), "Google", "Drive File Stream", "drive_fs.ico");
 
 				App.AppModel.GoogleDrivePath = path;
 
-				StorageFile iconFile = await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(iconPath).AsTask());
+				StorageFile? iconFile = await GetGoogleDriveIconFileAsync();
```
0x5bfa, line 109: Is var more appropriate here?
```suggestion
				var iconFile = await GetGoogleDriveIconFileAsync();
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+			var i = 0;
+			Debug.WriteLine("BEGIN LOGGING of " + contentsOf + " contents");
+			while (reader.Read())
+			{
+				var colVals = new object[reader.FieldCount];
+				reader.GetValues(colVals);
+				colVals.Select((val, j) => $"row {i}: column {j}: {colNamesList[j]}: {val}").ToList()
+					.ForEach(s => Debug.WriteLine(s));
+				i++;
+			}
```
0x5bfa, line 158:
```suggestion

#if DEBUG
			Debug.WriteLine($"BEGIN LOGGING of {contentsOf} contents");
#ifdef

			if (int index = 0; reader.Read() is not false; index++)
			{
				var colVals = new object[reader.FieldCount];
				reader.GetValues(colVals);

				colVals.Select((val, j) => $"row {index}: column {j}: {colNamesList[j]}: {val}")
					.ToList().ForEach(s => Debug.WriteLine(s));
			}
```

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -89,6 +120,218 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
...
+			Debug.WriteLine("END LOGGING of " + contentsOf + " contents");
+		}
+
+		private ILogger GetAltLogger(string filename)
```
0x5bfa, line 162: Could you please explain me why another logger necessary?

0x5bfa (changes requested):
Mostly looks good, please solve those QA issues.

Snd most importantly, please format changed files with tabs instead of spaces.
You can do that from the bottom of the VS editor (I think the formatting box says MIXED but please make sure it's TABS)

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
wharvex: Since I switched to Debug.WriteLine calls, this method is no longer necessary. I'll remove it.

wharvex:
@0x5bfa thanks for the review. I just pushed a commit that addresses all of your suggested changes. I hope this was the right way to do this, sorry I don't have much experience with PRs and code reviews. Please let me know if I should have done it differently or if you want me to do anything else.

wharvex:
Would you like me to click "Resolve conversation" on the reviews that I've addressed in the last couple commits?

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
0x5bfa: This is missing.

0x5bfa:
Yeah, and one thing is missed (while to for change).

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
wharvex: Thanks, somehow I missed the loop change. I just pushed a commit that changes it to a `for` loop.

0x5bfa (approved):
LGTM, code wise.

wharvex:
Great, thanks! I noticed one of the tests failed though when I pushed. The error message says:
`Element Button(50000) violated rule 'Focusable sibling elements must not have the same Name and LocalizedControlType.'.`
It looks like it's coming from an accessibility testing tool called Axe. Do you know if this is related to my code? I can try to debug it if you think it is.

wharvex:
It looks like the `AxeHelper.AssertNoAccessibilityErrors` assertion in `InteractionTests.Tests.GeneralAppTests` failed. I ran this test (along with the other Interaction tests) on my computer and they passed (see below).

I also noticed there are two calls to `AxeHelper.AssertNoAccessibilityErrors` in GeneralAppTests, and the first call is wrapped in a try-catch with a comment above it that says `First run can be flaky due to external components`. Maybe this has something to do with it? I could try pushing again with some inconsequential change to see if it fails again.

interaction_tests

0x5bfa:
@yaira2 can return the CI, or you can squash the latest two commits and force push.

wharvex:
Awesome, it looks like they're all passing now, thanks. I just squashed and force-pushed.

yaira2:
@0x5bfa @Josh65-2201 is this good from your end?

Josh65-2201:
Right now I'm still seeing the issue of Google Drive showing in the Drives section on the sidebar and home page when mounted using drive letter in Google Drives settings. Folder location is correct in all modes under the cloud section.

wharvex:
@Josh65-2201 Thanks for reminding me of this issue. I just pushed a commit that seems to fix it on my end. Please let me know what you think.

I realized the issue has to do with how we're setting/reading from `App.AppModel.GoogleDrivePath`.

In `GoogleDriveCloudDetector.GetProviders()`, we're setting `App.AppModel.GoogleDrivePath` right before each place where we "yield return" a CloudProvider, and we're setting it to the CloudProvider's `SyncFolder` path each time.

In `Files.App.Services.RemovableDrivesService.GetDrivesAsync()`, we're then checking the contents of `App.AppModel.GoogleDrivePath` against each DriveInfo instance returned from `System.IO.DriveInfo.GetDrives()`, ideally to filter out Google Drive from the potential adds to a "Drives" section.

However, `Files.App.Services.RemovableDrivesService.GetDrivesAsync()` actually gets called *before* `GoogleDriveCloudDetector.GetProviders()`, so the contents of `App.AppModel.GoogleDrivePath` is an empty string when the former is called, causing the Google Drive filtering attempt to fail.

My solution was to make my `GetGoogleDriveProvidersFromRegistryAsync` method public and static, and just enumerate over its yield returns directly in `Files.App.Services.RemovableDrivesService.GetDrivesAsync()` to filter out a `G:\\` path. I made a few minor adjustments to make this possible.

Obviously this introduces some redundancies/inefficiencies, but it seems like a viable quick fix for now, but let me know if you'd prefer a different approach.

Josh65-2201 (dismissed):
Working as expected now

wharvex:
I also wanted to mention that the call to `RemovableDrivesService.GetDrivesAsync()` that happens before the cloud drives get detected can be traced back to `Files.App.ViewModels.MainPageViewModel.OnNavigatedToAsync`, as you can see in this screen shot of my call stack when I was debugging.

callstack

wharvex:
> Working as expected now

Great, thanks!

src/Files.App/Services/Storage/StorageDevicesService.cs
```diff
@@ -22,8 +22,20 @@ public async IAsyncEnumerable<ILocatableFolder> GetDrivesAsync()
...
 			var pCloudDrivePath = App.AppModel.PCloudDrivePath;
 
+#if DEBUG
+			Debug.WriteLine($"In RDS.GDA: googleDrivePath: {googleDrivePath}");
+#endif
+
 			foreach (var drive in list)
 			{
+				var shouldSkip = false;
+				await foreach (var cloudProvider in GoogleDriveCloudDetector.GetGoogleDriveProvidersFromRegistryAsync(false))
```
yaira2, line 32: I’m a bit worried about the potential performance impact this change might have. That’s actually one of the reasons we initially didn’t display Google Drive in the cloud section.

wharvex: It looks like that code takes about a quarter of a second to run on my machine in debug mode, and it only runs once. I just pushed a commit with some timing in place so you can experiment with it if you want.

Output of `Get-ComputerInfo -Property "CsProscessors"` on my machine, for reference:

```
CsProcessors
------------
{Intel(R) Core(TM) i7-8700T CPU @ 2.40GHz}
```

Screenshot of the timing:

timing

yaira2: Doesn't it depend on the number of drives?

wharvex: Trial 1:

In RDS.GDA after registry check: Time elapsed for filter: 00:00:01.7072234
In RDS.GDA: drive.Name: C:\

In RDS.GDA after registry check: Time elapsed for filter: 00:00:00.6069444
In RDS.GDA: drive.Name: G:\

Trial 2:

In RDS.GDA after registry check: Time elapsed for filter: 00:00:01.6830505
In RDS.GDA: drive.Name: C:\

In RDS.GDA after registry check: Time elapsed for filter: 00:00:00.4277025
In RDS.GDA: drive.Name: G:\

Trial 3:

In RDS.GDA after registry check: Time elapsed for filter: 00:00:02.0122619
In RDS.GDA: drive.Name: C:\

In RDS.GDA after registry check: Time elapsed for filter: 00:00:00.8921746
In RDS.GDA: drive.Name: G:\

Trial 4:

In RDS.GDA after registry check: Time elapsed for filter: 00:00:01.9693083
In RDS.GDA: drive.Name: C:\

In RDS.GDA after registry check: Time elapsed for filter: 00:00:00.3338732
In RDS.GDA: drive.Name: G:\

Not great! I'll look into how I can improve this.

wharvex: > Doesn't it depend on the number of drives?

Yes it seems like it does. I was mistaken at first.

yaira2: Another option is to remove the drives after the app is fully loaded, but this is not ideal.

wharvex: I just pushed a change that improves the performance of that registry check quite a bit, please let me know if it looks alright. Sorry there are so many changes. I basically just split up some of the methods I wrote so I could reuse the smallest essential part of the code that gets the registry path.

Here are the trials I ran, all in Debug mode:

Trial 1

In RemovableDrivesService: Time elapsed for registry check: 00:00:00.0770886

Trial 2

In RemovableDrivesService: Time elapsed for registry check: 00:00:00.0609413

Trial 3

In RemovableDrivesService: Time elapsed for registry check: 00:00:00.0727057

Trial 4

In RemovableDrivesService: Time elapsed for registry check: 00:00:00.0918816

wharvex: Also it only checks the registry once now. I set `App.AppModel.GoogleDrivePath` in `RemovableDrivesService` now, since that's where we need the path first. And then GoogleDriveCloudDetector reads from there, since that gets called second.

yaira2: This looks a lot better

wharvex: Excellent, thanks.

wharvex: Should I click "Resolve conversation" on here?

wharvex:
I just brought the fork up-to-date with the main repo and made a few minor formatting changes. The CI was failing at first, it was that same issue with the Axe accessibility test with sibling elements, and then there was an x64 Release build fail for some reason (it wouldn't say why), but now everything seems good.

Please let me know if you want me to change anything else.

0x5bfa (approved):
Code wise, LGTM.
I haven't checked behavior yet.

Josh65-2201 (approved):
Still works as expected since my last review

wharvex:
Thanks for checking.

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
```diff
@@ -56,7 +66,10 @@ await FilesystemTasks.Wrap(() => StorageFile.GetFileFromPathAsync(Path.Combine(a
 				var folder = await StorageFolder.GetFolderFromPathAsync(path);
 				string title = reader["title"]?.ToString() ?? folder.Name;
 
-				App.AppModel.GoogleDrivePath = path;
+#if DEBUG
+				Debug.WriteLine("YIELD RETURNING from `GoogleDriveCloudDetector.GetProviders()` (roots): ");
+				Debug.WriteLine($"Name: Google Drive ({title}); SyncFolder: {path}");
+#endif
```
yaira2, line 72: Do we still need to output this data?

wharvex:
Is there anything I can do to help with testing? I made this screen recording of a test I did just now running the repo with my changes under four different config conditions, just to document how I ran it and what I looked for.

https://drive.google.com/file/d/1bUYS0T3HQcxmzROW_wlNlP8zQjxze035/view?usp=sharing

src/Files.App/Utils/Cloud/Detector/GoogleDriveCloudDetector.cs
wharvex: I can remove it, I just had it in there to help with testing. Should I remove all of the `Debug.WriteLine` calls and their related code?

yaira2: It wouldn't hurt to keep it, but I don't think we need to wrap it in the `#if DEBUG` statement.

wharvex: Sure I can remove the `#if debug` statement there. Although @0x5bfa told me to add those in his review, just to let you know. Maybe he was thinking of the potential performance hit the `Debug.WriteLine` calls could cause in release.

There are 7 `#if debug`s in GoogleDriveCloudDetector and one in StorageDevicesService. Do you want me to remove all of them?

0x5bfa: I think we should.
On release builds, those lines shouldn't be run and you can't read unless you debug it via VS or WinDbg without source code.

yaira2: That would be helpful, thanks!

wharvex: No problem, I just pushed the change.

0x5bfa: I meant we should keep they as are.

PR #15773, opened 2024-07-07
Code Quality: Introduced IStorageTrashBinService

0x5bfa:
### Resolved / Related Issues

- https://github.com/files-community/Files/issues/14572
- https://github.com/files-community/Files/issues/15000

1. Open Files
2. Go to the Recycle Bin
3. Empty the bin or Restore the bin
4. Move a file to the bin


src/Files.App/Services/Windows/WindowsRecycleBinService.cs
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		/// <inheritdoc/>
+		public bool RestoreAll()
+		{
+			return true;
+		}
+
+		/// <inheritdoc/>
+		public void StartWatcher()
```
0x5bfa, line 97: @hishitetsu This service implementation is the very first time to use IWather. I’m feeling this is odd because StartWatcher and StopWatcher is public and you can call them via the interface.
Do you think I should create `RecycleBinWatcher : ITrashWatcher` and create a property of it in the service?
When you want to hook up the addition event of it yo have to access via `IWindowsRecycleBinService.Watcher.ItemAdded` for example.
This also can be said to IWindowsLibraryService and IStorageDevicesService and IMutableFolder (storage abstraction, you can ignore for now) in the future, such as `LibraryFolderWatcher/ShellFolderWatcher : IFolderWatcher` and `DeviceWatcher : IDeviceWatcher`.

0x5bfa:
This is requiring a shell operation and I'm looking further since this can be used for other Vanara references including code porting such as ShellFileOperations and ShellFileOperations2.

0x5bfa:
Should be something like this

```c#
public abstract ShellStorable : ...
{
    public IShellItem item { get; set; }

    protected NativeStorable(string path);

    // Same as Environment.SpecialFolder enum but it's not complete enum
    protected NativeStorable(KNOWN_FOLDER_ID knownFolderId);
}
```

```c#
public class NativeFile/NativeFolder : ShellStorable, ...
{
    CopyAsync(IStorageOperationProgress handler, uint flags, CancellationToken token);
    DeleteAsync(IStorageOperationProgress handler, uint flags); // Can recycle with FILE_OP.ALLOW_UNDO
    MoveAsync(IStorageOperationProgress handler, uint flags);
}
```

```c#
public class ShellOperationProgress : IStorageOperationProgress
{
    public event EventHandler<StorageOperationEventArgs>? FinishedAll;

    public event EventHandler<StorageOperationEventArgs>? FinishedPer;

    public event EventHandler<StorageOperationEventArgs>? BeginAll;

    public event EventHandler<StorageOperationEventArgs>? BeginPer;

    // This will be given from caller which is instanced by StatusCenterItem. If progress report isn't needed this can be null.
    public IProgress? Progress { get; set; }
}
```

0x5bfa:
Tested locally (especially restore all), ready for review and merge.

src/Files.App/Actions/FileSystem/EmptyRecycleBinAction.cs
```diff
@@ -30,7 +36,31 @@ public EmptyRecycleBinAction()
 
 		public async Task ExecuteAsync(object? parameter = null)
 		{
-			await RecycleBinHelpers.EmptyRecycleBinAsync();
+			// TODO: Use AppDialogService
+			var confirmationDialog = new ContentDialog()
```
yaira2, line 40: Can you set the dialog root?

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
yaira2, line 1: Can you rename to `IRecycleBinService`?

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
```diff
@@ -0,0 +1,67 @@
...
+		/// </summary>
+		/// <returns></returns>
+		bool HasItems();
+
+		/// <summary>
+		/// Gets the file or folder specified is already moved to Recycle Bin.
+		/// </summary>
+		/// <param name="path">The path that indicates to a file or folder.</param>
+		/// <returns>True if the file or path is recycled; otherwise, false.</returns>
+		bool IsRecycled(string? path);
```
yaira2, line 46:
```suggestion
		bool IsPathUnderRecycleBin(string? path);
```

0x5bfa: why that should be that long?

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
0x5bfa: image

yaira2: The implementation name can be platform specific, but the interface should be generic.

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
yaira2: `IsRecycled` is vague

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
0x5bfa: I still not sure that this interface can be platform-independent.

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
0x5bfa: While this method is in Windows**RecycleBin**Service?

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
0x5bfa: Also everything in Files.App is not platform-dependent. Otherwise, should go Files.Core.*

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
yaira2: `IsRecycled` sounds likes it's fully deleted. Another option is `IsInRecycleBin`.

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
yaira2: > I still not sure that this interface can be platform-independent.

We'll cross that road if and when we get there, but for now, the name should be generic.

0x5bfa: I mean this is in Files.App project.
What's the reason 'now' we do that while you are not sure? I did it for ITrashWatcher because I was sure it uses platform independent classes.

In my opinion, I don't want to make things ambiguous. For now IWindowsRecycleBinService is better with me.

---

Or, IStorageTrashService is good.

src/Files.App/Data/Contracts/IWindowsRecycleBinService.cs
0x5bfa: The same problem persists as above.
Recycle Bin is the name only on Windows.
If you think platform independence, we should avoid that naming. IsInBin or something sounds better to me.

---

Sorry for taking time for this tiny things but I believe discussing naming is really important to me as I try to keep overlooking Files architecture to avoid ambiguity or misunderstanding.
Thanks!

src/Files.Core.Storage/Contracts/ITrashWatcher.cs
```diff
@@ -15,11 +15,6 @@ public interface ITrashWatcher : IWatcher
 		/// </summary>
 		event EventHandler<SystemIO.FileSystemEventArgs>? ItemDeleted;
 
-		/// <summary>
-		/// Gets invoked when an item changing is detected by the watcher
-		/// </summary>
-		event EventHandler<SystemIO.FileSystemEventArgs>? ItemChanged;
```
yaira2, line 21: Why was this removed?

0x5bfa: I'm not sure why added.
In fact, it isn't needed.
But maybe needed in other platform. Should we keep?

yaira2: Let's keep it

0x5bfa: Done

PR #15788, opened 2024-07-10
GitHub: Remove artifacts older than 24 hours automatically

0x5bfa:
### Resolved / Related Issues

*Planned internally*

Trigger the CI in the main branch later

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,6 +237,12 @@ jobs:
           $env:AUTOMATED_TESTS_ASSEMBLY_DIR\**\Files.InteractionTests.dll `
           --logger "trx;LogFileName=$env:AUTOMATED_TESTS_ASSEMBLY_DIR\testResults.trx"
 
+    - name: Remove artifacts
+      uses: c-hive/gha-remove-artifacts@v1.2.0
+      with:
+        age: 1 day
+        GITHUB_TOKEN: ${{ github.token }}
+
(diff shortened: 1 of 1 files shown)


0x5bfa:
This is failing because deleting all artifacts result in a massive request of GitHub API. @yaira2 do you already have an access token or can you create an access token in this organization settings?

yaira2:
Is there a benefit to this over setting the retention?

0x5bfa:
Didn’t say the organization is running out of Actions storage pool?
If setting Retention Setting to 1 day or something and everything is gonna be fine, it’s fine with me but I thought you want to remove artifacts published within 10 days (you set it to 15 days previously and now 5 days).
if we have this we don’t think about storage anymore when we are uploading other things in the future tho.

yaira2:
> Shared storage consists of Actions artifacts and Packages usage. This graph shows the account's storage usage in GB-months. Removing stored artifacts will not reduce this number, but it will lower its rate of growth. To see your account's current storage, download a usage report.

It doesn't seem that changing the retention or deleting the existing artifacts will make a difference. These are our options
- Find a way to pass the package to the testing job
- Require manually triggering the action
- Merge the testing job together with the build job

0x5bfa:
Sigh, I said it’s on the docs. Deleting of course makes difference since it free up spaces.
I don’t recommend to merge two jobs btw

PR #15795, opened 2024-07-13
Code Quality: Disable drag & drop when running as admin

### LINKED ISSUE #13394: Bug: App crashes when dragging a file while app is running with a high integrity level

0x5bfa:
### Summary

The service 'IWindowsSecurityService' gets the current process's token and checks if the token belongs to Administrators group.

- [x] Create `IWindowsSecurityService`
- [x] Replace ElevationHelpers with the service above
- [x] Cache the result in `WindowContext`

### Resolved / Related Issues

- Closes #13394

1. Open Files app as admin
2. Confirm drag and drop is disabled in TabBar and content views.
3. Check it also when UAC is disabled.

<details><summary>Open details that I learnt</summary><div>

### Why drag and drop doesn’t work?

This is because of UIPI (User Interface Privilege Isolation) using MIC (Mandatory Integrity Control), which blocks a process that has lower IL (integrity level) from interacting with a process that has higher IL. This was introduced in Vista with UAC introduction.

It is not caused by UIPI, btw, that UWP cannot access system resource and user data without user’s consent (it’s by AppContainer itself).

IL get higher from top to bottom

1. `Untrusted` (Chrome, IE)
2. `AppContainer/LowBox` (UWP)
4. `Low`
5. `Medium` (default)
6. `Medium Plus`
7. `High/Elevated` (elevated)
8. `System` (system services)
9. `Protected Process`
10. `Installer` (Windows)

### UAC settings

**Behavior**

Level|Modification<br/>by a process|Modification<br/>by user|On what session|Remarks
:-:|---|---|---|---
Highest|Yes|Yes|Secure Desktop (dimmed)|Default in Vista
2nd|Yes|No|Secure Desktop (dimmed)|Default in 7 onwards
3rd|Yes|No|Normal Desktop (not dimmed)|Not recommended
Lowest|No|No|N/A|Not recommended

- By process means: Installation of a software manifest of which calls for elevation, and invocation via RunAs verb
- By user means: modification via Windows Settings app and certain Control Panel applets

**Registry values to be modified**

Level|ConsentPrompt<br/>BehaviorAdmin|ConsentPrompt<br/>BehaviorUser|EnableLUA|PromptOn<br/>SecureDesktop
:-:|---|---|---|---
Highest|2 (AAC UAC)|3 (OTS UAC)|1|1
2nd|5|3|1|1
3rd|5|3|1|0
Lowest|0|3|0|0

- AAC (Admin Approval Consent) UAC displays only Yes or No since logon user is in Administrators group.
- OTS (Over The Shoulder) UAC displays text boxes to enter credential of an admin identity.

### Disabling UAC

When you disabled UAC through UAC Settings dialog (UIPI will be disabled as well), the UAC prompt won’t be shown.

However, it seems that DataExchangeHost.exe doesn’t accept dropping onto a window of a process that has higher IL (Windows OS Bug?) even though UAC is disabled. A contributor of Windows Terminal made a workaround for it.
Also, I should denote that we might be able to use DragQueryFile to workaround UIPI, while I’m still not sure.

### How UAC works (if it helps)?

On startup, the kernel executes explorer.exe with a token below:

**Standard User (not in Administrators group)**

- Only *restricted* token will be generated after logon.
- ShellExecute calls appinfo.exe (AIS) and then it calls consent.exe (UAC prompt) via its hosted inner service. When a valid credential of an administrator entered, AIS calls CreateProcessAsUser with that administrative identity; otherwise, it returns ERROR_ACCESS_DENIED.
- The token generated by this routine is called *filtered* token (not *privileged* token)

**Administrator User (in Administrators group)**

- Both *privileged* and *restricted* token will be generated.
- When Yes clicked, AIS calls CreateProcessAsUser with the logon user ‘s *privileged* token; otherwise, it returns ERROR_ACCESS_DENIED.

**BUILTIN/Administrator (the real administrator)**

- Only *privileged* token will be generated
- AIS doesn’t call consent.exe. But this behavior can be altered from group policy.

---

I've been described so far, for when user hasn't changed anything through Windows Registry or Group Policy. Some behavior can be changed otherwise.

</div></details>


yaira2:
Can you cache the result?

yaira2:
For #13394

0x5bfa:
Yeah do you think would it be better in WindowContext?

yaira2:
I would keep the code in the service, but we can cache it in the context.

0x5bfa:
@yaira2 I found that there's a function ChangeWindowMessageFilterEx to change the filtering level. In theory this workarounds UIPI behavior, but may expose some security concerns such as a process that has lower IL can run arbitrary code by injection code to DLL. This is maybe why Windows Terminal didn't do this (or maybe they didn't know this).

yaira2:
I would prefer to do the same as Terminal.

yaira2:
FYI @ahmed605

ahmed605:
> IL get higher from top to bottom

this order is kinda wrong, AC has higher IL than Untrusted, and it's actually Low IL but with lower TrustLevel than usual Low IL apps, and there's also LPAC (Less Privileged AppContainer) which is almost on the same level as Untrusted but a little bit higher, it's used by Chromium, Firefox, and Microsoft Edge Legacy

0x5bfa:
I was really uncertain in that point. I'm even not sure this is IL. But a docs i referred (not official) wrote as AppContainer < Untrusted. Thank you for the correction.

What about that function above to workaround this blocking?

ahmed605:
> What about that function above to workaround this blocking?
I'm a bit worried about the security concerns, you can still inject through only the HWNDs btw

0x5bfa:
I see, thank you for letting me know it.
I'll just check token type and disable drag and drop accordingly.

0x5bfa:
Looks like SettingsButton is inaccessible.

yaira2:
Is this ready for review?

0x5bfa:
I'll cache this in the context class.

0x5bfa:
FULLY TESTED.
100% WORKING.

yaira2:
Do we need to adjust the text in the "running as admin" prompt?

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -3267,7 +3267,7 @@
     <value>Files is running as administrator</value>
   </data>
   <data name="FilesRunningAsAdminContent" xml:space="preserve">
-    <value>Due to a limitation with Windows and WinAppSdk, drag and drop isn't available when running Files as admin. If you wish to use drag and drop, you can work around this limitation by opening UAC (User Account Control) from the Start Menu and selecting the third level, and restarting Windows. Otherwise, you can keep using Files without drag &amp; drop.</value>
+    <value>Due to a limitation with Windows, drag &amp; drop isn't available when running Files as admin. If you wish to use drag and drop, please restart Files without running as admin.</value>
```
0x5bfa, line 3270:
```suggestion
    <value>Due to a limitation with Windows, drag &amp; drop isn't available when running Files as admin or UAC is fully disabled. If you wish to use drag and drop, please restart Files without running as admin or enable UAC again.</value>
```

0x5bfa:
Would we like to remove the commented out code lines?

0x5bfa:
Ready again.

0x5bfa:
All good.

yaira2:
@0x5bfa I rebased the branch from `main`, can you confirm that everything is in order?

0x5bfa:
Yes, except that test failing

PR #15796, opened 2024-07-13
Code Quality: Fixed ArgumentNullException in ShellViewModel.CheckForBackgroundImage()

hishitetsu:
**Resolved / Related Issues**

- Closes #

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -1738,9 +1738,9 @@ namespace Files.App.ViewModels
 
 						await OrderFilesAndFoldersAsync();
 						await ApplyFilesAndFoldersChangesAsync();
+						await dispatcherQueue.EnqueueOrInvokeAsync(CheckForSolutionFile, Microsoft.UI.Dispatching.DispatcherQueuePriority.Low);
 						await dispatcherQueue.EnqueueOrInvokeAsync(() =>
 						{
-							CheckForSolutionFile();
 							GetDesktopIniFileData();
 							CheckForBackgroundImage();
 						},
@@ -1813,7 +1813,7 @@ namespace Files.App.ViewModels
 
 		public void CheckForBackgroundImage()
 		{
-			var filesAppSection = DesktopIni.FirstOrDefault(x => x.SectionName == "FilesApp");
+			var filesAppSection = DesktopIni?.FirstOrDefault(x => x.SectionName == "FilesApp");
 			if (filesAppSection is null || folderSettings.LayoutMode is FolderLayoutModes.ColumnView)
 			{
 				FolderBackgroundImageSource = null;


src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -1813,7 +1813,7 @@ private void GetDesktopIniFileData()
 
 		public void CheckForBackgroundImage()
 		{
-			var filesAppSection = DesktopIni.FirstOrDefault(x => x.SectionName == "FilesApp");
+			var filesAppSection = DesktopIni?.FirstOrDefault(x => x.SectionName == "FilesApp");
```
hishitetsu, line 1816: @0x5bfa I found that `LayoutModeChangeRequested()` might call this method before desktop.ini was loaded, so I added a null-safe operator to ignore the call.

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -1738,9 +1738,9 @@ private async Task<int> EnumerateItemsFromStandardFolderAsync(string path, Cance
 
 						await OrderFilesAndFoldersAsync();
 						await ApplyFilesAndFoldersChangesAsync();
+						await dispatcherQueue.EnqueueOrInvokeAsync(CheckForSolutionFile, Microsoft.UI.Dispatching.DispatcherQueuePriority.Low);
```
hishitetsu, line 1741: CheckForSolutionFile is not related to GetDesktopIniFileData and CheckForBackgroundImage, so I separated the queuing.

src/Files.App/ViewModels/ShellViewModel.cs
0x5bfa: Should we add the loader inside this method? This doesn't make much sense since even if ini file  exists you'll have to reload until it's loaded

hishitetsu: EnumerateItemsFromStandardFolderAsync should always load desktop.ini when the folder is successfully loaded, so I don't think it is necessary to load it every time.

0x5bfa: Correct me if I'm wrong

once this desktop background check method called ahead of time, the background is still null. So unless you don't refresh and pray that desktop ini data is loaded first, the background won't be loaded.

hishitetsu: CheckForBackgroundImage is called without loaded desktop.ini if LayoutModeChangeRequested is called before the folder load is completed or after the folder load fails. But the former will always call GetDesktopIniFileData and CheckForBackgroundImage in EnumerateItemsFromStandardFolderAsync, and the latter is an error state and can be ignored.

0x5bfa: Dang. Should we set a setter to call CheckForBackgroundImage in the property?

hishitetsu: That's one way. I don't have a preference.

mdtauk: If the Folder background has a fade in transition, then you can load the folder view, and once the contents are done loading, you can then check and fade in the folder image, if there is an ini file.  Lazy load essentially.

yaira2 (approved):
LGTM

PR #15808, opened 2024-07-16
Fix: Fixed FileNotFoundException in Recycle Bin watcher

yaira2:
**Resolved / Related Issues**

- https://files-org.sentry.io/issues/5610462140/events/0d15202895df4621bcde5b01ed4bc2e6/

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Utils/RecycleBin/RecycleBinManager.cs
+++ b/src/Files.App/Utils/RecycleBin/RecycleBinManager.cs
@@ -51,18 +51,21 @@ namespace Files.App.Utils.RecycleBin
 				if (drive.DriveType == SystemIO.DriveType.Network || !SystemIO.Directory.Exists(recyclePath))
 					continue;
 
-				SystemIO.FileSystemWatcher watcher = new()
+				SafetyExtensions.IgnoreExceptions(() =>
 				{
-					Path = recyclePath,
-					Filter = "*.*",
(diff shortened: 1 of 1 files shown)


0x5bfa (commented):
Only FileSystemWatcher.EnableRaisingEvents produces this exception when the path isn't locatable by CreateFile Win32 API, what about wrapping just it with SafetyExtention?

- https://learn.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.enableraisingevents?view=net-8.0
- https://github.com/dotnet/runtime/blob/9565550164fd63f1612141d221e04747d9742d4b/src/libraries/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Win32.cs#L42

hishitetsu (approved):
LGTM

yaira2:
@0x5bfa I don't think wrapping `EnableRaisingEvents` will help if the error doesn't occur until the watcher is already started.

0x5bfa:
Why not check before instancing it.

yaira2:
We check on line 51 but that doesn't seem to help.

0x5bfa:
This method doesn't work sometimes lol I'm wondering why.
It produces not found exception when CreateFile fails. Do you think we should have our own checker using that function?

```c#
// bool StorageHelper.IfExists(string path);
HANDLE hFile = CreateFileW(path, ...);
if (hFile == -1) return false;
```

yaira2:
It's something to look into

PR #15834, opened 2024-07-21
Code Quality: Removed WinUIEx dependency

### LINKED ISSUE #12358: Feature: remove WinuiEx dependency (when possible)

Mica is now (easily) available in the latest WASDK.

### Requirements

Pretty UI? :)

### Files Version

2.4.71

### Windows Version

Windows 11

### Comments

https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.media.micabackdrop?view=windows-app-sdk-1.3

```xaml
<Window x:Class="MyApp.MainWindow">

    <Window.SystemBackdrop>
        <MicaBackdrop MicaKind="BaseAlt"/>
    </Window.SystemBackdrop>

</Window>
```

I've heard there were issues with v1.3.0, so you might want to go up to 1.3.1

0x5bfa:
### Summary

As WinAppSdk has been improved, thanks to its team, we have reached at a point where we can move to WinAppSdk with a little Win32 API. Shout out to outgoing WinUIEx, which has been giving us the fundamental and powerful features since our initial migration for almost two years.

Note to someone who are looking for the same achievement:
- **Backdrop**: replaced with WinAppSdk
- **Window Size**: replaced with WinAppSdk
- **IsMaximizable/IsMinimizable**: replaced with WinAppSdk
- **MinWidth/MinHeight**: Implemented custom WndProc via CsWin32 using MINMAXSIZEINFO in WindowEx
- **Last window placement**: Saved/restored to ApplicationDataContainer (WINDOWPLACEMENT as base64) in WindowEx

### Resolved / Related Issues

- Closes #12358

1. Open Files app
2. Resize to be minimum (see if min size info is set)
3. Terminate the app
4. Lunch the app again (see if the last window placement is kept)
5. Open properties window of a file (see if you can resize but can't maximize)
6. See caption buttons of both windows have the correct brush now and even after you switch theme
7. Enable 'Automatically hide taskbar' in Windows settings
8. Enter F11 on the app
9. Try to hover over on the bottom of the monitor (see taskbar fades in)

### DIFF
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -6,11 +6,9 @@ using Microsoft.UI;
 using Microsoft.UI.Windowing;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media.Animation;
-using Microsoft.UI.Xaml.Navigation;
 using System.Runtime.InteropServices;
 using Windows.ApplicationModel.Activation;
 using Windows.Storage;
-using WinUIEx;
 using IO = System.IO;
 
 namespace Files.App
@@ -20,39 +18,17 @@ namespace Files.App
 		private static MainWindow? _Instance;
 		public static MainWindow Instance => _Instance ??= new();
 
-		public IntPtr WindowHandle { get; }
-
-		private MainWindow()
+		public MainWindow() : base(minWidth: 516, minHeight: 416)
 		{
-			WindowHandle = this.GetWindowHandle();
-
 			InitializeComponent();
 
-			EnsureEarlyWindow();
-		}
-
-		private void EnsureEarlyWindow()
-		{
-			// Set PersistenceId
-			PersistenceId = "FilesMainWindow";
-
-			// Set minimum sizes
-			MinHeight = 416;
-			MinWidth = 516;
-
-			AppWindow.Title = "Files";
-			AppWindow.SetIcon(AppLifecycleHelper.AppIconPath);
-			AppWindow.TitleBar.ExtendsContentIntoTitleBar = true;
+			ExtendsContentIntoTitleBar = true;
+			Title = "Files";
 			AppWindow.TitleBar.ButtonBackgroundColor = Colors.Transparent;
 			AppWindow.TitleBar.ButtonInactiveBackgroundColor = Colors.Transparent;
-
-			// Workaround for full screen window messing up the taskbar
-			// https://github.com/microsoft/microsoft-ui-xaml/issues/8431
-			// This property should only be set if the "Automatically hide the taskbar" in Windows 11,
-			// or "Automatically hide the taskbar in desktop mode" in Windows 10 is enabled.
-			// Setting this property when the setting is disabled will result in the taskbar overlapping the application
-			if (AppLifecycleHelper.IsAutoHideTaskbarEnabled()) 
-				Win32PInvoke.SetPropW(WindowHandle, "NonRudeHWND", new IntPtr(1));
+			AppWindow.TitleBar.ButtonPressedBackgroundColor = Colors.Transparent;
+			AppWindow.TitleBar.ButtonHoverBackgroundColor = Colors.Transparent;
+			AppWindow.SetIcon(AppLifecycleHelper.AppIconPath);
 		}
 
 		public void ShowSplashScreen()
@@ -209,8 +185,9 @@ namespace Files.App
 				Activate();
 			}
 
-			if (Windows.Win32.PInvoke.IsIconic(new(WindowHandle)))
-				Instance.Restore(); // Restore window if minimized
+			if (Windows.Win32.PInvoke.IsIconic(new(WindowHandle)) &&
+				AppWindow.Presenter is OverlappedPresenter presenter)
+				presenter.Restore(); // Restore window if minimized
 		}
 
 		private Frame? EnsureWindowIsInitialized()
@@ -224,7 +201,10 @@ namespace Files.App
 				{
 					// Create a Frame to act as the navigation context and navigate to the first page
 					rootFrame = new() { CacheSize = 1 };
-					rootFrame.NavigationFailed += OnNavigationFailed;
+					rootFrame.NavigationFailed += (s, e) =>
+					{
+						throw new Exception("Failed to load Page " + e.SourcePageType.FullName);
+					};
 
 					// Place the frame in the current Window
 					Instance.Content = rootFrame;
@@ -237,14 +217,6 @@ namespace Files.App
 				return null;
 			}
 		}
-
-		/// <summary>
-		/// Invoked when Navigation to a certain page fails
-		/// </summary>
-		/// <param name="sender">The Frame which failed navigation</param>
-		/// <param name="e">Details about the navigation failure</param>
-		private void OnNavigationFailed(object sender, NavigationFailedEventArgs e)
-			=> throw new Exception("Failed to load Page " + e.SourcePageType.FullName);
 
 		private async Task InitializeFromCmdLineArgsAsync(Frame rootFrame, ParsedCommands parsedCommands, string activationPath = "")
 		{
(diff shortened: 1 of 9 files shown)


Jay-o-Way:
@niels9001 something to think about for PowerToys maybe?

yaira2:
We have been very satisfied with WinUIEx, as it provided a crucial function during our initial migration to WinAppSdk. However, given the improvements in WinAppSdk, we have reached a point where this dependency is no longer necessary.

0x5bfa:
leaving them below, making good progress

image

src/Files.App/MainWindow.xaml.cs
```diff
@@ -3,53 +3,46 @@
...
-
 			InitializeComponent();
 
-			EnsureEarlyWindow();
-		}
-
-		private void EnsureEarlyWindow()
-		{
-			// Set PersistenceId
-			PersistenceId = "FilesMainWindow";
```
yaira2, line 35: PersistenceId is required to separate the main window size from the properties window.

0x5bfa: let me take a look further 👍

0x5bfa: I found interesting thing; this persistence id is store in application data settings which we marks as obsolete in Files (SettingsViewModel.Get and Set)
Do we want to use this or use our settings json?

yaira2: PersistenceId is just a string which I believe AppWindow handled internally, it shouldn't have anything to do with `SettingsViewModel`.

0x5bfa: It uses ApplicationData.Current?.LocalSettings.Container.
That's what we had been using in SettingsViewModel.

yaira2: > It uses ApplicationData.Current?.LocalSettings.Container.

This makes sense, but it's all handled internally. It doesn't have anything to do with SettingsViewModel.

0x5bfa: I know but I wanted to point out this is outdated technology in Files and ask you if we want to use GeneralSettingsViewModel. This property is only in WinUIEx and we have to implement ourselves.

0x5bfa: What we have to implement is this property and min window size (already done).

yaira2: > I know but I wanted to point out this is outdated technology in Files and ask you if we want to use GeneralSettingsViewModel.

We shouldn't be saving or reading any settings ourselves, that is assuming the property is available in WinAppSdk (I'd be surprised if it was removed). I'm not certain if it's the same as what we had in [UWP](https://learn.microsoft.com/en-us/uwp/api/windows.ui.windowmanagement.appwindow.persistedstateid?view=winrt-26100), but there is an ID property for [WinAppSdk](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.windowing.appwindow.id?view=windows-app-sdk-1.5).

src/Files.App/MainWindow.xaml.cs
```diff
@@ -18,38 +21,31 @@ public sealed partial class MainWindow : WindowEx
...
-			AppWindow.SetIcon(AppLifecycleHelper.AppIconPath);
-			AppWindow.TitleBar.ExtendsContentIntoTitleBar = true;
 			AppWindow.TitleBar.ButtonBackgroundColor = Colors.Transparent;
 			AppWindow.TitleBar.ButtonInactiveBackgroundColor = Colors.Transparent;
+			AppWindow.TitleBar.ButtonPressedBackgroundColor = Colors.Transparent;
+			AppWindow.TitleBar.ButtonHoverBackgroundColor = Colors.Transparent;
+			AppWindow.SetIcon(AppLifecycleHelper.AppIconPath);
+
+			MinHeight = 550;
+			MinWidth = 460;
```
yaira2, line 41: I'm not sure where the values 416 and 516 came from, but what's the reason for changing them?

yaira2:
@0x5bfa can you resolve the merge conflicts?

src/Files.App/MainWindow.xaml.cs
0x5bfa: I think we don't need it anymore.
`https://github.com/dotMorten/WinUIEx/issues/178#event-13751222808`

src/Files.App/MainWindow.xaml.cs
0x5bfa: This was not intentional so I'll revert back soon.

I find the good min width and height for splash screen not to overwrap the splash screen text with progress ring.
But we can use the same values.

yaira2:
@0x5bfa launching Files via this branch doesn't seem to work.

src/Files.App/MainWindow.xaml.cs
0x5bfa: Reverted

src/Files.App/Services/Settings/GeneralSettingsService.cs
```diff
@@ -315,6 +315,30 @@ public ShellPaneArrangement ShellPaneArrangementOption
...
+		{
+			get => Get(int.MaxValue);
+			set => Set(value);
+		}
+
+		public int MainWindowPositionY
+		{
+			get => Get(int.MaxValue);
+			set => Set(value);
+		}
```
yaira2, line 340: Can we use a tuple to condense the code?

0x5bfa: I'm gonna use ApplicationData as you suggested.
The key names would be those.

- MainWindowWidth
- MainWindowHeight
- MainWindowPositionX
- MainWindowPositionY

What do you think?

0x5bfa: Also, one more last thing we should concern is when monitor info has changed since last time the app launched.
If monitor resolution changed WinUIEx doesn't restore info and reset info.

yaira2: The key names aren't that important but we might as well use the same ones as WinUIEx for the least disruption.

0x5bfa: WinUIEx creates a key named $"WindowPersistance_{PersistenceId}" and puts value that is converted into base64 string of WINDOWPLACEMENT struct data.

Reference:

https://github.com/dotMorten/WinUIEx/blob/ea41e17bc3ecba48491db9920a9ed7348f62d44b/src/WinUIEx/WindowManager.cs#L428-L469

0x5bfa: I think we don't want such a complexity so a set of keys is preferable

yaira2: We would have to migrate the existing settings either way.

0x5bfa: You don't want to drop support of the window size before this removal? Even just one time?

Btw if so, when are we gonna such migration code?

yaira2: I meant that the existing window size settings shouldn't get lost.

0x5bfa: But it requires a migration code.
When are we gonna remove it(shouldn't exist forever right?)?

yaira2: A couple months is good, but if migrating is complicated, we can just start fresh.

0x5bfa: It's not that complicated, I'll make a code of it and then please take a look at later:)

0x5bfa:
Ready for testing.

yaira2:
@0x5bfa I’ve noticed that when I’m in full screen mode, the taskbar doesn’t show up when I move the cursor to the taskbar area. Everything else appears to be working though!

0x5bfa:
Fixed

yaira2 (approved):

yaira2:
@0x5bfa I found an issue where the navigation items are missing from the properties window.

0x5bfa:
What do you mean??
It gone? On my machine, it exists

yaira2:
Did you compile from `main`?

PR #15835, opened 2024-07-21
Code Quality: Added logging for ItemDisplayFrame

yaira2:
**Resolved / Related Issues**

- Closes #

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Views/Shells/BaseShellPage.cs
+++ b/src/Files.App/Views/Shells/BaseShellPage.cs
@@ -1,11 +1,11 @@
 // Copyright (c) 2024 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
+using Microsoft.Extensions.Logging;
 using Microsoft.UI.Input;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/Shells/BaseShellPage.cs
```diff
@@ -567,18 +567,29 @@ public virtual void Back_Click()
...
-			else
-				ItemDisplay.GoBack();
+			try
+			{
+				ItemDisplay.GoBack(new SuppressNavigationTransitionInfo());
+			}
+			catch (COMException ex)
+			{
+				App.Logger.LogWarning(ex, ex.Message);
+			}
```
hishitetsu, line 577: Is it intentional to eliminate the if-else?

hishitetsu (commented):
Since page transitions are a fundamental function of Files, I don't think it is appropriate to catch Exception here and ignore it.

src/Files.App/Views/Shells/BaseShellPage.cs
yaira2: Yes, we don't need the `EntranceNavigationTransitionInfo`

PR #15918, opened 2024-07-31
Code Quality: Log errors when setting FolderBackgroundImageSource

yaira2:
Prevent crash and log error when creating URI from FolderBackgroundImageSource


yaira2:
Would it be better to display a prompt with more information? This exception mainly occurs if the path isn't set correctly.

0x5bfa (approved):
Looks good 👍

PR #15925, opened 2024-08-02
Fix: Fixed a bug where empty pages were stealing focus from the Properties window

### LINKED ISSUE #13697: When opening empty folder in Files, property window for a drive (Right-click>Properties) moves to background & stays there

Qin-shihuang:
**Resolved / Related Issues**

- Closes #13697

1. Open an empty folder
2. Right-click a folder in the sidebar and open the Properties window
3. Before: The Properties window opened but lost focus immediately, and clicking on it didn't bring it front
	After: The window gains focus normally

### DIFF
--- a/src/Files.App/Views/Layouts/BaseLayoutPage.cs
+++ b/src/Files.App/Views/Layouts/BaseLayoutPage.cs
@@ -1285,6 +1285,14 @@ namespace Files.App.Views.Layouts
 				ItemManipulationModel.SetSelectedItem(rightClickedItem);
 		}
 
+		protected void FileList_GettingFocus(UIElement sender, GettingFocusEventArgs args)
+		{
+			// Refuse to gain focus if the file list is empty.
+			// Otherwise, it keeps stealing focus from the Properties window (#13697)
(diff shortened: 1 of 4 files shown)


Qin-shihuang:
I'm not sure why focusing on `FileList` causes issues, maybe it would be better to solve it directly on `FileList`.

This workaround prevents navigating before and after `FileList` using Tab, but I didn't figured out how to make the focus skip it

PR #15927, opened 2024-08-03
Code Quality: Introduced IWindowsAppLauncherService

0x5bfa:
### Resolved / Related Issues

This also involves with IShellItem, methods of which return HRESULT but CsWin32 generates those methods with void returned. But this issue can be detoured by not allowing marshaling by CsWin32. However, this is blocked by WinUIEx because of CsWin32 issue, conflicting generated code modifiers between internal (WinUIEx) and public (Files). WinUIEx removal is almost done so this should be able to be continued soon again.

- Related https://github.com/files-community/Files/issues/15000

_None yet_


0x5bfa:
@yaira2 Which one would you like: WindowsAppLauncher or WindowsLaunchApp?

0x5bfa:
@yaira2 Which one would you like: WindowsAppLauncherService or WindowsLaunchAppService?

0x5bfa:
I have CreateFile method using Vanara in this file yet as well.
Imo, I'd like to replace them after this to reduce the work of review because I already made complicated changes.
What do you think?

src/Files.App/Data/Contracts/IWindowsAppLauncherService.cs
yaira2, line 1: Could you please add some comments? I understand this code is based on existing code, but our standard practice is to include comments when creating an interface. Thanks!

src/Files.App/UserControls/Widgets/DrivesWidget.xaml.cs
```diff
@@ -52,7 +53,7 @@ private async void GoToStorageSense_Click(object sender, RoutedEventArgs e)
 				button.Tag.ToString() is not string path)
 				return;
 
-			await StorageSenseHelper.OpenStorageSenseAsync(path);
+			await WindowsLaunchAppService.LaunchStorageSensePolicySettingsAsync();
```
yaira2, line 56: Should you include the path in the arguments?

src/Files.App/Data/Contracts/IWindowsAppLauncherService.cs
0x5bfa: Ah I forgot that even though I appended inheritdoc to implementations lol

src/Files.App/UserControls/Widgets/DrivesWidget.xaml.cs
0x5bfa: The first implementation to include path is made by Gave but I'm not sure why because Storage Sense works for only system drive and for the whole system drive not for specifically a path.

0x5bfa: Actually path isn't used to be passed to settings to give extra context.

yaira2: There is a thread on GitHub about this with more details, but in any event, it makes a difference for which page is opened in Settings.

0x5bfa: Both open wrong page.

For other than C (system) drive:
image

For thers:
image

Both should open:
image

0x5bfa: My implementapention opens what should be opened.

yaira2: > Both open wrong page.

It's supposed to open the Storage Page for the `C:\` drive, and the "Storage Used on Other Drives" page for other drives.

0x5bfa: Even so, how would you free up storage for other drives? The "Storage used on other drives" doesn't contains any free up page.
As to C, why not open Policy page directly.

0x5bfa: I assume this weird page navigation might come from win10 when Gave implemented

yaira2: > Even so, how would you free up storage for other drives? The "Storage used on other drives" doesn't contains any free up page.

You can access the "Free up storage" page by clicking on a drive. Ideally we'd open directly to that page, but that's not supported by the Settings app.

0x5bfa:
This should not be DI'd.
Ill create a new one for this.

PR #15934, opened 2024-08-04
Fix: Fixed an issue where shortcuts couldn't be created for commands

### LINKED ISSUE #15928: Bug: Can't put commands in shortcuts files
### Description

In Windows, u can create command in a shortcuts and it worked well on file explorer however in Files u cant create a shortcut with a command in it (It only worked on the desktop).

### Steps To Reproduce

1. Right click in a random folder (e.g Music folder)
2. Click New and then Shortcut 
3. Try to put a command (e.g winget upgrade --all) and it would say "Invalid location"

### Requirements

Files would not allow to put commands in shortcut.

### Files Version

3.5.0.0

### Windows Version

22631.3958

### Log File

[debug.log](https://github.com/user-attachments/files/16478384/debug.log)

Qin-shihuang:
**Resolved / Related Issues**

- Closes #15928

1. Right click in a folder
2. Click New -> Shortcut in the context menu
3. Type "cmd /?" in the textbox
4. Before: You got a prompt saying "Invalid location"
	After: Shortcut successfully created
5. (After) Open the shortcut and see the cmd help text

**-- Update --**
These modifications make the behavior of creating shortcuts more consistent with File Explorer.

These modifications can be tested by creating the following shortcuts, all of which are supported by File Explorer but not supported by Files before:
1. `cmd`
2. `cmd.exe`
3. `cmd /?`
4. `  cmd`
5. `"C:\Windows\System32\cmd.exe" /?`
6. `C:\Windows\System32\cmd.exe /?`
7. `"C:\Program Files\PowerShell\7\pwsh.exe" -c cmd /?`
8. `"https://www.google.com"`

Also fixed the issue 4 mentioned below, which is caused by Path.Exists considering Files' startup path as relative path.

### DIFF
--- a/src/Files.App/ViewModels/Dialogs/CreateShortcutDialogViewModel.cs
+++ b/src/Files.App/ViewModels/Dialogs/CreateShortcutDialogViewModel.cs
@@ -92,8 +92,7 @@ namespace Files.App.ViewModels.Dialogs
 						// If the quoted part is a valid filename, try to find it in the PATH
 						if (quoted == Path.GetFileName(quoted)
 							&& quoted.IndexOfAny(Path.GetInvalidFileNameChars()) == -1
-							&& PathHelpers.TryGetFullPath(quoted, out var fullPath)
-							)
+							&& PathHelpers.TryGetFullPath(quoted, out var fullPath))
 						{
 							DestinationPathExists = true;
 							IsLocationValid = true;
@@ -112,12 +111,6 @@ namespace Files.App.ViewModels.Dialogs
 					}
 					else
 					{
-						if (trimmed == _previousShortcutTargetPath)
-						{
-							Arguments = trimmed.Split(' ')[1..].Aggregate(Arguments, (current, arg) => current + arg + " ");
-							return;
-						}
-
 						// Try to parse the whole text as path
 						if (Path.Exists(trimmed)
 						    && Path.IsPathFullyQualified(trimmed)
@@ -132,10 +125,16 @@ namespace Files.App.ViewModels.Dialogs
 						}
 
 						var filename = trimmed.Split(' ')[0];
+
+						if (filename == _previousShortcutTargetPath)
+						{
+							Arguments = trimmed.Split(' ')[1..].Aggregate(Arguments, (current, arg) => current + arg + " ");
+							return;
+						}
+
 						if (filename == Path.GetFileName(filename)
 							&& filename.IndexOfAny(Path.GetInvalidFileNameChars()) == -1
-							&& PathHelpers.TryGetFullPath(filename, out var fullPath)
-							)
+							&& PathHelpers.TryGetFullPath(filename, out var fullPath))
 						{
 							DestinationPathExists = true;
 							IsLocationValid = true;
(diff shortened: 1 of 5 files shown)


Qin-shihuang:
While working on this, I discovered several other issues:

1. This page cannot create shortcuts for drives
image

2. The default naming rule looks strange
image

src/Files.App/ViewModels/Dialogs/CreateShortcutDialogViewModel.cs
```diff
@@ -45,20 +46,35 @@ public string DestinationItemPath
...
+					IsLocationValid = uri.IsWellFormedOriginalString();
+					_isLocationInPath = false;
+					_fullPath = DestinationItemPath;
 				}
 				catch (Exception)
 				{
+					DestinationPathExists = false;
 					IsLocationValid = false;
+					_isLocationInPath = false;
 				}
```
hishitetsu, line 78: Is it unnecessary to set `_fullPath`?

src/Files.App/ViewModels/Dialogs/CreateShortcutDialogViewModel.cs
```diff
@@ -75,6 +91,9 @@ public bool IsLocationValid
 			}
 		}
 
+		private bool _isLocationInPath;
```
hishitetsu, line 94: Is the name of this field appropriate? I think a name like `_hasArguments` would be more appropriate.

src/Files.App/ViewModels/Dialogs/CreateShortcutDialogViewModel.cs
```diff
@@ -115,12 +134,17 @@ private async Task CreateShortcutAsync()
...
-				destinationName = Path.GetFileName(DestinationItemPath);
-				if (string.IsNullOrEmpty(destinationName))
+				destinationName = Path.GetFileName(_fullPath);
+				if (_isLocationInPath)
+					arguments = DestinationItemPath.Split(' ')[1..].Aggregate(arguments, (current, arg) => current + arg + " ");
+
+				if(string.IsNullOrEmpty(destinationName))
 				{
+					
 					var destinationPath = DestinationItemPath.Replace('/', '\\');
```
hishitetsu, line 148: Shouldn't `DestinationItemPath` be replaced with `_fullPath`?
Same for L158.

src/Files.Shared/Helpers/PathHelpers.cs
```diff
@@ -63,5 +64,39 @@ public static bool IsSpecialFolder(string path)
...
+				{
+					UseShellExecute = false,
+					CreateNoWindow = true,
+					FileName = "where.exe",
+					Arguments = exeName,
+					RedirectStandardOutput = true
+				};
+				p.Start();
+				var output = p.StandardOutput.ReadToEnd();
+				p.WaitForExit();
```
hishitetsu, line 83: A timeout period should be set to avoid stoppages.

Qin-shihuang:
Sorry for the low-quality code. I am rewriting a better version that is more consistent with Files Explorer's behavior.

hishitetsu:
Don't mind! Your contribution is very helpful. It's working and I don't consider it low-quality.

hishitetsu (commented):
This would cause PathHelpers.TryGetFullPath() to be called for every input, even when editing arguments.
It is better to call it only when the file name is changed.

hishitetsu (approved):
Nice work! LGTM
But as you say, there are many other things that need to be fixed in this feature.

PR #15947, opened 2024-08-05
Feature: Added a setting to hide the System Tray icon

### LINKED ISSUE #15907: Add option to hide files app from system tray

Please add the option to remove the system tray icon for the files app. My taskbar system tray is very cluttered and I'm trying to remove as many icons as I can. Hiding the icon so that it is only revealed when I click the up arrow on my taskbar is also very cluttered now, so being able to remove system tray utilities I never use would be nice.

### Requirements

- This proposal will help users with cluttered taskbar system trays clean up a little.

### Files Version

3.5.0.0

### Windows Version

10.0.22631.3958

### Comments

_No response_

yaira2:
**Resolved / Related Issues**

- Closes #15907

1. Confirmed that changing the setting updates the icon visibility
2. Confirmed the setting persists across sessions

### DIFF
--- a/src/Files.App/App.xaml.cs
+++ b/src/Files.App/App.xaml.cs
@@ -18,7 +18,7 @@ namespace Files.App
 	/// </summary>
 	public partial class App : Application
 	{
-		private static SystemTrayIcon? SystemTrayIcon { get; set; }
+		public static SystemTrayIcon? SystemTrayIcon { get; set; }
 
 		public static TaskCompletionSource? SplashScreenLoadingTCS { get; private set; }
 		public static string? OutputPath { get; set; }
--- a/src/Files.App/App.xaml.cs
+++ b/src/Files.App/App.xaml.cs
@@ -132,16 +132,18 @@ namespace Files.App
 					SplashScreenLoadingTCS = null;
 
 					// Create a system tray icon
+					SystemTrayIcon = new SystemTrayIcon();
 					if (userSettingsService.GeneralSettingsService.ShowSystemTrayIcon)
-						SystemTrayIcon = new SystemTrayIcon().Show();
+						SystemTrayIcon.Show();
 
 					_ = MainWindow.Instance.InitializeApplicationAsync(appActivationArguments.Data);
 				}
 				else
 				{
 					// Create a system tray icon
+					SystemTrayIcon = new SystemTrayIcon();
 					if (userSettingsService.GeneralSettingsService.ShowSystemTrayIcon)
-						SystemTrayIcon = new SystemTrayIcon().Show();
+						SystemTrayIcon.Show();
 
 					// Sleep current instance
 					Program.Pool = new(0, 1, $"Files-{AppLifecycleHelper.AppEnvironment}-Instance");
--- a/src/Files.App/App.xaml.cs
+++ b/src/Files.App/App.xaml.cs
@@ -18,7 +18,7 @@ namespace Files.App
 	/// </summary>
 	public partial class App : Application
 	{
-		public static SystemTrayIcon? SystemTrayIcon { get; set; }
+		public static SystemTrayIcon? SystemTrayIcon { get; private set; }
 
 		public static TaskCompletionSource? SplashScreenLoadingTCS { get; private set; }
 		public static string? OutputPath { get; set; }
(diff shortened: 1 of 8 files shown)


src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
```diff
@@ -139,6 +141,8 @@ public SystemTrayIcon()
 			_IconWindow = new SystemTrayIconWindow(this);
 
 			CreateOrModifyNotifyIcon();
+
+			GeneralSettingsService.PropertyChanged += GeneralSettingsService_PropertyChanged;
```
yaira2, line 145: @0x5bfa is this Constructor ever called multiple times in a single session?

src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
```diff
@@ -18,6 +18,8 @@ namespace Files.App.Utils.Taskbar
 	/// </summary>
 	public sealed class SystemTrayIcon : IDisposable
 	{
+		private IGeneralSettingsService GeneralSettingsService { get; } = Ioc.Default.GetRequiredService<IGeneralSettingsService>();
```
yaira2, line 21: I don't think this is good practice, is there a better way to go about this?

0x5bfa (commented):
If I were, I think I'd put the event handler in App.ActivateAsync or AppLifecycleHelper.InitializeAppComponentsAsync since the tray service should be independent from any settings

src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
yaira2: I pushed a commit with these changes

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
```diff
@@ -93,6 +93,8 @@ static Task OptionalTaskAsync(Task task, bool condition)
 
 				return Task.CompletedTask;
 			}
+
+			generalSettingsService.PropertyChanged += GeneralSettingsService_PropertyChanged;
```
yaira2, line 97: Will this cause a memory leak if the window is closed and reopened?

0x5bfa: I believe no. When you have running on background enabled, the process still alive, otherwise GC will finalize when the process is eliminated. FYI @hishitetsu

hishitetsu: No, this method is called only once at startup.

0x5bfa: By the way, how do you unhook an event in static class?
For non static class, we should create a event field and dispose it in destructor or Dispose method?

hishitetsu: Since static classes are not destroyed until the app is exited, the events associated with them should not need to be unsubscribed. If there is a need to unsubscribe in some time, the events should be subscribed/unsubscribed in objects that have another lifecycle.

0x5bfa (approved):
Code wise, LGTM.

src/Files.App/App.xaml.cs
```diff
@@ -18,7 +18,7 @@ namespace Files.App
 	/// </summary>
 	public partial class App : Application
 	{
-		private static SystemTrayIcon? SystemTrayIcon { get; set; }
+		public static SystemTrayIcon? SystemTrayIcon { get; set; }
```
hishitetsu, line 21:
```suggestion
		public static SystemTrayIcon? SystemTrayIcon { get; private set; }
```

src/Files.App/Helpers/Application/AppLifecycleHelper.cs
```diff
@@ -356,5 +358,21 @@ public static bool IsAutoHideTaskbarEnabled()
 			// The least significant bit of the 9th byte controls the auto-hide setting																		
 			return value != null && ((value[8] & 0x01) == 1);
 		}
+
+		/// <summary>
+		/// Updates the visibility of the system tray icon
+		/// </summary>
+		private static void GeneralSettingsService_PropertyChanged(object? sender, PropertyChangedEventArgs e)
+		{
+			var generalSettingsService = Ioc.Default.GetRequiredService<IGeneralSettingsService>();
```
hishitetsu, line 367:
```suggestion
			if (sender is not IGeneralSettingsService generalSettingsService)
				return;
```

hishitetsu (approved):
LGTM, except for some code improvements

PR #15949, opened 2024-08-05
Code Quality: Remove spacing from SidebarItems

yaira2:
**Resolved / Related Issues**

- Removed spacing from between sidebar items and added a margin instead.


yaira2:
@0x5bfa this introduced a new crash that seems to be affecting a lot of users in v3.6.1 but I can't repro it personally. #16026

Issue #12367, opened 2023-05-16
Feature: dissolve or flatten folder


Jay-o-Way:
### What feature or improvement do you think would benefit Files?

Would love an option to dissolve or flatten a folder. Placing all the content in the parent folder.
Simply said, it's the opposite of "Create folder with selection" and placing it above or below this entry would make the most sense.
Example:

- 📁 Parent folder
   - 📄 File A
   - 📁 Sub-folder _(target this folder)_
      - 📄 File B
      - 🖼 File C

Would give:

- 📁 Parent folder
   - 📄 File A
   - 📄 File B
   - 🖼 File C

### Requirements

To have a context menu item on a folder that will pull the content one (directory) level up. Must deal with existing/conflicting files (like Explorer does) with a prompt.

### Files Version

2.4.71

### Windows Version

Windows 10

### Comments

I've seen many, many requests for this in the PowerToys repo, so I **know** it's something people want.
- https://github.com/microsoft/PowerToys/issues/9269

yaira2:
We originally discussed this in #10619 but I think it's worth bringing back the discussion again.

AtlasHouse:
Hi, has this feature been incorporated in the latest Files release?

Jay-o-Way:
Nope

yaira2:
I updated the requirements to only include one selected folder. Based on the reception and usage of this feature, we'll consider adding support for additional options.

hishitetsu:
I would like to be more clear on the requirements.
* Should we also recursively flatten subfolders of the target folder?
* Should the target folder be deleted after the flattening is complete?

yaira2:
> Should we also recursively flatten subfolders of the target folder?

I can see scenarios where both would make sense, perhaps we can add a modal to let the user choose.

> Should the target folder be deleted after the flattening is complete?

That's my understanding.

hishitetsu:
> perhaps we can add a modal to let the user choose.

How about having each as a separate command?

yaira2:
If we add additional options, we're going to want a modal either way. An added benefit is that it allows the user to confirm what would otherwise be a destructive action.

yaira2:
Fyi @Tundra

smarquespt:
From a translator POV, is this really what "Flatten" means? I don't have a concrete wording but I think "Flatten" is not the best approach ...
https://dictionary.cambridge.org/dictionary/english/flatten

yaira2:
It combines the folder structure into a single folder.

[...]

PR #15992, opened 2024-08-11
Feature: Added experimental support for flattening folders

devin-slothower:
**Resolved / Related Issues**

- For https://github.com/files-community/Files/issues/12367

1. Tested many nested folders both blank and with data.
2. Tested with merging duplicate files.
3. Assured the Flatten popup only appears for a single selected folder.
4. Ran through a bunch of scenarios trying to crash with/without a debugger.

**Menu Preview**
image


yaira2:
This is amazing @devin-slothower! Just a heads-up, you only need to add the strings to the en-us file. The other files will automatically update when we sync with Crowdin

src/Files.App/Actions/FileSystem/FlattenRecursiveAction.cs
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		private async Task FlattenFolderAsync(string folderPath)
+		{
+			var containedFolders = await Task.Run(() => Directory.GetDirectories(folderPath));
+			var containedFiles = await Task.Run(() => Directory.GetFiles(folderPath));
+
+			foreach (var containedFolder in containedFolders)
+			{
+				await FlattenFolderAsync(containedFolder);
```
yaira2, line 55: Is it possible to avoid using `await` in the loop?

src/Files.App/Actions/FileSystem/FlattenRecursiveAction.cs
```diff
@@ -0,0 +1,110 @@
...
+			foreach (var item in items)
+			{
+				await FlattenFolderAsync(item.ItemPath);
+			}
+		}
+
+		private async Task FlattenFolderAsync(string folderPath)
+		{
+			var containedFolders = await Task.Run(() => Directory.GetDirectories(folderPath));
+			var containedFiles = await Task.Run(() => Directory.GetFiles(folderPath));
```
0x5bfa, line 51: How come you made this awaitable?

0x5bfa (commented):
Using Actions is great idea!

src/Files.App/Actions/FileSystem/FlattenSingleAction.cs
0x5bfa, line 1: FlattenCcurrent is better? @yaira2

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
```diff
@@ -539,6 +539,22 @@ private static bool Check(ContextMenuFlyoutItemViewModel item, CurrentInstanceVi
 					ShowItem = UserSettingsService.GeneralSettingsService.ShowCompressionOptions && StorageArchiveService.CanDecompress(selectedItems)
 				},
 				new ContextMenuFlyoutItemViewModel()
+				{
+					Text = "Flatten".GetLocalizedResource(),
+					ShowInSearchPage = true,
+					ThemedIconModel = new ThemedIconModel()
+					{
+						ThemedIconStyle = "App.ThemedIcons.Folder",
+					},
```
0x5bfa, line 548:
```suggestion
					ThemedIconModel = new() { ThemedIconStyle = "App.ThemedIcons.Folder" },
```

src/Files.App/Services/Settings/GeneralSettingsService.cs
```diff
@@ -297,6 +297,12 @@ public bool ShowSystemTrayIcon
 			set => Set(value);
 		}
 
+		public bool ShowFlattenOptions
+		{
+			get => Get(true);
```
0x5bfa, line 302: Show by default? @yaira2

yaira2: Probably off by default

src/Files.App/Actions/FileSystem/FlattenSingleAction.cs
yaira2: I've considered some alternatives, but I'm not sure they are any better than "Flatten single".
- Flatten
- Flatten {directory}

src/Files.App/Actions/FileSystem/FlattenRecursiveAction.cs
devin-slothower: > How come you made this awaitable?

My thinking was that a recursive search on a big enough directory may end up temporarily freezing the main thread including the UI, so I was using await Task.Run to run those functions on a background thread.

src/Files.App/Actions/FileSystem/FlattenRecursiveAction.cs
devin-slothower: > Is it possible to avoid using `await` in the loop?

Sure, I've pushed a commit changing this to run synchronously.

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
devin-slothower: I think this should stay the way it is currently, the coding style says to align curly brackets on new lines and the surrounding flyout items are written in the same way.

Jay-o-Way:
Wooohhh! Lemme check this out on my computer 😁

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
0x5bfa: It's ok, I was going to change here to:

new("App.ThemedIcons.Folder"),

It's easier if it is written like that I suggested.

src/Files.App/Actions/FileSystem/FlattenSingleAction.cs
devin-slothower: Some other alternatives I came up with are 
- Flatten Folder
- Flatten All

and
- Flatten
- Flatten All

I think the second one needs a simpler synonym for recursive. Any thoughts lmk!

src/Files.App/Actions/FileSystem/FlattenRecursiveAction.cs
0x5bfa: You don't have to use UI awaitable thread?

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
0x5bfa: @devin-slothower Please change here.

src/Files.App/Actions/FileSystem/FlattenRecursiveAction.cs
devin-slothower: > You don't have to use UI awaitable thread (DispatcherQueue)?

I'm new to C# so I'm don't quite have a grasp on how threads work, this is good to know thanks!

Jay-o-Way:
Hm, at the moment it doesn't show up for me. I opened the pr/code in GitHub Desktop and Visual Studio, then build the solution, then opened `Files.App.Launcher.exe`. Can't find the setting for it under Context Menu options.

Jay-o-Way:
How do I make sure I launch the correct way to test this?

> [!NOTE]
> Pressing "Files.App" to Debug gives an exeption:
> ```
> Exception thrown: 'System.InvalidOperationException' in WinRT.Runtime.dll
> Exception thrown: 'System.TypeInitializationException' in Files.dll
> Exception thrown: 'System.TypeInitializationException' in Files.dll
> An unhandled exception of type 'System.TypeInitializationException' occurred in Files.dll
> The type initializer for 'Files.App.Program' threw an exception.
>```

0x5bfa:
> Files.App

Build it from Files.Package

image

Jay-o-Way:
I think the resulting action needs to be clear: what files will end up where? Specifically - what will become the root folder? A reasonable expectation can be that, when applying this on _one_ folder, all the the SUBfolders will be dissolved, but the selected folder will remain. In stead, everything is "dumped" in the above folder. But I suppose this behavior makes sense to generalize the different triggering scenarios: selecting one or more folders, selecting one or more files...

yaira2:
Would an optional dialog with a preview of the changes help?

yaira2:
@devin-slothower what's the behavior for duplicate files?

devin-slothower:
> @devin-slothower what's the behavior for duplicate files?

The current behavior is just to keep the duplicate files where they are and move all non duplicates to their respective place.

yaira2:
I don’t want to complicate things, so this isn’t a blocker, but it would be great if we could handle this better. Here are some ideas:
- Show a conflict dialog and let the user choose an action.
- Add a prompt before flattening folders, allowing the user to set a default behavior for any conflicts.
- Display a prompt at the end of the operation with a list of files that weren’t moved.

Personally, I lean towards the second idea, but I’m open to discussion. We should probably handle this in a separate PR once we agree on a solution.

devin-slothower:
> I don’t want to complicate things, so this isn’t a blocker, but it would be great if we could handle this better. Here are some ideas:
> 
> * Show a conflict dialog and let the user choose an action.
> * Add a prompt before flattening folders, allowing the user to set a default behavior for any conflicts.
> * Display a prompt at the end of the operation with a list of files that weren’t moved.
> 
> Personally, I lean towards the second idea, but I’m open to discussion. We should probably handle this in a separate PR once we agree on a solution.

I think the best option would be to handle conflicts similar to the default file manager where the user is prompted for an action, should a conflict arise, prior to anything being moved. I would recommend the following options (with some tweaking to the names):

* Replace
* Rename (Either prompting for a new name, or automatically appending a suffix like "file - 1.txt", "file - 2.txt", etc)
* Skip
* Cancel

0x5bfa:
It's a bit confusing but https://github.com/files-community/Files/blob/main/src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs#L656 how to use. The dialog class is shared among multiple file system operations and dialog conflict resolve is one of them. There's an enum property to get how to resolve for each.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2027,6 +2027,21 @@
...
+    <value>Flatten</value>
+  </data>
+  <data name="FlattenDescription" xml:space="preserve">
+    <value>Flatten a folder contents into the current path.</value>
+  </data>
+  <data name="FlattenSingle" xml:space="preserve">
+    <value>Flatten Single</value>
+  </data>
+  <data name="FlattenRecursive" xml:space="preserve">
+    <value>Flatten Recursive</value>
```
yaira2, line 2040:
```suggestion
    <value>Flatten recursively</value>
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2027,6 +2027,21 @@
...
   </data>
+  <data name="Flatten" xml:space="preserve">
+    <value>Flatten</value>
+  </data>
+  <data name="FlattenDescription" xml:space="preserve">
+    <value>Flatten a folder contents into the current path.</value>
+  </data>
+  <data name="FlattenSingle" xml:space="preserve">
+    <value>Flatten Single</value>
+  </data>
```
yaira2, line 2038: Can you remove this string and use `Flatten` instead?

src/Files.App/Actions/FileSystem/FlattenRecursiveAction.cs
```diff
@@ -0,0 +1,114 @@
...
+			if (items is null || !items.Any() || items.Any(item => !item.IsFolder))
+				return Task.CompletedTask;
+
+			foreach (var item in items)
+			{
+				FlattenFolderAsync(item.ItemPath);
+			}
+
+			return Task.CompletedTask;
+		}
```
yaira2, line 48:
```suggestion
		public async Task ExecuteAsync(object? parameter = null)
		{
			if (context.ShellPage?.ShellViewModel is null)
				return;

			var items = context.SelectedItems;

			if (items is null || !items.Any() || items.Any(item => !item.IsFolder))
				return;

			await Task.WhenAll(items.Select(item => FlattenFolderAsync(item.ItemPath)));
		}
```
@0x5bfa would this be better for performance?

yaira2:
@devin-slothower I tested this and I'm very happy with the behavior. However, like @Jay-o-Way, I was a bit confused about where the files would end up and I think we need to do some more work. Are you interested in polishing the behavior in this PR, or would you prefer to merge as is and mark the feature as experimental?

yaira2:
Let's merge as experimental so that this work doesn't get lost. I also noticed that we don't display an progress of the file operation, which is something we'd want to do before marking the feature as stable.

yaira2:
@devin-slothower I made some adjustments based on user feedback. I also updated the list in #12367 to track improvements to make at a later time.

Updated the wording to “Flatten to root” and changed the destination to be the selected folder.
Removed the “Flatten single action.”
Added a “Flatten folder” action. This isn’t functional yet but opens a dialog that will contain more advanced options in the future.
From my perspective, this PR is ready to merge, but I’ll leave it open for a couple of hours to see if there is any other feedback.

src/Files.App/Actions/FileSystem/FlattenToRootAction.cs
```diff
@@ -0,0 +1,116 @@
...
+
+			var items = context.SelectedItems;
+
+			if (items is null || !items.Any() || items.Any(item => !item.IsFolder))
+				return Task.CompletedTask;
+
+			foreach (var item in items)
+			{
+				FlattenFolderAsync(item.ItemPath);
+			}
```
0x5bfa, line 47:
```suggestion
			foreach (var item in items)
				FlattenFolderAsync(item.ItemPath);
```

0x5bfa (approved):
Looks good!

src/Files.App/Actions/FileSystem/FlattenToRootAction.cs
```diff
@@ -0,0 +1,126 @@
...
+			if (items is null || !items.Any() || items.Any(item => !item.IsFolder))
+				return Task.CompletedTask;
+
+			foreach (var item in items)
+				FlattenFolderAsync(item.ItemPath);
+
+			return Task.CompletedTask;
+		}
+
+		private Task FlattenFolderAsync(string folderPath)
```
yaira2, line 51: @0x5bfa where do you recommend moving this code to?

hishitetsu: This method is not an async method and is not meant to return Task. The return value should be void.

0x5bfa: I think here is good.
Do you have something else in your mind?

yaira2: I thought we might want to move it to a service or helper.

src/Files.App/Actions/FileSystem/FlattenToRootAction.cs
```diff
@@ -0,0 +1,140 @@
+﻿// Copyright (c) 2024 Files Community
+// Licensed under the MIT License. See the LICENSE.
+
+using Microsoft.Extensions.Logging;
+using System.IO;
+using Windows.Storage;
+using Microsoft.UI.Xaml.Controls;
+using Windows.Foundation.Metadata;
+using Files.Shared.Helpers;
```
0x5bfa:
```suggestion
using Files.Shared.Helpers;
using Microsoft.Extensions.Logging;
using Microsoft.UI.Xaml.Controls;
using System.IO;
using Windows.Storage;
using Windows.Foundation.Metadata;
```

0x5bfa (approved):

hishitetsu (approved):
LGTM

PR #16025, opened 2024-08-17
Code Quality: Improved DependencyProperty generator

0x5bfa:
### Resolved / Related Issues

The primary achievement is to reduce lines of code and verbosity when declaring DependencyProperties.

_None_


0x5bfa:
Ready (tested locally) to merge.

@Poker-sang fyi. btw do you know how to emit ExpressionSyntax from TypedConstant that can be obtained from attribute properties list? This https://github.com/files-community/Files/pull/16025/commits/abb466106cec4a491169e46a3bc57869485a668f does its job but looks quite bad.

Poker-sang:
🤔 Since we've made it to the eve of the release of some of the properties, maybe we can wait for C#13 to come out before making adjustments? That way we can implement it in this syntax:

```cs
[DependencyProperty(PropertyChangedCallback, etc.)]
public partial object? DefaultValue { get; private set; } = new Object();
```

This way is much easier and more readable. And perhaps CommunityToolkit would like to implement a better version?

0x5bfa:
Yeah but it'll be implemented in WCT 8.x while we use 7.x and 8.x packages are incompatible with Files because some API from 7.x still missing. So I decided to make use of this for the time being (we're likely to create more controls in the near feature).
For more improvements, I'll wait until .NET 9 / C# 13 becomes GA.
I might ask you help again 😉

0x5bfa:
@yaira2 This is ready (tested locally), please rerun the CI to check.

Poker-sang:
> @Poker-sang fyi. btw do you know how to emit ExpressionSyntax from TypedConstant that can be obtained from attribute properties list? This [abb4661](https://github.com/files-community/Files/commit/abb466106cec4a491169e46a3bc57869485a668f) does its job but looks quite bad.
> 
> I have
> 
> ```cs
> object? DefaultValue { get; set; }
> ```
> 
> But AttributeData lets me get that value as object? with some extra info of its type. do i have to use switch statement and decide the expression based on type such as enum and primitive?

If you wish to do so, then indeed you can. You can get it through `((TypeDeclarationSyntax)generatorAttributeSyntaxContext.TargetNode).AttributeLists[0].Attributes[0].ArgumentList.Arguments[0].Expression`. However the arguments of the attributes only allow constants, which I don't think is very useful, so I used `string`

0x5bfa:
I think I'd like to use the latter one but there seems to be a cast error.

image

0x5bfa:
Worked thanks!!

Poker-sang (approved):
Looks good to me

yaira2 (dismissed):

0x5bfa:
That didn't work; somehow emitted property type string like "IconType" and caused an invalid cast exception.

Reverted to your original approach, I think we can wait for .NET partial properties to achieve this as you suggested.

Again, ready.

XTorLukas (approved):
LGTM

yaira2:
@0x5bfa are you ready for me to merge this?

0x5bfa:
YES!!

hishitetsu:
This PR disables changing the sort order by clicking on the header of the details layout.
image

0x5bfa:
Fixed

PR #16034, opened 2024-08-19
Code Quality: Deleted docs dir

0x5bfa:
### Resolved / Related Issues

We don't maintain here any more.
Also this dir includes api documentation that doesn't exists in current Files codebase, which can cause confusion.

_None_


yaira2:
Fyi @d2dyno1

PR #16048, opened 2024-08-21
Fix: Fixed crash from occuring when dropping items via the address bar

### LINKED ISSUE #15105: Bug: Crash after drag and drop on the addresses bar
### Description

Crash after drag and drop

### Steps To Reproduce

Tried to move a folder to address bar (trying to go up a level in the folder structure). That was not allowed. Then tried to move the folder down a level (because it had somehow moved up a level) and app crashed. Note that I had previously invoked the app with the key combination: WIN-E which btw caused all three opened tabs to duplicate.

### Requirements

Fix crashing issue

### Files Version

3.3.0.0

### Windows Version

W11 23H2

### Log File

[debug.log](https://github.com/files-community/Files/files/14841620/debug.log)

yaira2:
**Resolved / Related Issues**

- Closes #15105

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
+++ b/src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
@@ -339,7 +339,12 @@ namespace Files.App.ViewModels.UserControls
 
 			await signal.WaitAsync();
 
-			deferral.Complete();
+			try
+			{
+				deferral.Complete();
+			}
+			catch { }
+
 			await Task.Yield();
 
 			lockFlag = false;


src/Files.App/ViewModels/UserControls/AddressToolbarViewModel.cs
```diff
@@ -339,7 +339,12 @@ public async Task PathBoxItem_Drop(object sender, DragEventArgs e)
 
 			await signal.WaitAsync();
 
-			deferral.Complete();
+			try
+			{
+				deferral.Complete();
+			}
+			catch { }
```
hishitetsu, line 346: Most of the exceptions that occur in deferral are caused by Complete() being called even though there is an operation in progress. So we should make sure that all async operations are awaited correctly.
Even if an exception is caught here and a crash is avoided, the drop operation will not complete successfully and may behave unexpectedly.

hishitetsu: I think the crashes are occurring because the PathBoxItemDropped event invoked at L332 can't be awaited.

yaira2: I pushed some changes in [dcd1724](https://github.com/files-community/Files/pull/16048/commits/dcd17246901064a03f0c60cf146fa15d731ead4d)

hishitetsu: The similarity between this and #16026 is that both call FilesystemHelpers.PerformOperationTypeAsync() internally. So I suspect the problem is in this method.

hishitetsu: This may be the cause of the crash.
https://github.com/files-community/Files/blob/660a51e258ed4956680fa0aa6419da2b32144019/src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs#L478

PR #16069, opened 2024-08-24
Feature: Added support for Infomaniak kDrive

### LINKED ISSUE #15475: Feature: Add support for Infomaniak kDrive cloud

kDrive ([https://www.infomaniak.com/kdrive](https://www.infomaniak.com/kdrive)) doesn't show up in the cloud section. The folder is usually located in the user profile folder.

### Requirements

- This proposal will make the kDrive folder appear in the cloud section in sidebar
- The standard cloud-specific operations and icons should be supported

### Files Version

3.4.1.0

### Windows Version

10.0.22621.3593

### Comments

_No response_

mafra99:
**Resolved / Related Issues**

- Closes #15475

1. kDrive is displayed in Files and copying of files etc. is possible

Note: The Sync Status column is not displayed, but this does not work in Windows File Explorer either. I think kDrive does not set the property [“System.FilePlaceholderStatus”](https://github.com/files-community/Files/blob/main/src/Files.App/ViewModels/ShellViewModel.cs#L1885) correctly, which Files (and probably also Windows File Explorer) uses to find out the status.

image

image

### DIFF
--- a/src/Files.App/Utils/Cloud/CloudDrivesDetector.cs
+++ b/src/Files.App/Utils/Cloud/CloudDrivesDetector.cs
@@ -59,6 +59,7 @@ namespace Files.App.Utils.Cloud
 			var results = new List<ICloudProvider>();
 			using var clsidKey = Registry.ClassesRoot.OpenSubKey(@"CLSID");
 			using var namespaceKey = Registry.CurrentUser.OpenSubKey(@"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Desktop\NameSpace");
+			using var syncRootManagerKey = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\SyncRootManager");
 
 			foreach (var subKeyName in namespaceKey?.GetSubKeyNames() ?? [])
 			{
@@ -72,11 +73,23 @@ namespace Files.App.Utils.Cloud
 						continue;
 					}
 
-					//Nextcloud specific
-					var appName = (string)namespaceSubKey?.GetValue("ApplicationName");
-					if (!string.IsNullOrEmpty(appName) && appName == "Nextcloud")
+					var appNameFromNamespace = (string)namespaceSubKey?.GetValue("ApplicationName");
+					var appNameFromSyncRoot = (string)syncRootManagerKey?.OpenSubKey(driveType)?.GetValue(string.Empty);
+
+					// Nextcloud specific
+					if (!string.IsNullOrEmpty(appNameFromNamespace) && appNameFromNamespace == "Nextcloud")
 					{
-						driveType = appName;
+						driveType = appNameFromNamespace;
+					}
+
+					// kDrive specific
+					if (!string.IsNullOrEmpty(appNameFromNamespace) || appNameFromNamespace == "kDrive")
+					{
+						driveType = appNameFromNamespace;
+					}
+					else if (!string.IsNullOrEmpty(appNameFromSyncRoot) && appNameFromSyncRoot == "kDrive")
+					{
+						driveType = appNameFromSyncRoot;
 					}
 
 					// Drive specific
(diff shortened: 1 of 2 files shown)


src/Files.App/Utils/Cloud/CloudDrivesDetector.cs
```diff
@@ -108,6 +108,7 @@ private static Task<IEnumerable<ICloudProvider>> DetectGenericCloudDrive()
 						"Creative Cloud Files" => CloudProviders.AdobeCreativeCloud,
 						"ownCloud" => CloudProviders.ownCloud,
 						"ProtonDrive" => CloudProviders.ProtonDrive,
+						"kDrive" => CloudProviders.kDrive,
```
0x5bfa, line 111: Don't you have to add this?

```c#
if (driveType.StartsWith("kDrive"))
	driveType = "kDrive";
```

mafra99: I don't think this is necessary because the ApplicationName in the registry is exactly kDrive
image

For the others (e.g. iCloudDrive and OneDrive) there is sometimes an ID or name behind it
image
image

0x5bfa: Thanks for letting me know!

yaira2:
@mafra99 does this still work if the user changes the location from kDrive settings?

mafra99:
I have just reinstalled kDrive and now the entry in the registry no longer contains an application name, but only an ID. As a result, kDrive is no longer recognized on my device.
image


I can think of two other ways to recognize kDrive anyway: 
1. Using the DefaultIcon under Computer\HKEY_CLASSES_ROOT\CLSID\<>\DefaultIcon to check whether the path contains kDrive
image

2. Via the entry in Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\SyncRootManager
image

To use the entry from SyncRootManager seems safer to me, I have pushed a commit that contains the changes for this.


In the second commit I cleaned up the code a bit and fixed a few warnings.

The method GetDriveType is new and returns the drive type, e.g. iCloudDrive or kDrive. This should make the method DetectGenericCloudDrive a little simpler.
Compared to before, the DriveType is now returned directly as soon as one is detected and the checks for the other DriveTypes are no longer executed.

mafra99:
> @mafra99 does this still work if the user changes the location from kDrive settings?

@yaira2 thanks, that actually didn't work. should also be fixed

src/Files.App/Utils/Cloud/CloudDrivesDetector.cs
```diff
@@ -345,5 +329,33 @@ private static Task<IEnumerable<ICloudProvider>> DetectAutodeskDrive()
...
+			if (driveIdentifier.StartsWith("iCloudPhotos"))
+				return "iCloudPhotos";
+			if (driveIdentifier.StartsWith("ownCloud"))
+				return "ownCloud";
+			if (driveIdentifier.StartsWith("ProtonDrive"))
+				return "ProtonDrive";
+
+			// Nextcloud specific
+			if (!string.IsNullOrEmpty(appNameFromNamespace) && appNameFromNamespace == "Nextcloud")
+				return appNameFromNamespace;
```
hishitetsu, line 349: The order of Nextcloud specific and Drive specific are swapped, is it no problem? If it is, L335 should be moved just before here.

mafra99: The order of the checks should not matter.
But now the registry only has to be used if it cannot be identified by the driveIdentifier.

I have moved L335 👍

hishitetsu (approved):
LGTM, code wise.

0x5bfa (approved):
For the record, there's [GetSyncRootInformationForId WinRT method](https://learn.microsoft.com/uwp/api/windows.storage.provider.storageprovidersyncrootmanager.getsyncrootinformationforid) to retrieve sync root manager info but accessing the registry directly should be much faster.

LGTM, code wise.

PR #16075, opened 2024-08-25
Fix: Handle edge case where OpenSubKey results in SecurityException

yaira2:
**Resolved / Related Issues**

- https://files-org.sentry.io/issues/5560042390/events/f37fce2929aa4e259b7e88e2d51d9d03/

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Utils/FileTags/FileTagsDatabase.cs
+++ b/src/Files.App/Utils/FileTags/FileTagsDatabase.cs
@@ -7,6 +7,7 @@ using Windows.ApplicationModel;
 using JsonSerializer = System.Text.Json.JsonSerializer;
 using static Files.App.Helpers.RegistryHelpers;
 using static Files.App.Utils.FileTags.TaggedFileRegistry;
+using System.Security;
 
 namespace Files.App.Utils.FileTags
 {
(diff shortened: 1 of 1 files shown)


src/Files.App/Utils/FileTags/FileTagsDatabase.cs
```diff
@@ -210,28 +211,32 @@ public string Export()
 
 		private void IterateKeys(List<TaggedFile> list, string path, int depth)
```
hishitetsu, line 212: Isn't there any need to address this on the part of the caller? I'm afraid it will cause problems such as exporting empty settings if we simply ignore the exception.

yaira2: How would you suggest handling?

hishitetsu: Instead of `IterateKeys`, you can catch and ignore SecurityException in each of `GetAll` and `GetAllUnderPath`.
`Export` also calls `IterateKeys`, but the caller of `Export` catches the exception and ignores it, so it is better to throw the exception without catching it here.

hishitetsu (approved):

PR #16091, opened 2024-08-27
Code Quality: Removed Vanara from DriveHelpers

0x5bfa:
### Resolved / Related Issues

- https://github.com/files-community/Files/issues/15000

1. Set a long drive label
2. See it in Files properly shown


src/Files.Core.Storage/Contracts/IDeviceWatcher.cs
```diff
@@ -20,11 +20,6 @@ internal interface IDeviceWatcher
 		/// </summary>
 		event EventHandler<DeviceEventArgs>? ItemDeleted;
 
-		/// <summary>
-		/// Gets invoked when an item changing is detected by the watcher
-		/// </summary>
-		event EventHandler<DeviceEventArgs>? ItemChanged;
```
yaira2, line 26: Was this used for anything?

src/Files.App.Storage/Watchers/DeviceWatcher.cs
```diff
@@ -0,0 +1,148 @@
...
+		private Windows.Devices.Enumeration.DeviceWatcher? _winDeviceWatcher;
+
+		private ManagementEventWatcher? _mmiInsertWatcher, _mmiRemoveWatcher, _mmiModifyWatcher;
+
+		/// <inheritdoc/>
+		public bool CanBeStarted
+			=> _winDeviceWatcher?.Status is DeviceWatcherStatus.Created or DeviceWatcherStatus.Stopped or DeviceWatcherStatus.Aborted;
+
+		/// <inheritdoc/>
+		public event EventHandler<DeviceEventArgs>? ItemAdded;
```
yaira2, line 22: What do item refer to?

src/Files.App/Files.App.csproj
yaira2, line 1: Please revert the changes

src/Files.Core.Storage/Contracts/IDeviceWatcher.cs
0x5bfa: I can't remember why added but I assume there was the same in WindowsDeviceWatcher class, which wraps 'ejected' and 'inserted' events of DeviceManager to 'changed' event.
But I think we shouldn't do that in the watcher and better do that in the service, taking which event it wants.

src/Files.App.Storage/Watchers/DeviceWatcher.cs
0x5bfa: The device, in this context, the drive.
Do you think it should be simply 'Added'? Other watchers have 'Item' prefixes so I'd like to rename all after this PR in another PR.

yaira2: It should be `DeviceAdded`

src/Files.Core.Storage/Contracts/IDeviceWatcher.cs
0x5bfa: Reverted.

src/Files.App.Storage/Watchers/DeviceWatcher.cs
0x5bfa: Done.

src/Files.App/Files.App.csproj
0x5bfa: Done.

0x5bfa:
This has Vanara removal and is the phase 1. Phase 2 is to merge DriveHelpers into the service but I probably won't do that for the time being as this doesn't involve with Vanara removal.

https://github.com/files-community/Files/blob/af07aa37b62c8f4bc0ceda362b092921f3456cb9/src/Files.App/Utils/Storage/Helpers/DriveHelpers.cs#L135

yaira2 (approved):

PR #16098, opened 2024-08-28
Code Quality: Fixed null ref in SetAppThemeMode

yaira2:
I can't reliably reproduce the issue, but I encountered this issue when debugging Files.


hishitetsu (approved):
LGTM

PR #16122, opened 2024-09-02
Feature: Added an optional home button

### LINKED ISSUE #14755: Feature: Support for displaying "home" icon next to navigation buttons

Much like in other browsers, I would like to customize the toolbar. Notably, I would add a "Home" button to quickly go to ... home.

### Requirements

- This proposal will allow better navigation when Files is tucked away on the right side of my screen in a very narrow window (where the folders pane is hidden)

### Files Version

3.2.0.0

### Windows Version

10.0.22631.3155

### Comments

_No response_

yaira2:
**Resolved / Related Issues**

- Closes #14755

1. Confirmed the home button is only enabled when applicable
2. Confirmed the button updates to reflect the setting without a restart
3. Confirmed the button navigates home

### DIFF
--- a/src/Files.App/UserControls/AddressToolbar.xaml
+++ b/src/Files.App/UserControls/AddressToolbar.xaml
@@ -349,6 +349,19 @@
 					ToolTipService.ToolTip="{x:Bind Commands.RefreshItems.LabelWithHotKey, Mode=OneWay}">
 					<FontIcon FontSize="14" Glyph="{x:Bind Commands.RefreshItems.Glyph.BaseGlyph}" />
 				</Button>
+
+				<Button
+					x:Name="HomeButton"
+					x:FieldModifier="public"
+					AccessKey="H"
+					AccessKeyInvoked="Button_AccessKeyInvoked"
+					AutomationProperties.Name="{x:Bind Commands.NavigateHome.Label}"
+					Command="{x:Bind Commands.NavigateHome, Mode=OneWay}"
+					IsEnabled="{x:Bind Commands.NavigateHome.IsExecutable, Mode=OneWay}"
+					Style="{StaticResource AddressToolbarButtonStyle}"
+					ToolTipService.ToolTip="{x:Bind Commands.NavigateHome.LabelWithHotKey, Mode=OneWay}">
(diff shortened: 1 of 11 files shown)


src/Files.App/UserControls/AddressToolbar.xaml
```diff
@@ -349,6 +349,20 @@
 					ToolTipService.ToolTip="{x:Bind Commands.RefreshItems.LabelWithHotKey, Mode=OneWay}">
 					<FontIcon FontSize="14" Glyph="{x:Bind Commands.RefreshItems.Glyph.BaseGlyph}" />
 				</Button>
+
+				<Button
+					x:Name="HomeButton"
+					x:FieldModifier="public"
```
0x5bfa, line 355: What's the reason to make this public?

src/Files.App/UserControls/AddressToolbar.xaml
```diff
@@ -349,6 +349,20 @@
...
+					x:Load="{x:Bind ViewModel.ShowHomeButton, Mode=OneWay}"
```
0x5bfa, line 356: How about using IsExecutable property of NavigateHomeAction and put the user setting condition on it?

yaira2: The action should still be useable even if the home button is hidden.

src/Files.App/UserControls/AddressToolbar.xaml
yaira2: I based this on the other navigation buttons, but I'm not sure why they have it though.

0x5bfa: You can remove public from all.
I think this is a leftover of something.

src/Files.App/UserControls/AddressToolbar.xaml
0x5bfa: Fair enough.

src/Files.App/UserControls/AddressToolbar.xaml
0x5bfa: Also there's one more element that marked as public, TabBar's background l. That is marked as so to be assessable from MainPage to set drag regions. But this can also be improved

yaira2: I pushed a commit removing them

PR #16141, opened 2024-09-05
Fix: Fixed crash when opening folders with invalid git configs

yaira2:
**Resolved / Related Issues**

- Closes https://files-org.sentry.io/issues/5549705044/events/95f583157753427d9838cdb159b84d32/

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Utils/Git/GitHelpers.cs
+++ b/src/Files.App/Utils/Git/GitHelpers.cs
@@ -3,13 +3,11 @@
 
 using Files.App.Dialogs;
 using LibGit2Sharp;
-using Sentry;
 using Microsoft.Extensions.Logging;
+using Sentry;
 using System.Net.Http;
 using System.Net.Http.Json;
-using System.Net.Sockets;
 using System.Text;
-using System.Text.Json;
 
 namespace Files.App.Utils.Git
 {
@@ -146,8 +144,19 @@ namespace Files.App.Utils.Git
 
 		public static async Task<BranchItem?> GetRepositoryHead(string? path)
 		{
-			if (string.IsNullOrWhiteSpace(path) || !Repository.IsValid(path))
+			if (string.IsNullOrWhiteSpace(path))
 				return null;
+
+			// Workaround for https://github.com/libgit2/libgit2sharp/issues/1851
+			try
+			{
+				if (!Repository.IsValid(path))
+					return null;
+			}
+			catch (Exception)
+			{
+				return null;
+			}
 
 			var (_, returnValue) = await DoGitOperationAsync<(GitOperationResult, BranchItem?)>(() =>
 			{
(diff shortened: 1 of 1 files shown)


hishitetsu (commented):
* You can use `SafetyExtensions.IgnoreExceptions` as it returns false when an exception is thrown.
* Is there no need to consider other places that use `Repository.IsValid`?

yaira2:
> You can use SafetyExtensions.IgnoreExceptions as it returns false when an exception is thrown.

We have to return a string so I think try catch makes sense in this case.

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -646,11 +709,17 @@ public static bool IsRepositoryEx(string path, out string repoRootPath)
...
-			if (Repository.IsValid(repositoryRootPath))
+			// Workaround for https://github.com/libgit2/libgit2sharp/issues/1851
+
+			try
 			{
-				repoRootPath = repositoryRootPath;
-				return true;
+				if (Repository.IsValid(repositoryRootPath))
+				{
+					repoRootPath = repositoryRootPath;
```
yaira2, line 718: I would use `SafetyExtensions.IgnoreExceptions` here but it doesn't allow access to `repoRootPath`.

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -99,8 +97,19 @@ private set
...
+			if (string.IsNullOrWhiteSpace(path))
+				return string.Empty;
+
+			// Workaround for https://github.com/libgit2/libgit2sharp/issues/1851
+			try
+			{
+				if (!Repository.IsValid(path))
+					return string.Empty;
+			}
+			catch (Exception)
```
0x5bfa, line 109: Should catch specify Exception class derivatives?

hishitetsu:
> > You can use SafetyExtensions.IgnoreExceptions as it returns false when an exception is thrown.
> 
> We have to return a string so I think try catch makes sense in this case.

I was thnking the following code could be included in the if condition.
`SafetyExtensions.IgnoreExceptions(() => !Repository.IsValid(path))`

yaira2:
I didn't realize that works, it's certainly cleaner that way.

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -78,7 +76,7 @@ private set
 
 			try
 			{
-				if (Repository.IsValid(path))
+				if (SafetyExtensions.IgnoreExceptions(() => Repository.IsValid(path)))
```
hishitetsu, line 79: It could be left as is, but it would be even better to cut it out into a separate method and call it.

yaira2: What's the benefit of doing it like that?

hishitetsu: Don't have to surround each call to `isValid` with `SafetyExtensions.IgnoreExceptions`.

yaira2: Done 👍

hishitetsu (approved):
LGTM👍

PR #16160, opened 2024-09-09
Fix: Fixed NullReferenceException with ShellFileOperations2

### LINKED ISSUE #16153: Bug: NullReferenceException with ShellFileOperations2
Sentry Issue: [FILES-APP-1B](https://files-org.sentry.io/issues/5464643879/?referrer=github_integration)

```
  ?, in void ShellFileOperations2.Dispose(bool disposing)
  ?, in void ShellFileOperations2.Finalize()
```

hishitetsu:
**Resolved / Related Issues**

- Closes #16153 

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Utils/Shell/ShellFileOperations2.cs
+++ b/src/Files.App/Utils/Shell/ShellFileOperations2.cs
@@ -11,7 +11,7 @@ namespace Vanara.Windows.Shell;
 public class ShellFileOperations2 : IDisposable
 {
 	private const OperationFlags defaultOptions = OperationFlags.AllowUndo | OperationFlags.NoConfirmMkDir;
-	private bool disposedValue = false;
+	private int disposedValue = 0;
 	private IFileOperation op;
 	private OperationFlags opFlags = defaultOptions;
 	private HWND owner;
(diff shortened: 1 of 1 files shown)


src/Files.App/Utils/Shell/ShellFileOperations2.cs
```diff
@@ -569,7 +569,7 @@ public void QueueRenameOperation(IEnumerable<ShellItem> sourceItems, string newN
 	/// <param name="disposing"><c>true</c> to release both managed and unmanaged resources; <c>false</c> to release only unmanaged resources.</param>
 	protected virtual void Dispose(bool disposing)
 	{
-		if (!disposedValue)
+		if (Interlocked.CompareExchange(ref disposedValue, 1, 0) == 0)
```
0x5bfa, line 572: How come this has to be thread safe? Just for my knowledge.

hishitetsu: If `Dispose` is called by multiple threads at the same time, the other thread will check `disposedValue` before it becomes true, and the disposal process may be done twice.

0x5bfa: Doesn't this class get instantiated in *a* STA thread?

hishitetsu: But `Dispose` is public and can be called from any thread. And if a NullReferenceException is thrown here, I can only assume it is because non-thread-safe.

0x5bfa: That makes sense. Thanks!!

PR #16163, opened 2024-09-09
Code Quality: Use EncoderReplacementFallback in QuickLook preview provider

yaira2:
**Resolved / Related Issues**

- https://files-org.sentry.io/issues/5525344341/events/a340edbf9c384ccb9675ce5dbf155c37/

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
+++ b/src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
@@ -17,7 +17,6 @@ namespace Files.App.Services.PreviewPopupProviders
 		private static string pipeName = $"QuickLook.App.Pipe.{WindowsIdentity.GetCurrent().User?.Value}";
 		private static string pipeMessageSwitch = "QuickLook.App.PipeMessages.Switch";
 		private static string pipeMessageToggle = "QuickLook.App.PipeMessages.Toggle";
-		private static Encoding encoding = Encoding.GetEncoding("UTF-8", new EncoderReplacementFallback("?"), new DecoderExceptionFallback());
 
 		public async Task TogglePreviewPopupAsync(string path)
 		{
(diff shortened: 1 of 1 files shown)


src/Files.App/Services/PreviewPopupProviders/QuickLookProvider.cs
```diff
@@ -16,6 +17,7 @@ public sealed class QuickLookProvider : IPreviewPopupProvider
 		private static string pipeName = $"QuickLook.App.Pipe.{WindowsIdentity.GetCurrent().User?.Value}";
 		private static string pipeMessageSwitch = "QuickLook.App.PipeMessages.Switch";
 		private static string pipeMessageToggle = "QuickLook.App.PipeMessages.Toggle";
+		private static Encoding encoding = Encoding.GetEncoding("UTF-8", new EncoderReplacementFallback("?"), new DecoderExceptionFallback());
```
0x5bfa, line 20: What's the reason put this here? This should be instantiate right before using and GC'd right after the method finished.

0x5bfa, line 20: What's the reason why this is put in class fields instead of being a local variable?

yaira2: Do you prefer a local variable?

0x5bfa: Don't we have to put this in L61 as well?
Unless, yes.

yaira2: > Don't we have to put this in L61 as well?

It doesn't look like it, the message we send on L62 is very specific and shouldn't have issues. The message on line 42 can have issues though because it can change based on a number of factors.

0x5bfa (approved):

PR #16165, opened 2024-09-09
Fix: Fixed OverflowException in UpdateSelectionSize

### LINKED ISSUE #15500: Bug: BaseLayoutPage.UpdateSelectionSize OverflowException
### Description

**BaseLayoutPage.UpdateSelectionSize ()**
System.OverflowException: Arithmetic operation resulted in an overflow.

### Steps To Reproduce

Unknown

### Requirements

- Resolve crash

### Files Version

v3.4.4

### Windows Version

Windows 11

### Log File

```csharp
System.Linq
Enumerable.Sum[TSource,TResult,TAccumulator] (IEnumerable`1 source, Func`2 selector)
Files.App.Views.Layouts
BaseLayoutPage.UpdateSelectionSize ()
Files.App.Views.Shells
BaseShellPage.FilesystemViewModel_DirectoryInfoUpdated (Object sender, EventArgs e)
System.Threading.Tasks.Task
WinRT
Microsoft.AppCenter.Utils
```

yaira2:
**Resolved / Related Issues**

- Closes #15500

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Data/Models/SelectedItemsPropertiesViewModel.cs
+++ b/src/Files.App/Data/Models/SelectedItemsPropertiesViewModel.cs
@@ -183,8 +183,8 @@ namespace Files.App.Data.Models
 			set => SetProperty(ref isUncompressedItemSizeVisibile, value);
 		}
 
-		private long itemSizeBytes;
-		public long ItemSizeBytes
+		private double itemSizeBytes;
+		public double ItemSizeBytes
 		{
 			get => itemSizeBytes;
 			set => SetProperty(ref itemSizeBytes, value);
--- a/src/Files.App/Data/Models/SelectedItemsPropertiesViewModel.cs
+++ b/src/Files.App/Data/Models/SelectedItemsPropertiesViewModel.cs
@@ -183,8 +183,8 @@ namespace Files.App.Data.Models
 			set => SetProperty(ref isUncompressedItemSizeVisibile, value);
 		}
 
-		private double itemSizeBytes;
-		public double ItemSizeBytes
+		private decimal itemSizeBytes;
+		public decimal ItemSizeBytes
 		{
 			get => itemSizeBytes;
 			set => SetProperty(ref itemSizeBytes, value);
--- a/src/Files.App/Views/Layouts/BaseLayoutPage.cs
+++ b/src/Files.App/Views/Layouts/BaseLayoutPage.cs
@@ -702,7 +702,7 @@ namespace Files.App.Views.Layouts
 			var isSizeKnown = !items.Any(item => string.IsNullOrEmpty(item.FileSize));
 			if (isSizeKnown)
 			{
-				ulong size = items.Aggregate(0UL, (current, item) => current + (ulong)item.FileSizeBytes);
+				decimal size = items.Sum(item => item.FileSizeBytes);
 				SelectedItemsPropertiesViewModel.ItemSizeBytes = (long)size;
 				SelectedItemsPropertiesViewModel.ItemSize = size.ToSizeString();
 			}
--- a/src/Files.App/Views/Layouts/BaseLayoutPage.cs
+++ b/src/Files.App/Views/Layouts/BaseLayoutPage.cs
@@ -703,7 +703,7 @@ namespace Files.App.Views.Layouts
 			if (isSizeKnown)
 			{
 				decimal size = items.Sum(item => item.FileSizeBytes);
-				SelectedItemsPropertiesViewModel.ItemSizeBytes = (double)size;
+				SelectedItemsPropertiesViewModel.ItemSizeBytes = size;
 				SelectedItemsPropertiesViewModel.ItemSize = size.ToSizeString();
 			}
 			else
--- a/src/Files.App/Views/Layouts/BaseLayoutPage.cs
+++ b/src/Files.App/Views/Layouts/BaseLayoutPage.cs
@@ -702,8 +702,8 @@ namespace Files.App.Views.Layouts
 			var isSizeKnown = !items.Any(item => string.IsNullOrEmpty(item.FileSize));
 			if (isSizeKnown)
 			{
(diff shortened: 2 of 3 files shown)


src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -702,8 +702,8 @@ public void UpdateSelectionSize()
 			var isSizeKnown = !items.Any(item => string.IsNullOrEmpty(item.FileSize));
 			if (isSizeKnown)
 			{
-				long size = items.Sum(item => item.FileSizeBytes);
-				SelectedItemsPropertiesViewModel.ItemSizeBytes = size;
+				ulong size = items.Aggregate(0UL, (current, item) => current + (ulong)item.FileSizeBytes);
+				SelectedItemsPropertiesViewModel.ItemSizeBytes = (long)size;
```
0x5bfa, line 706: Is casting here ok when overflow exception is happening?

0x5bfa: I'd suggest 
```suggestion
				SelectedItemsPropertiesViewModel.ItemSizeBytes = size >= (ulong)long.MaxValue ? -1l : (long)size; // NOTE: Suppressed OverflowException in #16165
```
What do you think?

yaira2: Fyi @hishitetsu

hishitetsu: OverflowException may occur even in ulong.
How about using decimal?

yaira2: > OverflowException may occur even in ulong.

True, but it's already very uncommon with long (2 reports) and switching to ulong should take care of it.

yaira2: Does `ToSizeString` work with decimal?

0x5bfa: Hmm, in the first place WIN32_FIND_DATA, which we use for enumeration of items, can only retrieve up to sizeof(ulong)

0x5bfa: Compromise may be necessary.

Btw, should also be updated to ulong.
https://github.com/files-community/Files/blob/8ba4f4517a35e85547a89ec63a08805e00f60965/src/Files.App/Data/Items/ListedItem.cs#L281

hishitetsu: > Does `ToSizeString` work with decimal?

We can cast decimal to double.

hishitetsu: Casting decimal to long may throw an OverflowException.

0x5bfa: What's the reason to do so?
If I understand C#'s value types correctly, ulong can express larger value (20 digits) than double (16 digits). decimal can express much larger (like 29?) than ulong or double though.

0x5bfa: You are right regarding the first part. But isn't the significant number the same even when it comes to integers? Like 1.00 == 100

The max value that can be expressed accurately (according to an article):
Type|Max integer
---|--:
double|`0x20000000000000`
ulong|`0xFFFFFFFFFFFFFFFF`

double.MaxValue is 0x7FEFFFFFFFFFFFFF so even this is less than ulong.MaxValue...

But you seem to understand much better, I'm fine to merge.

0x5bfa: You are right 😄
How large is that digits?

hishitetsu: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types

> ±5.0 × 10^−324 to ±1.7 × 10^308

yaira2: @hishitetsu can you check my latest commit?

hishitetsu: Actually, I haven't been able to check it out yet. I think I will be able to see it tomorrow morning in Japan time.

src/Files.App/Data/Models/SelectedItemsPropertiesViewModel.cs
```diff
@@ -183,8 +183,8 @@ public bool IsUncompressedItemSizeVisibile
 			set => SetProperty(ref isUncompressedItemSizeVisibile, value);
 		}
 
-		private long itemSizeBytes;
-		public long ItemSizeBytes
+		private double itemSizeBytes;
+		public double ItemSizeBytes
```
hishitetsu, line 187:
```suggestion
		private decimal itemSizeBytes;
		public decimal ItemSizeBytes
```

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -702,8 +702,8 @@ public void UpdateSelectionSize()
 			var isSizeKnown = !items.Any(item => string.IsNullOrEmpty(item.FileSize));
 			if (isSizeKnown)
 			{
-				long size = items.Sum(item => item.FileSizeBytes);
-				SelectedItemsPropertiesViewModel.ItemSizeBytes = size;
+				decimal size = items.Sum(item => item.FileSizeBytes);
+				SelectedItemsPropertiesViewModel.ItemSizeBytes = (double)size;
```
hishitetsu, line 706:
```suggestion
				SelectedItemsPropertiesViewModel.ItemSizeBytes = size;
```

hishitetsu (commented):
I think it is better to unify with decimal.

hishitetsu (approved):

PR #16182, opened 2024-09-12
Code Quality: Assigned execution alias and URI for each build

### LINKED ISSUE #13112: Feature: Add app execution aliases and URL schemes for each build channel

0x5bfa:
### Resolved / Related Issues

- Closes #13112

_None_

### DIFF
--- a/.github/scripts/Configure-AppxManifest.ps1
+++ b/.github/scripts/Configure-AppxManifest.ps1
@@ -20,6 +20,9 @@ $nsmgr.AddNamespace("pkg", "http://schemas.microsoft.com/appx/manifest/foundatio
 $nsmgr.AddNamespace("rescap", "http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities")
 $nsmgr.AddNamespace("uap", "http://schemas.microsoft.com/appx/manifest/uap/windows10")
 $nsmgr.AddNamespace("uap5", "http://schemas.microsoft.com/appx/manifest/uap/windows10/5")
+$ap = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap:Extension[@Category='windows.protocol']/uap:Protocol", $nsmgr)
+$aea = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap5:Extension[@Category='windows.appExecutionAlias']/uap5:AppExecutionAlias", $nsmgr)
+$ea = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap5:Extension[@Category='windows.appExecutionAlias']/uap5:AppExecutionAlias/uap5:ExecutionAlias", $nsmgr)
 
 # Update the publisher
 $xmlDoc.Package.Identity.Publisher = $Publisher
@@ -33,10 +36,8 @@ if ($Branch -eq "Preview")
     $xmlDoc.Package.Applications.Application.VisualElements.DefaultTile.ShortName="Files - Preview"
 
     # Update app protocol and execution alias
-    $ap = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap:Extension[@Category='windows.protocol']/uap:Protocol", $nsmgr)
-    $aea = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap5:Extension[@Category='windows.appExecutionAlias']/uap5:AppExecutionAlias/uap5:ExecutionAlias", $nsmgr)
     $ap.SetAttribute("Name", "files-pre");
-    $aea.SetAttribute("Alias", "files-pre.exe");
+    $ea.SetAttribute("Alias", "files-pre.exe");
 
     # Save modified Package.appxmanifest
     $xmlDoc.Save($PackageManifestPath)
@@ -62,10 +63,8 @@ elseif ($Branch -eq "Stable")
     $xmlDoc.Package.Applications.Application.VisualElements.DefaultTile.ShortName="Files"
 
     # Update app protocol and execution alias
-    $ap = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap:Extension[@Category='windows.protocol']/uap:Protocol", $nsmgr)
-    $aea = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap5:Extension[@Category='windows.appExecutionAlias']/uap5:AppExecutionAlias/uap5:ExecutionAlias", $nsmgr)
     $ap.SetAttribute("Name", "files");
-    $aea.SetAttribute("Alias", "files.exe");
+    $aea.RemoveChild(aea.FirstChild); # Avoid duplication
     
     # Save modified Package.appxmanifest
     $xmlDoc.Save($PackageManifestPath)
@@ -95,10 +94,8 @@ elseif ($Branch -eq "Store")
     $xmlDoc.Package.Capabilities.RemoveChild($pm)
 
     # Update app protocol and execution alias
-    $ap = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap:Extension[@Category='windows.protocol']/uap:Protocol", $nsmgr)
-    $aea = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap5:Extension[@Category='windows.appExecutionAlias']/uap5:AppExecutionAlias/uap5:ExecutionAlias", $nsmgr)
     $ap.SetAttribute("Name", "files");
-    $aea.SetAttribute("Alias", "files.exe");
+    $aea.RemoveChild(aea.FirstChild); # Avoid duplication
 
     # Save modified Package.appxmanifest
     $xmlDoc.Save($PackageManifestPath)
--- a/.github/scripts/Configure-AppxManifest.ps1
+++ b/.github/scripts/Configure-AppxManifest.ps1
@@ -36,8 +36,8 @@ if ($Branch -eq "Preview")
     $xmlDoc.Package.Applications.Application.VisualElements.DefaultTile.ShortName="Files - Preview"
 
     # Update app protocol and execution alias
-    $ap.SetAttribute("Name", "files-pre");
-    $ea.SetAttribute("Alias", "files-pre.exe");
+    $ap.SetAttribute("Name", "files-preview");
+    $ea.SetAttribute("Alias", "files-preview.exe");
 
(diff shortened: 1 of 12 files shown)


0x5bfa:
Let me validate changes in the script locally.

.github/scripts/Configure-AppxManifest.ps1
```diff
@@ -11,7 +11,17 @@ param(
...
 
+# Load Package.appxmanifest
 [xml]$xmlDoc = Get-Content $PackageManifestPath
+
+# Add namespaces
+$nsmgr = New-Object System.Xml.XmlNamespaceManager($xmlDoc.NameTable)
+$nsmgr.AddNamespace("pkg", "http://schemas.microsoft.com/appx/manifest/foundation/windows10")
+$nsmgr.AddNamespace("rescap", "http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities")
+$nsmgr.AddNamespace("uap", "http://schemas.microsoft.com/appx/manifest/uap/windows10")
+$nsmgr.AddNamespace("uap5", "http://schemas.microsoft.com/appx/manifest/uap/windows10/5")
```
yaira2, line 22: What are these for?

0x5bfa: You have to add namespaces to select nodes.

src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
```diff
@@ -27,7 +27,19 @@ public sealed class SystemTrayIcon : IDisposable
 
 		// Fields
 
-		private readonly static Guid _trayIconGuid = new("684F2832-AC2B-4630-98C2-73D6AEBD46B7");
+		private readonly static Guid _trayIconGuid = new(
```
yaira2, line 30: Are these random values? Can they be a little more specific?

src/Files.App (Package)/Package.appxmanifest
```diff
@@ -119,12 +119,14 @@
...
                 <uap:Extension Category="windows.protocol">
-                    <uap:Protocol ReturnResults="none" Name="files-uwp" />
+                    <!--  This value changes based on the build branch (files, files-pre, files-dev).  -->
+                    <uap:Protocol ReturnResults="none" Name="files-dev" />
                 </uap:Extension>
 
                 <uap5:Extension Category="windows.appExecutionAlias">
                     <uap5:AppExecutionAlias>
-                        <uap5:ExecutionAlias Alias="files.exe" />
+                        <!--  This value changes based on the build branch (files, files-pre, files-dev).  -->
```
yaira2, line 128: This will break any existing scripts or shortcuts targeting `files.exe`.

0x5bfa: Yeah for preview and dev builds. Bad?

src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
0x5bfa: Sure, some COM contract GUIDs that are associated with the same program uses almost the same GUID but the last part.

For example, `684F2832-AC2B-4630-98C2-000000000001`.
I'm not sure this would break something.

src/Files.App (Package)/Package.appxmanifest
yaira2: I don't think there is a way around this. @hishitetsu

src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
yaira2: Does they need to be numbers?

src/Files.App (Package)/Package.appxmanifest
hishitetsu: How about having both an alias that specifies a build and an alias that does not?

yaira2: I wasn't aware of that option. Good solution!

src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
0x5bfa: Has to be a GUID since this is used by a native API to distinguish the existing tray icons and avoid creating the same icon.

src/Files.App (Package)/Package.appxmanifest
0x5bfa: Indeed. Let me add a static one

src/Files.App/Utils/Taskbar/SystemTrayIcon.cs
yaira2: In that case I would follow your suggestion to use similar GUIDs.

src/Files.App (Package)/Package.appxmanifest
0x5bfa: Btw, this PR was actually to cope with an issue where the alias is overridden when another Files build installed. So that script, if any, would result in malfunction and thus I don't think letting script writers to use ambiguous alias.

0x5bfa: @yaira2

yaira2: This is for backwards compatibility.

.github/scripts/Configure-AppxManifest.ps1
```diff
@@ -53,18 +91,29 @@ elseif ($Branch -eq "Store")
     $xmlDoc.Package.Applications.Application.VisualElements.DefaultTile.ShortName="Files"
 
     # Remove an capability that is used for the sideload
```
yaira2, line 93:
```suggestion
    # Removes a capability that is only used for the sideload package
```

.github/scripts/Configure-AppxManifest.ps1
```diff
@@ -53,18 +91,29 @@ elseif ($Branch -eq "Store")
...
-    $nsmgr = New-Object System.Xml.XmlNamespaceManager($xmlDoc.NameTable)
-    $nsmgr.AddNamespace("pkg", "http://schemas.microsoft.com/appx/manifest/foundation/windows10")
-    $nsmgr.AddNamespace("rescap", "http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities")
```
yaira2, line 58: Was this meant to be removed?

.github/scripts/Configure-AppxManifest.ps1
```diff
@@ -53,18 +91,29 @@ elseif ($Branch -eq "Store")
...
     $pm = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Capabilities/rescap:Capability[@Name='packageManagement']", $nsmgr)
     $xmlDoc.Package.Capabilities.RemoveChild($pm)
+
+    # Update app protocol and execution alias
+    $ap = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap:Extension[@Category='windows.protocol']/uap:Protocol", $nsmgr)
+    $aea = $xmlDoc.SelectSingleNode("/pkg:Package/pkg:Applications/pkg:Application/pkg:Extensions/uap5:Extension[@Category='windows.appExecutionAlias']/uap5:AppExecutionAlias/uap5:ExecutionAlias", $nsmgr)
+    $ap.SetAttribute("Name", "files");
+    $aea.SetAttribute("Alias", "files.exe");
```
yaira2, line 101: Considering we've decided to keep `files.exe` for backwards compatibility, will this cause issues?

.github/scripts/Configure-AppxManifest.ps1
0x5bfa: Moved to the outside if-else

.github/scripts/Configure-AppxManifest.ps1
0x5bfa: Maybe, maybe not. If that might happen we could remove that in Stable/Store release,

yaira2: I think we need to remove it for the stable release

0x5bfa: Done. Btw I renamed it to files-dev, files-preview and files instead of files-uwp. This alone doesn't need backward compatibility?

yaira2: Unless @hishitetsu sees an issue with it, I think `files-uwp` can be removed completely.

hishitetsu: I think `files-uwp` should also be kept for backward compatibility. It is not a public feature, but someone may have written scripts using this URI scheme.
Also, I think `files` is a too generic term for a URI scheme and may be confused with `file` URI scheme. So I think `files-app`, `files-app-stable`, `files-app-preview`, `files-app-dev` are better.

yaira2: I personally prefer `files`, as it's short and easy to type, but I'm not set on it.

0x5bfa: Easy to change before we merge this.
I'll test this again later.

hishitetsu: > Also, I'm not sure we can set more than one protocol.

I've confirmed that it works fine even if the app has mutiple protocols.

I believe each build should have these aliases and URI schemes.
* Alias

|Build|files.exe|files-stable.exe|files-preview.exe|files-dev.exe|
|---|---|---|---|---|
|Stable|o|o| | |
|Preview|o| |o| |
|Dev|o| | |o|

* URI scheme

|Build|files:|files-uwp:|files-stable:|files-preview:|files-dev:|
|---|---|---|---|---|---|
|Stable|o|o|o| | |
|Preview|o|o| |o| |
|Dev|o|o| | |o|

0x5bfa:
PTAL

src/Files.App (Package)/Package.appxmanifest
```diff
@@ -119,12 +119,26 @@
                 </uap3:Extension>
 
                 <uap:Extension Category="windows.protocol">
+                    <!--  This is kept for backward compatibility.  -->
                     <uap:Protocol ReturnResults="none" Name="files-uwp" />
                 </uap:Extension>
 
+                <uap:Extension Category="windows.protocol">
+                    <!--  This is kept for backward compatibility.  -->
+                    <uap:Protocol ReturnResults="none" Name="files" />
```
yaira2, line 128: Can this be moved to line 124?

0x5bfa: According to the docs, it indicates you can't put more than one child element to uap:Extension element.

yaira2 (approved):

hishitetsu (approved):
LGTM

PR #16197, opened 2024-09-15
Code Quality: Update `.gitignore`

Lamparter:
Updated the `.gitignore` file with the one in the head https://github.com/github/gitignore repo
This is not a strictly necessary change but contains some exclusions that add to the code quality.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,9 +7,6 @@
 Files.App.OpenDialog*.dll
 Files.App.SaveDialog*.dll
 Files.App.Launcher.exe
-CustomOpenDialog*.dll
-CustomSaveDialog*.dll
-FilesLauncher.exe
 Files.Extensions/nul
 
 # User-specific files
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@
 Files.App.OpenDialog*.dll
 Files.App.SaveDialog*.dll
 Files.App.Launcher.exe
-Files.Extensions/nul
 
 # User-specific files
 *.rsuser
--- a/.gitignore
+++ b/.gitignore
@@ -202,8 +202,6 @@ publish/
 # in these scripts will be unencrypted
 PublishScripts/
 
-# NuGet Packages
-*.nupkg
 # NuGet Symbol Packages
 *.snupkg
 # The packages folder can be ignored because of Package Restore
--- a/.gitignore
+++ b/.gitignore
@@ -7,8 +7,13 @@
 Files.App.OpenDialog*.dll
 Files.App.SaveDialog*.dll
 Files.App.Launcher.exe
+CustomOpenDialog*.dll
+CustomSaveDialog*.dll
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -7,8 +7,13 @@
 Files.App.OpenDialog*.dll
 Files.App.SaveDialog*.dll
 Files.App.Launcher.exe
+CustomOpenDialog*.dll
+CustomSaveDialog*.dll
+FilesLauncher.exe
```
yaira2, line 12: Why is this excluded?

0x5bfa: Yes, those aren't used any more.

Lamparter: Yeah I was going to say these used to be excluded

.gitignore
```diff
@@ -7,8 +7,10 @@
 Files.App.OpenDialog*.dll
 Files.App.SaveDialog*.dll
 Files.App.Launcher.exe
+Files.Extensions/nul
```
0x5bfa, line 10:
```suggestion
```

.gitignore
```diff
@@ -181,6 +204,9 @@ publish/
 PublishScripts/
 
 # NuGet Packages
+*.nupkg
```
yaira2, line 207: We have a package in our source code `SevenZipSharp.1.0.2.nupkg` so it's best to avoid this change.

.gitignore
yaira2: What is this?

0x5bfa: I have no idea.

Lamparter: It's being excluded right now. https://github.com/files-community/Files/blob/a8942f0341da13b0760de37fcdfefe6828fefe3b/.gitignore#L334
Two years ago I would've been able to comprehend the Files codebase, but I'm starting to lose that ability... 🫠

.gitignore
Lamparter: Why isn't this on NuGet? Isn't this deprecated?

0x5bfa: We use [customized version of SevenZipSharp](https://github.com/files-community/SevenZipSharp).

.gitignore
0x5bfa: > Files.Extensions

This namespace didnt existed when i started contributing. I remeber there was but it must be a few years ago.

.gitignore
Lamparter: > We use [customized version of SevenZipSharp](https://github.com/files-community/SevenZipSharp).

Oh I see, I'll remove it from the ignore - why don't you publish this library to NuGet?

.gitignore
Lamparter: Yeah me too

.gitignore
Lamparter: Wouldn't it be easier to update then?

yaira2 (approved):

PR #16102, opened 2024-08-29
Fix: Fixed issues with opening certain file formats

yaira2:
**Resolved / Related Issues**

To prevent extra work, all changes to the Files codebase must link to an approved issue marked as `Ready to build`. Please insert the issue number following the hashtag with the issue number that this Pull Request resolves.
- Closes #12950
- Closes #15164

**Steps used to test these changes**

Stability is a top priority for Files and all changes are required to go through testing before being merged into the repo. Please include a list of steps that you used to test this PR.

1. Created a zip file named "a.zip"
2. Created a text file named "a.zip.txt"
3. Set default text editor to VS Code and confirmed that the file opened correctly
4. Confirmed the python script @Josh65-2201 created outputs the correct folder

PR #16232, opened 2024-09-19
Fix: Fixed issue with navigating photos on Windows 10

yaira2:
**Resolved / Related Issues**

- Partial revert #16102

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Helpers/Navigation/NavigationHelpers.cs
+++ b/src/Files.App/Helpers/Navigation/NavigationHelpers.cs
@@ -5,8 +5,11 @@ using Files.Shared.Helpers;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media;
 using Microsoft.UI.Xaml.Media.Imaging;
+using Windows.Storage.Search;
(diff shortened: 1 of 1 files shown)


yaira2:
Fyi @hishitetsu, it turns out the NeighboringFilesQuery still works on Windows 10 so I added the conditional check I mentioned here https://github.com/files-community/Files/pull/16102#issuecomment-2318862961

yaira2:
Should continue to function the way it did before #16102. The truth is #16102 was really only needed to fix a couple of niche cases, none of which were known to affect photos.

hishitetsu (approved):
> Should continue to function the way it did before https://github.com/files-community/Files/pull/16102. The truth is https://github.com/files-community/Files/pull/16102 was really only needed to fix a couple of niche cases, none of which were known to affect photos.

I'm wondering if that niche case applies to other photo apps, but there are no code issues, so it is fine to see how it goes.

yaira2:
That's a fair point, we'll have to watch for feedback and adjust our approach if necessary.

Issue #16071, opened 2024-08-25
Code Quality: Dispose COM pointers automatically using IDisposable


0x5bfa:
### Description

I sometimes unbelievably forget to dispose unmanaged resources. If you do that for COM interfaces, it might cause a severe issue leaking memory.

Struct impl:
```c#
public unsafe struct ComPtr<T> : IDisposable
    where T : unmanaged, IUnknown
{
    private _ptr;

    public ComPtr(T* ptr)
    {
        _ptr = ptr;
    }

    public Dispose()
    {
        _ptr->Release();
    }
}
```

Usage:
```c#
public void Do()
{
    try
    {
        using ComPtr<IFileOperation> pFileOperation = default;

        HRESULT hRes = CoCreateInstance(..., *(void**)pFileOperation.GetAddressOf);
    }
    catch (Exception ex)
    {
    }
}
``` 

### Concerned code

All com interface pointers.

### Gains

- prevent memory leak

### Requirements

- Implement a struct that implements IDisposable and use it through using statement

0x5bfa:
This was implemented in #15911 and it works well.
I'm considering cherry picking into another PR to merge it first.

0x5bfa:
I'll consider to use [IDynamicInterfaceCastable](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.idynamicinterfacecastable?view=net-8.0) that would supersede ComQIPtr after .NET 9 comes out.

In the meantime, I'd propose:

```c#
readonly ComPtr<U> QueryFor<U>() where U : unmanaged, IComIID
{
    ComPtr<U> ptr = default;
    _ptr->QueryInterface(&U.Guid, (void**)ptr.GetAddressOf());
    return ptr;
}

// or

readonly HRESULT QueryFor<U>(U** ptr) where U : unmanaged, IComIID
{
    _ptr->QueryInterface(&U.Guid, (void**)ptr);
}
```

```c#
using ComPtr<IShellItem> pShellItem = default;
using ComPtr<IShellItem2> pShellItem2 = pShellItem.QueryFor<IShellItem2>();

// or

using ComPtr<IShellItem> pShellItem = default;
using ComPtr<IShellItem2> pShellItem2 = default;
pShellItem.QueryFor<IShellItem2>(pShellItem2.GetAddressOf());
```

Not sure which is better for now, but I'll implement either of them.

PR #16237, opened 2024-09-21
Code Quality: Introduced ComHeapPtr

0x5bfa:
### Resolved / Related Issues

- Related #16071
- Related #15000

_None_

### DIFF
--- a/src/Files.App.CsWin32/Windows.Win32.ComHeapPtr.cs
+++ b/src/Files.App.CsWin32/Windows.Win32.ComHeapPtr.cs
@@ -9,7 +9,7 @@ using Windows.Win32.System.Com;
 namespace Windows.Win32
 {
 	/// <summary>
-	/// Contains a COM heap-allocated pointer and a set of methods to work with the pointer safely.
+	/// Contains a heap pointer allocated via CoTaskMemAlloc and a set of methods to work with the pointer safely.
 	/// </summary>
 	public unsafe struct ComHeapPtr<T> : IDisposable where T : unmanaged
 	{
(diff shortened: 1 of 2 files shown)


0x5bfa:
As far as we enable friendlyOverloads and useSafeHandles we don't have to worry about handles and we should use SafeHandle derivatives.

@hez2010 fyi.

0x5bfa:
@hez2010 Let me know what you think of ComQIPtr which automatically QI's when initialized with a existing com object and disposes when no longer needed, when you got some time to review this:)

```c#
ComPtr<IShellItem> pShellItem = default;
...
ComQIPtr<IShellItem2> pShellItem2 = new(pShellItem);
```

src/Files.App.CsWin32/Windows.Win32.ComHeapPtr.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+	/// <summary>
+	/// Contains a heap pointer allocated via CoTaskMemAlloc and a set of methods to work with the pointer safely.
+	/// </summary>
+	public unsafe struct ComHeapPtr<T> : IDisposable where T : unmanaged
+	{
+		private T* _ptr;
+
+		public bool IsNull
+			=> _ptr == default;
```
hez2010, line 19: You can use `null` here.

hez2010 (approved):
Looks great.

src/Files.App.CsWin32/Windows.Win32.ComHeapPtr.cs
0x5bfa: Which is good?

0x5bfa: Which is better?

PR #16277, opened 2024-09-29
Fix: Fixed an issue when navigating the sidebar with the tab key

### LINKED ISSUE #16145: Bug: Focus indicator not displayed when navigating in the sidebar using keyboard
### Description

Navigate to the sidebar using `Tab`, then use up and down arrow keys to switch between items. It can observed that the focus indicator disappears and won't show until `Tab` is pressed again.

This is a regression issue introduced by #15949.

https://github.com/user-attachments/assets/da3ef91a-6a23-4531-ad2f-c5373df6f369

### Steps To Reproduce

As mentioned in the description.

### Requirements

The focus indicator should display

### Files Version

3.7.0.0

### Windows Version

10.0

### Log File

Not applicable

karman1111:
**Resolved / Related Issues**

> **Note:** The solution to the issue was a simple change: setting the `IsTabStop` attribute to `False`. This was necessary as we were excluding focus from each selected element.

- Closes #16145 

1. Opened Files
2. Press "Tab" (You can see focused element)
3. Use the 'Down' and 'Up' arrow keys (you won't be able to see the focused element, but it will change. When you press 'Tab' again, you will see the border on the element where the focus is. If you try to move to the next element, it will change, but without a visible border)."

### DIFF
--- a/src/Files.App/UserControls/Sidebar/SidebarItem.cs
+++ b/src/Files.App/UserControls/Sidebar/SidebarItem.cs
@@ -75,7 +75,7 @@ namespace Files.App.UserControls.Sidebar
 				border.DragOver += ItemBorder_DragOver;
 				border.Drop += ItemBorder_Drop;
 				border.AllowDrop = true;
-				border.IsTabStop = true;
+				border.IsTabStop = false;
 			}
 
 			if (GetTemplateChild("ChildrenPresenter") is ItemsRepeater repeater)


yaira2:
Is `ElementGrid` still a tab stop? We should confirm that only one element is a top stop.

karman1111:
Honestly, I'm not entirely sure I understood the question, but recorded a video to demonstrate it.

https://github.com/user-attachments/assets/21313281-61f3-41c9-9456-4f914504e07d

yaira2:
The border contains a grid, we need to confirm that only one of the elements is treated as a tab stop.

yaira2 (approved):
LGTM

PR #16361, opened 2024-10-18
GitHub: Make C++ headers actually detect as C++

Lamparter:
Absolutely _miniscule_ change, but this makes the C++ header files in the Visual C++ projects detected as C++ rather than C as GitHub Linguist assumes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -61,3 +61,5 @@
 #*.PDF   diff=astextplain
 #*.rtf   diff=astextplain
 #*.RTF   diff=astextplain
+
+*.h linguist-language=C++


0x5bfa:
What if we'd be using C? Better to rename hpp.

Lamparter:
@0x5bfa does the team intend to use C?

0x5bfa:
For these? No, it's hard to utilize com from c. I was referring to 'maybe' in the future. (Though I'm pretty sure we should be able to merge these c++ projects into Files.App, registering com in the appxmanifest and opening a window from com classes)

Btw what's the reason to make such change? It doesn't matter and no body cares.

Lamparter:
> Btw what's the reason to make such change? It doesn't matter and no body cares.

Yes, you're absolutely right. It was just a small thing that bugged me after a long time of looking at the repo...

But I don't see what's wrong with this PR since it wasn't a lot of effort to fix a small issue.

0x5bfa:
It's hard for me to approve and rather don't agree because .h file is C and .hpp is specifically for C++.
@yaira2 i'll leave to you.

yaira2:
I don't mind either way.

0x5bfa (approved):

yaira2:
LGTM

PR #15652, opened 2024-06-23
GitHub: Added README.md to src and tests directory

0x5bfa:
## Summary

Added README.md to ‘src’ and ‘tests’ dir for new contributors to be able to participate more easily.

### Resolved / Related Issues

*Planned internally*

**Please proofread.**

### Preview

- https://github.com/0x5bfa/Files/tree/5bfa/Docs-SourceReadme/src
- https://github.com/0x5bfa/Files/tree/5bfa/Docs-SourceReadme/tests

PR #16363, opened 2024-10-18
Code Quality: Document entire repository

Lamparter:
_Similar_ to #15652, I've taken the time to document the entire repository.
However, unlike the previous PR, this is far less descriptive codebase documentation and so is far less likely to become outdated.

To be honest, I wonder whether this might be more viable:
> I guess we should xml comments and generate api docs from it (I know there's a tool for it).

That might be a good idea, but it's down to @yair100 to express his plans for documentation.

### DIFF
--- a/builds/README.md
+++ b/builds/README.md
@@ -1,0 +1,3 @@
+# `builds`
+
+This folder contains Files build pipelines that are used by Azure DevOPS to seamlessly build the app and deploy it.
--- a/src/Files.App (Package)/README.md
+++ b/src/Files.App (Package)/README.md
@@ -1,0 +1,3 @@
+# `src/Files.App (Package)`
+
+This folder contains the Windows Application Packaging project that is used to deploy Files as an MSIX file and package it for association with the Microsoft Store.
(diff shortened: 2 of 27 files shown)


Jay-o-Way:
Gotta appreciate the effort. Just a nit: I read that it's unnecessary c.q. unrecommended to mix headers and formatting.

Lamparter:
> unrecommended to mix headers and formatting.

@Jay-o-Way what do you mean?

Jay-o-Way:
I mean using `#` _and_ `code` formatting
```md
# `text`
```

Lamparter:
Ah yes. I just used it to specify the directory.

0x5bfa:
So just let you know, I disagree with this idea. No one sees Strings dir, for example. Files codebase follows a normal MVVM design pattern, anyone can understand what's in these folders. Just the source is complicated, not the structure.

yaira2:
Let's discuss this proposal in #16364

0x5bfa:
@yaira2 @Lamparter i think we can close this for now, we haven't made a solid decision how to document the source code.

PR #16373, opened 2024-10-21
feat: add readme

jaydesl:


Lamparter:
There is already a `README.md` file inside the `.github` directory; it is moved there for simplicity.

PR #16382, opened 2024-10-24
Code Quality: Introduced ComPtr.As() to QI easily

0x5bfa:
### Resolved / Related Issues

- Related #16071
- Related #15000

> [!NOTE]
> While I settled on this method to QI for now, we may want to use IDynamicInterfaceCastable, which is Hez let me know, after .NET 9 (stack allocation for classes) becomes GA. I need more exploration but this enables us to utilize `is` and `as` with QI called implicitly.

_None_

### DIFF
--- a/src/Files.App.CsWin32/Windows.Win32.ComHeapPtr.cs
+++ b/src/Files.App.CsWin32/Windows.Win32.ComHeapPtr.cs
@@ -16,7 +16,7 @@ namespace Windows.Win32
 		private T* _ptr;
 
 		public bool IsNull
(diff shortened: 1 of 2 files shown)


0x5bfa:
While I'm gonna use this method to QI for the time being, we may want to use IDynamicInterfaceCastable after .NET 9 (stack allocation for classes) becomes GA. I need more exploration but this enables us to utilize `is` and `as` with QI called implicitly.

yaira2 (approved):

PR #16438, opened 2024-11-04
Feature: Added support for creating alternate data streams

### LINKED ISSUE #16337: Feature: Create alternate data streams

I noticed that Files could view and delete alternate data streams, so it could be useful to be able to **create** alternate data streams too.
The user could right-click an existing file, click `create alternate data stream`, the user would choose a name (`[orignal file name, not editable]:[file name chosen by user]`).

### Requirements

- Add a `Create alternate data stream` button when right-clicking on a file

### Files Version

3.7.11.0

### Windows Version

10.0.22631.4317

### Comments

_No response_

yaira2:
**Resolved / Related Issues**

- Closes #16337

1. Confirmed the option is disabled in the root path
2. Confirmed the feature works for single and multiple items
3. Confirmed the menu option is only shown when the setting is enabled

### DIFF
--- a/src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
+++ b/src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
@@ -1,10 +1,8 @@
 ﻿// Copyright (c) 2024 Files Community
 // Licensed under the MIT License. See the LICENSE.
 
-using Files.App.Dialogs;
 using Microsoft.UI.Xaml.Controls;
 using System.IO;
-using TagLib.Ape;
 using Windows.Foundation.Metadata;
 
 namespace Files.App.Actions
--- a/src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
+++ b/src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
@@ -74,6 +74,9 @@ namespace Files.App.Actions
 				if (isDateOk)
 					Win32Helper.SetFileDateModified(selectedItem.ItemPath, dateModified);
 			}));
+
+			if (context.ShellPage is not null)
+				await context.ShellPage.Refresh_Click();
 		}
 
 		private void Context_PropertyChanged(object? sender, PropertyChangedEventArgs e)
--- a/src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
+++ b/src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
@@ -11,6 +11,8 @@ namespace Files.App.Actions
 	{
 		private readonly IContentPageContext context;
 
+		private static readonly IFoldersSettingsService FoldersSettingsService = Ioc.Default.GetRequiredService<IFoldersSettingsService>();
+
 		public string Label
 			=> "CreateAlternateDataStream".GetLocalizedResource();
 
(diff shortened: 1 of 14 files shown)


src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
```diff
@@ -0,0 +1,90 @@
...
+		public string Label
+			=> "CreateAlternateDataStream".GetLocalizedResource();
+
+		public string Description
+			=> "CreateAlternateDataStreamDescription".GetLocalizedResource();
+
+		public override bool IsExecutable =>
+			context.HasSelection &&
+			context.CanCreateItem &&
+			(context.ShellPage?.ShellViewModel.WorkingDirectory != Path.GetPathRoot(context.ShellPage?.ShellViewModel.WorkingDirectory)) &&
```
hishitetsu, line 23: Is this condition necessary?

src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
```diff
@@ -0,0 +1,90 @@
...
+				if (isReadOnly)
+					Win32Helper.SetFileAttribute(selectedItem.ItemPath, System.IO.FileAttributes.ReadOnly);
+
+				// Restore date modified
+				if (isDateOk)
+					Win32Helper.SetFileDateModified(selectedItem.ItemPath, dateModified);
+			}));
+
+			if (context.ShellPage is not null)
+				await context.ShellPage.Refresh_Click();
```
hishitetsu, line 76: I don't know how to fix it, but if you fix ShellViewModel.WatchForDirectoryChanges to work with the addition of alternate data streams, you won't need to refresh the page here.
At least, refreshing should only be done if the setting to display alternate data streams is on.

src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
yaira2: It's to account for https://github.com/files-community/Files/issues/16337#issuecomment-2452221219

src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
yaira2: If that can be figured out, that would be ideal. In the meantime, I'll follow your suggestion to only refresh when the setting is enabled.

src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
hishitetsu: Then this condition should be deleted, because Create ADS will not appear when right-clicking on an empty space even without this condition,
This condition prevents the creation of ADSs for items in the root directory.

src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
```diff
@@ -0,0 +1,115 @@
...
+			else
+			{
+				var dialog = new ContentDialog
+				{
+					Title = Strings.DataStreamsAreHiddenTitle.GetLocalizedResource(),
+					Content = Strings.DataStreamsAreHiddenDescription.GetLocalizedResource(),
+					PrimaryButtonText = Strings.Yes.GetLocalizedResource(),
+					SecondaryButtonText = Strings.No.GetLocalizedResource()
+				};
```
hishitetsu, line 90: Isn't it necessary to add an option to not notify again?

src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
yaira2: Fair enough

src/Files.App/Actions/FileSystem/CreateAlternateDataStreamAction.cs
yaira2: Resolved in [801a858](https://github.com/files-community/Files/pull/16438/commits/801a8588357d8147779c6bc4db50aae7ed1f7496)

hishitetsu (approved):
LGTM

yaira2:
@hishitetsu thank you for the review.

PR #16480, opened 2024-11-18
Code Quality: Improved SetAs actions

0x5bfa:
### Resolved / Related Issues

- Annotated that the service methods may throw COMException
- Throw COMException when failed in all the com methods
- Release the PIDLs properly
- Use stackalloc

1. Set as wallpaper
2. Set as slideshow
3. Set as app background


src/Files.App/Actions/Content/Background/SetAsLockscreenBackgroundAction.cs
```diff
@@ -18,20 +20,21 @@ public override RichGlyph Glyph
...
 
 			try
 			{
-				return WindowsWallpaperService.SetLockScreenWallpaper(context.SelectedItem.ItemPath);
+				return WindowsWallpaperService.SetLockScreenWallpaper(ContentPageContext.SelectedItem.ItemPath);
 			}
 			catch (Exception ex)
 			{
 				ShowErrorDialog(ex.Message);
+				App.Logger.LogWarning(ex, ex.Message);
```
yaira2, line 37: I don't think logging helps us because there isn't anything to do for this type of exception. The best thing is to notify the user and prevent a crash.

src/Files.App/Actions/Content/Background/SetAsSlideshowBackgroundAction.cs
```diff
@@ -18,11 +20,14 @@ public override RichGlyph Glyph
 
 		public override bool IsExecutable =>
 			base.IsExecutable &&
-			context.SelectedItems.Count > 1;
+			ContentPageContext.SelectedItems.Count > 1;
 
 		public override Task ExecuteAsync(object? parameter = null)
 		{
-			var paths = context.SelectedItems.Select(item => item.ItemPath).ToArray();
+			if (ContentPageContext.SelectedItems.Count <= 1)
```
yaira2, line 27: Is there a scenario where this is ever true?

src/Files.App/Actions/Content/Background/SetAsWallpaperBackgroundAction.cs
```diff
@@ -18,20 +18,21 @@ public override RichGlyph Glyph
...
 
 			try
 			{
-				WindowsWallpaperService.SetDesktopWallpaper(context.SelectedItem.ItemPath);
+				WindowsWallpaperService.SetDesktopWallpaper(ContentPageContext.SelectedItem.ItemPath);
 			}
 			catch (Exception ex)
 			{
 				ShowErrorDialog(ex.Message);
+				App.Logger.LogWarning(ex, ex.Message);
```
yaira2, line 35: I don't think logging helps us because there isn't anything to do for this type of exception. The best thing is to notify the user and prevent a crash.

src/Files.App/Services/Windows/WindowsWallpaperService.cs
```diff
@@ -18,27 +18,31 @@ public sealed class WindowsWallpaperService : IWindowsWallpaperService
...
+			// Get total count of all monitors available
+			hr = pDesktopWallpaper.Get()->GetMonitorDevicePathCount(out var dwMonitorCount)
+				.ThrowOnFailure();
 
 			fixed (char* pszPath = szPath)
 			{
 				PWSTR pMonitorID = default;
 
-				for (uint dwIndex = 0; dwIndex < dwMonitorCount; dwIndex++)
+				// Set the selected image file as wallpaper to all monitors available
```
yaira2, line 40:
```suggestion
				// Set the selected image file as wallpaper for all available monitors
```

src/Files.App/Actions/Content/Background/SetAsSlideshowBackgroundAction.cs
0x5bfa: I dont know, just doing the same as other actions. Like, checking if `if (ContentPageContext.SelectedItem is null)` is true (while IsExecutable includes this checking).

src/Files.App/Services/Windows/WindowsWallpaperService.cs
0x5bfa: Done.

src/Files.App/Actions/Content/Background/SetAsSlideshowBackgroundAction.cs
0x5bfa: Done.

src/Files.App/Services/Windows/WindowsWallpaperService.cs
```diff
@@ -18,28 +18,32 @@ public sealed class WindowsWallpaperService : IWindowsWallpaperService
 		public unsafe void SetDesktopWallpaper(string szPath)
 		{
 			using ComPtr<IDesktopWallpaper> pDesktopWallpaper = default;
-			var desktopWallpaperIid = typeof(IDesktopWallpaper).GUID;
-			var desktopWallpaperInstanceIid = typeof(DesktopWallpaper).GUID;
+			var CLSID_DesktopWallpaper = typeof(DesktopWallpaper).GUID;
+			var IID_DesktopWallpaper = typeof(IDesktopWallpaper).GUID;
```
0x5bfa, line 22: @yaira2 we really need RVA Guid gen. FYI @hez2010

You have to put Guid variables into the local variable field first and do `&guid` as above. Plus, besides this IID's and CLSID's Guid issue, what CsWin32 generates for other Guid variables such as BHID are also not on RVA field.
This inconvenience persists everywhere we use COM generated by CsWin32.

I'm probably going to request this in CsWin32 tho, I'd like us to have SG for this as a workaround for now after we adopted .NET 9 (If the author of CsWin32 was quick to fix this, we would go with it ofc tho). I think [this SG](https://github.com/dongle-the-gadget/GuidRVAGen) does exactly what we want:

```c#
// In Files.App.CsWin32/CLSID.cs
public static unsafe partial class CLSID
{
    [RVAGuidGen("00000000-0000-0000-0000-000000000000")]
    public partial Guid* CLSID_DesktopWallpaper { get; }
}
```

The SG generates:
```c#
public static unsafe partial class CLSID
{
    public partial Guid* CLSID_DesktopWallpaper // I guess inlining too
        => (Guid*)Unsafe.AsPointer(ref MemoryMarshal.AsRef([0x00, 0x00, ...]));
}
```

---

Once CsWin32 generates Guid from ROS\<byte\>, we would create an extension method like this:

```c#
// Files.App.CsWin32/Extras.cs

public unsafe Guid* AsPointer(this iid)
{
    // NOTE: This works only with Guid variables stored in ".rodata".
    return (Guid*)Unsafe.AsPointer(ref Unsafe.AsRef(in iid));
}
```

src/Files.App/Actions/Content/Background/SetAsWallpaperBackgroundAction.cs
yaira2: It's not really relevant to users, the reason we have a dialog in the first place is so that the user is informed that the action didn't complete.

src/Files.App/Actions/Content/Background/SetAsAppBackgroundAction.cs
```diff
@@ -18,12 +18,11 @@ public override RichGlyph Glyph
 
 		public override bool IsExecutable =>
 			base.IsExecutable &&
-			context.SelectedItem is not null;
+			ContentPageContext.SelectedItem is not null;
 
 		public override Task ExecuteAsync(object? parameter = null)
 		{
-			if (context.SelectedItem is not null)
```
yaira2, line 25: @hishitetsu can you confirm if this is okay to remove? I'm pretty sure it's needed to prevent a crash.

hishitetsu: Yeah, this code is necessary because the item may be deselected between the time it is checked in `IsExecutable` and the time the action is actually executed.

0x5bfa: Why not use IsExecutable prop instead?

src/Files.App/Actions/Content/Background/SetAsWallpaperBackgroundAction.cs
0x5bfa: It's common to include HResult error code into user-facing dialog though.
I'll remove it then.

src/Files.App/Actions/Content/Background/SetAsAppBackgroundAction.cs
hishitetsu: Certainly, but I noticed we should do the following instead of simply checking to prevent NullReferenceException completely.
```
if (ContentPageContext.SelectedItem is ListedItem selectedItem)
	AppearanceSettingsService.AppThemeBackgroundImageSource = selectedItem.ItemPath;
```

src/Files.App/Actions/Content/Background/SetAsWallpaperBackgroundAction.cs
yaira2: It's okay if we think it'll be helpful

src/Files.App/Actions/Content/Background/SetAsAppBackgroundAction.cs
0x5bfa: Done.

src/Files.App/Actions/Content/Background/SetAsSlideshowBackgroundAction.cs
```diff
@@ -18,14 +20,16 @@ public override RichGlyph Glyph
 
 		public override bool IsExecutable =>
 			base.IsExecutable &&
-			context.SelectedItems.Count > 1;
+			ContentPageContext.SelectedItems.Count > 1;
 
 		public override Task ExecuteAsync(object? parameter = null)
 		{
-			var paths = context.SelectedItems.Select(item => item.ItemPath).ToArray();
+			if (!IsExecutable)
```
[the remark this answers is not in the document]
0x5bfa: I added this per https://github.com/files-community/Files/pull/16480#discussion_r1849428227
Sounds not good to you?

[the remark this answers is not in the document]
yaira2: @hishitetsu is this what you intended?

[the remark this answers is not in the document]
hishitetsu: No. We need to ensure that the `paths` local varibale contains at least one element.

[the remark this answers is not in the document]
0x5bfa: How can it be 0 or null while IsExecutable ensures SelectedItems has at least one element.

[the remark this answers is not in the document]
hishitetsu: Another thread may deselect items between checking 'IsExecutable' and assigning to 'paths'.

0x5bfa:
PTAL

yaira2 (dismissed):

src/Files.App/Actions/Content/Background/SetAsSlideshowBackgroundAction.cs
```diff
@@ -18,11 +20,12 @@ public override RichGlyph Glyph
...
 		public override bool IsExecutable =>
 			base.IsExecutable &&
-			context.SelectedItems.Count > 1;
+			ContentPageContext.SelectedItems.Count > 1;
 
 		public override Task ExecuteAsync(object? parameter = null)
 		{
-			var paths = context.SelectedItems.Select(item => item.ItemPath).ToArray();
+			if (!IsExecutable || ContentPageContext.SelectedItems.Select(item => item.ItemPath).ToArray() is not string[] paths || paths.Length is not 0)
+				return Task.CompletedTask;
```
hishitetsu, line 28:
```suggestion
			if (!IsExecutable || ContentPageContext.SelectedItems.Select(item => item.ItemPath).ToArray() is not string[] paths || paths.Length is 0)
				return Task.CompletedTask;
```

PR #16506, opened 2024-11-21
Feature: Naming a shortcut while creating it

### LINKED ISSUE #16387: Feature: Naming a shortcut while creating it

In File Explorer, when you create a shortcut after you put the path to the shortcut; File Explorer will next prompt you to name the shortcut, it look something like this:
Image

So I was wondering if we can add this to Files, it will make creating and naming shortcuts much easier & more user-friendly.

### Requirements

- Naming a shortcut while creating it.

### Files Version

3.7.11.0

### Windows Version

10.0.22631.4391

### Comments

_No response_

ferrariofilippo:
**Resolved / Related Issues**

- Closes #16387 

1. Click on `New`
2. Click on `Shortcut`
3. Try to break the modal

**Notes**
At the moment the `Name` is updated whenever the destination path changes.
> Or alternatively, we only auto fill the name if the field is empty.

I think this may not work as expected: 
let's say a user selects `SomePath/FileA`, then the name will be set to `File_A`. If now the destination is changed to `SomePath/File_B`, the name should not be updated according to above (because the field would not be empty).

Screenshot 2024-11-21 214958

### DIFF
--- a/src/Files.App/Dialogs/CreateShortcutDialog.xaml
+++ b/src/Files.App/Dialogs/CreateShortcutDialog.xaml
@@ -81,7 +81,6 @@
 				Grid.Row="4"
 				Grid.ColumnSpan="2"
 				HorizontalAlignment="Stretch"
-				IsEnabled="{x:Bind ViewModel.IsLocationValid, Mode=OneWay}"
 				Text="{x:Bind ViewModel.ShortcutName, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}">
 				<TextBox.Resources>
 					<TeachingTip
--- a/src/Files.App/Dialogs/CreateShortcutDialog.xaml
+++ b/src/Files.App/Dialogs/CreateShortcutDialog.xaml
@@ -9,7 +9,7 @@
 	Title="{helpers:ResourceString Name=NewShortcutDialogTitle}"
 	DefaultButton="Primary"
 	HighContrastAdjustment="None"
-	IsPrimaryButtonEnabled="{x:Bind ViewModel.IsLocationValid, Mode=OneWay}"
+	IsPrimaryButtonEnabled="{x:Bind ViewModel.IsShortcutValid, Mode=OneWay}"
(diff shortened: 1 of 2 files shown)


yaira2:
@ferrariofilippo can you add some space between the location text box and the name section?

ferrariofilippo:
Which one is better?

`12dp`
image

`8dp`
image

yaira2:
Fyi @mdtauk

yaira2 (approved):

yaira2:
I think there is an issue with the TeachingTip placement.

image

src/Files.App/Dialogs/CreateShortcutDialog.xaml
```diff
@@ -68,6 +70,27 @@
...
+			<!--  Name Box  -->
+			<TextBox
+				x:Name="ShortcutName"
+				Grid.Row="4"
+				Grid.ColumnSpan="2"
+				HorizontalAlignment="Stretch"
+				Text="{x:Bind ViewModel.ShortcutName, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}">
+				<TextBox.Resources>
+					<TeachingTip
+						x:Name="InvalidNameWarning"
```
yaira2, line 88:
```suggestion
						x:Name="InvalidNameTeachingTip"
```

src/Files.App/ViewModels/Dialogs/CreateShortcutDialogViewModel.cs
```diff
@@ -262,5 +264,34 @@ private async Task CreateShortcutAsync()
 
 			ShortcutCreatedSuccessfully = await FileOperationsHelpers.CreateOrUpdateLinkAsync(filePath, FullPath, Arguments);
 		}
+
+		private void SetDefaultName()
```
yaira2, line 268:
```suggestion
		private void AutoFillName()
```

yaira2 (approved):
LGTM

PR #16508, opened 2024-11-22
Feature: Improved behavior when dragging items from an unfocused Files window

### LINKED ISSUE #13255: Feature: Don't focus Files window when dragging a file/folder

When trying to drag a file (or folder) from a Files window to other window, the Files windows will be focused and automatically jump to the top, covering other window that was originally above. It will be better if the window could stay behind.

It is necessary for the app popping onto the top when clicking on it, or dragging on some empty spaces., title bar, etc.. However, it would be more convenient if it could stay behind when dragging on a single file or the files selected. (That's how windows files explorer behaves on this situation.)

### Requirements

- When dragging an item from an unfocused Files window, the window should not be focused, stay behind, until the item is finally drop to the original window.
- If some of the files have been selected, the window should stay behind when dragging on those files, but get to the top when dragging on other unselected files.

### Files Version

2.5.21.0

### Windows Version

10.0.22631.2262

### Comments

_No response_

Jack251970:
**Resolved / Related Issues**

- Added support for not focusing Files window when dragging a file/folder
- Closes #13255

1. Opened Files and File Explorer.
2. Select files in Files.
3. Move File Explorer on the top, which is topper than Files. Make sure it doesn't cover Files so that you can drag files from Files.
4. Drag the selected files to the File Explorer. And you can find the Files won't move to the top when dragging files.

### DIFF
--- a/src/Files.App.CsWin32/Windows.Win32.Extras.cs
+++ b/src/Files.App.CsWin32/Windows.Win32.Extras.cs
@@ -16,5 +16,51 @@ namespace Windows.Win32
 	{
 		[UnmanagedFunctionPointer(CallingConvention.Winapi)]
 		public delegate LRESULT WNDPROC(HWND hWnd, uint msg, WPARAM wParam, LPARAM lParam);
+
+		/// <summary>Contains information about the size and position of a window.</summary>
+		/// <remarks>
+		/// <para><see href="https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-windowpos">Learn more about this API from docs.microsoft.com</see>.</para>
+		/// </remarks>
+		public partial struct WINDOWPOS
+		{
+			/// <summary>
+			/// <para>Type: <b>HWND</b> A handle to the window.</para>
+			/// <para><see href="https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-windowpos#members">Read more on docs.microsoft.com</see>.</para>
+			/// </summary>
+			internal HWND hwnd;
+
+			/// <summary>
+			/// <para>Type: <b>HWND</b> The position of the window in Z order (front-to-back position). This member can be a handle to the window behind which this window is placed, or can be one of the special values listed with the <a href="https://docs.microsoft.com/windows/desktop/api/winuser/nf-winuser-setwindowpos">SetWindowPos</a> function.</para>
+			/// <para><see href="https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-windowpos#members">Read more on docs.microsoft.com</see>.</para>
+			/// </summary>
+			internal HWND hwndInsertAfter;
+
+			/// <summary>
+			/// <para>Type: <b>int</b> The position of the left edge of the window.</para>
+			/// <para><see href="https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-windowpos#members">Read more on docs.microsoft.com</see>.</para>
+			/// </summary>
+			internal int x;
+
+			/// <summary>
+			/// <para>Type: <b>int</b> The position of the top edge of the window.</para>
+			/// <para><see href="https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-windowpos#members">Read more on docs.microsoft.com</see>.</para>
+			/// </summary>
+			internal int y;
+
+			/// <summary>
+			/// <para>Type: <b>int</b> The window width, in pixels.</para>
+			/// <para><see href="https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-windowpos#members">Read more on docs.microsoft.com</see>.</para>
+			/// </summary>
+			internal int cx;
+
+			/// <summary>
+			/// <para>Type: <b>int</b> The window height, in pixels.</para>
+			/// <para><see href="https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-windowpos#members">Read more on docs.microsoft.com</see>.</para>
+			/// </summary>
+			internal int cy;
+
+			/// <summary>Type: <b>UINT</b></summary>
+			public SET_WINDOW_POS_FLAGS flags;
+		}
 	}
 }
--- a/src/Files.App/Helpers/Win32/Win32Helper.WindowManagement.cs
+++ b/src/Files.App/Helpers/Win32/Win32Helper.WindowManagement.cs
@@ -4,8 +4,10 @@
 using Microsoft.UI.Input;
 using Microsoft.UI.Xaml;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using Windows.Win32;
 using Windows.Win32.UI.WindowsAndMessaging;
+using static Vanara.PInvoke.User32;
 
 namespace Files.App.Helpers
 {
@@ -57,6 +59,17 @@ namespace Files.App.Helpers
 				uiElement,
 				[cursor]
 			);
+		}
+
+		/// <summary>
+		/// Force window to stay at bottom of other upper windows.
+		/// </summary>
+		/// <param name="lParam">The lParam of the message.</param>
+		public static void ForceWindowPosition(nint lParam)
+		{
+			var windowPos = Marshal.PtrToStructure<WINDOWPOS>(lParam);
+			windowPos.flags |= SetWindowPosFlags.SWP_NOZORDER;
+			Marshal.StructureToPtr(windowPos, lParam, false);
 		}
 	}
 }
--- a/src/Files.App/Helpers/Win32/Win32Helper.WindowManagement.cs
+++ b/src/Files.App/Helpers/Win32/Win32Helper.WindowManagement.cs
@@ -7,7 +7,6 @@ using System.Reflection;
 using System.Runtime.InteropServices;
 using Windows.Win32;
 using Windows.Win32.UI.WindowsAndMessaging;
-using static Vanara.PInvoke.User32;
 
 namespace Files.App.Helpers
 {
@@ -68,7 +67,7 @@ namespace Files.App.Helpers
 		public static void ForceWindowPosition(nint lParam)
 		{
 			var windowPos = Marshal.PtrToStructure<WINDOWPOS>(lParam);
-			windowPos.flags |= SetWindowPosFlags.SWP_NOZORDER;
+			windowPos.flags |= SET_WINDOW_POS_FLAGS.SWP_NOZORDER;
 			Marshal.StructureToPtr(windowPos, lParam, false);
 		}
 	}
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -19,6 +19,7 @@ namespace Files.App
 		public static MainWindow Instance => _Instance ??= new();
 
 		public nint WindowHandle { get; }
+		private bool CanWindowToFront { get; set; } = true;
 
 		public MainWindow()
 		{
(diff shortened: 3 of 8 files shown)


src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -1271,15 +1299,52 @@ selectedItems is not null &&
...
+				}
+			}
+		}
+
 		protected void FileListItem_RightTapped(object sender, RightTappedRoutedEventArgs e)
 		{
+			if (!itemDragging)
+			{
+				// No need to bring the window to the front again
+				if (MainWindow.Instance.SetCanWindowToFront(true))
```
yaira2, line 1342: Can these statements be combined?

src/Files.App/Helpers/Win32/Win32Helper.WindowManagement.cs
```diff
@@ -4,8 +4,10 @@
 using Microsoft.UI.Input;
 using Microsoft.UI.Xaml;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using Windows.Win32;
 using Windows.Win32.UI.WindowsAndMessaging;
+using static Vanara.PInvoke.User32;
```
yaira2, line 10: We're trying to avoid new references to Vanara. While it's not absolutely required, it would be preferable if this can be achieved differently. Refer to issue #15000 for more details.

Jack251970: Get it. Finished in 1aa2cd6.

yaira2 (dismissed):

yaira2:
@Jack251970 the behavior is working nicely when Files is maximized. Is there a way to also do this when Files isn't maximized?

Jack251970:
@yaira2  Sorry, I cannot get your meaning. What is the difference between the behaviour when File is maximized and when it is not?

Jack251970:
https://github.com/user-attachments/assets/91b17452-b706-47fe-be16-89460ec12508
I think the behaviour like this is enough compared to system File Explorer?

yaira2:
> @yaira2 Sorry, I cannot get your meaning. What is the difference between the behaviour when File is maximized and when it is not?

I meant that Files is brought to the foreground if I drag a file when the Files window isn't maximized.

Jack251970:
> > @yaira2 Sorry, I cannot get your meaning. What is the difference between the behaviour when File is maximized and when it is not?
> 
> I meant that Files is brought to the foreground if the window isn't maximized.

Sorry, still no idea about why Files needs to be brought to the foreground if the window isn't maximized. (Under what situation?)

As you can see in the video above, my codes implement a function that the Files will stay in the current z index if dragging selected files so that it will have the same experience as system File Explorer.

0x5bfa:
Putting aside the behavior, you can use DidPositionChanged from AppWindow.Changed event instead of utilizing win32 apis.

Jack251970:
> Putting aside the behavior, you can use DidPositionChanged from AppWindow.Changed event instead of utilizing win32 apis.

I am not quite familiar with this api. Will it make some differences?

yaira2:
@Jack251970 I'm referring to the behavior (around two seconds into your recording) where the Files window is brought to the foreground.
1. Open Files regularly (not maximized)
2. Select file
3. Switch focus to File Explorer
4. Start dragging selected file from Files
5. Files is brought to the foreground

Isn't the expected behavior for File Explorer to remain the topmost window?

Jack251970:
> @Jack251970 I'm referring to the behavior (around two seconds into your recording) where the Files window is brought to the foreground.
> 
> 1. Open Files regularly (not maximized)
> 2. Select a file
> 3. Switch the focus to File Explorer
> 4. Start dragging the selected file from Files
> 5. Files is brought to the foreground
> 
> Isn't the expected behavior for File Explorer to remain the topmost window?

Get it. I will try to make this work.

yaira2:
Thanks! I should note, your current changes already helped to resolve this issue when the window is maximized.

0x5bfa:
> Will it make some differences?

Yeah you don't have to hook/unhook win32 window message loop, there's high-level API wrapper called WinRT API as opposed to low-level Win32 API.

This API let you just subscribe an event. Go to MainWindow and subscribe this.Changed event and get e.DidPositionChanged property value.

Jack251970:
> > Will it make some differences?
> 
> Yeah you don't have to hook/unhook win32 window message loop, there's high-level API wrapper called WinRT API as opposed to low-level Win32 API.
> 
> This API let you just subscribe an event. Go to MainWindow and subscribe this.Changed event and get e.DidPositionChanged property value.

Sure, I will have a try.

Jack251970:
@yaira2 Sorry about the misunderstanding description (I have removed the useless step that asks you to maximize the Files window). This feature can work while Files isn't a maximized window.
In my recording around 2 seconds, the reason why Files is brought to the front is that I have selected a new item (09-20-37.mp4) instead of the original one (09-28-05.mp4). This behaviour is just like the system File Explorer.

Jack251970:
@0x5bfa Yeah I have found that AppWindow.Changed event contains information about DidPositionChange & ZOrderChange, but  I cannot handle these changes. (AppWindowChangedEventArgs does not have Handle property, and I cannot stop those changing event)

So I think WindowMessageReceived event in WinUIEx.WindowManager is the only way to implement this and WinUIEx package has provided a good solution for this.

yaira2:
@Jack251970 thank you for clarifying. I'll test again to confirm the behavior works on my side.

0x5bfa:
> Yeah I have found that AppWindow.Changed event contains information about DidPositionChange & ZOrderChange, but I cannot handle these changes. (AppWindowChangedEventArgs does not have Handle property, and I cannot stop those changing event)

Thanks, I missed that. I'll file a feature request on WinAppSdk for this later.

yaira2:
@Jack251970 I can confirm the behavior is working as expected. I'm going to wait for a review from @hishitetsu, but as far as I'm concerned I have no requested changes.

hishitetsu:
This improvement does not work once the app goes to the background and then reopens. 
I also found a few behaviors that differ from Explorer.
* In Files, if you try to drag a file that is not selected, the Files window comes into focus; this is not the case in Explorer.
* Once dragged to another window, returning the cursor and releasing the mouse button on the original window will focus the original window in Explorer, but not in Files.

hishitetsu (commented):
This improvement does not work once the app goes to the background and then reopens. 
I also found a few behaviors that differ from Explorer.
* In Files, if you try to drag a file that is not selected, the Files window comes into focus; this is not the case in Explorer.
* Once dragged to another window, returning the cursor and releasing the mouse button on the original window will focus the original window in Explorer, but not in Files.

Jack251970:
@hishitetsu Thank you for testing this function so carefully!

As for few behaviors that differ from Explorer above, I don not think it is necessary to ensure Files has the same behavior as the Explorer and I wonder what is the difference in user experience. (Sorry, I am not familiar with focus behaviors.)

For `This improvement does not work once the app goes to the background and then reopens`, do you mean close the Files window into tray icon? (I have test it when minimize File window into task bar, and I found it can work.)

hishitetsu:
> can you provide the reproduction steps?

1. Set `Leave app running in the background when the window is closed` option to on.
2. Close (not minimize) the Files window.
3. Click the task tray icon to reopen Files.
4. Test the behavior.

Jack251970:
@hishitetsu Now fixed, I think it can work when you close window and keep it in the background.

yaira2:
> As for few behaviors that differ from Explorer above, I don not think it is necessary to ensure Files has the same behavior as the Explorer and I wonder what is the difference in user experience. (Sorry, I am not familiar with focus behaviors.)

For what it's worth, it's probably better if we only implement the behaviors as they are requested. Thanks for all your work!

yaira2 (approved):

src/Files.App/MainWindow.xaml.cs
```diff
@@ -339,5 +342,28 @@ x.tabItem.NavigationParameter.NavigationParameter is PaneNavigationArguments pan
...
 		}
+
+		public bool SetCanWindowToFront(bool canWindowToFront)
+		{
+			if (CanWindowToFront != canWindowToFront)
+			{
+				CanWindowToFront = canWindowToFront;
+				return true;
+			}
+			return false;
```
hishitetsu, line 353: This does not guarantee atomic operation when multiple threads handle it at the same time. It is better to use `Interlocked` class or `lock` statement.

src/Files.App/MainWindow.xaml.cs
```diff
@@ -339,5 +342,28 @@ x.tabItem.NavigationParameter.NavigationParameter is PaneNavigationArguments pan
...
+		private void WindowManager_WindowMessageReceived(object? sender, WinUIEx.Messaging.WindowMessageEventArgs e)
+		{
+			if (!CanWindowToFront)
+			{
+				if (e.Message.MessageId == WM_WINDOWPOSCHANGING)
+				{
+					Win32Helper.ForceWindowPosition(e.Message.LParam);
+					e.Handled = true;
+				}
+			}
```
hishitetsu, line 366:
```suggestion
			if (!CanWindowToFront && e.Message.MessageId == WM_WINDOWPOSCHANGING)
			{
				Win32Helper.ForceWindowPosition(e.Message.LParam);
				e.Handled = true;
			}
```

hishitetsu (commented):
It seems to work fine, but there are a few comments about the code quality.

Jack251970:
@hishitetsu Resolved.

src/Files.App/MainWindow.xaml.cs
```diff
@@ -339,5 +343,28 @@ x.tabItem.NavigationParameter.NavigationParameter is PaneNavigationArguments pan
...
+            lock (_canWindowToFrontLock)
+            {
+                if (CanWindowToFront != canWindowToFront)
+                {
+                    CanWindowToFront = canWindowToFront;
+                    return true;
+                }
+                return false;
+            }
+        }
```
hishitetsu, line 358: Use tabs for indentation.
```suggestion
		public bool SetCanWindowToFront(bool canWindowToFront)
		{
			lock (_canWindowToFrontLock)
			{
				if (CanWindowToFront != canWindowToFront)
				{
					CanWindowToFront = canWindowToFront;
					return true;
				}
				return false;
			}
		}
```

hishitetsu (approved):
LGTM

src/Files.App.CsWin32/Windows.Win32.Extras.cs
```diff
@@ -16,5 +16,51 @@ namespace UI.WindowsAndMessaging
 	{
 		[UnmanagedFunctionPointer(CallingConvention.Winapi)]
 		public delegate LRESULT WNDPROC(HWND hWnd, uint msg, WPARAM wParam, LPARAM lParam);
+
+		/// <summary>Contains information about the size and position of a window.</summary>
+		/// <remarks>
+		/// <para><see href="https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-windowpos">Learn more about this API from docs.microsoft.com</see>.</para>
+		/// </remarks>
+		public partial struct WINDOWPOS
```
0x5bfa, line 24: Don't add any to this file. Use CsWin32's one.

src/Files.App/MainWindow.xaml.cs
```diff
@@ -339,5 +343,28 @@ x.tabItem.NavigationParameter.NavigationParameter is PaneNavigationArguments pan
...
+				if (CanWindowToFront != canWindowToFront)
+				{
+					CanWindowToFront = canWindowToFront;
+					return true;
+				}
+				return false;
+			}
+		}
+
+		private const int WM_WINDOWPOSCHANGING = 0x0046;
```
0x5bfa, line 360: Add this to CsWin32 and use it thru PInvoke.WM_WINDOWPOSCHANGING

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -1200,6 +1216,10 @@ private void RefreshItem(SelectorItem container, object item, bool inRecycleQueu
 
 		protected internal void FileListItem_PointerPressed(object sender, PointerRoutedEventArgs e)
 		{
+			// Set can window to front and bring the window to the front if necessary
```
0x5bfa, line 1219: Please add issue number like this.
There's also 5 places to add this below this line.
```suggestion
			// Set can window to front and bring the window to the front if necessary (#13255)
```

Issue #16477, opened 2024-11-17
Feature: RTL interface


sofianhit:
### What feature or improvement do you think would benefit Files?

hi
I wish this application was compatible with right-to-left languages. For example, I use Windows with a right-to-left language.
The vertical sidebar list is on the right side .
can you please add this feature ? 
i cant purchase the app without RTL direction .
thanks 
![Image](https://github.com/user-attachments/assets/6683abfc-7b69-41fa-b3bc-7f1808c38bc7)

### Requirements

1- move the sidebar menu ( that display files and folders on the left side ) to the right side .
2- move the main big section (that show hard desks and all folders + files) to the left .
3- let tabs start from the right (first tab open on the right not on the left)
4- the options , arrows and buttons start from right to left .

![Image](https://github.com/user-attachments/assets/16652b40-481b-469f-80b7-0e4db582f7f8)

### Files Version

3.7.11.0

### Windows Version

10.0.19045.5131

### Comments

people using RTL platform , all the time they have RIGHT-TO-LEFT look .
their language also start from right .
i wonder if "files app" become has this feature .

thanks

yaira2:
Fyi @XTorLukas

0x5bfa:
I thought we already do this? If not we need extra care to sidebar visual states, while other WinUI controls support RTL such as Breadcrumb, TabView, AppBar, ListView and GridView afaik

yaira2:
We already do, but the code stopped working after the upgrade to WASDK.

0x5bfa:
FYI https://github.com/microsoft/microsoft-ui-xaml/issues/8746

sofianhit:
Will your words make me feel hopeless about this matter?
I see many programs and browsers that I use daily go from right to left automatically immediately after installation.
Will this feature return to the program again or not?
Is the program developer interested in doing this task?

yaira2:
@XTorLukas has a workaround but I think he's still working on it.

sofianhit:
i hope finish it soon .

very thanks

XTorLukas:
I'm still working on it, but I have many more serious issues to deal with.
First of all, of course, I need to fix any problems with incompatible GUI components. The problem occurred after switching to WASDK.

sofianhit:
> I'm still working on it, but I have many more serious issues to deal with. First of all, of course, I need to fix any problems with incompatible GUI components. The problem occurred after switching to WASDK.

I wish you success and good luck.
[...]

Issue #2786, opened 2020-12-31
File properties page does not support RTL


Don-Vito:
**Describe the bug**
File details do not support RTL

**To Reproduce**
1. Open file properties by right-clicking the file and choosing properties

**Expected behavior**
The page is RTL

**Screenshots**
![image](https://user-images.githubusercontent.com/4639110/103390796-ee0f1500-4b1e-11eb-911b-f74cab62ddb2.png)

**Desktop (please complete the following information):**
 - App version: v0.23.9.0

yaichenbaum:
@Don-Vito I will work on fixing this issue. Also, do we finally have someone who can help with the Hebrew translation??

yaichenbaum:
@Don-Vito This issue should be resolved with #2788.

Don-Vito:
> 
> 
> @Don-Vito I will work on fixing this issue. Also, do we finally have someone who can help with the Hebrew translation??

Yep, I can help :)

Issue #14958, opened 2024-03-13
Bug: Drag gap between tab bar and windows controls doesn't always happen


Josh65-2201:
### Description

Currently sometimes show like this
![image](https://github.com/files-community/Files/assets/26977075/078e5237-0f9b-4b31-935c-ab48122e7fd3)

### Steps To Reproduce

1. Tested on resolution 1920x1080 and scaling 100%
2. Maximize window
3. Open new tabs until tab bar is full

### Requirements

- Make sure a drag gap is always present between the tabs and window controls

### Files Version

3.2.7.0

### Windows Version

10.0.22621.3155

### Log File

[debug.log](https://github.com/files-community/Files/files/14591229/debug.log)

0x5bfa:
I can repro.

![image](https://github.com/files-community/Files/assets/62196528/35565e72-b41e-4bce-88b4-cc0a7c78758f)

You can drag there and so you cannot add a new tab by the button, right?

yaira2:
A fix will be included in the next release. Thank you @XTorLukas for your research on this issue and for contributing the fix.

PR #16514, opened 2024-11-24
Feature: Added proper support for RTL languages

XTorLukas:
**Resolved / Related Issues**

- From #16477
- From #2786

### Microsoft issues
- From https://github.com/microsoft/microsoft-ui-xaml/issues/8746
- From https://github.com/microsoft/microsoft-ui-xaml/issues/9899

> [!IMPORTANT]  
> This service function is only used to display the layout correctly.

1. Opened Files
2. Set RTL or LTR culture
3. Automaticaly change flow direction layout

## Tests for RTL
- [x] All Dialogs show correctly
- [x] All Tooltip show correctly
- [x] All Windows show correctly
- [x] All Drag position works correctly
- [x] Forward & Back swipe in touchscreen works correctly
- [x] Position dragging works correctly on MainProperties page
- [ ] Tab action button is clickable


XTorLukas:
fyi @0x5bfa

src/Files.App/App.xaml.cs
```diff
@@ -97,6 +100,9 @@ async Task ActivateAsync()
 				var userSettingsService = Ioc.Default.GetRequiredService<IUserSettingsService>();
 				var isLeaveAppRunning = userSettingsService.GeneralSettingsService.LeaveAppRunning;
 
+				Ioc.Default.GetRequiredService<IRealTimeLayoutService>().UpdateCulture(new(AppLanguageHelper.PreferredLanguage.Code));
```
0x5bfa, line 103:
```suggestion
				var realTimeLayoutService = Ioc.Default.GetRequiredService<IRealTimeLayoutService>();
				realTimeLayoutService.UpdateCulture(new(AppLanguageHelper.PreferredLanguage.Code));
```

src/Files.App/App.xaml.cs
```diff
@@ -97,6 +100,9 @@ async Task ActivateAsync()
...
+				MainWindow.Instance.InitializeContentLayout();
```
0x5bfa, line 104: Shouldn't this be called from MainWindow's constructor after InitializeComponent call?

src/Files.App/Data/Contracts/IRealTimeWindow.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+	/// <summary>
+	/// Defines an interface for real-time window components that manage content layout updates.
+	/// </summary>
+	internal interface IRealTimeWindow
+	{
+		/// <summary>
+		/// Initializes the content layout for the window.
+		/// </summary>
+		void InitializeContentLayout() { }
```
0x5bfa, line 14:
```suggestion
		void InitializeContentLayout();
```

src/Files.App/Data/Contracts/IRealTimeWindow.cs
```diff
@@ -0,0 +1,21 @@
...
+
+		/// <summary>
+		/// Updates the content layout of the window.
+		/// </summary>
+		void UpdateContentLayout() { }
```
0x5bfa, line 19:
```suggestion
		void UpdateContentLayout();
```

src/Files.App/Dialogs/AddBranchDialog.xaml.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Files.App.Dialogs
 {
-	public sealed partial class AddBranchDialog : ContentDialog, IDialog<AddBranchDialogViewModel>
+	public sealed partial class AddBranchDialog : ContentDialog, IDialog<AddBranchDialogViewModel>, IRealTimeControl
```
0x5bfa, line 10: FYI, While this looks somewhat weird just for this but I'm planning to create an all-in-one service for ContentDialog & DynamicDialog to generate every dialog we use. The dialog contents will be Page class derivatives instead of the ContentDialog class.

For example:
```c#
IAppDialogService service = new();
var res = service.Show_AddBranchDialog();
```

src/Files.App/UserControls/AddressToolbar.xaml
```diff
@@ -263,63 +263,68 @@
 					</AnimatedIcon>
 				</ToggleButton>
 
-				<Button
-					x:Name="Back"
```
0x5bfa, line 267: Why is this changed?

src/Files.App/UserControls/Sidebar/SidebarView.xaml.cs
```diff
@@ -130,7 +130,10 @@ private void SidebarResizer_ManipulationStarted(object sender, ManipulationStart
 
 		private void SidebarResizer_ManipulationDelta(object sender, ManipulationDeltaRoutedEventArgs e)
 		{
-			var newWidth = preManipulationSidebarWidth + e.Cumulative.Translation.X;
+			var rtls = Ioc.Default.GetRequiredService<IRealTimeLayoutService>();
+			var rePos = rtls.FlowDirection == FlowDirection.LeftToRight ? 1 : -1;
```
0x5bfa, line 134: Why not make these service and rePos private static property?

src/Files.App/Utils/Storage/Helpers/FilePropertiesHelpers.cs
```diff
@@ -136,11 +133,12 @@ public static void OpenPropertiesWindow(object item, IShellPage associatedInstan
 
 			// WINUI3: Move window to cursor position
 			PInvoke.GetCursorPos(out var pointerPosition);
+			var rePos = RealTimeLayoutService.FlowDirection is FlowDirection.LeftToRight ? 1 : -1;
```
0x5bfa, line 136: As above.

src/Files.App/ViewModels/Settings/GeneralViewModel.cs
```diff
@@ -77,6 +77,10 @@ public int SelectedAppLanguageIndex
 				{
 					selectedAppLanguageIndex = value;
 					OnPropertyChanged(nameof(SelectedAppLanguageIndex));
+
+					var RTLayoutService = Ioc.Default.GetRequiredService<IRealTimeLayoutService>();
```
0x5bfa, line 81: Why not make this a private readonly field

0x5bfa (changes requested):
Looks great. Does this also work for generated content dialogs such as the add new dialog that is shown when Ctrl+Shift+I entered?

src/Files.App/App.xaml.cs
XTorLukas: For calling, the culture must be already set by realTimeLayoutService.UpdateCulture

src/Files.App/UserControls/AddressToolbar.xaml
XTorLukas: To correct arrow direction in RTL
image

0x5bfa: @yaira2 fyi

src/Files.App/App.xaml.cs
0x5bfa: I mean InitializeContentLayout shouldn't be called in MainWindow's constructor as others?

src/Files.App/UserControls/AddressToolbar.xaml
yaira2: What happens without this change?

src/Files.App/App.xaml.cs
XTorLukas: Yes I understand, the generator works with the IoC as part of the initialization to implement the service. But the MainWindow constructor is called before the IoC is implemented.

src/Files.App/UserControls/AddressToolbar.xaml
XTorLukas: The arrows are facing each other

yaira2: Would it work to wrap these in a StackPanel (as you've done) and hardcode the FlowDirection of the StackPanel?

XTorLukas: Yes this is exactly the solution: `FlowDirection="{Binding Path=FlowDirection, ElementName=Root}"` that I use

src/Files.App/Views/SplashScreenPage.xaml
```diff
@@ -42,6 +42,7 @@
 			x:Name="SplashScreenLoadingProgressRing"
 			Margin="0,0,0,48"
 			VerticalAlignment="Bottom"
+			FlowDirection="LeftToRight"
```
yaira2, line 45: What happens if this isn't hardcoded, does the progress ring spin in the wrong direction?

XTorLukas: inverse spin

src/Files.App/UserControls/TabBar/TabBar.xaml
```diff
@@ -237,6 +237,7 @@
 						Grid.Column="0"
 						Width="30"
 						Height="30"
+						Margin="0,0,6,0"
```
yaira2, line 240: Was this changed on purpose?

XTorLukas: Add a small space to avoid overlapping elements
image

yaira2: There should really be a larger gap there (I think that's what the left inset is for).

XTorLukas: That seemed like a quick fix, if that's the problem I'd add it in a different way.

yaira2: The area next to the caption buttons is supposed to be reserved as a drag region.

yaira2: @XTorLukas I think you're experiencing this issue #14958. Can you confirm?

XTorLukas: Yes, it will solve the problem, but I don't think it's part of the problem of this PR, so I decided to solve it as part of issue #16532.

yaira2: Can you please remove the gap that was added in this PR as it's unrelated to these changes?

XTorLukas: Ok

yaira2 (approved):

src/Files.App/Data/Enums/ArrowSymbolType.cs
yaira2, line 1: What's this enum used for?

src/Files.App/UserControls/AddressToolbar.xaml
```diff
@@ -273,7 +274,7 @@
 					IsEnabled="{x:Bind ViewModel.Commands.NavigateBack.IsExecutable, Mode=OneWay}"
 					Style="{StaticResource AddressToolbarButtonStyle}"
 					ToolTipService.ToolTip="{x:Bind ViewModel.Commands.NavigateBack.LabelWithHotKey, Mode=OneWay}">
-					<FontIcon FontSize="14" Glyph="{x:Bind ViewModel.Commands.NavigateBack.Glyph.BaseGlyph, Mode=OneWay}" />
+					<s:ArrowGlyph Arrow="Back" FontSize="14" />
```
yaira2, line 277: What's happening here?

src/Files.App/Views/Properties/MainPropertiesPage.xaml
```diff
@@ -69,14 +70,7 @@
...
 				</Button.Resources>
-				<AnimatedIcon x:Name="BackAnimatedIcon">
-					<AnimatedIcon.Source>
-						<animatedvisuals:AnimatedBackVisualSource />
-					</AnimatedIcon.Source>
-					<AnimatedIcon.FallbackIconSource>
-						<SymbolIconSource Symbol="Back" />
-					</AnimatedIcon.FallbackIconSource>
-				</AnimatedIcon>
+				<s:ArrowGlyph Arrow="Back" UseAnimatedIcon="True" />
```
yaira2, line 73: Same as above

src/Files.App/Data/Enums/ArrowSymbolType.cs
XTorLukas: Used to set `Arrow` in `<s:ArrowGlyph Arrow="Back" />`

src/Files.App/UserControls/AddressToolbar.xaml
XTorLukas: Rewinds the direction of the arrows for the RTL culture, this is necessary for correction functions. Simply setting the flow direction in xaml is not enough.

marcelwgn:
Thank you for all the effort you put into this PR @XTorLukas. I think it is a good step forward and good you made these investigations, I think there are a few things I would approach differently for this.

The approach with the code generator is quite interesting, however it seems quite complicated to what we would like to achieve here. From what I can see, there are three things we need to do to properly support RTL:
1. Update the caption buttons to be on the left in RTL
2. Set the FlowDirection of the MainPage to RTL
3. Get ContentDialogs to respect our RTL layout

While your approach does achieve this, it also creates a lot of complexity through the interfaces and the code generators associated with them. I would probably prefer a lightweight FlowDirectionService that essentially provides the following API:

1. A property called FlowDirection that you can bind to (and a function generating bindings for it in code behind)
2. A method to update the state from LTR to RTL (and vice versa) which sets the correct flags to update the titlebar/caption buttons and sets the FlowDirection property (which updates the bindings)

An open question for me (and something that is also not addressed in this PR right now I think) is how to deal with multiple windows. Does every window have its own FDS? How do all of these get notified of the change?

XTorLukas:
I'll close this PR for now, more implementation study is needed, I'm busy with other work at the moment.

yaira2:
@XTorLukas thank you for all your work on this. Hopefully we can bring this back in the future.

Issue #14952, opened 2024-03-13
Add spacing for tabs to allow moving window


emkaix:
### What feature or improvement do you think would benefit Files?

Normally, when you want to move the Files window, you click somewhere on the title bar to drag it.
However, I noticed that it is difficult to move the window when the tab list is full, as the area left for the mouse to click is very small. See screenshot below.

![full_tab_list](https://github.com/files-community/Files/assets/9194057/312be813-0751-43a9-a7ff-796cd61fe21c)

The area marked in red is the only area that is left to click with a full tab list.

Other applications that use tabs, like Google Chrome or the default Windows Explorer, leave some space between the "+" and the "Minimize" buttons, making it easy to move the window:

![full_tab_list_explorer](https://github.com/files-community/Files/assets/9194057/e79779a8-3eee-4e43-a232-88a40a8b0950)

### Requirements

- Adding spacing to the right-hand side of the "+" button would make it easier to move the window when the tab list is full.

### Files Version

3.2.2.0

### Windows Version

10.0.22631.3155

0x5bfa:
Thanks for the feedback!

I assume we’d had this ability but somehow removed. I’d add back if @yaira2 thinks it’s good.

emkaix:
A restart didn't fix it, it shows no gap as seen in the first screenshot. I also downloaded and installed the latest version inside a clean Windows 10 VM with the same effect (3.2.4.0, 10.0.19045.4046). Easily reproducible for me (as well as in the VM). Open Files and start adding tabs using the "+" button until the tabs bar is full.

yaira2:
I'm not able to repro this on my end
![image](https://github.com/files-community/Files/assets/39923744/f137d434-bf6b-4bab-86ab-9269f4920a29)

Josh65-2201:
I can on 3.2.7,0, It might be based on screen resolution and scaling. I'm on 1920x1080 scaling 100%

yaira2:
@Josh65-2201 can you open a bug report and include steps to repro?

Josh65-2201:
Sure

Josh65-2201:
https://github.com/files-community/Files/issues/14958

Issue #14598, opened 2024-01-30
Add sftp integration, or better yet rclone.


paolosezart:
### What feature or improvement do you think would benefit Files?

Rclone is a very functional program for mounting cloud disks. Unfortunately the configuration of this program is quite complicated (for a home user). Is it not possible to integrate rclone into your wonderful program?

### Requirements

- This proposal will accomplish X
- This proposal will accomplish Y
- This proposal will accomplish Z

### Files Version

3.1.2

### Windows Version

10.0.19045.2965

Josh65-2201:
Thanks for the feedback, We're tracking this in #2407 . You can check there or click subscribe in the right sidebar for updates.

PR #16549, opened 2024-11-29
Fix: Fixed an issue where the title bar drag region was sometimes missing

### LINKED ISSUE #14958: Bug: Drag gap between tab bar and windows controls doesn't always happen

### LINKED ISSUE #16532: Bug:
### Description

I found a problem that appears in all versions of Files.
It is a problem with not keeping the distance after the application starts if more tabs are open than the width of the window.
Image

### Steps To Reproduce

Ensuring the correct padding distance

### Files Version

3.8.0.0

### Windows Version

10.0.26100.2314

### User ID

e7721cff-79b9-4d35-941d-e59b2b4fcc91

### Log File

There is no need

XTorLukas:
**Resolved / Related Issues**

- Closes #16532
- Closes #14958

1. Open enough tabs that the scroll buttons are displayed
2. Fully close Files (enable continue where you left off)
3. Reopen Files
4. Tabs don't overflow and the gap is maintained.

image

- [x] Gap is identical at different dpi's e.g. 1.25

### DIFF
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -9,9 +9,6 @@ using Microsoft.UI.Xaml.Media.Animation;
 using System.Runtime.InteropServices;
 using Windows.ApplicationModel.Activation;
 using Windows.Storage;
-using Windows.Win32;
-using Windows.Win32.Graphics.Gdi;
-using Windows.Win32.UI.HiDpi;
 using IO = System.IO;
 
 namespace Files.App
@@ -341,16 +338,6 @@ namespace Files.App
 						break;
 				}
 			}
-		}
-
-		public double GetDpiScale()
-		{
-			var hMonitor = PInvoke.MonitorFromWindow(
-				new(WinRT.Interop.WindowNative.GetWindowHandle(this)),
-				MONITOR_FROM_FLAGS.MONITOR_DEFAULTTONEAREST);
-
-			var res = PInvoke.GetDpiForMonitor(hMonitor, MONITOR_DPI_TYPE.MDT_EFFECTIVE_DPI, out var dpiX, out var _);
-			return res == 0 ? ((float)dpiX / Win32PInvoke.USER_DEFAULT_SCREEN_DPI) : 1;
 		}
 	}
 }
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -9,6 +9,9 @@ using Microsoft.UI.Xaml.Media.Animation;
 using System.Runtime.InteropServices;
 using Windows.ApplicationModel.Activation;
 using Windows.Storage;
+using Windows.Win32;
+using Windows.Win32.Graphics.Gdi;
+using Windows.Win32.UI.HiDpi;
 using IO = System.IO;
 
 namespace Files.App
@@ -338,6 +341,16 @@ namespace Files.App
 						break;
 				}
 			}
+		}
+
+		public double GetDpiScale()
+		{
+			var hMonitor = PInvoke.MonitorFromWindow(
+				new(WinRT.Interop.WindowNative.GetWindowHandle(this)),
+				MONITOR_FROM_FLAGS.MONITOR_DEFAULTTONEAREST);
+
+			var res = PInvoke.GetDpiForMonitor(hMonitor, MONITOR_DPI_TYPE.MDT_EFFECTIVE_DPI, out var dpiX, out var _);
+			return res == 0 ? ((float)dpiX / Win32PInvoke.USER_DEFAULT_SCREEN_DPI) : 1;
 		}
 	}
 }
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -60,7 +60,7 @@ namespace Files.App.UserControls.TabBar
 		public GridLength TitleBarWidth
 		{
 			get => _titleBarWidth;
-			set => _titleBarWidth = new((value.Value / MainWindow.Instance.GetDpiScale()) + _gap);
+			set => _titleBarWidth = new((value.Value / App.AppModel.AppWindowDPI) + _gap);
 		}
 
 		// Events
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -407,7 +407,15 @@ namespace Files.App.UserControls.TabBar
 		}
 
 		private void DragAreaRectangle_Loaded(object sender, RoutedEventArgs e)
-			=> HorizontalTabView.Measure(new(HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - TitleBarWidth.Value, HorizontalTabView.ActualHeight));
+		{
+			if (HorizontalTabView.ActualWidth > 0 && TabBarAddNewTabButton.Width > 0 && TitleBarWidth.Value > 0)
+			{
+				HorizontalTabView.Measure(new(HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - TitleBarWidth.Value, HorizontalTabView.ActualHeight));
+				return;
+			}
+
+			DispatcherQueue.TryEnqueue(() => DragAreaRectangle_Loaded(sender, e));
+		}
 
 		public int SetTitleBarDragRegion(InputNonClientPointerSource source, SizeInt32 size, double scaleFactor, Func<UIElement, RectInt32?, RectInt32> getScaledRect)
 		{
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -60,7 +60,7 @@ namespace Files.App.UserControls.TabBar
 		public GridLength TitleBarWidth
 		{
 			get => _titleBarWidth;
-			set => _titleBarWidth = new(value.Value + _gap);
+			set => _titleBarWidth = new((value.Value / MainWindow.Instance.GetDpiScale()) + _gap);
 		}
 
 		// Events
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -362,25 +362,14 @@ namespace Files.App.UserControls.TabBar
 
 		private void DragAreaRectangle_Loaded(object sender, RoutedEventArgs e)
 		{
-			if (HorizontalTabView.ActualWidth > 0 && TabBarAddNewTabButton.Width > 0)
-			{
-				var appWindow = MainWindow.Instance.AppWindow;
-				var titleBarInset = (FilePropertiesHelpers.FlowDirectionSettingIsRightToLeft
-					? appWindow.TitleBar.LeftInset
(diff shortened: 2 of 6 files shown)


src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -125,8 +125,10 @@ private void UserSettingsService_OnSettingChangedEvent(object? sender, SettingCh
 		}
 
 		private void HorizontalMultitaskingControl_Loaded(object sender, RoutedEventArgs e)
+			=> TabControl.TabControlAreaSizeChanged += TabControlArea_SizeChanged;
+
+		private void TabControlArea_SizeChanged(object sender, SizeChangedEventArgs e)
```
0x5bfa, line 133: I don't think you should hook size event in here. This method hooks CurrentInstanceChanged every time the size changed. I think loaded event was good.

XTorLukas: Yea,  but I've already committed to a more elegant solution.

src/Files.App/MainWindow.xaml.cs
```diff
@@ -339,5 +342,15 @@ x.tabItem.NavigationParameter.NavigationParameter is PaneNavigationArguments pan
 				}
 			}
 		}
+
+		public double GetDpiScale()
```
yaira2, line 346: We already have `AppWindowDPI` in AppModel, is there any way to reuse it?

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -28,6 +30,8 @@ public sealed partial class TabBar : BaseTabBar, INotifyPropertyChanged
 
 		private bool _lockDropOperation = false;
 
+		private static int _gap = 6;
```
yaira2, line 33: We don't need to add an additional gap.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -51,6 +55,14 @@ public bool AllowTabsDrag
 		public Rectangle DragArea
 			=> DragAreaRectangle;
 
+		private GridLength _titleBarWidth;
+
+		public GridLength TitleBarWidth
+		{
+			get => _titleBarWidth;
+			set => _titleBarWidth = new((value.Value / MainWindow.Instance.GetDpiScale()) + _gap);
```
yaira2, line 63: Please see if you can use `App.AppModel.AppWindowDPI`.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
yaira2, line 33: Is this still needed?

src/Files.App/MainWindow.xaml.cs
XTorLukas: Thanks, I didn't know.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
XTorLukas: Thanks, I didn't know.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
XTorLukas: Yes, I would still leave a gap, most programmes use it. On a touchscreen it's very useful to have a space to catch the window. But I would probably use a gap (minimum size) of 40px.

yaira2: I meant that there should already be the correct size gap from the inset properties.

XTorLukas: Without a gap, the space is exactly the size of TitleBar.Inset(Right or Left), depending on the culture.

> image

Green: TabBar.Inset
Red: minimum size `_gap`

yaira2: I'm seeing a different behavior on my side 
image

XTorLukas: it's the same, the size of TitleBar + _gap 🤔

yaira2: As you can see, the gap is already there, however if you follow the steps listed in #14598, the gap is not set correctly.

XTorLukas: > As you can see, the gap is usually there by default, however if you follow the steps listed in #14598, the gap is not set correctly.

For me so far the gap is always set correctly, I haven't found a problem yet

XTorLukas: If you mean this problem #14952 , I have already solved it in this PR

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -84,6 +89,40 @@ public TabBar()
 			};
 		}
 
+		/// <summary>
+		/// Initializes the title bar asynchronously. This method attempts to get the title bar's left and right insets.
+		/// If the values are properly initialized, it sets the <see cref="TitleBarWidth"/> property. If the initialization fails,
+		/// it falls back to a default width.
+		/// </summary>
+		private async void InitializeTitleBarAsync()
```
yaira2, line 97: Do we still need multiple attempts now that you added `UIElement.Measure`?

XTorLukas: UIElement.Measure I added this before this test. But then you described a problem that never appeared during my testing. If this problem was only caused by a missing UIElement.Measure, then this test would no longer be necessary.

yaira2: Let's see what happens if we remove it, hopefully `UIElement.Measure` is all we need.

XTorLukas: In my personal testing, this has not manifested itself, but I plan to test this further on a weaker laptop where slower loading and thus the problem you describe may manifest itself.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -51,6 +55,14 @@ public bool AllowTabsDrag
 		public Rectangle DragArea
 			=> DragAreaRectangle;
 
+		private GridLength _titleBarWidth;
+
+		public GridLength TitleBarWidth
```
yaira2, line 60:
```suggestion
		public GridLength TitleBarRightInsetWidth
```

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -366,5 +374,25 @@ private void TabViewItem_Loaded(object sender, RoutedEventArgs e)
...
+			if (HorizontalTabView.ActualWidth > 0 && TabBarAddNewTabButton.Width > 0 && TitleBarWidth.Value > 0)
+			{
+				HorizontalTabView.Measure(new(HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - TitleBarWidth.Value, HorizontalTabView.ActualHeight));
+				return;
+			}
+
+			DispatcherQueue.TryEnqueue(() => DragAreaRectangle_Loaded(sender, e));
+		}
+
+		public int SetTitleBarDragRegion(InputNonClientPointerSource source, SizeInt32 size, double scaleFactor, Func<UIElement, RectInt32?, RectInt32> getScaledRect)
```
yaira2, line 389: Is it possible to move this method back to `MainPage`?

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -51,6 +55,14 @@ public bool AllowTabsDrag
 		public Rectangle DragArea
 			=> DragAreaRectangle;
 
+		private GridLength _titleBarWidth;
+
+		public GridLength TitleBarWidth
+		{
+			get => _titleBarWidth;
+			set => _titleBarWidth = new((value.Value / App.AppModel.AppWindowDPI) + _gap);
```
yaira2, line 63: Assuming the `TitleBar.Inset` property isn't null, I think it's already DPI aware and we can use the value as is.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
XTorLukas: It is possible to change, but a gap is already added to the value.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
XTorLukas: @0x5bfa  suggested to convert all TabBar operations into one file, and for me it is reasonable

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
XTorLukas: No, I tested with 1.25 dpi and the value is not recalculated.

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
yaira2: I don't think we want to restrict the logic to the TabBar, this way all our options are open if we ever add a scenario where the TabBar is hidden.

XTorLukas: I don't know, my preference is to keep the changes

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -366,5 +359,39 @@ private void TabViewItem_Loaded(object sender, RoutedEventArgs e)
...
+					titleBarInset += 40; // Add 40px gap
+					RightPaddingColumn.Width = new(titleBarInset);
+
+					HorizontalTabView.Measure(new(
+						HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - titleBarInset,
+						HorizontalTabView.ActualHeight));
+
+					return;
+				}
+			}
```
yaira2, line 383:
```suggestion
		private void DragAreaRectangle_Loaded(object sender, RoutedEventArgs e)
		{
			double scaleAdjustment = DragAreaRectangle.XamlRoot.RasterizationScale;
    		double titleBarInset = ((FilePropertiesHelpers.FlowDirectionSettingIsRightToLeft
        		? MainWindow.Instance.AppWindow.TitleBar.LeftInset
        		: MainWindow.Instance.AppWindow.TitleBar.RightInset) / scaleAdjustment) + 48;

   			HorizontalTabView.Measure(new(HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - titleBarInset, HorizontalTabView.ActualHeight));
    		RightPaddingColumn.Width = new(titleBarInset >= 0 ? titleBarInset : 0);
		}
```

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
yaira2: I would prefer if this is reverted

yaira2 (approved):
LGTM

XTorLukas:
Solved problems in this revert PR #16595

PR #16559, opened 2024-12-04
GitHub: Added workflow to submit package to Store

0x5bfa:
### Resolved / Related Issues

Planned internally

_None_


.github/workflows/cd-store.yml
yaira2, line 1: Can you add this to the preview CD first?

PR #16566, opened 2024-12-06
Code Quality: Added Store_Preview configuration

yaira2:
**Resolved / Related Issues**

To prevent extra work, all changes to the Files codebase must link to an approved issue marked as `Ready to build`. Please insert the issue number following the hashtag with the issue number that this Pull Request resolves.
- Closes #

**Steps used to test these changes**

Stability is a top priority for Files and all changes are required to go through testing before being merged into the repo. Please include a list of steps that you used to test this PR.

1. Opened Files ...
2. ...

PR #16569, opened 2024-12-06
Code Quality: Removed unnecessary configurations

0x5bfa:
## Summary

We removed Store, Preview and Stable configurations in #16566. This is follow-up PR for it in order to remove leftovers.

We now have only **Debug** and **Release**.
While we have 4 release destinations in **Release** configuration: SideloadPreview, SideloadStable, StorePreview and StoreStable, these are configured within our CD pipeline for the simplicity in project configurations since these have differences only in packaging and publish destination.

### DIFF
--- a/src/Files.App (Package)/Files.Package.wapproj
+++ b/src/Files.App (Package)/Files.Package.wapproj
@@ -28,61 +28,25 @@
       <Configuration>Debug</Configuration>
       <Platform>x86</Platform>
     </ProjectConfiguration>
-    <ProjectConfiguration Include="Preview|ARM64">
-      <Configuration>Preview</Configuration>
-      <Platform>ARM64</Platform>
-    </ProjectConfiguration>
-    <ProjectConfiguration Include="Preview|x64">
-      <Configuration>Preview</Configuration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
       <Platform>x64</Platform>
     </ProjectConfiguration>
-    <ProjectConfiguration Include="Preview|x86">
-      <Configuration>Preview</Configuration>
-      <Platform>x86</Platform>
+    <ProjectConfiguration Include="Debug|arm64">
+      <Configuration>Debug</Configuration>
+      <Platform>ARM64</Platform>
     </ProjectConfiguration>
     <ProjectConfiguration Include="Release|x86">
       <Configuration>Release</Configuration>
       <Platform>x86</Platform>
     </ProjectConfiguration>
-    <ProjectConfiguration Include="Debug|x64">
-      <Configuration>Debug</Configuration>
-      <Platform>x64</Platform>
-    </ProjectConfiguration>
     <ProjectConfiguration Include="Release|x64">
       <Configuration>Release</Configuration>
       <Platform>x64</Platform>
     </ProjectConfiguration>
-    <ProjectConfiguration Include="Debug|ARM64">
-      <Configuration>Debug</Configuration>
-      <Platform>ARM64</Platform>
-    </ProjectConfiguration>
-    <ProjectConfiguration Include="Release|ARM64">
+    <ProjectConfiguration Include="Release|arm64">
       <Configuration>Release</Configuration>
       <Platform>ARM64</Platform>
-    </ProjectConfiguration>
-    <ProjectConfiguration Include="Stable|ARM64">
-      <Configuration>Stable</Configuration>
-      <Platform>ARM64</Platform>
-    </ProjectConfiguration>
-    <ProjectConfiguration Include="Stable|x64">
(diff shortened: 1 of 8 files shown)


0x5bfa:
Should I include follow up for CD for store submission to add a checkbox

src/Files.App.Server/Files.App.Server.csproj
```diff
@@ -1,12 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <OutputType>WinExe</OutputType>
-        <DefaultLanguage>en-US</DefaultLanguage>
-        <AppxBundleAutoResourcePackageQualifiers>Scale|DXFeatureLevel</AppxBundleAutoResourcePackageQualifiers>
-        <AppxDefaultResourceQualifiers>Language=en-US;af;ar;be-BY;bg;ca;cs-CZ;da;de-DE;el;en-GB;es-ES;es-419;fa-IR;fi-FI;fil-PH;fr-FR;he-IL;hi-IN;hr-HR;hu-HU;id-ID;it-IT;ja-JP;ka;km-KH;ko-KR;lt-LT;lv-LV;ms-MY;nb-NO;nl-NL;pl-PL;pt-BR;pt-PT;ro-RO;ru-RU;sk-SK;sq-AL;sr-Cyrl;sv-SE;ta;th-TH;tr-TR;uk-UA;vi;zh-Hans;zh-Hant</AppxDefaultResourceQualifiers>
```
yaira2, line 7: Shouldn't these be kept?

0x5bfa: I don't think so. These are for WAP for us for now. @hez2010 fyi.

src/Files.App/Files.App.csproj
```diff
@@ -6,19 +6,8 @@
...
-        <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
-        <AppxPackageSigningTimestampDigestAlgorithm>SHA256</AppxPackageSigningTimestampDigestAlgorithm>
-        <AppxAutoIncrementPackageRevision>False</AppxAutoIncrementPackageRevision>
-        <GenerateTestArtifacts>False</GenerateTestArtifacts>
-        <AppxBundle>Always</AppxBundle>
-        <DisableXbfLineInfo>False</DisableXbfLineInfo>
-        <AppxBundlePlatforms>x86|x64|arm64</AppxBundlePlatforms>
-        <HoursBetweenUpdateChecks>0</HoursBetweenUpdateChecks>
         <Nullable>Enable</Nullable>
-        <ApplicationManifest>app.manifest</ApplicationManifest>
```
hez2010, line 21: This manifest need to be kept otherwise it can result in blurring on high-DPI devices.

src/Files.App.Server/Files.App.Server.csproj
hez2010: I'm not sure about this, but I think it can be removed if localization is not broken by this change.

src/Files.App/Files.App.csproj
0x5bfa: L45 already has it, should I remove that one instead?

src/Files.App.Server/Files.App.Server.csproj
0x5bfa: This even doesn't UseWinUI

src/Files.App/Files.App.csproj
hez2010: Maybe, `ApplicationManifest` seems to be the one need to be kept, while I cannot find any document about `Manifest`.

0x5bfa: Thanks. Looking back my FluentHub, I kept ApplicationManifest only and it is still DPI-Aware

src/Files.App.Server/Files.App.Server.csproj
yaira2: I think it can be removed from the server project, but I believe it's needed in both the packaging and app projects.

0x5bfa:
PTAL

0x5bfa:
PTAL

src/Files.App (Package)/Files.Package.wapproj
```diff
@@ -10,7 +10,7 @@
     <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
     <AppxAutoIncrementPackageRevision>False</AppxAutoIncrementPackageRevision>
     <AppxSymbolPackageEnabled>True</AppxSymbolPackageEnabled>
-    <AppxBundlePlatforms>x86|x64|arm64</AppxBundlePlatforms>
```
yaira2, line 13: We should still have x86

PR #16572, opened 2024-12-06
Fix: Fixed issue where changing a folder's grouping preference wouldn't update other open tabs

### LINKED ISSUE #16570: Bug: Grouping and sorting options are not consistent between tabs/panes
### Description

Currently, groups and the sort order are not updated between tabs and panes.
If, for example, I change the `GroupOption` to `Type` in a folder, I'd expect to see that layout applied to the other tabs as well (if they are showing the same folder). The video below will show you what I mean.

I'm trying to work on #11759, but I think we should first address this one.

https://github.com/user-attachments/assets/400032e5-87bb-41e3-b82e-211ec04518cd

### Steps To Reproduce

1. Open two tabs with the same folder
2. Change `sort/group` `option/order` in a tab
3. Switch to the second tab
4. See that the sorting/grouping has not changed

### Files Version

3.8.0.0

### Windows Version

10.0.22631.4541

### User ID

8467f272-08cd-4a50-93e0-39458990bdc4

### Log File

Not relevant

ferrariofilippo:
**Resolved / Related Issues**

- Closes #16570 

**Notes**
I'm not sure how to handle multiple panes: currently the layout is updated only when the pane is focused, to avoid wasting resources (see video below). Is this fine? Would it be better to update the layout istantly?

1. Open two tabs with the same folder
2. Change `sort/group` `option/order` in a tab
3. Switch to the second tab
4. Repeat using the `Sync preferences between folders` setting

https://github.com/user-attachments/assets/11a4b97d-7516-4cb7-803b-b926ffe2ec2c


yaira2:
> Would it be better to update the layout istantly?

How much of a performance difference is there?

ferrariofilippo:
> How much of a performance difference is there?

As you may think, the issue involves mainly big folders: with small ones (~ 50 items) I saw no difference.
I tested a folder with ~550 items on my laptop (i7-13650HX, 16GB): sometimes it got slightly laggy when updating both panes.
I'll leave two recordings below (sped-up 2x).

The real issue is with RAM usage: updating two panes uses twice the RAM. 
In my case, updating the 550 items folder I reached a difference of ~350 MB after 7 updates (I guess we have a memory leak).
Each update required more or less 50MB per pane.
I imagine this would get worse using some layouts such as `Grid`

https://github.com/user-attachments/assets/33fefe2f-c1d5-4222-9e35-1573b0da82cf

https://github.com/user-attachments/assets/fa936939-a12e-4f73-8c93-258d83301b08

Should I push the code to update both the panes so that you can test by yourself?

yaira2:
I think it's important to update both panes, but we should make sure that the focus remains on the active pane. While memory usage is a valid concern, I don't think it's common for users to have the same folder open in both panes and then decide to change the grouping.

ferrariofilippo:
> I think it's important to update both panes, but we should make sure that the focus remains on the active pane. While memory usage is a valid concern, I don't think it's common for users to have the same folder open in both panes and then decide to change the grouping.

Focus is fine, it was me changing it

src/Files.App/Views/ShellPanesPage.xaml.cs
```diff
@@ -185,7 +185,9 @@ public IShellPage? ActivePane
 						secondShellPage.IsCurrentInstance = false;
 
 					if (ActivePane is not null)
+					{
 						ActivePane.IsCurrentInstance = IsCurrentInstance;
+					}
```
yaira2, line 190:
```suggestion
						ActivePane.IsCurrentInstance = IsCurrentInstance;
```

0x5bfa:
Code wise, looks great otherwise.

src/Files.App/Data/Contexts/DisplayPage/DisplayPageContext.cs
```diff
@@ -131,6 +131,20 @@ public DisplayPageContext()
 			layoutSettingsService.PropertyChanged += Settings_PropertyChanged;
 		}
 
+		public void UpdateAllTabsAndPanesLayout()
```
yaira2, line 134:
```suggestion
		public void UpdateOpenTabsPreferences()
```

As a side note, should this method be under the page context? It seems it might be better as a helper.

ferrariofilippo:
As a side note, I found out there's a similar issue with layouts not being updated between panes... I remember it used to work, didn't it?

src/Files.App/Data/Contexts/ContentPage/ContentPageContext.cs
```diff
@@ -62,6 +62,20 @@ public ContentPageContext()
 			Update();
 		}
 
+		public void UpdateOpenTabsPreferences()
```
yaira2, line 65: Should this method really be under the page context? It seems it might be better as a helper.

ferrariofilippo: Is one between `LayoutPreferencesManager` (as a static method), `UIHelpers` fine or should I create a new `Helper` class?

yaira2: Let's go with a new class

src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
```diff
@@ -265,6 +265,24 @@ public Type GetLayoutType(string path, bool changeLayoutMode = true)
 			};
 		}
 
+		public void UpdateGroupAndSortOptions(string? path)
+		{
+			if (string.IsNullOrWhiteSpace(path))
+				return;
+
+			var preferencesItem = GetLayoutPreferencesForPath(path);
```
yaira2, line 273: Do we need to reload the preferences, or can we simplify this further by just invoking the related events (eg `GroupDirectionPreferenceUpdated`)?

ferrariofilippo: We cannot just invoke the event because the properties are not updated unless we reload them. I can add a parameter that tells which is the only one to update

ferrariofilippo: BTW should we sync sorting preferences even when they are set clicking on the details view headers? If so I'll implement it

yaira2: > We cannot just invoke the event because the properties are not updated unless we reload them.

I see, in that case I would just leave it as is.

> BTW should we sync sorting preferences even when they are set clicking on the details view headers? If so I'll implement it

It might be best to wrap up this PR first, but we can look into this next.

src/Files.App/Data/Contracts/IShellPanesPage.cs
```diff
@@ -58,5 +58,12 @@ public interface IShellPanesPage : IDisposable, INotifyPropertyChanged
 		/// Focuses the other pane.
 		/// </summary>
 		public void FocusOtherPane();
+
+		/// <summary>
+		/// Updates the layout of open panes.
+		/// </summary>
+		/// <param name="targetPath">The path of panes to update</param>
+		/// <param name="includeActivePane">Whether the active pane should be updated or not</param>
+		public void UpdatePanesLayout(string targetPath, bool includeActivePane = false);
```
yaira2, line 67:
```suggestion
		public void UpdateOpenPanesPreferences(string targetPath, bool includeActivePane = false);
```

src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
```diff
@@ -265,6 +265,24 @@ public Type GetLayoutType(string path, bool changeLayoutMode = true)
 			};
 		}
 
+		public void UpdateGroupAndSortOptions(string? path)
```
yaira2, line 268:
```suggestion
		public void ReloadGroupAndSortPreferences(string? path)
```

src/Files.App/Views/ShellPanesPage.xaml.cs
```diff
@@ -356,6 +357,23 @@ public void FocusOtherPane()
 				GetPane(0)?.Focus(FocusState.Programmatic);
 		}
 
+		/// <inheritdoc/>
+		public void UpdateOpenPanesPreferences(string targetPath, bool includeActivePane = false)
+		{
+			foreach (var pane in GetPanes())
```
yaira2, line 363: I'm not certain this is doable, but it might be worth seeing if this code could be put in the layout helper as well.

ferrariofilippo: I'll see what I can do

ferrariofilippo: I just had a look at `ModernShellPage` implementation and it seems there are no "dangerous" public things, so I guess we can make it public

src/Files.App/Helpers/Layout/LayoutHelpers.cs
```diff
@@ -0,0 +1,36 @@
...
+					if ((isNotCurrentTab || pane != shPage.ActivePane) &&
+						(layoutSettingsService.SyncFolderPreferencesAcrossDirectories ||
+						path is not null &&
+						path.Equals(activePath, StringComparison.OrdinalIgnoreCase)))
+					{
+						var page = pane.SlimContentPage as BaseLayoutPage;
+						page?.FolderSettings?.ReloadGroupAndSortPreferences(path);
+					}
+				}
+			}
```
yaira2, line 33:
```suggestion
public static void UpdateOpenTabsPreferences()
{
	// Services
	var multitaskingContext = Ioc.Default.GetRequiredService<IMultitaskingContext>();
	var layoutSettingsService = Ioc.Default.GetRequiredService<ILayoutSettingsService>();

	// Get all tab instances and active path
	var tabs = multitaskingContext.Control?.GetAllTabInstances();
	var activePath = (multitaskingContext.CurrentTabItem?.TabItemContent as ShellPanesPage)?.ActivePane?.TabBarItemParameter?.NavigationParameter as string;

	// Return if required data is missing
	if (tabs is null || activePath is null)
		return;

	for (int i = 0; i < tabs.Count; i++)
	{
		var isNotCurrentTab = i != multitaskingContext.CurrentTabIndex;
		var shPage = tabs[i] as ShellPanesPage;

		if (shPage is not null)
		{
			foreach (var pane in shPage.GetPanes())
			{
				var path = pane.ShellViewModel?.CurrentFolder?.ItemPath;

				// Skip panes without a valid path
				if (path is null)
					continue;

				// Check if we need to update preferences for this pane
				if ((isNotCurrentTab || pane != shPage.ActivePane) &&
					(layoutSettingsService.SyncFolderPreferencesAcrossDirectories ||
					 path.Equals(activePath, StringComparison.OrdinalIgnoreCase)))
					if (pane.SlimContentPage is BaseLayoutPage page)
						page.FolderSettings?.ReloadGroupAndSortPreferences(path);
			}
		}
	}
}
```

yaira2 (approved):
LGTM

PR #16580, opened 2024-12-08
Fix : Bug: Soft link file details layout information misaligned #15998

### LINKED ISSUE #15998: Soft link file misaligned
### Description

image

### Steps To Reproduce

1. create a softlink file (via mklink, not a Windows shortcut file) under a git project
2. switch to `Details` layout

### Requirements

soft link file has the same alignment with other files

### Files Version

3.6.0.0

### Windows Version

10.0.22631.3958

### Log File

no

ashrafmansuri:
**Resolved / Related Issues**
- Closes #15998

1. Opened Files 
2. Create a soft link in a git repo.
3. Change to detail layout.
4. check the information alignment.

### DIFF
--- a/src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
+++ b/src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
@@ -277,26 +277,49 @@ namespace Files.App.Utils.Storage
 			if (isSymlink)
 			{
 				var targetPath = NativeFileOperationsHelper.ParseSymLink(itemPath);
-
-				return new ShortcutItem(null)
+				if (isGitRepo)
 				{
-					PrimaryItemAttribute = StorageItemTypes.File,
-					FileExtension = itemFileExtension,
-					IsHiddenItem = isHidden,
-					Opacity = opacity,
-					FileImage = null,
-					LoadFileIcon = itemThumbnailImgVis,
-					ItemNameRaw = itemName,
-					ItemDateModifiedReal = itemModifiedDate,
-					ItemDateAccessedReal = itemLastAccessDate,
-					ItemDateCreatedReal = itemCreatedDate,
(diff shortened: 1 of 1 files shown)


ashrafmansuri:
From my observation if (isSymlink) is called before else if (isGitRepo), isGitRepo is true so I have just added a check in the isSymlink block.

ashrafmansuri:
The icon is different which can be changed. By adding properties to GitItem(). Can I get this issue assigned to me? And if any other review please let me know.

hishitetsu (changes requested):
This still leaves the alignment wrong for shortcut files (.lnk and .url).
I don't believe this approach is correct. The problem is that ShortcutItem doesn't support Git repositories, so we should create something like GitShortcutItem.

0x5bfa:
@ashrafmansuri Look at my earlier comment https://github.com/files-community/Files/issues/15998#issuecomment-2308829838 Since checking if item is shortcut or not is done before if it is git item or not, details view fails to x:Load text blocks for git info.

ashrafmansuri:
> This still leaves the alignment wrong for shortcut files (.lnk and .url).
> 
> I don't believe this approach is correct. The problem is that ShortcutItem doesn't support Git repositories, so we should create something like GitShortcutItem.

Alright. Understood.

ashrafmansuri:
> @ashrafmansuri Look at my earlier comment https://github.com/files-community/Files/issues/15998#issuecomment-2308829838

Yes I saw that. Your comment helped me a lot. So should I check for isGitRepo before or should I create a new item as suggested by @hishitetsu ?

0x5bfa:
Yeah. Create a new class that derives from both ShortcutItem and GitItem named GitShortcutItem while this isn't ideal way.

ashrafmansuri:
I'll create a new pull request

Issue #16583, opened 2024-12-08
Code Quality: Files bundles all 7z.dll for all three architectures


Pinguin2001:
### Description

Files bundles by default all three 7z dlls regardless of the architecture.

### Concerned code

src\Files.App\Files.App.csproj

### Gains

To be fair, only 3 mbs reduction in bundle size, but it doesn't hurt

### Requirements

Add three platform conditions

yaira2:
For future reference https://github.com/files-community/Files/pull/16584#issuecomment-2591338955

PR #15690, opened 2024-06-25
Fix: Fixed an issue with extracting archives on ARM64 devices

yaira2:
**Resolved / Related Issues**

To prevent extra work, all changes to the Files codebase must link to an approved issue marked as `Ready to build`. Please insert the issue number following the hashtag with the issue number that this Pull Request resolves.
- Closes #15689

**Steps used to test these changes**

Stability is a top priority for Files and all changes are required to go through testing before being merged into the repo. Please include a list of steps that you used to test this PR.

1. Opened Files ...
2. ...

PR #16584, opened 2024-12-08
Code Quality: Only bundle architecture specific 7z dll

Pinguin2001:
**Resolved / Related Issues**

- Closes #https://github.com/files-community/Files/issues/16583

1. Build files for all three architectures
2. Verify only the architecture specific dll has been copied

While looking into the outdated 7z dlls, I noticed files bundles all three dlls regardless of the architecture being built. This pr addresses this issue

### DIFF
--- a/src/Files.App/Files.App.csproj
+++ b/src/Files.App/Files.App.csproj
@@ -55,12 +55,13 @@
         <Content Update="Assets\FilesOpenDialog\Files.App.Launcher.exe">
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </Content>
-        <Content Include="7z64.dll">
-            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-        </Content><Content Include="7z.dll">
+        <Content Condition="'$(Platform)' == 'x64'" Include="7z64.dll">
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </Content>
-        <Content Include="7zArm64.dll">
+		<Content Condition="'$(Platform)' == 'x86'" Include="7z.dll">
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        </Content>
+        <Content Condition="'$(Platform)' == 'arm64'" Include="7zArm64.dll">
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </Content>
     </ItemGroup>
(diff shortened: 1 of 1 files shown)


src/Files.App/Files.App.csproj
```diff
@@ -55,12 +55,13 @@
         <Content Update="Assets\FilesOpenDialog\Files.App.Launcher.exe">
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </Content>
-        <Content Include="7z64.dll">
+        <Content Condition="'$(Platform)' == 'x64'" Include="7z64.dll">
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-        </Content><Content Include="7z.dll">
+        </Content>
+		<Content Condition="'$(Platform)' == 'x86'" Include="7z.dll">
```
yaira2, line 61: Can you fix the spacing?

Pinguin2001: Sure, VisualStudio somehow didn't display that and it only showed up in git, fixed!

Lamparter (approved):

yaira2:
Btw, I tried this once before, but it broke the installation.
https://github.com/files-community/Files/pull/15690/commits/7e288451026515d9bcfb4c82317185446c1d79cc

Pinguin2001:
> Btw, I tried this once before, but it broke the installation. [7e28845](https://github.com/files-community/Files/commit/7e288451026515d9bcfb4c82317185446c1d79cc)

Which issue occurred? I don't see how this could affect msix-based app installation

yaira2:
Users couldn't update or install the new version due to resource mapping issue.

Pinguin2001:
> Users couldn't update or install the new version due to resource mapping issue.

This seems odd, can you generate a signed test appx with and without the change and try it for yourself? This really doesn't seem like intended behavoir at all

Pinguin2001:
I will generate a test signed msix as well and update from that to the one with the other two dlls removed in a virtual machine and on a secondary pc. Will report back my tests

0x5bfa:
This doesn't work in CD because appxbundle we generate is configured to generate x64|arm64 and build is configured to generate artifacts for x64.
I like this size reduction improvement but unless we configure CD to build 2 times for each x64 and arm64 and then bundle them into appxbundle we can't enjoy this improvement.

yaira2:
Aside from the workaround to use msixbundler, are there any other paths forward?

0x5bfa:
I don't know. I guess it's easier.

yaira2:
@Pinguin2001 sorry for the delay in responding. We're definitely interested in implementing this feature, but our current packaging setup isn't compatible with this change, as @0x5bfa pointed out. We hope to make the necessary adjustments in the future, but right now, we lack the resources to do so, which means we'll have to put this on hold. 

We'll keep you updated on our progress in #16583. Also, it's worth mentioning that we're looking forward to a significant reduction in package size once we enable Native AOT, something we've been working towards for some time now. Keep an eye on the issues related to Vanara and the Windows Community Toolkit for more details.

PR #16585, opened 2024-12-08
Feature: Updated 7zip dependencies

yaira2:
**Resolved / Related Issues**

- Closes #

1. Opened Files ...
2. ...


hishitetsu (approved):

yaira2:
@hishitetsu I noticed the changelog for 7z mentions support for calculating file hashes. Is this of any use for us?

hishitetsu:
Currently, we are using the .NET implementation to compute the hash. I don't think it's worth to switch to 7-zip, since we still support other than SHA3-256. I don't think SHA3-256 is used much yet.

hishitetsu:
I just noticed that .NET also supports SHA3 since 8. So we can add it if we want to.

PR #16589, opened 2024-12-09
Feature: Removed toolbar from homepage

### LINKED ISSUE #16577: Feature: Hide toolbar when in homepage

Currently, the toolbar displays "home" when in the homepage:

Image

This isn't really necessary as "Home" is already displayed in the address bar and also the side pane. The toolbar should be hidden when inside the homepage for maximum UI simplicity

### Requirements

- Hide toolbar when in homepage

### Files Version

3.8.1.0

### Windows Version

10.0.22631.4541

### Comments

_No response_

Lamparter:
**Resolved / Related Issues**

- Closes #16577

1. Opened Files
2. Saw toolbar is hidden


0x5bfa:
ShowToolbar is completely different setting here. I don't think this works when you switch between tabs.

yaira2:
> ShowToolbar is completely different setting here. I don't think this works when you switch between tabs.

This is correct

Lamparter:
> ShowToolbar is completely different setting here. I don't think this works when you switch between tabs.

It does work

yaira2:
This will cause the toolbar to show even if the user turned it off from settings.

yaira2:
@Lamparter you can do this by modifying the toolbar visibility in `MainPage.xaml`.
```xml
Visibility="{x:Bind converters:MultiBooleanConverter.AndConvertToVisibility(SidebarAdaptiveViewModel.PaneHolder.ActivePaneOrColumn.InstanceViewModel.IsPageTypeNotHome, ViewModel.ShowToolbar), Mode=OneWay}" />
```

0x5bfa:
Well, I would do the same as what I do for displaying "Home" inside Toolbar.

yaira2:
Looking good! Can you please remove the header from the toolbar control now that it's no longer used?

Lamparter:
What do you mean?

yaira2:
In `InnerNavigationToolbar.xaml`.

PR #16595, opened 2024-12-10
Revert: "Fix: Fixed an issue where the title bar drag region was sometimes missing"

XTorLukas:
**Resolved / Related Issues**

- Closes #

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -60,7 +60,7 @@ namespace Files.App.UserControls.TabBar
 		public GridLength TitleBarWidth
 		{
 			get => _titleBarWidth;
-			set => _titleBarWidth = new((value.Value / MainWindow.Instance.GetDpiScale()) + _gap);
+			set => _titleBarWidth = new((value.Value / App.AppModel.AppWindowDPI) + _gap);
 		}
 
 		// Events
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -407,7 +407,15 @@ namespace Files.App.UserControls.TabBar
 		}
 
 		private void DragAreaRectangle_Loaded(object sender, RoutedEventArgs e)
-			=> HorizontalTabView.Measure(new(HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - TitleBarWidth.Value, HorizontalTabView.ActualHeight));
+		{
+			if (HorizontalTabView.ActualWidth > 0 && TabBarAddNewTabButton.Width > 0 && TitleBarWidth.Value > 0)
+			{
+				HorizontalTabView.Measure(new(HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - TitleBarWidth.Value, HorizontalTabView.ActualHeight));
+				return;
+			}
+
+			DispatcherQueue.TryEnqueue(() => DragAreaRectangle_Loaded(sender, e));
+		}
 
 		public int SetTitleBarDragRegion(InputNonClientPointerSource source, SizeInt32 size, double scaleFactor, Func<UIElement, RectInt32?, RectInt32> getScaledRect)
 		{
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -60,7 +60,7 @@ namespace Files.App.UserControls.TabBar
 		public GridLength TitleBarWidth
 		{
 			get => _titleBarWidth;
-			set => _titleBarWidth = new(value.Value + _gap);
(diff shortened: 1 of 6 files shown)


src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -360,13 +360,27 @@ private void TabViewItem_Loaded(object sender, RoutedEventArgs e)
 
 		private void DragAreaRectangle_Loaded(object sender, RoutedEventArgs e)
 		{
-			double scaleAdjustment = DragAreaRectangle.XamlRoot.RasterizationScale;
-			double titleBarInset = ((FilePropertiesHelpers.FlowDirectionSettingIsRightToLeft
-				? MainWindow.Instance.AppWindow.TitleBar.LeftInset
-				: MainWindow.Instance.AppWindow.TitleBar.RightInset) / scaleAdjustment) + 40;
+			if (HorizontalTabView.ActualWidth > 0 && TabBarAddNewTabButton.Width > 0)
```
yaira2, line 363: Isn't HorizontalTabView already loaded at this point?

yaira2: The downside to this logic is that it can result in a loop if an issue causes the value to remain 0. I suggest adding a short delay to accommodate slower devices and, if the value is still 0, we can use a fallback value for that device. However, I expect a short delay to be more than adequate.

XTorLukas: The solution may be to add only one repeat, since it is queued at the end and called when everything is ready.

yaira2: That's better, however I still prefer the logic of adding a short delay.

XTorLukas: If I add a delay, the problem is solved on slow machines, but the delay may be visible on other machines. This workaround will take care of the fast machines, and the slow machines will have a delay due to queuing, but at the same time the queuing will only be done once. "Personally, I try to avoid Delay when programming."

yaira2: I mean like this
```cs
private async void DragAreaRectangle_Loaded(object sender, RoutedEventArgs e)
{
    if (HorizontalTabView.ActualWidth <= 0 || TabBarAddNewTabButton.Width <= 0)
        await Task.Delay(100);

    var appWindow = MainWindow.Instance.AppWindow;
    var titleBarInset = (FilePropertiesHelpers.FlowDirectionSettingIsRightToLeft
        ? appWindow.TitleBar.LeftInset
        : appWindow.TitleBar.RightInset) / DragAreaRectangle.XamlRoot.RasterizationScale;

    if (titleBarInset > 0)
    {
        titleBarInset += 40; // Add 40px gap
        RightPaddingColumn.Width = new(titleBarInset);

        HorizontalTabView.Measure(new(
            HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - titleBarInset,
            HorizontalTabView.ActualHeight));
    }
}
```

XTorLukas: OK, sorry, it wasn't in my mind, sure, maybe I'm too invasive in my thoughts. 😅

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -358,15 +358,28 @@ private void TabViewItem_Loaded(object sender, RoutedEventArgs e)
...
+				titleBarInset += 40; // Add 40px gap
+				RightPaddingColumn.Width = new(titleBarInset);
+
+				HorizontalTabView.Measure(new(
+					HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - titleBarInset,
+					HorizontalTabView.ActualHeight));
+				return;
+			}
+
+			RightPaddingColumn.Width = new(138); // fallback
```
yaira2, line 382:
```suggestion
			RightPaddingColumn.Width = new(titleBarInset >= 0 ? titleBarInset : 138);
			HorizontalTabView.Measure(new(HorizontalTabView.ActualWidth - TabBarAddNewTabButton.Width - titleBarInset, HorizontalTabView.ActualHeight));
```

src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -358,15 +358,19 @@ private void TabViewItem_Loaded(object sender, RoutedEventArgs e)
 			}
 		}
 
-		private void DragAreaRectangle_Loaded(object sender, RoutedEventArgs e)
+		private async void DragAreaRectangle_Loaded(object? sender, RoutedEventArgs e)
```
yaira2, line 361:
```suggestion
		private async void DragAreaRectangle_Loaded(object sender, RoutedEventArgs e)
```

yaira2 (approved):

yaira2:
Thank you

PR #16603, opened 2024-12-13
Fix: Fixed an issue where tags would sometimes display with the wrong color

### LINKED ISSUE #16548: Bug: Tag icons in the details layout may use the wrong color
### Description

When items are ordered in the `DetailsLayout` (change the sorting order or the grouping option), tag icons may use the wrong color.
The issue is caused by `src/Files.App.Controls/ThemedIcon/ThemedIcon.xaml`. The binding on the color property at line `153-154` is not working.
I know how to fix this only by manually setting the property, but I don't think that is the right way to proceed.

https://github.com/user-attachments/assets/4d932548-d35d-42ee-bed3-c59558af2a5b

### Steps To Reproduce

1. Open a folder containing some tagged items
2. Change sort order
3. See that colors get messed up

### Files Version

3.8.0.0

### Windows Version

10.0.22631.4541

### User ID

8467f272-08cd-4a50-93e0-39458990bdc4

### Log File

Not relevant

ferrariofilippo:
**Resolved / Related Issues**

- Closes #16548 

1. Open a folder containing some tagged items
2. Change sort order
3. Then, check that tag icons in the following controls are using the appropiate color : `Settings/Tags`, `Sidebar>Tags`, `Home>Tags Widget`, `Preview Pane`, `Right click context menu`

**Notes**
I changed the bindings (I think we may remove that `VisualState` - our `DependencyProperty` doesn't seem to be bindable).
I don't know if this is the appropriate way to address the issue

### DIFF
--- a/src/Files.App.Controls/ThemedIcon/ThemedIcon.Properties.cs
+++ b/src/Files.App.Controls/ThemedIcon/ThemedIcon.Properties.cs
@@ -32,7 +32,7 @@ namespace Files.App.Controls
 		protected virtual void OnColorPropertyChanged(Brush oldValue, Brush newValue)
 		{
 			OnIconTypeChanged();
-			UpdateColor(newValue);
+			OnIconColorChanged(newValue);
 		}
 
 		protected virtual void OnIconTypePropertyChanged(ThemedIconTypes oldValue, ThemedIconTypes newValue)
--- a/src/Files.App.Controls/ThemedIcon/ThemedIcon.Properties.cs
+++ b/src/Files.App.Controls/ThemedIcon/ThemedIcon.Properties.cs
@@ -32,6 +32,7 @@ namespace Files.App.Controls
 		protected virtual void OnColorPropertyChanged(Brush oldValue, Brush newValue)
 		{
 			OnIconTypeChanged();
+			UpdateColor(newValue);
 		}
 
 		protected virtual void OnIconTypePropertyChanged(ThemedIconTypes oldValue, ThemedIconTypes newValue)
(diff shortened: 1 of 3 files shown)


src/Files.App.Controls/ThemedIcon/ThemedIcon.xaml
```diff
@@ -150,8 +150,8 @@
 								</VisualState>
 								<VisualState x:Name="Custom">
 									<VisualState.Setters>
-										<Setter Target="PART_OutlinePath.Fill" Value="{Binding RelativeSource={RelativeSource Mode=TemplatedParent}, Path=Color}" />
-										<Setter Target="PART_FilledPath.Fill" Value="{Binding RelativeSource={RelativeSource Mode=TemplatedParent}, Path=Color}" />
+										<Setter Target="PART_OutlinePath.Fill" Value="{Binding Color, Mode=OneTime}" />
```
yaira2, line 153: Shouldn't `mode` be `OneWay`?

ferrariofilippo: There are some issues with binding because of how the `Color` property is implemented. Still I just found out my current fix is causing another issue in the home page...
I'll mark this as `Ready` as soon as I find a solution

src/Files.App.Controls/ThemedIcon/ThemedIcon.Properties.cs
```diff
@@ -32,6 +32,7 @@ protected virtual void OnOutlineIconPropertyChanged(string oldValue, string newV
 		protected virtual void OnColorPropertyChanged(Brush oldValue, Brush newValue)
 		{
 			OnIconTypeChanged();
+			UpdateColor(newValue);
```
yaira2, line 35: I feel like this should probably be 
```suggestion
			OnIconColorChanged(newValue);
```

0x5bfa:
I think you don't need newValue parameter, just use DP value

yaira2 (approved):

PR #16653, opened 2024-12-23
Fix: Fixed keyboard navigation in the address path bar

### LINKED ISSUE #7077: Bug: Can't enter edit mode on path bar using the keyboard
### Description

Its seems not possible to enter edit mode in the navigation bar when using only keyboard. Since the focus jumps to a bread crump and from there between other crumps and other controls, there is no way to actually edit the location.
I expected F2 to work, or having a stop at an extra edit field after the crumps, but nope. nothing of the likes. 

To be fair, explorer doesnt seem to offer this either, but at least in explorer i can select folders from the bread crumps.

### Steps To Reproduce

Use the tab key to jump between the controls in the app until you reach the navigation bar and the bread crumps. 

### Expected behavior

Some way to edit the location in the navigation bar with keyboard only.

### Files Version

Version: 2.0.34.0 

### Windows Version

OS Version: 10.0.19042.1348

### Relevant Assets

_No response_

yaira2:
**Resolved / Related Issues**

- Closes #7077

1. Pressing the Tab key when the Refresh button (or Home button) is focused will switch the focus to the breadcrumb bar and enable edit mode.
2. Pressing the Tab key again will turn off edit mode and allow navigation within the breadcrumb bar.

### DIFF
--- a/src/Files.App/UserControls/AddressToolbar.xaml
+++ b/src/Files.App/UserControls/AddressToolbar.xaml
@@ -428,6 +428,7 @@
 				BorderBrush="{ThemeResource ControlElevationBorderBrush}"
 				BorderThickness="1"
 				CornerRadius="{StaticResource ControlCornerRadius}"
+				GettingFocus="ClickablePath_GettingFocus"
 				PointerPressed="ManualPathEntryItem_Click"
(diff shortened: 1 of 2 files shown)


0x5bfa:
Can you use nameof(HomeButton) instead.

Issue #12490, opened 2023-05-30
✨ Dropzone


Lamparter:
### What feature or improvement do you think would benefit Files?

Dropzone: a space to drag files into temporarily and then move them/copy them into different places
###### 📝 Not sure whether the files should be _copied_ into their new location or _moved_ into their new location.
The Dropzone will change size/disappear based on the amount of space in the sidebar (see the **mockup**)
Ideally it would also be inside an expander like these that is _open_ by default.
![image](https://github.com/files-community/Files/assets/71598437/63718b4f-4ae6-47ee-b0d6-e47a0051f52f)

---
Once inside the dropzone, there will be a preview of the file and manipulation options to clear the dropzone, etc. 
![image](https://github.com/files-community/Files/assets/71598437/1f10ab64-a2f9-4485-86a5-590cc6dfb0f2)
### Requirements

- Implement model to find files in the temporary folder 
	- Implement storage model for storing new dropzone files
###### 📝 Either Files can manually index the folder or a `json` file can be created with details about the Dropzone contents.
- Implement model to switch between the 'Drag and drop files here' and `GridView` views
- Implement automatic size changing/hiding in XAML
- Add support for manipulating temporary files: delete, open, etc.
- Add a view within Files; an extended page with temp files in it (or just open the appropriate `AppData` folder) when the 'open' button is clicked in the dropzone file manipulation view
![image](https://github.com/files-community/Files/assets/71598437/478bdb36-cb37-4def-9ad5-ba94a7370aa7)

### Files Version

2.4.83.0

### Windows Version

Windows 11 22H2 Insider Beta 10.0.22631.1825

### Comments

Quick mockup:
![image](https://github.com/files-community/Files/assets/71598437/ab56a3e4-cead-426a-a9b4-5d71714f3146)
(in the sidebar)

###### 🗃️ Dropzone files should be stored in a temp folder in `%appdata%`

meemore:
![Screenshot_19](https://github.com/files-community/Files/assets/124497263/033f9aa4-eff0-403c-a8d1-637b754b8ff7)
Drop zone call it what you want.  you want a FIXED area to drag files too,  I prefer it to be located in the details and preview panel.    So it doesn't  matter what folder you go to anyplace, it stays there until you need to drag it into a new folder.   All good news  other have been thinking about this also.  Cheers

yaira2:
Some details that were discussed internally:
- 'Dropzone' is not an ideal name for this feature. Suggestions are welcome.
[...]

PR #16673, opened 2025-01-06
Code Quality: Added design and basic functionality for the upcoming Shelf feature

d2dyno1:
**Resolved / Related Issues**
This is a continuation for the Shelf feature #12490

- Dropped a couple of files and folders to validate they're being added to the list

**Screenshots**
image

### DIFF
--- a/src/Files.App/Data/Items/ShelfItem.cs
+++ b/src/Files.App/Data/Items/ShelfItem.cs
@@ -1,0 +1,32 @@
+﻿using Files.Shared.Utils;
+
+namespace Files.App.Data.Items
+{
+	[Bindable(true)]
+	public sealed partial class ShelfItem : ObservableObject, IWrapper<IStorable>, IAsyncInitialize
+	{
+		private readonly IImageService _imageService;
+
+		[ObservableProperty] private IImage? _Icon;
+		[ObservableProperty] private string? _Name;
+		[ObservableProperty] private string? _Path;
+
+		/// <inheritdoc/>
+		public IStorable Inner { get; }
+
+		public ShelfItem(IStorable storable, IImage? icon = null)
+		{
+			_imageService = Ioc.Default.GetRequiredService<IImageService>();
+			Inner = storable;
+			Icon = icon;
+			Name = storable.Name;
+			Path = storable.Id;
+		}
+
+		/// <inheritdoc/>
+		public async Task InitAsync(CancellationToken cancellationToken = default)
+		{
+			Icon = await _imageService.GetIconAsync(Inner, cancellationToken);
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


src/Files.App/UserControls/Pane/ShelfPane.xaml.cs
```diff
@@ -1,15 +1,102 @@
 // Copyright (c) Files Community
 // Licensed under the MIT License.
 
+using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
+using System.Runtime.InteropServices.ComTypes;
+using System.Windows.Input;
+using Vanara.PInvoke;
```
yaira2, line 8: Is `Vanara` supposed to be referenced?

src/Files.App/UserControls/Pane/ShelfPane.xaml
yaira2, line 1: Is it okay if I push a commit to localize the strings?

src/Files.App/UserControls/Pane/ShelfPane.xaml.cs
d2dyno1: Yes, it's for the `ListView_DragItemsStarting` method (Just like `FileList_DragItemsStarting` references it too)

src/Files.App/UserControls/Pane/ShelfPane.xaml
d2dyno1: I don't mind

src/Files.App/Data/Items/ShelfItem.cs
```diff
@@ -0,0 +1,32 @@
...
+			Inner = storable;
+			Icon = icon;
+			Name = storable.Name;
+			Path = storable.Id;
+		}
+
+		/// <inheritdoc/>
+		public async Task InitAsync(CancellationToken cancellationToken = default)
+		{
+			Icon = await _imageService.GetIconAsync(Inner, cancellationToken);
```
yaira2, line 29: Are we still using image service? I believe we've switched to `FileThumbnailHelper.GetIconAsync`.

d2dyno1: `FileThumbnailHelper.GetIconAsync` returns a `byte[]` array that is hard to work with and would break (or require workarounds) in the view model to support images. `ImageService`, on the other hand, returns an `IImage` that already contains a BitmapImage ready to be parsed and extracted by the UI side. It uses FileThumbnailHelper in the underlying implementation anyway

src/Files.App/UserControls/Pane/ShelfPane.xaml.cs
yaira2: Ideally we'd like to avoid adding references to Vanara but I think we can address this when we address the reference in `FileList_DragItemsStarting`.

src/Files.App/UserControls/Pane/ShelfPane.xaml
yaira2: Thanks

yaira2 (approved):
LGTM

PR #16695, opened 2025-01-09
Feature: Display storage information for phones and tablets

### LINKED ISSUE #6872: Feature: Display smartphone/tablet storage size and free space
### What's the Problem?

The only thing missing for me is having those informations for Internal or SD Card .

**------------EXPLORER:**

**------------FILES:**

**Requirements**
- Display space progress bar for phones and tablets

### Files Version

v2.0.34.0

### Windows Version

Windows 11 21H2 22000.318

### Additional comment

_No response_

ferrariofilippo:
**Resolved / Related Issues**

- Closes #6872 

**Notes**
I proposed two possible working solutions (one commented out):
- the first one performs a double check
- the second goes with only one check (which should be enough in most cases)

Tell me which one do you find more appropriate

**Screenshots**
Screenshot 2025-01-09 215554
Screenshot 2025-01-09 215559

1. Open `Files` Home
2. Enable `Drives` widget
3. Connect a phone/tablet
4. Check that storage is shown
5. *IMPORTANT*: if you have the opportunity, test this feature using a phone with an SD card (may be source of bugs)

### DIFF
--- a/src/Files.App/Data/Items/DriveItem.cs
+++ b/src/Files.App/Data/Items/DriveItem.cs
@@ -6,6 +6,7 @@ using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Media.Imaging;
 using Windows.Storage;
+using Windows.Storage.Provider;
 using Windows.Storage.Streams;
 using ByteSize = ByteSizeLib.ByteSize;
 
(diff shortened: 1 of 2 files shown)


src/Files.App/Data/Items/DriveItem.cs
```diff
@@ -278,7 +279,37 @@ public async Task UpdatePropertiesAsync()
...
+				//}
+				//----------------------------------------------
+
+				var propertiesSource = Root;
+				if (string.IsNullOrEmpty(Root.Path) &&
+					Path.StartsWith(@"\\?\", StringComparison.Ordinal) &&
+					(await Root.GetFoldersAsync())[0] is StorageFolder systemFolder)
+				{
+					propertiesSource = systemFolder;
+				}
```
yaira2, line 310:
```suggestion
				var propertiesSource = (string.IsNullOrEmpty(Root.Path) &&
                        Path.StartsWith(@"\\?\", StringComparison.Ordinal) &&
                        (await Root.GetFoldersAsync())[0] is StorageFolder systemFolder) 
                        ? systemFolder 
                        : Root;
```

yaira2:
> IMPORTANT: if you have the opportunity, test this feature using a phone with an SD card (may be source of bugs)
 
This appears to be an issue because the widget only displays storage information for the device, and not the SD card. Instead of showing this info on the drives widget, it might be better to display it within the folder view when you open the device in Files, just like File Explorer.

ferrariofilippo:
> The details layout should be the easiest to update. We can add some additional columns and make them visible by default when opening the right type of drive. Regarding the implementation, is there a way to determine when to load the details?

At the moment we're just checking whether the drive path starts with `\\?\` , and if so we consider it an MTP device. Following this approach we can see if the folder path is at the `root` level for its drive.
Still, I see this is not the best approach (not all paths starting with `\\?\` refers to an MTP device) and I'll try to find a different solution

yaira2:
Using the foundation added in #16717, we should be able to display the storage information for the contextual property.

yaira2:
@ferrariofilippo thank you for this PR. Building on this foundation and the work in #16717, I've opened #16729.

PR #16730, opened 2025-01-23
Feature: Added an option to change the default IDE

### LINKED ISSUE #12857: Feature: Add a setting to change the "open" ide on the status bar

Hey there!

I'm really enjoying the new git integration with Files, but there's one small thing I'm missing.
I use code-insiders as my daily driver, the beta version of visual studio code, and the "VS Code" button doesn't seem to work. Since insiders functions the same as regular VS Code, I tried a couple of ways to make this work, including creating a shortcut called code.exe and adding it to the path, but it doesn't seem like that does the trick.

I did read that adding support for other IDE's is being considered, but I think having an optional setting pointing at the code.exe you wish to use would be a great consideration for those of us who use VS code alternatives like Insiders or VS Codium.

### Requirements

- Add an optional setting to user_settings.json that can point at an exe.
- When the setting is filled in, instead of using the discovered VS Code installation, use the exe the new setting points at to launch VS Code when the VS Code button is pressed in the right bottom of the screen.
- When the setting is an empty string (I assume this is the way this file usually functions, when I look at "AppThemeAddressBarBackgroundColor" and similar settings) or the setting is absent, use the originally discovered VS Code executable.

### Files Version

2.5.11.0

### Windows Version

10.0.23486.1000

### Comments

Thanks for your time and hard work!

ferrariofilippo:
**Resolved / Related Issues**

- Closes #12857 

1. Open `Settings>Dev Tools`
2. Play with IDE settings
3. Open a git repo and use the `Open in ...` command

Screenshot 2025-01-23 152220 Screenshot 2025-01-23 152835
Screenshot 2025-01-23 152237
Screenshot 2025-01-23 152829
Screenshot 2025-01-23 152604
Screenshot 2025-01-23 152810
Screenshot 2025-01-23 153205


src/Files.App/Views/Settings/DevToolsPage.xaml
```diff
@@ -21,6 +22,14 @@
...
 
 			<converters:BoolNegationConverter x:Key="BoolNegationConverter" />
+			<wctconverters:BoolToVisibilityConverter
+				x:Key="BoolVisibilityConverter"
+				FalseValue="Collapsed"
+				TrueValue="Visible" />
+			<wctconverters:BoolToVisibilityConverter
+				x:Key="InvertedBoolVisibilityConverter"
+				FalseValue="Visible"
+				TrueValue="Collapsed" />
```
Lamparter, line 32: WinUI now allows you to implicitly convert bool to visibility, so using WCT's one is unnecessary here.
As for the inverted bool converter, there's a local one.

ferrariofilippo: Actually it doesn't seem to work. When I open the page I get an `InvalidCastException` saying `Unable to cast object of type 'System.Boolean' to type 'Microsoft.UI.Xaml.Visibility'`

src/Files.App/Actions/Open/OpenRepoInIDEAction.cs
```diff
@@ -0,0 +1,82 @@
...
+				OnPropertyChanged(nameof(IsExecutable));
+			}
+			else if (e.PropertyName == nameof(IDevToolsSettingsService.FriendlyIDEName))
+			{
+				OnPropertyChanged(nameof(Label));
+				OnPropertyChanged(nameof(Description));
+			}
+		}
+
+		private async Task ShowErrorDialog()
```
yaira2, line 63: Can this be moved to the dialog helper? This way the dialog can be reused.

src/Files.App/Services/Settings/DevToolsSettingsService.cs
```diff
@@ -18,6 +20,20 @@ public OpenInIDEOption OpenInIDEOption
 			set => Set(value);
 		}
 
+		/// <inheritdoc/>
+		public string IDEPath
+		{
+			get => Get(_isVSCodeInstalled ? "code" : string.Empty) ?? string.Empty;
```
yaira2, line 26:
```suggestion
			get => Get(SoftwareHelpers.IsVSCodeInstalled() ? "code" : string.Empty) ?? string.Empty;
```

src/Files.App/ViewModels/Settings/DevToolsViewModel.cs
```diff
@@ -9,10 +9,16 @@ public sealed partial class DevToolsViewModel : ObservableObject
...
+		private readonly ICommonDialogService CommonDialogService = Ioc.Default.GetRequiredService<ICommonDialogService>();
 
 		public Dictionary<OpenInIDEOption, string> OpenInIDEOptions { get; private set; } = [];
 		public ICommand RemoveCredentialsCommand { get; }
 		public ICommand ConnectToGitHubCommand { get; }
+		public ICommand StartEditingIDECommand { get; }
+		public ICommand CancelIDEChangesCommand { get; }
+		public ICommand SaveIDEChangesCommand { get; }
+		public ICommand OpenFilePickerForIDECommand { get; }
+		public ICommand TestIDECommand { get; }
```
yaira2, line 21: Is this supposed to be removed before merging the PR?

ferrariofilippo: Currently not, why do you think we should remove them? 🤔

yaira2: My understanding was that `TestIDECommand` was to make it easier to debug without having to constantly reopen the settings dialog.

ferrariofilippo: Ideally yes, but I'm not sure how we should handle the testing: we need to launch the editor to see if it works, and it's not a good idea to kill the process (as it is now - I can change the code to get the ID of the launched  process). We cannot test it at each input event (obviously).
Should we test it only on save?

src/Files.App/ViewModels/Settings/DevToolsViewModel.cs
```diff
@@ -62,5 +144,51 @@ public async void DoConnectToGitHubAsync()
...
+		{
+			CommonDialogService.Open_FileOpenDialog(
+				MainWindow.Instance.WindowHandle,
+				false,
+				["*.exe;*.bat;*.cmd;*.ahk"],
+				Environment.SpecialFolder.ProgramFiles,
+				out var filePath
+			);
+
+			IDEPath = filePath;
```
yaira2, line 181: We may want to check the dialog result in case the user canceled the picker.

src/Files.App/Views/Settings/DevToolsPage.xaml
```diff
@@ -49,6 +58,200 @@
...
+								Text="{x:Bind ViewModel.IDEPath, Mode=TwoWay}"
+								TextTrimming="CharacterEllipsis"
+								Visibility="{x:Bind ViewModel.IsEditingIDEConfig, Converter={StaticResource InvertedBoolVisibilityConverter}, Mode=OneWay}" />
+
+							<StackPanel
+								Grid.Column="1"
+								Orientation="Horizontal"
+								Spacing="8"
+								Visibility="{x:Bind ViewModel.IsEditingIDEConfig, Converter={StaticResource BoolVisibilityConverter}, Mode=OneWay}">
+								<TextBox
```
yaira2, line 98: Can you add placeholder text?

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4043,4 +4043,31 @@
...
+    <value>Invalid path</value>
+  </data>
+  <data name="Test" xml:space="preserve">
+    <value>Test</value>
+  </data>
+  <data name="SelectedIDENotValid" xml:space="preserve">
+    <value>Selected IDE is not valid</value>
+  </data>
+  <data name="IDEError" xml:space="preserve">
+    <value>IDE Error</value>
```
yaira2, line 4062:
```suggestion
    <value>{0} could not be located</value>
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4043,4 +4043,31 @@
...
+  <data name="InvalidPath" xml:space="preserve">
+    <value>Invalid path</value>
+  </data>
+  <data name="Test" xml:space="preserve">
+    <value>Test</value>
+  </data>
+  <data name="SelectedIDENotValid" xml:space="preserve">
+    <value>Selected IDE is not valid</value>
+  </data>
+  <data name="IDEError" xml:space="preserve">
```
yaira2, line 4061:
```suggestion
  <data name="IDENotLocatedTitle" xml:space="preserve">
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4043,4 +4043,31 @@
...
+  <data name="PathOrAlias" xml:space="preserve">
+    <value>Path or alias</value>
+  </data>
+  <data name="InvalidPath" xml:space="preserve">
+    <value>Invalid path</value>
+  </data>
+  <data name="Test" xml:space="preserve">
+    <value>Test</value>
+  </data>
+  <data name="SelectedIDENotValid" xml:space="preserve">
```
yaira2, line 4058:
```suggestion
  <data name="IDENotLocatedContent" xml:space="preserve">
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4043,4 +4043,31 @@
...
+    <value>Selected IDE is not valid</value>
```
yaira2, line 4059:
```suggestion
    <value>The configured location for {0} is not valid. Please update the path and try again.</value>
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4043,4 +4043,31 @@
...
+  </data>
+  <data name="IDEError" xml:space="preserve">
+    <value>IDE Error</value>
+  </data>
+  <data name="EditInSettings" xml:space="preserve">
+    <value>Edit in settings</value>
+  </data>
```
yaira2, line 4066:
```suggestion
  <data name="OpenSettings" xml:space="preserve">
    <value>Open settings</value>
  </data>
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4043,4 +4043,31 @@
...
+  <data name="VisualStudioCode" xml:space="preserve">
+    <value>Visual Studio Code</value>
+  </data>
+  <data name="IDEPathPlaceholder" xml:space="preserve">
+    <value>code</value>
```
yaira2, line 4071:
```suggestion
    <value>Enter a path or launch alias</value>
```

src/Files.App/Views/Settings/DevToolsPage.xaml
```diff
@@ -49,6 +59,166 @@
...
+									x:Name="IDEPathTextBox"
+									Width="300"
+									VerticalAlignment="Center"
+									PlaceholderText="{helpers:ResourceString Name=IDEPathPlaceholder}"
+									Text="{x:Bind ViewModel.IDEPath, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}">
+									<TextBox.Resources>
+										<SolidColorBrush x:Key="TextControlBackgroundFocused" Color="{ThemeResource SolidBackgroundFillColorSecondary}" />
```
yaira2, line 156: Is there any reason not to use the default TextBox style?

src/Files.App/Views/Settings/DevToolsPage.xaml
```diff
@@ -76,5 +246,38 @@
 			</local:SettingsBlockControl>
 
 		</StackPanel>
+
+		<VisualStateManager.VisualStateGroups>
+			<VisualStateGroup>
+				<VisualState x:Name="DefaultState">
```
yaira2, line 252: If the main goal is to adjust the browse button's position, we can simplify the visual states by using a StackPanel and changing the orientation based on the page width.

src/Files.App/Views/Settings/DevToolsPage.xaml
ferrariofilippo: I didn't use the default style to stay consistent with other UI elements (e.g. TextBoxes in `Settings>Tags`)

src/Files.App/Views/Settings/DevToolsPage.xaml
ferrariofilippo: I'm using a grid to place elements in the read-only UI, which flows in both directions.
I think that removing the grid would require other 2/3 stackpanels.

BTW I found there are a couple of states to be removed (`PickIDEExe.(Grid.*)`)

src/Files.App/Views/Settings/DevToolsPage.xaml
yaira2: That's probably an oversight, we can remove that as well.

src/Files.App/Views/Settings/DevToolsPage.xaml
yaira2: 👍

src/Files.App/Views/Settings/DevToolsPage.xaml
ferrariofilippo: Just seen is used in the `Edit Tag Name` textbox but not in the `New Tag Name` textbox. Those are placed on different surfaces, so that might be by design and not by oversight.

For the moment I'll start removing the style only from `DevsPage`

yaira2: Thanks

src/Files.App/Views/Settings/DevToolsPage.xaml
```diff
@@ -49,6 +59,164 @@
...
+								<TextBox.Resources>
+									<TeachingTip
+										x:Name="InvalidIDENameWarning"
+										Title="{helpers:ResourceString Name=InvalidIDENameError}"
+										IsLightDismissEnabled="True"
+										IsOpen="{x:Bind ViewModel.IsIDENameValid, Converter={StaticResource BoolNegationConverter}, Mode=OneWay}"
+										PreferredPlacement="Bottom"
+										Target="{x:Bind IDENameTextBox}" />
+								</TextBox.Resources>
+								<TextBox.Header>
```
yaira2, line 115: We can simplify this by using `Header=""`

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4043,4 +4043,31 @@
...
+    <value>Open settings</value>
+  </data>
+  <data name="VisualStudioCode" xml:space="preserve">
+    <value>Visual Studio Code</value>
+  </data>
+  <data name="IDEPathPlaceholder" xml:space="preserve">
+    <value>Enter a path or launch alias</value>
+  </data>
+  <data name="InvalidIDENameError" xml:space="preserve">
+    <value>Please, enter a name for the IDE</value>
```
yaira2, line 4071:
```suggestion
    <value>Please enter a name for the IDE</value>
```

src/Files.App/Views/Settings/DevToolsPage.xaml
```diff
@@ -11,6 +11,8 @@
 	xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
 	xmlns:uc="using:Files.App.UserControls"
 	xmlns:vm="using:Files.App.ViewModels.Settings"
+	xmlns:wctconverters="using:CommunityToolkit.WinUI.Converters"
+	SizeChanged="Page_SizeChanged"
```
yaira2, line 15: Can we use the built-in triggers for screen size?

yaira2: > If you mean something like this, it works!

Yes

> is there a predefined width we should use as a breakpoint?

It should be able to fit the text box and browse button, with a little extra space for other localizations where the browse text may be longer.

ferrariofilippo: This is the min size in the latest commit
image

ferrariofilippo: BTW I used two `AdaptiveTriggers` to prevent a bug where the `Compact` layout was not triggered

yaira2 (approved):
LGTM

Issue #16658, opened 2024-12-27
Bug: Files no longer clears recent files


NoxStellans:
### Description

Before version 3.8.0.0, I could clear the recent files list on the home screen by right-clicking an item on the list and selecting 'Clear all items'. This no longer works; nothing happens if I press the button. Closing and reopening Files also doesn't do anything to clear the list.

### Steps To Reproduce

1. Open Files.
2. Right-click an item in the recent files list.
3. Choose 'clear all items'.

### Files Version

3.8.0.0

### Windows Version

10.0.19045.5247

### User ID

964c0dea-8575-4c4e-a05b-17ac3469f72c

### Log File

[debug.log](https://github.com/user-attachments/files/18257273/debug.log)

0x5bfa:
While we have changed P/Invoker from Vanara to CsWin32, haven't made changes to Clear All recent files functionality itself.

Before:
https://github.com/files-community/Files/blob/0debeee681f3443808ca49d47dc3e829c8a5e795/src/Files.App/Utils/RecentItem/RecentItems.cs#L187

After:
https://github.com/files-community/Files/blob/1ee9de3c86dfe7af47cee957cfa66317c3d36136/src/Files.App/Services/Windows/WindowsRecentItemsService.cs#L171

Josh65-2201:
It does clear it for me just doesn't update the widget and a full restart is needed. @NoxStellans can you confirm that it cleared if you close Files from the system tray and reopen.

NoxStellans:
Hi, the icon doesn't show in the system tray for me, even when I have that option enabled in advanced settings.

0x5bfa:
Can you kill the Files process from Task Manager?

NoxStellans:
Yes, and that updates it.

yaira2:
I can't reproduce this in v3.8, clearing all recent files updates the widget even without a restart.

NoxStellans:
Apologies, this is still happening for me on version 3.9.

PR #16732, opened 2025-01-24
Feature: Update the Drives Widget & Tags Widget when refreshing the Home Page

### LINKED ISSUE #6182: Feature: Auto refresh drive widget & tags widget when changes are detected

ferrariofilippo:
**Resolved / Related Issues**

- Closes #6182 

**Steps**
- [x] Make the `Refresh` button actually refresh the home page
- [x] Refresh `Tags` when changes are detected
- [x] Refresh `Tags` when files are tagged
- [ ] Refresh drives when files are added/removed

1. Open the home page
2. Play with tags/drives and recent files
3. See the home page widgets are refreshed


ferrariofilippo:
@yaira2 I'm not sure we should use the filewatcher to update drives. If a user is copying many items this would slow down the process

yaira2:
That could be a concern, we might be better off reloading the details only when the user refreshes the page.

Josh65-2201:
Couldn't it refresh when the home page is opened, that way it auto refreshes without the performance issue

ferrariofilippo:
> Couldn't it refresh when the home page is opened, that way it auto refreshes without the performance issue

This seems a good workaround, I'll look into it

yaira2:
>  Refresh Tags when changes are detected

Is this automatic, or does it require refreshing the page?

ferrariofilippo:
> > Refresh Tags when changes are detected
> 
> Is this automatic, or does it require refreshing the page?

It was automatic but I reverted it in the last commit. I can re-revert it if it's needed 
(It still works automatically when their name/colors are updated from settings)

yaira2:
So just to confirm, aside from when the settings change, the only live updating is when manually refreshing & when navigating to the home page?

ferrariofilippo:
> So just to confirm, aside from when the settings change, the only live updating is when manually refreshing & when navigating to the home page?

Currently the behaviour is as follows:
- When a drive is (un)mounted it is added to the `Drives` widget [even for network drives]
- When files are opened they are added to the `Recent files` widget
- When pinned folders are changed the `Quick Access` widget is updated

Introduced in this PR:
- Update the `Tags` widget when tags are changed from settings
- When manually refreshing the `Home` page or navigating to it,  update all the widgets

I can add back the code to update the `Tags` widget when items are tagged.
As mentioned above I don't think we should update the drives widget whenever items are moved

src/Files.App/ViewModels/UserControls/Widgets/RecentFilesWidgetViewModel.cs
```diff
@@ -68,6 +68,7 @@ public RecentFilesWidgetViewModel()
 		public async Task RefreshWidgetAsync()
 		{
 			IsRecentFilesDisabledInWindows = !CheckIsRecentItemsEnabled();
+			Items.Clear();
```
yaira2, line 71: The recent files widget already has a watcher that updates the list, is there any reason to manually clear the list?

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -171,6 +171,13 @@ public async void MultitaskingControl_CurrentInstanceChanged(object? sender, Cur
 			{
 				SidebarAdaptiveViewModel.PaneHolder = currentInstance;
 				SidebarAdaptiveViewModel.PaneHolder.PropertyChanged += PaneHolder_PropertyChanged;
+
+				if (currentInstance.ActivePaneOrColumn is ModernShellPage msh &&
+					msh.ItemDisplayFrame.Content is HomePage homePage &&
+					homePage.ViewModel.RequiresUpdate)
+				{
+					await homePage.ViewModel.RefreshWidgetProperties();
```
yaira2, line 179: Ideally this logic would be under `HomPage.xaml.cs`.

src/Files.App/ViewModels/UserControls/Widgets/RecentFilesWidgetViewModel.cs
ferrariofilippo: There's just one case where that is needed: when you clear `File Explorer` history (`File Explorer>Options>Privacy>Clear`).
In this scenario the watcher doesn't work and items remain visible after updating the home page

src/Files.App/Views/MainPage.xaml.cs
ferrariofilippo: Done

src/Files.App/ViewModels/UserControls/Widgets/RecentFilesWidgetViewModel.cs
yaira2: It works on my end, but there is a known issue where it doesn't work for some users. We're tracking this in #16658.

src/Files.App/Views/HomePage.xaml.cs
```diff
@@ -82,20 +83,23 @@ protected override void OnNavigatedFrom(NavigationEventArgs e)
 			Dispose();
 		}
 
-		private async void ToolbarViewModel_RefreshRequested(object? sender, EventArgs e)
+		private async void MultitaskingControl_CurrentInstanceChanged(object? sender, CurrentInstanceChangedEventArgs e)
```
yaira2, line 86: Does this fire when switching tabs?

yaira2: I haven't tested, but it looks like we can put this code under `OnNavigatedTo`.

ferrariofilippo: I tried it and it doesn't work. `OnNavigatedTo` is executed only when it's the first time you open the page

yaira2: What happens if you turn off caching for the page?

ferrariofilippo: > Does this fire when switching tabs?

Yes it is fired

ferrariofilippo: > What happens if you turn off caching for the page?

Just checked and it's already off, so it doesn't work

yaira2: Ideally it would only update when navigating or refreshing.

yaira2: > I tried it and it doesn't work. OnNavigatedTo is executed only when it's the first time you open the page

It appears to fire when I renavigate as well.

ferrariofilippo: I'll try to figure out why mine behaves differently

ferrariofilippo: I think we are talking about two different things:
- I'm saying it doesn't fire when you switch tab
- You are saying it fires when you navigate back/forward (which works for me too), am I right?

yaira2: Currently, this will reload the home page whenever a user switches between tabs. The expected behavior is for it to reload only if the user navigates to the home page or refreshes manually.

ferrariofilippo: In this case I can remove the call since widgets are already reloaded on navigation

PR #16745, opened 2025-01-28
Fix: Fixed issue where the size wasn't displayed after creating a new archive

### LINKED ISSUE #11580: Bug: Archive size is not updated at the creation
### Description

When creating an archive, the size of the archive is not displayed correctly in the view. Refreshing the page displays the correct size.

### Steps To Reproduce

1- Create an archive with at least one item.
2- See that the size is not correctly displayed.
3- Refresh the page to see the size correctly displayed.

SizeError

### Steps to fix

- Display the correct size when creating 

### Files Version

Version 2.4.40 Dev

### Windows Version

W10 Home 22H2 19045.2251

### Log file

Irrelevant.

Roilin-Lab:
**Resolved / Related Issues**

The problem is related to the fact that the archive, which is perceived as a folder, is not assigned a size in the `ShellViewModel.GetFileOrFolderUpdateInfoAsync()` method, unlike files, but we also do not want to assign a size to regular folders, for which this method is called, so for them the size should be left as null.
- Closes #11580

1. Create an archive using any available actions.
2. Make sure that the size is displayed correctly.

### DIFF
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -2496,6 +2496,7 @@ namespace Files.App.ViewModels
 				else if (storageItem.IsOfType(StorageItemTypes.Folder))
 				{
 					var properties = await storageItem.AsBaseStorageFolder().GetBasicPropertiesAsync();
+					size = item.IsArchive ? (long)properties.Size : null;
 					modified = properties.DateModified;
 					created = properties.DateCreated;
 				}
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -2496,7 +2496,7 @@ namespace Files.App.ViewModels
 				else if (storageItem.IsOfType(StorageItemTypes.Folder))
 				{
 					var properties = await storageItem.AsBaseStorageFolder().GetBasicPropertiesAsync();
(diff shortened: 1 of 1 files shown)


src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -2496,6 +2496,7 @@ private async Task AddFileOrFolderAsync(ListedItem? item)
 				else if (storageItem.IsOfType(StorageItemTypes.Folder))
 				{
 					var properties = await storageItem.AsBaseStorageFolder().GetBasicPropertiesAsync();
+					size = item.IsArchive ? (long)properties.Size : null;
```
yaira2, line 2499:
```suggestion
					size = (long)properties.Size;
```
This should do the trick

yaira2 (approved):
@Roilin-Lab thank you!

src/Files.App/ViewModels/ShellViewModel.cs
yaira2: GetBasicPropertiesAsync only returns the size for files, so it doesn't need to be conditional.

yaira2: On second thought, it seems to depend on the folder. I've added back the check just in case.

yaira2: That was the issue I had as well. In any case, I added back the check and it's working as expected.

PR #16557, opened 2024-12-04
Code Quality: Migrated shared properties to props file

Lamparter:
**Resolved / Related Issues**

- Closes #14826

**Steps used to test these changes**

Built the app; checked for Roslyn analyser warnings 

---

This PR finally adds a props file to manage mutual properties where appropriate. It has been built and tested and works perfectly. 

I also wanted to implement CPM however haven't succeeded yet.

PR #16770, opened 2025-02-02
Fix: Fixed an issue where `SetWindowLongPtr` didn't work on x86

Lamparter:
**Resolved / Related Issues**

This was previously fixed alongside #16557 however because it was closed this fix was lost.

1. Built Files

> Files doesn't build without this fix.

### DIFF
--- a/src/Files.App.CsWin32/Files.App.CsWin32.csproj
+++ b/src/Files.App.CsWin32/Files.App.CsWin32.csproj
@@ -11,7 +11,7 @@
         <RuntimeIdentifiers>win-x86;win-x64;win-arm64</RuntimeIdentifiers>
         <DefineConstants Condition="'$(Configuration)' == 'Debug'">TRACE;DEBUG;NETFX_CORE</DefineConstants>
         <DefineConstants Condition="'$(Configuration)' != 'Debug'">TRACE;RELEASE;NETFX_CORE</DefineConstants>
-		<DefineConstants>$(DefineConstants);$(Platform)</DefineConstants>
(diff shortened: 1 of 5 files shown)


0x5bfa:
I would define interop by myself instead of using the deprecated one.

0x5bfa:
Tagging https://github.com/microsoft/CsWin32/issues/882 tbh this should be platofrm-agnostically generating all requested.

Lamparter:
Does it really matter?

Lamparter:
@yaira2 can you review?

0x5bfa:
Why would you think it doesn't matter?

IMG_0440

0x5bfa:
Not sure yet what binding type is used for LONG_PTR but we should be able to cast or something to long

0x5bfa:
Why would I tell verbose

```c#
public static nint SetWindowLongPtrW(HWND hWnd, int nIndex, nint dwNewLong)
{
    if (sizeof(nint) == 4)
    {
        [DllImport("user32", EntryPoint = "SetWindowLongW", ExactSpelling = true)]
        static extern nint _SetWindowLong(HWND hWnd, int nIndex, nint dwNewLong);

        return _SetWindowLong(hWnd, nIndex, (int)dwNewLong);
    }
    else
    {
        [DllImport("user32", EntryPoint = "SetWindowLongPtrW", ExactSpelling = true)]
        static extern nint _SetWindowLongPtr(HWND hWnd, int nIndex, nint dwNewLong);

        return _SetWindowLongPtr(hWnd, nIndex, dwNewLong);
    }
}
```

0x5bfa:
#16781 supersedes.

PR #14423, opened 2024-01-10
Feature: Switched away from GetThumbnailAsync API

yaira2:
**Resolved / Related Issues**
- Simplified the code and improved the thumbnail quality
- Closes #14071
- Closes #10067

**Validation**
How did you test these changes?
- [x] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [ ] Did you implement any design changes to an existing feature?
   - [ ] Was this change approved?
- [x] Are there any other steps that were used to validate these changes?
   1. Check that folder thumbnails display a preview of their contents (except for in the Details & Column layouts)
   2. Clear the thumbnail cache in Windows
   3. Check that opening a folder with images or office documents generates thumbnails (in OneDrive you might have to refresh the folder to display the thumbnails)
   4. Open a location with a lot of images (in the grid layout) and confirm that all the thumbnails are loaded
   5. Check that turning off the option to show thumbnails works as expected (it should auto refresh the current tab)
   6. Check that the preview/details pane shows the thumbnail for folders and images
   7. Check that the Recent Items widget loads the thumbnails as expected

**Screenshots (optional)**
image

PR #16782, opened 2025-02-05
Fix: Fixed thumbails not loading in MTP devices

### LINKED ISSUE #15084: Bug: Thumbnails don't load for MTP devices (Phones)
### Description

I've connected my phone to my PC to share some files, but it's been several minutes, and the file app hasn't loaded the file thumbnails. Additionally, the file app is running very slowly.

### Steps To Reproduce

- Connect the phone to the PC using a USB cable or any designated method for file sharing.
- Open the file app on the phone or navigate to the file directory where thumbnails should appear.
- Wait for several minutes while observing the file app interface.
- Note that the file thumbnails fail to load within the expected timeframe.
- Additionally, observe any sluggishness or delays in the responsiveness of the file app during this 

### Requirements

- Ensure that the file app loads file thumbnails promptly after connecting the phone to the PC for file sharing
- The thumbnails should appear within a reasonable timeframe, ideally within a few seconds of connecting the devices.
- Additionally, optimize the performance of the file app to minimize any slowdowns or delays experienced during file browsing and sharing activities.

### Files Version

3.3.0.0

### Windows Version

10.0.22631.3374

### Log File

[debug.log](https://github.com/files-community/Files/files/14814408/debug.log)

gave92:
**Resolved / Related Issues**

- Closes #15084

Added back some code removed in #14423 which is necessary for loading thumbails in MTP.

1. Verify that icons are loaded in MTP devices
2. Verify that loading on normal drives is not impacted

### DIFF
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -973,7 +973,7 @@ namespace Files.App.ViewModels
 			return shieldIcon;
 		}
 
-		private async Task LoadThumbnailAsync(ListedItem item, CancellationToken cancellationToken)
+		private async Task<bool> LoadThumbnailAsync(ListedItem item, CancellationToken cancellationToken)
 		{
 			var loadNonCachedThumbnail = false;
 			var thumbnailSize = LayoutSizeKindHelper.GetIconSize(folderSettings.LayoutMode);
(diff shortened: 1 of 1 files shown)


src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -1086,6 +1086,40 @@ await dispatcherQueue.EnqueueOrInvokeAsync(async () =>
 					}
 				}, cancellationToken);
 			}
+
+			return result is not null;
+		}
+
+		private async Task<bool> LoadThumbnailAsync(ListedItem item, IStorageItem matchingStorageItem, CancellationToken cancellationToken)
```
yaira2, line 1093: Would it be worth implementing this logic as part of `FileThumbnailHelper`? This way the fix will apply throughout the app, and not just the file list.

yaira2:
I was thinking we could take this a step forward with a single method in `FileThumbnailHelper` that would first attempt using the Win32 method, and StorageItem as a fallback.

gave92:
Yeah so it should work e.g also when loading the thumbnail for the porperties window.

src/Files.App/Utils/Storage/Helpers/FileThumbnailHelper.cs
```diff
@@ -17,6 +18,22 @@ public static class FileThumbnailHelper
 			return await Win32Helper.StartSTATask(() => Win32Helper.GetIcon(path, (int)size, isFolder, iconOptions));
 		}
 
+		/// <summary>
+		/// Returns thumbnail for given file or folder using Storage API
+		/// </summary>
+		public static async Task<byte[]?> GetIconAsync(IStorageItem item, uint requestedSize, ThumbnailMode thumbnailMode, ThumbnailOptions thumbnailOptions)
```
yaira2, line 24: Is thumbnail mode always single?

gave92: Both ListView and SingleItem are used

gave92:
Ok now thumbnails work in file list, in properties window and preview pane 💪

src/Files.App/Utils/Storage/Helpers/FileThumbnailHelper.cs
yaira2: I think we can reduce complexity by just using SingleItem.

src/Files.App/ViewModels/UserControls/Previews/BasePreviewModel.cs
```diff
@@ -86,6 +86,7 @@ public async virtual Task<List<FileProperty>> LoadPreviewAndDetailsAsync()
 		{
 			var result = await FileThumbnailHelper.GetIconAsync(
 				Item.ItemPath,
+				Item.ItemFile,
```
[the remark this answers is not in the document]
yaira2: @gave92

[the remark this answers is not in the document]
gave92: I want to always try first with the path and only if this fails to try with StorageItem. I could use Item.ItemFile.Path but on some locations the Path property is empty.

src/Files.App/Utils/Storage/Helpers/FileThumbnailHelper.cs
gave92: Done!

yaira2: Thanks!

src/Files.App/ViewModels/UserControls/Previews/BasePreviewModel.cs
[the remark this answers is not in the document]
yaira2: Fair enough.

yaira2:
I'm concerned about maintaining two separate thumbnail systems (we did in the past, with much trouble). 
Is there anything we can do with `Win32Helper.GetIcon` and avoid the need to use StorageThumbnail?

yaira2:
> I'm concerned about maintaining two separate thumbnail systems (we did in the past, with much trouble). Is there anything we can do with `Win32Helper.GetIcon` and avoid the need to use StorageThumbnail?

@gave92 this is potentially solvable when we switch ShellViewModel to the new storage abstractions.

yair100:
@gave92 I opened a follow up PR in #18238.

PR #16804, opened 2025-02-12
Code Quality: Removed sln file

hez2010:
VS 17.13 and .NET 9.0.200 is out.
Let's migrate to the new slnx solution file.


hez2010:
We need to wait the windows-latest image to upgrade VS 17.13.

global.json
yaira2, line 1: Do we still need the global.json file?

0x5bfa:
Linking https://github.com/actions/runner-images/issues/11591

global.json
hez2010: We need to use `global.json` to ensure a minimum version of .NET SDK so that people won't use an incorrect one that doesn't support slnx.

yaira2: We'll need to update our docs to reflect this change

0x5bfa:
The latest toolset has 17.13 now. Can you rerun to test if it goes well?

hez2010:
Reopen to trigger CI.

yaira2 (approved):
LGTM 🎉

Issue #12704, opened 2023-06-23
SMB servers load 10x slower than in Windows' File Explorer


life036:
### Description

Just like the title says. In addition, I have a folder on my local HD that just has shortcuts to about a dozen SMB servers (standard shortcuts to locations like \\ServerName), and it takes 30-40 seconds for that folder to load, vs instantly in Windows' File Explorer.

Something strange is going on with the way Files interacts with SMB servers.

### Steps To Reproduce

1. Connect to any SMB server on your LAN (\\ServerName) and compare how quickly it loads to the native file explorer.

2. Create a local folder on your HD with shortcuts to a handful of SMB servers on your network, and compare how quickly it loads to the native file explorer

### Requirements

SMB locations, and folders full of shortcuts to SMB locations, should load just as fast (or nearly as fast) as the native File Explorer.

### Files Version

2.5.1.0

### Windows Version

10.0.22621.1848

### Log File

[debug.log](https://github.com/files-community/Files/files/11852798/debug.log)

Josh65-2201:
Thanks for the report. Can you give the following information.
- Is the issue with loading the contents of a network drive or the folder containing the links to the network drives
- How many items does the folder contain

life036:
> Thanks for the report. Can you give the following information.
> 
> * Is the issue with loading the contents of a network drive or the folder containing the links to the network drives

Both situations have similar issues

> * How many items does the folder contain

My particular folder has 10 items, but the issue happens no matter how many shortcuts you have in there. It's just more pronounced when you have more items in the folder. It also seems to stutter longer on servers that aren't available; like it's checking the validity of the shortcut before loading it.

Josh65-2201:
Thanks for the information, I've added this to the project board.

KFlab:
Since I am a non-native English speaker, I am sorry for my poor English。
I have same problem and I think I found some useful information。

I found a lot of download traffic coming from my SMB server，
so I use wireshark to capture analysis.

Here's the information I got:

step 1) 
there's a <GetInfo Request> package try to read one of my file's info,
The target file is a mod pack for a game, and it's a 47.53GB size and zip type.
I will call this file "a.zip" 

step 2)
SMB server send a TCP package back 
and a <GetInfo Response> shows success

step 3)
There 's a <Read Request> to "a.zip" with Length 65536 Offset 65536

step 4)
There 's a <Read Request> to "a.zip" again with Length 69632 Offset 51039236096
The file here seems Central Directory Record part of a zip file

[...]

PR #16812, opened 2025-02-15
Fix: Fixed performance issue with SMB shares

KFlab:
**Resolved / Related Issues**

Related to #12704
I'm not pro in c_sh,
This is only a temporary solution.
Because my work relies heavily on various mapped network drives,
Till now, this patch work for me.

1） Enter an smb that has been mapped as a network drive like "Y:\Download"
2）enter the directory containing a large number of large compressed files and wait for them to be lod.
3)  Observe whether the loading speed is the same with "\\192.168.0.1\nfs\Download"

ps:
The root cause may be the reading of the zip file，
Although it has not been tested, this is just a guess. Will huge files on slow devices have similar phenomena?
I also think that the compressed file recognition problem I mentioned in #12704 may aggravate this phenomenon.
Additionally, I would suggest that the file system type detection could be more unified into one function, rather than being everywhere.

### DIFF
--- a/src/Files.App/Services/Storage/StorageNetworkService.cs
+++ b/src/Files.App/Services/Storage/StorageNetworkService.cs
@@ -208,6 +208,25 @@ namespace Files.App.Services
 
 			unsafe
 			{
+
+   				if (!path.StartsWith(@"\\", StringComparison.Ordinal))
+				{
+					//  Special handling for network drives
+	 				//  This part will change path from "y:\Download" to "\\192.168.0.1\nfs\Download"
+					[DllImport("mpr.dll", CharSet = CharSet.Auto)]
+					static extern int WNetGetConnection(string lpLocalName, StringBuilder lpRemoteName, ref int lpnLength);
+					
+					StringBuilder remoteName = new StringBuilder(300);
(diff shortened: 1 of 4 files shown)


yaira2:
@KFlab thank you for the PR. Does this fully resolve #12704?

KFlab:
> @KFlab thank you for the PR. Does this fully resolve #12704?

Worked for me, the patch just adds support for network mapped drives in the directory check condition.
I think the root of this problem is that under the SMB protocol, reading compressed packages will take up a lot of IO
After trying, I found that directly using "//" to access the directory will not cause such a phenomenon
I'm not sure why you're using string matching instead of the SystemIO.DriveInfo function in the codebase though.
Due to the need for code quality, I think it is necessary to refactor most of the drive letter judgment parts

yaira2:
@KFlab thank you for your contribution. I merged the PR, but I would still like a second review before releasing to prod. Please be on the lookout in case further comments are posted.

0x5bfa:
This is causing an issue where loading a virtualized shell folder may not finish forever and this is because these folders dont start from a drive label.

PR #16815, opened 2025-02-15
Code Quality: Added root files to projects to make them easier to edit

Lamparter:
Adds `Directory.Packages.props`, `Directory.Build.props` and `.editorconfig` to all projects as `None` to make them appear in Solution Explorer. This makes them easier to edit.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,4 +7,9 @@
     <WindowsSdkPackageVersion>10.0.22621.57</WindowsSdkPackageVersion>
     <WindowsTargetFramework>$(TargetFrameworkVersion)-windows$(TargetWindowsVersion)</WindowsTargetFramework>
   </PropertyGroup>
+  <ItemGroup>
+    <None Include="$(MSBuildThisFileDirectory)\.editorconfig" />
+    <None Include="$(MSBuildThisFileDirectory)\Directory.Build.props" />
+    <None Include="$(MSBuildThisFileDirectory)\Directory.Packages.props" />
+  </ItemGroup>
 </Project>


Directory.Build.props
```diff
@@ -6,4 +6,9 @@
     <MinimalWindowsVersion>10.0.19041.0</MinimalWindowsVersion>
     <WindowsSdkPackageVersion>10.0.22621.57</WindowsSdkPackageVersion>
   </PropertyGroup>
+  <ItemGroup>
+    <None Include="$(MSBuildThisFileDirectory)\.editorconfig" />
+    <None Include="$(MSBuildThisFileDirectory)\Directory.Build.props" />
+    <None Include="$(MSBuildThisFileDirectory)\Directory.Packages.props" />
+  </ItemGroup>
```
hez2010, line 13: This will make them appear on all projects. Instead you should add them as solution items.
I would recommend holding up the change until we migrate to slnx.

hez2010 (changes requested):

Directory.Build.props
Lamparter: I'm aware of that

Lamparter: IMO it's a better experience to add them to the project directly as they apply to every project
If it was an explicit `.props` file that was in the actual project structure then I would add them to the solution, though since these files apply everywhere it makes sense to put them inside every project as None.

hez2010: You can create a virtual folder and add them to it in the solution file.

yaira2: @Lamparter we've just migrated to slnx and can move forward with these changes now.

Lamparter: Migrating to SLNX doesn't affect these changes.

hez2010:
I personally don't like this change as solution-wide files shouldn't appear as project-wide items, instead it should be placed under the solution hierarchy directly IMO. So I'm sorry but I hesitate to approve this change. 
While I'm not strongly against it if other reviewers think it's fine.

Lamparter:
They apply to all the projects; that's my logic for them being included in all of them.

marcofranzen99:
I would have also expected the files to be in a Solution Items folder. I think the files are also rarely needed and would rather distract from the individual projects when navigating in the Solution Explorer than in a separate Solution Items folder.

This is how i usually do it my private projects:
image

0x5bfa:
I don't approve either. It doesn't need to be everywhere and should not.

> This is how i usually do it my private projects:

I think this is also what Hez has in his mind too.

yaira2:
@Lamparter thank you for the PR. Currently the community would prefer not to adopt this change, but we appreciate the contribution as always 🙂

PR #16830, opened 2025-02-19
Fix: Fixed an issue where the status center didn't display the size

### LINKED ISSUE #16538: Bug: Status Center shows 0 transfer speed when copying folder
### Description

I was copying a folder from disk C: to D: , while it does show me the file being copied, transfer speed (MB/s) or item process speed (items/s) remains 0 during the entire process.
Image
I tested this using 3 different folders, and all resulted in the same. (around 300GB, 180GB and 50GB respectively)

It seems like a similar issue to #14974 

### Steps To Reproduce

1. copy a folder
2. go to another directory and paste it
3. open the Status Center

### Files Version

3.8

### Windows Version

10.0.22631.4541

### User ID

96a498d2-7152-4a22-8212-8e8ad62fdeb1

### Log File

[debug.log](https://github.com/user-attachments/files/17916121/debug.log)

primetime43:
**Resolved / Related Issues**

- Closes #16538

1. Opened Files ...
2. Copy a large folder to another drive and you should see the items being processed correctly on the UI
3. 
image


src/Files.App/Utils/Storage/Operations/FileSizeCalculator.cs
primetime43, line 1: Could probably get rid of the try on line 67, and either do something with the var error or get rid of it

yaira2: Please remove the try

src/Files.App/Utils/Storage/Operations/FileSizeCalculator.cs
```diff
@@ -73,24 +74,30 @@ unsafe void ComputeSizeRecursively(string path, CancellationToken token)
...
+										fileName.Equals("..", StringComparison.OrdinalIgnoreCase))
 									{
-										ComputeFileSize(itemPath);
+										continue;
 									}
-									else if (findData.cFileName.ToString() is string fileName &&
-										fileName.Equals(".", StringComparison.OrdinalIgnoreCase) &&
-										fileName.Equals("..", StringComparison.OrdinalIgnoreCase))
+
+									if (attributes.HasFlag(FILE_FLAGS_AND_ATTRIBUTES.FILE_ATTRIBUTE_DIRECTORY))
```
yaira2, line 85: Will this work for zip files?

primetime43: > Will this work for zip files?

Yes because it works with files and folders. At that point a zip file would be considered just a "file".

Josh65-2201 (commented):
Works for me with file only, folder only, file and folder on local drive, USB drives and network drives. 
~~It doesn't work when the SMB location is mapped as a drive, which works in v3.9.1~~ Ignore, it does work, I got version mixed up when testing

Josh65-2201 (approved):

src/Files.App/Utils/Storage/Operations/FileSizeCalculator.cs
yaira2: Great! Thank you.

PR #16853, opened 2025-02-24
Code Quality: Retarget BladeView to ContentControl

hez2010:
`Expander` is buggy and its `OnApplyTemplate` is not usable. Retarget it to `ContentControl` and remove the unused titlebar and expand (enlarge) features.

Resolves the missing styles and crashing issue.


hez2010:
> If you you like this SG for BladeItem for the sake of the simplicity:)

Nope. This would hide those properties from being seen by other source generators. Today a source generator cannot see anything generated by another source generator.

PR #16880, opened 2025-02-28
Fix: Fixed issue where the Recycle Bin was sometimes unpinned

### LINKED ISSUE #16676: Bug: Recycle bin removed from pinned on first install of any version
### Description

I have found when any version of Files is install and first opened `Shell:RecycleBinFolder` is removed

### Steps To Reproduce

1. Install Files stable
2. Open and pin `Shell:RecycleBinFolder`
3. Install Files Preview
4. Open Preview see Recycle bin missing from pins

### Files Version

3.8.10

### Windows Version

10.0.26100.2605

### User ID

3372b393-b294-495c-9814-4e5b48d9fd8d

### Log File

[debug.log](https://github.com/user-attachments/files/18320824/debug.log)

marcofranzen99:
**Resolved / Related Issues**
- Closes #16676

1. Uninstalled Files Dev
2. Pinned `Shell:RecycleBinFolder` via File Explorer
3. Installed Files Dev (Launched with Visual Studio)
4. Checked that `Shell:RecycleBinFolder` is still pinned

Additionally i have checked that `Shell:RecycleBinFolder` gets pinned at the first start of Files and that pinning/unpinning of folders via the context menu still works

### DIFF
--- a/src/Files.App/Services/Windows/WindowsQuickAccessService.cs
+++ b/src/Files.App/Services/Windows/WindowsQuickAccessService.cs
@@ -26,7 +26,12 @@ namespace Files.App.Services
 		private async Task PinToSidebarAsync(string[] folderPaths, bool doUpdateQuickAccessWidget)
 		{
 			foreach (string folderPath in folderPaths)
-				await ContextMenu.InvokeVerb("pintohome", [folderPath]);
+			{
+				// make sure that the item has not yet been pinned
+				// the verb 'pintohome' is for both adding and removing
+				if (!IsItemPinned(folderPath))
+					await ContextMenu.InvokeVerb("pintohome", folderPath);
+			}
 
 			await App.QuickAccessManager.Model.LoadAsync();
 			if (doUpdateQuickAccessWidget)
--- a/src/Files.App/Utils/Global/QuickAccessManager.cs
+++ b/src/Files.App/Utils/Global/QuickAccessManager.cs
@@ -42,11 +42,10 @@ namespace Files.App.Utils
 		public async Task InitializeAsync()
 		{
 			PinnedItemsModified += Model.LoadAsync;
+			await Model.LoadAsync();
 
 			if (!Model.PinnedFolders.Contains(Constants.UserEnvironmentPaths.RecycleBinPath) && AppLifecycleHelper.IsFirstRun)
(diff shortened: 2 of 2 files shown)


src/Files.App/Utils/Global/QuickAccessManager.cs
```diff
@@ -42,11 +42,10 @@ private void PinnedItemsWatcher_Changed(object sender, FileSystemEventArgs e)
 		public async Task InitializeAsync()
 		{
 			PinnedItemsModified += Model.LoadAsync;
+			await Model.LoadAsync();
```
yaira2, line 45: Is there a reason this was reorderd?

marcofranzen99: In the line below we check with Model.PinnedFolers.Contains.... if the Recycle Bin is already pinned. LoadAsync needs to be called first, otherwise the PinnedFolders list is empty and the check would not work.

yaira2: Makes sense.

src/Files.App/Services/Windows/WindowsQuickAccessService.cs
```diff
@@ -26,7 +26,12 @@ public async Task<IEnumerable<ShellFileItem>> GetPinnedFoldersAsync()
 		private async Task PinToSidebarAsync(string[] folderPaths, bool doUpdateQuickAccessWidget)
 		{
 			foreach (string folderPath in folderPaths)
-				await ContextMenu.InvokeVerb("pintohome", [folderPath]);
+			{
+				// make sure that the item has not yet been pinned
+				// the verb 'pintohome' is for both adding and removing
+				if (!IsItemPinned(folderPath))
```
yaira2, line 32: I suppose this doesn't hurt, but from what I understand, the changes to `QuickAccessManager` should be enough to resolve the issue.

marcofranzen99: That's correct.
I just thought it would be good to add this check here to prevent problems like this. 

Since the method is called 'PinToSidebarAsync', someone using the method would not assume that the item gets unpinned if it is already pinned.

yaira2 (approved):
LGTM

Issue #10001, opened 2022-09-15
Calculated OneDrive size is wrong


ferrariofilippo:
### Description

When displaying OneDrive's properties, space free/used/total refers to the disk where the local folder is stored, rather than the actual size of the folder

### Steps To Reproduce

1. Open OneDrive Properties

### Steps to fix

- Fix how OneDrive used space is calculated

### Files Version

Version: 2.3.28.0 OS Version: 10.0.22000.978

### Windows Version

Windows 11 Home 21H2 22000.856

### Relevant Assets/Logs

![OneDrive](https://user-images.githubusercontent.com/102259289/190513056-a6f417c4-7fa2-4f35-9455-884026b47864.png)

Josh-65:
Asking @yaichenbaum for approval.

yaichenbaum:
Approved

yaichenbaum:
@cinqmilleans do you want to look at this?

cinqmilleans:
Ok I analyze the bug.

cinqmilleans:
@yaichenbaum No, because the issue already exists in the preview version, before my changes. I was able to reproduce the problem, I will analyze.

cinqmilleans:
@yaichenbaum There is a logic to this behavior. OneDrive uses a folder placed by default in the C: drive.
The sizes displayed therefore correspond to the disk that hosts the folder synchronized with OneDrive. However, the expected, intuitive behavior is to display the remote sizes. This requires code to retrieve these values. It appears that this code does not currently exist for any cloud provider.

It would be interesting to add this information in the CloudProvider class. The tooltip and the popup can display 2 sections, one for local sizes and another for remote sizes. It would be a very practical added value.

yaichenbaum:
> No, because the issue already exists in the preview version, before my changes. I was able to reproduce the problem, I will analyze.

That's correct, this issue has been around for a while and isn't related to any of your recent changes.

> It would be interesting to add this information in the CloudProvider class. The tooltip and the popup can display 2 sections, one for local sizes and another for remote sizes. It would be a very practical added value.

I imagine the expected behavior for cloud drives is to only display the remote values and to fall back on the local values if that's the only information available.

hecksmosis:
So this should be fixed for now by just calculating the local folder size for OneDrive?

ferrariofilippo:
This would fix only `UsedSpace` not `TotalSpace`

hecksmosis:
Oh true

[...]

PR #16882, opened 2025-03-02
Report OneDrive size on 23H2.

dongle-the-gadget:
**Resolved / Related Issues**

#10001 (not using the "closes" word as I don't believe this is a complete fix)

Checked by opening the Properties window of OneDrive Cloud Drives and local drives. Both reports correct sizes.

### DIFF
--- a/src/Files.App/Utils/Storage/Helpers/DriveHelpers.cs
+++ b/src/Files.App/Utils/Storage/Helpers/DriveHelpers.cs
@@ -2,18 +2,25 @@
 // Licensed under the MIT License.
 
 using DiscUtils.Udf;
+using Files.App.Services.SizeProvider;
 using Microsoft.Management.Infrastructure;
+using Microsoft.Win32;
+using System.Runtime.InteropServices;
 using Windows.Devices.Enumeration;
 using Windows.Devices.Portable;
 using Windows.Storage;
 using Windows.Storage.FileProperties;
+using Windows.Storage.Provider;
 using Windows.Win32;
 using Windows.Win32.Foundation;
+using WinRT;
 
 namespace Files.App.Utils.Storage
 {
 	public static class DriveHelpers
 	{
+		private static readonly Guid IID_IStorageProviderStatusUISourceFactory = new Guid("12e46b74-4e5a-58d1-a62f-0376e8ee7dd8");
+
 		public static async void EjectDeviceAsync(string path)
 		{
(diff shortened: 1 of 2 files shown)


0x5bfa:
I did a quick research and this seems to be the right way of querying size of cloud storages. Surprisingly there's no legacy api or properties for this AFAIK.

@dongle-the-gadget do you think we can code like this (I mean using our ComPtr and interface instead)? I'm not good at WinRT binding and also this might need a raw binding like [this](https://github.com/terrafx/terrafx.interop.windows/blob/main/sources/Interop/Windows/WinRT/winrt/windows.storage.provider/IStorageProviderStatusUISourceFactory.cs) from CsWin32 which may not have it (I can't check rn) so I may be wrong.

```
ComPtr<IStorageProviderStatusUISourceFactory> pFactory = default;
PInvoke.CoCreateInstance(
    &factoryClsid,
    null,
    CLSCTX.CLSCTX_LOCAL_SERVER,
    &factoryIid,
    (void**)pFactory.GetAddressOf());

var source = pFactory.Get()->GetStatusUISource(syncRootInfo.Id);
var quotaUI = source.GetStatusUI().QuotaUI;
```

0x5bfa:
LGTM!

dongle-the-gadget:
> There seems to be an issue where the free space is higher than the capacity.

I wasn't able to replicate on my machine, can you send me over a screenshot (and what File Explorer reports if it's available)?

yaira2:
The issue is visible in your screenshot as well, capacity is shown as 1tb and the free space is listed as more than 1tb.

dongle-the-gadget:
Looked into it, and it's a separate formatting bug within Files. Files internally process sizes in base two but uses base ten suffixes.

0x5bfa:
The more correct suffix is TiB (1 TiB = 1024 GiB).

yaira2:
How do we want to move forward with this?

dongle-the-gadget:
I think that issue should be filed and resolved separately, since it's out of scope for this. This affects regular drives as well, not just this new implementation.

yaira2 (approved):

yaira2:
@dongle-the-gadget thank you!

yaira2:
There are a couple of things to follow up on, but they aren't blocking.
- Enable the storage overview in the details pane for OneDrive
- Investigate the incorrect display for capacity
- Support for older versions of Windows

PR #16908, opened 2025-03-08
Code Quality: Use ComPtr for all D3D11 pointers

0x5bfa:
### Resolved / Related Issues

General improvements for code that uses raw unmanaged pointers.

1. Build the Files.
2. Open a preview pane with a PDF or a DOCX, etc.

CC @hez2010


0x5bfa:
@hez2010 What exactly was the issue that caused the crash in Preview Pane? Looks like this PR also has a crash when disposing. Perhaps should I dispose these com objects manually?

hez2010:
> Looks like this PR has a crash when disposing CEOL again. Perhaps should I dispose these com objects manually?

Some resources shouldn't be released eagerly when the visual content is still visible.

yaira2 (approved):

src/Files.App/ViewModels/UserControls/Previews/ShellPreviewViewModel.cs
```diff
@@ -161,61 +161,61 @@ private unsafe bool ChildWindowToXaml(nint parent, UIElement presenter)
...
-
-			foreach (var driveType in driverTypes)
+			Windows.Win32.Foundation.HRESULT hr = default;
+			Guid IID_IDCompositionDevice = typeof(IDCompositionDevice).GUID;
+			using ComPtr<ID3D11Device> pD3D11Device = default;
+			using ComPtr<ID3D11DeviceContext> pD3D11DeviceContext = default;
+			using ComPtr<IDXGIDevice> pDXGIDevice = default;
+			using ComPtr<IDCompositionDevice> pDCompositionDevice = default;
+			using ComPtr<IUnknown> pControlSurface = default;
+			ComPtr<IDCompositionVisual> pChildVisual = default; // Don't dispose this one, it's used by the compositor
```
hez2010, line 171: It seems that we never release it and would cause memory leaking. I think we should do something around this to free it eventually when user switched or closed the preview pane.

hez2010 (approved):
LGTM!

0x5bfa:
@hez2010 I just tried to dispose pChildVisual but it seems to be already disposed by CEOL. What should we do, shoul we leave as was?

image

image

PR #16926, opened 2025-03-13
Code Quality: Replaced our own Shimmer control with WCT

0x5bfa:
### Resolved / Related Issues

_None_

1. Open Files app
2. Go to a folder that has numerous items
3. Change layouts and see Shimmer in the all llayouts works fine


src/Files.App/Views/Layouts/DetailsLayoutPage.xaml
```diff
@@ -931,8 +932,9 @@
 											</ContentPresenter>
 
 											<!--  Loading indicator  -->
-											<uc:Shimmer
+											<wctlabs:Shimmer
 												x:Name="TypeUnknownGlyph"
+												MinWidth="8"
```
yaira2, line 937: Does `MinHeight` need to be set as well?

0x5bfa: Yeah when I tried this change locally, Shimmer was clipped because default style defines the min width 40px instead of 8px which was set in our own Shimmer style.
We can set this globally, do you want me to do so?

yaira2: Please do

yaira2 (approved):
LGTM

zp7D (commented):
[t]()

PR #16929, opened 2025-03-14
Code Quality; Added global files to the solution file

0x5bfa:
### Resolved / Related Issues

_None_

_None_

image

### DIFF
--- a/Files.slnx
+++ b/Files.slnx
@@ -4,6 +4,10 @@
     <Platform Name="x64" />
     <Platform Name="x86" />
   </Configurations>
+  <Folder Name="/.solution/" Id="ddb4ef12-cc4e-5e88-e6fc-71a802f4de97">
+    <File Path="Directory.Build.props" />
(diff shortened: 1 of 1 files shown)


0x5bfa:
the goal here is not to match the folder structure but to keep developer experience the greatest possible. contrary, i dont encourage the team to "core" and "platform" folders since they unnecessarily make nested structure and hide the projects inside.

@yaira2 please review, @hez2010 based on your review, is this good for you?

yaira2:
This looks good to me. Fyi @hez2010

0x5bfa:
> LGTM. While don't like the name ".solution" btw. Maybe we can use some terms like "global" etc.?

The reason why I put the period `.` is to always put that folder at the very top. `global` is still bfore `src` and `tests` but we should keep in mind this. Let me change to it shortly.

hez2010 (approved):
LGTM

Issue #15283, opened 2024-05-01
Code Quality: Broader SidebarView control usage


0x5bfa:
### Description

We're planning to use Sidebar control for consistency and simplicity. I already confirmed this can reduce 200 lines.

@marcelwgn It be nice if you have idea around this:)

### Concerned code

- Around SidebarView class

### Gains

Consistency and simplicity

### Requirements

In this case, we should do:
- Separate view-unique values such as ExpandedWidthThreshold, IsPaneOpen and ViewModel. Those examples are just examples, we probably need to separate almost all dependency properties in SidebarView.
- Separate view-unique events such as ItemInvokedAsync.
- Add generation method for each Sidebar type: Default, Properties, Settings.
- Create FrameworkElement property to hold icon element for both BitmapImage and OpacityIcon

### Comments

My prototype:
https://github.com/files-community/Files/compare/main...0x5bfa:Files:5bfa/CQ-RefactorSidebarViewModel

0x5bfa:
@yaira2 Can I work on refactoring SidebarViewModel? It has been unchanged from UWP.

0x5bfa:
https://discord.com/channels/725513575971684472/958783217941487616/1236699756265214052

At the end of quick discussion, we settled on the idea to generalize usage of SidebarView, moving view-specific properties, such as item collection, pane mode, pane width. This also enables us to bind values to user settings to make them keep up with the latest info all the time.
The new usage will be almost the same as the one of NavigationView, such as invoking an event to notify item invocation, using two-way binding for selected item.

In this issue, I guess we also might as well discuss the idea to make items larger to the height probably based on a user setting - EnableCompactModeInSidebar.

PR #16952, opened 2025-03-18
Code Quality: Converted SidebarView to a custom control

0x5bfa:
### Resolved / Related Issues

This is the first PR to improve the codebase of Sidebar controls in order to use this in the Settings pages and the Properties window.

- #15283
  - Converted SidebarView to a custom control
  - Refactored accompanyying classes, such as sidebar automation peer classes

The last thing we have to do for SidebarView is to remove ViewModel DP from that and create some events instead.

I have tested and confirmed working well.

1. Open Files
2. Test tab stop: Move tab stop around the sidebar and see no issues/exceptions come up
3. Test the behavior:
    1. Double tap the sizer
    2. Move the sizer
    3. Shrink the height of the window to see if the ScrollViewer position and behavior is correct
    4. Click the Settings button placed at the footer.
4. See other things

CC: @marcelwgn


marcelwgn:
PR looks mostly fine. A few notes:
1. Why are we switching to a templated control?
2. Naming and structuring wise, not sure I agree with moving events into a separate file and prefixing all template parts with "PART_".
3. If we rightfully rename the internal event handlers, we should match the control exactly (_paneSizer vs PaneResizer_...)

0x5bfa:
1. That is because we are planning to use this control in other places like Settings NavView and Props window NavView. Originally SidebarView as a UC had a background brush that canno be overridden from outside, now it can be.
2. Tbh, I don't what the PART_ means lol. I've seen in the WinUI repo and controls used in codebehind seems to be named like that. The reason why I separated event methods are simply for the sake of simplicitiy, I can merge it into SidebarView.cs
3. Of course, sounds like I still missed some of things, let me find and fix.

0x5bfa:
1. That is because we are planning to use this control in other places like Settings NavView and Props window NavView. For example, originally SidebarView as a UC had a background brush that canno be overridden from outside, now it can be.
2. Tbh, I don't know what the PART_ means lol. I've seen in the WinUI repo and controls used in codebehind seem to be named like that. The reason why I separated event methods is simply for the sake of simplicitiy, I can merge it into SidebarView.cs
3. Of course, sounds like I have missed some of things, let me find and fix.

marcelwgn:
Regarding 1, does lightweight styling not work for UserControls? I hope we dont aim to retemplate the entire sidebar for settings...

For the second, yes some WinUI controls use PART_ (which stands for template part, i.e. parts that should be included in the control) but working on the WinUI 2 repository, I can remember only one or two controls that did that. From what I remember most controls did not use that schema, not using that prefix was way more common.

mdtauk:
``PART_`` Was a convention for controls back in the early Xaml controls, silverlight, and wpf.

Its for parts that the code relies on, and so when the control is re-templated it acts as a warning to include it in the template.  It would give a warning in the designer if it was not present.

0x5bfa:
> Regarding 1, does lightweight styling not work for UserControls? I hope we dont aim to retemplate the entire sidebar for settings...

We will not do that absolutely, I'm trying to make it flexible for future use, largely taling about overring resources not the whole style, I don't think we use the same value as to the sidebar max threshold width and pane sizer position margin, etc:
image

> For the second, yes some WinUI controls use PART_ (which stands for template part, i.e. parts that should be included in the control) but working on the WinUI 2 repository, I can remember only one or two controls that did that. From what I remember most controls did not use that schema, not using that prefix was way more common.

> PART_ Was a convention for controls back in the early Xaml controls, silverlight, and wpf.
> Its for parts that the code relies on, and so when the control is re-templated it acts as a warning to include it in the template. It would give a warning in the designer if it was not present.

I see them across the WinUI 3 repo and even the newest controls (e.g. SelectorBar and TitleBar) have them. I'm convinced the reason why they add and I feel like we should use them.

0x5bfa:
Theres an anooying issue that persists where i didnt touch. Let me close this and re-convert again...

PR #16960, opened 2025-03-20
Fix: Fixed an issue where drive capacity was reported incorrectly

### LINKED ISSUE #16919: Bug: Files uses incorrect size unit mappings
### Description

Related to #10075, opening as a separate issue as this is a bug and not a feature.

Previously found in #16882, Files uses incorrect size unit mappings, as it internally calculates sizes in base-two (so the appropriate units are kibibytes, mebibytes, etc.) but it shows to the user base-ten units (kilobytes, megabytes) instead.

This causes a disparity in the screenshot of #16882 where Files UI determined that 1007 GB is larger than 1 TB, so it showed the size in GB rather than TB.

Relevant code:
https://github.com/files-community/Files/blob/283345a335777b9ef7051e722747694f8c161b90/src/Files.App/Data/Models/ByteSize.cs#L10

### Steps To Reproduce

See screenshot in #16882.

### Files Version

Found on main branch, might have affected released versions

### Windows Version

Not relevant

### User ID

_No response_

### Log File

Not relevant

dongle-the-gadget:
**Resolved / Related Issues**

Closes #16919

Tried opening the Properties window, the size is correctly reported.

Files properties window, reporting correct sizes

### DIFF
--- a/src/Files.App/Data/Models/ByteSize.cs
+++ b/src/Files.App/Data/Models/ByteSize.cs
@@ -9,13 +9,13 @@ namespace Files.App.Data.Models
 	{
 		private static readonly FrozenDictionary<string, string> units = new Dictionary<string, string>
 		{
-			[ByteSizeLib.ByteSize.BitSymbol] = "ByteSymbol".ToLocalized(),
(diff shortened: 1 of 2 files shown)


yaira2 (approved):
LGTM

Thanks for updating the localization strings to the constants!

PR #16984, opened 2025-03-25
Fix: Fixed issue where the extract button was sometimes not displayed in the toolbar

### LINKED ISSUE #15457: Bug: When a ZIP file is opened via file association, the Extract Button is not displayed on the toolbar
### Description

I have Files set as my default handler for Zip files.  When I download a zip file in MS Edge and then open that zip file from Edge, Files is opened as expected, but the Extract button is not visible.

image

If Files loses focus and regains focus, the extract button will show up on the refresh.

image

### Steps To Reproduce

1. Set Files as the default handler for ZIP files
2. Download a ZIP File in MS Edge
3. Open the downloaded ZIP file in MS Edge
4. Once Files opens, notice that there is no Extract button
5. Switch focus to a different application in Windows.
6. Switch focus back to Files, notice that the Extract button will appear.

### Requirements

At step 4, the Extract button should be visible.

### Files Version

3.4.0.0

### Windows Version

10.0.22621.3593

### Log File

[debug.log](https://github.com/files-community/Files/files/15418409/debug.log)

marcofranzen99:
**Resolved / Related Issues**

- Closes #15457

1. Set Files Dev as default application for .zip files
2. Opened .zip file via files explorer
3. .zip files gets opened in Files and the extract button is visible

I have also checked that the buttons are still displayed in the other cases

https://github.com/user-attachments/assets/41996f06-88af-43a4-a0a4-2ae747dd1900

### DIFF
--- a/src/Files.App/UserControls/Toolbar.xaml
+++ b/src/Files.App/UserControls/Toolbar.xaml
@@ -36,6 +36,7 @@
 				x:Key="NegatedBoolToVisibilityConverter"
 				FalseValue="Visible"
 				TrueValue="Collapsed" />
+			<wctconverters:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter" />
 
 			<x:Boolean x:Key="True">True</x:Boolean>
 			<x:Boolean x:Key="False">False</x:Boolean>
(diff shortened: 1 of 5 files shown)


src/Files.App/UserControls/Toolbar.xaml
marcofranzen99, line 1: I could not find a case where any of the MenuFlyoutItems was visible and disabled, so i removed the `IsEnabled` from all of them. The visibility is now directly controlled by the decompress actions and the properties that are no longer necessary have been removed from the ViewModel

yaira2:
Can you please resolve the conflicts?

marcofranzen99:
Sometimes one of the MenuFlyoutItems is visible but disabled 🤔 I don't know why yet, will try to debug it if I have time in the next few days
image

marcofranzen99:
> Sometimes one of the MenuFlyoutItems is visible but disabled 🤔 I don't know why yet, will try to debug it if I have time in the next few days

fixed

yaira2 (approved):
LGTM

PR #16996, opened 2025-03-30
Feature: Added support for changing the size format

### LINKED ISSUE #10075: Feature: Ability to customize the way file sizes are displayed (units and decimal/binary bytes)

Being able to configure Files so that the unit for the file sizes is accurate to the way Windows measures file sizes _(eg. GiB instead of GB)_, or even to customize the way file sizes are displayed entirely: letting users choose between Decimal bytes _(1kB = 1000B, universal standard)_ and Binary bytes _(1kiB = 1024B, which windows has been using forever, despite displaying the units for Decimal bytes instead)_.

### Requirements

_I'm not exactly sure how to use this field, but here goes_
- New options in the "Preferences" settings for changing the way file sizes are displayed: either "Use windows default" _(combination of binary bytes and decimal units, which should likely be the default option, as to not confuse new users)_; "Decimal"; and "Binary"
- Changing this setting would then reflect on the drive sizes in the "Drives" view, file sizes anywhere when browsing, as well as when opening "Properties", and folder sizes if enabled.

### Files Version

Version: 2.3.12.0 OS Version: 10.0.22622.601

### Windows Version

Edition	Windows 11 Pro Version	22H2 Installed on	‎07/‎08/‎2022 OS build	22622.601 Experience	Windows Feature Experience Pack 1000.22634.1000.0

### Comments

I don't know how possible this actually is to implement. I've seen programs use the correct units (KiB instead of KB) when on Windows, but I don't remember seeing any that uses decimal bytes, or gives users the option to do so.

_I'm having a lot of fun using "Files" so far (about a week), and while it's generally a bit slower than the default Windows 11 file explorer when fetching large files/folder lists (although the Windows Explorer tends to freeze and crash quite a bit in the insider Betas), it's made up for it in appearance, layout, and features. This app makes so much sense, and I wouldn't hesitate to recommend it to anyone. Thank you for creating and maintaining it!_

yaira2:
**Resolved / Related Issues**

- Closes #10075

1. Confirmed the size unit reflects the selected option in settings

### DIFF
--- a/src/Files.App/Data/Contracts/IFoldersSettingsService.cs
+++ b/src/Files.App/Data/Contracts/IFoldersSettingsService.cs
@@ -84,5 +84,10 @@ namespace Files.App.Data.Contracts
 		/// Gets or sets a value indicating whether or not to show checkboxes when selecting items.
 		/// </summary>
 		bool ShowCheckboxesWhenSelectingItems { get; set; }
+
+		/// <summary>
+		/// Gets or sets a value indicating which format to use when displaying item sizes.
+		/// </summary>
+		SizeUnitTypes SizeUnitFormat { get; set; }
 	}
 }
--- a/src/Files.App/Data/Enums/SizeUnitTypes.cs
+++ b/src/Files.App/Data/Enums/SizeUnitTypes.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) Files Community
+// Licensed under the MIT License.
+
+namespace Files.App.Data.Enums
+{
+	public enum SizeUnitTypes
+	{
+		/// <summary>
+		/// Displays sizes using Binary values.
+		/// </summary>
+		BinaryUnits,
+
+		/// <summary>
+		/// Displays sizes using Decimal values.
+		/// </summary>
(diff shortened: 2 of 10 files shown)


0x5bfa (approved):
Tested locally, confirmed working well.
Settings title could be more descriptive maybe?

0x5bfa:
> On my machine, the toggle doesn't immediately refresh the view, causing the page to show units based on the old setting until the user navigates to a new page or manually refresh.

I clicked the Refresh button to test, wouldnt it b challenging to reflect the change automatically? @yaira2

yaira2:
I'll look into this

yaira2:
I pushed a commit to auto refresh the file area. I'll look into the drives widget a little later.

yaira2:
> On my machine, the toggle doesn't immediately refresh the view, causing the page to show units based on the old setting until the user navigates to a new page or manually refresh.

This should be addressed with my latest commit.

yaira2:
> Tested locally, confirmed working well. Settings title could be more descriptive maybe?

We can technically display an example, but let's see how this works first.

PR #17015, opened 2025-04-03
Fix: Fixed errors that occurred with shortcut items in git repositories

### LINKED ISSUE #17008: Bug: Selecting a shortcut in a git folder crashes Files
### Description

Trying to click/select a link file in any way will cause the entire program to crash and restart.
This is the command in link file. Used to convert pictures in the same directory to webp through ffmpeg
```
C:\Windows\System32\cmd.exe /C for %i in (*.jpg *.png *.jpeg) do ffmpeg -i "%i" "%~ni.webp"
```

### Steps To Reproduce

1. Create a new shortcut and fill in the above command
2. Any attempt to select this file in files will result in a crash.(no need running)

### Files Version

3.9.1.0

### Windows Version

10.0.22631.5039

### User ID

9a9d7334-3ac0-4870-a4ae-9503d5f3c117

### Log File

[debug.log](https://github.com/user-attachments/files/19561738/debug.log)

marcofranzen99:
**Resolved / Related Issues**

- Closes #17008 

**Description**
The problem was that GitShortcutItem cannot be cast to ShortcutItem because GitShortcutItem only inherits from GitItem and IShortcutItem. It cannot additionally inherit from ShortcutItem, since a class in C# cannot have multiple base classes.

Simply casting to IShortcutItem everywhere did not work, as the IShortcutItem interface would lack properties of the ListedItem (such as ItemPath).

That's why I decided to create an interface IListedItem. IShortcutItem & IGitItem now both inherit from this interface.

Problems only occurred with shortcut items that are part of a git repository. Not all of them led to a crash, some only displayed error messages or led to incorrect behavior. Here is a list of affected actions:
- Selecting a shortcut item
- Selecting a shortcut item with preview pane open
- Open Properties Window for a shortcut item
- Save modified properties of a shortcut via properties window
- Open in new tab
- Open in new pane
- Open in new Window
- Pin To Start
- Unpin From Start
- Share Item

1. tested if the actions listed above work correctly

### DIFF
--- a/src/Files.App/Data/Items/IListedItem.cs
+++ b/src/Files.App/Data/Items/IListedItem.cs
@@ -1,0 +1,79 @@
+// Copyright (c) Files Community
+// Licensed under the MIT License.
+
+using Files.App.ViewModels.Properties;
+using Microsoft.UI.Xaml.Media.Imaging;
+using Windows.Storage;
+
+namespace Files.App.Utils
+{
+	public interface IListedItem
+	{
+		GitItem AsGitItem { get; }
+		RecycleBinItem AsRecycleBinItem { get; }
+		bool ContainsFilesOrFolders { get; set; }
+		string ContextualProperty { get; set; }
+		BitmapImage CustomIcon { get; set; }
+		Uri CustomIconSource { get; set; }
+		ObservableCollection<FileProperty> FileDetails { get; set; }
+		string FileExtension { get; set; }
+		ulong? FileFRN { get; set; }
+		BitmapImage FileImage { get; set; }
+		string FileSize { get; set; }
+		long FileSizeBytes { get; set; }
+		string FileSizeDisplay { get; }
+		string[] FileTags { get; set; }
+		IList<TagViewModel>? FileTagsUI { get; }
+		string FileVersion { get; set; }
+		string FolderRelativeId { get; set; }
+		bool HasTags { get; set; }
+		BitmapImage IconOverlay { get; set; }
+		string ImageDimensions { get; set; }
+		bool IsAlternateStream { get; }
+		bool IsArchive { get; }
+		bool IsDriveRoot { get; }
+		bool IsElevationRequired { get; set; }
+		bool IsExecutable { get; }
+		bool IsFolder { get; }
+		bool IsFtpItem { get; }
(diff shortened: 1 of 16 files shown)


src/Files.App/Data/Items/IListedItem.cs
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) Files Community
+// Licensed under the MIT License.
+
+using Files.App.ViewModels.Properties;
+using Microsoft.UI.Xaml.Media.Imaging;
+using Windows.Storage;
+
+namespace Files.App.Utils
```
marcofranzen99, line 8: This interface was automatically generated by VS.
Should all of the properties be in here? 
It might also be better to move the interfaces and classes from `ListedItem.cs` to separate files.

yaira2: Fyi @0x5bfa

0x5bfa: > Should all of the properties be in here?

All props inside the ListView class should be in here.

> It might also be better to move the interfaces and classes from ListedItem.cs to separate files.

You mean like `IShortcutItem`? Personally I dont but if you like, feel free.

marcofranzen99: > You mean like IShortcutItem? Personally I dont but if you like, feel free.

yes exactly. if you wouldn't do it, we can also leave it as it is

src/Files.App/ViewModels/Properties/Items/FolderProperties.cs
```diff
@@ -48,9 +48,8 @@ public override void GetBaseProperties()
 				ViewModel.LoadFileIcon = Item.LoadFileIcon;
 				ViewModel.ContainsFilesOrFolders = Item.ContainsFilesOrFolders;
 
-				if (Item.IsShortcut)
+				if (Item.IsShortcut && Item is IShortcutItem shortcutItem)
```
0x5bfa, line 51: I think this is the important part. FYI for others.

0x5bfa (approved):
I can repro that thehe latest stable version of Files crashes when selecting but in this PR it works well.

LGTM.

image

Issue #14937, opened 2024-03-11
Bug: File names for Japanese zips aren't shown correctly on Chinese language


oxygen-dioxide:
### What feature or improvement do you think would benefit Files?

My windows is in simplified Chinese locale. Currently, when opening zip files produced by Japanese users, file names will be garbled. 
![image](https://github.com/files-community/Files/assets/54425948/785482a7-69b8-4b2e-ac32-25a02b3cc0e7)

### Requirements

When opening zip files, the user can choose which encoding to use for reading file names.

### Files Version

3.2.2.0

### Windows Version

windows 11 23H2 build 26040

0x5bfa:
Thanks for the feedback!
I'm on japanese env. I'll repro later

Josh65-2201:
Thanks for the report, I've added this to the project board.

oxygen-dioxide:
Further idea: let the user choose which encoding to use when creating a .zip file, and use UTF-8 by default so that it won't become garbled when opened on another locale

yaira2:
What happens when you browse the archives in File Explorer?

oxygen-dioxide:
Do you mean the explorer.exe shipped with windows system? 
![image](https://github.com/files-community/Files/assets/54425948/b6af4ddf-f8f6-4a33-8869-95609e0af1a3)
They are shown as incorrect Chinese characters, but they should be Japanese hiraganas.

0x5bfa:
Looking from what FE shows it decodes as ShiftJIS while it encoded as UTF8

0x5bfa:
@gave92 fyi, do you think this is from seven zip sharp or 7zip?

yaira2:
> @gave92 fyi, do you think this is from seven zip sharp or 7zip?

That was my thought as well

yair100:
@oxygen-dioxide is this still an issue?

oxygen-dioxide:
It is still an issue, because we still can't view a zip file in user-specified encoding. We can only extract it.

yair100:
Thank you for the response, do you have any suggestions for how to resolve this?

oxygen-dioxide:
> Thank you for the response, do you have any suggestions for how to resolve this?

[...]

PR #17022, opened 2025-04-06
Feature: Added support for selecting encoding when extracting ZIP files

oxygen-dioxide:
ZIP file specifications historically use the system's local encoding (e.g., CP936 for Chinese, Shift_JIS for Japanese) to store filenames, rather than UTF-8. This causes cross-language compatibility issues: for example, filenames created by Japanese users (Shift_JIS) will appear as garbled text when extracted by Chinese users (using GBK/CP936), and vice versa. English-speaking users rarely encounter this issue since most legacy encodings are Latin-compatible, but it remains critical for multilingual users.

This PR adds an encoding selection option during ZIP extraction. Users can manually override the default system encoding (e.g., to Shift_JIS, UTF-8, GBK, Big5, etc.) to ensure filenames are decoded correctly. (defaults to system encoding to reserve backward compatibility)

**Resolved / Related Issues**

- #14937 is partially solved in this PR. (See the "TODO" part below)

I'm a Chinese user, so I don't know whether and how the encoding issue affects English users. Below is how I test this on Simplified Chinese windows.

1. Download this zip file: 
[japanese-example.zip](https://github.com/user-attachments/files/19619832/japanese-example.zip)
2. In Files, right click this file, Extract -> Extract Files
3. Select "Japanese (Shift-JIS)" encoding
image

The file names in this zip file are Japanese encoded in Shift-JIS. Here are the results of extraction:
Extracted with Shift-JIS (correct)
image

Extracted with default encoding (previous behaviour which is incorrect):
image

**TODO**
- Autometically guess encoding
- Being able to select encoding when viewing a zip file


Directory.Packages.props
```diff
@@ -32,6 +32,7 @@
     <PackageVersion Include="OwlCore.Storage" Version="0.12.2" />
     <PackageVersion Include="Sentry" Version="5.1.1" />
     <PackageVersion Include="SevenZipSharp" Version="1.0.2" />
+    <PackageVersion Include="SharpZipLib" Version="1.4.2" />
```
yaira2, line 35: First of all, great work on this PR! Before reviewing, have you looked into whether this can be using `SevenZipSharp`?

oxygen-dioxide: I've tried using SevenZipSharp first, but it doesn't provide encoding option. We can only use system default encoding when extracting zip files with SevenZipSharp.

yaira2: In that case, once we're implementing a new library, can it fully replace `SevenZipSharp`?

oxygen-dioxide: I'm new to this project. Do you mean that files community will implement a brand-new archive library to replace SevenZipSharp?

yaira2: Assuming `SharpZipLib` can do everything we need, I meant if we should remove the `SevenZipSharp` dependency and avoid extra dependencies.

oxygen-dioxide: If `SharpZipLib` can cover all the functions of `SevenZipSharp`, we can migrate to `SharpZipLib` of course. But we have to look into `SharpZipLib` to see if there are some functions of `SevenZipSharp` that Files uses and can't be replaced by `SharpZipLib`
Also `SharpZipLib` is written in c#, so the performance may be worse than `SevenZipSharp` which is an wrapper of 7zip's c++ code

yaira2: Those are all good points. I'm not the biggest fan of overlapping dependencies, but it might have to do for now.

oxygen-dioxide: Oh, `SharpZipLib` doesn't support RAR or 7z, so we still need `SevenZipSharp` for these file types.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2096,6 +2096,12 @@
   <data name="ArchivePassword" xml:space="preserve">
     <value>Archive password</value>
   </data>
+  <data name="ArchiveEncoding" xml:space="preserve">
+    <value>Archive encoding</value>
```
yaira2, line 2100:
```suggestion
    <value>Encoding</value>
```

src/Files.App/Data/Items/EncodingItem.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EncodingItem"/> class.
+        /// </summary>
+        /// <param name="code">The code of the language.</param>
+        public EncodingItem(string code)
+        {
+            if (string.IsNullOrEmpty(code))
+            {
+                Encoding = null;
+                Name = "ArchiveEncodingSystemDefault".GetLocalizedResource();
```
yaira2, line 30:
```suggestion
                Name = Strings.Default.GetLocalizedResource();
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -2096,6 +2096,12 @@
...
+  </data>
+  <data name="ArchiveEncodingSystemDefault" xml:space="preserve">
+    <value>Windows default</value>
+  </data>
```
yaira2, line 2104:
```suggestion
```

src/Files.App/Dialogs/DecompressArchiveDialog.xaml
```diff
@@ -80,6 +82,34 @@
...
+				Orientation="Vertical"
+				Spacing="8">
+
+				<TextBlock
+					x:Name="EncodingHeader"
+					HorizontalAlignment="Stretch"
+					VerticalAlignment="Stretch"
+					Text="{helpers:ResourceString Name=ArchiveEncoding}" />
+
+				<uc:ComboBoxEx
```
yaira2, line 98:
```suggestion
				<ComboBox
```
ComboBoxEx is mainly used for settings pages.

src/Files.App/Dialogs/DecompressArchiveDialog.xaml
```diff
@@ -80,6 +82,34 @@
...
+					VerticalAlignment="Stretch"
+					Text="{helpers:ResourceString Name=Encoding}" />
+
+				<ComboBox
+					x:Name="EncodingBox"
+					HorizontalAlignment="Stretch"
+					VerticalAlignment="Stretch"
+					ItemsSource="{x:Bind ViewModel.EncodingOptions, Mode=OneWay}"
+					SelectedItem="{x:Bind ViewModel.SelectedEncoding, Mode=TwoWay}">
+					<uc:ComboBoxEx.ItemTemplate>
```
yaira2, line 104: This needs to be updated to `ComboBox` as well.

yaira2 (approved):
LGTM

PR #17025, opened 2025-04-06
Code Quality: Introduced a new build experience

### LINKED ISSUE #16680: Code Quality: Add a solution configuration for each build of the app
### Description

Here's an interesting idea: what if instead of using the `Configure-AppxManifest.ps1` script, each "branch" or release channel of the app (Release, Preview, Dev) were literally a solution configuration? For example, instead of choosing `Debug|x64`, you'd choose `Dev|x64`. Resources such as the app icon and name would dynamically change rather than needing to be updated using a script.

### Concerned code

- The solution file
- Areas where there are differences per branch
  - Different features
  - Different app icons

### Gains

- eg. A better readability.
- eg. Uncoupling concepts X and Y.
- eg. Clarifying the responsibility of class C.

### Requirements

- Add `#if` compiler blocks to hide features that are restricted to a channel
- Using MSBuild property files, the associated app icons could also be dynamically changed

### Comments

This would be a good thing to do when a new solution is created.

Lamparter:
**Resolved / Related Issues**

- Closes #16680

1. Build Files
2. ...and if it works, this PR works

---

This PR introduces a new build experience for Files, where you can select the app environment using Visual Studio's built-in interface (this use case is the reason Configuration settings exist in Visual Studio)

image

I've created MSBuild targets that generate an AppX manifest file on the spot, and add correct assets appropriate to the build.

Ultimately, this PR simplifies the build experience by:
- Removing the CD scripts `Configure-*.ps1` (mostly)
- Generating the AppX manifest file on the spot, so no clunky designer or copied files
- Removing unnecessary configurations (aka #16569)
- Copying app assets at compile time
  - (no need to replace them manually, and all references to the app tiles are now `src/Files.App/Assets/AppTiles/Current/**` which contains the app tiles relevant to the build type.
- Hiding unnecessary generated files through the ignore, such as AppX manifest files and automatically generated app tile assets.


Lamparter:
I'm aware that the issue #16680 wasn't marked as ready to build, but the exact benefits and differences introduced by this are hard to outline in the issue.
It is easier to just implement and then the reviewer can try it out for themselves.

Lamparter:
This level of integration is tricky to describe, as the actual benefits of this are unrealistic to measure outside of the actual implementation.
Perhaps this can serve as a "demo" for how it could be implemented in other apps.

PR #17029, opened 2025-04-08
Fix: Fixed an issue where name sorting was case sensitive

### LINKED ISSUE #16912: Bug: Name sorting is case sensitive
### Description

Hi,
not sure if this is a bug or a feature ;p
Looks like sorting by name is case sensitive : 

Image

Replacing the "s" with "S" everywhere : 

Image

Can that be turned off if it's a feature ?

Thanks :)

### Steps To Reproduce

Name things like shown above

### Files Version

3.9.1.0

### Windows Version

22631.4890

### User ID

_No response_

### Log File

[debug.log](https://github.com/user-attachments/files/19152605/debug.log)

yaira2:
**Resolved / Related Issues**

- Closes #16912

1. Confirmed sorting is no longer case sensitive

### DIFF
--- a/src/Files.App/Helpers/NaturalStringComparer.cs
+++ b/src/Files.App/Helpers/NaturalStringComparer.cs
@@ -8,7 +8,7 @@ namespace Files.App.Helpers
 	{
 		public static IComparer<object> GetForProcessor()
 		{
-			return new NaturalComparer(StringComparison.CurrentCulture);
+			return new NaturalComparer(StringComparison.CurrentCultureIgnoreCase);
 		}
 
 		/// <summary>


src/Files.App/Helpers/NaturalStringComparer.cs
```diff
@@ -8,7 +8,7 @@ public sealed class NaturalStringComparer
 	{
 		public static IComparer<object> GetForProcessor()
 		{
-			return new NaturalComparer(StringComparison.CurrentCulture);
+			return new NaturalComparer(StringComparison.CurrentCultureIgnoreCase);
```
0x5bfa, line 11: Btw this is instantiated every time `SortingHelper.OrderFileList` is called. We prob should create a static field for this (if null, for the first time, create an instance).

0x5bfa (approved):

src/Files.App/Helpers/NaturalStringComparer.cs
yaira2: Can you please start an issue so we can track that?

PR #17034, opened 2025-04-09
Code Quality: Switch Storage Provider Status UI APIs to use MemberFunction

dongle-the-gadget:
**Resolved / Related Issues**

Switch the Storage Provider Status UI APIs from the `Stdcall` calling convention to `MemberFunction`, for correctness.

Tested the Properties window of OneDrive, works as expected.

### DIFF
--- a/src/Files.App.CsWin32/IStorageProviderQuotaUI.cs
+++ b/src/Files.App.CsWin32/IStorageProviderQuotaUI.cs
@@ -16,13 +16,13 @@ namespace Windows.Win32.System.WinRT
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		public HRESULT GetQuotaTotalInBytes(ulong* value)
 		{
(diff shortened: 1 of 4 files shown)


0x5bfa:
Since CsWin32 uses StdCall for COM functions, I thought okay with this convention. It uses a wrong conversion?

dongle-the-gadget:
`MemberFunction` is more strict, and for our purposes here it differentiates between C# `int` and C++ `HRESULT` (which is actually a type alias for `int`) versus CsWin32 `HRESULT` type, which is a wrapper around `int`.
The fact that `Stdcall` lets us use CsWin32 `HRESULT` instead of `int` is kind of by accident, and `MemberFunction` disallows that.

0x5bfa:
Is that because the layout the HRESULT from CsWin32 matches the one of int (2 bytes)?

dongle-the-gadget:
Quote from Jeremy (.NET):
> MemberFunction “doesn’t work” with an HResult struct as that’s technically wrong. That’s the whole reason we needed to add it (as an HResult struct works with Stdcall)

As for why it worked, it has to do with registers (that I’m not qualified to talk about)

0x5bfa (approved):
Thanks.

PR #17075, opened 2025-04-27
Code Quality: Ignored `Files.App.Launcher.exe.sha256`

Lamparter:
Correctly ignores the `Files.App.Launcher.exe.sha256` file that was not correctly ignored previously.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 Files.App.OpenDialog*.dll
 Files.App.SaveDialog*.dll
 Files.App.Launcher.exe
+Files.App.Launcher.exe.sha256
 
 # User-specific files
 *.rsuser
@@ -408,4 +409,3 @@ FodyWeavers.xsd
 # JetBrains Rider
 *.sln.iml
 .idea/
-src/Files.App/Assets/FilesOpenDialog/Files.App.Launcher.exe.sha256


src/Files.App/Assets/FilesOpenDialog/Files.App.Launcher.exe.sha256
yaira2, line 1: This shouldn't be deleted.

yaira2: From the repo, or it won't be included when the repo is cloned?

0x5bfa: TBH theres no reason to keep this file in the repo or on one's local because everytime you run CD, it'll be overridden and either:
- The CD should embed the SHA to the files.exe binary.
- The CD should output the SHA and you should copy&paste to the corresponding release blog.

CC: @hishitetsu I believe you added because we or users better can check the SHA of the binary right?

yaira2: > because everytime you run CD

Originally the file was only updated when the launcher project was modified, but we lost this behavior with one of our other changes. I'm hoping we can fix the build step at some point, but it's pretty low priority issue since it doesn't have any impact.

0x5bfa: The important thing is that we should keep the SHA so users can compare on their local but in the first place we have never done so (if Hishitetsu meant so).

Lamparter: > It'll be removed from the repo, or it just won't be included when the repo is cloned?

If it's in the ignore but still in the repository Git will let you track changes to it.

hishitetsu: The hash value is provided to see if the locally deployed FilesLauncher.exe needs to be updated.
If the hash is not correct, FilesLauncher.exe will not be properly updated.
https://github.com/files-community/Files/pull/11750

yair100:
@Lamparter this doesn't seem to be working, the file is not being ignored.

PR #17100, opened 2025-05-05
Code Quality: Removed extra code for async ops in Program class

0x5bfa:
### Resolved / Related Issues

- #15000

1. Open Files app
2. Enable the setting to run Files in the background
3. Start debebugging Files
4. Close the window
5. Launch Files from Start Menu

### DIFF
--- a/src/Files.App/Program.cs
+++ b/src/Files.App/Program.cs
@@ -246,10 +246,7 @@ namespace Files.App
 		/// </remarks>
 		public static void RedirectActivationTo(AppInstance keyInstance, AppActivationArguments args)
 		{
-			STATask.RunAsSync(() =>
-			{
-				keyInstance.RedirectActivationToAsync(args).AsTask().Wait();
-			});
+			keyInstance.RedirectActivationToAsync(args).AsTask().Wait();
 		}
 
 		public static void OpenShellCommandInExplorer(string shellCommand, int pid)
@@ -259,10 +256,7 @@ namespace Files.App
 
 		public static void OpenFileFromTile(string filePath)
 		{
-			STATask.RunAsSync(() =>
-			{
-				LaunchHelper.LaunchAppAsync(filePath, null, null).Wait();
-			});
+			LaunchHelper.LaunchAppAsync(filePath, null, null).Wait();
 		}
 	}
 }
--- a/src/Files.App/Program.cs
+++ b/src/Files.App/Program.cs
@@ -9,7 +9,6 @@ using System.IO;
 using System.Text;
 using Windows.ApplicationModel.Activation;
 using Windows.Storage;
-using static Files.App.Helpers.Win32PInvoke;
 
 namespace Files.App
 {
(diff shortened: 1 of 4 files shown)


0x5bfa:
Requesting your review @dongle-the-gadget @hez2010 I also would like to ask you review in:

Before:

After:

src/Files.App.Storage/Storables/WindowsStorage/STATask.cs
```diff
@@ -143,5 +146,46 @@ public static Task Run(Func<Task> func)
...
+				{
+					tcs.SetException(ex);
+				}
+				finally
+				{
+					PInvoke.SetEvent(hEventHandle);
+				}
+			});
+
+			HANDLE* pEventHandles = stackalloc HANDLE[1];
```
dongle-the-gadget, line 175: Do you even need to allocate a stack here? I thought your HANDLE variable is in stack already.

src/Files.App/Program.cs
```diff
@@ -273,20 +259,10 @@ public static void OpenShellCommandInExplorer(string shellCommand, int pid)
 
 		public static void OpenFileFromTile(string filePath)
 		{
-			IntPtr eventHandle = CreateEvent(IntPtr.Zero, true, false, null);
-
-			Task.Run(() =>
+			STATask.RunAsSync(() =>
```
dongle-the-gadget, line 262: Is there a reason why async void isn't desirable here?

0x5bfa: So you mean adding Wait() in that helper method instead?

src/Files.App.Storage/Storables/WindowsStorage/STATask.cs
0x5bfa: Related to your 2nd comment, it looks like you can't take address of a local variable that goes thru an async lambda/function. The current code doesn't have async and initially did so this can be refactored but IIUC if we wanna use async void we may have to do this.

0x5bfa:
This is ready, I, for now, keep the scope of this PR just to wrap the unsafe code into a safe method.

src/Files.App/Program.cs
0x5bfa: Id like to keep this PR's scope just wrapping the unsafe code into a separate method.

dongle-the-gadget:
I think it's better to do something like
```cs
public Task Something(Task task)
{
    async Task Call()
    {
        await task;
        SetEvent();
    }
}
```

yair100:
@0x5bfa fyi

0x5bfa:
@ahmed605 What is your opinion too? I can't be sure the current code and the above code is really non blocking.

yair100:
What's the state of this PR?

0x5bfa:
Committed the requested change. Itested the change in `RedirectActivationTo` on my local but didn't test  the one in `OpenFileFromTile` as I don't have Windows 10 rn.

PTAL. CC: @ahmed605

ahmed605:
LGTM

PR #17118, opened 2025-05-11
Code Quality: Replaced DllImport calls with CsWin32 generations for DPI calculation operations

Lamparter:
**Resolved / Related Issues**

- Related #15000 

1. Opened the app
2. Tested components that reference the migrated `AppModel.AppWindowDPI` property

---

`GetDpiForWindow` is already being used in some places however the legacy method pointing to a `DllImport` call remains in the Win32 PInvoke helper.
This PR removes it from the helper and replaces the final reference to `Win32PInvoke.GetDpiForWindow()` with the appropriate method generated by CsWin32.

### DIFF
--- a/src/Files.App/Data/Models/AppModel.cs
+++ b/src/Files.App/Data/Models/AppModel.cs
@@ -131,7 +131,7 @@ namespace Files.App.Data.Models
 		/// Gets or sets a value indicating the AppWindow DPI.
 		/// </summary>
 		private float _AppWindowDPI = PInvoke.GetDpiForWindow((HWND)MainWindow.Instance.WindowHandle) / 96f;
-		public float AppWindowDPI // TODO: Update value if the DPI changes
+		public float AppWindowDPI
 		{
 			get => _AppWindowDPI;
 			set => SetProperty(ref _AppWindowDPI, value);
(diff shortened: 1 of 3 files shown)


src/Files.App/Data/Models/AppModel.cs
```diff
@@ -127,10 +129,9 @@ public string PCloudDrivePath
 
 		/// <summary>
 		/// Gets or sets a value indicating the AppWindow DPI.
-		/// TODO update value if the DPI changes
 		/// </summary>
-		private float _AppWindowDPI = Win32PInvoke.GetDpiForWindow(MainWindow.Instance.WindowHandle) / 96f;
-		public float AppWindowDPI
+		private float _AppWindowDPI = PInvoke.GetDpiForWindow((HWND)MainWindow.Instance.WindowHandle) / 96f;
+		public float AppWindowDPI // TODO: Update value if the DPI changes
```
yaira2, line 134: Regarding the TODO, I don't believe this isn't such a common occurrence that we need to update when the value changes.

Lamparter: I can remove it if you like.

yaira2: Thanks!

PR #17120, opened 2025-05-11
Code Quality: Removed unused P/Invoke methods from helper class

Lamparter:
**Resolved / Related Issues**

- Related #15000 

1. Built Files

---

Removes a lot of methods, structs and enums that are no longer necessary or in use by Files.


yaira2:
@0x5bfa can confirm, but many of these were only disabled temporarily and are supposed to be readded back.

0x5bfa:
How come supposed to be re-added? If they aren't in use, I guess they are safe to be removed.

Lamparter:
All of the types, methods, enums and structures in the Win32 PInvoke helper should be replaced with CsWin32 generations unless they are undocumented APIs.
In that case they should be moved to the Extras class in Files.App.CsWin32.

0x5bfa:
If none of these are referenced from else where, I think it's safe to remove.

LGTM.

yaira2 (approved):

PR #14253, opened 2023-12-17
Feature: Added option to prioritize files when sorting

hishitetsu:
**Resolved / Related Issues**
- [X] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
   Closes #14226 

**Validation**
How did you test these changes?
- [X] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [X] Did you remove any strings from the en-us resource file?
   - [X] Did you search the solution to see if the string is still being used? 
- [X] Did you implement any design changes to an existing feature?
   - [X] Was this change approved?
- [ ] Are there any other steps that were used to validate these changes?

PR #17127, opened 2025-05-13
Fix: Ensure Home key moves cursor to start of filename during rename

AkhilReddy-Ramireddy:
This commit addresses the issue where pressing the Home key during file renaming in List View (and other views
inheriting from BaseGroupableLayoutPage) did not move the cursor to the absolute beginning of the filename if the
filename was longer than the visible TextBox. The fix ensures that textBox.SelectionStart is set to 0 and
textBox.SelectionLength is set to 0 when the Home key is pressed, providing behavior consistent with Windows
Explorer.

**Resolved / Related Issues**
Resolves #14253

### DIFF
--- a/src/Files.App/Views/Layouts/BaseGroupableLayoutPage.cs
+++ b/src/Files.App/Views/Layouts/BaseGroupableLayoutPage.cs
@@ -312,6 +312,11 @@ namespace Files.App.Views.Layouts
 					await CommitRenameAsync(textBox);
 					e.Handled = true;
 					break;
+				case VirtualKey.Home:
+					textBox.SelectionStart = 0;
+					textBox.SelectionLength = 0;
+					e.Handled = true;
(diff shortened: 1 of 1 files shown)


Lamparter:
Did you mean to link to #14253?

Josh65-2201:
@AkhilReddy-Ramireddy Please link to the correct issue, #14253 is a PR

AkhilReddy-Ramireddy:
Hi @Lamparter, thanks for checking! Yes, the PR description correctly links to issue #14253. I've also updated the PR title to remove any confusion.

yaira2:
@AkhilReddy-Ramireddy can you please update the PR description with the correct issue link and steps used to verify these changes? Thank you!

Lamparter:
#15048 is the correct issue that this PR should link to.

PR #17135, opened 2025-05-17
Code Quality: Added a static method to the PCWSTR struct that allows constructing from a string

Lamparter:
**Resolved / Related Issues**

- Related #15000 

1. Built the app
> This only introduces a new API that can be used in future PRs to make migrating from DllImport calls to CsWin32 generations easier, and does not have any effect over the app as it is.

---

This method makes it easier to pass in a `PCWSTR` struct to a P/Invoke method that might previously require a string input.
A lot of methods generated by CsWin32 that require unsafe will prefer a `PCWSTR` passed as an argument over a literal `System.String` however this method allows using a string value (via fixing the `string` in memory as a `char*`) for simplicity and so that code isn't reused.

e.g.:
```cs
unsafe void Example()
{
	HANDLE file = PInvoke.CreateFile(lpFileName: PCWSTR.FromString("example") ...);
}
```

### DIFF
--- a/src/Files.App.CsWin32/Extras.cs
+++ b/src/Files.App.CsWin32/Extras.cs
@@ -44,6 +44,20 @@ namespace Windows.Win32
 		public const int PixelFormat32bppARGB = 2498570;
 	}
 
+	namespace Foundation
+	{
+		public partial struct PCWSTR
+		{
+			public static unsafe PCWSTR FromString(string value)
+			{
+				fixed (char* p = value)
+				{
+					return new PCWSTR(p);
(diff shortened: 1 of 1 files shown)


0x5bfa (changes requested):
This can cause memory corruption and thats why CsWin32 doesn't provide something like this.

Lamparter:
> This can cause memory corruption and thats why CsWin32 doesn't provide something like this.

@0x5bfa how do you propose that the PCWSTR struct is created from a managed string value then?

PR #17136, opened 2025-05-17
Code Quality: Replaced Vanara with CsWin32 in ContextMenu class

Lamparter:
**Resolved / Related Issues**

- Related #15000 

WIP


src/Files.App.CsWin32/Extras.cs
```diff
@@ -44,6 +44,25 @@ public static unsafe nint SetWindowLongPtr(HWND hWnd, WINDOW_LONG_PTR_INDEX nInd
...
 
+	namespace Foundation
+	{
+		public partial struct PCSTR
+		{
+			public static unsafe PCSTR FromString(string value)
+			{
+				fixed (byte* p = global::System.Text.Encoding.UTF8.GetBytes(value))
+				{
+					return new PCSTR(p);
```
0x5bfa, line 55: I have no idea what you doing. Once getting out of `fixed`, the `p` won't be guaranteed to be fixed any longer.

src/Files.App.CsWin32/Extras.cs
```diff
@@ -44,6 +44,25 @@ public static unsafe nint SetWindowLongPtr(HWND hWnd, WINDOW_LONG_PTR_INDEX nInd
...
+				}
+			}
+
+			public static unsafe PCSTR FromInt(int value)
+			{
+				return new PCSTR((byte*)&value);
```
0x5bfa, line 61: Here too, taking pointer of an int value? For what.

0x5bfa (changes requested):
ContextMenu will retire in favor of my new group of helper tho.

src/Files.App.CsWin32/Extras.cs
Lamparter: How do I then construct the struct from a string/int value?

Lamparter:
> ContextMenu will retire in favor of my new group of helper tho.

Will it not reuse the code here?
If not, I'll close this PR.

src/Files.App.CsWin32/Extras.cs
0x5bfa: You have to pin the address unless the method you want to call is returned.

```
fixed (char* pName = name)
{
    // PInvoke.Method(..., pName, ...);
}
```

So, this case you can't create a helper like that.

0x5bfa:
I don't think so, I'll create a static ContextMenuManager along with JumpListManager, TrayIconManager, etc, all of which are AoT-safe.

That said, I don't think I'd be against this PR as far as this works and because I don't think mine will be implemented any time soon.

Issue #16029, opened 2024-08-18
Feature: Omnibar


yaira2:
### What feature or improvement do you think would benefit Files?

TBA

### Requirements

- Touch friendly
- Proper keyboard navigation #7077 #7086
- Responsive design #5513
- Add RichGlyph to command palette items #14707

## Bug tracker

- [x] Pressing esc in any mode should remove focus from Omnibar
- [x] Pressing enter in any mode should remove focus from Omnibar
- [x] Command Palette should display a message when no results are found
- [x] Focus Omnibar when using the Edit Path action
- [x] Pressing the down arrow when focused on a breadcrumb should expand its flyout
- [x] Filter out the "Command Palette" action from the Command Palette
- [x] Reset scroll position when opening the Command Palette flyout
- [x] Mode icons aren't updating when switching themes
- [x] Switching tabs while focused on the Omnibar causes a crash
- [x] Tab navigation isn't properly focusing on the Breadcrumb Bar
- [x] Chevrons are disappearing after adding a new tab
- [x] Switching between Path mode and Command mode a couple of times causes crash
- [x] Entering text in Command mode, switching to Edit path mode, and then back to Command mode doesn't clear the original text.
- [x] Pressing tab in Command mode closes flyout but doesn't switch to the next control
- [x] Switching to Path mode from Command mode isn't selecting the text
- [x] When breadcrumb flyout is open, clicking on Command mode switches to path mode.
- [x] Pressing enter key while focused on a mode button isn't switching modes (`PART_ModeButton` appears to be missing the keyboard events)
- [x] When Command mode flyout is open, clicking a button in the UI will not close the flyout
- [x] Add drag & drop support
- [x] On a fresh start, focusing on the Path Mode textbox doesn't open the suggestion flyout (but only when multiple tabs are open)
- [x] Suggestions don't update while typing
- [x] Holding the arrow key in Command Palette mode sometimes removes focus from the suggestions and instead filters the commands
- [x] Suggestions in path mode don't update when switching tabs
- [x] Display search suggestions in search mode
- [x] Omnibar text is incorrect when selecting suggestions via arrow keys in search mode
- [x] Update chevron tooltip to include folder name
- [x] Add tooltip to the root item chevron
- [x] If a file is selected, focusing on Command mode button via tab navigation and pressing enter opens the selected file
- [x] When using Command mode, switching tabs and then switching back reopens the Omnibar flyout
- [x] Switching folders via breadcrumb bar is incorrectly using a navigation transition
[...]

PR #17227, opened 2025-06-29
Code Quality: Use action description when selecting via arrow keys

yaira2:
**Resolved / Related Issues**

- For #16029

1. Confirmed selecting suggestions in Command Palette sets text to the description
2. Confirmed that it works to invoke commands

### DIFF
--- a/src/Files.App/UserControls/NavigationToolbar.xaml.cs
+++ b/src/Files.App/UserControls/NavigationToolbar.xaml.cs
@@ -258,36 +258,35 @@ namespace Files.App.UserControls
 
 		private async void Omnibar_QuerySubmitted(Omnibar sender, OmnibarQuerySubmittedEventArgs args)
 		{
-			if (Omnibar.CurrentSelectedMode == OmnibarPathMode)
+			var mode = Omnibar.CurrentSelectedMode;
+
+			// Path mode
+			if (mode == OmnibarPathMode)
 			{
 				await ViewModel.HandleItemNavigationAsync(args.Text);
 				(MainPageViewModel.SelectedTabItem?.TabItemContent as Control)?.Focus(FocusState.Programmatic);
+				return;
 			}
-			else if (Omnibar.CurrentSelectedMode == OmnibarCommandPaletteMode)
+
+			// Command palette mode
+			if (mode == OmnibarCommandPaletteMode)
 			{
-				if (args.Item is not NavigationBarSuggestionItem item)
-					return;
+				var item = args.Item as NavigationBarSuggestionItem;
 
 				// Try invoking built-in command
-				foreach (IRichCommand command in Commands)
+				foreach (var command in Commands)
 				{
-					if (item.Text == command.Description)
-					{
-						if (command == Commands.None)
-							await DialogDisplayHelper.ShowDialogAsync(Strings.InvalidCommand.GetLocalizedResource(),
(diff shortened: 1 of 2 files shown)


src/Files.App/UserControls/NavigationToolbar.xaml.cs
```diff
@@ -258,32 +258,35 @@ private void ClickablePath_GettingFocus(UIElement sender, GettingFocusEventArgs
...
+				return;
 			}
-			else if (Omnibar.CurrentSelectedMode == OmnibarCommandPaletteMode)
+
+			// Command palette mode
+			if (mode == OmnibarCommandPaletteMode)
 			{
-				if (args.Item is not NavigationBarSuggestionItem item)
-					return;
+				var item = args.Item as NavigationBarSuggestionItem;
```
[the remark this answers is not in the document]
yair100: We fallback to `args.Text` in the event that `item` is null.

src/Files.App/UserControls/NavigationToolbar.xaml.cs
```diff
@@ -258,32 +258,38 @@ private void ClickablePath_GettingFocus(UIElement sender, GettingFocusEventArgs
 
 		private async void Omnibar_QuerySubmitted(Omnibar sender, OmnibarQuerySubmittedEventArgs args)
 		{
-			if (Omnibar.CurrentSelectedMode == OmnibarPathMode)
+			var mode = Omnibar.CurrentSelectedMode;
```
0x5bfa, line 261: What's the reason to introduce a local variable here for the current mode instance?

src/Files.App/UserControls/NavigationToolbar.xaml.cs
```diff
@@ -258,32 +258,38 @@ private void ClickablePath_GettingFocus(UIElement sender, GettingFocusEventArgs
...
+			if (mode == OmnibarPathMode)
 			{
 				await ViewModel.HandleItemNavigationAsync(args.Text);
 				(MainPageViewModel.SelectedTabItem?.TabItemContent as Control)?.Focus(FocusState.Programmatic);
+				return;
 			}
-			else if (Omnibar.CurrentSelectedMode == OmnibarCommandPaletteMode)
+
+			// Command palette mode
+			if (mode == OmnibarCommandPaletteMode)
```
0x5bfa: What's the reason to move away from "else if"?

yair100: The code contains numerous if statements. My intention was to clearly separate the logic for each mode.

src/Files.App/UserControls/NavigationToolbar.xaml.cs
yair100: It's not a major difference, but this helps make the if statements a little cleaner.

src/Files.App/UserControls/NavigationToolbar.xaml.cs
0x5bfa: You can still use "else if" with the comment. if you use "if" here, even the current mode is Path mode, `mode == Palette` and `mode == Search` are evaludated and concequently takes a few milisec.

yair100: I pushed a change

PR #17248, opened 2025-07-04
Fixed an issue where the UI thread can fall into dead-lock state when switching Omnibar modes

0x5bfa:
### Resolved / Related Issues

- #16029

1. Open Files app
2. Switch modes between Path and Palette multiple times.


src/Files.App/UserControls/NavigationToolbar.xaml
```diff
@@ -445,6 +445,7 @@
 							<keyboard:KeyboardShortcut
 								x:Name="RightAlignedKeyboardShortcut"
 								Grid.Column="2"
+								Margin="0,0,8,0"
```
yair100: It might be cleaner to move the margin to the containing grid.

src/Files.App/ViewModels/UserControls/NavigationToolbarViewModel.cs
```diff
@@ -1200,49 +1148,41 @@ void AddNoResultsItem()
...
 		public void PopulateOmnibarSuggestionsForCommandPaletteMode()
 		{
-			OmnibarCommandPaletteModeText ??= string.Empty;
-			OmnibarCommandPaletteModeSuggestionItems.Clear();
-
 			if (ContentPageContext.SelectedItems.Count == 1 && ContentPageContext.SelectedItem is not null && !ContentPageContext.SelectedItem.IsFolder)
 			{
-				var dispatcherQueue = Microsoft.UI.Dispatching.DispatcherQueue.GetForCurrentThread();
-
-				dispatcherQueue.TryEnqueue(() =>
```
yair100, line 1210: I found that `DispatcherQueue` is required for actions to work properly. Unfortunately, we can't do any further testing in this area right now because the experimental API doesn't seem to work in the latest Insider build.

0x5bfa: Please look at the L439 of NavigationToolbar.xaml.cs, I've wrapped the whole method with it.

yair100: I see, thanks.

yaira2 (approved):
LGTM

src/Files.App/ViewModels/UserControls/NavigationToolbarViewModel.cs
yair100: It looks like it's missing from the `Omnibar_TextChanged` event. To prevent this from being missed in the future, it would be safer to revert this change.

PR #17258, opened 2025-07-08
Code Quality: Use newer syntaxes for ArgumentException handling

corvinsz:
Updated code to use newer syntaxes and APIs.
There is no issue related to this PR.

Some of the exceptions I purposely didn't change to preserve the custom message, e.g.:
```csharp
if (string.IsNullOrEmpty(backingFile.Path))
{
    throw new ArgumentException("Backing file Path cannot be null");
}
```


yair100:
Thank you for your interest in Files. Please link to an approved issue as per the [contributing guidelines](https://github.com/files-community/Files/blob/main/.github/CONTRIBUTING.md).

corvinsz:
> Where is this? I think it's good to change them as well.

https://github.com/corvinsz/Files/blob/fe5215713dba4302a5a263a5309025bed79a707c/src/Files.App/Utils/Storage/StorageItems/ZipStorageFolder.cs?plain=1#L50-L53

I can fix it this evening, if you are not faster than me 😄

0x5bfa:
I'm tied up with something else, so feel free 🙂

PR #17260, opened 2025-07-09
Code Quality: Fixed selecting Omnibar suggestions via keyboard

yaira2:
**Resolved / Related Issues**

- For #16029
- Fixed 'Holding the arrow key in Command Palette mode sometimes removes focus from the suggestions and instead filters the commands'

### DIFF
--- a/src/Files.App.Controls/Omnibar/Omnibar.Events.cs
+++ b/src/Files.App.Controls/Omnibar/Omnibar.Events.cs
@@ -131,16 +131,12 @@ namespace Files.App.Controls
 
 			// UpdateSuggestionListView();
 
-			if (_textChangeReason is not OmnibarTextChangeReason.SuggestionChosen and
-				not OmnibarTextChangeReason.ProgrammaticChange)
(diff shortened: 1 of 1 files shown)


src/Files.App.Controls/Omnibar/Omnibar.Events.cs
```diff
@@ -131,14 +131,12 @@ private void AutoSuggestBox_TextChanged(object sender, TextChangedEventArgs e)
 
 			// UpdateSuggestionListView();
 
-			if (_textChangeReason is not OmnibarTextChangeReason.SuggestionChosen and
-				not OmnibarTextChangeReason.ProgrammaticChange)
+			if (_textChangeReason is OmnibarTextChangeReason.ProgrammaticChange)
+				_textBox.SelectAll();
+			else
 			{
-				_textChangeReason = OmnibarTextChangeReason.UserInput;
```
0x5bfa, line 137: Removing this alright? When text is being changed, how it being changed would be still needed

yair100: It's causing issues, do you remember why you added it?

PR #17274, opened 2025-07-16
Code Quality: Use UI thread when populating suggestions

0x5bfa:
### Resolved / Related Issues

- #16029

1. Open Files app
2. Load suggestions in Omnibar

### DIFF
--- a/src/Files.App/ViewModels/UserControls/NavigationToolbarViewModel.cs
+++ b/src/Files.App/ViewModels/UserControls/NavigationToolbarViewModel.cs
@@ -1171,7 +1171,7 @@ namespace Files.App.ViewModels.UserControls
 			}
 		}
 
-		public void PopulateOmnibarSuggestionsForCommandPaletteMode()
+		public async Task PopulateOmnibarSuggestionsForCommandPaletteMode()
 		{
 			var newSuggestions = new List<NavigationBarSuggestionItem>();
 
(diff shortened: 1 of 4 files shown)


src/Files.App/ViewModels/UserControls/NavigationToolbarViewModel.cs
```diff
@@ -236,11 +243,7 @@ public string? PathText
 		public string? OmnibarSearchModeText { get => _OmnibarSearchModeText; set => SetProperty(ref _OmnibarSearchModeText, value); }
 
 		private string _OmnibarCurrentSelectedModeName = OmnibarPathModeName;
-		public string OmnibarCurrentSelectedModeName
-		{
-			get => _OmnibarCurrentSelectedModeName;
-			set => SetProperty(ref _OmnibarCurrentSelectedModeName, value);
-		}
+		public string OmnibarCurrentSelectedModeName { get => _OmnibarCurrentSelectedModeName; set => SetProperty(ref _OmnibarCurrentSelectedModeName, value); }
```
yair100, line 246: Was the formatting changed intentionally?

0x5bfa: Yes and others done already.

This is because this will be replaced with a SG'd [ObservableProperty] which uses `field` keyword, which is still not out of C# language preview.

```cs
[ObservableProperty]
public partial string Name { get; set; }
```

It will be -1/+1 change and should look better in a PR diff where we implement this in a future.

yaira2 (approved):
LGTM

PR #17283, opened 2025-07-20
Code Quality: Improved Axe assertion message

0x5bfa:
### Resolved / Related Issues

_None_

_None_


0x5bfa:
This PR fails with a unknown AXE issue but because now we know the poisiton of items causing that issue, I figured out that it was caused by an element located in the right top (and presumably Mode buttons not having a11y name.), which was fixed in the other PR.

yair100:
Why is this failing if no xaml has been changed?

PR #17300, opened 2025-07-23
Code Quality: Improved display of plural strings in Status Center

### LINKED ISSUE #17295: Bug: No plural forms in Status Center actions
### Description

The current strings in Status Center actions are not plural form compatible.
For example: StatusCenter_CopyInProgress_Header -> Copying {0} item(s) to "{1}"
It should be Copying {0, plural, one {# item} other {# items}} to "{1}"

I tried changing the string, but the program crashes with System.FormatException as `card.HeaderStringResource.GetLocalizedResource();` does not give the right format.

I cannot figure out how to use `GetLocalizedFormatResource` (as it is done in other places) in this context to provide the necessary plural form correctly.
I am open to suggestions on how to fix this :D

### Steps To Reproduce

-

### Files Version

Dev

### Windows Version

11

### User ID

_No response_

### Log File

-

SilverGreen93:
**Changes**

Update Status Center strings to include plural forms for `item(s)` and `font(s)`.
Use the `GetLocalizedFormatResource` instead of `GetLocalizedResource` and move the `headerString` and `subHeaderString` initialization to each corresponding case.
Also removed the redundant `if (headerString is not null)` checks, as the string cannot be null at that point.

**Resolved / Related Issues**

- Closes #17295

Tested the changed strings in file operations.

image

### DIFF
--- a/src/Files.App/Utils/StatusCenter/StatusCenterHelper.cs
+++ b/src/Files.App/Utils/StatusCenter/StatusCenterHelper.cs
@@ -640,218 +640,207 @@ namespace Files.App.Utils.StatusCenter
 				destinationDirName = destinationPath.Split('\\').Last();
 			}
 
-			string headerString = string.IsNullOrWhiteSpace(card.HeaderStringResource) ? string.Empty : card.HeaderStringResource.GetLocalizedResource();
-			string subHeaderString = string.IsNullOrWhiteSpace(card.SubHeaderStringResource) ? string.Empty : card.SubHeaderStringResource.GetLocalizedResource();
-
 			// Update string resources
 			switch (card.Operation)
 			{
 				case FileOperationType.Copy:
 					{
-						if (headerString is not null)
+						string headerString = string.IsNullOrWhiteSpace(card.HeaderStringResource) ? string.Empty
+							: card.HeaderStringResource.GetLocalizedFormatResource(card.TotalItemsCount, destinationDirName);
+						card.Header = card.FileSystemOperationReturnResult switch
 						{
-							card.Header = card.FileSystemOperationReturnResult switch
-							{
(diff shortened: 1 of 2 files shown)


yair100:
This looks very good. Did you confirm that each operation state is displaying correctly? This status center is prone to crashing if a string is configured wrong 🙂

src/Files.App/Utils/StatusCenter/StatusCenterHelper.cs
```diff
@@ -640,218 +640,207 @@ public static void UpdateCardStrings(StatusCenterItem card)
...
-								ReturnResult.InProgress => string.Format(headerString, card.TotalItemsCount, destinationDirName),
-								_ => string.Format(headerString, card.TotalItemsCount, destinationDirName),
-							};
-						}
-						if (subHeaderString is not null)
+							ReturnResult.Cancelled => string.Format(headerString, card.TotalItemsCount, destinationDirName),
+							ReturnResult.Success => string.Format(headerString, card.TotalItemsCount, destinationDirName),
+							ReturnResult.Failed => string.Format(headerString, card.TotalItemsCount, destinationDirName),
+							ReturnResult.InProgress => string.Format(headerString, card.TotalItemsCount, destinationDirName),
+							_ => string.Format(headerString, card.TotalItemsCount, destinationDirName),
```
0x5bfa, line 656: Looks like they do the same thing, what is this code for?

yair100: It's best to keep it as is in case we want to change the strings in the future.

SilverGreen93:
> This looks very good. Did you confirm that each operation state is displaying correctly? This status center is prone to crashing if a string is configured wrong 🙂

Tested all except for failure cases. They all seem to work.

yair100:
@SilverGreen93 thank you!

PR #17334, opened 2025-07-27
Feature: Added `Digital Signatures` tab to the properties window

### LINKED ISSUE #3723: Show digital signatures info in properties.
image

I want to view and verify the digital signatures info in Files. I think we can add a tab in properties, and show the signature details directly.

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

ferrariofilippo:
**Resolved / Related Issues**

- Closes #3723 

1. Download some `.exe`s (or any other signed file)
2. Open their properties
3. Open the `Digital Signatures` tab and check you see the right data
4. Click on the `Details` button

https://github.com/user-attachments/assets/0e79fe84-c616-40b8-aafb-f50c5cd44569


src/Files.App/Data/Enums/PropertiesNavigationViewItemType.cs
```diff
@@ -47,5 +47,10 @@ public enum PropertiesNavigationViewItemType
 		/// Shortcut page type
 		/// </summary>
 		Shortcut,
-	}
+
+        /// <summary>
+        /// Signatures page type
+        /// </summary>
+        Signatures,
```
yair100: The tab spacing looks a little off, should be an easy fix 🙂

src/Files.App/Data/Factories/PropertiesNavigationViewItemFactory.cs
```diff
@@ -107,7 +115,10 @@ public static ObservableCollection<NavigationViewItemButtonStyleItem> Initialize
 					PropertiesNavigationViewItems.Remove(securityItem);
 
 				if (!hashItemEnabled)
+				{
 					PropertiesNavigationViewItems.Remove(hashesItem);
+					PropertiesNavigationViewItems.Remove(signaturesItem);
```
yair100, line 126: Is there ever a time when the hashes page is displayed, but signatures should be hidden?

ferrariofilippo: Good call 😄
I added a check to show the section only for supported file types

BTW, when an item has no signature, should we show a message like below or leave the section empty?
image

yair100: Not showing anything might look like a bug, I think it's best to show a message.

ferrariofilippo: FYI https://github.com/files-community/Files/pull/17334/commits/571d0a2f12ae9a3fabfb3f5ec4065776794ff33b already shows the message

yair100: @mdtauk do you have any suggestions for the empty state?

ferrariofilippo:
BTW I think we should reference this project (https://github.com/leeqwind/PESignAnalyzer/) somewhere, since the business logic is theirs translated to C#

yair100:
We have space on the about page 👍

src/Files.App/Helpers/Win32/Win32PInvoke.Methods.cs
Lamparter, line 1: Isn't adding more methods and types to this class an anti-pattern?
As @0x5BFA said before [here](https://github.com/files-community/Files/issues/15000#issuecomment-2404012228), all win32 platform invoke functions must be from CsWin32 projections and not hardcoded. This would be creating more work for the future.

ferrariofilippo: I think it's ready for review now. I replaced all the methods/structures which had an equivalent `PInvoke.` version.
It would be really helpful if you two could help me reviewing this.
Everything seems to work but I'm not sure if there's any leak or similar

0x5bfa: You can add `CertGetNameString` to `NativeMethods.txt` and replace. I understand CryptUIDlgViewSignerInfo is no longer included to the Windows SDK so it's fine to leave it.

0x5bfa (approved):
There're some tiny improvements that can be made but this is already working very well. I can take care of unmanaged stuff.

Behavior wise, LGTM.

src/Files.App/Helpers/Win32/Win32PInvoke.Methods.cs
0x5bfa: And if you would like, please add it here as Yair mentioned. We have a backend list in the `AboutViewModel`.

image

src/Files.App/Data/Factories/PropertiesNavigationViewItemFactory.cs
mdtauk: Other than disabling the Signature sidebar item, I think it depends on what UI is added.  I was sharing my thoughts with @0x5bfa and we could have empty lists if its something like this.

image

src/Files.App/Helpers/Win32/Win32PInvoke.Methods.cs
ferrariofilippo: Done

src/Files.App/Data/Factories/PropertiesNavigationViewItemFactory.cs
ferrariofilippo: > Other than disabling the Signature sidebar item

To do this we should pre-load signatures (currently they are loaded when you open the tab). Is this ok?


> I was sharing my thoughts with @0x5bfa and we could have empty lists if its something like this.

Do I have to implement this now?

yair100: We already agreed on the current UI so I wouldn't make changes right now.

src/Files.App/Data/Factories/PropertiesNavigationViewItemFactory.cs
```diff
@@ -3,12 +3,25 @@
...
 using Files.Shared.Helpers;
 using Microsoft.UI.Xaml;
+using System.Collections.Frozen;
 using Windows.Storage;
 
 namespace Files.App.Data.Factories
 {
 	public static class PropertiesNavigationViewItemFactory
 	{
+		private static readonly FrozenSet<string> _signableTypes = new HashSet<string>()
```
yair100: We have a `FileExtensionHelpers`, would it make sense to move this there?

ferrariofilippo: Yes, I'll do that, but I'll keep it as a `Set` instead of using the existing methods because it would require many calls

yair100: Would it be worth refactoring the other extension groups as well (in a separate PR)?

yaira2 (approved):
LGTM

PR #17345, opened 2025-07-29
Feature: Major Speed Optimizations

elliotttate:
Performance improvements for file navigation and enumeration:

- Increased concurrency limits: 4 concurrent folder enumerations, 20 concurrent thumbnail loads
- Implemented smart preloading for visible items with viewport-based loading
- Added differential collection updates to prevent UI flashing
- Increased batch size from 32 to 200 items for better throughput
- Added icon caching with WeakReference for memory efficiency
- Removed expensive async operations during file enumeration
- Added performance monitoring for slow operations
- Fixed icon flashing by setting LoadFileIcon = true immediately
- Optimized zip file detection to avoid unnecessary async calls

These changes significantly improve perceived performance and reduce UI lag during file navigation.

### DIFF
--- a/src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
+++ b/src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
@@ -15,6 +15,9 @@ namespace Files.App.Utils.Storage
 		private static readonly IStorageCacheService fileListCache = Ioc.Default.GetRequiredService<IStorageCacheService>();
 
 		private static readonly string folderTypeTextLocalized = Strings.Folder.GetLocalizedResource();
+		
+		// Performance optimization: Increased batch size for better throughput
+		private const int BATCH_SIZE = 200;
 
 		public static async Task<List<ListedItem>> ListEntries(
 			string path,
@@ -89,7 +92,7 @@ namespace Files.App.Utils.Storage
 				if (cancellationToken.IsCancellationRequested || count == countLimit)
 					break;
 
-				if (intermediateAction is not null && (count == 32 || sampler.CheckNow()))
+				if (intermediateAction is not null && (count == BATCH_SIZE || sampler.CheckNow()))
 				{
 					await intermediateAction(tempList);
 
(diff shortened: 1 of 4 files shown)


src/Files.App/Utils/Storage/Enumerators/Win32StorageEnumerator.cs
```diff
@@ -391,14 +397,16 @@ CancellationToken cancellationToken
 			}
 			else
 			{
-				if (ZipStorageFolder.IsZipPath(itemPath) && await ZipStorageFolder.CheckDefaultZipApp(itemPath))
```
yair100, line 394: Please note, changing this logic is likely to cause issues.
#6457

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -1328,9 +1600,9 @@ await dispatcherQueue.EnqueueOrInvokeAsync(() =>
 						{
 							_ = Task.Run(async () =>
 							{
-								await Task.Delay(500);
```
yair100, line 1331: @hishitetsu can confirm, but if I recall correctly, this delay is to allow time for Windows to finish caching the thumbnail for items in cloud drives.

Josh65-2201:
There's a few issues, 
Thumbnails won't load if the folder is refreshed
Git information doesn't load at all

I had one time where all the folders icons would be the same in medium layout size, I have restarted and not been able to reproduce it

It's also slower for me then in preview, clicking on a folder takes a while to start highlighting it and then only after it's highlighted will it open

src/Files.App/ViewModels/ShellViewModel.cs
hishitetsu: Correct. Without this delay, thumbnails will not be able to load.

0x5bfa: We need a comment for this fr...

yair100: There is one a little higher up. That said, there are a lot of intricacies and this code is difficult to follow even with comments.

neil1111: Sorry for jumping in here, but is this related to the very slow performance in a business environment where OneDrive is continuously syncing with the local C: drive?  Opening folders is extremely slow (multiple seconds) even when a folder was recently open.  Could this be changed to a setting preference?  Many users (like me) do not need thumbnails or would sacrifice them for extremely high performance.

yair100:
Closing due to inactivity.

PR #17348, opened 2025-07-29
Fix: Fixed issue where delete key wouldn't work after removing item from selection 

### LINKED ISSUE #17298: Bug: Delete key doesn't work when removing one item from selection
### Description

When You Select Multiple Files From a list and deselect one in the middle You are then unable to delete the files

### Steps To Reproduce

1. Select multiple files
2. Remove one file from the selection
3. Press delete on the Keyboard
4. Nothing happens

### Files Version

3.9.1.0

### Windows Version

10.0.26100.4652

### User ID

84cc6017-b5fa-4a79-b028-9bf23472b03b

### Log File

Can not find the log file

ferrariofilippo:
**Resolved / Related Issues**

- Closes #17298 

**Notes**
It seems the issue is caused by the checkbox, which when focused handles all `Key`-related events.
For some reason I don't understand, the bug affects only `Deselection` and not `Selection`
The code below removes the focus from the checkbox once its state is updated

1. Open a folder with multiple files (test different layouts)
2. Select some of them
3. Deselect one clicking on the checkbox
4. Hit the `Del` key and see that the modal pops-up
5. Repeat 3 and 4 as many times as you wish

https://github.com/user-attachments/assets/edf8b79c-d962-45fd-b08c-5ac6368f5e60

### DIFF
--- a/src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
+++ b/src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
@@ -670,18 +670,18 @@ namespace Files.App.Views.Layouts
 
 		private void ItemSelected_Unchecked(object sender, RoutedEventArgs e)
 		{
-			if (sender is CheckBox checkBox)
-			{
-				if (checkBox.DataContext is ListedItem item && FileList.SelectedItems.Contains(item))
-					FileList.SelectedItems.Remove(item);
+			if (sender is not CheckBox checkBox)
+				return;
 
-				// Workaround for #17298
-				checkBox.IsTabStop = false;
-				checkBox.IsEnabled = false;
-				checkBox.IsEnabled = true;
-				checkBox.IsTabStop = true;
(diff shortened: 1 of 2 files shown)


src/Files.App/Views/Layouts/GridLayoutPage.xaml.cs
```diff
@@ -670,10 +670,18 @@ checkBox.DataContext is ListedItem item &&
 
 		private void ItemSelected_Unchecked(object sender, RoutedEventArgs e)
 		{
-			if (sender is CheckBox checkBox &&
-				checkBox.DataContext is ListedItem item &&
-				FileList.SelectedItems.Contains(item))
-				FileList.SelectedItems.Remove(item);
+			if (sender is CheckBox checkBox)
+			{
```
yair100: Can you return early if this statement is false? This way we can avoid a nested if statement and make it more readable at the same time.

yaira2 (approved):

PR #17351, opened 2025-07-29
GitHub: Update FUNDING.yml

Lamparter:
I just realised this entry was never added to the sponsorship file when @0x5BFA launched their sponsors profile 😊


yair100:
Thank you for the PR. We looked into this earlier but there is a limit of four profiles.

Issue #6534, opened 2021-10-17
Resizable column view


jainal09:
### What's the Problem?

For files having long names. In column view it only covers some part of name. Having a feature to resize the column by clicking is absolutely necessary and needed!
(SEE SCREEN SHOT)
![image](https://user-images.githubusercontent.com/34179361/137611071-cc7152f2-e75a-4934-91fc-c22b8be5a411.png)

### Solution/Idea

Having a feature to resize the column by clicking.

### Alternatives

Not using column mode. But I really like it.

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish W | Must |
| This proposal will allow end users to accomplish X | Should |
| This proposal will allow developers to accomplish Y | Could |
| This proposal will allow end users to accomplish Z | Won't |

### Files Version

Version: 2.0.13.0 OS Version: 10.0.22478.1012

### Windows Version

Windows 11 22478.1012

jainal09:
@yaichenbaum at least not a scroll bar. But just a  feature to select and drag the line

sergiubacioiu:
An option in settings menu to resize all the columns would be great too. I mostly use a maximum of 2/3 subfolders so most of the time 30% of my screen is empty.

yaira2:
Are the custom column sizes supposed to be saved for whenever the column is opened again?

andrefromsbg:
hello, is there anything happening with this task? files is really a great file browser for windows. but the fact that you can't change the column width keeps me, and probably many others who are used to this rudimentary function of apple finder, from using files as a permanent alternative...

yaira2:
> Are the custom column sizes supposed to be saved for whenever the directory is opened again?

@jainal09

michael-oz-smith:
Saving column sizes would be helpful, but simply having the ability to resize columns is an ASAP necessity.

Users coming over from Macs are very accustomed to having this capability since its been an OS X feature since launch.

yaira2:
Does MacOS save the sizes?

Lukiluc29:
> Does MacOS save the sizes?

I believe and I even think that if this function is not on MacOS it would be better to put it like that we don't need to readjust the columns all the time to see the names of the documents

jainal09:
> 

No macos doesnt save the resized column sizes

Lukiluc29:
Yes but it's still better to include this option if possible and if it's not too long to code

redhotideas:
[...]

PR #17380, opened 2025-07-31
Feature: Added support for resizing columns in the Columns View

workbysaran:
**Resolved / Related Issues**

Partially addresses #6534
- [x] Support for resizing columns  
- [ ] Columns sizes should be saved across sessions 
- [x] Double clicking the resize bar should size the column to fit

1. Opened Files and confirmed that:
   - Columns are resizable using the splitter between them
   - Cursor changes to a resize icon when hovering over the column divider
   - Double-clicking the divider resets to default width
   - Content inside each column adjusts correctly to new width
   - No visual glitches appear when resizing multiple columns rapidly

2. Verified that:
   - Column resizing works with mouse.
   - Resizing does not interfere with item selection, list scroll or navigation
   - Minimum and maximum column widths are respected


src/Files.App.Controls/BladeView/BladeItem.cs
```diff
@@ -50,5 +86,177 @@ private void CloseButton_Click(object sender, RoutedEventArgs e)
...
+				return maxWidth;
+			}
+			catch (Exception ex)
+			{
+				Debug.WriteLine($"Error measuring content width: {ex.Message}");
+				// Fallback calculation
+				return 200; // Default reasonable width
+			}
+		}
+		private List<TextBlock> GetTextBlocksFromVisualTree(DependencyObject parent)
```
yair100, line 208:
```suggestion

		private List<TextBlock> GetTextBlocksFromVisualTree(DependencyObject parent)
```

src/Files.App.Controls/BladeView/BladeItem.cs
```diff
@@ -36,7 +48,31 @@ protected override void OnApplyTemplate()
 
 			_closeButton.Click -= CloseButton_Click;
 			_closeButton.Click += CloseButton_Click;
+
+			_bladeResizer = GetTemplateChild("BladeResizer") as Border;
+
+			if (_bladeResizer != null)
```
yair100, line 54: These events are already registered in `BladeView.xaml`, is there any reason to register them again from code behind?

workbysaran: I've removed the event handlers from the XAML and kept them only in the code-behind using OnApplyTemplate, as it's required for better reusability of the control.

yair100: Thanks.

yaira2 (approved):
LGTM

PR #17421, opened 2025-08-08
Fix: Fixed issue where repository names lost casing and retained .git suffix

### LINKED ISSUE #17386: Bug: Git repository directory name contains .git suffix
### Description

If you clone a repository by the full URL (i.e. `https://github.com/files-community/Files.git`), the destination directory is named `files.git` instead of `Files`.

So there are really 2 issues here:
1. The directory name should NOT contain the `.git` suffix
2. The directory name should be capitalized exactly as it is in the URL: `Files`, not `files`

### Steps To Reproduce

1. Open command palette
2. Select Clone Git repo
3. Paste url `https://github.com/files-community/Files.git`
4. Clone and observe the result as described above.

### Files Version

3.9.13.0 - dev

### Windows Version

10.0.26100.4770

### User ID

2a411168-c5ac-4f2d-92f9-be000c77eac5

### Log File

-

marcofranzen99:
**Resolved / Related Issues**
- Closes #17386

1. Used the clone action from the command palette (or drag and drop the URL)
2. Pasted url `https://github.com/files-community/Files.git` or `https://github.com/files-community/Files` 
3. Verified that casing was preserved (uppercase F) and the .git suffix was removed.

image

image

### DIFF
--- a/src/Files.App/Utils/Git/GitHelpers.cs
+++ b/src/Files.App/Utils/Git/GitHelpers.cs
@@ -7,10 +7,11 @@ using Microsoft.Extensions.Logging;
 using System.Net.Http;
 using System.Net.Http.Json;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace Files.App.Utils.Git
 {
-	internal static class GitHelpers
+	internal static partial class GitHelpers
 	{
 		private static readonly StatusCenterViewModel StatusCenterViewModel = Ioc.Default.GetRequiredService<StatusCenterViewModel>();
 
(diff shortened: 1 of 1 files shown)


yair100:
These changes appear to block the ability to drag URLs like `https://github.com/files-community/Files/pull/17421`. The feature was initially implemented to support dragging URLs even if they weren't the root of the repo.

marcofranzen99:
> These changes appear to block the ability to drag URLs like `https://github.com/files-community/Files/pull/17421`. The feature was initially implemented to support dragging URLs even if they weren't the root of the repo.

Good point, I didn't know that feature exists.
I have adjusted the regex so that the URL doesn't have to end with the repo name or a slash. Everything after the repo name gets ignored now

image

I think it would be useful to add a unit test for the IsValidRepoUrl method, for example. Since there is no UnitTests project, I left it out for now. What do you think of this?

yaira2 (approved):
LGTM

yair100:
> I think it would be useful to add a unit test for the IsValidRepoUrl method, for example. Since there is no UnitTests project, I left it out for now. What do you think of this?

It would be helpful 👍

PR #17026, opened 2025-04-07
Feature: Use UTF-8 encoding for ZIP creation to ensure cross-language compatibility

oxygen-dioxide:
Currently, when creating ZIP archives, `Files` defaults to the system local encoding (e.g., GBK for Chinese, Shift_JIS for Japanese). UTF-8 is used only when there are characters in file name that can't be expressed in system encoding. This causes filenames to become unreadable when the archive is opened on systems with a different language locale. For example, a Chinese user creates a ZIP with filenames in GBK → Japanese users extract it as Shift_JIS → garbled text.

After this PR, `Files` always creates ZIP archive in UTF-8 encoding, which is consistent with windows explorer's behaviour. The archive created by `Files` will never become garbled no matter who downloads it.

**Resolved / Related Issues**

To prevent extra work, all changes to the Files codebase must link to an approved issue marked as `Ready to build`. Please insert the issue number following the hashtag with the issue number that this Pull Request resolves.
- #14937 is partially solved in this PR.

**Steps used to test these changes**

Stability is a top priority for Files and all changes are required to go through testing before being merged into the repo. Please include a list of steps that you used to test this PR.

I'm a Chinese developer, so below is how I test it in Simplified Chinese windows. I don't know whether this affects English users and how they'll test this.

1. install python, and `pip install zipunicode`
2. Create a file with Simplified Chinese filename. for example, `中文.txt`
3. In `Files`, right click on the file, Compress -> Create 中文.zip
4. In command line, `zipu 中文.zip`

Previous behaviour: 
```
* Detected encoding  :  GB2312 | Language:Chinese | Confidence:86%
```

New Behaviour:
```
* All file names are properly in UTF8 encoding
```

PR #17426, opened 2025-08-09
Fix: Fixed issue where 7z archive files could not be previewed or compressed

### LINKED ISSUE #17257: Bug: Unable to compress/preview 7z archive files.
### Description

As mentioned, in **Files - Preview**, I can't properly create a **7z archive file** and **open files within it**.

### Steps To Reproduce

1. Select multiple files.
2. Right-click the mouse and select "Create *xxx.7z*" under the "Compress" option in the context menu.
3. An error appears in the Notification Center: "Failed to compress *n* items into '*xxx*'".
4. Open the newly created 7z archive file using **Files - Preview**.
5. Try opening any file in the preview view.
6. The file is empty after opening.

### Files Version

3.9.11.0

### Windows Version

10.0.26120.0

### User ID

fee4d11a-4248-4e98-a8f2-09fbcd837b11

### Log File

Creating a 7z archive file:

[debug.log](https://github.com/user-attachments/files/21126270/debug.log)

Previewing the 7z archive file and open files within it:

[debug.log](https://github.com/user-attachments/files/21126318/debug.log)

NicholasChin28:
**Resolved / Related Issues**

- Closes #17257

1. Trace the logic of 7z archive compression by Right-clicking the mouse and selecting "Create xxx.7z" under the "Compress" option in the context menu.
2. Upon step debugging, found out that adding the custom parameter `cu` was throwing an exception from the file with the error of `System.ArgumentException: Value does not fall within the expected range`
3. After applying the fixes, compress multiple files by selecting `Create xxx.7z` under the `Compress` option in the context menu and then extract the created `7z` archive.

**Fix Description**
- Upon checking the documentation from the [site](https://7-zip.opensource.jp/chm/cmdline/switches/method.htm#7Z), I noticed that the `7z` archive does not contain the parameter `cu`.
- I assume this could be the one causing the issue as when I tested without the `cu` archive, the bug reported by the original poster is fixed.
- I also checked the `.chm` file from my downloaded `7zip` instance. The version is 7-Zip 24.09 (x64)

### DIFF
--- a/src/Files.App/Data/Models/CompressArchiveModel.cs
+++ b/src/Files.App/Data/Models/CompressArchiveModel.cs
@@ -161,7 +161,7 @@ namespace Files.App.Data.Models
 			//Use UTF-8 encoding. 
 			//References: 7-zip chm --> Command Line Version --> Switches
 			//--> -m --> cu=[off | on].
-			//See https://github.com/files-community/Files/issues/17257
+			//Don't add "cu" parameter for 7zip files, see https://github.com/files-community/Files/issues/17257
 			if (FileFormat != ArchiveFormats.SevenZip)
 				compressor.CustomParameters.Add("cu", "on");
 
--- a/src/Files.App/Data/Models/CompressArchiveModel.cs
+++ b/src/Files.App/Data/Models/CompressArchiveModel.cs
@@ -161,6 +161,7 @@ namespace Files.App.Data.Models
 			//Use UTF-8 encoding. 
 			//References: 7-zip chm --> Command Line Version --> Switches
 			//--> -m --> cu=[off | on].
+			//See https://github.com/files-community/Files/issues/17257
 			if (FileFormat != ArchiveFormats.SevenZip)
 				compressor.CustomParameters.Add("cu", "on");
 
--- a/src/Files.App/Data/Models/CompressArchiveModel.cs
+++ b/src/Files.App/Data/Models/CompressArchiveModel.cs
@@ -161,7 +161,6 @@ namespace Files.App.Data.Models
 			//Use UTF-8 encoding. 
 			//References: 7-zip chm --> Command Line Version --> Switches
 			//--> -m --> cu=[off | on].
-			compressor.CustomParameters.Add("cu", "on");
 
 			compressor.Compressing += Compressor_Compressing;
 			compressor.FileCompressionStarted += Compressor_FileCompressionStarted;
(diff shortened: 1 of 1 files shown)


Josh65-2201:
The CU code needed for zips for https://github.com/files-community/Files/pull/17026

yair100:
> The CU code needed for zips for #17026

Can it be made conditional?

NicholasChin28:
Updated the code to exclude adding the parameter if its 7z archive

src/Files.App/Data/Models/CompressArchiveModel.cs
```diff
@@ -161,7 +161,8 @@ public async Task<bool> RunCreationAsync()
 			//Use UTF-8 encoding. 
 			//References: 7-zip chm --> Command Line Version --> Switches
 			//--> -m --> cu=[off | on].
-			compressor.CustomParameters.Add("cu", "on");
+			if (FileFormat != ArchiveFormats.SevenZip)
```
yair100, line 165: This looks good, can you please add a code comment linking to the issue for future reference?

yair100: `// Don't add "cu" parameter for 7zip files, see https://github.com/files-community/Files/issues/17257`

src/Files.App/Data/Models/CompressArchiveModel.cs
```diff
@@ -161,7 +161,9 @@ public async Task<bool> RunCreationAsync()
 			//Use UTF-8 encoding. 
 			//References: 7-zip chm --> Command Line Version --> Switches
 			//--> -m --> cu=[off | on].
-			compressor.CustomParameters.Add("cu", "on");
+			//See https://github.com/files-community/Files/issues/17257
```
yair100:
```suggestion
			// Don't add "cu" parameter for 7zip files
			// See https://github.com/files-community/Files/issues/17257
```

src/Files.App/Data/Models/CompressArchiveModel.cs
NicholasChin28: Updated with the comments

yair100: Thanks again!

src/Files.App/Data/Models/CompressArchiveModel.cs
```diff
@@ -161,7 +161,9 @@ public async Task<bool> RunCreationAsync()
 			//Use UTF-8 encoding. 
 			//References: 7-zip chm --> Command Line Version --> Switches
 			//--> -m --> cu=[off | on].
```
yaira2, line 163:
```suggestion
// Use UTF-8 encoding. 
// References: 7-zip chm --> Command Line Version --> Switches
// --> -m --> cu=[off | on].
```

PR #17436, opened 2025-08-12
Fix: Fixed issue where the 'layout' icon didn't match the select layout

### LINKED ISSUE #17347: Bug: Toolbar's `Layout` icon doesn't change
### Description

The `Layout` icon in the toolbar isn't changing when the user switches layout

### Steps To Reproduce

1. Open a folder
2. Change the layout using the toolbar's button
3. See that the icon doesn't change

### Files Version

3.9.12.0

### Windows Version

10.0.26100.0

### User ID

86d9d7a3-828c-4e8e-ab92-966b01d872d5

### Log File

https://github.com/user-attachments/assets/b1cdcc37-4390-4204-a3cd-2032cc23fe91

ferrariofilippo:
**Resolved / Related Issues**

- Closes #17347 

**Notes**
- It seems that `Style` changes do not invoke `OnApplyTemplate`, so this solution is basically a workaround.
- I can't remember if we have anything against `Finalizer`s... should I dispose the object in a different way?

1. Open a folder
2. Change the layout using the toolbar's button
3. See that the icon changes
4. Try swapping panes
5. Check if this fixes a similar issue with `Cloud Sync Status` (I can't test this myself)

### DIFF
--- a/src/Files.App.Controls/ThemedIcon/ThemedIcon.Properties.cs
+++ b/src/Files.App.Controls/ThemedIcon/ThemedIcon.Properties.cs
@@ -97,5 +97,22 @@ namespace Files.App.Controls
 		{
 			UpdateVisualStates();
 		}
+
+		private void OnStylePropertyChanged(DependencyObject sender, DependencyProperty dp)
+		{
+			if (dp != StyleProperty)
+				return;
+
+			DispatcherQueue.TryEnqueue(() =>
+			{
+				GetTemplateParts();
+				OnFilledIconChanged();
+				OnOutlineIconChanged();
+				OnLayeredIconChanged();
+				OnIconTypeChanged();
+				OnIconColorTypeChanged();
+				OnIconSizeChanged();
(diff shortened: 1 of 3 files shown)


yair100:
> Check if this fixes a similar issue with Cloud Sync Status (I can't test this myself)

I expect it does 👍 

It likely fixes an issue we were having with the Omnibar as well.

yaira2:
Is the project running on your side?

yair100:
> I can't remember if we have anything against Finalizers... should I dispose the object in a different way?

It seems to be causing a crash, it might be more reliable to use the `Unloaded` event.

ferrariofilippo:
> It seems to be causing a crash, it might be more reliable to use the `Unloaded` event.

My bad, I pushed without testing. Your fix seems to work 👍

yair100:
@ferrariofilippo thank you for fixing this! I didn't think about debugging this from the control side.

PR #17476, opened 2025-08-20
Code Quality: Added support for middle clicking in the new BreadcrumbBar

0x5bfa:
### Resolved / Related Issues

- #16029

1. Open Files app
2. navigate to a nested folder
3. Middle-click on the parent folder
4. See a new tab opens with the parent folder

### DIFF
--- a/src/Files.App.Controls/BreadcrumbBar/EventArgs.cs
+++ b/src/Files.App.Controls/BreadcrumbBar/EventArgs.cs
@@ -3,7 +3,7 @@
 
 namespace Files.App.Controls
 {
-	public record class BreadcrumbBarItemClickedEventArgs(BreadcrumbBarItem Item, int Index, bool IsRootItem = false);
+	public record class BreadcrumbBarItemClickedEventArgs(BreadcrumbBarItem Item, int Index, bool IsRootItem = false, bool IsMiddleButtonPressed = false);
 
 	public record class BreadcrumbBarItemDropDownFlyoutEventArgs(MenuFlyout Flyout, BreadcrumbBarItem? Item = null, int Index = -1, bool IsRootItem = false);
 }
--- a/src/Files.App/UserControls/NavigationToolbar.xaml.cs
+++ b/src/Files.App/UserControls/NavigationToolbar.xaml.cs
@@ -279,7 +279,7 @@ namespace Files.App.UserControls
 				ViewModel.PathComponents[args.Index].Path is not { } path)
 				return;
 
-			await ViewModel.HandleFolderNavigationAsync(path);
+			await ViewModel.HandleFolderNavigationAsync(path, args.IsMiddleButtonPressed);
 		}
 
 		private async void BreadcrumbBar_ItemDropDownFlyoutOpening(object sender, BreadcrumbBarItemDropDownFlyoutEventArgs e)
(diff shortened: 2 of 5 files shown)


src/Files.App.Controls/BreadcrumbBar/EventArgs.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Files.App.Controls
 {
-	public record class BreadcrumbBarItemClickedEventArgs(BreadcrumbBarItem Item, int Index, bool IsRootItem = false);
+	public record class BreadcrumbBarItemClickedEventArgs(BreadcrumbBarItem Item, int Index, bool IsRootItem = false, bool IsMiddleButtonPressed = false);
```
yair100: Instead of passing a Boolean, would it be possible to pass the pointer args directly so that this custom event can do everything the `PointerTapped` event can do?

src/Files.App.Controls/BreadcrumbBar/BreadcrumbBarItem.Events.cs
```diff
@@ -10,7 +10,7 @@ public partial class BreadcrumbBarItem
 	{
 		private void ItemContentButton_Click(object sender, RoutedEventArgs e)
 		{
-			OnItemClicked();
+			//OnItemClicked();
```
yair100: Is this supposed to be commented out?

src/Files.App.Controls/BreadcrumbBar/EventArgs.cs
0x5bfa: Done

src/Files.App.Controls/BreadcrumbBar/BreadcrumbBarItem.Events.cs
0x5bfa: Thought it would be better than removing the event handler to let poeple notice that this is already handled. Should we remove?

src/Files.App/UserControls/NavigationToolbar.xaml.cs
```diff
@@ -279,7 +279,8 @@ private async void BreadcrumbBar_ItemClicked(Controls.BreadcrumbBar sender, Cont
 				ViewModel.PathComponents[args.Index].Path is not { } path)
 				return;
 
-			await ViewModel.HandleFolderNavigationAsync(path);
+			var pointerUpdateKind = args.PointerRoutedEventArgs.GetCurrentPoint(null).Properties.PointerUpdateKind;
+			await ViewModel.HandleFolderNavigationAsync(path, pointerUpdateKind is PointerUpdateKind.MiddleButtonReleased);
```
yair100: Instead of passing `PointerUpdateKind.MiddleButtonReleased` directly, I think we can make the code easier to read if we add another variable like this `var openInNewTab = PointerUpdateKind.MiddleButtonReleased`.

src/Files.App.Controls/BreadcrumbBar/BreadcrumbBarItem.Events.cs
yair100: If it's already handled, it should be removed.

src/Files.App/UserControls/NavigationToolbar.xaml.cs
0x5bfa: Done.

src/Files.App.Controls/BreadcrumbBar/BreadcrumbBarItem.Events.cs
0x5bfa: Done

PR #17487, opened 2025-08-23
Code Quality: Use Satori GC for x64 and arm64 package

hez2010:
This can improve the overall performance, memory footprint and responsiveness.
Note: this only affects the package. It won't kick in during the build process.

I'm maintaining the Satori GC build in a separate repository. 

Also, bump the Windows SDK version to 26100.

### DIFF
--- a/src/Satori.targets
+++ b/src/Satori.targets
@@ -1,7 +1,7 @@
 ﻿<Project>
   <Target Name="FetchSatori" AfterTargets="ResolveRuntimePackAssets">
     <DownloadFile
-        SourceUrl="https://github.com/hez2010/Satori/releases/latest/download/windows_$(Platform).zip"
+        SourceUrl="https://github.com/files-community/Satori/releases/latest/download/windows_$(Platform).zip"
         DestinationFolder="$(IntermediateOutputPath)$([System.Guid]::NewGuid().ToString('N'))"
         Retries="3">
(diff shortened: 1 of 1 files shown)


0x5bfa:
Will we set latency mode for the compaction?

hez2010:
> Will we set latency mode for the compaction?

Maybe not. For the desktop app, we prefer a smaller memory footprint, especially when the latency of interactive mode is already low.

hez2010:
Hmmm based on my observation it seems that it would be better to enable LowLatency mode. Need to verify the memory consumption.

hez2010:
With LowLatency mode the memory grows very aggressively.
I think we need to stick with the default interactive mode.

It's slightly better than WKS GC btw, but not that observable.

yair100:
The package size is 1mb smaller.

src/Satori.targets
```diff
@@ -0,0 +1,44 @@
+﻿<Project>
+  <Target Name="FetchSatori" AfterTargets="ResolveRuntimePackAssets">
+    <DownloadFile
+        SourceUrl="https://github.com/hez2010/Satori/releases/latest/download/windows_$(Platform).zip"
```
yair100:
```suggestion
        SourceUrl="https://github.com/files-community/Satori/releases/latest/download/windows_$(Platform).zip"
```

PR #17490, opened 2025-08-25
Code Quality: Added DisableRuntimeMarshalling to the all trim-safe projects

0x5bfa:
### Resolved / Related Issues

Disabled runtime marshaling in the projects that are already trim-safe in order to prevent people from adding a piece of code that uses runtime marshaling.

- #15000

_N/A_


yair100:
LGTM

PR #17491, opened 2025-08-26
Fix: Command Palette sometimes freezes the UI (#16029)

workbysaran:
**Resolved / Related Issues**

Partially addresses #16029

- Opened the Files app and navigated to one of the available folders.
- Used the Ctrl + Shift + P keyboard shortcut to open the command palette.
- Before the command appeared, performed other actions such as maximize, resize the window, or right-click.
- Verified that the window did not freeze and resizing still worked as expected.


yaira2 (approved):
LGTM

PR #17564, opened 2025-09-07
Code Quality: Sync the jump list with Explorer

### LINKED ISSUE #14526: Bug: JumpList Not Populating
### Description

Right clicking of Files on the taskbar does not show favorites in the jump list.

image

### Steps To Reproduce

1. Install Files (either from Store or classic installer)
2. Right click on Files in the taskbar
3. Don't see favorites because the jump list didn't populate

(I'm experiencing this on 3 different machines (2 Win10, 1 Win11) and trying both Store and Classic install methods)

### Requirements

Right clicking in taskbar should show favorites in jump list

### Files Version

3.1.1.0

### Windows Version

10.0.19045.3930

### Log File

[debug.log](https://github.com/files-community/Files/files/14016962/debug.log)

### LINKED ISSUE #17488: Code Quality: Copy jump list of Explorer to Files via IAutomaticDestinationList
### Description

With the WinRT API, `JumpList`, we are not capable of having proper collections of jump list items because it heavily wraps the base documented functionalities. Even with the documented Win32 APIs, they heavily narrow down the actual functionalities provided by the undocumented APIs (`IAutomaticDestinationList` & `IInternalCustomDestinationList`). These undocumented APIs have not changed since Windows 7 up until Windows 11 and thus I'm pretty sure that the APIs are stable enough to use in Files.

### Concerned code

- https://github.com/files-community/Files/blob/main/src/Files.App/Services/Windows/WindowsJumpListService.cs

### Gains

- Better UX on the jump list.
- Unlimited control over the jump list.

### Requirements

- Remove the existing class for the jump list, `WindowsJumpListService`.
- Add a whole new manager class under `Files.App.Storage`.

### Comments

I made a POC in https://github.com/0x5bfa/JumpListViewer

### LINKED ISSUE #17631: Feature: Files is pinned to my taskbar. When I right click on it, I'd  like to see my quick access pins and pins I make

Show quick access pins via the taskbar right-click menu when Files is pinned to the toolbar.

### Requirements

I have a lot of quick-access pins I use. I'd like to be able to access them from right-clicking my taskbar 

### Files Version

4.0.0.0 

### Windows Version

10.0.26100.6584

### Comments

_No response_

### LINKED ISSUE #17659: Feature: Show actual folder icon in jump list

In the jump list shown when right-clicking the Files icon on the taskbar, display the correct folder icons instead of always using the default icon.

### Requirements

### Files Version

4.0.4.0

### Windows Version

10.0.26120.6682

### Comments

_No response_

0x5bfa:
### Resolved / Related Issues

- Related: #15000
- Closes #17488
- Fixes #14526
- Fixes #17659
- Fixes #17631

### TODOs

- [x] Completely deletes the binaries for the Files' jump list on startup and on every update
- [x] Watches the changes to the Explorer's jump list
- [x] Watches the changes to the Files' jump list
- [x] Replicates the Explorer's jump list to the Files' jump list
- [x] Replicates the Files' jump list to the Explorer's jump list

- Build the app
- Cold-launch the app
- See its jump list being the same as the Explorer's jump list
- Pin/unpin an item from the Explorer's jump list
- See that of Files also gets updated
- Do the way around of above

image

image


src/Files.App.BackgroundTasks/UpdateTask.cs
yair100, line 1: Is this code used at all? If not, we can remove it.

src/Files.App/Assets/FolderIcon.png
yair100, line 1: What is this?

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4325,4 +4319,7 @@
   <data name="FailedToOpenLogFile" xml:space="preserve">
     <value>Unable to open the log file</value>
   </data>
+  <data name="Recent" xml:space="preserve">
```
yair100: Renaming the string will require users to retranslate it. I would leave the name as is unless there is a reason to change it.

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -163,7 +161,6 @@ showDialog is DeleteConfirmationPolicies.PermanentOnly &&
 
 			// Execute removal tasks concurrently in background
 			var sourcePaths = source.Select(x => x.Path);
-			_ = Task.WhenAll(sourcePaths.Select(jumpListService.RemoveFolderAsync));
```
yair100: Does this code need a replacement from our end?

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -228,8 +227,6 @@ public async Task SetWorkingDirectoryAsync(string? value)
 
 			if (value == "Home" || value == "ReleaseNotes" || value == "Settings")
 				currentStorageFolder = null;
-			else
-				_ = Task.Run(() => jumpListService.AddFolderAsync(value));
```
yair100: Does this need a replaement from our end?

src/Files.App/Assets/FolderIcon.png
0x5bfa: This was used in the jump list folder icon. we no longer use it as we use the original icons.

src/Files.App.BackgroundTasks/UpdateTask.cs
0x5bfa: Yeah used.

https://github.com/files-community/Files/blob/b549af2e151d5c054ed464e47efae0fa3e31e132/src/Files.App%20(Package)/Package.appxmanifest#L141

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
0x5bfa: I do not think so. We do not want track jump list by ourselves, we just sync with Explorer and if you use the shell operation API, Explorer updates its jump list anyway.

src/Files.App/ViewModels/ShellViewModel.cs
0x5bfa: As above, no.

src/Files.App/Strings/en-US/Resources.resw
0x5bfa: Reverted

src/Files.App/ViewModels/ShellViewModel.cs
yair100: I think that if we don't manually add folders when opening them in Files, they won't get added to the frequent access list.

src/Files.App/Strings/en-US/Resources.resw
yair100: Thanks

src/Files.App.BackgroundTasks/UpdateTask.cs
yair100: We may need to put something in place to prevent stale jumplists after updating Files or Windows.

src/Files.App/ViewModels/ShellViewModel.cs
0x5bfa: Ah right, but do we really want to add a folder just because you browsed it? In Explorer one gets added only when you use one of some verbs (`open`, for example) on a file. If you just keep clicking recursively, then all of them get added, it may be excessive.

yair100: In my opinion, yes. It's more useful when recently opened folders are actually displayed.

0x5bfa: That case, to keep the synchronization, we should update Explorer's jump list, then the watcher (I will create in the next PR) would notice it and update Files' one too.
What you think?

src/Files.App.BackgroundTasks/UpdateTask.cs
0x5bfa: We can do the same as what we do at the startup, clear all and fetch items from Explorer.
Should we do it?

src/Files.App/ViewModels/ShellViewModel.cs
yair100: Sounds good. That said, I would prefer it if the watcher was included in this PR as well so that we don't have any gap in functionality.

0x5bfa: Oh I totally thought I should separate into smaller PRs for easier review. But it's just adding a watcher class so should be easy to review either way. I'll implement it in this PR today.

src/Files.App.BackgroundTasks/UpdateTask.cs
yair100: Does the update task run even if Files is closed? In that case, I think we should do this.

src/Files.App/ViewModels/ShellViewModel.cs
yair100: Thanks!

src/Files.App.BackgroundTasks/UpdateTask.cs
0x5bfa: I think yeah.

0x5bfa: I think we should make sure to reset our previous jump list.

src/Files.App/ViewModels/ShellViewModel.cs
0x5bfa: Done.

src/Files.App.BackgroundTasks/UpdateTask.cs
0x5bfa: Now when the app is updated, the old jump list will be deleted and re-fetched from Explorer.

yair100:
This might help with #14526

Josh65-2201:
Would this fix https://github.com/files-community/Files/issues/17659

yair100:
> Would this fix #17659

I think so. @0x5bfa can you confirm?

0x5bfa:
Yes. This is one of the motivation. (I updated the PR to include this)

yair100:
1. Pinning or unpinning an item from the jumplist doesn't update the sidebar until I restart Files.
2. If the jumplist is modified while the app is closed, opening Files will revert the changes.

src/Files.App.BackgroundTasks/UpdateTask.cs
```diff
@@ -34,30 +32,18 @@ private void DeleteLogFiles()
...
+			// Make sure to delete the Files' custom destinations binary files
+			var recentFolder = WindowsStorableHelpers.GetRecentFolderPath();
+			File.Delete($"{recentFolder}\\CustomDestinations\\3b19d860a346d7da.customDestinations-ms");
+			File.Delete($"{recentFolder}\\CustomDestinations\\1265066178db259d.customDestinations-ms");
+			File.Delete($"{recentFolder}\\CustomDestinations\\8e2322986488aba5.customDestinations-ms");
+			File.Delete($"{recentFolder}\\CustomDestinations\\6b0bf5ca007c8bea.customDestinations-ms");
+			File.Delete($"{recentFolder}\\AutomaticDestinations\\3b19d860a346d7da.automaticDestinations-ms");
+			File.Delete($"{recentFolder}\\AutomaticDestinations\\1265066178db259d.automaticDestinations-ms");
+			File.Delete($"{recentFolder}\\AutomaticDestinations\\8e2322986488aba5.automaticDestinations-ms");
+			File.Delete($"{recentFolder}\\AutomaticDestinations\\6b0bf5ca007c8bea.automaticDestinations-ms");
```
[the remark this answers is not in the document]
0x5bfa: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/files-community/Files/pull/17564#discussion_r2740614965)

0x5bfa:
Copilot review is so spammy :(

Anyway, all done. PTAL.

yair100:
The list of pinned items in the sidebar no longers updates when I make a change from File Explorer.

yair100:
After the most recent commit, any changes to the jumplist causes the app to freeze.

yair100:
@0x5bfa what's the current status of this PR? Do you think it can be split into several smaller PRs?

PR #17611, opened 2025-09-15
Fix: Fixed crash that sometimes occured when opening pinned items 

### LINKED ISSUE #17539: Bug: Clicking pinned item while tab is loading causes app to crash
### Description

I've always had occasional crashes when opening a new tab or rearranging tabs, but it has been happening really frequently since updating to 4.0.0.0.

### Steps To Reproduce

1. Open a new tab
2. Click a folder on the pinned list - EDIT 2025/9/8
3. Files freezes for a second or two - EDIT 2025/9/8
4. Files crashes

#### Other steps
1. Click a folder on the pinned list - EDIT 2025/9/8
2. Trys to rearrange tabs
3. Files freezes for a second or two - EDIT 2025/9/8
3. Files crashes

### Files Version

4.0.0.0

### Windows Version

10.0.22631.5768

### User ID

402463a3-d7d9-47cb-9146-35f01189c9ce

### Log File

[debug.log](https://github.com/user-attachments/files/22140886/debug.log)

workbysaran:
**Resolved / Related Issues**

Closes #17539 
Fixed the crash by ensuring the WidgetItems is accessed only on the UI thread.

1. Open the Files app
2. Add a new tab
3. Immediately click on a pinned item in the left panel before the tab completes loading
4. Verify that no exception is thrown and the folder opens correctly
5. Verify that normal tab navigation and pinned item access work as expected

### DIFF
--- a/src/Files.App/ViewModels/HomeViewModel.cs
+++ b/src/Files.App/ViewModels/HomeViewModel.cs
@@ -105,7 +105,7 @@ namespace Files.App.ViewModels
 
 		public void RefreshWidgetList()
 		{
-			for (int i = 0; i < WidgetItems.Count; i++)
+			for (int i = WidgetItems.Count - 1; i >= 0; i--)
 			{
 				if (!WidgetItems[i].WidgetItemModel.IsWidgetSettingEnabled)
 					RemoveWidgetAt(i);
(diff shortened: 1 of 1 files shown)


src/Files.App/ViewModels/HomeViewModel.cs
```diff
@@ -105,7 +105,7 @@ private void ReloadWidgets()
 
 		public void RefreshWidgetList()
 		{
-			for (int i = 0; i < WidgetItems.Count; i++)
```
0x5bfa, line 108: Why is this reversed?

workbysaran: Backward iteration avoids skipped items from index shifts and works safely with DispatcherQueue UI updates.

Issue #17636, opened 2025-09-17
Bug: Cannot open mounted Cryptomator volumes


fabioromeo:
### Description

When I click on a mounted cryptomator volume it asks me to put a user name and password in network credentials. Neither user and windows password works.

<img width="935" height="407" alt="Image" src="https://github.com/user-attachments/assets/5da9e96d-d03d-4c9a-8eaa-aaabf6d9be66" />

### Steps To Reproduce

1. Mount cryptomator volume 
2. Open Files
3. Click on Cryptomator volume

### Files Version

4.0.10.0

### Windows Version

24H2, 26100.6584, Windows 11 Pro

### Log File

[debug.log](https://github.com/user-attachments/files/22391297/debug.log)

Josh65-2201:
Thanks for the report, Does it work in Windows File Explorer?

fabioromeo:
> Thanks for the report, Does it work in Windows File Explorer?

Yes, no issues at all. It also works on OneCommander.

Josh65-2201:
@fabioromeo Can you see if this is an issue in the preview version. https://files.community/download

fabioromeo:
@Josh65-2201 I just checked the preview, it works without issues now!

Josh65-2201:
Great, thanks for checking

Issue #17521, opened 2025-09-03
Bug: Cant open Egnyte


rayyesboulous:
### Description

I used Files app in the past and never had a problem.. I wanted to try the new  version today but i wasent able to open egnyte

### Steps To Reproduce

when i click on the egnyte folder i want to used from the pinned folder i receive an error message: ERROR_BAD_NET_NAME

<img width="417" height="274" alt="Image" src="https://github.com/user-attachments/assets/91c739c4-fa94-4c29-bdfc-3f296200287b" />

### Files Version

4.0.0.0

### Windows Version

Edition	Windows 11 Enterprise Version	24H2 Installed on	‎5/‎14/‎2025 OS build	26120.5751 Experience	Windows Feature Experience Pack 1000.26100.234.0

### User ID

0a95775a-664c-4f6e-88a9-d7dd0c2cf7e8

### Log File

NO LOG FILE

yair100:
Thank you for the report.

Josh65-2201:
Can you confirm this https://www.egnyte.com is what you are using

rayyesboulous:
> Can you confirm this https://www.egnyte.com is what you are using

Yes. Using the egnyte desktop app

yair100:
@rayyesboulous Is this still an issue in the latest preview version? https://files.community/download

rayyesboulous:
Hi Yair,

Thank you very much for the reply. I downloaded the preview app and I can
confirm it works.
But just out of curiosity, why does it appear under network and not cloud
drives?

Josh65-2201:
Thanks for checking, Egnyte mounts it as a network drive and Files has no changes to make it go under cloud drives

rayyesboulous:
Hi YAir,

Thank you for the answer.

I love to use the app but the only thing missing is the drag and drop from
outlook. I know thus is on another ticket but do u have an idea of when
that will be fixed?

Thanks

yair100:
> I love to use the app but the only thing missing is the drag and drop from
outlook. I know thus is on another ticket but do u have an idea of when
that will be fixed?

We haven't found a reliable fix yet. Please subscribe to that issue to be notified of any changes.

PR #17664, opened 2025-09-20
Fix: Added conditional checking for virtual network path

### LINKED ISSUE #17144: Bug: Opening network drive from RaiDrive shows ERROR_BAD_NET_NAME
### Description

I have a network drive J:\ connected via RaiDrive, on the normal Files version it works fine, but on the Files Preview version it doesn't and I get this error message: ERROR_BAD_NET_NAME

Image

### Steps To Reproduce

1. Install RaiDrive
2. Connect a network connection (any) for my part SFTP
3. Open the network drive via Files Preview

### Files Version

3.9.9.0

### Windows Version

10.0.22631.0

### User ID

aadae2f1-0446-46ad-a301-899f73adb44a

### Log File

[debug.log](https://github.com/user-attachments/files/20346569/debug.log)

NicholasChin28:
**Resolved / Related Issues**

- Closes #17144

1. Download and install RaiDrive client application
2. Configure RaiDrive to mount an SFTP server connection via the RaiDrive GUI client
3. Verify the SFTP drive appears in Windows File Explorer as a network location
image

4. Attempt to access the RaiDrive SFTP drive using the Files application
5. Upon step debugging through AuthenticateNetworkShare function, observe that
RaiDrive SFTP paths follow the format \\RaiDrive-{username}\{drive-custom-name}\

image

image

8. Notice that WNetAddConnection3W fails when attempting to resolve the RaiDrive
remoteName with either ERROR_BAD_NET_NAME or ERROR_NO_NET_OR_BAD_PATH

image

9. After applying the fix, successfully access RaiDrive SFTP drives through Files
application without errors

10. Checked with SSHFS to test behavior. 

image

image

image

11. SSHFS connecting to a SFTP server seems fine as it does prompt user to key in credentials and after inputting details the folder is accessible.

### DIFF
--- a/src/Files.App/Services/Storage/StorageNetworkService.cs
+++ b/src/Files.App/Services/Storage/StorageNetworkService.cs
@@ -18,6 +18,14 @@ namespace Files.App.Services
 
 		private readonly static string guid = "::{f02c1a0d-be21-4350-88b0-7367fc96ef3c}";
 
+		// Virtual disk path prefixes that don't work with Windows networking APIs
+		private readonly static string[] VirtualDiskPrefixes =
+		[
+			@"\\RaiDrive-",
+			@"\\cryptomator-vault\",
+			@"\\EgnyteDrive\"
+		];
+
 
 		private ObservableCollection<IFolder> _Computers = [];
 		/// <inheritdoc/>
@@ -227,10 +235,9 @@ namespace Files.App.Services
 
 				}
 
-				// Skip authentication for RaiDrive virtual shares
-				// RaiDrive creates virtual UNC paths that don't work with Windows networking APIs
-				// Format is \\RaiDrive-{user}\{drive-custom-name}\
-				if (path.StartsWith(@"\\RaiDrive-", StringComparison.OrdinalIgnoreCase))
+				// Skip authentication for virtual disk shares
+				// These providers create virtual disk paths that don't work with Windows networking APIs
+				if (VirtualDiskPrefixes.Any(prefix => path.StartsWith(prefix, StringComparison.OrdinalIgnoreCase)))
 				{
 					return true;
 				}


src/Files.App/Services/Storage/StorageNetworkService.cs
```diff
@@ -227,6 +227,14 @@ public async Task<bool> AuthenticateNetworkShare(string path)
 
 				}
 
+				// Skip authentication for RaiDrive virtual shares
+				// RaiDrive creates virtual UNC paths that don't work with Windows networking APIs
+				// Format is \\RaiDrive-{user}\{drive-custom-name}\
+				if (path.StartsWith(@"\\RaiDrive-", StringComparison.OrdinalIgnoreCase))
```
yair100: Should we create a constant for this?

Josh65-2201:
Could this also apply to https://github.com/files-community/Files/issues/17636 based on the site it's a separate program that adds a virtual network drive too

src/Files.App/Services/Storage/StorageNetworkService.cs
NicholasChin28: I have updated the code to use a string array constant for a list of virtual drive providers

NicholasChin28:
I had a look at [#17636 ](https://github.com/files-community/Files/issues/17636) and it seems to have a similar issue. When testing:

- The lock/authenticated status of the virtual drive is controlled from the Cryptomator desktop app
- In the Windows File Explorer, once the drive has been unlocked, it immediately allows access without prompting for login credentials
- I assume this is because its already unlocked so it should not prompt/require the user to re-enter the details


Another potentially related issue is [#17521 ](https://github.com/files-community/Files/issues/17521)

- For this one, it has a similar behavior to RaiDrive in the original comment

NicholasChin28:
I have updated the code to maintain a list of providers that seems to be using a virtual network drive. I'm not sure if we want to include the checks for both Cryptomator and Egnyte in this PR.

src/Files.App/Services/Storage/StorageNetworkService.cs
yair100: Thank you

yair100:
> I have updated the code to maintain a list of providers that seems to be using a virtual network drive. I'm not sure if we want to include the checks for both Cryptomator and Egnyte in this PR.

I think it's fine to include them in the same PR.

yaira2 (approved):
LGTM

CherryDT:
Note: This fails if a public key is used instead of a password.

PR #17669, opened 2025-09-22
Fixed issue where clicking home icon on the Omnibar from Tags panel search results would not navigate to home

workbysaran:
**Resolved / Related Issues**

Fixed issue by modifying NavigationToolbarViewModel to allow home navigation when viewing search results

1. Open the Files app and navigate to one of the Tags.
2. Click the Home button.
3. Verified that it navigated to the home page.
4. Also verified home navigation by navigating to other folders and from search results, and all worked as expected.


yaira2 (approved):
LGTM

PR #17682, opened 2025-09-26
Code Quality: Optimize large folder enumeration performance

workbysaran:
**Resolved / Related Issues**

Fixed performance issues when enumerating large folders that would cause the UI to freeze or become unresponsive during folder scanning operations.

1. Open the Files app and navigate to a directory containing a large number of files (100,000+).
2. Observe the folder enumeration process and UI responsiveness.
3. The UI remains responsive during large folder operations without freezing.


Josh65-2201:
It could be confusing to people that `this folder is empty` is shown for a few seconds while it loading. That doesn't happen in v4.0.11

yair100:
It seems to me that `C:\Windows\WinSxS` takes longer to load with these changes.

PR #17792, opened 2025-10-23
Fixed an issue where opening a deleted folder from the Recycle Bin using Column View displayed the physical file system hierarchy

### LINKED ISSUE #13654: Feature: Streamlined "Open File Location" for Recycle Bin Items in Windows Taskbar Search

When I search for an item in the Windows Taskbar search, and that item is in the Recycle Bin, clicking "Open file location" opens the Recycle Bin but follows the path of the item, which is "C:\$Recycle.Bin\*item's name*". It first opens the drive, then an empty folder, and finally displays what's inside the Recycle Bin. It would be great if, instead of displaying the drive folder, an empty folder, and then the Recycle Bin's contents, it would only display the contents of the Recycle Bin.

### Requirements

Modify the behavior of the "Open file location" function to directly open and display the contents of the Recycle Bin, bypassing intermediate folders.

### Files Version

2.5.35

### Windows Version

10.0.22621.2506

### Comments

Before
Capture d'écran 2023-11-03 184403
After
Capture d'écran 2023-11-03 184453

workbysaran:
**Resolved / Related Issues**
Fixed an issue where opening a deleted folder from the Recycle Bin using Column View displayed the physical file system hierarchy (C:\$Recycle.Bin\user SID folder) instead of treating the Recycle Bin as the root location.

Closes #13654

2. Delete some folders with nested subfolders to create test items (or use existing deleted folders)
4. Switch to Column View layout
5. Right-click on a deleted folder in the Recycle Bin (e.g., "Everything-SDK")
6. Select "Open in new tab" from the context menu
7. Verify that the new tab shows:
  - Column 0: Recycle Bin root (showing all deleted items)
  - Column 1: The selected folder's contents (automatically opened)
8. Close and open the Files app and notice the path always shown as Recycle Bin

### DIFF
--- a/src/Files.App/Actions/Navigation/OpenInNewWindow/BaseOpenInNewWindowAction.cs
+++ b/src/Files.App/Actions/Navigation/OpenInNewWindow/BaseOpenInNewWindowAction.cs
@@ -11,6 +11,7 @@ namespace Files.App.Actions
 		protected IContentPageContext ContentPageContext { get; } = Ioc.Default.GetRequiredService<IContentPageContext>();
 		protected IHomePageContext HomePageContext { get; } = Ioc.Default.GetRequiredService<IHomePageContext>();
 		protected ISidebarContext SidebarContext { get; } = Ioc.Default.GetRequiredService<ISidebarContext>();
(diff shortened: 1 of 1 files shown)


yair100:
This PR has been replaced by #17803.

PR #17803, opened 2025-10-27
Fix: Fixed an issue where Columns View was incorrectly enabled for Recycle Bin

### LINKED ISSUE #13654: Feature: Streamlined "Open File Location" for Recycle Bin Items in Windows Taskbar Search

workbysaran:
**Resolved / Related Issues**
Fixed an issue where using Column View in the Recycle Bin caused unexpected behavior with file system hierarchy display. The Columns layout is now disabled when viewing the Recycle Bin to prevent navigation issues.

Closes #13654

2. Delete some folders to create test items in Recycle Bin
4. Open the Layout options menu from the toolbar
5. Verify that the Columns layout button is grayed out and disabled
6. Try using the keyboard shortcut (Ctrl+Shift+5) and verify it doesn't switch to Columns view
7. Switch to other layouts (Details, List, Cards, Grid) and verify they work normally in Recycle Bin
8. Navigate away from Recycle Bin to any regular folder
9. Verify that the Columns layout button is now enabled and functional again
10. Navigate back to Recycle Bin and confirm Columns layout is disabled again

### DIFF
--- a/src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
+++ b/src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
@@ -486,6 +486,8 @@ namespace Files.App.Helpers
 
 		private static LayoutPreferencesItem? GetLayoutPreferencesForPath(string path)
 		{
+			//Recycle Bin does not support Column View due to navigation conflicts with hierarchical display
+			//Fall back to Details View when Column View is configured
 			if (path.StartsWith(Constants.UserEnvironmentPaths.RecycleBinPath, StringComparison.Ordinal))
 			{
 				var trimmedPath = path.TrimPath() ?? string.Empty;


yair100:
@workbysaran what happens if the user has the 'sync layout preferences' option enabled and has the default layout set to Columns?

workbysaran:
I have updated the LayoutPreferencesManager to handle the scenario when the "Sync layout preferences" setting is enabled.

yair100:
Thanks! Does this PR supersede #17792?

yair100:
I found a scenario where Recycle Bin still uses Columns:
1. Install Files (fresh install)
2. Open Settings
3. Change default layout to Columns (without enabling syncing)
4. Open Recycle Bin
5. Notice how it uses Columns

workbysaran:
> Thanks! Does this PR supersede #17792?

Yes, it does!

src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
```diff
@@ -486,6 +486,28 @@ public static void SetLayoutPreferencesForPath(string path, LayoutPreferencesIte
 
 		private static LayoutPreferencesItem? GetLayoutPreferencesForPath(string path)
 		{
+			if (path.StartsWith(Constants.UserEnvironmentPaths.RecycleBinPath, StringComparison.Ordinal))
```
yair100, line 491: Can this code be reused on line 128 to reduce duplication?

workbysaran: @yaira2 A merge conflict on my end had introduced a duplicate RecycleBinPath condition. I’ve now removed the redundant code.

yaira2 (approved):
LGTM

src/Files.App/Helpers/Layout/LayoutPreferencesManager.cs
yair100, line 491: Can you add a code comment explaining why we disable Columns View in Recycle Bin?

PR #17813, opened 2025-10-30
Fix: Fixed an issue where incorrect actions were displayed in the Recycle Bin context menu

workbysaran:
**Resolved / Related Issues**
Fixed an issue where the context menu displayed inappropriate navigation and action options for items in the Recycle Bin. Since items in the Recycle Bin cannot be properly navigated or used until restored, the following menu items are now hidden when right-clicking Recycle Bin items:
- "Open in New Tab", "Open in New Window", "Open in New Pane" - Hidden for all items in Recycle Bin
- "Create Folder with Selection" - Hidden for items in Recycle Bin
- "Open" - Hidden only for folders in Recycle Bin

2. Delete some files and folders to create test items in Recycle Bin
4. Test folder context menu
   - Right-click on a deleted folder in the Recycle Bin
   - Verify that "Open", "Open in New Tab", "Open in New Window", "Open in New Pane", and "Create Folder with Selection" do not appear in the context menu.
5. Test file context menu
   - Right-click on a deleted file in the Recycle Bin
   - Verify that "Open" is visible and "Create Folder with Selection" do not appear in the context menu.

### DIFF
--- a/src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
+++ b/src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
@@ -479,7 +479,7 @@ namespace Files.App.Data.Factories
 				}.Build(),
 				new ContextMenuFlyoutItemViewModelBuilder(Commands.CreateFolderWithSelection)
 				{
-					IsVisible = UserSettingsService.GeneralSettingsService.ShowCreateFolderWithSelection && itemsSelected && !currentInstanceViewModel.IsPageTypeRecycleBin
+					IsVisible = UserSettingsService.GeneralSettingsService.ShowCreateFolderWithSelection && itemsSelected
 				}.Build(),
 				new ContextMenuFlyoutItemViewModelBuilder(Commands.CreateShortcut)
 				{


yair100:
Thank you for the PR. Would it be possible to set the IsExecutable property from the actions code? Otherwise, these actions will still be available from the Command Palette.

workbysaran:
> Thank you for the PR. Would it be possible to set the IsExecutable property from the actions code? Otherwise, these actions will still be available from the Command Palette.

Thank you for the feedback! I've updated the code to set the IsExecutable property.

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
```diff
@@ -380,7 +380,10 @@ public static List<ContextMenuFlyoutItemViewModel> GetBaseItemMenuItems(
 				{
 					IsVisible = currentInstanceViewModel.IsPageTypeRecycleBin && itemsSelected,
 				}.Build(),
-				new ContextMenuFlyoutItemViewModelBuilder(Commands.OpenItem).Build(),
+				new ContextMenuFlyoutItemViewModelBuilder(Commands.OpenItem)
+				{
+					IsVisible = !(currentInstanceViewModel.IsPageTypeRecycleBin && areAllItemsFolders)
```
yair100: Does this need to be set manually? If I recall correctly, it'll automatically disable if the action is disabled.

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
```diff
@@ -409,15 +412,15 @@ public static List<ContextMenuFlyoutItemViewModel> GetBaseItemMenuItems(
 				new ContextMenuFlyoutItemViewModelBuilder(Commands.OpenFileLocation).Build(),
 				new ContextMenuFlyoutItemViewModelBuilder(Commands.OpenInNewTab)
 				{
-					IsVisible = UserSettingsService.GeneralSettingsService.ShowOpenInNewTab && Commands.OpenInNewTab.IsExecutable
+					IsVisible = UserSettingsService.GeneralSettingsService.ShowOpenInNewTab && Commands.OpenInNewTab.IsExecutable && !currentInstanceViewModel.IsPageTypeRecycleBin
```
yair100: Same as above.

src/Files.App/Actions/FileSystem/OpenItemAction.cs
```diff
@@ -27,7 +27,9 @@ public HotKey HotKey
 		public bool IsExecutable =>
 			context.HasSelection &&
 			!(context.ShellPage is ColumnShellPage &&
-			context.SelectedItem?.PrimaryItemAttribute == StorageItemTypes.Folder);
+			context.SelectedItem?.PrimaryItemAttribute == StorageItemTypes.Folder) &&
+			!(context.PageType == ContentPageTypes.RecycleBin &&
+			context.SelectedItems.All(x => x.IsFolder));
```
yair100: Was `context.SelectedItems.All(x => x.IsFolder))` added to the logic intentionally?

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
workbysaran: The latest changes to IsExecutable in the action classes supersede the previous manual IsVisible checks . Removed the redundant code accordingly.

src/Files.App/Actions/FileSystem/OpenItemAction.cs
workbysaran: Upon further checking, I found that NavigationHelpers prevents the Open and Open With actions from executing in the Recycle Bin. Also, I noticed that Pin to Sidebar is not working as expected, and I believe it should not be shown for items within the Recycle Bin. What do you think?

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
yair100: Thank you. By the way, would it be possible to leave the GitHub conversations unresolved until I have a chance to view them? This way it will be easier to follow your response 🙂

src/Files.App/Actions/FileSystem/OpenItemAction.cs
yair100: > Also, I noticed that Pin to Sidebar is not working as expected, and I believe it should not be shown for items within the Recycle Bin. What do you think?

Agreed.

src/Files.App/Data/Factories/ContentPageContextFlyoutFactory.cs
```diff
@@ -479,7 +479,7 @@ public static List<ContextMenuFlyoutItemViewModel> GetBaseItemMenuItems(
 				}.Build(),
 				new ContextMenuFlyoutItemViewModelBuilder(Commands.CreateFolderWithSelection)
 				{
-					IsVisible = UserSettingsService.GeneralSettingsService.ShowCreateFolderWithSelection && itemsSelected
+					IsVisible = UserSettingsService.GeneralSettingsService.ShowCreateFolderWithSelection && itemsSelected && !currentInstanceViewModel.IsPageTypeRecycleBin
```
yair100: Same as above

workbysaran: You are correct! The check is redundant now and I've removed it.

PR #17819, opened 2025-11-04
fix: allow Ctrl+C to copy selected text in Details Pane

Rahul2322-P:
Fixed the bug where Ctrl+C was copying the file instead of the selected text in the Details Pane.
Changed args.Handled = true to args.Handled = false in OnCopyInvoked().
Now, when a text field is focused, Ctrl+C copies the text as expected.


yair100:
@Rahul2322-P thank you for the PR. I'm not seeing how this solves the issue, can you please clarify?

yair100:
This PR is superseded by #17825

PR #14101, opened 2023-11-28
Fix: Fixed issue where batch files were opened as system32

yaira2:
**Resolved / Related Issues**
- [ ] Were these changes approved in an issue or discussion with the project maintainers? In order to prevent extra work, feature requests and changes to the codebase must be approved before the pull request will be reviewed. This prevents extra work for the contributors and maintainers.
   Closes #12139...

**Validation**
How did you test these changes?
- [ ] Did you build the app and test your changes?
- [ ] Did you check for accessibility? You can use Accessibility Insights for this.
- [ ] Did you remove any strings from the en-us resource file?
   - [ ] Did you search the solution to see if the string is still being used? 
- [ ] Did you implement any design changes to an existing feature?
   - [ ] Was this change approved?
- [ ] Are there any other steps that were used to validate these changes?
   1. Open batch file and confirm it runs with the correct username

**Screenshots (optional)**
Add screenshots here.

Issue #12139, opened 2023-04-21
Bug: Batch files are executed as SYSTEM user


Cromagnon31:
### Description

When double clicking on a batch file or an executable, it's executed as SYSTEM user.

USERNAME=SYSTEM
USERPROFILE=C:\Users\OW

When you open a shell on a directory is is executed as the session user.

USERNAME=OW
USERPROFILE=C:\Users\OW

USERPROFILE is good but not USERNAME !

### Steps To Reproduce
0. log in Windows session as standard user (eg not SYSTEM user) and open Files.

1 Edit a batch files with the 2 lines :
SET
PAUSE
Open(double click) the batch file.
Look at USERNAME variable, it's SYSTEM
or 

2 Open(double click) %windir%\system32\cmd.exe and run set command in the newly opened shell.
Look at USERNAME variable, it's SYSTEM

3. Open a shell and run the batch file or set command
Look at USERNAME it it's the logged user name.

### Requirements

When opening a batch file  or an executable it should run in the user environmement eg environment, variable USERNAME should be equal to the logged user name, not SYSTEM.

### Files Version

2.4.62.0

### Windows Version

Window 11 pro 22H2 22261.1555

### Log file

[debug.log](https://github.com/files-community/Files/files/11296129/debug.log)

Josh65-2201:
Thanks for the report. Can you give the following information.
- Are you running the bat files as admin?

yaira2:
@Cromagnon31 can you confirm if this is the same as #5648?

Cromagnon31:
No, I just click on it.

If I open a shell at the directory place, the user is the session user.

 

Strange isn't it ? 

 

 

De : Josh65-2201 ***@***.*** 
Envoyé : vendredi 21 avril 2023 15:32
À : files-community/Files
Cc : OlivierW; Author
Objet : Re: [files-community/Files] Batch files are executed as SYSTEM user (Issue #12139)

 

Thanks for the report. Can you give the following information.

*	Are you running the bat files as admin?

Cromagnon31:
No. Quite a bit different I guess.

Batch is executed normally (ie not as administrator) from data storage outside Windows directories.  

 

 

De : Yair ***@***.*** 
Envoyé : vendredi 21 avril 2023 15:37
À : files-community/Files
Cc : OlivierW; Mention
Objet : Re: [files-community/Files] Batch files are executed as SYSTEM user (Issue #12139)

 

[...]

PR #17872, opened 2025-11-16
Code Quality: Skip USERNAME env var when launching processes

yair100:
**Resolved / Related Issues**

This is a new approach to solving #12139 and replaces the solution implemented in #14101. This PR adds back the code removed in #14101 and instead skips the USERNAME environment variable when getting the Machine variables.

### DIFF
--- a/src/Files.App/Utils/Shell/LaunchHelper.cs
+++ b/src/Files.App/Utils/Shell/LaunchHelper.cs
@@ -105,7 +105,7 @@ namespace Files.App.Utils.Shell
 						string key = (string)ent.Key;
 
 						// Skip USERNAME to avoid issues where files were executed as SYSTEM user (#12139)
-						if (key == "USERNAME")
+						if (string.Equals(key, "USERNAME", StringComparison.OrdinalIgnoreCase)) 
 							continue;
 
 						process.StartInfo.EnvironmentVariables[key] = (string)ent.Value;


src/Files.App/Utils/Shell/LaunchHelper.cs
```diff
@@ -100,6 +100,17 @@ private static async Task<bool> HandleApplicationLaunch(string application, stri
 					process.StartInfo.Arguments = arguments;
 
 					// Refresh env variables for the child process
+					foreach (DictionaryEntry ent in Environment.GetEnvironmentVariables(EnvironmentVariableTarget.Machine))
+					{
+						string key = (string)ent.Key;
+
+						// Skip USERNAME to avoid issues where files were executed as SYSTEM user (#12139)
+						if (key == "USERNAME")
```
0x5bfa: Use the ordinal case ignoring string comparison.

0x5bfa: `==` differentiates A and a but OrdinalIgnoreCase doesn't.

0x5bfa: just in case

yair100: As far as I know, this variable is always uppercase.

yair100: I pushed a commit

PR #17893, opened 2025-11-21
Fix: Fixed crash that would sometimes occur when resizing the window

### LINKED ISSUE #17739: Bug: Resizing window in Dual Pane mode causes crash
### Description

When dual panes are enabled and you reduce the window size horizontally, it crashes the the app and then reopens itself.

### Steps To Reproduce

1. Open app
2. reduce window horizontally

https://github.com/user-attachments/assets/3ecb65a4-ef5f-4864-af0d-5c9b68cf7e35

### Files Version

4.0.11.0

### Windows Version

Edition	Windows 11 Pro
Version	25H2
OS build	26200.5761
Experience	Windows Feature Experience Pack 1000.26100.234.0

### User ID

45bd3559-e32a-4c9f-b0d2-791fb492a322

### Log File

[debug.log](https://github.com/user-attachments/files/22731730/debug.log)

workbysaran:
**Resolved / Related Issues**
Fixed a crash (DivideByZeroException) that occurred when resizing the window with multiple panes split vertically.

Closes #17739

1. Open Files with dual pane (vertical split)
2. Resize the window to smaller sizes
3. Resize window back to larger size and verify right pane restores with correct folder
4. Test multiple resize cycles and layouts to ensure stability

### DIFF
--- a/src/Files.App/MainWindow.xaml.cs
+++ b/src/Files.App/MainWindow.xaml.cs
@@ -28,7 +28,7 @@ namespace Files.App
 		{
 			InitializeComponent();
 
-			WindowHandle = WinUIEx.WindowExtensions.GetWindowHandle(this); 
+			WindowHandle = WinUIEx.WindowExtensions.GetWindowHandle(this);
 			MinHeight = 316;
 			MinWidth = 416;
 			ExtendsContentIntoTitleBar = true;


yair100:
Thank you for the PR. Can you please provide some details on why you went with this approach? Ideally the minimum size wouldn't have to change, and we would instead fix the underlying cause.

workbysaran:
@yaira2 
The application was crashing with a System.DivideByZeroException in Microsoft's WinUI framework when the window was resized too small with multiple panes displayed vertically. The crash occurs in WinUI's internal layout engine(FrameworkElement.MeasureOverride) when the available window space becomes insufficient to handle the grid's minimum height and width constraints.

In fact, the same issue happens even when using a single pane but the crash threshold is much lower (nearly 0px) compared to the original default size set by the code (316×416). With dual panes, the crash occurs around 700-850px width. So setting MinWidth = 850 and MinHeight = 646 on the main window itself prevents users from resizing below the threshold before WinUI's layout engine runs, avoiding the crash condition entirely.

If you don't want to set the size explicitly, we need to spend more time on research to find out why the internal Microsoft component fails, find out if there is a way to handle the validation ourselves, and confirm whether it's a Microsoft bug and report it accordingly.

Furthermore, we can enhance my previous fix to set the minimum size dynamically based on whether multiple panes are enabled.

Please let me know how you’d like to proceed.

workbysaran:
With further investigation, the DivideByZeroException occurs because the restored pane has no navigation parameters, leaving it uninitialized. When WinUI's layout engine attempts to measure and arrange the empty pane, it encounters invalid state, which can lead to a division-by-zero error during its dimension calculations.

I fixed the issue by saving the navigation parameters and restoring them when the pane is shown again.

yair100:
@workbysaran I don't recall the original behavior (maybe @winston-de or @gave92 remembers) but the other option is to use 'home' when adding back the second pane. Does anyone have a preference?

Issue #17915, opened 2025-11-27
Bug: Renaming file the same as an older deleted one causes new file to be grouped incorrectly


thepistoneer:
### Description

General description of issue: Say I have a file e.g. "file1" in my Downloads folder under the Last week tab and another file e.g. "file2" also in the Downloads folder under another time tab (e.g. Earlier this year). I delete file2 and rename file1 to file2. Instead of file1 (now called file2) remaining under the Last week tab it moves to the Earlier this year tab where the original file2 just was. Obviously annoying because a file I've made changes to in the past week is in a different time period category.

What happened for me specifically (shown in screenshot): I had an Adobe Illustrator file named "Coursework level design.ai" that was under the "Earlier this year" time category in the Downloads folder. I then downloaded a newer version of this file with the same name. To avoid confusion I deleted the older file version, and this caused the newer file to take its place under the "Earlier this year" category even though the date modified was that day.

Shown below is a screenshot of my downloads folder, as you can see under "Earlier this year" the "Coursework level design.ai" file has the date modified as 3 days ago and so shouldn't be in that category.

<img width="948" height="367" alt="Image" src="https://github.com/user-attachments/assets/c7c8e5e3-28f1-49d5-87b4-f0522e848cc0" />

### Steps To Reproduce

- Have a file in your Downloads folder that is under any time category that isn't "Today"
- Download a file that is the same file type as the first file (doesn't have to be a copy of the first file)
- Delete the first file
- Rename the second file to have the same name as the first file
- The second file will be put under whatever time category the first file was in even though the date modified is today and the file should be under "Today"

### Files Version

4.0.11.0

### Windows Version

10.0.19045.6466

### User ID

d13aeb6a-f925-40b6-ab9c-1bda6579c9c9

### Log File

[debug.log](https://github.com/user-attachments/files/23809181/debug.log)

yair100:
Thank you for the report.

Mythindia:
Created PR #17920  . The fix was to Register property changed handler to react to date changes so the item can be moved and Unregister handlers for all items before clearing. Also I created Unit Test case for testing the fix. Please review and approve.

thepistoneer:
Still having the issue but not sure if that's because I need to do anything other than update the program (now on version 4.0.21.0 and issue persists). Please let me know if I need to do anything else, thanks!

yair100:
[...]

PR #17920, opened 2025-11-29
Fix: Renaming file the same as an older deleted one causes new file to be grouped incorrectly

Mythindia:
**Resolved / Related Issues**

- Closes   [#17915](https://github.com/files-community/Files/issues/17915)

Fix - Register property changed handler to react to date changes so the item can be moved and Unregister handlers for all items before clearing.

Run Unit Test case or follow repro steps as per bug
UT available in 
1. tests/Files.App.UnitTests/BulkConcurrentObservableCollectionTests.cs


yair100:
@Mythindia thank you for the PR. It looks like there are some unrelated changes to the sln file. Can you please revert the unrelated changes?

Mythindia:
Thank you for the quick check and noticing the issue, Unrelated changes have been reverted. Please check.

yair100:
@Mythindia it looks like the CI pipelines are failing to compile this PR. Can you take a look?

Mythindia:
The Test files have been deleted which may have caused compilation issue of the CI pipeline. Prioritizing only the bug fix and relevant changes with this PR.

yair100:
@Mythindia it seems that this branch doesn't compile in Visual Studio. Can you confirm if you were able to compile it on your end?

yair100:
@Mythindia as per https://github.com/files-community/Files/issues/17915#issuecomment-3630885770, it looks like the current behavior is working as expected.

PR #17942, opened 2025-12-08
Feature: Added default name when creating new files

### LINKED ISSUE #17845: Feature: Creating a folder should have `new folder` as name by default

**Translated**
It's inconvenient when creating new folders, documents, etc.

**Original**
新建文件夹、文档等的时候不方便

### Requirements

**Translated**
Creating new folders, documents, etc., is inconvenient because it requires entering a name each time, unlike Windows Explorer where you can create them first and modify the filename later. When you don't want to enter a filename, it feels cumbersome.

**Original**
新建文件夹、文档等的时候不方便，每次要求输入项目名称，而不是像windows资源管理器一样直接新建了然后留文件名待修改，不想输入文件名的时候就感觉很繁琐。

### Files Version

4.0.15.0

### Windows Version

10.0.19045.6456

### Comments

_No response_

marlenhalvorsen:
Issue:
Closes #17845

Changes:
- Adds subtitle "Enter an item name" in create-item dialog
- Sets default name to "New folder" when creating folder
- Selects name by default to allow fast rename

Test steps: 
1. Launch Files (debug build)
2. Create new folder via Ctrl+Shift+N
3. Check that:
   - The dialog shows subtitle "Enter an item name"
   - "New folder" is prefilled
   - Text is fully selected on open

### DIFF
--- a/src/Files.App/Helpers/Dialog/DynamicDialogFactory.cs
+++ b/src/Files.App/Helpers/Dialog/DynamicDialogFactory.cs
@@ -55,7 +55,7 @@ namespace Files.App.Helpers
 			return dialog;
 		}
 
-		public static DynamicDialog GetFor_CreateItemDialog(string itemType)
+		public static DynamicDialog GetFor_CreateItemDialog(string itemType, string? itemName)
 		{
 			DynamicDialog? dialog = null;
 			TextBox inputText = new()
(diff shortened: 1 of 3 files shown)


yair100:
@marlenhalvorsen it looks like this is missing the `Enter an item name` subtitle. Can you adjust this?

marlenhalvorsen:
Subtitle added. “Enter an item name” now appears correctly in the create-item dialog.

yair100:
Thanks! Should we apply this change for other file formats as well?

marlenhalvorsen:
Thanks for the suggestion!

I've looked into the handling of `itemType` in `GetFor_CreateItemDialog`, and currently the dialog only receives two categories: `"Folder"` and `"File"`. All file templates (e.g. .pptx, .xlsx, .txt) are mapped to `"File"` without any specific extension or template metadata.

Because of this, only the folder case can receive a unique default label at the moment. Differentiating per file type (e.g. "New .pptx", "New .xlsx") would require passing additional ShellNewEntry/extension information into the dialog.

yaira2:
Thanks for checking. We can merge as is and can reconsider if we receive more feedback regarding other file formats.

yaira2 (approved):
LGTM

yair100:
@marlenhalvorsen thanks again for the PR. I know we discussed leaving this as is, but I decided to add support for other file formats for consistency.

Issue #17701, opened 2025-09-29
Bug: App crashes frequently


kiromikhaeel:
### Description

When the app stays idle for a little bit and I click on the window, it crashes and closes sometimes

### Steps To Reproduce

1. open app and browse somewhere
2. move on to a different app to do something
3. click on the windows again 

### Files Version

4.0.11.0

### Windows Version

10.0.26100.6584

### User ID

dacee906-5239-4788-bcdd-9754ee380c1e

Josh65-2201:
Thanks for the report, Can you please upload the debug.log file (DON'T copy and paste the text) directly to GitHub (email attachments don't work).

kiromikhaeel:
[debug.log](https://github.com/user-attachments/files/22644881/debug.log)

Josh65-2201:
Thanks. The log doesn't show any crash errors, can you
1. Type Event Viewer into the Windows start menu search 
2. Go to Windows logs > Application 
3. Screenshot the first `Application Error` under the source column with this symbol ![Image](https://github.com/user-attachments/assets/ca054409-fd2b-46f0-bafc-72fa40edc741) that has Files.exe shown on the first line 
4. Upload this on GitHub comment box (Email attachments don't work)

armandn:
I am also getting lots of crashes. I tried the 3.x branch, now 4.011. Application log says:
```
Faulting application name: Files.exe, version: 1.0.0.0, time stamp: 0x68760000
Faulting module name: ntdll.dll, version: 10.0.26100.5074, time stamp: 0x0e967e89
Exception code: 0xc0000374
Fault offset: 0x000000000011a3b5
Faulting process id: 0x6490
Faulting application start time: 0x1DC350FC819D3B0
Faulting application path: C:\Program Files\WindowsApps\49306atecsolution.FilesUWP_4.0.11.0_x64__et10x9a9vyk8t\Files.App\Files.exe
Faulting module path: C:\WINDOWS\SYSTEM32\ntdll.dll
Report Id: 117810a1-9850-48b6-ab82-a6e08ffa2745
Faulting package full name: 49306atecsolution.FilesUWP_4.0.11.0_x64__et10x9a9vyk8t
Faulting package-relative application ID: App
```

The app crashes by itself, without any user input, after a few minutes after launch. It crashes every time, making it unusable.

Josh65-2201:
[...]

PR #17955, opened 2025-12-11
Code Quality: Additional logs to troubleshoot crashes

workbysaran:
**Resolved / Related Issues**
Added additional debug logging to help troubleshoot a crash that occurs when switching apps. The issue cannot be reproduced locally, so these logs will help diagnose the problem on affected user machines.

Related to #17701

1. Launch the Files app, enable the Preview Pane
2. Navigate to multiple files and switch to other applications using Alt+Tab
3. Switch back to Files app and verify no crash occurs
4. Check debug logs for proper logging of window activation states

### DIFF
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -2796,7 +2796,7 @@ namespace Files.App.ViewModels
 		public void Dispose()
 		{
 			CancelLoadAndClearFiles();
-			App.Logger.LogInformation($"ShellViewModel.Dispose: CurrentFolder={LogPathHelper.GetDirectoryName(CurrentFolder?.ItemPath)}");
+			App.Logger.LogInformation($"ShellViewModel.Dispose: CurrentFolder={LogPathHelper.GetPathIdentifier(CurrentFolder?.ItemPath)}");
 
 			StorageTrashBinService.Watcher.ItemAdded -= RecycleBinItemCreatedAsync;
 			StorageTrashBinService.Watcher.ItemDeleted -= RecycleBinItemDeletedAsync;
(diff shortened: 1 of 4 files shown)


src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -2792,6 +2796,8 @@ public void UpdateDateDisplay(bool isFormatChange)
 		public void Dispose()
 		{
 			CancelLoadAndClearFiles();
+			App.Logger.LogInformation($"ShellViewModel.Dispose: CurrentFolder={CurrentFolder?.ItemPath}");
```
yair100: Is there a way to only save the folder name without the complete path?

workbysaran: I have updated the code to log only the file/directory name

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -2792,6 +2796,8 @@ public void UpdateDateDisplay(bool isFormatChange)
 		public void Dispose()
 		{
 			CancelLoadAndClearFiles();
+			App.Logger.LogInformation($"ShellViewModel.Dispose: CurrentFolder={LogPathHelper.GetDirectoryName(CurrentFolder?.ItemPath)}");
```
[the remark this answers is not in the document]
workbysaran: The namespace Files.App.Helpers is already included in GlobalUsings.

src/Files.App/ViewModels/ShellViewModel.cs
yair100: @workbysaran Upon further thought, is there a way to simply use a placeholder? It would be best to avoid any personal details in the log file.

workbysaran: I have changed it to use an MD5 hash. It does not expose the file or folder details, but it is sufficient to verify whether it is the same file and to identify the file extension.

yair100:
@workbysaran thank you.

PR #17991, opened 2025-12-24
Feature: Add disk based thumbnail caching system

workbysaran:
## Resolved / Related Issues
Related to #4180

Addresses performance issues caused by repeated Shell API thumbnail generation by introducing a disk based LRU cache and a centralized ThumbnailService.

1. Navigate to a folder with mixed file types (images, PDFs, etc.) and verify that thumbnails are generated and cached
2. Scroll down and then back up, and verify that thumbnails load instantly from the cache
3. Navigate away from the folder and return to it, and verify that thumbnails load instantly from the cache
4. Restart the app, navigate to the same folder, and verify that thumbnails load instantly from the cache
5. Modify an image file and verify that the thumbnail is regenerated
6. Switch between different views and verify that thumbnails are generated correctly for each view

TODO
- Fix issue where small thumbnails are being cached even when larger thumbnails are requested
- Fix issue where folder thumbnails are not being cached properly in OneDrive


0x5bfa (commented):
Changes are looking good as far as I can tell as of now. This code will still remain for the time being but we should use [the built-in windows API to get thumbnail cache](https://learn.microsoft.com/windows/win32/api/thumbcache/nn-thumbcache-ithumbnailcache) in the future. (Looks like IThumbnailProvider in the new Storage Abstraction API WindowsStorableHelpers.GetThumbnail already calls, nice)

src/Files.App/Data/Contracts/IThumbnailCache.cs
yair100, line 1: @workbysaran please add docs for the tasks.

src/Files.App/Utils/Storage/Helpers/FileThumbnailHelper.cs
```diff
@@ -1,12 +1,20 @@
...
 // Licensed under the MIT License.
 
+using Files.App.Data.Contracts;
 using Windows.Storage.FileProperties;
 
 namespace Files.App.Utils.Storage
 {
 	public static class FileThumbnailHelper
 	{
+		private static IThumbnailService? _thumbnailService;
```
yair100:
```suggestion
		private static IThumbnailService thumbnailService = Ioc.Default.GetRequiredService<IThumbnailService>();
```

src/Files.App/App.xaml.cs
```diff
@@ -111,6 +111,9 @@ async Task ActivateAsync()
 				Logger = Ioc.Default.GetRequiredService<ILogger<App>>();
 				AppModel = Ioc.Default.GetRequiredService<AppModel>();
 
+				var thumbnailService = Ioc.Default.GetRequiredService<IThumbnailService>();
+				FileThumbnailHelper.Initialize(thumbnailService);
```
yair100: Does this need to be initialized here or can it be done from `InitializeAppComponentsAsync` in `AppLifecycleHelper.cs`?

workbysaran: Agreed. This has been moved to AppLifecycleHelper.

src/Files.App/Data/Contracts/IThumbnailCache.cs
workbysaran: I have added xml documentation

src/Files.App/App.xaml.cs
yair100: Is there a reason to pass the service into FileThumbnailHelper instead of just using regular dependency injection?

workbysaran: Initially thought explicit init would give more control for a static class. But you're right , this pattern is already used in ContentPageContextFlyoutFactory. Checked the flow & DI is configured well before any access, so no timing issues. Updated the code.

mdtauk:
Will we need a button in settings to rebuild the thumbnail cache?

Issue #17970, opened 2025-12-17
Code Quality: Implement SidebarView2


0x5bfa:
### Description

The current sidebar implementation has several issues, which is simply because we ported the controls over to custom controls and was fairly a hard work.

1. The whole instance of ViewModel is passed to the control
2. The control uses `Binding`-s in the styles, which is AOT-incompatible
3. There's a dedicated view model for the user control, `SidebarView` with 1.3k loc
4. There's an interface for the view model
5. There's an interface for data template of the `SidebarItem`
6. Item clicking event is heavily tied to the view model
7. Item changed event is heavily tied to the related services like `QuickAccessService`

Here's the proposed version of the sidebar.

1. The control implementation design will be similar to that of `NavigationView`
2. The `SidebarViewModel` will be removed
3. Bindings will be done in `MainPageViewModel`
4. The control will make use of events for dragging and clicking
5. The view model will make use of events for data source from e.g. `QuickAccessService`
6. The control will allow to expand folders toward the tree view proposal

Here's the API spec:

- **SidebarView2**
  - prop: `SidebarViewDisplayMode2 DisplayMode`
  - prop: `bool IsPaneOpen`
  - prop: `object MenuItemsSource`
  - prop: `object FooterItemsSource`
  - prop: `double OpenPaneLength`
  - prop: `object SelectedItem`
- **SidebarView2SectionItem**
  - prop: `object Icon`
  - prop: `object Text`
  - prop: `IList<object> Items`
- **SidebarView2Item**
  - prop: `bool CanDrag` = true
  - prop: `bool AllowDrop` = true
  - prop: `object Icon`
  - prop: `object Text`
  - prop: `bool IsExpanded`
  - prop: `bool CanExpand`
  - prop: `object Decorator`
  - prop: `IList<object> Items`

Here's an example of usage:

```xml
<controls:SidebarView2
    DisplayMode="{x:Bind ViewModel.SidebarView2DisplayMode, Mode=TwoWay}"
    IsPaneOpen="{x:Bind ViewModel.SidebarView2IsPaneOpen, Mode=TwoWay}"
    MenuItemsSource="{x:Bind ViewModel.SidebarView2SectionItems, Mode=OneWay}"
    FooterItemsSource="{x:Bind ViewModel.SidebarView2FooterItems, Mode=OneWay}"
    OpenPaneLength="{x:Bind ViewModel.SidebarView2OpenPaneLength, Mode=OneWay}"
[...]

PR #17992, opened 2025-12-24
Code Quality: Removed ISidebarItemModel 1

0x5bfa:
### Resolved / Related Issues

This is part of the phase 1 mentioned in https://github.com/files-community/Files/issues/17970#issuecomment-3667936651

- #15283

- [x] Check if an item has an icon
- [x] Check if an item has a text
- [x] Check if an item has a tooltip
- [x] Check if an item has a correct decorator (pin icon, or eject button, etc)

### DIFF
--- a/src/Files.App.Controls/Sidebar/SidebarStyles.xaml
+++ b/src/Files.App.Controls/Sidebar/SidebarStyles.xaml
@@ -200,9 +200,14 @@
 											<ItemsRepeater.ItemTemplate>
 												<DataTemplate>
 													<local:SidebarItem
-														AutomationProperties.AutomationId="{Binding Text, Mode=OneWay}"
+														AutomationProperties.AutomationId="{Binding Text}"
+														Decorator="{Binding ItemDecorator}"
+														Icon="{Binding IconElement}"
+														IsExpanded="{Binding IsExpanded, Mode=TwoWay}"
 														IsInFlyout="True"
-														Item="{Binding}" />
+														Item="{Binding}"
+														Text="{Binding Text}"
+														ToolTip="{Binding ToolTip}" />
 												</DataTemplate>
 											</ItemsRepeater.ItemTemplate>
 										</ItemsRepeater>


tests/Files.App.UITests/Views/SidebarViewPage.xaml
```diff
@@ -6,8 +6,4 @@
 	xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
 	xmlns:local="using:Files.App.UITests.Views"
 	xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-	Background="{ThemeResource ApplicationPageBackgroundThemeBrush}"
-	mc:Ignorable="d">
-
-	<controls:SidebarView x:Name="Sidebar" MenuItemsSource="{x:Bind sidebarModels}" />
```
marcelwgn, line 12: Why are we removing this?

src/Files.App.Controls/Sidebar/SidebarItem.cs
```diff
@@ -277,13 +262,6 @@ private void UpdateSelectionState()
 			UpdatePointerState();
 		}
 
-		private void UpdateIcon()
-		{
-			Icon = Item?.IconSource?.CreateIconElement();
-			if (Icon is not null)
```
marcelwgn, line 283: Why are we no longer doing this?

src/Files.App/Data/Items/DriveItem.cs
```diff
@@ -191,12 +191,13 @@ public object ToolTip
 
 		public bool IsExpanded { get => false; set { } }
 
-		public IconSource? IconSource
+		public IconElement? IconElement
```
marcelwgn, line 194: Why are switching to an icon element instead of icon source now?

tests/Files.App.UITests/Views/SidebarViewPage.xaml
0x5bfa: The control is not yet ready for other ways than one in MainPage.
We'll bring this back after finishing this refactoring.

src/Files.App.Controls/Sidebar/SidebarItem.cs
0x5bfa: We are now using binding for icons, we cannot call CreateIconElement in XAML.

src/Files.App/Data/Items/DriveItem.cs
0x5bfa: As above. And is there a reason why we should use IconSource?

marcelwgn: Sharing UI objects like that is a very bad idea, especially since it raises ownership questions and cause issues as soon as someone accidentally shares the element.

src/Files.App.Controls/Sidebar/SidebarItem.cs
marcelwgn: No, why are we no longer setting the UIA property appropriately? Also just because something doesnt work in xaml doesnt mean we should just ignore what the code did before.

src/Files.App/Data/Items/DriveItem.cs
0x5bfa: Yeah that's because we hardcode the binding in the resource dictionary. We'll change it to define in MainPage.xaml using template selector (one for DriveItem, another for LocationItem, another for WslDistroItem, etc). We'll also inherit that template selector to descendants.

src/Files.App.Controls/Sidebar/SidebarItem.cs
0x5bfa: Don't we set it in XAML today already?

https://github.com/files-community/Files/blob/01ddde98edbf514bd4458aba64d2237d9ffc8fc2/src/Files.App.Controls/Sidebar/SidebarStyles.xaml#L114-L124

marcelwgn: AccessibilityView is not a transitive property so every element needs to set it themselves. Just setting it on the presenter does not exclude its children from showing up in the UIA tree

src/Files.App/Data/Items/DriveItem.cs
marcelwgn: Even if for that one use case, we might do datatemplates, I would still consider this not best practice to take in raw icon elements like that. And if the goal is to make this a more widely used control, I think we should do what WinUI does, which is using IconSource in such cases.

0x5bfa: I mean, yeah, we'll create data templates in MainPage.xaml instead of defining it in the resource dictionary by default. But as of now, we do so, we have no other way of binding icons than this.

Once I create DataTemplateSelector/DataTemlate properties in SidebarView, this will be resolved.

```xml
<!-- In MainPage.xaml -->

<DataTemplate x:Key="SidebarDriveItemTemplate" x:DataType="dataitems:DriveItem">
    <SidebarItem ...>
        <SidebarItem.Icon>
            <IconSource Source="{x:Bind ...}">
        </SidebarItem.Icon>
    </SidebarItem>
</DataTemplate>

<SidebarItemTemplateSelector
    x:Key="SidebarItemTemplateSelector"
    DriveItem="{StaticResource SidebarDriveItemTemplate}"
    FolderItem="..."
    TagItem="..."
    .../>

<!-- ... -->

<SidebarView
    ItemTemplateSelector="{StaticResource SidebarItemTemplateSelector}"
    ...>
    ...
</SidebarView>
```

Does this have an issue for you?

marcelwgn: I dont see why we need to switch to IconElement and then back to IconSource while we currently already are using IconSource. And what is the source property on IconSource? Can't find that in the documentation.

0x5bfa: We are **temporarily** doing this.

> then back to IconSource

No, it will be icon data (byte array), glyph name, etc in the future. In this current Sidebar-related codebase, we cannot avoid directly creating IconElement because we don't have template selector (remember items of each section have different kind of icon).

marcelwgn: Lets discuss this offline first.

marcelwgn (approved):

PR #18030, opened 2026-01-08
Feature: Added setting to disable smooth scrolling

### LINKED ISSUE #15962: Feature: Add setting to disable smooth scrolling

Smooth scrolling is visually pleasing, but for me I find that it slows me down and adds input lag. I end up disabling it in every app that has the ability to do so.

We talked about this on Discord a little today, apparently WinUI's built-in scroll viewer doesn't support disabling smooth scrolling 😕 Optimally we could submit a PR to add a property for it over there, but it may also be possible to fork the scroll viewer control and implement it directly.

### Requirements

This would be a checkbox in settings. Once enabled, scrolling would happen instantly without animations or inertia. So if I move my mouse wheel by one detent, scrolling would happen over precisely 1 frame. The scroll viewer would go from OffsetY=y to OffsetY=y+delta without any frames in-between.

### Files Version

3.6.0.0

### Windows Version

10.0.26100.1301

### Comments

_No response_

workbysaran:
**Resolved / Related Issues**
Added a new setting to disable smooth scrolling across all layouts and scrollable components in the application

Closes #15962

1. Navigate to Settings > Appearance
2. Toggle "Enable smooth scrolling" setting
3. Verify scrolling behavior in all layout modes:
   - Details layout
   - List layout
   - Cards layout
   - Grid layout
   - Columns layout
4. Verify scrolling behavior in the sidebar navigation
5. Verify scrolling behavior on the home page
6. Verify scrolling behavior on the Settings page
7. Verify setting persists after app restart
8. Verify smooth scrolling is enabled by default for new users

### DIFF
--- a/src/Files.App/Services/Settings/AppearanceSettingsService.cs
+++ b/src/Files.App/Services/Settings/AppearanceSettingsService.cs
@@ -166,13 +166,6 @@ namespace Files.App.Services.Settings
 			set => Set(value);
 		}
 
-		/// <inheritdoc/>
-		public bool EnableSmoothScrolling
-		{
-			get => Get(true);
-			set => Set(value);
-		}
-
 		protected override void RaiseOnSettingChangedEvent(object sender, SettingChangedEventArgs e)
 		{
 			base.RaiseOnSettingChangedEvent(sender, e);
--- a/src/Files.App/Views/Settings/AppearancePage.xaml
+++ b/src/Files.App/Views/Settings/AppearancePage.xaml
@@ -283,17 +283,6 @@
 					AutomationProperties.Name="{helpers:ResourceString Name=ShowStatusBar}"
 					IsOn="{x:Bind ViewModel.ShowStatusBar, Mode=TwoWay}" />
 			</wctcontrols:SettingsCard>
-
-			<!--  Enable Smooth Scrolling  -->
-			<wctcontrols:SettingsCard
-				Header="{helpers:ResourceString Name=EnableSmoothScrolling}">
-				<wctcontrols:SettingsCard.HeaderIcon>
-					<FontIcon Glyph="&#xEC8F;" />
-				</wctcontrols:SettingsCard.HeaderIcon>
-				<ToggleSwitch
-					AutomationProperties.Name="{helpers:ResourceString Name=EnableSmoothScrolling}"
-					IsOn="{x:Bind ViewModel.EnableSmoothScrolling, Mode=TwoWay}" />
-			</wctcontrols:SettingsCard>
 		</StackPanel>
 	</Grid>
 </Page>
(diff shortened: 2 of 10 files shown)


src/Files.App/Views/Settings/AppearancePage.xaml
```diff
@@ -283,6 +283,18 @@
 					AutomationProperties.Name="{helpers:ResourceString Name=ShowStatusBar}"
 					IsOn="{x:Bind ViewModel.ShowStatusBar, Mode=TwoWay}" />
 			</wctcontrols:SettingsCard>
+
+			<!--  Enable Smooth Scrolling  -->
+			<wctcontrols:SettingsCard
+				Header="Enable smooth scrolling"
+				Description="Enable smooth scrolling experience">
```
yaira2: Most settings don't have a description, unless this adds important context, we can probably leave it out.

yair100:
@workbysaran instead of manually setting each ScrollViewer, do you think we can follow the approach that we use for other overrides in `AppResourcesService.cs`?

workbysaran:
> @workbysaran instead of manually setting each ScrollViewer, do you think we can follow the approach that we use for other overrides in `AppResourcesService.cs`?

@yaira2 
Good suggestion! I've implemented it using the AppResourcesService pattern.

src/Files.App/App.xaml
```diff
@@ -49,6 +51,18 @@
...
+				<Setter Property="IsScrollInertiaEnabled" Value="{ThemeResource App.ScrollInertiaEnabled}" />
+			</Style>
+
+			<Style TargetType="ListView">
+				<Setter Property="ScrollViewer.IsScrollInertiaEnabled" Value="{ThemeResource App.ScrollInertiaEnabled}" />
+			</Style>
+
+			<Style TargetType="GridView">
+				<Setter Property="ScrollViewer.IsScrollInertiaEnabled" Value="{ThemeResource App.ScrollInertiaEnabled}" />
+			</Style>
```
0x5bfa: Do you need this explicitly set even if we already set this for ScrollViewer itself?

src/Files.App/Services/Settings/AppearanceSettingsService.cs
```diff
@@ -166,6 +166,13 @@ public StatusCenterVisibility StatusCenterVisibility
 			set => Set(value);
 		}
 
+		/// <inheritdoc/>
+		public bool EnableSmoothScrolling
```
0x5bfa: Maybe IsScrollInertiaEnabled to align with the name in the codebase?

src/Files.App/App.xaml
workbysaran: I added them defensively since the previous implementation used attached properties on ListView/GridView, but testing confirms they're unnecessary. Removing them now to simplify.

src/Files.App/Services/Settings/AppearanceSettingsService.cs
yair100: SmoothScrolling is more understandable.

yair100:
> Good suggestion! I've implemented it using the AppResourcesService pattern.

@workbysaran thanks! This looks a lot cleaner.

PR #18044, opened 2026-01-11
Code Quality: Optimized FileExtensionHelpers.HasExtension

### LINKED ISSUE #18043: Code Quality: Unnecessary allocations in FileExtensionHelpers.HasExtension
### Description

FileExtensionHelpers.HasExtension is a frequently called method, but each call results in several easily avoided memory allocations:

- All callers use the `params` feature to pass the extensions to check, which implies that every caller allocates a `string[]` for each call. This can be avoided by declaring the parameter a `ReadOnlySpan<string>`.
- A closure is allocated for the delegate passed to `Enumerable.Any`. Note that if we use a `ReadOnlySpan<string>`, we can't use `Enumerable.Any` anyway. A simple `foreach` loop avoids the allocation and the delegate call.
- `Path.GetExtension(filePathToCheck)` is executed for each extension passed in the `extensions` parameter. Calling it once before the loop fixes that.

More optimizations are possible, but these are easy local fixes with minimal impact on code readability/maintainability.

### Concerned code

- src\Files.Shared\Helpers\FileExtensionHelpers.cs

### Gains

- Reduced memory allocations.
- Reduced CPU usage.

### Requirements

- Declaring the `extensions` parameter as `ReadOnlySpan<string>`.
- Use a `foreach` loop to check each extension.
- Call `Path.GetExtension(filePathToCheck)` once before the loop.

### Comments

I'll make a PR.

KrisVandermotten:
**Resolved / Related Issues**

- Closes #18043

Open any folder containing files, and observe that file type detection still works.

### DIFF
--- a/src/Files.Shared/Helpers/FileExtensionHelpers.cs
+++ b/src/Files.Shared/Helpers/FileExtensionHelpers.cs
@@ -32,7 +32,7 @@ namespace Files.Shared.Helpers
 		/// <param name="filePathToCheck">Path or name or extension of the file to check.</param>
 		/// <param name="extensions">List of the extensions to check.</param>
 		/// <returns><c>true</c> if the filePathToCheck has one of the specified extensions; otherwise, <c>false</c>.</returns>
-		public static bool HasExtension(string? filePathToCheck, params string[] extensions)
+		public static bool HasExtension(string? filePathToCheck, params ReadOnlySpan<string> extensions)
 		{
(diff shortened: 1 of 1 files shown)


yair100:
Thank you for this improvement, optimizations like these can go a long way!

By the way, for future PRs, it's recommended to check with us before doing the work. This helps prevent extra work in the event we don't implement a proposal 🙂

yair100:
@KrisVandermotten congrats on your first PR!

PR #18048, opened 2026-01-11
Code Quality: Made FileExtensionHelpers.IsSignableFile case insensitive

### LINKED ISSUE #18047: Bug: Detecting signable files fails when file extension is not lowercase
### Description

Signable files are detected using `FileExtensionHelpers.IsSignableFile`, which does a case-sensitive compare of the file extension against a list of known, lowercase extensions. As a result, if a filename has a non-lowercase extension, it will not be considered as signable, and the properties window will not show its signatures.

Root cause is that `FileExtensionHelpers._signableTypes` is constructed with the default (ordinal) comparer. The fix is to construct it using `new HashSet<string>(StringComparer.OrdinalIgnoreCase)`.

### Steps To Reproduce

1. Find a signed executable file, e.g. C:\Windows\explorer.exe
2. View the file properties and observe that signatures are shown.
3. Make a copy of the file, and change its extension to uppercase, e.g. C:\Windows\explorer - copy.EXE
4. View the copy's properties and observe signatures are not shown.

### Files Version

4.0.24.0

### Windows Version

10.0.26200.7462

### User ID

_No response_

### Log File

-

KrisVandermotten:
**Resolved / Related Issues**

- Closes #18047 

Followed Steps to Reproduce in the linked issue.

### DIFF
--- a/src/Files.Shared/Helpers/FileExtensionHelpers.cs
+++ b/src/Files.Shared/Helpers/FileExtensionHelpers.cs
@@ -14,7 +14,7 @@ namespace Files.Shared.Helpers
 	/// </summary>
 	public static class FileExtensionHelpers
 	{
-		private static readonly FrozenSet<string> _signableTypes = new HashSet<string>()
+		private static readonly FrozenSet<string> _signableTypes = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
 		{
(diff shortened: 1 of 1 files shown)


yaira2 (approved):
LGTM

Issue #18024, opened 2026-01-06
Feature: Add more descriptive error message when pull fails because of none commited changed files


Josh65-2201:
### What feature or improvement do you think would benefit Files?

Currently it show a `git pull timed out` when there are files that have been changed and are not committed.

### Requirements

- Add `Can not pull due to none committed changed files` dialog
- Could show files that are pending and allow discarding then retrying pull

### Files Version

4.0.24.0

### Windows Version

10.0.26200.7462

yair100:
Thank you for the feedback.

PR #18063, opened 2026-01-14
Feature: Added SSH support for Git and progress reporting in Status Center 

### LINKED ISSUE #17692: Feature: Implement SSH Authentication Support for Git Integration

Currently, the Git integration in Files only supports authentication over the HTTPS protocol, primarily using GitHub Personal Access Tokens (PATs) obtained via an OAuth login flow.

This means that any repository that has been cloned using an SSH URL (e.g., `git@github.com:user/repo.git`) cannot perform authenticated network operations like `pull`, `push`, or `fetch`. These operations fail because the application is unable to provide the SSH key credentials required by the Git remote.

This is a significant limitation for developers who prefer using SSH for its security and convenience, effectively preventing them from using the integrated Git features for their existing projects.

### Requirements

I propose to implement full authentication support for the SSH protocol. This would involve the following key changes, primarily within the `GitHelpers.cs` file:
1. **Enhance the** `CredentialsProvider`: The existing `CredentialsProvider` callback will be updated to intelligently handle different credential requests from `LibGit2Sharp`.
2. **Detect SSH Requests**: The handler will now check if the required credential type is `SupportedCredentialTypes.SshKey`.
3. **Provide SSH Key Credentials**: When an SSH key is requested, the handler will locate the user's default SSH private key (e.g., ~/.ssh/id_rsa).
4. **Support for Passphrases**: If the private key is encrypted with a passphrase, a ContentDialog will be displayed, securely prompting the user to enter it.
5. **Complete Authentication**: The key path and the (optional) passphrase will be used to create and return an `SshUserKeyCredentials` object, allowing the SSH authentication to succeed.

This solution would seamlessly enable SSH support while retaining the existing functionality for HTTPS authentication.

### Files Version

4.0.11.0

### Windows Version

10.0.26100.6584

### Comments

Adding SSH support would significantly enhance the Git functionality in Files.

I have already spent time investigating the codebase and have a clear implementation plan. I am willing to contribute the code for this feature and submit a Pull Request.

Before I begin development, I wanted to open this issue to discuss the plan with the project maintainers. I would greatly appreciate any feedback or guidance on this approach.

Thank you in advance!

NyanHeart:
**Resolved / Related Issues**

Closes #17692

**Changes**

- **SSH Support**: Implemented SSH key authentication via the system OpenSSH Agent. This resolves compatibility issues with LibGit2Sharp 0.31.0 and adds support for passphrase-protected keys without custom dialogs.
- **Status Center Integration**: Git Push, Fetch, and Pull operations now report real-time progress (items processed, transfer bytes) to the Status Center.
- **Real-time Refresh**: Fixed the Git Status Bar (commits ahead/behind) not refreshing automatically after remote operations.
- **Performance**: Refactored [FetchOrigin](cci:1://file:///c:/Github/Fork/Files/src/Files.App/Utils/Git/GitHelpers.cs:357:2-442:3) and [PullOriginAsync](cci:1://file:///c:/Github/Fork/Files/src/Files.App/Utils/Git/GitHelpers.cs:444:2-538:3) to use `async/await` and run on background threads, preventing UI freezes during network operations.
- **Cleanup**: Removed the obsolete `GitPassphraseDialog` and custom SSH credential logic.

1. **SSH Push/Pull**:
   - Cloned a repository using an SSH URL (`git@github.com:...`).
   - Verified that `git push` and `git pull` work seamlessly using the system SSH Agent (with both passphrase-protected and standard keys).
2. **Status Center Feedback**:
   - Performed a long-running push operation.
   - Observed the Status Center card appearing with "Pushing..." status.
   - Verified that progress percentages and item counts updated in real-time.
   - Verified that success/failure cards are displayed correctly after completion.
3. **UI Refresh**:
   - Committed a change locally.
   - Pushed the change to origin.
   - Verified that the "Commits Ahead" count in the status bar (bottom right) automatically cleared to 0 without needing a manual refresh.
4. **HTTPS Fallback**:
   - Verified that standard HTTPS repositories still work correctly with username/password (Personal Access Token) authentication.

### DIFF
--- a/src/Files.App/Utils/Git/GitHelpers.cs
+++ b/src/Files.App/Utils/Git/GitHelpers.cs
@@ -355,7 +355,7 @@ namespace Files.App.Utils.Git
 				branch.FriendlyName.Equals(branchName, StringComparison.OrdinalIgnoreCase));
 		}
 
-		public static async void FetchOrigin(string? repositoryPath, CancellationToken cancellationToken = default)
+		public static async Task FetchOrigin(string? repositoryPath, CancellationToken cancellationToken = default)
 		{
 			if (string.IsNullOrWhiteSpace(repositoryPath))
 				return;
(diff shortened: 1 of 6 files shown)


yair100 (dismissed):

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -355,65 +355,86 @@ public static bool ValidateBranchNameForRepository(string branchName, string rep
...
-					=> new UsernamePasswordCredentials
-					{
-						Username = signature.Name,
-						Password = token
-					};
-			}
+            var remoteName = repository.Network.Remotes.FirstOrDefault()?.Name ?? "origin";
+            
+            // Add Status Center Card
+            var banner = StatusCenterHelper.AddCard_GitFetch(remoteName, ReturnResult.InProgress);
```
workbysaran: Auto-fetch on folder navigation silently fails due to being called from a background thread in BaseShellPage.cs.

src/Files.App/Utils/Git/GitHelpers.cs
workbysaran: This is no longer used after the refactoring.

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -355,65 +355,86 @@ public static bool ValidateBranchNameForRepository(string branchName, string rep
...
+            var fsProgress = new StatusCenterItemProgressModel(banner.ProgressEventSource, enumerationCompleted: true, FileSystemStatusCode.InProgress);
+            
+            // Link CancellationTokens
+            using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, banner.CancellationToken);
+            var token = linkedCts.Token;
+
+            var fetchOptions = new FetchOptions
```
workbysaran: The older code used Prune = true in _fetchOptions, but that appears to have been removed. Is this intentional?

workbysaran:
I am able to successfully pull and push using an SSH URL (and HTTPS). However, I encountered the following problems,

1. When the Windows "OpenSSH Authentication Agent" service is not running, there is no user-friendly error message. Instead, it reports: "git pull failed due to a timeout error."
2. The same error message is displayed when there are uncommitted changes in the working directory and a pull is attempted.
3. Auto-fetch triggered during folder navigation fails with the following exception:
System.AggregateException: A Task's exception was not observed by awaiting the Task or by accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread.
---> System.Runtime.InteropServices.COMException (0x8001010E)
at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|38_0(Int32 hr)
at ABI.Microsoft.UI.Xaml.IApplicationMethods.get_Resources(IObjectReference _obj)
at Files.App.Utils.StatusCenter.StatusCenterItem..ctor(...)
at Files.App.ViewModels.UserControls.StatusCenterViewModel.AddItem(...)
at Files.App.Utils.StatusCenter.StatusCenterHelper.AddCard_GitFetch(...)
at Files.App.Utils.Git.GitHelpers.FetchOrigin(String repositoryPath, CancellationToken cancellationToken)

NyanHeart:
@workbysaran Thanks for the detailed review! I've pushed a fix that addresses your feedback:

1. **Threading Crash**: Wrapped all Status Center UI updates and dialog calls in DispatcherQueue.TryEnqueue to safely handle background fetch operations. This should resolve the COMException 0x8001010E / RPC_E_WRONG_THREAD crash.
2. **Error Handling**:
    - Added specific handling for `LibGit2SharpException` to surface underlying errors directly to the user via a dialog.
    - Added handling for `CheckoutConflictException` (uncommitted changes) with a localized error message.
4. **Logic**: Restored Prune = true for fetch operations.
5. **Cleanup**: Removed the unused variables as requested.

Could you please take another look? Thanks!

src/Files.App/Utils/Git/GitHelpers.cs
```diff
@@ -363,34 +358,39 @@ public static async Task FetchOrigin(string? repositoryPath, CancellationToken c
...
-                        fsProgress.AddProcessedItemsCount(progress.ReceivedObjects);
-                        fsProgress.Report((int)((progress.ReceivedObjects / (double)progress.TotalObjects) * 100));
-                    }
-                    return true;
-                }
-            };
+			var remoteName = repository.Network.Remotes.FirstOrDefault()?.Name ?? "origin";
+
+			// Add Status Center Card
+			var banner = StatusCenterHelper.AddCard_GitFetch(remoteName, ReturnResult.InProgress);
```
workbysaran, line 364: The call to StatusCenterHelper.AddCard_GitFetch needs to be dispatched to the UI thread. An exception is thrown during auto-fetch.

workbysaran:
@NyanHeart 

Thanks for the update. I tested again, and here are my findings.

>When the Windows "OpenSSH Authentication Agent" service is not running, there is no user-friendly error message. Instead, it reports: "git pull failed due to a timeout error."

It appears that the LibGit2Sharp library hides transport-level errors, so only the generic message "could not read refs from remote repository" is returned. We may need to investigate and address this in a separate PR.

The Git CLI shows the following error:

git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

I think we should accept this behavior with the current limitation.

>The same error message is displayed when there are uncommitted changes in the working directory and a pull is attempted.

Now it shows, "Please commit or stash your changes before pulling," which is good!

Josh65-2201:
> Added handling for CheckoutConflictException (uncommitted changes) with a localized error message.

Does this fix https://github.com/files-community/Files/issues/18024 ?

yair100:
@NyanHeart fyi

yair100:
@NyanHeart are you still planning to work on this?

Issue #18081, opened 2026-01-24
Code Quality: Fix performance related warnings from static code analysis


KrisVandermotten:
### Description

Static code analysis reports a number of performance relates warnings, some of which are trivial and risk free to fix.

- [CA1834: Use StringBuilder.Append(char) for single character strings](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1834)
- [CA1846: Prefer AsSpan over Substring](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1846)
- [CA1847: Use String.Contains(char) instead of String.Contains(string) with single characters](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1847)
- [CA1853: Unnecessary call to 'Dictionary.ContainsKey(key)'](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1853)
- [CA1854: Prefer the IDictionary.TryGetValue(TKey, out TValue) method](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1854)
- [CA1865-CA1867: Use 'string.Method(char)' instead of 'string.Method(string)' for string with single char](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1865-ca1867)

In addition, using analyzers I wrote myself and have been using for years, I found a set of similar risk free opportunities:

- Use StringComparison.Ordinal where applicable
- Use nameof instead of ToString on constant enum values
- Use interpolated strings instead of string.Format where possible
- Replace chars instead of single character strings where applicable
- Search for chars instead of single character strings where applicable

This is low hanging fruit, and we should fix these warnings.

### Concerned code

These warnings occur throughout the code base.

### Gains

- Improved performance
- Often improved readability

### Requirements

Fix the warnings, often using automated code fixers.

### Comments

Optimizations like these are often criticized for not being necessary, or not providing enough benefits. However, all of these fixes do have benefit, and the cost of applying them is virtually zero. IMHO, there is just no reason not to apply the fixes.

I can make a PR, one commit per fix type. Virtually all these fixes occur on single lines, making the PR very easy to review.

yair100:
I'm okay with this in theory but would prefer the changes be split across multiple PRs (as opposed to multiple commits in the same PR).

yair100:
[...]

PR #18089, opened 2026-01-25
Code Quality: Fix CA1853

KrisVandermotten:
**Resolved / Related Issues**

- Contributes to #18081

### DIFF
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -218,8 +218,7 @@ namespace Files.App.UserControls.TabBar
 			else
 				HorizontalTabView.SelectedItem = args.Tab;
 
-			if (ApplicationData.Current.LocalSettings.Values.ContainsKey(TabDropHandledIdentifier))
-				ApplicationData.Current.LocalSettings.Values.Remove(TabDropHandledIdentifier);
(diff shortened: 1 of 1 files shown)


src/Files.App/UserControls/TabBar/TabBar.xaml.cs
```diff
@@ -218,8 +218,7 @@ private void TabView_TabDragCompleted(TabView sender, TabViewTabDragCompletedEve
 			else
 				HorizontalTabView.SelectedItem = args.Tab;
 
-			if (ApplicationData.Current.LocalSettings.Values.ContainsKey(TabDropHandledIdentifier))
-				ApplicationData.Current.LocalSettings.Values.Remove(TabDropHandledIdentifier);
+			ApplicationData.Current.LocalSettings.Values.Remove(TabDropHandledIdentifier);
```
yair100, line 221: Are you sure this won't cause an exception if the key doesn't exist?

KrisVandermotten: Yes, the method simply returns false when the key doesn't exist.
I tested it by dragging and dropping a tab. It's also in line with [the interface documentation](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.idictionary-2.remove?view=net-9.0#returns).

yaira2 (approved):

PR #18096, opened 2026-01-25
Code Quality: Search for chars instead of single-char strings where possible

### LINKED ISSUE #18081: Code Quality: Fix performance related warnings from static code analysis

KrisVandermotten:
**Resolved / Related Issues**

- Closes #18081

### DIFF
--- a/src/Files.App.Storage/Ftp/FtpHelpers.cs
+++ b/src/Files.App.Storage/Ftp/FtpHelpers.cs
@@ -10,7 +10,7 @@ namespace Files.App.Storage
 	{
 		public static string GetFtpPath(string path)
 		{
-			path = path.Replace("\\", "/", StringComparison.Ordinal);
+			path = path.Replace('\\', '/');
 
 			var schemaIndex = path.IndexOf("://", StringComparison.Ordinal) + 3;
 			var hostIndex = path.IndexOf("/", schemaIndex, StringComparison.Ordinal);
@@ -44,7 +44,7 @@ namespace Files.App.Storage
 
 		public static string GetFtpAuthority(string path)
 		{
-			path = path.Replace("\\", "/", StringComparison.Ordinal);
+			path = path.Replace('\\', '/');
 			var schemaIndex = path.IndexOf("://", StringComparison.Ordinal) + 3;
 			var hostIndex = path.IndexOf("/", schemaIndex, StringComparison.Ordinal);
 
--- a/src/Files.App/Actions/FileSystem/CopyItemPathAction.cs
+++ b/src/Files.App/Actions/FileSystem/CopyItemPathAction.cs
@@ -39,7 +39,7 @@ namespace Files.App.Actions
 				: context.ShellPage.ShellViewModel.WorkingDirectory;
 
 				if (FtpHelpers.IsFtpPath(path))
-					path = path.Replace("\\", "/", StringComparison.Ordinal);
+					path = path.Replace('\\', '/');
 
 				SafetyExtensions.IgnoreExceptions(() =>
 				{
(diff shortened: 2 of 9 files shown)


yair100:
Can you please fill out the steps for how this change was tested?

KrisVandermotten:
> Can you please fill out the steps for how this change was tested?

Honestly, I didn't test all of these call sites. Why should I? There is no behavioral change.

Indeed, when called with `StringComparison.Ordinal`, `string.Replace(string, string, StringComparison` [calls `string.Replace(string, string)`](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/[String.Manipulation.cs](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/String.Manipulation.cs,b61602cc05c2db6e),b61602cc05c2db6e). And when called with two strings of length 1, that [calls `string.Replace(char, char)`](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/[String.Manipulation.cs](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/String.Manipulation.cs,69fc1d0aa6df8a90),69fc1d0aa6df8a90). This change merely optimizes the call chain, and it makes the code clearer. Note also that [the documentation for `string.Replace(string, string)`](https://learn.microsoft.com/en-us/dotnet/api/system.string.replace?view=net-9.0#system-string-replace(system-string-system-string)) says that it performs an ordinal search, so the behavior is documented and very unlikely to change in a future version. 

The same is true for `string.IndexOf(string, int, StringComparison` when called with a string of length 1 and `StringComparison.Ordinal`. That ultimately calls into the same method on `SpanHelpers` that `string.IndexOf(char, int)` calls into. Again, we're merely optimizing the call chain and making to code clearer.

Also, I used an analyzer to find these call sites, and a code fixer to fix them. That's why I'm sure I found all call sites, and can't have made a mistake by manually fixing them.

PR #18097, opened 2026-01-25
Code Quality: Performance improvements in NaviagtionHelpers.OpenPath

### LINKED ISSUE #18082: Code Quality: Inefficiencies in PathHelpers.OpenPath
### Description

There are some inefficiencies in PathHelpers.OpenPath.

- The method starts with obtaining a lot of information about the path, even though in many scenarios the information won't be used. We should obtain information only when it's relevant. For example, the call to `(FileExtensionHelpers.IsScreenSaverFile(path)` should be made only when we know `path` is a file that we need to open.
- There are several calls to `Win32Helper.HasFileAttribute`, all of which repeat the same Windows API call. We should introduce a `Win32Helper.GetFileAttributes(string path)` method and make the call only once.

### Concerned code

- src\Files.App\Helpers\Navigation\NavigationHelpers.cs
- src\Files.App\Helpers\Win32\Win32Helper.Storage.cs

### Gains

- Improved performance

### Requirements

- Introduce `Win32Helper.GetFileAttributes(string path)` and use it.
- Move some calls to where their result is needed, to avoid unnecessary work.

### Comments

I can make a PR.

KrisVandermotten:
**Resolved / Related Issues**

- Closes #18082

1. Opened Files
2. Navigate to libraries and folders, open files, ...

### DIFF
--- a/src/Files.App/Helpers/Navigation/NavigationHelpers.cs
+++ b/src/Files.App/Helpers/Navigation/NavigationHelpers.cs
@@ -403,16 +403,7 @@ namespace Files.App.Helpers
 			if (associatedInstance.ShellViewModel is null)
 				return false;
 
-			string previousDir = associatedInstance.ShellViewModel.WorkingDirectory;
-			bool isHiddenItem = Win32Helper.HasFileAttribute(path, System.IO.FileAttributes.Hidden);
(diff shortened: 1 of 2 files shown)


workbysaran:
Reviewed. Looks good!

PR #18127, opened 2026-02-04
GitHub: Add PR check for non-en-US resource changes

yair100:
Add a GitHub Actions workflow that runs on pull requests to detect changes to .resw files under the Strings folder that are not en-US. The job lists changed files against the base ref, filters for non-en-US .resw files, and posts a comment on the PR (unless the author is crowdin[bot]) asking authors to revert and use Crowdin for translations. This prevents accidental direct edits to localized resource files.

### DIFF
--- a/.github/workflows/pr-resource-check.yml
+++ b/.github/workflows/pr-resource-check.yml
@@ -1,0 +1,55 @@
+
+name: Resource File Check
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  check-non-enus-resources:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Get PR branch info
+        id: pr_branch
+        run: |
+          echo "pr_head_repo=$(jq -r .pull_request.head.repo.full_name < $GITHUB_EVENT_PATH)" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-resource-check.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Detect non-enus resource changes in Strings
+        id: detect
+        run: |
+          # Only check for .resw files (including subfolders)
+          grep -E "\.resw$" changed_files.txt | grep -vi "en-us" > non_enus_resources.txt || true
+          count=$(cat non_enus_resources.txt | wc -l)
+          echo "count=$count" >> $GITHUB_OUTPUT
+
+      - name: Comment on PR if needed
+        if: steps.detect.outputs.count != '0' && steps.pr_author.outputs.pr_author != 'crowdin[bot]'
```
0x5bfa: I dont think the check for "crowdin[bot]" is needed because the PR author has been always you.

yair100: This is addressed in e69b30415ad4bad34b5169dba50522e6b83eb370.

0x5bfa:
See my previous review, or you won't try?

yair100:
I addressed your code comments, is there something else?

0x5bfa:
> We should try custom instructions of Copilot code review instead for this. https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#enabling-or-disabling-custom-instructions-for-copilot-code-review

PR #18148, opened 2026-02-11
Code Quality: Avoid redownloading Satori package

yair100:
Add date-based archive path and flags to skip downloading when a matching Satori archive already exists. DownloadFile now targets a dated subfolder and only runs if the archive is missing; a SatoriDownloadedThisRun flag ensures the archive is only deleted when it was downloaded during this run. This reduces redundant downloads across builds and preserves existing archives.

### DIFF
--- a/src/Satori.targets
+++ b/src/Satori.targets
@@ -16,7 +16,7 @@
       <SatoriDownloadedThisRun>true</SatoriDownloadedThisRun>
     </PropertyGroup>
     <DownloadFile
-        SourceUrl="https://github.com/files-community/Satori/releases/latest/download/windows_$(Platform).zip"
+        SourceUrl="https://github.com/files-community/Satori/releases/latest/download/$(SatoriArchiveName)"
         DestinationFolder="$(IntermediateOutputPath)SatoriPackage\$(SatoriDownloadDate)"
         Retries="3"
         Condition="!Exists('$(SatoriArchivePath)')">


src/Satori.targets
```diff
@@ -1,9 +1,25 @@
 ﻿<Project>
+  <PropertyGroup>
+    <SatoriDownloadDate>$([System.DateTime]::UtcNow.ToString('yyyyMMdd'))</SatoriDownloadDate>
+    <SatoriArchiveName>windows_$(Platform).zip</SatoriArchiveName>
```
hez2010, line 4: `_$(Platform)` is not necessary as it's already included in `IntermediateOutputPath`. So we can always set the name to `Satori.zip`.

yair100: Isn't this the name of the downloaded archive?

hez2010: You are right.. then maybe use it in the source url as well?

yair100: This is part of the source url as well.

src/Satori.targets
```diff
@@ -1,9 +1,25 @@
...
+    <ItemGroup Condition="Exists('$(SatoriArchivePath)')">
+      <SatoriArchive Include="$(SatoriArchivePath)" />
+    </ItemGroup>
+
+    <!-- Download if the archive does not exist -->
+    <PropertyGroup Condition="!Exists('$(SatoriArchivePath)')">
+      <SatoriDownloadedThisRun>true</SatoriDownloadedThisRun>
+    </PropertyGroup>
     <DownloadFile
         SourceUrl="https://github.com/files-community/Satori/releases/latest/download/windows_$(Platform).zip"
```
hez2010:
```suggestion
        SourceUrl="https://github.com/files-community/Satori/releases/latest/download/$(SatoriArchiveName)"
```

yair100:
That's a good idea.

PR #18163, opened 2026-02-15
Code Quality: Improve CommandManager

### LINKED ISSUE #18136: Code Quality: Inefficiencies in CommandManager
### Description

The `CommandManager` is a very central concept in Files: all commands are registered there. Unfortunately, there are several inefficiencies:

- Commands are stored in a `FrozenDictionary<CommandCodes, IRichCommand>`, but `CommandCodes` is a generated `enum` with underlying values from 0 to n. That means we might as well store the commands in an `ImmutableArray<IRichCommand>`. That would require less memory and be more efficient to loop over or index into. Moreover, the `FrozenDictionary` is constructed using a LINQ query over an intermediate, dynamically growing `Dictionary`. The `ImmutableArray` can be constructed directly, which would be much faster.
- The `this[string code]` indexer uses `Enum.Parse<CommandCodes>` where `Enum.TryParse<CommandCodes>` could be used.
- `OverwriteKeyBindings` uses a `ForEach` method to loop over command collections, requiring delegate calls to `RestoreKeyBindings`. A regular `foreach` loop is much faster.
- The `ActionCommand` constructor has an unused `CommandManager` parmeter.
- The `GeneratedRichCommandAttribute` has an unused `AssociatedActions` property.

The `CommandManagerGenerator` also has some significant performance problems.

- The call to `context.SyntaxProvider.ForAttributeWithMetadataName` returns an `IncrementalValuesProvider<GeneratorAttributeSyntaxContext>`, which is not cacheable. This causes frequent code regeneration. Given that all we need are the action names, it should return an `IncrementalValuesProvider<string>`. A good article about this is [Avoiding performance pitfalls in incremental generators](https://andrewlock.net/creating-a-source-generator-part-9-avoiding-performance-pitfalls-in-incremental-generators/) by Andrew Lock.
- The actual code generating methods should be `static`, use a strongly typed parameter, and use the `StringBuilder` more efficiently.

### Concerned code

- src\Files.App\Data\Commands\Manager\CommandManager.cs
- src\Files.App\Data\Commands\ActionCommand.cs
- src\Files.Core.SourceGenerator\Generators\CommandManagerGenerator.cs
- src\Files.Shared\Attributes\GeneratedRichCommandAttribute.cs

### Gains

- Reduced startup time.
- Faster execution of everything involving commands.
- Reduced memory usage.
- Improved code readability.
- Improved IDE responsiveness (this is just one generator though).

### Requirements

- Fix the issues described above.

### Comments

I have a tested PR ready, I can push it a soon as you give the green light.

KrisVandermotten:
**Resolved / Related Issues**

- Closes #18136

1. Inspect the generated files, and compare them to the previously generated files
2. Run Files and use commands


yair100:
Thank you for the PR.

0x5bfa (approved):
LGTM. This looks to be incremental now.

KrisVandermotten:
> This looks to be incremental now.

It is indeed.

Do you want me to update the StringBuilder logic in StringsPropertyGenerator using raw strings too (in a separate PR, obviously)?

yair100:
Does this have a positive impact on startup performance?

0x5bfa:
Generators run on VS threads at compilation time, no.

> Do you want me to update the StringBuilder logic in StringsPropertyGenerator using raw strings too (in a separate PR, obviously)?

I think it's already better.

KrisVandermotten:
> I think it's already better.

I think it's indeed improved in this PR. But my question was about #18161, where I did not touch the StringBuilder code (yet). Do you want me to make a new PR, that improves that code similarly to the improvements here?

KrisVandermotten:
> Does this have a positive impact on startup performance?

Not sure what you are referring to?

This PR does have a positive impact on startup performance. The newly generated code is more efficient than the previously generated code. Just to be clear, I have no plans for further changes to this PR.

On the other hand, changing the way in which the code is generated, i.e. replacing multiple calls to `AppendLine` by a single call to `Append` with a multiline raw string in the generator, has no impact on the generated code whatsoever. This is primarily about readability of the generating code. The generation is also slightly more efficient, but the generated code is the same. I already did this in this PR, but not (yet) in #18161.

yair100:
Thanks for confirming. Did you run any benchmarks to compare the difference in startup performance? If it's noticeable, I can include a mention in the next update changelog.

KrisVandermotten:
I did not run any formal benchmark. Measuring cold startup time correctly is not exactly trivial.

I would not expect the speedup to be immediately noticeable with the naked eye. There are a lot of things going on during startup, and building the CommandManager is just one of them. You know, Amdahl's law.

yair100:
Thanks again.

PR #18177, opened 2026-02-18
Code Quality: Enable NuGet Transitive Pinning

dfederm:
**Resolved / Related Issues**
N/A

Build (no behavioral changes expected)

**Changes**
This enables [Transitive Pinning](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#transitive-pinning) which essentially makes your `Directory.Packages.props` a partial lock file. If a package is listed in `Directory.Packages.props`, that will be the version, even if a project doesn't explicitly reference that package and it's just a transitive dependency. This is the primary benefit of transitive pinning, to ensure all projects within the repo are using the same version of all packages. It also allows security issues in transitive dependencies to be more easily addressed by simply updating `Directory.Packages.props` instead of adding a top-level dependency to all impacted projects, which is far messier and less maintainable.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -2,6 +2,7 @@
 <Project>
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="ByteSize" Version="2.1.2" />


yair100:
Thank you for the pull request.

In the future, it's recommended to only open pull requests for issues marked 'Ready to build', this helps prevent extra work in the event we don't end up accepting a change.

In any event, since the work is already done, we'll take a look 🙂

yair100:
@dfederm thank you again for opening this PR. After discussion with the team, we've decided not to move forward with this proposal.

PR #18188, opened 2026-02-21
Add tab button shifts when non-tab item is dragged over tab strip

### LINKED ISSUE #14143: BUG: add tab button shifts when a non-tab item is dragged next to it
### Description

related to feature #4710 which I will be PRing shortly
I will take on this bug fix in the next couple days

### Steps To Reproduce

1. Drag anything between the + button and the rightmost tab
2. The button skips to the right even if a non tab item is dragged next to it

### Requirements

Button should only skip when dragging a tab

### Files Version

3.0.13.0

### Windows Version

Windows 11 22H2

### Log File

[debug.log](https://github.com/files-community/Files/files/13538039/debug.log)

WarisMaqbool-ship-it:
Fixes #14143

When a non-tab item is dragged over the tab strip, the add tab button 
incorrectly shifts to the right. This happens because the drag over 
event was only setting CanReorderTabs to false but not explicitly 
rejecting the drag operation.

The fix sets AcceptedOperation to None in the else branch of 
TabView_TabStripDragOver, which correctly signals that non-tab items 
are not valid drop targets on the tab strip.

Note: Developed on Mac, unable to run the app locally. 
Would appreciate maintainer verification on Windows.

### DIFF
--- a/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
+++ b/src/Files.App/UserControls/TabBar/TabBar.xaml.cs
@@ -170,6 +170,7 @@ namespace Files.App.UserControls.TabBar
 			else
 			{
 				HorizontalTabView.CanReorderTabs = false;
+				e.AcceptedOperation = DataPackageOperation.None;
 			}
 		}
 


yair100:
@WarisMaqbool-ship-it thank you for the PR. Unfortunately, this doesn't seem to have resolved the linked issue. I recommend setting up a VM for testing purposes.

Issue #17763, opened 2025-10-14
Bug: Files won't launch


Lacki23:
### Description

This happened after updating the Microsoft Store, I trying reinstalling Files, and replacing .NET versions, but unfortunately, this did not fix the problem. [same problem in Preview version]

### Steps To Reproduce

Open program

### Files Version

4.0.11.0

### Windows Version

19045.6332 [22H2]

### Log File

[logs.txt](https://github.com/user-attachments/files/22903080/logs.txt)

yair100:
- What happens when you click the Files icon, does a window display at all?
- Do you see anything in Event Viewer?
- Does Files Preview work? https://files.community/download

Lacki23:
> * What happens when you click the Files icon, does a window display at all?
> * Do you see anything in Event Viewer?
> * Does Files Preview work? https://files.community/download

1. Only a loading ring near the cursor for a moment
2. Check logs.txt
3. Nope, same situation

yair100:
> Check logs.txt

I mean in the system Events Viewer.

Josh65-2201:
Log.txt is event viewer output.

yair100:
Thanks

workbysaran:
@Lacki23
We've improved the start-up error handling in v4.0.34. Instead of silently crashing, the app should now show an error dialog with more details about what went wrong.

Could you update to v4.0.34 and try again?

Josh65-2201:
@Lacki23 Can you do the above

PR #18208, opened 2026-02-26
Code Quality: Improve startup error handling for Windows App Runtime 

workbysaran:
**Resolved / Related Issues**
Improved the startup experience when the Windows App Runtime is missing or corrupted. Previously the app would silently terminate showing only a brief loading cursor with no feedback. The improvement catches the COM registration failure ('REGDB_E_CLASSNOTREG') on the first Windows App Runtime call, shows  a user-facing error dialog and rethrows a descriptive exception so the failure is recorded in Windows Event Viewer.

Related #17763

1. Install Files on a machine with the Windows App Runtime missing or corrupted
2. Launch Files and confirm the error dialog appears with a clear message instead of a silent crash
3. Open Event Viewer > Windows Logs > Application and confirm a descriptive entry is recorded


src/Files.App/Program.cs
```diff
@@ -20,6 +21,11 @@ namespace Files.App
 	/// </remarks>
 	internal sealed class Program
 	{
+		// These strings are intentionally hardcoded and cannot be moved to resource files.
+		// The Windows App Runtime (which powers the resource loading system) may itself be unavailable at this point
+		private const string MissingRuntimeMessage = "Files failed to start. A required Windows component could not be loaded. Try reinstalling Files from the Microsoft Store.";
```
yair100: Files can also be installed from outside the Store, is it possible to add a link to the download page? https://files.community/download

src/Files.App/Helpers/Win32/Win32PInvoke.Methods.cs
```diff
@@ -341,5 +341,8 @@ out IntPtr pszPath
 		// cryptui.dll
 		[DllImport("cryptui.dll", SetLastError = true, CharSet = CharSet.Auto)]
 		public unsafe static extern bool CryptUIDlgViewSignerInfo(CRYPTUI_VIEWSIGNERINFO_STRUCT* pViewInfo);
+
+		[DllImport("user32.dll", CharSet = CharSet.Unicode)]
+		public static extern int MessageBoxW(IntPtr hWnd, string text, string caption, uint type);
```
0x5bfa: Use CsWin32. Add `MessageBox` to NativeMethods.txt in Files.App.CsWin32.

src/Files.App/Program.cs
```diff
@@ -20,6 +21,11 @@ namespace Files.App
...
+		private const string MissingRuntimeTitle = "Files - Startup Error";
```
0x5bfa: You should move these consts to Constats.cs file.

src/Files.App/Constants.cs
```diff
@@ -282,5 +282,13 @@ public static class Distributions
 				"FilesDev", // dev
 			};
 		}
+
+		public static class Startup
+		{
+			// These strings are intentionally hardcoded and cannot be moved to resource files.
+			// The Windows App Runtime (which powers the resource loading system) may itself be unavailable at this point
+			public const string MissingRuntimeMessage = "Files failed to start. A required Windows component could not be loaded. Try reinstalling Files from the Microsoft Store or from https://files.community/download";
```
yair100, line 290: Is the link clickable?

workbysaran: No, the link is not clickable. If you need a clickable link, we need to use TaskDialog instead.

yair100: It's okay for now, hopefully this doesn't come up very often.

PR #18218, opened 2026-03-03
Fix: Fixed wrong scroll position when switching out of semantic zoom

### LINKED ISSUE #18203: Bug: Scroll position changes when switching out of semantic zoom
### Description

When using Files Stable/Preview in Windowed Mode with Details enabled, clicking a group header like "Today" to collapse it and then clicking it again to expand it causes the view to jump to the wrong position. The UI doesn't stay centered.

https://github.com/user-attachments/assets/1a11e296-8631-4bf0-af46-e8952a8d2dc9

### Steps To Reproduce

1. Open Files Stable/Preview.
2. Ensure It is in Windowed Mode and not maximized.
3. Enable "Details".
4. Go into any folder but make sure it contains files/folders inside the Folder (Downloads, Documents, Desktop...)
5. Click the "Today" group header then click again.
6. The UI will be in different position, but not in center.

### Files Version

4.0.28.0

### Windows Version

10.0.26100.7922

### User ID

9bdf90bc-38fe-4333-9390-4d45eda8a643

### Log File

[debug.log](https://github.com/user-attachments/files/25533498/debug.log)

workbysaran:
**Resolved / Related Issues**
Fixed an issue where navigating to a group from the Group Selection page in the Details Layout scrolled to an incorrect position, causing the group header to be partially hidden and the Name column to be out of view.

Closes #18203

1. Open a folder with grouping enabled in Details Layout
2. Click the group header to open the zoomed-out group overview
3. Click a group header (e.g. "November 2025") to zoom back in
4. Verify the group header is visible just below the column header bar
5. Verify the Name column is visible (horizontal scroll reset to 0)

### DIFF
--- a/src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
+++ b/src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
@@ -911,6 +911,20 @@ namespace Files.App.Views.Layouts
 		private void FileList_Loaded(object sender, RoutedEventArgs e)
 		{
 			ContentScroller = FileList.FindDescendant<ScrollViewer>(x => x.Name == "ScrollViewer");
+			const double OffsetCorrection = 88; // HeaderGrid (40) + ListViewHeaderItem (44 + 4 margin)
+
+			RootGridZoom.ViewChangeStarted += (_, args) =>
+			{
+				var scroller = ContentScroller;
+				if (args.IsSourceZoomedInView || scroller is null)
+					return;
+				void OnZoomScrolled(object? s, ScrollViewerViewChangedEventArgs ve)
+				{
+					scroller.ViewChanged -= OnZoomScrolled; 
+					scroller.ChangeView(0, Math.Max(0, scroller.VerticalOffset - OffsetCorrection), null, true);
+				}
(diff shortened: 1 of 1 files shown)


src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
```diff
@@ -911,6 +911,20 @@ private bool IsCorrectColumn(FrameworkElement element, int columnIndex)
 		private void FileList_Loaded(object sender, RoutedEventArgs e)
 		{
 			ContentScroller = FileList.FindDescendant<ScrollViewer>(x => x.Name == "ScrollViewer");
+			const double OffsetCorrection = 88; // HeaderGrid (40) + ListViewHeaderItem (44 + 4 margin)
```
yair100, line 914: Does this take into account the different item heights?

workbysaran: The values are calculated based on the existing header and row height margin values from DetailsLayoutPage.xaml. These are the defaults. Although it should work across most scenarios, the actual height may slightly vary depending on the font and resolution. A more complex solution would be to calculate the actual rendered heights dynamically.

PR #18262, opened 2026-03-15
Add 'Remove all tags' ption for files and folders

### LINKED ISSUE #18256: Feature: Add "Remove all tags" option for files and folders

Currently, when removing tags from files in Files, tags must be removed individually using the context menu.

Tags are stored in an Alternate Data Stream (ADS) attached to the file. If a user wants to remove all tags from a file (for example before sharing it with someone who does not use Files), the current workflow requires either removing them one by one or manually deleting the ADS after enabling ADS file visibility in the settings.

It would be helpful to have a built-in option to remove all tags associated with a file or folder in a single action.

### Use case

In a collaborative workflow, users may tag files locally using Files (for example with colored tags). Before sending a file to someone outside the team, the user may want to ensure that no tag metadata remains attached to the file and that the file appears completely unmodified.

Currently this requires either manually removing the ADS or deleting tags individually while toggling ADS visibility on and off.

A **"Remove all tags"** action in the context menu would make this process simpler, faster, and safer.

### Requirements

- Add a "Remove all tags" option for files and folders.
- The option should remove all tag metadata associated with the selected item.
- Internally this would delete the ADS used by Files to store tags.
- The action should not modify the file contents themselves.
- The option should be available in both:
  - the context menu
  - the tag management UI
- Support multi-selection so tags can be removed from multiple files and folders at once.
- The feature should allow users to remove all tags from selected files or folders in a single action.

### Files Version

4.0.34.0

### Windows Version

10.0.26200.8037

### Comments

_No response_

shivansh31414:
**Resolved / Related Issues**
Closes Feature: Add “Remove all tags” option for files and folders 

- Closes #18256

Applied multiple tags to files and folders.

Used the new “Remove all tags” option from the context menu.

Verified all tags were cleared in one action.

Confirmed UI updated correctly in the Status Center.

Tested with items that had no tags to ensure graceful handling.

Checked multiple selections to confirm consistent behavior.

Ran accessibility checks with Accessibility Insights for Windows.

Validated no regressions or memory leaks occurred.

### DIFF
--- a/src/Files.App/Strings/en-US/Resources.resw
+++ b/src/Files.App/Strings/en-US/Resources.resw
@@ -1697,6 +1697,9 @@
   <data name="EditTags" xml:space="preserve">
     <value>Edit tags</value>
   </data>
+  <data name="RemoveAllTags" xml:space="preserve">
+    <value>Remove all tags</value>
+  </data>
   <data name="PropertyPartOfSet" xml:space="preserve">
     <value>Part of set</value>
   </data>
--- a/src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
+++ b/src/Files.App/UserControls/Menus/FileTagsContextMenu.cs
@@ -5,14 +5,15 @@ using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
 using Microsoft.UI.Xaml.Markup;
 using Microsoft.UI.Xaml.Media;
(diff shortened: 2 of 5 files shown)


yair100:
Please see here for the latest proposal https://github.com/files-community/Files/issues/18256#issuecomment-4232904113

Closing due to inactivity.

PR #18270, opened 2026-03-16
Code Quality: Created Action groups

yair100:
**Resolved / Related Issues**

- Added support for grouping actions. This will allow us to create flyouts without manually adding each item to the flyout. It will allow us to reduce code in the Context Menu, Toolbar, and possibly add some new features in the future.
- Created a few basic groups to start the process (New, Set as Background, & Extract).
- The AccessKey property can now be defined from the RichCommand class.

More work is needed to apply these changes, but I don't want to complicate this PR.

1. Confirmed the modified toolbar behaves the same as `main`.


src/Files.App/Actions/CommandGroups.cs
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) Files Community
+// Licensed under the MIT License.
+
+using Files.App.Data.Commands;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+
+namespace Files.App.Actions;
```
0x5bfa: Better to put this (preferably separating each class) in https://github.com/files-community/Files/tree/main/src/Files.App/Data/Commands/Manager

src/Files.Core.SourceGenerator/Generators/CommandManagerGenerator.cs
```diff
@@ -134,5 +172,33 @@ private static ImmutableArray<IRichCommand> CreateCommands() =>
...
+				commandGroups.Select(static commandGroup => $"\t\t{commandGroup},"));
+
+			return $$"""
+				// <auto-generated/>
+				
+				using global::System.Collections.Generic;
+				
+				namespace Files.App.Actions;
+				
+				public sealed class CommandGroups
```
0x5bfa, line 189: Preferably `CommandGroupManager` with namespace `Files.App.Data.Commands`

yair100:
> Overall looks good to me. Bad that there's no ItemsSource in MenuFlyout for us to bind properties.

It would be nice, agreed.

PR #18295, opened 2026-03-24
Code Quality: Update the CI and CDs to use the VS2026-installed runner image

0x5bfa:
### Resolved / Related Issues

Update the CI and CDs to use the VS2026-installed runner image

_None_

### DIFF
--- a/.github/workflows/cd-sideload-preview.yml
+++ b/.github/workflows/cd-sideload-preview.yml
@@ -18,7 +18,7 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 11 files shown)


0x5bfa:
We should also update the MSVC toolchain used in the C++ projects to v145 from v143.

yair100:
> We should also update the MSVC toolchain used in the C++ projects to v145 from v143.

Can you please do that in in this PR? The reason we were still using v143 was for the CI.

PR #18315, opened 2026-03-28
Re-evaluate Windows Terminal availability dynamically

raashish1601:
Follow-up to #18285.


Josh65-2201:
You should make this a suggestion on the PR rather than making your own

PR #18317, opened 2026-03-28
Code Quality: Optimized GetModifierCode

### LINKED ISSUE #18316: Code Quality: HotKey GetModifierCode implementations can be optimized
### Description

`HotKey.LocalizedLabel` uses a local function `GetModifierCode` to calculate a localized string representation of the HotKey's modifiers.

It's a frequently called method. During startup alone, it's called 55 times on my test machine. Moreover, `LocalizedLabel` is called whenever the HotKey is displayed, e.g. when a local menu is opened.

That means the function should be efficient. Unfortunately, the `builder.Remove(0, 1)` call at the end is relatively expensive. Luckily, is can be avoided easily, if we append the `+` at the end of every modifier string instead of at the beginning.

While we're at it, we should also fix the `GetModifierCode` local function in `RawLabel`, which has the same problem. We should also append string constants there, instead of enum values which require an expensive runtime `ToString` call.

### Concerned code

- src\Files.App\Data\Commands\HotKey\HotKey.cs

### Gains

- Better performance (CPU)

### Requirements

- refactor the `GetModifierCode` local functions

### Comments

I'll make a PR. It's simple and straightforward. The effect won't be enormous, but everything helps.

KrisVandermotten:
**Resolved / Related Issues**

- Closes #18316 

1. Opened Files
2. Observed that HotKey labels still work as before

### DIFF
--- a/src/Files.App/Data/Commands/HotKey/HotKey.cs
+++ b/src/Files.App/Data/Commands/HotKey/HotKey.cs
@@ -206,14 +206,14 @@ namespace Files.App.Data.Commands
 				{
 					StringBuilder builder = new();
 					if (modifiers.HasFlag(KeyModifiers.Alt))
-						builder.Append($"+{KeyModifiers.Alt}");
+						builder.Append("Alt+");
 					if (modifiers.HasFlag(KeyModifiers.Ctrl))
-						builder.Append($"+{KeyModifiers.Ctrl}");
+						builder.Append("Ctrl+");
 					if (modifiers.HasFlag(KeyModifiers.Shift))
-						builder.Append($"+{KeyModifiers.Shift}");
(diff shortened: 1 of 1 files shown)


yair100:
What happens if you change languages at run time?

KrisVandermotten:
> What happens if you change languages at run time?

The exact same thing as before. As you can see, `LocalizedLabel` is still called in exactly the same places, and it is still using `LocalizedModifiers`. Nothing has changed there.

Just to be clear: `LocalizedModifiers` is initialized at startup, and never updated. So the `LocalizedLabel` will not be updated either, as before. If you think that's a bug, I can fix it in this PR by not caching `LocalizedModifiers`.

yair100 (approved):
LGTM

PR #18320, opened 2026-03-29
Feature: Enable drag and drop reordering for Pinned Sidebar items

### LINKED ISSUE #15430: Request: pinned folders may be moved

Would like the option to rearrange pinned folders (drag and drop). Another possibility would be that the program would sort them in some order (alphabetically).

### Requirements

- This proposal will accomplish reordering the pinned folders as the user desires
- This proposal will accomplish sort the pinned folders in a logical order of some type (eg. alphabetically)

### Files Version

3.4.1.0

### Windows Version

10.0.22631

### Comments

That's all folks.

kalmix:
**Resolved / Related Issues**
- Closes #15430

**Code Changes**
* **Implemented Inline Drag-and-Drop Reordering:** Removed the standalone `ReorderSidebarItemsDialog` entirely and replaced it with native inline drag-and-drop directly within the sidebar.
* **Added Visual Drop Indicators:** Updated `SidebarItem` UI and logic to calculate insertion positions (top, bottom, or center) and display the appropriate visual feedback when hovering during a drag operation.
* **Resolved OS-Level Race Conditions:** Implemented a robust `FileSystemWatcher` in `WindowsQuickAccessService` to monitor the hidden `automaticDestinations-ms` file. This guarantees that the Windows Shell has finished asynchronously saving the new pin order before the app triggers a UI refresh, preventing snapping and stuttering.
* **Optimized Widget Rendering:** Updated `QuickAccessWidgetViewModel` to use `Items.Move()` when diffing and updating reordered items, rather than destroying and rebuilding the UI collection. This eliminates screen flickering during reorders.
* **Refined Exception Handling:** Added safe interception for stale OLE payload crashes locally in the sidebar drag-and-drop event handlers. This uses the native `SafetyExtensions.IgnoreExceptions` helper, strictly targeting `COMExceptions` to ensure the app doesn't crash during drag race conditions.

**Before**:

https://github.com/user-attachments/assets/b77989dd-ea9e-4ed6-a8d3-3e2b297324be

**After**:

https://github.com/user-attachments/assets/20596055-a075-415a-8a50-118cb54c2bfb

1. Opened Files
2. Pinned multiple folders to the sidebar
3. Dragged a pinned folder to a different position in the sidebar
4. Verified the folder drops and stays in the new position naturally
5. Checked the Home Page Quick Access widget to confirm it matches the new order without flickering or reloading
6. Opened Windows File Explorer in the background
7. Started dragging a pinned folder in the sidebar to reorder it, and then quickly grabbed another sidebar item while the first drag was resolving
8. Confirmed the app securely handles the underlying `COMException` and no longer crashes
9. Restarted the app to verify the new pinned folder order persists


yair100:
@kalmix I haven't reviewed the code yet, but I tested the changes. It works pretty well but I noticed that I can't drag Recycle Bin to a different position. I'm not sure if this is specific to Recycle Bin or if it's because it's the last item in the list.

As a side note, can you remove the old content dialog since it's no longer needed?

kalmix:
@yair100 Sorry, I missed `Shell:` paths like Recycle Bin. I fixed it now; I also removed the old dialog.

kalmix:
I've also updated the "drop on center" behavior to better align with the Explorer. For pinned items, dropping on the center now performs a Create Link action instead of a move operation, while non-pinned items continue to be moved as before. In addition, I added support for drag-and-drop pinning.

Josh65-2201:
Seems to be no way to drag an item to the first position. I have to drag the item to the second from top and then the first one down. 

Also has issues keeping Gallery pined when the list is updated
```
\SHELL\OgAfAOTpNjks2e5OqFq8FtXqCBkmAAEAJgDvvhAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAUAAAA\_AAAAPYAycAeDOsAAQAAVgAAADFTUFOmamMoPZXSEbXWAMBP2RjQKQAAACAAAAAAERAAABYAAAAUAB9B6mWI6BwOIE6apu3NAhLIfAAAAAARAAAAGQAAAAATAAAAAAAAIAAAAABgAAAAMVNQU_TuyDAyqOJBqzLjw8oo_SkRAAAAEwAAAAATAAAAAAAAABEAAAAEAAAAAAsAAAD__wAAEQAAAAMAAAAAEwAAAAUAAAARAAAAFAAAAAALAAAAAAAAAAAAAAAxAAAAMVNQU35WhU_w__VNsdmYsxT_BykVAAAACAAAAAAfAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAA=
```

kalmix:
@Josh65-2201 Yes, sorry I blocked users from draging to the top because I was running into a UI virtualization bug that left a blank space and also because the file explorer doesn't allow you to do that either but now i've fixed it using `Manager_DataChanged(`) instead of `Move()`, and since we're already "reseting" the pinned section we can get around that.

Regarding the Gallery issue, thanks for pointing that out, in this case Gallery is an outlier that also causes problem with the current state of Files (try it on the prod app), I gave my best shot at fixing this by normalizing the PIDL paths and also adding a fallback remove COM verb and then to keep everything in sync I added a FileSystemWatcher that tracks the Quick Access registry.

Josh65-2201:
Working perfectly now.

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
```diff
@@ -47,6 +48,8 @@ public IFilesystemHelpers FilesystemHelpers
 		public PinnedFoldersManager SidebarPinnedModel => App.QuickAccessManager.Model;
 		public IQuickAccessService QuickAccessService { get; } = Ioc.Default.GetRequiredService<IQuickAccessService>();
 
+		private readonly SemaphoreSlim reorderSemaphore = new(1, 1);
```
marcelwgn: Why do we need a semaphore?

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -474,28 +474,38 @@ private void SidebarControl_ItemContextInvoked(object sender, ItemContextInvoked
...
 			{
-				await SidebarAdaptiveViewModel.HandleItemDroppedAsync(e);
-			}, App.Logger);
+				deferral = e.RawEvent.GetDeferral();
 
-			deferral.Complete();
+				await SafetyExtensions.IgnoreExceptions(
+					() => SidebarAdaptiveViewModel.HandleItemDroppedAsync(e), App.Logger);
+			}
+			catch (Exception ex)
```
marcelwgn: What error case are you expecting to catch here? SafetyExtensions.IgnoreExceptions is already catching things, do we think the deferral wil throw?

src/Files.App/Utils/Storage/Operations/FilesystemHelpers.cs
```diff
@@ -779,7 +812,18 @@ public static async Task<IEnumerable<IStorageItemWithPath>> GetDraggedStorageIte
 
 			// workaround for GetStorageItemsAsync() bug that only yields 16 items at most
 			// https://learn.microsoft.com/windows/win32/shell/clipboard#cf_hdrop
-			if (packageView.Contains("FileDrop"))
+			bool containsFileDrop;
+			try
+			{
+				containsFileDrop = packageView.Contains("FileDrop");
```
marcelwgn: Fairly sure this shouldnt throw

src/Files.App.Controls/Sidebar/SidebarView.xaml.cs
```diff
@@ -53,13 +54,36 @@ internal void RaiseContextRequested(SidebarItem item, Point e)
...
+				dataView = rawEvent.DataView;
+			}
+			catch (Exception ex) when (DragDropExceptionHelper.IsExpectedStaleDragData(ex))
+			{
+				DragDropExceptionHelper.LogStaleDrag(ex, "Stale OLE drag payload reading DataView in RaiseItemDragOverAsync.");
+				return;
+			}
+			var args = new ItemDragOverEventArgs(sideBarItem.Item, dataView, dropPosition, rawEvent);
+			ItemDragOver?.Invoke(this, args);
+			if (args.CompletionTask is not null)
```
marcelwgn: This is a suspicious pattern, why does the handler not call its own completion task?

src/Files.App.Controls/Sidebar/ISidebarViewModel.cs
```diff
@@ -9,6 +9,12 @@ namespace Files.App.Controls
 {
 	public record ItemInvokedEventArgs(PointerUpdateKind PointerUpdateKind) { }
 	public record ItemDroppedEventArgs(object DropTarget, DataPackageView DroppedItem, SidebarItemDropPosition dropPosition, DragEventArgs RawEvent) { }
-	public record ItemDragOverEventArgs(object DropTarget, DataPackageView DroppedItem, SidebarItemDropPosition dropPosition, DragEventArgs RawEvent) { }
+	public record ItemDragOverEventArgs(object DropTarget, DataPackageView DroppedItem, SidebarItemDropPosition dropPosition, DragEventArgs RawEvent)
+	{
+		/// <summary>
+		/// Set by the event handler to signal async completion
+		/// </summary>
+		public Task? CompletionTask { get; set; }
```
marcelwgn: This is very suspicious, both from the fact that it is settable and also is just passing around things to execute. What is the problem we are trying to solve here?

src/Files.App.Controls/Sidebar/SidebarView.xaml.cs
```diff
@@ -53,13 +54,36 @@ internal void RaiseContextRequested(SidebarItem item, Point e)
...
+			try
+			{
+				dataView = rawEvent.DataView;
+			}
+			catch (Exception ex) when (DragDropExceptionHelper.IsExpectedStaleDragData(ex))
+			{
+				DragDropExceptionHelper.LogStaleDrag(ex, "Stale OLE drag payload reading DataView in RaiseItemDropped.");
+				return;
+			}
+			ItemDropped?.Invoke(this, new(sideBarItem.Item, dataView, dropPosition, rawEvent));
```
marcelwgn: Two things:
a) Do we expect rawEvent.DataView to throw?
b) Why not move the invocation of ItemDropped into the try catch, as is this is hard to read.

src/Files.App/Services/Windows/WindowsQuickAccessService.cs
```diff
@@ -41,74 +83,317 @@ private async Task PinToSidebarAsync(string[] folderPaths, bool doUpdateQuickAcc
...
+			try
 			{
-				string pathStr = (string)fi.Path;
+				Type? shellAppType = Type.GetTypeFromProgID("Shell.Application");
+				if (shellAppType == null)
+					return;
 
-				if (ShellStorageFolder.IsShellPath(pathStr))
+				object? shell = Activator.CreateInstance(shellAppType);
+				dynamic? f2 = shellAppType.InvokeMember("NameSpace", System.Reflection.BindingFlags.InvokeMethod, null, shell, [$"shell:{guid}"]);
```
marcelwgn: Dynamic seems very fishy to use, if you are calling .Items() on it, it seems you have a very good idea of what you are expecting here.

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
```diff
@@ -280,23 +282,30 @@ private async void Manager_DataChanged(object sender, NotifyCollectionChangedEve
 			if (dispatcherQueue is null)
 				return;
 
-			await dispatcherQueue.EnqueueOrInvokeAsync(async () =>
+			try
```
marcelwgn: This try catch almost certainly is not doing what you want it to unless you expect await dispatcherQueue.EnqueueOrInvokeAsync to actually throw

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
```diff
@@ -324,6 +333,29 @@ private async Task SyncSidebarItemsAsync(LocationItem section, Func<IReadOnlyLis
 					}
 
 				case NotifyCollectionChangedAction.Move:
+					{
+						if (e.OldItems?.Count == 1 && e.NewItems?.Count == 1)
```
marcelwgn, line 395: If this is not the case, our view now just stays outdated?

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
```diff
@@ -1286,9 +1383,144 @@ public async Task HandleItemDroppedAsync(ItemDroppedEventArgs args)
...
+						if (locationItem.IsHeader)
+							return;
+
+						if (args.dropPosition == SidebarItemDropPosition.Center)
+						{
+							await FilesystemHelpers.PerformOperationTypeAsync(DataPackageOperation.Link, args.DroppedItem, locationItem.Path, false, true);
+							return;
+						}
+
+						await reorderSemaphore.WaitAsync();
```
marcelwgn: Why are we waiting here and what are we expecting to be done after this wait?

marcelwgn (changes requested):
As a general thought, why are we now introducing a bunch of try catches for functions that should REALLY not throw and why not leverage the SafetyExtensions for this?

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
kalmix: Nope, I'll remove this and rely on SuspendSync() instead.

src/Files.App/Services/Windows/WindowsQuickAccessService.cs
kalmix: Lazy shortcut, will fix this.

src/Files.App.Controls/Sidebar/SidebarView.xaml.cs
kalmix: a) yes, it throws COMExceptions occasionally if the OLE payload goes stale.
b) Good point.

src/Files.App/ViewModels/UserControls/SidebarViewModel.cs
kalmix: I originally added the semaphore as an async lock to prevent race conditions from corrupting the ChildItems ObservableCollection. I'll use isReordering instead...

src/Files.App/Views/MainPage.xaml.cs
kalmix: GetDeferral() and Complete() can both throw COMException when the OLE drag payload goes stale mid-operation (common when Explorer is open).

marcelwgn (changes requested):
I think we should only try/catch what we expect to fail, a lot of the try catches catch a more than is necessary/expected to throw, e.g. SideBarItem.cs:440 or SideBarItem.cs:499.

Also regarding all of these try catches, what is the need for the try catches based on? I couldnt find any documentation saying that those are expected to throw and the fact that you introduced them now in a lot of places that we didnt have them before is a bit suspicious to me.

Adding to that, what do you mean by this part in the PR description:
> Replaced the previous app-level COM exception workaround [...]

yair100:
@kalmix

kalmix:
> I think we should only try/catch what we expect to fail, a lot of the try catches catch a more than is necessary/expected to throw, e.g. SideBarItem.cs:440 or SideBarItem.cs:499.
> 
> Also regarding all of these try catches, what is the need for the try catches based on? I couldnt find any documentation saying that those are expected to throw and the fact that you introduced them now in a lot of places that we didnt have them before is a bit suspicious to me.
> 
> Adding to that, what do you mean by this part in the PR description:
> 
> > Replaced the previous app-level COM exception workaround [...]

@marcelwgn You make a very fair point. I just pushed a new commit that completely cleans up all of those broad try-catches.

To answer your question on why those exceptions were happening at all, during drag-and-drop operations (especially when reordering pinned items while the background Quick Access syncs are running) Windows occasionally invalidates the OLE drag payload mid-flight. If the UI tries to read a property like `e.GetDeferral()` from that stale payload, WinUI throws a hard `COMException` that crashes the app. Since WinUI doesn't provide a way to check if an external drag object is still valid, catching that specific `COMException` when reading the payload is the only way to avoid a crash when that race condition triggers. I've added comments in the code to document this so it's clear for anyone reading it in the future.

You were totally right that catching broad Exceptions or wrapping huge blocks of code was the wrong approach. I've switched everything over to use the existing `SafetyExtensions.IgnoreExceptions`. Now, it only wraps the exact single-line property read that fails, and it strictly targets `typeof(COMException)` so we aren't swallowing actual bugs anymore. I also removed the broad try-catches that were hiding in the background sync logic.

Let me know what you think of the latest commit.

marcelwgn (approved):
Sorry for the delay, changes look good to me. Thank you for your help with this!

src/Files.App/Views/MainPage.xaml.cs
yair100, line 1: @kalmix can you please explain the changes to this file?

yair100:
I noticed the dragging a file to the pinned section shows the reorder visual. This should probably be hidden when dragging files.

src/Files.App/Views/MainPage.xaml.cs
kalmix: Sure, I added targeted exception handling to the drag-and-drop handlers to prevent the app from crashing when drag operations are invalidated. Calling GetDeferral or Complete on a drag payload throws a COMException if the underlying OLE payload goes stale mid-flight (for example, when a background Quick Access sync refreshes the sidebar elements while a drag/drop operation is actively resolving). I wrapped those calls in SafetyExtensions to ignore just COMExceptions specifically.

kalmix:
Noted. Working on it.

src/Files.App/Views/MainPage.xaml.cs
yair100: It looks like there were other changes as well. Please take a look.

kalmix: @yair100 Yes, sorry about that. It looks like those other changes were from an accidental merge regression. I've restored it in my latest commit.

kalmix: I've looked back and the issue was a merge conflict resolution mismatch when merging main that mistakenly kept my branch's version of the file.

kalmix: using `typeof(COMException)` is still a good idea here.

yair100:
Thanks! I noticed another issue where reordering items sometimes shifts existing sidebar items to the right.

Josh65-2201:
Not sure if there intended but dragging a pin over another pin will make a shortcut in the hovered folder or open it

Josh65-2201 (dismissed):

yair100:
@Josh65-2201 that isn't intended.

kalmix:
> Thanks! I noticed another issue where reordering items sometimes shifts existing sidebar items to the right. 

@yair100 Could I get steps to reproduce this issue?

kalmix:
> Not sure if there intended but dragging a pin over another pin will make a shortcut in the hovered folder or open it

@yair100 @Josh65-2201 I intentionally set it to create a shortcut because it's the safest UX. Before this PR, accidentally dropping a pin onto another pin triggered a Move operation, which can trigger a lot of accidental move operations for example of a drive to another... and it's a much less common use case.

yair100:
> accidentally dropping a pin onto another pin triggered a Move operation, which can trigger a lot of accidental move operations for example of a drive to another...

We can have a discussion about this behavior in general but it's out of scope for this PR.

kalmix:
> > accidentally dropping a pin onto another pin triggered a Move operation, which can trigger a lot of accidental move operations for example of a drive to another...
> 
> We can have a discussion about this behavior in general but it's out of scope for this PR.

Noted. Will revert those changes.

yair100:
Thanks!

src/Files.App.Controls/Sidebar/ISidebarItemModel.cs
```diff
@@ -30,5 +30,10 @@ public interface ISidebarItemModel : INotifyPropertyChanged
 		/// Expansion participant that keeps the regular row appearance (icon + normal text) instead of the section-header style.
 		/// </summary>
 		bool IsLeafWithChildren => false;
+
+		/// <summary>
+		/// Indicates whether the item supports reorder dropping.
+		/// </summary>
+		bool IsReorderDropItem => false;
```
0x5bfa, line 37: CanBeReordered?

src/Files.App.Controls/Sidebar/SidebarItem.cs
```diff
@@ -190,32 +206,36 @@ private void Item_PropertyChanged(object? sender, PropertyChangedEventArgs e)
 			}
 		}
 
+		private static bool IsValidDropPath(string? path)
+			=> path is not null && (System.IO.Path.IsPathRooted(path) || path.StartsWith("Shell:", StringComparison.OrdinalIgnoreCase));
```
0x5bfa, line 210: Please do not put system specific code inside Files.App.Controls. It should be done within Files.App. Sidebar reordering should work in any other scenario.

0x5bfa (changes requested):
Drag and drop stuff should be similar implementation as the WinUI one. We dont want to handle drop validation inside a control. Have to keep in mind that these controls have to be standalone and any of functionalitis of them have to be able to be used in any scenario.

When you try to drag an item but drag back to the original position, it shows a dialog saying this. This can be a bad UX.

Image

src/Files.App.Controls/Sidebar/ISidebarItemModel.cs
kalmix: @yair100 This is the kind of behavior I was trying to avoid with Center Drop => Create Shortcut.

src/Files.App.Controls/Sidebar/SidebarItem.cs
kalmix: @0x5bfa Thanks for the review, I will work on it this week.

src/Files.App.Controls/Sidebar/ISidebarItemModel.cs
hez2010: > CanBeReordered?

Or just simply `CanReorder`.

yair100:
I opened a PR based on this branch #18951.

PR #18322, opened 2026-03-29
Code Quality: Simplified logic for adaptive layout

Josh65-2201:
**Resolved / Related Issues**

- Closes https://discord.com/channels/725513575971684472/1487913349017636944

1. Opened video, audio and image only folders they went to grid layout
2. Opened just folders and mixed media (image and video/audio less that 85% of the total contents) they went to details layout

Added .ogg (.oga is the same format just different file extension) to the isAudio check. Was only in the isVideo check and as it can have video in it (although rarely) I have kept it there.

Note that the existing isMedia checks audio and video files in a folder. Since audio can have album art for the thumbnail I believe it's fine to have them also use grid layout

### DIFF
--- a/src/Files.App/Helpers/Layout/AdaptiveLayoutHelpers.cs
+++ b/src/Files.App/Helpers/Layout/AdaptiveLayoutHelpers.cs
@@ -4,7 +4,6 @@
 using Files.App.ViewModels.Previews;
 using Files.Shared.Helpers;
 using Windows.Storage;
-using static Files.App.Constants.AdaptiveLayout;
 
 namespace Files.App.Helpers
 {
@@ -70,7 +69,7 @@ namespace Files.App.Helpers
 			float imagePercentage = 100f * filesAndFolders.Count(IsImage) / itemCount;
 			float mediaPercentage = 100f * filesAndFolders.Count(IsMedia) / itemCount;
 
-			if (imagePercentage > MediumThreshold || mediaPercentage > MediumThreshold)
+			if (imagePercentage > 60.0f || mediaPercentage > 60.0f)
 				return Layouts.Grid;
 			return Layouts.Detail;
 
--- a/src/Files.App/Helpers/Layout/AdaptiveLayoutHelpers.cs
+++ b/src/Files.App/Helpers/Layout/AdaptiveLayoutHelpers.cs
@@ -70,7 +70,7 @@ namespace Files.App.Helpers
 			float imagePercentage = 100f * filesAndFolders.Count(IsImage) / itemCount;
 			float mediaPercentage = 100f * filesAndFolders.Count(IsMedia) / itemCount;
 
-			if (imagePercentage > ExtraLargeThreshold || mediaPercentage > ExtraLargeThreshold)
+			if (imagePercentage > MediumThreshold || mediaPercentage > MediumThreshold)
 				return Layouts.Grid;
 			return Layouts.Detail;
 
--- a/src/Files.App/Helpers/Layout/AdaptiveLayoutHelpers.cs
+++ b/src/Files.App/Helpers/Layout/AdaptiveLayoutHelpers.cs
@@ -67,22 +67,12 @@ namespace Files.App.Helpers
 			if (filesAndFolders.Count is 0)
 				return Layouts.None;
 
-			float folderPercentage = 100f * filesAndFolders.Count(IsFolder) / itemCount;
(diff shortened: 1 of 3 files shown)


src/Files.App/Helpers/Layout/AdaptiveLayoutHelpers.cs
```diff
@@ -67,25 +67,12 @@ private static Layouts GetContentLayout(IList<ListedItem> filesAndFolders)
...
 
-			float folderPercentage = 100f * filesAndFolders.Count(IsFolder) / itemCount;
 			float imagePercentage = 100f * filesAndFolders.Count(IsImage) / itemCount;
 			float mediaPercentage = 100f * filesAndFolders.Count(IsMedia) / itemCount;
-			float miscPercentage = 100f - (folderPercentage + imagePercentage + mediaPercentage);
 
-			if (folderPercentage + miscPercentage > LargeThreshold)
-				return Layouts.Detail;
-			if (imagePercentage > ExtraLargeThreshold)
+			if (imagePercentage > ExtraLargeThreshold || mediaPercentage > ExtraLargeThreshold)
```
yair100: The if statement should also make sure `folderPercentage + miscPercentage > LargeThreshold` is false. Otherwise, it will incorrectly switch to grid mode.

src/Files.Shared/Helpers/FileExtensionHelpers.cs
```diff
@@ -77,7 +77,7 @@ public static bool IsCompatibleToSetAsWindowsWallpaper(string? fileExtensionToCh
 		/// <returns><c>true</c> if the fileExtensionToCheck is an audio file; otherwise, <c>false</c>.</returns>
 		public static bool IsAudioFile(string? fileExtensionToCheck)
 		{
-			return HasExtension(fileExtensionToCheck, ".mp3", ".m4a", ".wav", ".wma", ".aac", ".adt", ".adts", ".cda", ".flac");
+			return HasExtension(fileExtensionToCheck, ".mp3", ".m4a", ".ogg", ".oga", ".wav", ".wma", ".aac", ".adt", ".adts", ".cda", ".flac");
```
yair100, line 80: This list is also used to enable album cover editing in the properties window and the media player in the preview pane. Did you check that both these formats work with those features?

Josh65-2201: .ogg play in the info pane preview already in main and still does with this change as well as .oga. Settings a cover image works with .ogg/.oga too.

src/Files.App/Helpers/Layout/AdaptiveLayoutHelpers.cs
Josh65-2201: As far as I can tell it checks how many folder and none media items there is compared to media/images, If that was above large threshold (>80%) then imagePercentage and mediaPercentage would be less than 20% combined and wouldn't pass ExtraLargeThreshold (>85%) anyway as it's already making a percentage based on how many total items there are in the folder.

yair100: You’re right.

One other thing, previously, when the image percentage was between 60% and 75%, the layout switched to Grid. It seems odd to me then that 75% to 85% would use Details, it would make more sense if we simply used Grid for anything above 60%.

Josh65-2201: ExtraLargeThreshold (85%) is what main currently uses, Do you want it changed to medium (I think that 60%)?

yair100: I think that makes more sense, I'm just curious why the original behavior excluded 75% - 85%.

Josh65-2201: Probably because it default was grid so it worked as intended in use

Josh65-2201: Set the check to medium (60%)

yair100: That could be

Josh65-2201:
Could also remove the below as it's now only using one value from it for adaptive layout helper
image

yair100:
Agreed

yair100 (approved):
LGTM

PR #18346, opened 2026-04-06
Code Quality: Improved HotKey.GetKeyCharacter

### LINKED ISSUE #18345: Bug: Incorrect HotKey rendering
### Description

HotKeys are rendered incorrectly.

For example, CTRL + OEM3, the key combination for the `OpenTerminalAction`, is displayed on my machine as "Ctrl+Ù". That is not correct, it should be "Ctrl+ù". Indeed, there is no "Ù" key on my keyboard, and the character requires multiple keystrokes to be typed.

The bug is caused by the implementation of `HotKey.GetKeyCharacter`, and the fact that its result is capitalized using `string.ToUpper()`.

### Steps To Reproduce

1. Run Files on a machine with a configured keyboard layout where OEM keys are not capitalized.
2. Observe that HotKeys using OEM keys are renderend with a capitalized key character.

### Files Version

4.0.38.0

### Windows Version

10.0.26200.7462

### User ID

02442ba2-f9ef-489e-bd00-444220895388

### Log File

N/A

KrisVandermotten:
**Resolved / Related Issues**

- Closes #18345 

The new implementation is not only correct where the previous one wasn't, it is also much more efficient.

The new implementation will, for each of the 40 calls:
- Make one P/Invoke call requiring no marshalling of parameters or return values;
- Allocate one string for existing keys.

The previous implementation would, for each of the 40 calls:
- Allocate a 256 byte array;
- Allocate a StringBuilder, which in turn allocates a 4 char array;
- Perform 4 P/Invoke calls, one of them with complex parameter marshalling;
- Allocate a string for existing keys;
- Capitalize that string resulting in another string allocation for each existing key.

There is no need for any of that. The [MapVirtualKey](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-mapvirtualkeyw) function can do the conversion directly (and correctly) when used with the `MAPVK_VK_TO_CHAR` flag.

1. Opened Files on a machine with a configured keyboard layout where OEM keys are not capitalized.
2. Observed that all hotkeys are now renderend correctly, including those using OEM keys.
3. Verified that hotkeys still work correctly, including those that use OEM keys (such as the hotkey for Open Terminal).
4. Ran Files under the debugger and observed that `HotKey.GetKeyCharacter` returns the same strings as before for all 40 calls, except that they are now all correctly capitalized (including the simple alphabetic keys `A` to `Z`).

### DIFF
--- a/src/Files.App.CsWin32/NativeMethods.txt
+++ b/src/Files.App.CsWin32/NativeMethods.txt
@@ -223,9 +223,7 @@ _SICHINTF
 RoGetAgileReference
 IQueryInfo
 QITIPF_FLAGS
-GetKeyboardState
 MapVirtualKey
-GetKeyboardLayout
 S_FALSE
 IExecuteCommand
 IObjectWithSelection
--- a/src/Files.App/Data/Commands/HotKey/HotKey.cs
+++ b/src/Files.App/Data/Commands/HotKey/HotKey.cs
@@ -4,6 +4,7 @@
 using System.Collections.Frozen;
 using System.Text;
 using Windows.Win32;
+using Windows.Win32.UI.Input.KeyboardAndMouse;
 using Forms = System.Windows.Forms;
 
 namespace Files.App.Data.Commands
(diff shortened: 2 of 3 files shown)


0x5bfa (approved):
Lgtm.

PR #18356, opened 2026-04-12
GitHub: Mass-update workflow dependencies

A-Loot:
A lot of the GitHub Actions dependencies haven't been updated in quite a while, so this PR is meant to update all dependencies to their latest major version. All Actions now use Node 24 instead of Node 20, which will reach EOL by the end of this month.

I also noticed that the YAML wasn't formatted correctly, so I fixed that, too. If necessary, the changes to formatting can be reverted.

**Resolved / Related Issues**

None. This PR simply updates dependencies.

Testing most of the workflows is not easily possible. I only properly tested `ci.yml`, as it does not rely on Secrets and can be enabled on a fork by simply removing the org check.

Instead, I read through all relevant release notes for all dependencies to find breaking changes that may affect the workflows. Luckily, it seems like none of the breaking changes actually break any of the wokflows.

### DIFF
--- a/.github/workflows/cd-sideload-preview.yml
+++ b/.github/workflows/cd-sideload-preview.yml
@@ -159,7 +159,7 @@ jobs:
         timestamp-digest: SHA256
 
     - name: Upload packages to Cloudflare
-      uses: ryand56/r2-upload-action@v1
+      uses: ryand56/r2-upload-action@latest
       with:
         r2-account-id: ${{ secrets.R2_ACCOUNT_ID }}
         r2-access-key-id: ${{ secrets.R2_ACCESS_KEY_ID }}
--- a/.github/workflows/cd-sideload-preview.yml
+++ b/.github/workflows/cd-sideload-preview.yml
@@ -33,7 +33,7 @@ jobs:
       WORKING_DIR:                '${{ github.workspace }}' # D:\a\Files\Files\
       ARTIFACTS_STAGING_DIR:      '${{ github.workspace }}\artifacts'
       APPX_PACKAGE_DIR:           '${{ github.workspace }}\artifacts\AppxPackages\'
-      APP_PROJECT_PATH:            'src\Files.App\Files.App.csproj'
+      APP_PROJECT_PATH:           'src\Files.App\Files.App.csproj'
       PACKAGE_MANIFEST_PATH:      'src\Files.App\Package.appxmanifest'
       LAUNCHER_PROJECT_PATH:      'src\Files.App.Launcher\Files.App.Launcher.vcxproj'
       TEST_PROJECT_PATH:          'tests\Files.InteractionTests\Files.InteractionTests.csproj'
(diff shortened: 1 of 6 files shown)


0x5bfa (commented):
There are many unrelated changes. Please revert whitespace changes.

A-Loot:
I reverted the indentation changes, but kept the changes to trailing whitespace. If necessary, I can undo these, too.

yair100:
We can update the first party GitHub dependencies, but I think all third-party dependencies should be kept as is for security reasons.

A-Loot:
Which third-party actions are you talking about exactly? Assuming actions from `actions`, `Azure`, `microsoft`, and `NuGet` all count as first-party:

- `ryand56/r2-upload-action`: The change from `@latest` to `@v1` does not actually change the underlying version (v1.4). While it is a subjective choice, switching away from `@latest` should prevent potential issues with future SemVer-major (breaking) changes by requiring the major version to be explicitly bumped in the workflows.
- `nick-fields/retry`: The only major/breaking change from `@v3` to `@v4` is the update from Node 20 to Node 24.
- `GeekyEggo/delete-artifact`: The only major/breaking changes from `@v5` to `@v6` are the update from Node 20 to Node 24 and the switch from CommonJS to ESM, which shouldn't affect the workflows.

Switching from Node 20 to 24 is important to ensure the workflows don't break when GitHub deprecates Node 20 on June 2nd, 2026 (https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/).

yair100:
> Which third-party actions are you talking about exactly? Assuming actions from actions, Azure, microsoft, and NuGet all count as first-party:

Correct. All other changes need to wait.

.github/workflows/cd-sideload-preview.yml
```diff
@@ -159,7 +159,7 @@ jobs:
         timestamp-digest: SHA256
 
     - name: Upload packages to Cloudflare
-      uses: ryand56/r2-upload-action@latest
+      uses: ryand56/r2-upload-action@v1
```
0x5bfa: This is a 3rd party. Please revert.

.github/workflows/cd-sideload-stable.yml
```diff
@@ -159,17 +159,17 @@ jobs:
         timestamp-digest: SHA256
 
     - name: Upload packages to Cloudflare
-      uses: ryand56/r2-upload-action@latest
+      uses: ryand56/r2-upload-action@v1
```
0x5bfa: Above.

.github/workflows/ci.yml
```diff
@@ -251,7 +251,7 @@ jobs:
 
     # Retry integration tests if first attempt fails
     - name: Run interaction tests
-      uses: nick-fields/retry@v3
+      uses: nick-fields/retry@v4
```
0x5bfa: Above.

.github/workflows/ci.yml
```diff
@@ -265,9 +265,9 @@ jobs:
             --report-trx-filename testResults.trx
 
     - if: github.event_name == 'pull_request'
-      uses: geekyeggo/delete-artifact@v5
+      uses: GeekyEggo/delete-artifact@v6
```
0x5bfa: Above.

0x5bfa (changes requested):
All of others, looks good to me.

0x5bfa (approved):
LGTM!

yair100 (approved):

Issue #17128, opened 2025-05-13
Bug: Active tags widget = massive slowdown


RinineFV2:
### Description

I thought it was some known general performance issue, but I realized it's the tags widget's fault.
If it's active, when I open the app it freezes for no less than 20 seconds. (On a 14700k).

And it always loads all tag elements in duplicate.

### Steps To Reproduce

The exact steps to reproduce the issue are:

1. Configure the files app to load the home page, or continue a session, or load pages that include the home page
2. Have the tags widget enabled
3. Have several tags with elements (6 tags in my case, with about 4–5 elements per tag, nothing excessive)
4. Start a new instance from scratch (not running in the background)

Then the app becomes unresponsive for 20 seconds or more until the tag widget finishes "doing its thing" (leaving all items duplicated as well).

If I disable the tag widget, everything works fine, even when creating new instances.

### Files Version

3.9.9.0

### Windows Version

10.0.26100.0

### User ID

9baec1c2-1793-4175-a258-57a34bc58c31

### Log File

No logs needed.

RinineFV2:
I only use the preview version so I don’t know, but I stopped using the app in version 3.9.2 because of this issue.
When I saw the update to 3.9.9, I tested it to see if it had been fixed, and that’s when I discovered the tag widget was the culprit.

Given how old it is, I can confirm it’s not a bug introduced in recent versions, but something that was already there.

Josh65-2201:
Please try reproducing it in stable https://files.community/download

Josh65-2201:
@RinineFV2 Can you answer the above question

RinineFV2:
Testing in the stable version, the issue doesn’t occur, but I’m not sure if that’s because it’s an earlier version than 3.9.2 (I see it’s 3.9.1) or due to another reason, as I explain below.

When exporting the configuration from the preview version and importing it into the stable version, the tags are missing, but the tagged files appear with a tag labeled “unknown” (and the stable version crashed once).
Since my tags weren’t there, I created new ones, and the error didn’t occur.

I went back to the preview version, to my first tag “Main” (which had about 10 files/directories associated with it), and opened the first one, “Downloads.” I navigated up one level (to my user folder) to view the Downloads directory and saw that it didn’t display any tag beside it. After reapplying the tag, all items previously tagged as “Main” were removed.

So the bug might happen due to some kind of tag corruption.
I’m going to try recreating my tags and reassigning them from scratch in the preview version, but it’ll take me a bit of time.

[...]

PR #18390, opened 2026-04-20
Fix: Fixed an issue where tags widget had duplicate items

workbysaran:
**Resolved / Related Issues**
Fixed an issue where tagged items in the home page tags widget appeared duplicated on first load. 

refs #17128

1. Enable the Tags widget on the home page
2. Create several tags with a few tagged files/folders each
3. Close the app completely (not running in background)
4. Launch a fresh instance and verify items appear once, not duplicated
5. Navigate away from home and back and verify no duplicates on return


yair100 (approved):
LGTM

PR #18409, opened 2026-04-22
Fix: Fixed home page freeze caused by active tags widget

workbysaran:
**Resolved / Related Issues**
Fixed an issue where the home page would freeze or load slowly when the file tags widget was active and large number of files tagged, caused by the widget performing a separate database scan for each tag.

refs #17128

1. Open the app with the file tags widget enabled on the home page and verify it loads without freezing
2. Add multiple tags with several files each and verify all containers populate correctly
3. Tag a new file and check it appears in the widget after refresh


yair100 (approved):
LGTM

yair100:
@workbysaran were you able to reproduce the issue in #17128? If this fixes it, can we close the issue?

workbysaran:
@yair100 Yes, I was able to reproduce issue #17128 with a large number of tags on the Home page, and this PR is able to fix the issue.

Issue #18420, opened 2026-04-24
Bug: Slow file selection when Recycle Bin has many files


muko:
### Description

Selecting files can cause noticeable delays, especially when the Recycle Bin contains a large number of items.

`SHQueryRecycleBin()` can be slow (over 1 second) when the Recycle Bin is large, and `ToolBar.PopulateToolbarItems()` is called multiple times per selection leads to UI lag.
```
ToolBar.PopulateToolbarItems()
   -> ...
        -> TrashBinService.QueryRecycleBin()
            -> PInvoke.SHQueryRecycleBin()
```

### Steps To Reproduce

1. Fill the Recycle Bin with a large number of files
2. Execute Files
3. Navigate to a folder
4. Change file selection repeatedly

### Files Version

4.0.40.0

### Windows Version

10.0.26100

### Log File

no log files

Josh65-2201:
Thanks for the feedback, we're tracking this in the below linked issue. You can check there or click subscribe in the right sidebar for updates.

yair100:
@muko are you able to reproduce this in `main`? There is a change that should help but it would be great if you can confirm.

yair100:
> Thanks for the feedback, we're tracking this in the below linked issue. You can check there or click subscribe in the right sidebar for updates.

@Josh65-2201 #18289 is a different issue with the selection rectangle. The issue @muko is reporting is a little different.

muko:
I’ve opened a PR for this: https://github.com/files-community/Files/pull/18422

I linked #18289, bu it should also help with this issue.
It reduces redundant calls to PopulateToolbarItems(), which seem to be a major source of the slowdown.

yair100:
Did you check if this still occurs in `main`? I would hold off creating a PR for now.

muko:
This issue still occurs on the latest main (commit: 1425850a2b8cf1fe5e8a17dc0928d9ad0a08040d).

When PInvoke.SHQueryRecycleBin() returns queryBinInfo.i64NumItems = 3378, the API call takes about 7.5 seconds.

As a result, selecting a file can take up to 40 seconds.

With PR #18422, this is reduced to about 24 seconds, and with PR #18424, it is reduced to under 1 second.

yair100:
Thanks

PR #18424, opened 2026-04-25
Improve Recycle Bin item existence check performance

muko:
**Resolved / Related Issues**

Bug: Slow file selection when Recycle Bin has many files #18420

* When a file is selected, the state updates in under 1 second.
* StorageTrashBinService.HasItems() returns true when the Recycle Bin contains items, and false when it is empty.

### DIFF
--- a/src/Files.App/Services/Storage/StorageTrashBinService.cs
+++ b/src/Files.App/Services/Storage/StorageTrashBinService.cs
@@ -1,7 +1,9 @@
 // Copyright (c) Files Community
 // Licensed under the MIT License.
 
+using System.IO;
 using System.Runtime.InteropServices;
+using System.Security.Principal;
 using Windows.Win32;
 using Windows.Win32.Foundation;
 using Windows.Win32.System.Com;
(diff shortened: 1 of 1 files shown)


src/Files.App/Services/Storage/StorageTrashBinService.cs
```diff
@@ -42,7 +44,34 @@ public ulong GetSize()
 		/// <inheritdoc/>
 		public bool HasItems()
 		{
-			return QueryRecycleBin().NumItems > 0;
```
yair100, line 45: The QueryRecycleBin method is used in other places as well. IF there is an alternative, we should update the method directly so these improvements apply across the board.

muko: QueryRecycleBin() returns whether the Recycle Bin exists, along with the total item count and total size.
Since HasItems() only needs to check whether any items exist, calling QueryRecycleBin() is unnecessary here.

0x5bfa (commented):
I'm very against using the C# APIs over the Shell APIs but if this does improve the performance, lgtm.

PR #18432, opened 2026-04-27
Fix: Fixed issue where double click to rename had a delay

### LINKED ISSUE #17332: Bug: Click to rename has a delay before text box is displayed
### Description

When clicking on a file name to rename it, the filename becomes editable after a very long delay, compared to Windows Explorer where it is almost instant. This is not productive as it introduces a lot of lag when renaming multiple files.

### Steps To Reproduce

1. Click to select a file
2. Click again on the filename to make the filename editable to rename it.

Notice the 1-2 seconds delay between the second click and when the filename becomes editable.

In Windows Explorer the delay is under 1 second. Same in Total Commander, Directory Opus etc.

### Files Version

3.9.1.0 and 3.9.12.0

### Windows Version

10.0.26100.4770

### User ID

_No response_

### Log File

-

SamHatami:
**Resolved / Related Issues**
Replace hardcoded 1500ms rename delay with Windows system double-click time retrieved via GetDoubleClickTime() Win32 API.

The rename delay now respects the user's Windows double-click speed setting (Control Panel → Mouse → Buttons tab), making the app more accessible and consistent with File Explorer behavior.

Changes:
- Added GetDoubleClickTime to CsWin32 NativeMethods.txt
- Replaced hardcoded 1500ms with PInvoke.GetDoubleClickTime() + 150ms buffer
- Extracted 150ms buffer to RenameDelayBufferMs constant

This allows users with motor control issues who set slower double-click speeds system-wide to have Files respect their preference.

Closes #17332

Pre-check the delay before testing this PR and note the delay.

1. Open file, try renaming expected mouse behavior. The delay should be less now.
2. To check if the correct settings is being read from the system, go to Accessibility -> Mouse and change the double click time. And test again.
3. A comparison can be done by pulling up a file explore window and files and check directly.

### DIFF
--- a/src/Files.App.CsWin32/NativeMethods.txt
+++ b/src/Files.App.CsWin32/NativeMethods.txt
@@ -266,3 +266,4 @@ HCERTSTORE
 CERT_QUERY_ENCODING_TYPE
 CertGetNameString
 MessageBox
+GetDoubleClickTime
--- a/src/Files.App/Constants.cs
+++ b/src/Files.App/Constants.cs
@@ -104,6 +104,9 @@ namespace Files.App
 			public const double ContextMenuItemsMaxWidth = 250;
 
 			public const double MultiplePaneWidthThreshold = 750;
+
+			public const int RenameDelayBufferMs = 150;
+
 		}
 
 		public static class Appearance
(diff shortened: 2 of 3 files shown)


yair100:
Thank you for the PR. It appears that there is still a delay before the rename box becomes visible.

yair100:
@SamHatami a pushed a solution that seems to work, can you please take a look? Please feel free to revert if you have a better idea in mind.

SamHatami (commented):
Agreed!

yair100:
Thank you! LGTM

yair100:
@SamHashtami it appears that double click to rename is too sensitive now. I would like to revert this until we can include a patch. Would you like to continue working on this?

PR #18436, opened 2026-04-28
Fix: Fixed issue where thumbnails didn't refresh when file was updated

### LINKED ISSUE #18352: Bug: PDF thumbnail not displaying until refresh
### Description

The update won't start unless you manually press [F5]. This bug has been occurring since the update around last fall.

### Steps To Reproduce

Image

### Files Version

4.0.38

### Windows Version

エディション	Windows 11 Pro バージョン	23H2 インストール日	‎2023/‎07/‎17 OS ビルド	22631.6199 エクスペリエンス	Windows 機能エクスペリエンス パック 1000.22700.1108.0

### User ID

_No response_

### Log File

[debug.log](https://github.com/user-attachments/files/26566153/debug.log)

workbysaran:
**Resolved / Related Issues**
Fixed an issue where files saved to an open folder by external applications (e.g. exporting a PDF from Microsoft Word)  would show only the generic file type icon instead of the actual thumbnail.

Closes #18352

1. Open any folder in Grid View 
2. Open an external application (e.g., Microsoft Word), create a document with a few pages, and export or save it as a PDF in the open folder
3. Verify the PDF thumbnail appears correctly without needing to refresh
4. Repeat with other file types and applications
5. Verify files copied via File Explorer still show thumbnails correctly

### DIFF
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -35,6 +35,7 @@ namespace Files.App.ViewModels
 		private readonly SemaphoreSlim getFileOrFolderSemaphore;
 		private readonly SemaphoreSlim bulkOperationSemaphore;
 		private readonly SemaphoreSlim loadThumbnailSemaphore;
+		private readonly ConcurrentDictionary<string, CancellationTokenSource> thumbnailRetryDebounce;
 		private readonly ConcurrentQueue<(uint Action, string FileName)> operationQueue;
 		private readonly ConcurrentQueue<uint> gitChangesQueue;
 		private readonly ConcurrentDictionary<string, bool> itemLoadQueue;
@@ -561,6 +562,7 @@ namespace Files.App.ViewModels
 			operationQueue = new ConcurrentQueue<(uint Action, string FileName)>();
 			gitChangesQueue = new ConcurrentQueue<uint>();
 			itemLoadQueue = new ConcurrentDictionary<string, bool>();
+			thumbnailRetryDebounce = new ConcurrentDictionary<string, CancellationTokenSource>();
 			addFilesCTS = new CancellationTokenSource();
 			semaphoreCTS = new CancellationTokenSource();
 			loadPropsCTS = new CancellationTokenSource();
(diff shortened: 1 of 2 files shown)


yair100:
What's the underlying issue?

workbysaran:
When an external app (especially one that takes longer to finish writing) writes a file, the Windows thumbnail cache isn’t ready when the file first appears. The initial fetch returns null because the file is still being written, and later modify events don’t refresh the thumbnail.

Fix: If the first fetch returned null, the item was marked NeedsDelayedThumbnailLoad and retried on FILE_ACTION_MODIFIED events with a 500 ms debounce timer.

yair100:
I need to do a more thorough review, but it looks good at first glance.

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -2650,6 +2662,34 @@ private async Task AddFileOrFolderAsync(ListedItem? item)
...
+					{
+						existingCts.Cancel();
+						existingCts.Dispose();
+					}
+
+					var debounceCts = new CancellationTokenSource();
+					thumbnailRetryDebounce[path] = debounceCts;
+					var debounceToken = debounceCts.Token;
+
+					_ = Task.Delay(500, debounceToken)
```
yair100, line 2680: Is this enough of a delay?

workbysaran: We are not just relying on the delay. Each FILE_ACTION_MODIFIED resets a 500ms timer. When no new event arrives within that window, the thumbnail load is retried, and if it still fails, the next event triggers another attempt, repeating until the thumbnail is successfully loaded.

yair100: What if it keeps failing? We don't want a continuous loop.

workbysaran: Retries stop naturally once the external app finishes writing, since no more FILE_ACTION_MODIFIED events will fire.

yair100:
Thank you!

yamachari2025-ops:
The bug where PDF previews do not update has not been fixed.
4.0.51.0
10.0.22631.6199

PR #18467, opened 2026-05-08
Code Quality: Reduce datetime update UI dispatch overhead

muko:
**Resolved / Related Issues**

- #4180

This PR improves date/time display refresh performance by reducing excessive UI-thread queue buildup.
It introduces a unified UI dispatch path and consolidates date property updates into batched/differential updates.
 
- Unified UI dispatch handling to simplify update paths
- Replaced self-assignment style updates with differential updates (update only when the rendered string actually changes)

- Confirmed that date display updates

### DIFF
--- a/src/Files.App/Data/Items/ListedItem.cs
+++ b/src/Files.App/Data/Items/ListedItem.cs
@@ -287,11 +287,26 @@ namespace Files.App.Utils
 
 		public long FileSizeBytes { get; set; }
 
-		public string ItemDateModified { get; private set; }
+		private string itemDateModified;
+		public string ItemDateModified
+		{
+			get => itemDateModified;
+			private set => SetProperty(ref itemDateModified, value);
+		}
 
-		public string ItemDateCreated { get; private set; }
+		private string itemDateCreated;
+		public string ItemDateCreated
+		{
+			get => itemDateCreated;
+			private set => SetProperty(ref itemDateCreated, value);
+		}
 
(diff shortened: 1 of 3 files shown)


yair100:
I recall we used to have a lot of issues with this code, hopefully this change helps.

src/Files.App/Data/Items/ListedItem.cs
```diff
@@ -474,7 +497,14 @@ public RecycleBinItem(string folderRelativeId) : base(folderRelativeId)
 		{
 		}
 
-		public string ItemDateDeleted { get; private set; }
+		private string itemDateDeleted;
+		public string ItemDateDeleted {
+			get => itemDateDeleted;
```
0x5bfa:
```suggestion
		public string ItemDateDeleted
		{
			get => itemDateDeleted;
```

0x5bfa (approved):
I don't have anything to say but if I were to, `ToShortLabel` is called for every date time field every second (tick by dispatcher). and it seems to call `ToString(offset, "D")` which converts to something like `Wednesday, October 31, 2007` (can impact performance). This can be concerning. Just an FYI.

src/Files.App/Data/Items/ListedItem.cs
muko: Thank you for the review. I checked whether similar formatting issues exist in other properties, but found no other occurrences.

yair100:
Thank you.

Issue #18110, opened 2026-01-30
Bug: BitLocker drive does not appear in sidebar or home view when locked


vzabrodin:
### Description

When a BitLocker-encrypted drive is locked, it does not appear in the Files sidebar or the "Drives" section of the Home page. The drive is correctly detected and visible in the native Windows File Explorer as a locked volume, but it is completely missing from the Files app interface until manually unlocked via other means.

<details>
<summary>Screenshots</summary>
Files app

<img width="1854" alt="Files app" src="https://github.com/user-attachments/assets/a302c379-e7ec-473a-aad7-d6a63309edd6" />

Windows Explorer

<img width="1563" height="989" alt="Windows Explorer" src="https://github.com/user-attachments/assets/7b035924-2260-47dc-8875-5e39e6972532" />
</details>

The issue appears to be a regression or oversight related to #17895. Currently, the application likely filters out drives where the `IsReady` property is `false`. Because BitLocker-encrypted volumes return `IsReady: false` while locked, they are excluded from the `WindowsStorageDeviceWatcher` or `RemovableDrivesService` logic.

I’ve also observed inconsistent behavior with SD card adapters that stay connected while media is swapped:
- Inserting an SD card into a previously empty (already connected) adapter does not trigger the drive to appear.
- Removing an SD card while leaving the adapter plugged in leaves a "ghost" drive in the sidebar/widget that results in an error when clicked.

I suggest reverting #17895 to fix this bug and creating a new feature/issue to implement a more sophisticated empty drive detection logic.

### Steps To Reproduce

1. Connect a BitLocker-encrypted external drive (or have a locked internal partition).
2. Ensure the drive is in a Locked state.
3. Open the Files app.
4. Observe that the drive is missing from the sidebar and Home view.
5. Open native Windows File Explorer and observe that the drive appears with a lock icon.

### Files Version

4.0.24.0

### Windows Version

10.0.26200.7623

### User ID

88ef8261-64d0-4208-b774-f43bfe22a0ec

### Log File

[debug.log](https://github.com/user-attachments/files/24973451/debug.log)

Josh65-2201:
Thanks for the report

seve7-china:
I have also encountered this problem.

yair100:
A fix will be included in the next update.

PR #18468, opened 2026-05-08
Fix: Fixed issue where BitLocker locked drive wasn't displayed

muko:
**Resolved / Related Issues**

- #18110

This PR fixes an issue where locked BitLocker drives were not shown.

Note: This PR updates only `WindowsStorageDeviceWatcher`; however, commit 6674a8f also changes `StorageDeviceServices`.

- a locked BitLocker drive is displayed
- an SD card slot without media is not displayed

### DIFF
--- a/src/Files.App/Utils/Global/WindowsStorageDeviceWatcher.cs
+++ b/src/Files.App/Utils/Global/WindowsStorageDeviceWatcher.cs
@@ -52,7 +52,7 @@ namespace Files.App.Utils
 		private async void Win32_OnDeviceAdded(object? sender, DeviceEventArgs e)
 		{
 			var driveAdded = new DriveInfo(e.DeviceId);
-			if (!driveAdded.IsReady)
+			if (!driveAdded.IsReady && !IsUnauthorizedDrive(driveAdded))
 				return;
 
 			var rootAdded = await FilesystemTasks.Wrap(() => StorageFolder.GetFolderFromPathAsync(e.DeviceId).AsTask());
@@ -101,7 +101,7 @@ namespace Files.App.Utils
 			{
 				// Check if this drive is associated with a drive letter
 				var driveAdded = new DriveInfo(root.Path);
-				if (!driveAdded.IsReady)
+				if (!driveAdded.IsReady && !IsUnauthorizedDrive(driveAdded))
 					return;
 
 				type = DriveHelpers.GetDriveType(driveAdded);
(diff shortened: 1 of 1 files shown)


src/Files.App/Utils/Global/WindowsStorageDeviceWatcher.cs
```diff
@@ -52,7 +52,7 @@ private void Win32_OnDeviceRemoved(object? sender, DeviceEventArgs e)
 		private async void Win32_OnDeviceAdded(object? sender, DeviceEventArgs e)
 		{
 			var driveAdded = new DriveInfo(e.DeviceId);
-			if (!driveAdded.IsReady)
+			if (!driveAdded.IsReady && !IsUnauthorizedDrive(driveAdded))
```
d2dyno1, line 55: I think this should be "or" condition (`||`).
If the drive is not ready, `IsUnauthorizedDrive` will still be called, causing an exception on an unready drive

src/Files.App/Utils/Global/WindowsStorageDeviceWatcher.cs
```diff
@@ -101,7 +101,7 @@ private async void Watcher_Added(DeviceWatcher sender, DeviceInformation args)
 			{
 				// Check if this drive is associated with a drive letter
 				var driveAdded = new DriveInfo(root.Path);
-				if (!driveAdded.IsReady)
+				if (!driveAdded.IsReady && !IsUnauthorizedDrive(driveAdded))
```
d2dyno1, line 104: Same here

d2dyno1 (changes requested):

src/Files.App/Utils/Global/WindowsStorageDeviceWatcher.cs
muko: `IsUnauthorizedDrive()` has all code paths wrapped in try-catch, so no exception will be thrown when `IsReady` == false.
In addition, with `||`, a locked BitLocker drive (`IsReady` == false, `IsUnauthorizedDrive()` == true) and a normal drive (IsReady == true, IsUnauthorizedDrive() == false) would both incorrectly return early.

| IsReady | IsUnauthorizedDrive | && (PR) | \|\| (suggested) |
| --- | --- | --- | --- |
| false | false | return | return |
| false | true | continue | return **(WRONG)** |
| true | false | continue | return **(WRONG)** |
| true | true | continue | continue |

d2dyno1 (approved):

yair100:
Thank you

PR #18486, opened 2026-05-14
Fix several issues in Command Palette

### LINKED ISSUE #18395: Bug: Duplicate entries in command list
### Description

Image

Every time you search for a given command, an extra entry appears in the list.

### Steps To Reproduce

1. Open files
2. Navigate to a folder
3. Search for a command, e.g. type "terminal". Observe how the dropdown shows matching commands, but do not execute a command.
4. Click somewhere else, e.g. in the folder pane. Ensure the command palette closes.
5. Search again for the same command. Observe how the dropdown now shows matching commands twice

### Files Version

4.0.38.0

### Windows Version

10.0.26200.8246

### User ID

ad3a123c-a37b-4ac2-a1b1-be4a75740a70

### Log File

-

KrisVandermotten:
**Resolved / Related Issues**

Closes #18395 

- The main problem was a bug in `NavigationToolbarViewModel.UpdateCommandPaletteSuggestions`, more specifically in the `else` branch loop. That `else` branch was completely removed. We now always replace all items in the `OmnibarCommandPaletteModeSuggestionItems` collection. This also results in simpler and more efficient code.
- In fact, `NavigationToolbarViewModel.UpdateCommandPaletteSuggestions` has become so simple that it was inlined into its caller, `PopulateOmnibarSuggestionsForCommandPaletteMode`. Because of that change, we no longer need the intermediate `newSuggestions` collection, saving memory and resulting in a more responsive UI update.
- The number of `await Task.Yield()` calls is reduced. There were too many, resulting is too much pointless overhead.
- Building the `commandsToProcess` list has been optimized, especially for the case where the `OmnibarCommandPaletteModeText` search string is empty, i.e. the first list to be calculated when the command palette is opened.
- Because the calculation of the the `commandsToProcess` list happens on a background thread, we now test whether `OmnibarCommandPaletteModeText` has not changed by the time the list calculation is finished, and prevent further processing of stale results if it did.
- The `NavigationBarSuggestionItem` objects are now immutable. They were never mutated anyway. This also simplifies their construction.
- The `NavigationBarSuggestionItem.ActionIconSource` property was removed as it was never assigned and always had the value `null`.
- The `NavigationBarSuggestionItem.PrimaryDisplay` property was removed as it always had the same value as the `Text` property.
- The `NavigationBarSuggestionItem.SearchText` property was removed as it was never read.
- A `NavigationBarSuggestionItem.Command` property was added, storing the command used to construct the `NavigationBarSuggestionItem`. That property is used in `NavigationToolbar.Omnibar_QuerySubmitted`, eliminating the need to search for the command.
- Because `NavigationBarSuggestionItem` is now immutable, it no longer inherits from `ObservableObject` and it no longer tries to raise pointless `PropertyChanging` and `PropertyChanged` events during its construction.
- Also because `NavigationBarSuggestionItem` is now immutable, data binding in NavigationToolbar.xaml now uses `Mode=OneTime` instead of `Mode=OneWay`. As a result, the UI no longer needs to wire up listeners for events that will never be raised anyway.
- Finally, in `Omnibar.AutoSuggestBox_TextChanged`, we set `_textChangeReason` to `OmnibarTextChangeReason.UserInput` when it's not `OmnibarTextChangeReason.ProgrammaticChange` and `_userInput.Length == 0`. This fixes the bug where clicking the `X` button at the right of the textbox to clear it did not cause a recalculation of the suggestions list.

As a result of the above
- several bugs were fixed,
- the code is significantly simplified,
- and UI responsiveness and memory usage have improved considerably.

1. Opened Files
2. Opened the command palette using Ctrl-Shift-P and tested several scenarios:
   1. Searched for existing commands by typing `terminal` one character at a time, as well as removing characters using Backspace, and observed that suggestions are now correct.
   2. Executed commands after searching for them. Observed that the correct command is still executed when selecting the suggestion. 
   3. Searched for non-existing commands by typing `xyz`. Observed that the `NoCommandsFound` message is still displayed correctly.
   4. Tried to execute non-existing commands such as `xyz`. Observed that the dialog box is still displayed correctly.
   5. Clicked the `X` button in the textbox to clear it. Observed that this now correctly causes suggestions to be recalculated.
   6. Type the full description of a command manually and execute it without selecting a suggestion. Observe that the correct command is still executed.

### DIFF
--- a/src/Files.App.Controls/Omnibar/Omnibar.Events.cs
+++ b/src/Files.App.Controls/Omnibar/Omnibar.Events.cs
@@ -134,13 +134,13 @@ namespace Files.App.Controls
 			if (string.Compare(_textBox.Text, CurrentSelectedMode!.Text, StringComparison.OrdinalIgnoreCase) is not 0)
 				CurrentSelectedMode!.Text = _textBox.Text;
 
-			// UpdateSuggestionListView();
-
 			if (_textChangeReason is OmnibarTextChangeReason.ProgrammaticChange)
 				_textBox.SelectAll();
 			else
 			{
 				_userInput = _textBox.Text;
+				if (_userInput.Length == 0)
+					_textChangeReason = OmnibarTextChangeReason.UserInput;
 			}
 
 			TextChanged?.Invoke(this, new(CurrentSelectedMode, _textChangeReason));
--- a/src/Files.App/Data/Items/NavigationBarSuggestionItem.cs
+++ b/src/Files.App/Data/Items/NavigationBarSuggestionItem.cs
@@ -3,97 +3,60 @@
 
 using Files.App.Controls;
 using Microsoft.UI.Xaml;
-using Microsoft.UI.Xaml.Media;
 
 namespace Files.App.Data.Items
 {
 	[Obsolete("Remove once Omnibar goes out of experimental.")]
-	public sealed partial class NavigationBarSuggestionItem : ObservableObject, IOmnibarTextMemberPathProvider
+	public sealed partial class NavigationBarSuggestionItem : IOmnibarTextMemberPathProvider
 	{
-		private ImageSource? _ActionIconSource;
-		public ImageSource? ActionIconSource { get => _ActionIconSource; set => SetProperty(ref _ActionIconSource, value); }
+		public IRichCommand? Command { get; }
 
-		private Style? _ThemedIconStyle;
-		public Style? ThemedIconStyle { get => _ThemedIconStyle; set => SetProperty(ref _ThemedIconStyle, value); }
+		public Style? ThemedIconStyle { get; }
 
-		private string? _Glyph;
-		public string? Glyph { get => _Glyph; set => SetProperty(ref _Glyph, value); }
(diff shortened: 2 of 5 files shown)


KrisVandermotten:
That failed test seems unrelated to my changes. Can you take a look @yair100 ?

yair100:
The test is passing now. That being said, this PR seems to include a lot of changes that weren't discussed beforehand. It will speed up the review process if this focuses on the fix needed for #18395. We can discuss the other changes as well, but they shouldn't be part of this PR.

KrisVandermotten:
> That being said, this PR seems to include a lot of changes that weren't discussed beforehand.

True. I described them in the description of this PR.

> It will speed up the review process if this focuses on the fix needed for #18395.

I made #18496. It fixes the one issue.
 
> We can discuss the other changes as well, but they shouldn't be part of this PR.

I described them in the description of this PR. Feel free to discuss.

IMHO it's easier to discuss changes when you see them, instead of when they're still an abstract concept.

yair100:
#18496 has been merged into `main`.

PR #18495, opened 2026-05-17
Code Quality: Reduce redundant SelectedItems updates in SelectionChanged

muko:
**Resolved / Related Issues**

- https://github.com/files-community/Files/issues/4180

When selecting files or folders, `SelectionChanged` is raised multiple times even when the actual selection does not change.
For example, when opening a folder by double-clicking, `SelectionChanged` is fired repeatedly without any real change in `SelectedItems`.

The root cause is that `SelectedItems` was always assigned from a new list instance:

```
SelectedItems = FileList.SelectedItems.Cast<ListedItem>().Where(x => x is not null).ToList();
```

Since `ToList()` creates a new list instance on every call, the setter was always triggered and caused unnecessary side effects (such as PropertyChanged-driven updates).

- Open a file by double-clicking
- Open a folder by double-clicking
- Open a folder using keyboard navigation
- Select multiple items by mouse

### DIFF
--- a/src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
+++ b/src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
@@ -362,9 +362,20 @@ namespace Files.App.Views.Layouts
 
 		protected override void FileList_SelectionChanged(object sender, SelectionChangedEventArgs e)
 		{
-			SelectedItems = FileList.SelectedItems.Cast<ListedItem>().Where(x => x is not null).ToList();
+			if (e is null && SelectedItems?.Count == 0)
+				return;
 
-			if (e != null)
+			if (e is not null && e.AddedItems.Count == 0 && e.RemovedItems.Count == 0)
+				return;
+
+			var selectedItems = FileList.SelectedItems.Cast<ListedItem>().Where(x => x is not null).ToList();
+
+			if (SelectedItems is not null && SelectedItems.SequenceEqual(selectedItems))
+				return;
+
+			SelectedItems = selectedItems;
+
+			if (e is not null)
 			{
 				foreach (var item in e.AddedItems)
 					SetCheckboxSelectionState(item);
--- a/src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
+++ b/src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
@@ -360,29 +360,13 @@ namespace Files.App.Views.Layouts
 
 		}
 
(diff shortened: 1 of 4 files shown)


src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
```diff
@@ -362,9 +362,20 @@ private void FolderSettings_LayoutModeChangeRequested(object? sender, LayoutMode
...
 		protected override void FileList_SelectionChanged(object sender, SelectionChangedEventArgs e)
 		{
-			SelectedItems = FileList.SelectedItems.Cast<ListedItem>().Where(x => x is not null).ToList();
+			if (e is null && SelectedItems?.Count == 0)
+				return;
+
+			if (e is not null && e.AddedItems.Count == 0 && e.RemovedItems.Count == 0)
+				return;
+
+			var selectedItems = FileList.SelectedItems.Cast<ListedItem>().Where(x => x is not null).ToList();
```
workbysaran: Since BaseGroupableLayoutPage.FileList_SelectionChanged is the base handler all other layouts use, can we move the SequenceEqual check there instead?

src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
```diff
@@ -362,9 +362,20 @@ private void FolderSettings_LayoutModeChangeRequested(object? sender, LayoutMode
 
 		protected override void FileList_SelectionChanged(object sender, SelectionChangedEventArgs e)
 		{
-			SelectedItems = FileList.SelectedItems.Cast<ListedItem>().Where(x => x is not null).ToList();
+			if (e is null && SelectedItems?.Count == 0)
+				return;
+
+			if (e is not null && e.AddedItems.Count == 0 && e.RemovedItems.Count == 0)
```
workbysaran: The SequenceEqual check already covers this since it compares the final selection state rather than how it changed. If AddedItems and RemovedItems are both empty, SequenceEqual will return true anyway. Can you clarify why this separate check is needed?

src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
muko: The intent is to return early before allocating a new list instance via `ToList()`.

src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
muko: I’ll think about moving it to the base class.

yair100 (approved):
LGTM

PR #17045, opened 2025-04-15
Feature: Automatically detect zip encoding

oxygen-dioxide:
After this PR, when decompressing zip file and the encoding of zip file is undetermined, Files will detect the encoding autometically. 

image

Before this PR, SharpZipLib decompressing (decompressing with specified encoding) blocks the UI. After this PR, it no longers blocks the UI and works like SevenZipSharp.

All possible Windows system encodings are added into the encoding list.

**Resolved / Related Issues**

To prevent extra work, all changes to the Files codebase must link to an approved issue marked as `Ready to build`. Please insert the issue number following the hashtag with the issue number that this Pull Request resolves.
- This PR is a part of #14937

**Steps used to test these changes**

Stability is a top priority for Files and all changes are required to go through testing before being merged into the repo. Please include a list of steps that you used to test this PR.

1. Download this zip file: 
[japanese-example.zip](https://github.com/user-attachments/files/19619832/japanese-example.zip)
2. In Files, right click this file, Extract -> Extract Files
3. Select "Japanese (Shift-JIS)" encoding
image

PR #18529, opened 2026-05-30
Feature: Add ZIP encoding selection support for browsing archives with non-UTF-8 filenames

### LINKED ISSUE #14937: Bug: File names for Japanese zips aren't shown correctly on Chinese language

oxygen-dioxide:
**Problem**
When opening a ZIP archive whose filenames use an encoding other than the user's OS encoding, the filenames appear garbled

**Solution**
Added a ComboBox in the navigation toolbar that appears when browsing a ZIP archive with undetermined encoding, allowing the user to select the appropriate character encoding. When a custom encoding is selected, the app switches from SevenZipSharp to SharpZipLib to read filenames and file contents.

**Resolved / Related Issues**

- Closes #14937

1. Download this zip file: [japanese-example.zip](https://github.com/user-attachments/files/19619832/japanese-example.zip)
2. In Files, double click this file to open it
3. Select "Japanese (Shift-JIS)" encoding

previous behaviour:
image

current behaviour:
image


oxygen-dioxide:
Problem: users still can't double-click a file in the archive to open it because process.Start() doesn't support custom zip encoding. (However, this bug also occurs for 7z files, so I may create another PR to fix it.)

src/Files.App/UserControls/NavigationToolbar.xaml
```diff
@@ -395,6 +395,17 @@
 			Orientation="Horizontal"
 			Spacing="4">
 
+			<!--  ZIP Encoding Selector  -->
+			<ComboBox
```
yair100: Please send a screenshot of this UI. I'm not sure the toolbar is the correct placement for a ComboBox.

oxygen-dioxide: image

yair100: What are the other options in the dropdown?

oxygen-dioxide: It's the same as in my PR #17045, which is all possible ANSI encoding in Windows.
image

https://github.com/oxygen-dioxide/Files/blob/48e97f20266720dbcc15ec07fd270f1e4bacef64/src/Files.App/Data/Items/EncodingItem.cs#L45-L76
 
 ```cs
 public static EncodingItem[] Defaults = new string?[] {
			null,//System Default
            "UTF-8",
			
            //All possible Windows system encodings
            //reference: https://en.wikipedia.org/wiki/Windows_code_page
            //East Asian
            "shift_jis",       //Japanese
            "gb2312",          //Simplified Chinese
            "big5",            //Traditional Chinese
            "ks_c_5601-1987",  //Korean
            
            //Southeast Asian
            "Windows-1258",    //Vietnamese
            "Windows-874",     //Thai
            
            //Middle East
            "Windows-1256",    //Arabic
            "Windows-1255",    //Hebrew
            "Windows-1254",    //Turkish
            
            //European
            "Windows-1252",    //Western European
            "Windows-1250",    //Central European
            "Windows-1251",    //Cyrillic
            "Windows-1253",    //Greek
            "Windows-1257",    //Baltic
            
            "macintosh",
		}
			.Select(x => new EncodingItem(x))
			.ToArray();
```

Comment (yair100) [reply src/Files.App/UserControls/NavigationToolbar.xaml]: The bottom right corner of the status bar might be a better location. Fyi @mdtauk

**Comment by oxygen-dioxide** [comment]
moved the combobox to bottom right corner
image

**Comment by yair100** [comment]
Thanks! I'd like to take a closer look and get some feedback from @mdtauk but I can already tell that the overall location looks a lot better in that screenshot.

**Comment by yair100** [comment]
What do you think about replacing the ComboBox with a Button and Flyout like we do for git branches? Please let me know if you want me to push a commit.
image

**Comment by oxygen-dioxide** [comment]
> What do you think about replacing the ComboBox with a Button and Flyout like we do for git branches? Please let me know if you want me to push a commit. image

Looks good to me

File: src/Files.App/ViewModels/UserControls/NavigationToolbarViewModel.cs
Comment (yair100) [inline src/Files.App/ViewModels/UserControls/NavigationToolbarViewModel.cs:1]: These changes can be reverted.

**Comment by yair100** [comment]
What's the expected behavior when opening a regular zip file?

**Comment by oxygen-dioxide** [comment]
- **`src/Files.App/Data/Items/EncodingItem.cs`** — Changed the built-in Chinese encoding from `gb2312` to `gb18030`, which is a strict superset and matches the encoding name returned by `DetectEncodingAsync`.
- **`src/Files.App/Services/Storage/StorageArchiveService.cs`** — In `IsEncodingUndeterminedAsync`, entries whose names consist entirely of ASCII characters are now considered **determined** (not undetermined), even if they don't declare Unicode. This avoids showing the encoding picker when the zip content is pure ASCII and doesn't need a specific non-UTF-8 encoding.
- **`src/Files.App/Data/Models/DirectoryPropertiesViewModel.cs`** (the StatusBarViewModel) — When auto-detect finds an encoding not present in the built-in list, it dynamically creates an `EncodingItem` for it and appends it to the options, ensuring the detected encoding always appears in the dropdown.

**Comment by oxygen-dioxide** [comment]
> What's the expected behavior when opening a regular zip file?

For ascii or utf-8 zip file, the encoding selector shouldn't show up. I've fixed my code.

**Review by yair100 — DISMISSED** [review]

File: src/Files.App/Utils/Storage/StorageItems/ZipStorageFolder.cs [hunk 2]
```diff
@@ -21,6 +23,12 @@ public sealed partial class ZipStorageFolder : BaseStorageFolder, ICreateFileWit
 		private readonly string containerPath;
 		private BaseStorageFile backingFile;
 
+		/// <summary>
+		/// Gets or sets the encoding to use when browsing ZIP files.
+		/// When set, SharpZipLib is used instead of SevenZipSharp.
+		/// </summary>
+		internal static Encoding? CurrentEncoding { get; set; }
```
0x5bfa: Fine this to be static? The same encoding would be shared across the process.

src/Files.App/Utils/Storage/StorageItems/ZipStorageFile.cs
0x5bfa, line 1: Do these newly added methods support password requesting? Apparently, it silently fails?

src/Files.App/Utils/Storage/StorageItems/ZipStorageFolder.cs
0x5bfa, line 1: As above.

src/Files.App/Utils/Storage/StorageItems/ZipStorageFolder.cs
```diff
@@ -255,6 +322,54 @@ public override IAsyncOperation<IReadOnlyList<IStorageItem>> GetItemsAsync()
...
+						var split = winPath.Substring(Path.Length).Split('\\', StringSplitOptions.RemoveEmptyEntries);
+						if (split.Length <= 0)
+							continue;
+
+						if (entry.IsDirectory || split.Length > 1) // Not all folders have a ZipEntry
+						{
+							var itemPath = System.IO.Path.Combine(Path, split[0]);
+							if (!items.Any(x => x.Path == itemPath))
+							{
+								var folder = new ZipStorageFolder(itemPath, containerPath, backingFile);
```
0x5bfa: Is GetBasicProperties encoding aware as well? If not, it may show incorrect date/size info.

src/Files.App/Utils/Storage/StorageItems/ZipStorageFolder.cs
```diff
@@ -255,6 +322,54 @@ public override IAsyncOperation<IReadOnlyList<IStorageItem>> GetItemsAsync()
...
+								((IPasswordProtectedItem)folder).CopyFrom(this);
+								items.Add(folder);
+							}
+						}
+						else
+						{
+							var file = new ZipStorageFile(winPath, containerPath, backingFile);
```
0x5bfa: As above.

0x5bfa (changes requested):

0x5bfa:
CI is now failing and most of my reviews are still present.

yair100:
@oxygen-dioxide fyi

src/Files.App/UserControls/StatusBar.xaml
```diff
@@ -193,10 +194,57 @@
...
+
+						<!--  Encodings List  -->
+						<ListView
+							x:Name="ZipEncodingList"
+							Padding="4"
+							Background="{ThemeResource CardBackgroundFillColorDefaultBrush}"
+							IsItemClickEnabled="True"
+							ItemClick="ZipEncodingList_ItemClick"
+							ItemsSource="{x:Bind StatusBarViewModel.ZipEncodingOptions, Mode=OneWay}"
+							SelectedItem="{x:Bind StatusBarViewModel.SelectedZipEncoding, Mode=OneWay}"
```
yair100: Shouldn't this be TwoWay?

oxygen-dioxide: When a ListView's `ItemsSource` is replaced (which happens every time `StatusBarViewModel` is reassigned during a tab switch), the ListView internally resets `SelectedItem = null`. With a **TwoWay** binding, this null propagates back to the ViewModel property, overwriting the previously selected encoding:

```
UpdateStatusBarProperties()
  → StatusBarViewModel DP set → DataContext changes
    → ItemsSource binding evaluates → ListView resets SelectedItem = null
    → SelectedItem TwoWay binding writes null → SelectedZipEncoding = null  ← bug
```

This happens despite `UpdateZipEncodingStateAsync()` already having restored the correct encoding synchronously earlier in the same UI cycle.

Changing to `Mode=OneWay` breaks the ListView → ViewModel propagation path, preventing the spurious null write. The user's selection is still synced to the ViewModel via the existing `ItemClick` handler (called when the user picks an encoding from the flyout):

https://github.com/oxygen-dioxide/Files/blob/3d0b867e34da179111a34691291fe28cf962d0d1/src/Files.App/UserControls/StatusBar.xaml.cs#L71-L77

`ItemClick` → `SelectedZipEncoding.set` → triggers `OnZipEncodingChangedAsync()` → stores the encoding and refreshes the view, just as before.

The ViewModel → ListView direction (OneWay) still works, so on tab switch or encoding restore, the ListView correctly highlights the matching item from the ViewModel's `SelectedZipEncoding` value.

There is still a problem: when a zip file is opened, the detected zip encoding is shown on the button, but isn't shown as selected in the flyout menu. Let me fix it.

oxygen-dioxide: fixed

src/Files.App/Utils/Storage/StorageItems/ZipStorageFolder.cs
oxygen-dioxide: It is encoding aware now

src/Files.App/Utils/Storage/StorageItems/ZipStorageFolder.cs
oxygen-dioxide: It is encoding aware now

src/Files.App/UserControls/StatusBar.xaml
0x5bfa: We should find the root cause and fix this to use the TwoWay binding.

oxygen-dioxide:
Now it supports encrypted archives

src/Files.App/UserControls/StatusBar.xaml
oxygen-dioxide: Now I use TwoWay. In SelectedZipEncoding, now it does nothing if the value is null (because normally it would never be null. The default encoding is a `EncodingItem(null)`, not a `null`.

src/Files.App/Utils/Storage/StorageItems/ZipStorageFolder.cs
oxygen-dioxide: Currently it works, but repeatly asks for password. This is another bug because it also happens on the original SevenZipSharp extractor. See #18656

oxygen-dioxide:
I've fixed all the problems you listed here. I think we can continue now.

0x5bfa:
@codex review

oxygen-dioxide:
@codex review

oxygen-dioxide:
@codex review

yair100 (dismissed):
LGTM

yair100:
While everything works fine on my end, Claude requested a couple of changes. It might be worth addressing them before we merge.
- [ ] `StatusBarViewModel` is created per layout page but the new `ContentPageContext.PropertyChanged += ...` subscription is never removed, so every navigation leaks a view model instance. Needs an unsubscribe path.
- [ ] `IPasswordProtectedItem.cs`: `ZipException.Message.Contains("encrypted")` matches "No password available for encrypted stream" but not "Invalid password", so a mistyped password won't re-prompt. Also match "password" (or check `entry.IsCrypted`).
- [ ] `ZipStorageFolder.GetBasicPropertiesWithEncodingAsync` / `GetItemWithEncodingAsync` compare entry paths with case-sensitive `==` while everything else uses `OrdinalIgnoreCase` — case-different names miss.
- [ ] `CopyWithEncodingAsync` buffers the whole entry into a `MemoryStream` before writing (the SevenZip path streams directly) — extracting a large file will spike memory. Copy `zipFile.GetInputStream(entry)` straight to the destination stream.
- [ ] `GetItemsWithEncodingAsync`, `GetItemWithEncodingAsync`, and `OpenSequentialReadWithEncodingAsync` are missing the `SafetyExtensions.Wrap(..., RetryWithCredentialsAsync)` wrapper all sibling methods use — no password prompt on failure.

Josh65-2201 (dismissed):
Only issue is that the folder doesn't reload correctly when changing the encoding, I have to back out and go back in. It's selecting the correct type and loading correctly when opened so it minor.

oxygen-dioxide:
> Only issue is that the folder doesn't reload correctly when changing the encoding, I have to back out and go back in. It's selecting the correct type and loading correctly when opened so it minor.

fixed

yair100:
Please resolve the merge conflicts.

oxygen-dioxide:
> Please resolve the merge conflicts.

resolved

yair100:
It looks like we need some additional logic to make this responsive. Is it okay if I push a commit?
image

oxygen-dioxide:
> It looks like we need some additional logic to make this responsive. Is it okay if I push a commit? image

OK

mdtauk:
Ideally this functionality would not be in the status bar, but as part of an **Archive** header, like with have the filter header - I have discussed this with @yair100 and for now, we have a work around until more time is available to create the header.

yair100:
Thank you @mdtauk for helping come up with a quick workaround for now.

PR #18531, opened 2026-05-30
Code Quality: Update StorageBar.xaml - Replace Storyboard with VisualState Setters

mdtauk:
https://github.com/user-attachments/assets/08d51890-0eb7-4e5e-abf8-8796e48db0b9

Based on what we are seeing with the WinUI Perf2026 changes, I thought I would start by updating our StorageBar to use VisualStateSetters, instead of Storyboards.

It does not resolve or close any existing issues

I ran the UITests app and the control behaves the same as it did before.

If possible, @0x5bfa could make some kind of test to ensure there is nothing that has been overlooked.  It should just be 1:1 behaviour though, and I suppose if Microsoft felt moving from storyboards offered a performance improvement, there may be some gains from this change.

### DIFF
--- a/src/Files.App.Controls/Storage/StorageBar/StorageBar.xaml
+++ b/src/Files.App.Controls/Storage/StorageBar/StorageBar.xaml
@@ -137,36 +137,24 @@
 								<VisualState x:Name="Safe" />
 
 								<VisualState x:Name="Caution">
-									<Storyboard>
-										<ObjectAnimationUsingKeyFrames Storyboard.TargetName="PART_ValueBar" Storyboard.TargetProperty="Background">
-											<DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource StorageBarValueCautionBrush}" />
-										</ObjectAnimationUsingKeyFrames>
-										<ObjectAnimationUsingKeyFrames Storyboard.TargetName="PART_TrackBar" Storyboard.TargetProperty="Background">
-											<DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource StorageBarTrackCautionBrush}" />
-										</ObjectAnimationUsingKeyFrames>
-									</Storyboard>
+									<VisualState.Setters>
+										<Setter Target="PART_ValueBar.Background" Value="{ThemeResource StorageBarValueCautionBrush}" />
+										<Setter Target="PART_TrackBar.Background" Value="{ThemeResource StorageBarTrackCautionBrush}" />
+									</VisualState.Setters>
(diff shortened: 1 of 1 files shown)


0x5bfa (approved):
LGTM!

yair100:
Does this have a noticeable impact on performance? Have you run any benchmarks? Fyi @marcelwgn

mdtauk:
> Does this have a noticeable impact on performance? Have you run any benchmarks? Fyi @marcelwgn

I don't think we have enough testing in the app to benchmark it - but if Microsoft has decided it offers a performance improvement - and has no functional difference making the change - I see no reason not to make the change.

There are other uses of Storyboard within the app, such as the Sidebar, etc - but I only felt comfortable touching the code I wrote, and the other uses would be others.

yair100:
> I don't think we have enough testing in the app to benchmark it - but if Microsoft has decided it offers a performance improvement - and has no functional difference making the change - I see no reason not to make the change.

VisualState Setters are easier to work with in general, but I don't see a reason to change something that's already working. That is unless we know that it makes a difference.

mdtauk:
> > I don't think we have enough testing in the app to benchmark it - but if Microsoft has decided it offers a performance improvement - and has no functional difference making the change - I see no reason not to make the change.
> 
> VisualState Setters are easier to work with in general, but I don't see a reason to change something that's already working. That is unless we know that it makes a difference.

Microsoft are making the changes to all their default WinUI control templates. and list the new xaml resource dictionaries as Perf2026, and its part of their effort to improve WinUI performance.

The changes reduce lines of Xaml, and involve no behavioural changes.  If it does boost performance as Microsoft seems to think, and reduces lines of code - it should be a no brainer to make these minimal changes.

marcelwgn:
I did some naive benchmarking, there is a slight performance gain. Question is though, if that matters for us. Its not like we are going to show dozens or hundreds of storagebars is it?

yair100:
> I did some naive benchmarking, there is a slight performance gain. Question is though, if that matters for us. Its not like we are going to show dozens or hundreds of storagebars is it?

Thank you for checking. I don't expect this to have much impact in Files, but it's good to confirm that this does have a performance gain. It's also something we can keep in mind for future development.

mdtauk:
> I did some naive benchmarking, there is a slight performance gain. Question is though, if that matters for us. Its not like we are going to show dozens or hundreds of storagebars is it?

Dozens is possible, Hundreds is less so.  I think there is a plan to maybe include them in a column within the details view, so for however many drives you display there, they could each have a storage bar.

PR #18532, opened 2026-05-30
Code Quality: Update StorageRing.xaml - Replace Storyboard with VisualState Setters

mdtauk:
https://github.com/user-attachments/assets/9471c2b9-d199-4639-b013-e77519c8e5d0

Based on what we are seeing with the WinUI Perf2026 changes, I thought I would start by updating our StorageRing to use VisualStateSetters, instead of Storyboards.

This does not close or resolve any existing issues

I ran the UITests app and the control behaves the same as it did before.

If possible, @0x5bfa  could make some kind of test to ensure there is nothing that has been overlooked. It should just be 1:1 behaviour though, and I suppose if Microsoft felt moving from storyboards offered a performance improvement, there may be some gains from this change.

### DIFF
--- a/src/Files.App.Controls/Storage/StorageRing/StorageRing.xaml
+++ b/src/Files.App.Controls/Storage/StorageRing/StorageRing.xaml
@@ -157,36 +157,24 @@
 								<VisualState x:Name="Safe" />
 
 								<VisualState x:Name="Caution">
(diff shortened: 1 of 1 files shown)


0x5bfa (approved):
LGTM!

PR #18540, opened 2026-06-03
Feature: Add Skip and Skip All buttons to File In Use dialog

### LINKED ISSUE #8239: Feature: Add skip button when deleting multiple files that are in use by another application
### What's the Problem?

_No response_

### Solution/Idea

When deleting a lot of items from for example Temp folder, a lot of items are usually in use, would be convenient if there was a way to say "ok" to every single file in there, so there's no need to say "ok" every time

Also would be nice if when deleting items, there was a way to see what process is using it

image

### Alternatives

na

### Priorities

| Capability | Priority |
- Replace "okay" with a skip button
- Add skip all if there are multiple files in use

### Files Version

_No response_

### Windows Version

_No response_

### Comments

_No response_

codenano077:
**Resolved / Related Issues**

- Closes #8239

1. Opened Files Dev app
2. Locked a file using PowerShell: $file = [System.IO.File]::Open("D:\testlock.txt", "OpenOrCreate", "ReadWrite", "None")
3. Navigated to the locked file in Files Dev
4. Attempted to delete the locked file
5. Confirmed that Retry, Skip, and Skip All buttons appear correctly in the File In Use dialog
6. Clicked Skip to verify it skips the current file
7. Clicked Skip All to verify it skips all locked files

### DIFF
--- a/src/Files.App/Helpers/Dialog/DynamicDialogFactory.cs
+++ b/src/Files.App/Helpers/Dialog/DynamicDialogFactory.cs
@@ -146,8 +146,9 @@ namespace Files.App.Helpers
 				TitleText = Strings.FileInUseDialog_Title.GetLocalizedResource(),
 				SubtitleText = lockingProcess.IsEmpty() ? Strings.FileInUseDialog_Text.GetLocalizedResource() :
 					string.Format(Strings.FileInUseByDialog_Text.GetLocalizedResource(), string.Join(", ", lockingProcess.Select(x => $"{x.AppName ?? x.Name} (PID: {x.Pid})"))),
-				PrimaryButtonText = "OK",
-				DynamicButtons = DynamicDialogButtons.Primary
+				PrimaryButtonText = Strings.Retry.GetLocalizedResource(),
+				SecondaryButtonText = Strings.Skip.GetLocalizedResource(),
+				DynamicButtons = DynamicDialogButtons.Primary | DynamicDialogButtons.Secondary | DynamicDialogButtons.Cancel
 			});
 			return dialog;
 		}
--- a/src/Files.App/Dialogs/FilesystemOperationDialog.xaml
+++ b/src/Files.App/Dialogs/FilesystemOperationDialog.xaml
@@ -12,6 +12,7 @@
 	xmlns:vm="using:Files.App.ViewModels.Dialogs.FileSystemDialog"
 	x:Name="RootDialog"
 	Title="{x:Bind ViewModel.Title, Mode=OneWay}"
+	CloseButtonText="{x:Bind ViewModel.CloseButtonText, Mode=OneWay}"
 	Closing="RootDialog_Closing"
 	CornerRadius="{StaticResource OverlayCornerRadius}"
 	DefaultButton="Primary"
--- a/src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
+++ b/src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
@@ -857,10 +857,17 @@ namespace Files.App.Utils.Storage
 				? Strings.FileInUseDialog_Text.GetLocalizedResource()
 				: string.Format(Strings.FileInUseByDialog_Text.GetLocalizedResource(), string.Join(", ", lockingProcess.Select(x => $"{x.AppName ?? x.Name} (PID: {x.Pid})")));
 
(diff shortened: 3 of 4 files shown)


yair100:
Can you please share screenshot? Also, can you confirm if the skip option only shows when deleting multiple items?

codenano077:
Here's a screenshot of the dialog with the new buttons:
Screenshot 2026-06-04 152323
Screenshot 2026-06-04 152337


Regarding your question — the Skip and Skip All buttons currently show for both single and multiple file deletions. However, I think it makes more sense to only show Skip and Skip All when deleting multiple files, since skipping a single locked file is essentially the same as cancelling. Would you like me to update the behavior so Skip only appears when multiple files are being deleted?

yair100:
> Would you like me to update the behavior so Skip only appears when multiple files are being deleted?

Yes, that would be great.

yair100 (dismissed):

Josh65-2201 (commented):
I'm not certain why there is `Skip` and `Skip all` since all files are attempted at the same time so both buttons do the same thing.

Permanently deletes file in use prompt also only has `Try again` or `Cancel` buttons like 4.1.3, shouldn't it have the same?

codenano077:
Thanks for the feedback!

Regarding Skip vs Skip All — you raise a valid point. Could you clarify the intended behavior? Should Skip dismiss the dialog for the current file only and show it again for the next locked file, while Skip All dismisses it for all locked files at once?

Regarding the permanent delete dialog — I can update that to be consistent with this change if that's the desired behavior.

Josh65-2201:
I think this was fixed while ago but the issue wasn't closed. All in use files are grouped together to show them and what's using it. Unless Yair wants to ungroup them so one file is shown at a time?

yair100:
> I think this was fixed while ago but the issue wasn't closed. All in use files are grouped together to show them and what's using it. Unless Yair wants to ungroup them so one file is shown at a time?

@Josh65-2201 in the latest stable build, what happens with the non-conflicting files, do they finish transferring? I thought that we needed to add a skip button to continue with the non-conflicting items.

Josh65-2201:
None blocking ones are completed for me

codenano077:
I tested this as well — non-locked files complete automatically while the dialog shows for locked ones.
Given that all locked files are shown together in one dialog, I agree that Skip All doesn't add much value. I'd suggest simplifying to just a single Skip button that dismisses the dialog and skips all the locked files shown. This keeps it simple and consistent.
Should I update the PR to remove Skip All and keep only Skip?

Josh65-2201:
Sure, all that's need is changing the Cancel button text to Skip

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -165,6 +165,9 @@
   <data name="Skip" xml:space="preserve">
     <value>Skip</value>
   </data>
+    <data name="SkipAll" xml:space="preserve">
+    <value>Skip all</value>
+</data>
```
Josh65-2201: Seem to be unused, Single file prompt show cancel and multi file prompt show as skip

src/Files.App/Helpers/Dialog/DynamicDialogFactory.cs
```diff
@@ -146,8 +146,10 @@ public static DynamicDialog GetFor_FileInUseDialog(List<Win32Process> lockingPro
 				TitleText = Strings.FileInUseDialog_Title.GetLocalizedResource(),
 				SubtitleText = lockingProcess.IsEmpty() ? Strings.FileInUseDialog_Text.GetLocalizedResource() :
 					string.Format(Strings.FileInUseByDialog_Text.GetLocalizedResource(), string.Join(", ", lockingProcess.Select(x => $"{x.AppName ?? x.Name} (PID: {x.Pid})"))),
-				PrimaryButtonText = "OK",
-				DynamicButtons = DynamicDialogButtons.Primary
+				PrimaryButtonText = Strings.Retry.GetLocalizedResource(),
+				SecondaryButtonText = Strings.Skip.GetLocalizedResource(),
+				CloseButtonText = Strings.SkipAll.GetLocalizedResource(),
```
Josh65-2201: Line 151 seem to be unused, Single file prompt show cancel and multi file prompt show as skip

Josh65-2201 (commented):
Works as expected but looking at the changes there seems to be a few unneeded ones.

Josh65-2201:
Looks like there is a lot of file formatting changes, I can retest tomorrow

Josh65-2201 (approved):

AzaelMew:
Oh my god. After 3 long years, its finally done. Shame im no longer using windows.

PR #18551, opened 2026-06-05
Feature: Optional indexed search backend (C# search service)

### LINKED ISSUE #5845: Feature request: Everything search integration
The current solution for searching for files does work, but I believe we can add value if we use Everything Search when we detect it's installed. Everything Search can index and search for files much faster than the current method and will help users be more productive.

Wingingbump:
**Resolved / Related Issues**

There is no `Ready to build` issue number for this yet — this PR is opened at a maintainer's request in Discord to start the design discussion around a faster, optional search backend. Happy to file/link a tracking issue if preferred.

Closes #5845

**Summary**

Adds an **opt-in indexed search backend** alongside the existing `Windows.Storage.Search` path. The current provider remains the default; nothing changes unless the user explicitly enables indexed search.

- A standalone C# Windows Service (`files-search-service.exe`) maintains an in-memory inverted + trigram filename index over the user's home directory, kept live by a `ReadDirectoryChangesW` watcher with process throttling so it stays out of the way.
- Files.App talks to it over gRPC through a new `ISearchProvider` abstraction (`Files.SearchAbstraction`), with two implementations: `Files.IndexedSearch.Client` (new) and `Files.LegacySearch` (wraps today's behavior).
- A `SearchRouter` selects the provider, caches service health, and falls back to legacy if the service is unavailable.
- Opt-in via **Settings → Advanced → "Use indexed search"** (or `FILES_SEARCH_PROVIDER=Indexed`).
- In Debug, a `Package.Debug.appxmanifest` omits the SCM service declaration so VS can sideload without admin; the service is spawned as a child process instead.

Benchmark on a ~1M-file corpus: first-result latency ~11 ms indexed vs ~2025 ms legacy. Pinned baseline checked in at `bench-results/baseline.json`; reproduce with `run-bench.ps1`. Design notes in `docs/csharp-search-service.md`.

1. `dotnet run -c Release --project tests/Files.Search.Correctness/Files.Search.Correctness.csproj` → 92/92 pass (tokenizer, scorer, index, persistence, corpus correctness).
2. Built and launched Files.App; enabled **Settings → Advanced → Use indexed search**.
3. Searched in a large directory and confirmed results stream in with the indexed provider; verified ranking/order against the legacy provider for the same queries.
4. Stopped the search service and confirmed `SearchRouter` falls back to legacy search without errors.
5. Toggled the setting off and confirmed behavior returns to the stock `Windows.Storage.Search` path.


src/Files.SearchService/proto/files_search.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
```
d2dyno1, line 1: Does using Google's Protocol Buffers in gRPC have any benefit over JSON? IIRC, protobuf is not a requirement for gRPC, and I'd insist on not adding more dependencies.

Need a decision on this @yair100

Josh65-2201:
How do you get the index service installed? Enabling `Files settings > Advanced > Use index search engine` doesn't prompt for admin to install one or create a process in Task manager.

Also fails to do a search
image

src/Files.SearchService/Usn/NativeMethods.cs
```diff
@@ -0,0 +1,170 @@
...
+using System.Runtime.InteropServices;
+
+namespace Files.SearchService.Usn;
+
+/// <summary>
+/// P/Invoke declarations for NTFS USN Change Journal access.
+/// All structures match the Windows SDK definitions for USN_RECORD_V2
+/// and MFT_ENUM_DATA_V0 used by FSCTL_ENUM_USN_DATA.
+/// </summary>
+internal static partial class NativeMethods
```
d2dyno1, line 14: Perhaps `Files.App.CsWin32` should be used instead?

tests/corpora/Program.cs
```diff
@@ -0,0 +1,348 @@
...
+        }
+
+        if (outDir is null) { Console.Error.WriteLine("error: --out is required"); return null; }
+
+        // Presets — small targets a quick local run; medium/large need real disk.
+        (string name, int count, long avg) = preset switch
+        {
+            "small"  => ("small",  50_000,    40L * 1024),       // ~2 GiB
+            "medium" => ("medium", 500_000,   100L * 1024),      // ~50 GiB
+            "large"  => ("large",  2_000_000, 250L * 1024),      // ~500 GiB
```
d2dyno1, line 88: 500gb is an overkill for testing

tests/corpora/Program.cs
```diff
@@ -0,0 +1,348 @@
...
+        }
+        return needles;
+    }
+}
+
+// Deterministic RNG — xorshift64*. Single-threaded; no hidden state.
+internal sealed class Xorshift64
+{
+    private ulong _s;
+    public Xorshift64(ulong seed) { _s = seed == 0 ? 0xDEADBEEFCAFEBABEUL : seed; }
```
d2dyno1, line 304: Is this seed fallback intentional?

tests/Files.Search.Correctness/PersistenceTests.cs
```diff
@@ -0,0 +1,128 @@
...
+            Assert.AreEqual(records[i].ModifiedUtc, loaded[i].ModifiedUtc);
+        }
+    }
+
+    [TestMethod]
+    public async Task RoundTrip_Unicode_PathAndFilename()
+    {
+        var records = new List<DocRecord>
+        {
+            new(@"C:\root\测试\测试_file.txt", "测试_file.txt", 512UL, DateTime.UtcNow),
```
d2dyno1, line 55: Verify: Those characters *may* cause some issues with different environments and IDEs. Perhaps it's better to use a constant byte array and convert it to a string at runtime?

hez2010:
Instead of pulling the dependencies on gRPC, I think for such purpose we should build on top of WinRT so that the search service becomes an WinRT out-off-proc server. See the project `Files.App.Server`.

src/Files.SearchService/proto/files_search.proto
d2dyno1: https://github.com/files-community/Files/pull/18551#issuecomment-4643117801

Wingingbump:
@Josh65-2201 
The service isn't starting because its binary doesn't get deployed on a clean build — Files.App
  copies `files-search-service.exe` via a content glob that's evaluated *before* the `Files.SearchService` project
  builds, so the first build leaves the `SearchService\` folder empty.
  
  1. Build the **Files.SearchService** project first (or just rebuild the solution once more).
  2. Confirm the exe landed here:
  `...\Files.App\bin\<platform>\Debug\net10.0-windows10.0.26100.0\AppX\SearchService\files-search-service.exe`
  3. Relaunch Files — it spawns the service itself as a child process over local TCP (no admin needed), and search
  should work.
  
  will need to fix

bench-results/baseline.json
Josh65-2201, line 1: If this is just a benchmark results file, it should be deleted.

src/Files.App/Package.Debug.appxmanifest
Josh65-2201, line 1: Should be removed, https://github.com/files-community/Files/blob/main/src/Files.App/Package.appxmanifest is what used

Josh65-2201:
@Wingingbump Are you able to continue working on this?

yair100:
@Wingingbump following up.

Wingingbump:
@yair100 yeah I can keep working on it. will need to get caught up on the plan moving forward though.

yair100:
@Wingingbump sounds good! I think there are also some comments on the Discord thread.

Wingingbump:
What exactly are our next steps? I saw some of the messages on discord but want to clarify them.

yair100:
Let's continue on Discord, @d2dyno1 shared some feedback that needs to be addressed.

PR #18554, opened 2026-06-06
Fix: Fixed issue with opening files from jar archives

### LINKED ISSUE #18533: Bug: Cannot open files inside a .jar archive by double-clicking like in .zip archive
### Description

Files allows users to browse the contents of a .jar file just like a .zip archive. However, double-clicking any file inside the archive does nothing.

### Steps To Reproduce

1. Set a default application to open .class files (e.g., Notepad).
2. Download [yamlbeans-1.17.jar](https://github.com/EsotericSoftware/yamlbeans/releases/download/1.17/yamlbeans-1.17.jar).
3. Open yamlbeans-1.17.jar using Files, navigate to the com\esotericsoftware\yamlbeans directory, and double-click any .class file.

### Files Version

4.1.1.0

### Windows Version

Windows 11 26200.8246

### User ID

_No response_

### Log File

[debug.log](https://github.com/user-attachments/files/28431487/debug.log)

marko1olo:
**Resolved / Related Issues**

- Closes #18533

**Summary**

- Launch files inside browsable archives through their streamed `StorageFile` representation instead of passing the virtual archive path to the Win32 shell path.
- Initialize `LauncherOptions` with the Files window handle for the archive-entry launch path, matching the nearby user-initiated launcher flows.
- Keep the existing Win32 fallback if the WinRT file launch fails.
- Leave normal filesystem files and the existing image-specific neighboring-files flow unchanged.

**Why**

Files inside archives such as `.jar` have virtual paths like `archive.jar\path\File.class`. Passing that path to the Win32 shell does not point at a real filesystem item, so double-clicking non-image files inside `.jar` archives can do nothing. `ZipStorageFile.ToStorageFileAsync()` already materializes archive entries as streamed `StorageFile` objects, which is the launch path used for archive images.

1. Ran `git diff --check`.
2. Restored the missing server assets with `dotnet restore src\Files.App.Server\Files.App.Server.csproj -p:Platform=x64 --nologo`.
3. Ran `dotnet build src\Files.App\Files.App.csproj -c Debug -p:Platform=x64 --no-restore --nologo`; the build reached the app XAML compilation stage, then failed with `WMC9999` because the local XAML compiler package could not load `Microsoft.UI.Xaml.Markup.Compiler.ErrorMessages.resources`.
4. Compared the new archive-entry launch path with the existing image and app-picker launch paths in `NavigationHelpers.OpenFile`.

### DIFF
--- a/src/Files.App/Helpers/Navigation/NavigationHelpers.cs
+++ b/src/Files.App/Helpers/Navigation/NavigationHelpers.cs
@@ -759,6 +759,24 @@ namespace Files.App.Helpers
 									if (!launchSuccess)
 										await Win32Helper.InvokeWin32ComponentAsync(path, associatedInstance, args);
 								}
+								else if (childFile.Item is ZipStorageFile zipStorageFile)
+								{
+									var options = InitializeWithWindow(new LauncherOptions());
+									var storageItem = (StorageFile)await FilesystemTasks.Wrap(() => zipStorageFile.ToStorageFileAsync().AsTask());
+									if (storageItem is null)
+									{
+										await Win32Helper.InvokeWin32ComponentAsync(path, associatedInstance, args);
+									}
+									else if (!await Launcher.LaunchFileAsync(storageItem, options))
+									{
+										var pickerOptions = InitializeWithWindow(new LauncherOptions
+										{
+											DisplayApplicationPicker = true
+										});
+										if (!await Launcher.LaunchFileAsync(storageItem, pickerOptions))
+											await Win32Helper.InvokeWin32ComponentAsync(path, associatedInstance, args);
+									}
+								}
 								else
 								{
 									await Win32Helper.InvokeWin32ComponentAsync(path, associatedInstance, args);


Josh65-2201:
You should test your changes in Visual Studio https://files.community/docs/contributing/building-from-source what you put in the steps to test is not testing anything related to the change you have made

Josh65-2201:
This doesn't show the open with prompt for files with no program associated

marko1olo:
Good catch. I pushed `26df283` to cover that case as well.

The archive branch now keeps the direct `LaunchFileAsync(storageItem, options)` path for entries that already have an association, but if that returns `false`, it retries the same extracted `StorageFile` with `LauncherOptions.DisplayApplicationPicker = true` before falling back to the old Win32 launch path. That should make unassociated archive entries show the Windows app picker instead of trying to shell-launch the virtual `archive.jar\entry` path.

Suggested manual checks for this PR:
- Open a `.jar` in Files and double-click an entry whose extension has no default app. Expected: Windows "Open with" picker appears.
- Pick an app such as Notepad and verify the extracted archive entry opens.
- Re-check the original #18533 scenario with a `.class` entry that has a default app. Expected: it still opens directly without forcing the picker.

Local checks:
- `git diff --check` passed.
- `dotnet build src\Files.App\Files.App.csproj --no-restore -p:Configuration=Debug -p:Platform=x64` did not finish within 5 minutes in this checkout, so I stopped it and checked that no repo build processes were left running.

Josh65-2201:
Working for me

yair100:
Fyi @oxygen-dioxide

oxygen-dioxide:
It works

PR #18560, opened 2026-06-07
Feature: Added banner when Network Discovery is turned off in Windows

### LINKED ISSUE #9735: Prompt to enable network discovery when clicking the Network option
### What's the Problem?

Currently if it disabled it just shows a blank folder.

### Solution/Idea

There should be a prompt to enable network discovery on windows

### Alternatives

*

### Priorities

_No response_

### Files Version

_No response_

### Windows Version

_No response_

### Comments

_No response_

0x5bfa:
### Resolved / Related Issues

- Closes #9735

1. Turn off the Network Discovery from `Settings > Network & internet > Advanced network settings > Advanced sharing settings`
2. Go to `Network` from the sidebar

image

### DIFF
--- a/src/Files.App/Strings/en-US/Resources.resw
+++ b/src/Files.App/Strings/en-US/Resources.resw
@@ -4328,6 +4328,15 @@
   <data name="OpenSettings" xml:space="preserve">
     <value>Open settings</value>
   </data>
+  <data name="NetworkDiscoveryTurnedOffInfoBarMessage" xml:space="preserve">
+    <value>Network discovery is turned off. Network computers and devices are not visible.</value>
+  </data>
+  <data name="FileSharingTurnedOffInfoBarMessage" xml:space="preserve">
+    <value>File sharing is turned off. Some network computers and devices might not be visible.</value>
+  </data>
+  <data name="NetworkDiscoveryAndFileSharingTurnedOffInfoBarMessage" xml:space="preserve">
+    <value>Network discovery and file sharing are turned off. Network computers and devices are not visible.</value>
+  </data>
   <data name="VisualStudioCode" xml:space="preserve">
     <value>Visual Studio Code</value>
   </data>
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -172,6 +172,7 @@ namespace Files.App.ViewModels
 		private CancellationTokenSource searchCTS;
 		private CancellationTokenSource updateTagGroupCTS;
 		private CancellationTokenSource? filterDebounceCS;
+		private CancellationTokenSource? networkAvailabilityCTS;
 
 		public event EventHandler FocusFilterHeader;
 
@@ -308,6 +309,38 @@ namespace Files.App.ViewModels
 		{
 			get => emptyTextType;
 			set => SetProperty(ref emptyTextType, value);
+		}
+
+		private bool isNetworkDiscoveryInfoBarOpen;
+		public bool IsNetworkDiscoveryInfoBarOpen
+		{
+			get => isNetworkDiscoveryInfoBarOpen;
+			set => SetProperty(ref isNetworkDiscoveryInfoBarOpen, value);
+		}
+
+		private NetworkAvailability networkAvailability = NetworkAvailability.All;
+		public NetworkAvailability NetworkAvailability
+		{
+			get => networkAvailability;
+			private set
+			{
+				if (SetProperty(ref networkAvailability, value))
+					OnPropertyChanged(nameof(NetworkDiscoveryInfoBarMessage));
+			}
+		}
+
+		public string NetworkDiscoveryInfoBarMessage
+		{
+			get
+			{
+				return NetworkAvailability switch
+				{
+					NetworkAvailability.None => Strings.NetworkDiscoveryAndFileSharingTurnedOffInfoBarMessage.GetLocalizedResource(),
+					NetworkAvailability.Discovery => Strings.FileSharingTurnedOffInfoBarMessage.GetLocalizedResource(),
+					NetworkAvailability.Sharing => Strings.NetworkDiscoveryTurnedOffInfoBarMessage.GetLocalizedResource(),
+					_ => string.Empty,
+				};
+			}
 		}
 
 		public async Task UpdateFolderThumbnailImageSource()
@@ -735,6 +768,8 @@ namespace Files.App.ViewModels
 		{
 			Debug.WriteLine("CancelLoadAndClearFiles");
 			CloseWatcher();
+			CancelNetworkAvailabilityUpdate();
+			IsNetworkDiscoveryInfoBarOpen = false;
 			if (IsLoadingItems)
 			{
 				IsLoadingCancelled = true;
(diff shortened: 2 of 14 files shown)


0x5bfa:
We need better check for the network discovery being turned off. I have an idea.

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4304,6 +4304,12 @@
   <data name="OpenSettings" xml:space="preserve">
     <value>Open settings</value>
   </data>
+  <data name="NetworkDiscoveryInfoBarTitle" xml:space="preserve">
+    <value>Network discovery might be turned off</value>
```
yair100:
```suggestion
    <value>Network discovery is turned off</value>
```

src/Files.App/Strings/en-US/Resources.resw
```diff
@@ -4304,6 +4304,12 @@
...
+  </data>
+  <data name="NetworkDiscoveryInfoBarMessage" xml:space="preserve">
+    <value>Network devices may not appear when network discovery is turned off. Open Windows Settings to enable it.</value>
```
yair100:
```suggestion
    <value>Network devices will not appear when network discovery is turned off. Open Windows Settings to enable it.</value>
```

src/Files.App/Views/Layouts/GridLayoutPage.xaml
```diff
@@ -973,13 +973,9 @@
 	</local:BaseGroupableLayoutPage.Resources>
 
 	<Grid x:Name="RootGrid" ContextFlyout="{x:Bind BaseContextMenuFlyout}">
-		<Grid.RowDefinitions>
```
yair100, line 975: Why were the RowDefinitions removed?

src/Files.App/Views/MainPage.xaml
```diff
@@ -229,18 +230,34 @@
 						TabIndex="2"
 						Visibility="{x:Bind ViewModel.ShowToolbar, Mode=OneWay}" />
 
+					<InfoBar
```
yair100, line 234: Is there a chance we'll want to reuse this info bar for other purposes as well? If so, we'll want to make it more generic with support for customizing the content and action.

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -203,6 +203,11 @@ private void UpdateNavToolbarProperties()
 				InnerNavigationToolbar.ViewModel = SidebarAdaptiveViewModel.PaneHolder?.ActivePaneOrColumn.ToolbarViewModel;
 		}
 
+		private async void OpenNetworkAdvancedSettingsButton_Click(object sender, RoutedEventArgs e)
+		{
+			await Windows.System.Launcher.LaunchUriAsync(new Uri("ms-settings:network-advancedsettings"));
```
Josh65-2201: It would be better to have this directly enable it for the current network like Windows File Explorer does or go to `ms-settings:network-advancedsharing` as that where it can be enabled.

src/Files.App/Views/Layouts/GridLayoutPage.xaml
0x5bfa: THis wasn't used and thought its not worth creating a separate PR.
Should I revert?

src/Files.App/Views/MainPage.xaml
0x5bfa: I'm not sure. Since there's a Button specific to this scenario, I don't think we can reuse this InfoBar.

src/Files.App/Views/Layouts/GridLayoutPage.xaml
yair100: We might as well remove it, thanks for noticing!

src/Files.App/Views/MainPage.xaml
yair100: Sounds good for now, just something to keep in mind for the future.

src/Files.App/Views/MainPage.xaml.cs
yair100: This would be ideal, but we should also consider the risks of modifying a setting in Windows from Files.

Josh65-2201: Not sure what risk there would be if it's done the same way as Windows File Explorer.

0x5bfa: File Explorer is signed by Windows, which means auto-elevation is supported by default. However Files is not signed by Windows so it requires to prompt a UAC dialog. I found the implementation of this pop bar in Explorer and did the same, it didn't work for Files ([it did work on CLI though](https://github.com/0x5bfa/windows-shell-pocs/blob/main/DtshPoc.cs#L289-L321)). I'm not sure it's that worth it to look into deeper.

maihcx: I've discovered that if you create a Task Scheduler task with "Run with highest privileges" enabled, you can launch certain processes with administrative privileges simply by invoking that task.

image

mdtauk:
Just a slight aesthetic nit pick.  Could you please adjust the corner radius for the InfoBar so it matches the pane rounding, and the toolbar rounding.

src/Files.App/Views/MainPage.xaml.cs
0x5bfa: We have an exact process for that purpose in Files.App.Server but it's not currently working (works but doesn't work for a certain case; when the consumer is launched as admin). When it gets resolved, we would certainly should add an option to directly enable this.

But for the time being, we should just let users open Settings and enable it.

Josh65-2201: In that case, it can be change to go to `ms-settings:network-advancedsharing` as it's more direct way to enable it.

yair100 (dismissed):
LGTM

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -787,7 +789,14 @@ public string? SearchHeaderTitle
 
 		public void UpdateEmptyTextType()
 		{
-			EmptyTextType = FilesAndFolders.Count == 0 ? (IsSearchResults ? EmptyTextType.NoSearchResultsFound : EmptyTextType.FolderEmpty) : EmptyTextType.None;
+			var isFolderEmpty = FilesAndFolders.Count == 0;
+
+			IsNetworkDiscoveryInfoBarOpen =
```
yair100: Is there a more accurate way to determine the network discovery state than by checking the item count?

Josh65-2201:
As my Windows install of network page is broken I ran in it in a Windows 10 VM, 
19045 doesn't have support for `ms-settings:network-advancedsharing` and would need to open control panel using `control /name Microsoft.NetworkAndSharingCenter` to go to the correct area


Windows 11 works perfectly

yair100:
I wouldn't worry about 19045.

Josh65-2201:
Doesn't Files still support Windows 10? 19045 is the build number for the last version of it.

Josh65-2201 (dismissed):

0x5bfa:
> As my Windows install of network page is broken I ran in it in a Windows 10 VM,
> 19045 doesn't have support for `ms-settings:network-advancedsharing` and would need to open control panel using `control /name Microsoft.NetworkAndSharingCenter` to go to the correct area

There's an undocumented API to open Control Panel, which in this scenario File Explorer also uses. I've just introduced it and implemented the same detection logic as File Explorer does.

PTAL @yair100 @Josh65-2201

0x5bfa:
Three states:

image

image

image

src/Files.App/ViewModels/ShellViewModel.cs
0x5bfa: Implemented the same logic as the Explorer.

Josh65-2201 (dismissed):

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -787,7 +816,69 @@ public string? SearchHeaderTitle
 
 		public void UpdateEmptyTextType()
 		{
-			EmptyTextType = FilesAndFolders.Count == 0 ? (IsSearchResults ? EmptyTextType.NoSearchResultsFound : EmptyTextType.FolderEmpty) : EmptyTextType.None;
+			var isFolderEmpty = FilesAndFolders.Count == 0;
+
+			var shouldCheckNetworkAvailability =
```
yair100, line 832: Is there a reason we call this code from `UpdateEmptyTextType()`?

0x5bfa: Good catch, fixed.

src/Files.App/Views/MainPage.xaml.cs
```diff
@@ -28,6 +28,7 @@ public sealed partial class MainPage : Page
 		public IUserSettingsService UserSettingsService { get; }
 		private readonly IWindowContext WindowContext = Ioc.Default.GetRequiredService<IWindowContext>();
 		private readonly ICommandManager Commands = Ioc.Default.GetRequiredService<ICommandManager>();
+		private readonly INetworkService NetworkService = Ioc.Default.GetRequiredService<INetworkService>();
```
yair100: I think this can be reverted.

0x5bfa: Reverted.

yair100:
LGTM

PR #18581, opened 2026-06-12
Feature: Added an 'Open With...' toolbar flyout

### LINKED ISSUE #17291: Feature: "Open with" toolbar dropdown

Add an "Open with" dropdown option next to the "New" button in the toolbar.
When a file is selected, the associated applications would show up in the dropdown.

### Requirements

This helps in quickly open a file in different programs. For example, you may open a C file in Notepad++ or Visual Studio Code.

### Files Version

3.9.12.0

### Windows Version

10.0.26100.4652

### Comments

_No response_

0x5bfa:
### Resolved / Related Issues

- Close #17291

1. Open Files app
2. Select a file
3. Click the "Open With" flyout

image

### DIFF
--- a/src/Files.App/Data/Items/ToolbarSections.cs
+++ b/src/Files.App/Data/Items/ToolbarSections.cs
@@ -26,8 +26,6 @@ namespace Files.App.Data.Items
 				[AlwaysVisibleContextId] =
 				[
 					new(commandGroup: nameof(CommandGroups.NewItem), showLabel: true),
-					new(commandGroup: nameof(CommandGroups.OpenWith), showLabel: true),
-					new(commandCode: ToolbarItemDescriptor.SeparatorCommandCode, showIcon: false),
 					new(commandCode: nameof(CommandCodes.CutItem)),
 					new(commandCode: nameof(CommandCodes.CopyItem)),
 					new(commandCode: nameof(CommandCodes.PasteItem)),
--- a/src/Files.App/Data/Items/ToolbarSections.cs
+++ b/src/Files.App/Data/Items/ToolbarSections.cs
@@ -26,6 +26,7 @@ namespace Files.App.Data.Items
 				[AlwaysVisibleContextId] =
 				[
 					new(commandGroup: nameof(CommandGroups.NewItem), showLabel: true),
+					new(commandGroup: nameof(CommandGroups.OpenWith), showLabel: true),
 					new(commandCode: ToolbarItemDescriptor.SeparatorCommandCode, showIcon: false),
 					new(commandCode: nameof(CommandCodes.CutItem)),
 					new(commandCode: nameof(CommandCodes.CopyItem)),
--- a/src/Files.App/Data/Items/ToolbarSections.cs
+++ b/src/Files.App/Data/Items/ToolbarSections.cs
@@ -26,6 +26,7 @@ namespace Files.App.Data.Items
 				[AlwaysVisibleContextId] =
 				[
 					new(commandGroup: nameof(CommandGroups.NewItem), showLabel: true),
+					new(commandCode: ToolbarItemDescriptor.SeparatorCommandCode, showIcon: false),
 					new(commandCode: nameof(CommandCodes.CutItem)),
 					new(commandCode: nameof(CommandCodes.CopyItem)),
 					new(commandCode: nameof(CommandCodes.PasteItem)),
--- a/src/Files.App/Properties/launchSettings.json
+++ b/src/Files.App/Properties/launchSettings.json
@@ -1,6 +1,6 @@
 {
   "profiles": {
-    "MsixPackage": {
+    "Run": {
       "commandName": "MsixPackage"
     }
   }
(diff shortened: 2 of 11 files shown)


src/Files.App/Data/Items/ToolbarSections.cs
```diff
@@ -26,6 +26,7 @@ public static class ToolbarDefaultsTemplate
 				[AlwaysVisibleContextId] =
 				[
 					new(commandGroup: nameof(CommandGroups.NewItem), showLabel: true),
+					new(commandGroup: nameof(CommandGroups.OpenWith), showLabel: true),
```
yair100: @0x5bfa can you please remove this from the always visible group? It should only be visible if the user manually adds it to the toolbar.

src/Files.App/Properties/launchSettings.json
```diff
@@ -1,6 +1,6 @@
 {
   "profiles": {
-    "MsixPackage": {
+    "Run": {
```
yair100, line 3: What's this change for?

0x5bfa: The text "MsixPackage" is shown on the VS code debug button and I thought "Run" is kinda easy to understand. and thought it's not worth creating a separate PR for this. Should we revert this?

image

src/Files.App/Data/Items/ToolbarSections.cs
0x5bfa: Done.

src/Files.App/Data/Items/ToolbarSections.cs
```diff
@@ -26,7 +26,6 @@ public static class ToolbarDefaultsTemplate
 				[AlwaysVisibleContextId] =
 				[
 					new(commandGroup: nameof(CommandGroups.NewItem), showLabel: true),
-					new(commandCode: ToolbarItemDescriptor.SeparatorCommandCode, showIcon: false),
```
yair100, line 29: Is there a reason the separator was removed?

src/Files.App/Properties/launchSettings.json
yair100: It's fine with me.

yair100:
@0x5bfa I tested the flyout and it works quite well. One issue I noticed is that for files with no supported applications, clicking 'Open with' displays an empty flyout. Are we able to disable the toolbar button for these files?

Josh65-2201:
> 'Open with' displays an empty flyout. Are we able to disable the toolbar button for these files?

Instead of disabling the button the drop down could be instead and then it opens the select prompt like it does in 4.1.3. That would also match how the context menu works

yair100:
@Josh65-2201 this PR introduces a new action group 'Open with...' that users can pin to the toolbar, it doesn't change the behavior of the existing 'Open with' action.

Josh65-2201:
Seem odd to have two, Is it not possible to have them combined or keep the `chose another app` option.

yair100:
> or keep the chose another app option.

Displaying 'Choose another app' when no options are available sounds like a good solution.

src/Files.App/Data/Items/ToolbarSections.cs
0x5bfa: No, reverted.

0x5bfa:
Done.

image

yair100:
Looks good. One more thing to consider is the behavior when multiple files are selected. Since the 'Choose another app' picker only supports a single file, we should disable when multiple files are selected.

mdtauk:
> Looks good. One more thing to consider is the behavior when multiple files are selected. Since the 'Choose another app' picker only supports a single file, we should disable when multiple files are selected.

What if all the selected files are of the same or compatible types?  Like a group of JPG and PNG images, all being openable in Photoshop or some other multi-document graphics package?

yair100:
> What if all the selected files are of the same or compatible types? Like a group of JPG and PNG images, all being openable in Photoshop or some other multi-document graphics package?

The open file picker only supports a single file. It doesn't make a difference if all of the items are the same type.

0x5bfa:
The context menu in the Explorer just hides that menu item. May want to disable this drop down.

0x5bfa:
Done.

src/Files.App/UserControls/Toolbar.xaml.cs
```diff
@@ -393,13 +414,88 @@ private void PopulateGroupFlyout(MenuFlyout flyout, CommandGroup group)
...
+		}
+
+		private static async Task<MenuFlyoutItem> CreateShellNewEntryMenuItemAsync(ShellNewEntry entry)
 		{
 			if (!string.IsNullOrEmpty(entry.IconBase64))
 			{
 				using var ms = new MemoryStream(Convert.FromBase64String(entry.IconBase64));
 				var image = new BitmapImage();
-				_ = image.SetSourceAsync(ms.AsRandomAccessStream());
+				await image.SetSourceAsync(ms.AsRandomAccessStream());
```
yair100, line 498: Is there a reason this was changed to await?

0x5bfa: https://github.com/files-community/Files/blob/ae2e045d1102f1d42defcd89364f5bdd5d2a48dd/src/Files.App/UserControls/Toolbar.xaml.cs#L492-L502

If you don't wait, `ms` can be disposed when it's out of scope.

yair100 (approved):
LGTM

Josh65-2201 (approved):

PR #18602, opened 2026-06-16
Feature: Improved 'Open in new pane' context menu

yair100:
**Resolved / Related Issues**

Improved the 'Open in new pane' context menus throughout the app to allow choosing the split arrangement.

image

### DIFF
--- a/src/Files.App/Actions/Navigation/OpenInNewPane/BaseOpenInNewPaneAction.cs
+++ b/src/Files.App/Actions/Navigation/OpenInNewPane/BaseOpenInNewPaneAction.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Files Community
+// Copyright (c) Files Community
 // Licensed under the MIT License.
 
 namespace Files.App.Actions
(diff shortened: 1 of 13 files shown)


Josh65-2201 (commented):
The action is disabled if a pane is open, 4.1.3 doesn't disable it.

yair100:
This is by design, we only allow two panes at a given time. If there is enough interest, we can add another action to 'Open in other pane'.

Josh65-2201:
Why not rename it `Open in pane` that wording would allow both new and existing, then the action can replace the other panes folder like 4.1.3 does

yair100:
The current behavior is outdated and dates back to before Files had support for changing the pane arrangement. The reason the behavior needs to change with this PR is that it allows users to choose the arrangement of the new pane. If Dual Pane is already active, we can't let users choose the pane arrangement since they aren't opening a new pane, they are opening a folder in an existing pane.

That being said, we can replace 'Open in new pane' with 'Open in other pane' when Dual Pane is already active.

Josh65-2201 (approved):

PR #18644, opened 2026-06-25
Code Quality: Removed the MMI reference for AOT compatibility

0x5bfa:
### Resolved / Related Issues

This project has been abandoned and replaced by `System.Management` but this is also AOT unsafe. Plus, we no longer have to use this safe code after we migrated to WASDK. I've replaced the MMI references with COM APIs.

- Related to #4180

1. Open Files app
2. Insert/eject a removable media
3. See the drive list is updated


yair100:
This needs some testing.

Josh65-2201 (dismissed):

src/Files.App.Storage/Windows/Managers/WindowsDriveManager.cs
```diff
@@ -0,0 +1,93 @@
...
+// Licensed under the MIT License.
+
+using Windows.Win32;
+using Windows.Win32.UI.Shell;
+
+namespace Files.App.Storage
+{
+	public unsafe sealed class WindowsDriveManager : IDisposable
+	{
+		private static readonly Lazy<WindowsDriveManager> lazy = new(() => new WindowsDriveManager());
```
hez2010: You don't need to use a lazy here. static fields are already only initializing on the first access.

src/Files.App.Storage/Windows/Managers/WindowsFolderChangeWatcher.cs
```diff
@@ -0,0 +1,227 @@
...
+			fixed (char* pszWndClassName = _windowClassName)
+			{
+				_wndProc ??= new(WndProc);
+
+				WNDCLASSEXW wndClass = default;
+				wndClass.cbSize = (uint)sizeof(WNDCLASSEXW);
+				wndClass.hInstance = PInvoke.GetModuleHandle(default(PCWSTR));
+				wndClass.lpszClassName = pszWndClassName;
+				wndClass.lpfnWndProc = (delegate* unmanaged[Stdcall]<HWND, uint, WPARAM, LPARAM, LRESULT>)
+					Marshal.GetFunctionPointerForDelegate(_wndProc);
```
hez2010, line 74: Is it possible to make `WndProc` static, marking it as `UnmanagedCallersOnly` so that we can take a function pointer directly instead of acquiring a function pointer from an instance delegate?

src/Files.App.Storage/Windows/Managers/WindowsDriveManager.cs
0x5bfa: Thanks! I've removed Lazy and just used `new`. In the future, I'd like to put this to the root app model and avoid `Default` but we can't do it but do this.

src/Files.App.Storage/Windows/Managers/WindowsFolderChangeWatcher.cs
0x5bfa: No, since the window proc accesses instance-only `FolderChanged` event.

hez2010 (approved):
Thanks.

yair100 (approved):

PR #18645, opened 2026-06-25
Fix: Fixed issue where toolbar buttons were sometimes missing

### LINKED ISSUE #18577: Bug: Initial-load toolbar state race that could leave the right-side toolbar buttons showing only the filter button instead of all 5 buttons.
### Description

There is a race condition in the UI logic that results in only the toolbar right side `Filter` button showing instead of all 5 (`Filter`, `Selection Options`, `Sort`, `Layout`, `Toggle the info pane`) showing when a new tab is opened and switched to the first time.

This may only affect slower systems as it was tested inside a VM.

### Steps To Reproduce

1. Open `Files`
2. Open a folder in a new tab
3. Switch to the new tab

## Expected behavior:
Toolbar right side should show all 5 buttons.

## Actual behavior:
Only the filter button is shown until the tab refreshes from switching or opening another new tab.

### Files Version

4.1.1.0

### Windows Version

Win10 19045.4529

### User ID

46719796-b353-47af-baa3-0df7ebf5ff6d

### Log File

[debug.log](https://github.com/user-attachments/files/28844047/debug.log)

yair100:
**Resolved / Related Issues**

- Closes #18577

Verification required


Josh65-2201 (changes requested):
This is still happen to the info pane toggle when starting Files on a network drive, others are working

Josh65-2201 (approved):

PR #18646, opened 2026-06-25
Fix: Fixed issue where search continued running after navigating away or closing a tab

### LINKED ISSUE #18393: Bug: Search functions continue querying even after exiting search or closing tab.
### Description

Image
Continued querying after tab close.

### Steps To Reproduce

1. search for a file that might take longer to query (eg, search for a filename in the C: drive itself
2. Clear the search, still continues
3. Navigate to other folder, Search continues
4. close tab, Search is still occuring

### Files Version

4.0.40.0 Files-Preview and 4.0.38.0 Files-Stable

### Windows Version

Windows 11 25H2

### User ID

n/r

### Log File

[debug.txt](https://github.com/user-attachments/files/26918979/debug.txt)

workbysaran:
**Resolved / Related Issues**
Fixed an issue where search queries kept running in the background even after the user cleared the search, navigated to a different folder, or closed the tab, causing sustained CPU usage.

Closes #18393

1. Open Files and navigate to a large drive
2. Open the search bar and type a query slowly to confirm suggestion searches cancel per keystroke and only fire after a ~200ms pause
3. Type a full query and press Enter and confirm search results load correctly
4. While results are loading, navigate to a different folder and confirm the search stops promptly

**Note:** A brief CPU in Files.exe after cancellation is expected due to Windows Search runs as an in-process COM server and its native thread pool work drains briefly after the .NET task unblocks.  Please find the profiler screenshot attached
image

### DIFF
--- a/src/Files.App/UserControls/NavigationToolbar.xaml.cs
+++ b/src/Files.App/UserControls/NavigationToolbar.xaml.cs
@@ -422,6 +422,11 @@ namespace Files.App.UserControls
 			}
 			else
 			{
+				if (Omnibar.CurrentSelectedMode == OmnibarSearchMode)
+				{
+					ViewModel?.CancelSuggestionSearch();
+				}
+
 				// When Omnibar loses focus, revert to Path Mode to display BreadcrumbBar
 				Omnibar.CurrentSelectedMode = OmnibarPathMode;
 			}
--- a/src/Files.App/Utils/Storage/Search/FolderSearch.cs
+++ b/src/Files.App/Utils/Storage/Search/FolderSearch.cs
@@ -92,6 +92,7 @@ namespace Files.App.Utils.Storage
 			}
 			catch (OperationCanceledException)
 			{
+				return;
 			}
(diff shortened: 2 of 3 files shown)


yair100 (approved):

Josh65-2201:
I'm not certain the search is stopping, Files shows that it's not loading anything, but CPU usage goes from 0.1% when idle to around 10% start of search and stays there for about 10 seconds before decreasing only a lit bit around 5-8% when backing out to the home page or empty folder. Although this is much less than 4.1.5 staying around 25-30% after backing out

Search results are shown same as 4.1.5

workbysaran:
@Josh65-2201 

The remaining CPU after cancellation comes from Windows Search running as an in-process COM server inside Files.exe. When cancelled, the .NET task stops immediately but the native search work already in the thread pool drains briefly.

Beyond this fix, moving the search execution out-of-process would eliminate this entirely.

Josh65-2201 (approved):
As long as that amount is expected, all good

PR #18669, opened 2026-07-01
Code Quality: Moved `Checkout` operation into Git abstraction

Lamparter:
**Resolved / Related Issues**

- Related #16738

**Steps used to test these changes**

The implementation is logically identical to the previous, it has just been moved.

Issue #16738, opened 2025-01-26
Feature: Use installed git.exe instead of libgit2


heftymouse:
### What feature or improvement do you think would benefit Files?

Having this as an option would allow Files to easily support the full range of Git features, including automatically respecting the user's Git settings, using any authentication methods they have already configured, and pushing to ssh remotes. This is a fairly common way of implementing Git support, seen in software like [vscode](https://code.visualstudio.com/docs/sourcecontrol/faq#_why-is-vs-code-warning-me-that-the-git-repository-is-potentially-unsafe). Files can fall back to libgit2 for basic support if Git is not already installed (unlikely on a developer's machine)

### Requirements

- [ ] Introduce an `IGitService` abstraction and refactor the existing `GitHelpers` implementation to use it
- [ ] Implement existing Git operations through git.exe and use this implementation if git.exe is found
- [ ] Add configuration options

### Files Version

3.8.13

### Windows Version

10.0.27774.1000

Lamparter:
I wrote [just the thing](https://github.com/Lamparter/CommitSharp) for this!
Might need a bit more work, but it's a start.

0x5bfa:
I remember someone told me porcelain is better, perhaps you, and I think this is a good idea. Here's my thoughts:

I have been thinking about this being a standalone class like below and I think this would be instantiated through GitRepositoryFactory static class (e.g. `GetOrCreate()`) to retrieve one identical instance per the same git repo among tabs, panes and windows.

```c#
public static class GitRepositoryFactory
{
    private Dictionary<string, GitRepository> lookupTable = [];

    public GitRepository? GetOrCreate(string path) {}
}
```

```c#
public partial sealed class GitRepository : IDisposable
{
    public GitRepository(string rootPath) {} // Ideally, not string but IFolder in the future.
}
```

I still haven't evaluate how difficult this is and how suitable this is so I can't say decisive thing but IGitRepositorySevice would work in this current codebase.

yaira2:
> Introduce an IGitService abstraction and refactor the existing GitHelpers implementation to use it

I need more time to review the other proposals, but we can certainly implement an abstraction for the existing git implemention.

Lamparter:
@yaira2 what do you think?
I have a [working `IGitService` implementation](https://github.com/Lamparter/CommitSharp/blob/main/src/Riverside.CommitSharp.Abstractions/IVersionControl.cs) that I can implement here quite easily if you like.

yair100:
[...]

PR #18679, opened 2026-07-05
Code Quality: Refactored LibGit2 integration as a DI service for future-proofing

Lamparter:
**Resolved / Related Issues**

- Related #16738

The implementation is logically identical to the previous, it has just been moved.

---

In this PR, I:
- converted the previous LibGit2.cs (that aims to implement IVersionControl) into a DI service named LibGit2Service
- fixed the IsExecutingGitAction field as well as both event handlers IsExecutingGitActionChanged and GitFetchCompleted, as raised by @yair100 in https://github.com/files-community/Files/pull/18669#issuecomment-4871166917
- moved FetchOrigin to the new service (because it invokes the event handler, it must be part of the type containing the event handler [as per C# convention](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0079?f1url=%3FappId%3Droslyn%26k%3Dk(CS0079)))

this future proofs the git abstraction because we will want to move to a DI service later anyway (because the end goal is to allow the user to choose how the app controls git, i.e. they can switch between libgit2sharp and git.exe) and fixes the issue Yair raised before at the same time :)

### DIFF
--- a/src/Files.App/Data/Contracts/IVersionControlService.cs
+++ b/src/Files.App/Data/Contracts/IVersionControlService.cs
@@ -4,7 +4,7 @@
 using LibGit2Sharp;
 using System.ComponentModel;
 
-namespace Files.App.Utils.Git
+namespace Files.App.Data.Contracts
 {
 	/// <summary>
 	/// Defines a version control abstraction.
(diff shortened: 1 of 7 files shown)


yair100 (approved):

0x5bfa (approved):
lgtm

PR #18733, opened 2026-07-18
Fix fr-FR: rename legacy dotted keys, add 9 missing translations

Bpl5966:
This PR fixes 64 strings that Crowdin reports as missing in fr-FR/Resources.resw.

Investigation showed that 55 of these already had a French translation, but under
an old dotted naming convention (e.g. `RenameError.ItemDeleted.Text`) that no longer
matches the current en-US key names (e.g. `RenameErrorItemDeletedText`). This PR
renames those 55 keys to match en-US, preserving the existing translations.

The remaining 9 strings were genuinely untranslated and have been translated:
- AutoFitColumns / AutoFitColumnsDescription
- AutoSizeColumnsInDetailsLayout
- CannotRunFileDialogText / CannotRunFileDialogTitle
- CompressSkippedItemsDialogText / CompressSkippedItemsDialogTitle
- OpenCurrentFolderInOtherPane / OpenCurrentFolderInOtherPaneDescription

After this change, fr-FR has 0 missing keys compared to en-US.

### DIFF
--- a/src/Files.App/Strings/fr-FR/Resources.resw
+++ b/src/Files.App/Strings/fr-FR/Resources.resw
@@ -180,10 +180,10 @@
   <data name="Accessed" xml:space="preserve">
     <value>Accédé :</value>
   </data>
-  <data name="RecentItemClearAll.Text" xml:space="preserve">
+  <data name="RecentItemClearAllText" xml:space="preserve">
     <value>Effacer tous les éléments</value>
   </data>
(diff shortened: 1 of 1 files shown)


yair100:
@Bpl5966 thank you for contributing to Files. Translations for Files are done through the popular cloud based platform Crowdin. Modifying strings outside of Crowdin causes issues with syncing.

PR #18735, opened 2026-07-18
Fix: Fixed an issue where start menu pins had the wrong icon

### LINKED ISSUE #12191: Bug: Pinning an EXE to Start menu results in displaying the wrong icon (Files app icon) in Start menu
### Description

When we pin an app (specifically, an EXE) to Start, the app icon in Start menu is that of Files app instead of it being that of the pinned app.

### Steps To Reproduce

1. Open Files app
2. Locate any executable file 
3. Right click > Show more options
4. Tap "Pin to the Start menu"
5. Open Start menu and observe the issue. While the app is correctly pinned, the icon is that of Files app instead of the EXE's app icon.

🧓🏻☁️⚠️
Rafael provided this GUID for your records: c65c0d4d-7e6a-4b3a-a2f6-9181f0f0743d

### Requirements

Ideally, the pinned icon should be that of the executable file, not the app that made the pin.

### Files Version

Version: 2.4.68.0, OS Version: 22624.1616

### Windows Version

Windows 11 Beta 10.0.22624.1616

### Log file

[debug.log](https://github.com/files-community/Files/files/11314220/debug.log)

Lamparter:
### Resolved / Related Issues

- Closes #12191

_This PR is pretty complicated, but this is how you can test every part of it:_

- **The pin/unpin from start actions are now only executable if they are appropriate (you can only pin if the item is not already pinned, and you can only unpin if the item is already pinned)**

	1. Opened the app
	2. Selected a single item in the file browser
	3. Observed that the 'unpin' action cannot be executed whereas the 'pin' action can
	4. Ran the 'pin' action on the file via the omnibar
	5. Observed that the 'pin' action can no longer be executed on that item whereas the 'unpin' action can
	6. Ran the 'unpin' action on the file via the omnibar
	7. Observed that the state reset and now the unpin button is greyed out again

- **Pinning/unpinning is possible and works correctly when multiple files/folders are selected**
- **Pinning an item to the start menu now shows the item's icon in the start menu**

	1. Pinned a file to the Start menu via the 'pin' action
	2. Ran the 'pin' action and accepted the dialogue
	3. Opened the Start menu
	4. Observed that the item is now pinned, and that the file thumbnail is its icon

		image

- **Start menu tile IDs are now fully unique**

	1. Pinned a file to the Start menu via the 'pin' action
	2. Opened `%localappdata%\FilesDev_ykqwq8d6ps0ag\LocalState\startmenu` in File Explorer
	3. Observed that there are now thumbnail images with a GUID name (e.g. `5a346596-28e7-bf9a-1743-91335747ad8e.png`) - each image name corresponds to its identical pinned start tile ID (this ensures that start tile IDs will _always_ be the same)
	5. Observed that unpinning a start tile (using the 'unpin' action) deletes the image file

### DIFF
--- a/src/Files.App/Services/Windows/WindowsStartMenuService.cs
+++ b/src/Files.App/Services/Windows/WindowsStartMenuService.cs
@@ -103,9 +103,6 @@ namespace Files.App.Services
 			{
 				using (var managedIcon = Icon.ExtractAssociatedIcon(file.Id))
 				{
-					//if (managedIcon is null)
-					//	return new Uri($"ms-appx:///Assets/AppTiles/Release/Square44x44Logo.scale-100.png");
-
 					using (var bitmap = managedIcon!.ToBitmap())
 					{
 						bitmap.Save(iconPath, ImageFormat.Png);
@@ -120,9 +117,6 @@ namespace Files.App.Services
 
 				using (var managedIcon = Icon.ExtractIcon(Path.Combine(Environment.SystemDirectory, "shell32.dll"), shell32IconId))
 				{
-					if (managedIcon is null)
-						return new Uri($"ms-appx:///Assets/AppTiles/Release/Square44x44Logo.scale-100.png");
-
 					using (var bitmap = managedIcon!.ToBitmap())
 					{
 						bitmap.Save(iconPath, ImageFormat.Png);
--- a/src/Files.App/Services/Windows/WindowsStartMenuService.cs
+++ b/src/Files.App/Services/Windows/WindowsStartMenuService.cs
@@ -29,8 +29,11 @@ namespace Files.App.Services
 
 			try
 			{
-				var path150x150 = new Uri("ms-appx:///Assets/tile-0-300x300.png");
-				var path71x71 = new Uri("ms-appx:///Assets/tile-0-250x250.png");
+				var path150x150 = new Uri("ms-appx:///Assets/AppTiles/Preview/Square150x150Logo.scale-100.png");
+				var path71x71 = new Uri("ms-appx:///Assets/AppTiles/Preview/Small71x71Logo.scale-100.png");
+				var path44x44 = new Uri("ms-appx:///Assets/AppTiles/Preview/Square44x44Logo.scale-100.png");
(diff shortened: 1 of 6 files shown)


Lamparter (commented):
- [x] You can no longer pin any item that is not an executable
- [ ] There is a bad framing border (might not be possible to fix)
- [x] Clicking on tiles that link to a file shows an error in the app
  - [x] #18740
- [x] The app cannot process file open events **(change activation handler)**
- [x] Use completely unique start menu IDs (Guid)
- [ ] It's possible to pin an item inside an archive
- [x] Some files do not open correctly

image

The top zip file opens correctly, but the bottom one does not. Both open in File Explorer when they ought to open in Files (because they are zip files)

src/Files.App/Services/Windows/WindowsStartMenuService.cs
```diff
@@ -78,5 +99,37 @@ private static string GetNativeTileId(string id)
 
 			return str;
 		}
+
+		private static unsafe Uri ExtractFileIcon(IStorable file, string id)
```
Lamparter: I created my own method but perhaps using a combination of [`TryExtractImageFromDll`](https://github.com/files-community/Files/blob/d3b400fcb0802d1502073701804bf9ea7a2fdcb0/src/Files.App.Storage/Windows/Helpers/WindowsStorableHelpers.Icon.cs#L102) and `TryGetThumbnail` would be more appropriate

Lamparter:
If a file icon cannot be obtained, the tile defaults to the app icon

image

Josh65-2201:
> You can no longer pin any item that is not an executable

Folders are meant to be allowed to pin

Lamparter:
Yes, that checklist is for the bugs I accidentally introduced in this PR that I'm working to fix / parts of the PR that aren't finished yet

Lamparter:
IsZipFile detects certain files like ".appxbundle" as zip files even though Files doesn't actually support them. This causes issues when trying to programmatically open those types of files (i.e. when pinning one of those files to the start menu). While I admit a scenario in which someone would try to pin that type of file to the start menu is rare, it's an issue that should still be resolved.

Issue #17454, opened 2025-08-16
Bug: Launching a corrupted executable from Files doesn't display the "This app can't run on your PC" dialog


IviriusMain:
### Description

Usually, a dialog is displayed when this happens. It should also be the case for Files, but it fails silently instead.

### Steps To Reproduce

1. Create a file named "a.exe"
2. Double click it
3. Nothing happens

### Files Version

3.9.16.0

### Windows Version

10.0.26100.3775

### Log File

[debug.log](https://github.com/user-attachments/files/21817933/debug.log)

Says right here as well: `System.ComponentModel.Win32Exception (193): [...] The specified executable is not a valid application for this OS platform.`

IviriusMain:
This should be an easy fix, I'll add it to my to-do list for tomorrow.

Josh65-2201:
Thanks for the report

IviriusMain:
> This should be an easy fix, I'll add it to my to-do list for tomorrow.

I was wrong, it's not an easy fix

Lamparter:
It's being discussed on discord.

yair100:
@IviriusMain this issue will be fixed in the next release.

PR #18739, opened 2026-07-18
Code Quality: Fixed an issue where the "This app can't run on your PC" dialogue displayed inappropriately

### LINKED ISSUE #18737: Bug: "This app can't run on your PC" dialogue sometimes inappropriately displays
### Description

Since #18721, sometimes the "This app can't run on your PC" dialogue from #17454 will appear despite it being inappropriate to do so.

Image

_^^^ for example, when opening an SVG file_

It's possible that this is an issue with my machine, but otherwise it will require immediate rollback of #18721.

### Steps To Reproduce

1. Open an SVG, PDF, HTML, MP4, ISO, APPXBUNDLE, WOFF2, NUPKG, DOCX, PUB, etc. file
2. See the dialogue open inappropriately

### Files Version

v4.2/[`d3b400f`](https://github.com/files-community/Files/commit/d3b400fcb0802d1502073701804bf9ea7a2fdcb0)

### Windows Version

10.0.26200.8875

### Log File

[debug.log](https://github.com/user-attachments/files/30153713/debug.log)

Lamparter:
**Resolved / Related Issues**

- Closes #18737 
- Related #17454

1. Created a file "dnskjdnjkf.exe" with bad binary contents
2. Opened Files
3. Opened the executable via Files and observed that the "This app can't run on your PC" dialogue still displays
4. Opened one of the erroneous file types as listed in #18737 (e.g. an HTML file)
5. Observed that the file opens correctly and the "This app can't run on your PC" did not appear

Note that creating an executable file with _no_ contents will not result in the dialogue showing anymore. However, this is justified by the fact that in practice, a user will never be opening an executable file with no binary contents unless they are just playing around with their PC.

### DIFF
--- a/src/Files.App/Utils/Shell/LaunchHelper.cs
+++ b/src/Files.App/Utils/Shell/LaunchHelper.cs
@@ -128,9 +128,10 @@ namespace Files.App.Utils.Shell
 
 				return true;
 			}
-			catch (Win32Exception ex) when (ex.NativeErrorCode is 193 or 216)
+			catch (Win32Exception ex) when (ex.NativeErrorCode is 216)
 			{
-				// ERROR_BAD_EXE_FORMAT (193) / ERROR_EXE_MACHINE_TYPE_MISMATCH (216)
+				// ERROR_EXE_MACHINE_TYPE_MISMATCH (216)
+				// "This app can't run on your PC"
 				await DialogDisplayHelper.ShowDialogAsync(DynamicDialogFactory.GetFor_CannotRunFileDialog());
 				return false;
 			}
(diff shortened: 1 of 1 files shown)


src/Files.App/Utils/Shell/LaunchHelper.cs
```diff
@@ -128,9 +128,10 @@ private static async Task<bool> HandleApplicationLaunch(string application, stri
 
 				return true;
 			}
-			catch (Win32Exception ex) when (ex.NativeErrorCode is 193 or 216)
+			catch (Win32Exception ex) when (ex.NativeErrorCode is 216)
```
yair100:
```suggestion
			catch (Win32Exception ex) when (ex.NativeErrorCode is 193 or 216 && FileExtensionHelpers.IsExecutableFile(application, exeOnly: true))
```

This seems like a more reliable solution.

yair100 (approved):
LGTM

PR #18742, opened 2026-07-19
Feature: Added Robocopy Multithreaded support

### LINKED ISSUE #7275: Please add multi-threaded copy support
### What's the Problem?

I was comparing both windows explorer and Files app copy speed of 1600 files. Here are the results. 
image
Files app took 43 seconds to copy 1600 files of 300 kb each
image
Whereas windows just took 5-6 seconds to fully copy the folder. 

### Solution/Idea

Add customizable multi-threaded support in the Files app. I want to set it as 16 or 32 threads (especially when I am copying from google drive to the desktop, multi-threading is essential). 

### Alternatives

You may use robocopy to copy files fast and multi-threaded (worked for my 150 Mbps internet, google drive downloads maxed 145 Mbps, normally it just gives 20 Mbps). 

You may also include file explorer's file handling option in settings to copy files, etc. 

### Priorities

| Capability | Priority |
| Multi-threaded Support | Must |
| Option to choose the number of threads in settings | Should |
| Option to change default file handler to native windows or robocopy | Could |

### Files Version

_No response_

### Windows Version

_No response_

### Comments

Well, it's an important feature to copy files fast.

DarkEden-coding:
This PR implements support for Robocopy multithreading in file operations. It should be able to robustly handle a variety of situations, such as copy‑pasting, moving, and file deletion as well. During testing for large numbers of small files, this can increase transfer rate by up to ten times. In general, this has been a requested feature for a while and will increase parity with the Windows File Explorer.

This is my first pr so sorry if I messed anything up!

**Basic code changes**

| File | Change | Methodology |
|---|---|---|
| `IDevToolsSettingsService.cs` | Added `UseRobocopyForFileOperations` and `RobocopyThreads` | Exposes Robocopy preferences on the existing Dev Tools settings contract |
| `DevToolsSettingsService.cs` | Persists those settings, defaulting to disabled and 8 threads | Stores them with the other Dev Tools JSON settings |
| `DevToolsViewModel.cs` | Binds the toggle and thread count to the service | Provides the MVVM surface for the Dev Tools page |
| `DevToolsPage.xaml` | Adds a Robocopy expander and thread-count `NumberBox` | Adds experimental settings UI under Dev Tools |
| `Resources.resw` | Adds Robocopy label and description strings | Provides localized UI text |
| `ShellFilesystemOperations.cs` | Routes large copy, move, and delete operations to Robocopy when enabled | Uses Shell operations for small operations and renames, and Robocopy for bulk operations when safe |
| `FileOperationsHelpers.cs` | Adds a Robocopy process runner, batching, progress reporting, and delete-via-`/MIR` | Calls `robocopy.exe` with `/MT`, verifies results, and supports cancellation through a `CancellationTokenSource` |
| `FilesystemHelpers.cs` | Awaits batched Jump List folder removal after delete and move operations | Avoids bursts of COM updates after bulk operations |
| `IWindowsJumpListService.cs` | Adds `RemoveFoldersAsync` | Provides a batch API for multi-folder removal |
| `WindowsJumpListService.cs` | Uses a single Jump List load/save operation for multiple paths | Deduplicates removals; `RemoveFolderAsync` delegates to the batch method |
| `StatusCenterItem.cs` | Retains only the first source and destination path | Reduces memory retained by large completed operations |
| `CopyEngineResult.cs` | Maps `COPYENGINE_E_USER_CANCELLED` to `Generic` | Treats cancellation as a known status rather than an unknown HRESULT |

**Resolved / Related Issues**

- Closes #7275 

1. Opened Files
2. Created 2 folders with random content (10 gb of random sized files)
3. Tested by repeatedly copying these files back and forth using cut, paste, regular copy and paste, and trash as well as permanent deletion.
4. Also tested for between-drive copy and pastes.

Required testing:

1. Copying between a local drive and a cloud drive.
2. In general, more variety in the amount of items transferred and different locations transferred to.


TechyAditya:
Hi team, please review and merge this PR so that I could close the issue I have raised 5 years ago. As this is a dev option, the testing could be done by public users and according to the users feedback, this implementation can be made as default choice if this implementation turns out to be a better choice. This could make me uninstall teracopy application.

Josh65-2201:
Your beach is quite old to build, could you rebase it with main.

DarkEden-coding:
@Josh65-2201 its updated, seems to be working the same with a basic test

Josh65-2201:
What is considered large for testing, the amount of files or size? Below is what i can think of should be tested, if you have anything else please let me know

Delete: Local `to test`, MTP (phone) `to test`, SMB network `to test`, Alternative data steams `to test`
Move: Local `to test`, MTP (phone) `to test`, SMB network `to test`
Copy: Local `to test`, MTP (phone) `to test`, SMB network `to test`
Replace: Local `to test`, MTP (phone) `to test`, SMB network `to test`

TechyAditya:
Hi Josh, the number of files are considered as large for testing because bigger files have consistent speeds of transfers, and this PR has 0 advantage on few file transfers. This would shine if you try transferring something like python venv, node_modules folder of nodejs apps, etc. on a very bloated developer project, for example.

src/Files.App/Views/Settings/AdvancedPage.xaml
```diff
@@ -206,6 +206,36 @@
...
+			<wctcontrols:SettingsExpander
+				Description="{helpers:ResourceString Name=UseRobocopyForFileOperationsDescription}"
+				Header="{helpers:ResourceString Name=UseRobocopyForFileOperations}"
+				IsExpanded="False">
+				<wctcontrols:SettingsExpander.HeaderIcon>
+					<FontIcon Glyph="&#xE8B5;" />
+				</wctcontrols:SettingsExpander.HeaderIcon>
+				<ToggleSwitch AutomationProperties.Name="{helpers:ResourceString Name=UseRobocopyForFileOperations}" IsOn="{x:Bind ViewModel.UseRobocopyForFileOperations, Mode=TwoWay}" />
+				<wctcontrols:SettingsExpander.ItemsHeader>
+					<Grid
```
yair100: Please use a SettingsCard for this. See the cache thumbnails setting for reference.

Josh65-2201:
Tested with 50 .mp3 files, haven't had any issues. I can test MTP if that is need later

- Delete: 
  - From local `Works`
  - From MTP (phone) `to test`
  - From SMB network `Works`
  - Alternative data steams `Works`

- Move: 
  - To local `Works`
  - From local `Works`
  - To MTP (phone) `to test`, 
  - From MTP (phone) `to test`, 
  - To SMB network `Works`
  - From SMB network `Works`

- Copy: 
  - To local `Works`
  - From local `Works`
  - To MTP (phone) `to test`, 
  - From MTP (phone) `to test`, 
  - To SMB network `Works`
  - From SMB network `Works`

- Replace
  - To local `Works`
  - From local `Works`
  - To MTP (phone) `to test`, 
  - From MTP (phone) `to test`, 
  - To SMB network `Works`
  - From SMB network `Works`

Josh65-2201 (approved):

Josh65-2201 (changes requested):
Just found that a folder in use dialog doesn't show with Robocopy enabled. 
I selected a single folder on my SMB drive, status center show it was successful deleted but will show in the file list again after a second.
This isn't affecting local folders and doesn't happen on 4.2.2.0

DarkEden-coding:
Got it, I am on vacation at the moment, I can fix that in a little less than a week

DarkEden-coding:
> Just found that a folder in use dialog doesn't show with Robocopy enabled. I selected a single folder on my SMB drive, status center show it was successful deleted but will show in the file list again after a second. This isn't affecting local folders and doesn't happen on 4.2.2.0

Should be fixed. The best I can see is that that was an issue where robocopy errors during deletion weren't being propagated correctly, but it could have been something else. If you could test out my new change and see if that's fixed it.

Josh65-2201 (dismissed):

src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
```diff
@@ -373,7 +385,9 @@ public async Task<IStorageHistory> DeleteItemsAsync(IList<IStorageItemWithPath>
 			var operationID = Guid.NewGuid().ToString();
 			await using var r = cancellationToken.Register(CancelOperation, operationID, false);
 
-			var (success, response) = await FileOperationsHelpers.DeleteItemAsync(deleteFilePaths.ToArray(), permanently, MainWindow.Instance.WindowHandle.ToInt64(), asAdmin, progress, operationID);
+			var (success, response) = permanently && Ioc.Default.GetRequiredService<IDevToolsSettingsService>().UseRobocopyForFileOperations
```
d2dyno1: Can the GetRequiredService actually cache the service in the class? (And in other places too)

src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
```diff
@@ -511,17 +525,35 @@ public async Task<IStorageHistory> MoveItemsAsync(IList<IStorageItemWithPath> so
...
 			var moveResult = new ShellOperationResult();
 
+			// Small operations stay on the shell path to avoid Robocopy process overhead.
+			var preferRobocopy = source.Count >= ROBOCOPY_MIN_ITEM_COUNT &&
+				Ioc.Default.GetRequiredService<IDevToolsSettingsService>().UseRobocopyForFileOperations;
+
 			if (sourceRename.Any())
 			{
-				var (status, response) = await FileOperationsHelpers.MoveItemAsync(sourceRename.Select(s => s.Path).ToArray(), destinationRename.ToArray(), false, MainWindow.Instance.WindowHandle.ToInt64(), asAdmin, progress, operationID);
+				var renameItems = sourceRename.Zip(destinationRename, (src, dest) => (src, dest)).ToList();
```
d2dyno1: Use ToArray

src/Files.App/Services/Windows/WindowsJumpListService.cs
```diff
@@ -109,19 +109,37 @@ public async Task RefreshPinnedFoldersAsync()
...
+				if (pathsToRemove.Count == 0)
+					return;
+
+				var instance = await JumpList.LoadCurrentAsync();
+				// Disable automatic jumplist. It doesn't work.
+				instance.SystemGroupKind = JumpListSystemGroupKind.None;
 
-					var itemToRemove = instance.Items.Where(x => x.Arguments == path).Select(x => x).FirstOrDefault();
-					instance.Items.Remove(itemToRemove);
+				var itemsToRemove = instance.Items
```
d2dyno1, line 135: ToArray

src/Files.App/Utils/StatusCenter/StatusCenterItem.cs
```diff
@@ -202,8 +202,10 @@ public StatusCenterItem(
 			SpeedGraphValues = [];
 			CancelCommand = new RelayCommand(ExecuteCancelCommand);
 			Message = Strings.DiscoveringItems.GetLocalizedResource();
-			Source = source;
-			Destination = destination;
+			// Status text only uses the first path. Retaining every path keeps large completed
+			// operations and their source storage objects alive for the lifetime of the card.
+			Source = source?.Take(1).ToArray();
+			Destination = destination?.Take(1).ToArray();
```
d2dyno1, line 210: No changes needed here, but I wonder whether we actually need `Source` and `Destination`. At the very least, they could be made into arrays. There are no usages of them here; however, I didn't check other files that might use them

src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
DarkEden-coding: You right, I switched it to store it once at the top of the relevant files.

src/Files.App/Utils/StatusCenter/StatusCenterItem.cs
DarkEden-coding: They are used in src/Files.App/Utils/StatusCenter/StatusCenterHelper.cs:624-640 and are used to build the Status Center card, I have converted it to arrays.

d2dyno1 (approved):
No more remarks on my end :+1:

yair100:
This seems to be failing the build pipelines. Can you please take a look?

DarkEden-coding:
Looks like some commits to main changed some contract stuff, should work now

src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
```diff
@@ -11,39 +11,40 @@ namespace Files.App.Utils.Storage
 	/// </summary>
 	public sealed partial class ShellFilesystemOperations : IFilesystemOperations
 	{
-		private readonly IStorageTrashBinService StorageTrashBinService = Ioc.Default.GetRequiredService<IStorageTrashBinService>();
+		/// <summary>
+		/// Operations at or below this item count stay on the shell path to avoid Robocopy process overhead.
+		/// </summary>
+		private const int ROBOCOPY_MIN_ITEM_COUNT = 21;
```
yair100: Named `MIN` with value 21, but the doc comment describes 20 ("at or below this item count"). Either `= 20` with `>`, or rename to `ROBOCOPY_ITEM_COUNT_THRESHOLD`.

src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
```diff
@@ -76,8 +77,13 @@ public ShellFilesystemOperations(IShellPage associatedInstance)
 			var result = (FilesystemResult)true;
 			var copyResult = new ShellOperationResult();
 
+			// Small operations stay on the shell path to avoid Robocopy process overhead.
+			var canUseRobocopy = source.Count >= ROBOCOPY_MIN_ITEM_COUNT &&
```
yair100: `source.Count` is the top-level selection, not the file count. Copying a single `node_modules` folder is `source.Count == 1` and goes to the shell — exactly the case this PR is meant to speed up. Needs to account for directories in the selection.

src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
```diff
@@ -76,8 +77,13 @@ public ShellFilesystemOperations(IShellPage associatedInstance)
...
+				DevToolsSettingsService.UseRobocopyForFileOperations;
+
 			if (sourceRename.Any())
 			{
+				// Rename operations always use shell operations for proper incremental naming
```
yair100: `GetCollisions` defaults non-conflicting items to `GenerateNewName`, and `sourceRename` is everything that isn't `ReplaceExisting` — so a normal paste with no name conflicts lands here and never reaches Robocopy. Copy only accelerates when overwriting.

`MoveItemsAsync` handles this correctly (same leaf name + destination doesn't exist). Copy needs the same guard.

src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
```diff
@@ -87,7 +93,9 @@ public ShellFilesystemOperations(IShellPage associatedInstance)
 
 			if (sourceReplace.Any())
 			{
-				var resultItem = await FileOperationsHelpers.CopyItemAsync(sourceReplace.Select(s => s.Path).ToArray(), destinationReplace.ToArray(), true, MainWindow.Instance.WindowHandle.ToInt64(), asAdmin, progress, operationID);
+				var resultItem = canUseRobocopy
+					? await FileOperationsHelpers.CopyItemWithRobocopyAsync(sourceReplace.Select(s => s.Path).ToArray(), destinationReplace.ToArray(), true, MainWindow.Instance.WindowHandle.ToInt64(), asAdmin, progress, operationID, _associatedInstance)
```
yair100, line 112: `asAdmin` is passed in and never used by the Robocopy path, so the access-denied → "retry elevated" flow below is dead here. Either fall back to the shell path when `asAdmin` is true, or wire up elevation.

src/Files.App/Utils/Storage/Operations/ShellFilesystemOperations.cs
```diff
@@ -377,13 +380,15 @@ public async Task<IStorageHistory> CreateShortcutItemsAsync(IList<IStorageItemWi
...
-				deleteFilePaths = deleteFilePaths.Concat(source.Select(x => GetRecycleMetadataPath(x.Path)).WhereNotNull()).Distinct();
+				deleteFilePaths = deleteFilePaths.Concat(source.Select(x => Path.Combine(Path.GetDirectoryName(x.Path), Path.GetFileName(x.Path).Replace("$R", "$I", StringComparison.Ordinal)))).Distinct();
 			}
 
 			var operationID = Guid.NewGuid().ToString();
 			await using var r = cancellationToken.Register(CancelOperation, operationID, false);
 
-			var (success, response) = await FileOperationsHelpers.DeleteItemAsync(deleteFilePaths.ToArray(), permanently, MainWindow.Instance.WindowHandle.ToInt64(), asAdmin, progress, operationID);
+			var (success, response) = permanently && DevToolsSettingsService.UseRobocopyForFileOperations
+				? await FileOperationsHelpers.DeleteItemWithRobocopyAsync(deleteFilePaths.ToArray(), MainWindow.Instance.WindowHandle.ToInt64(), asAdmin, progress, operationID, _associatedInstance)
```
yair100: Delete has no item-count gate — unlike copy/move it's just `permanently && UseRobocopyForFileOperations`. Permanently deleting one large file routes through the whole temp-move path. Should share the threshold.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -6,54 +6,61 @@
...
 
 namespace Files.App.Utils.Storage
 {
 	public sealed partial class FileOperationsHelpers
 	{
-		private static readonly PROPERTYKEY PKEY_FilePlaceholderStatus = new() { fmtid = new("B2F9B9D6-FEC4-4DD5-94D7-8957488C807B"), pid = 2 };
+		private static readonly Ole32.PROPERTYKEY PKEY_FilePlaceholderStatus = new Ole32.PROPERTYKEY(new Guid("B2F9B9D6-FEC4-4DD5-94D7-8957488C807B"), 2);
 		private const uint PS_CLOUDFILE_PLACEHOLDER = 8;
 
+		private static readonly IDevToolsSettingsService DevToolsSettingsService = Ioc.Default.GetRequiredService<IDevToolsSettingsService>();
```
yair100: Static field initializer resolves from the container at type-init time. If anything touches this class before the container is built you get a `TypeInitializationException` that's painful to diagnose. Make it lazy.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -672,8 +679,746 @@ public static Task SetClipboard(string[] filesToCopy, DataPackageOperation opera
...
+				};
+
+				using var process = new Process { StartInfo = psi };
+				var outputCompleted = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+				var remainingItemNames = expectedItemNames is null
+					? null
+					: new HashSet<string>(expectedItemNames, StringComparer.OrdinalIgnoreCase);
+				var initialProcessedSize = progressModel?.ProcessedSize ?? 0;
+				long batchProcessedSize = 0;
+				var hResult = -1;
```
yair100, line 771: Only sharing violations get a real HRESULT; everything else returns `-1`, which `CopyEngineResult.Convert` maps to unknown. Access-denied never surfaces as `Unauthorized`, so the elevation retry can't trigger. Parse `(0x00000005)` the same way you parse `0x00000020`.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -672,8 +679,746 @@ public static Task SetClipboard(string[] filesToCopy, DataPackageOperation opera
...
+						progressModel.Report(percentage);
+					}
+				};
+
+				process.Start();
+				process.BeginOutputReadLine();
+
+				var errorTask = process.StandardError.ReadToEndAsync();
+				using var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+				timeoutCts.CancelAfter(TimeSpan.FromMinutes(30));
```
yair100: Wall-clock 30 min, then `Kill(entireProcessTree: true)`. A legitimate large transfer over a slow share gets terminated mid-`/MOV`: sources already deleted for some files, truncated file at the destination. Should be an idle timeout keyed off `OutputDataReceived`.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -672,8 +679,746 @@ public static Task SetClipboard(string[] filesToCopy, DataPackageOperation opera
...
+				using var registration = timeoutCts.Token.Register(() =>
+				{
+					try
+					{
+						if (!process.HasExited)
+							process.Kill(entireProcessTree: true);
+					}
+					catch
```
yair100: Bare `catch { }` — you correctly replaced the silent catch in `WindowsJumpListService`; same treatment here.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -672,8 +679,746 @@ public static Task SetClipboard(string[] filesToCopy, DataPackageOperation opera
...
+					}
+
+					App.Logger?.LogWarning($"Robocopy operation {operationID}: Timed out");
+					return (false, -2);
+				}
+
+				await Task.WhenAll(outputCompleted.Task, errorTask);
+
+				var exitCode = process.ExitCode;
+				// Bit 2 means mismatched files; treating it as success can hide a partial move.
```
yair100: Exit code 2 is extra files in the destination; 4 is mismatched. The `<= 3` check is right, the comment isn't.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -672,8 +679,746 @@ public static Task SetClipboard(string[] filesToCopy, DataPackageOperation opera
...
+								cts.Cancel();
+								break;
+							}
+
+							var batchOk = true;
+							var hResult = 0;
+
+							var argsList = new List<string>
+							{
+								$"\"{sourceDir}\"",
```
yair100: When `sourceDir` is a drive root, this emits `"C:\"` — the trailing backslash escapes the closing quote under `CommandLineToArgvW` and the argument runs into the next one. Pasting into or out of a drive root is common. Trim trailing separators or double the backslash.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -672,8 +679,746 @@ public static Task SetClipboard(string[] filesToCopy, DataPackageOperation opera
...
+							else
+							{
+								// A move with replace semantics must process files Robocopy considers unchanged.
+								argsList.Add("/IS");
+								argsList.Add("/IT");
+							}
+
+							// Add operation-specific flags
+							if (isMoveOperation)
+								argsList.Add("/MOV");
```
yair100, line 1123: No `/XJ` or `/SL`, so junctions and symlinks get followed and their contents duplicated. A junction pointing at an ancestor is a recursion hazard with `/E`.

Separately, this file already handles `PS_CLOUDFILE_PLACEHOLDER` for the shell path — Robocopy will hydrate dehydrated OneDrive files, so a move can trigger an unbounded download. Worth gating placeholder items to the shell path rather than testing around it.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -672,8 +679,746 @@ public static Task SetClipboard(string[] filesToCopy, DataPackageOperation opera
...
+							(batchOk, hResult) = await RunRobocopyAsync(robocopyArgs, fsProgress, itemNames, operationID, cts.Token);
+
+							// Robocopy exit codes describe the batch, so verify every requested item before
+							// reporting success. A skipped move otherwise looks successful while its source remains.
+							var batchVerified = true;
+							foreach (var itemName in itemNames)
+							{
+								var sourcePath = Path.Combine(sourceDir, itemName);
+								var destinationPath = Path.Combine(destDir, itemName);
+								var itemOk = batchOk && StorageHelpers.Exists(destinationPath) &&
```
yair100: `StorageHelpers.Exists` is a `GetFileAttributesExFromApp` P/Invoke per item, twice for moves — up to 16k synchronous round trips per 8k batch. On SMB this can cost more than the transfer saved. A single directory enumeration of the destination would verify the whole batch.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -672,8 +679,746 @@ public static Task SetClipboard(string[] filesToCopy, DataPackageOperation opera
...
+				// Initial progress update
+				fsProgress.Report(0);
+
+				try
+				{
+					progressHandler.AddOperation(operationID);
+
+					// Step 1: Create temp folder structure
+					var tempBasePath = Path.GetTempPath();
+					var tempDeleteFolder = Path.Combine(tempBasePath, $"Files_Delete_{Guid.NewGuid()}");
```
yair100: `%TEMP%` is on the system drive. Permanently deleting 200 GB from `D:` or an SMB share reads all of it and writes it to `C:` before deleting — and fails outright when the system drive is short on space.

Robocopy also has no rename fast path; `/MOV` is copy-then-delete even within one volume, so this is slower than a plain delete in every case (worth timing to confirm). If this stays, the scratch folder has to be on the same volume as the source. I'd rather see delete split into a follow-up PR.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
```diff
@@ -672,8 +679,746 @@ public static Task SetClipboard(string[] filesToCopy, DataPackageOperation opera
...
+
+						fsProgress.Report(100); // 100% complete after robocopy MIR
+					}
+
+					// Step 4: Clean up temp folders
+					try
+					{
+						App.Logger?.LogInformation($"Robocopy delete operation {operationID}: Cleaning up temp folders");
+						if (Directory.Exists(tempDeleteFolder))
+							Directory.Delete(tempDeleteFolder, true);
```
yair100: This unconditional recursive delete is what actually removes the data — the `/MIR` above is redundant, and this runs single-threaded. It also fires when `success == false`, so a failed delete still destroys the files that made it to temp.

src/Files.App/Utils/Storage/Operations/FileOperationsHelpers.cs
d2dyno1:
```suggestion
		private static IDevToolsSettingsService DevToolsSettingsService => field ??= Ioc.Default.GetRequiredService<IDevToolsSettingsService>();
```

DarkEden-coding:
I remove the deletion part of this pr, the implementation was not clean and I don't have the bandwidth atm to try and figure out a better solution to that part

DarkEden-coding:
@yair100 any update on this? I just started the college school year and would like to get it wrapped up if possible, no worries if it takes a while though!

yair100:
@DarkEden-coding sorry for the delay, we're currently focused on stabilizing the changes in `main` and we're a little behind on code reviews. I would like to discuss this PR with a couple of other contributors to see if they agree with the approach or if they would prefer a different solution for file transfer operations.

DarkEden-coding:
@yair100 Got it, thank you!

PR #18762, opened 2026-07-22
Fix: Fixed issue where the Customize tab was shown for folders that don't support custom icons

### LINKED ISSUE #6689: Improvements for custom folder icons

workbysaran:
**Resolved / Related Issues**

Closes #6689

1. Opened Files, navigated to a UNC network share (\\server\share\folder), right-clicked a subfolder > Properties > confirmed the Customize tab no longer appears.
2. Opened Properties on a local folder (e.g. under C:\) > confirmed the Customize tab still appears and setting a custom icon still works.

### DIFF
--- a/src/Files.App/Utils/Storage/Helpers/DriveHelpers.cs
+++ b/src/Files.App/Utils/Storage/Helpers/DriveHelpers.cs
@@ -93,6 +93,27 @@ namespace Files.App.Utils.Storage
 			return null;
 		}
 
+		public static bool IsMtpPath(string path)
+		{
+			return path.StartsWith(@"\\?\", StringComparison.Ordinal);
+		}
+
+		public static bool IsNetworkPath(string path)
+		{
+			if (IsMtpPath(path))
+				return false;
+
+			try
+			{
+				return path.StartsWith(@"\\", StringComparison.Ordinal) ||
+					GetDriveType(new SystemIO.DriveInfo(path)) is Data.Items.DriveType.Network;
+			}
+			catch
+			{
+				return false;
+			}
+		}
+
 		public static Data.Items.DriveType GetDriveType(System.IO.DriveInfo drive)
 		{
 			if (drive.DriveType is System.IO.DriveType.Unknown)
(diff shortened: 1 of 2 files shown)


src/Files.App/Utils/Storage/Helpers/DriveHelpers.cs
yair100, line 1: I'm pretty sure we already have a helper for detecting network paths, can you check if it can be reused?

workbysaran: I checked. The only thing close is ShellViewModel.IsNetworkFolder, but that just compares against the Shell:NetworkPlacesFolder root, so it doesn't cover UNC paths or mapped network drives. I also noticed GetRootFromPathAsync is checking network path, but if needed I can refactor it.

Josh65-2201 (changes requested):
Folders in network drives/shares can have the folder icons customized. The only place it should be disabled is the root for device shares `\\127.0.0.1` or mapped drive 

Image

workbysaran:
@Josh65-2201 
What you're showing is Folder pictures, a legacy Windows XP feature that doesn't work consistently, and we don't implement it anyway, The Folder icon isn't offered for network folders in native Explorer (see the pic below), root or not, so disabling Customize for all network paths matches Explorer's own behaviour, not just at the root. The original behaviour is correct and I don't think we should narrow it to root-only.
image

Josh65-2201 (approved):
I see I haven't used Windows File Explorer for setting the icon for a long time.
In that case the only reason it wouldn't work is if the group policy allow remove paths for icons isn't enabled. I have use Files to set folder icons fine on network so maybe a settings or if possible detecting that group policy could be done even if it isn't in this PR

PR #18769, opened 2026-07-26
Code Quality: Moved file tags database reconciliation off the startup path

### LINKED ISSUE #18616: Bug: Empty file tag registry keys accumulate and block startup
### Description

On Files 4.1.3, a cold launch can take roughly 15-20 seconds before the UI becomes reliably usable. The main window is created in under one second, but the UI thread later stops responding for about 5.1 seconds while the file tag database is scanned.

Two 50 ms process samples attributed most managed startup CPU time to `FileTagsHelper.UpdateTagsDb()`:

| Measurement | Trace 1 | Trace 2 |
| --- | ---: | ---: |
| Window handle created | 0.91 s | 0.77 s |
| Managed startup CPU | 7.35 s | 6.38 s |
| Tag database scan CPU | 6.46 s | 5.55 s |
| Share of managed startup CPU | 87.8% | 87.0% |
| Peak working set | 380 MB | 365 MB |

`InitializeAppComponentsAsync()` resumes on the UI context after quick access initialization and calls `UpdateTagsDb()` synchronously. The method recursively enumerates the registry-backed tag index and validates every entry against its FRN, path, and alternate data stream.

This is related to #17678, which reported the same behavior despite the reporter never assigning a tag. Reinstalling fixed that case, suggesting that clearing the package's registry hive removed the accumulated data.

Further investigation found that Files creates registry entries even when no tag has ever been assigned:

- `ShellViewModel.SetFileTag()` calls `FileTagsDatabase.SetTags()` for each item after reading its ADS, including when the result is an empty tag array.
- `SetTags()` calls `CreateSubKey()` for the file path before checking whether the tag array is empty.
- `SaveValues(key, null)` deletes values but leaves the created key hierarchy in place.
- `FindTag()` also uses `CreateSubKey()` for lookups, so a read can create additional empty path and FRN keys.
- `GetAll()`, called by `UpdateTagsDb()`, recursively walks the path hierarchy even when every key is empty.

The user has never assigned a file tag. A newly installed Dev package already contained 954 empty path keys and 945 empty FRN keys after limited use, also with zero value-bearing tag records. This confirms that normal browsing grows the tag index without tag usage.

### Steps To Reproduce

1. Install or reset Files without assigning any file tags.
2. Browse folders and allow extended item properties to load.
3. Fully exit Files, including the tray process.
4. Inspect the package's `SystemAppData\Helium\User.dat` hive and observe empty entries under the `FileTags` path and FRN indices.
5. Repeat normal browsing and observe that the number of empty keys increases.
6. Launch Files and observe `FileTagsHelper.UpdateTagsDb()` spending startup time walking the empty path hierarchy.

### Files Version

4.1.3

### Windows Version

10.0.26200.0

### User ID

_Not available_

### Log File

No application log captured this CPU-bound UI stall. The profiler, registry-hive, and controlled A/B measurements are included above.

yair100:
**Resolved / Related Issues**

- Closes #18616

1. Confirmed the app launches, enumerates, and shuts down normally with `UpdateTagsDb` running via `Task.Run` off the startup path.

### DIFF
--- a/src/Files.App/Helpers/Application/AppLifecycleHelper.cs
+++ b/src/Files.App/Helpers/Application/AppLifecycleHelper.cs
@@ -127,7 +127,7 @@ namespace Files.App.Helpers
 				);
 			});
 
-			FileTagsHelper.UpdateTagsDb();
+			_ = Task.Run(FileTagsHelper.UpdateTagsDb);
 
 			_ = Task.Run(async () =>
 			{
--- a/src/Files.App/Utils/FileTags/FileTagsDatabase.cs
+++ b/src/Files.App/Utils/FileTags/FileTagsDatabase.cs
@@ -21,19 +21,23 @@ namespace Files.App.Utils.FileTags
 			if (FileTagsKey is null)
 				return;
 
-			using var filePathKey = Registry.CurrentUser.CreateSubKey(CombineKeys(FileTagsKey, filePath));
-
 			if (tags is [])
 			{
-				SaveValues(filePathKey, null);
+				using var existingFilePathKey = Registry.CurrentUser.OpenSubKey(CombineKeys(FileTagsKey, filePath), writable: true);
+				if (existingFilePathKey is not null)
+					SaveValues(existingFilePathKey, null);
+
 				if (frn is not null)
 				{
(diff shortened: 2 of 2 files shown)


yair100:
Fyi @woojh3690

Josh65-2201 (approved):

PR #18771, opened 2026-07-27
Fix: Fixed issue where the Customize tab was hidden for network folders that support custom icons

workbysaran:
Related to #6689

Follow-up to #18762, which hid the Customize tab for every network location. Windows only allows folder icon customization on remote paths when the respective group policy is enabled, so the tab is now hidden only when that policy is off. The policy value doesn't exist by default, which is treated as disabled.

1. Created 'HKLM\SOFTWARE\Policies\Microsoft\Windows\Explorer\EnableShellShortcutIconRemotePath'   as a 'REG_DWORD' set to '1', restarted Files, and confirmed the Customize tab appears for the same network folder and that setting a custom icon works.
2. Set the value to `0` and confirmed the tab is hidden again.


yair100 (approved):

Josh65-2201 (approved):

yair100:
@workbysaran thank you.

PR #18772, opened 2026-07-27
Fix: Fixed issue where password popup was displayed multiple times when copying from an encrypted archive

### LINKED ISSUE #18656: Bug: Files repeatly asks for password when copying a folder from an encrypted archive
### Description

When copying a folder with many files from an encrypted archive to outside of the archive, Files asks for password for each file in the folder.

expected behaviour: Files only asks for password once

### Steps To Reproduce

1. Download [encrypted-example.zip](https://github.com/user-attachments/files/29471602/encrypted-example.zip). (Password: aaaaaa)
2. Open it with Files
3. Copy the "code" folder inside it to any folder in your filesystem

### Files Version

4.1.3.0

### Windows Version

10.0.26200.8246

### User ID

_No response_

### Log File

[debug.log](https://github.com/user-attachments/files/29472174/debug.log)

yair100:
**Resolved / Related Issues**

- Closes #18656

1. Confirmed password prompt was only displayed once

### DIFF
--- a/src/Files.App/Utils/Storage/StorageBaseItems/IPasswordProtectedItem.cs
+++ b/src/Files.App/Utils/Storage/StorageBaseItems/IPasswordProtectedItem.cs
@@ -12,6 +12,9 @@ namespace Files.App.Utils.Storage
 
 		Func<IPasswordProtectedItem, Task<StorageCredential>> PasswordRequestedCallback { get; set; }
 
+		// Called after an operation succeeds with credentials obtained from PasswordRequestedCallback
+		void OnCredentialsVerified() { }
+
 		async Task<TOut> RetryWithCredentialsAsync<TOut>(Func<Task<TOut>> func, Exception exception)
 		{
 			var handled = exception is SevenZipOpenFailedException szofex && szofex.Result is OperationResult.WrongPassword ||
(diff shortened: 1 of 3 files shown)


Josh65-2201:
This also keeps it unlock after opening and then backing out (No operation was done)

Trying to extract (7z encrypted from Files), It gives a file not found prompt. This is also happening in 4.2.1 on the same Windows install so I can not test further

Josh65-2201:
> This also keeps it unlock after opening and then backing out (No operation was done)

This isn't happening anymore.


> Trying to extract (7z encrypted from Files), It gives a file not found prompt. This is also happening in 4.2.1 on the same Windows install so I can not test further

This still is with a new Windows VM.

yair100:
Do you have a different file to test with? This is working properly on my side.

Josh65-2201:
I have made multiple 7z using Files to test with. I haven't tested the one provided in the issue, I will try that

Josh65-2201:
With the issue provide zip it works as expected, although used what looked like a Windows version of the password prompt and not Files design. ~~It's now always unlocked and can be moved even after replacing the zip with a new copy so I can't get a screenshot of the dialog it showed.~~ Got it to show
image



Holding Ctrl to copy instead of moving crashes it. This also happens when opening a encrypted 7z archive.
[debug.log](https://github.com/user-attachments/files/30721123/debug.log)

Issue #18744, opened 2026-07-19
Bug:When opening the **Recycle Bin** in Files, restoring deleted files results in incorrect file names.


zbt00123:
### Description

https://github.com/user-attachments/assets/5887b600-1b0f-4d9c-bc83-877a27b7bb89

As shown in the attached video, I created two files on the Desktop: `Text_A.txt` and `Pic.bmp`.

After deleting them:

- If I open the Recycle Bin using **Windows File Explorer** and drag the files back to the Desktop, they are restored with their original file names (`Text_A.txt` and `Pic.bmp`).
- However, if I open the Recycle Bin using **Files** and restore the same files, they are incorrectly restored with names such as `$R174J5H.txt` and `$RF3KD96.bmp` instead of their original names.

### Expected behavior

Files should restore deleted files with their original file names, matching the behavior of Windows File Explorer.

### Actual behavior

The restored files are renamed to their internal Recycle Bin names (for example, `$R174J5H.txt` and `$RF3KD96.bmp`) instead of their original names.

### Steps To Reproduce

1. Create `Text_A.txt` and `Pic.bmp` on the desktop.  
2. Delete these two files.  
3. Use Files to open the Recycle Bin.  
4. Select the two deleted files and drag them back to the desktop.

### Files Version

4.2.0.0

### Windows Version

10.0.26200.8875

### User ID

5be1ab97-a6f8-4ceb-8e51-64d2d86df27a

### Log File

[debug.log](https://github.com/user-attachments/files/30165410/debug.log)

Josh65-2201:
Thanks for the report

PR #18785, opened 2026-07-31
Fix: Folder shortcuts could not be restored from Recycle Bin

### LINKED ISSUE #17110: Bug: Folder shortcuts in the recycle bin can not be restored
### Description

Right clicking and restoring them doesn't do anything and doesn't show anything in the Status center. 

Side note the shortcut size isn't calculated when in the recycle bin and causes the whole folder to not show a total size on the status bar

### Steps To Reproduce

1. Create a folder shortcut
2. Delete folder shortcut
3. Open recycle bin
4. Try to restore shortcut

### Files Version

3.9.8

### Windows Version

10.0.26100.0

### User ID

3372b393-b294-495c-9814-4e5b48d9fd8d

### Log File

[debug.log](https://github.com/user-attachments/files/20106261/debug.log)

Lins-ju:
- Fixes: #17110 

**Resolved / Related Issues**

- Folder shortcuts could not be restored at all.
- When folders are thrown in the recycle bin, before refreshing the recycle bin page, it will show the originalPath as the current folder path the file is in.

*What i've found*
After investigating the cascading functions used in order to perform a "Restore" action in the Recycle Bin, certain problems were found:
1. FromShellItem was supposed to return a BinStorageFolder instead of a ShortcutItem, hiding the originalPath.
2. Vanara could not retrieve the "DeletedFrom" path that could be used for an originalPath.
3. The filePath used for restoring a recycle bin file was pointing to the recycle bin itself.

*Why did they happen?:*
1. FromShellItem had an inverted if statement. For the function to work properly, it needs to check a "$Recycle.Bin" presence in the path first than it checks if the type equals "ShellLinkItem".
2. That's probably because Recycle Bin files are not like any other generic files. For Recycle Bin files, you have a "$RXX..." that has its names changed to avoid name conflicts in Recycle Bin and a "$IXX..." file containing the metadata (OriginalPath, DateDeleted, etc). I haven't find an easy way to do that properly.
3. "GetShellFileItem" receives a folderItem. But after debugging and testing, the folderItem would not have a correct originalPath, instead the originalPath would point to the current file path which would be in the recycle bin.

*Other problems found*
- #18744. Items are changing names when restored. It had nothing to do with with the file being able to restore or not. Also, i haven't looked to fix it because it is not marked as "Ready to build".

*What did i do exacly and why*
- Used a FileStream to read its bytes directly from this metadata file. It seems to be the most reliable way to get the originalPath.
- It has checks to prevent non recycle bin files to be read.
- It has fallbacks to prevent passing a null filePath to the ShellFileItem. It will use the filePath itself when the "GetOriginalPathFromRecycleBinFile" returns null.

- Create a folder shortcut
- Delete folder shortcut
- Open recycle bin
- Try to restore shortcut

### DIFF
--- a/src/Files.App/Utils/Shell/ShellFolderExtensions.cs
+++ b/src/Files.App/Utils/Shell/ShellFolderExtensions.cs
@@ -1,7 +1,9 @@
 // Copyright (c) Files Community
 // Licensed under the MIT License.
 
+using FluentFTP.Helpers;
 using System.IO;
+using System.Text;
 using Vanara.PInvoke;
 using Vanara.Windows.Shell;
 
@@ -42,6 +44,52 @@ namespace Files.App.Utils.Shell
 			return value;
 		}
 
+		private static string? GetOriginalPathFromRecycleBinFile(string? physicalPath)
+		{
+			if (string.IsNullOrEmpty(physicalPath))
+				return null;
+
+			var fileNameFromPath = Path.GetFileName(physicalPath);
+
+			if (!fileNameFromPath.StartsWith("$R", StringComparison.OrdinalIgnoreCase) &&
+				!fileNameFromPath.StartsWith("$I", StringComparison.OrdinalIgnoreCase))
+				return null;
+
+			try
+			{
+				string directory = Path.GetDirectoryName(physicalPath)!;
+
+				string iFileName = fileNameFromPath.StartsWith("$R", StringComparison.OrdinalIgnoreCase) ?
+					"$I" + fileNameFromPath.RemovePrefix("$R") : fileNameFromPath.ToString();
+
+				string filePath = Path.Combine(directory, iFileName);
+
+				if (!File.Exists(filePath) && !Directory.Exists(filePath))
+					return null;
+
+
+				using var stream = new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+				using var reader = new BinaryReader(stream, Encoding.Unicode);
+
+				// From the 24th to the 28th, contains the number of characters for the originalPath.
+				reader.BaseStream.Seek(24, SeekOrigin.Begin);
+				int pathCharCount = reader.ReadInt32();
+
+				// From the 28th byte, that's where the OriginalPath is.
+				// Windows is UTF-16, so each character takes 2 bytes, hence the multiplication.
+				byte[] pathBytes = reader.ReadBytes(pathCharCount * 2);
+				string originalPath = Encoding.Unicode.GetString(pathBytes).TrimEnd('\0');
+
+				return originalPath;
+			}
+			catch
+			{
+
+			}
+
+			return null;
+		}
+
 		public static ShellFileItem GetShellFileItem(ShellItem folderItem)
 		{
 			if (folderItem is null)
@@ -56,6 +104,8 @@ namespace Files.App.Utils.Shell
 
 			// True path on disk
 			parsingPath ??= folderItem.FileSystemPath;
+
+			string? recycleBinFileOriginalPath = GetOriginalPathFromRecycleBinFile(folderItem.FileSystemPath);
 
 			if (parsingPath is null || !Path.IsPathRooted(parsingPath))
 			{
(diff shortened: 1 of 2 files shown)


src/Files.App/Utils/Shell/ShellFolderExtensions.cs
```diff
@@ -42,6 +44,52 @@ private static T TryGetProperty<T>(this ShellItemPropertyStore sip, Ole32.PROPER
...
+
+				if (!File.Exists(filePath) && !Directory.Exists(filePath))
+					return null;
+
+
+				using var stream = new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+				using var reader = new BinaryReader(stream, Encoding.Unicode);
+
+				// From the 24th to the 28th, contains the number of characters for the originalPath.
+				reader.BaseStream.Seek(24, SeekOrigin.Begin);
```
0x5bfa, line 75: This is not preferable. Read `System.Recycle.DeletedFrom` prop on the deleted item to get the path to the original folder.

0x5bfa (changes requested):

src/Files.App/Utils/Shell/ShellFolderExtensions.cs
Lins-ju: Noted. I was in a hurry and that day and forgot to mention.
`System.Recycle.DeletedFrom` was my first choice to solve this problem. It didn't work and had some issues:
- When you freshly delete a folder and file, the `System.Recycle.DeletedFrom` will return empty.
- `System.Recycle.DeletedFrom` is always returning empty for a shortcut of any kind.
- **AFTER** updating the trash bin folder (F5), Folders and Files will have their correct original path. Other shortcuts will keep the original path as the recycle bin even after F5.

Will investigate it further to try to pin point why.

0x5bfa: Looks like it's broken for links.

image

I'll use some time to investigate this in the explorer's implementation too.

Lins-ju: I may have found a fix.
`System.Recycle.DeletedFrom` does not exist for *Shortcuts*. I haven't realized before but, the original path lies in the target path itself.
I thought this would not work because if you used the target path as the *Original Path* for links, it would change the *Original Path* as soon as you moved the *File* or *Folder* it's linking itself to (But after testing, it does not!).
Will use that and see if it works.
I'll update this soon.

0x5bfa: Here's what I found:

**For the "Original Location" column in the Details view**
There looks like a hidden item ID in an item's PIDL. An internal function uncovers that hidden item ID then restores the original path. This can be legally called via `IShellFolder::GetDetailsOfEx`.

image

**For the "Restore the selected items" command**
There's a dedicated internal class for that command.

image

---

If we just wanna execute the restore operation, we should do the latter (while both of them can be done in Files legally).

Here's the definition.

```c#
[GeneratedComInterface(StringMarshalling = StringMarshalling.Utf16), InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
[Guid("5869092d-8af9-4a6c-ae84-1f03be2246cc")]
public partial interface IRecycleBinManager
{
	[PreserveSig] HRESULT Compact();
	[PreserveSig] HRESULT GetFileData(string path, out DELETEDITEM deletedItem);
	[PreserveSig] HRESULT GetItemCount(out ulong itemCount);
	[PreserveSig] HRESULT GetUsedSpace(out ulong usedSpace);
	[PreserveSig] HRESULT IsEmpty();
	[PreserveSig] HRESULT PurgeAll(IFileOperation fileOperation);
	[PreserveSig] HRESULT PurgeItems([MarshalAs(UnmanagedType.LPWStr)] string path, IFileOperation fileOperation);
	[PreserveSig] HRESULT SuspendUpdating(int suspend);
	[PreserveSig] HRESULT HandleMenuMsg2(uint message, nuint wParam, nint lParam, out nint result);
	[PreserveSig] HRESULT RestoreItems(IShellItemArray items, IFileOperation fileOperation);
	[PreserveSig] HRESULT RestoreItem(IShellItem item, IFileOperation fileOperation);
	[PreserveSig] HRESULT IsRecycled([MarshalAs(UnmanagedType.LPWStr)] string path, out int isRecycled);
	[PreserveSig] HRESULT EnumItems(uint flags, [MarshalAs(UnmanagedType.Interface)] out IEnumRecycleItems enumerator);
	[PreserveSig] HRESULT WillRecycle([MarshalAs(UnmanagedType.LPWStr)] string path);
	[PreserveSig] HRESULT DelayCompaction(int delay);
	[PreserveSig] HRESULT GetRecycleBinCount(out int count);
	[PreserveSig] HRESULT GetRecycleBinAt(int index, in Guid interfaceId, [MarshalAs(UnmanagedType.Interface)] out object result);
	[PreserveSig] HRESULT GetRecycleBin([MarshalAs(UnmanagedType.LPWStr)] string path, in Guid interfaceId, [MarshalAs(UnmanagedType.Interface)] out object result);
	[PreserveSig] HRESULT Refresh();
}

[Guid("4A04656D-52AA-49DE-8A09-0B178760E748")]
partial class RecycleBinManager {}
```

Unfortunately the codebase in this area is not good enough to implement this easily but hope you find it useful.

PR #18803, opened 2026-08-06
Feature: Improved context menu performance

### LINKED ISSUE #18677: Bug: Duplicate context menu
### Description

When “Move shell extensions into a submenu” is enabled, duplicate menu items appear in the context menu.

Image

However, when “Move shell extensions into a submenu” is disabled, this no longer happens.

Image

These duplicate options in Explorer feature both the new Windows 11 new menu and the classic win32 menu.

Image

Image

I don't know if it has anything to do with [this](https://github.com/files-community/Files/issues/8251). #8251

### Steps To Reproduce

1. Enable “Move shell extensions into a submenu”
2. Right-click to open the context menu

### Files Version

4.1.7.0

### Windows Version

10.0.26200.8737

### User ID

e938aee8-aed9-4d61-960b-d8db7ab1d415

### Log File

[debug.log](https://github.com/user-attachments/files/29656849/debug.log)

yair100:
**Resolved / Related Issues**

- Related #4180
- Closes #18677

- Tested all context menus and compared behaviors with Files Preview
- Ran the new context menu test from the test project

### DIFF
--- a/src/Files.App/Helpers/MenuFlyout/FastContextFlyout.cs
+++ b/src/Files.App/Helpers/MenuFlyout/FastContextFlyout.cs
@@ -1,0 +1,717 @@
+// Copyright (c) Files Community
+// SPDX-License-Identifier: MPL-2.0
+
+using CommunityToolkit.WinUI;
+using Files.App.UserControls.Menus;
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Controls.Primitives;
+using Windows.Foundation;
+
+namespace Files.App.Helpers.ContextFlyouts
+{
+	/// <summary>
+	/// Lightweight MenuFlyout-based context menu host. Renders commands as regular menu items plus a horizontal
+	/// icon-button row for the primary commands. Owns the open-direction handling: placement is resolved before
+	/// the first render so the primary row sits next to the pointer (top when opening downward, bottom when
+	/// opening upward) without visibly moving after the menu appears.
+	/// </summary>
+	public sealed class FastContextFlyout
+	{
+		private static ControlTemplate? primaryRowTemplate;
+		private static IUserSettingsService? userSettingsService;
+
+		private MenuFlyoutItem? primaryRow;
+		private MenuFlyoutSeparator? primarySeparator;
+		private List<ContextMenuFlyoutItemViewModel>? primaryModels;
+		private bool openedUp;
+		private bool placementResolved;
+		private double estimatedWidth;
+		private double referenceScreenY = double.NaN;
+		private FrameworkElement? invocationAnchor;
+		private Point? invocationPosition;
+
+		// Supplies the live invocation point for framework-shown menus, whose ContextRequested can arrive after the
+		// pre-render guess; pulled in Flyout_Opened so the direction correction uses the finger, not the stale cursor.
+		public Func<(FrameworkElement Anchor, Point Position)?>? InvocationPointProvider { get; set; }
+
+		public MenuFlyout Flyout { get; } = new()
+		{
+			Placement = FlyoutPlacementMode.Right,
+			AreOpenCloseAnimationsEnabled = false,
+		};
+
+		public IList<MenuFlyoutItemBase> Items
+			=> Flyout.Items;
+
+		public FastContextFlyout()
+		{
+			Flyout.Opening += (sender, e) => App.LastOpenedFlyout = Flyout;
+			Flyout.Opened += Flyout_Opened;
+		}
+
+		/// <summary>
+		/// Clears the menu and the per-open placement state. Call at the start of every (re)build.
+		/// </summary>
+		public void Reset()
+		{
+			Flyout.Items.Clear();
+			primaryRow = null;
+			primarySeparator = null;
+			primaryModels = null;
+			openedUp = false;
+			placementResolved = false;
+			referenceScreenY = double.NaN;
+			invocationAnchor = null;
+			invocationPosition = null;
+		}
+
+		/// <summary>
+		/// Builds the given models into the menu: the primary commands as an icon-button row on top, the rest as
+		/// regular menu items. The overflow placeholder models ("ItemOverflow"/"OverflowSeparator") are markers
+		/// only and are skipped; shell loaders append a real "Show more options" submenu instead.
+		/// </summary>
+		public void Build(List<ContextMenuFlyoutItemViewModel> models)
+		{
+			Reset();
+
+			var primary = models.Where(x => x.IsPrimary && !x.IsHidden && x.ShowItem).ToList();
+			var secondary = models
+				.Where(x => !x.IsPrimary && x.ID != "ItemOverflow" && (x.Tag as string) is not ("ItemOverflow" or "OverflowSeparator"))
+				.ToList();
+
+			if (primary.Count > 0)
+				SetPrimaryModels(primary);
+
+			ContextFlyoutModelToElementHelper.GetMenuFlyoutItemsFromModel(secondary)?.ForEach(Flyout.Items.Add);
+
+			ApplyEstimatedWidth(secondary, primary.Count);
+		}
+
+		/// <summary>
+		/// Adds the primary-command icon-button row (and its separator) at the top of the menu.
+		/// </summary>
+		public void SetPrimaryModels(List<ContextMenuFlyoutItemViewModel> models)
+		{
+			primaryModels = models;
+			primaryRow = BuildPrimaryCommandRow(models);
+			primarySeparator = new MenuFlyoutSeparator();
+			Flyout.Items.Insert(0, primaryRow);
+			Flyout.Items.Insert(1, primarySeparator);
+		}
+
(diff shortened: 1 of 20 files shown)


Josh65-2201:
Also fixes #18677

src/Files.App/Helpers/MenuFlyout/FastContextFlyout.cs
```diff
@@ -0,0 +1,665 @@
...
+			{
+				var button = new Button
+				{
+					Style = buttonStyle,
+					Command = model.Command,
+					CommandParameter = model.CommandParameter,
+					Content = model.ThemedIconModel.IsValid ? model.ThemedIconModel.ToThemedIcon() : null,
+					IsEnabled = model.IsEnabled,
+					AccessKey = model.AccessKey ?? string.Empty,
+				};
```
0x5bfa, line 675: **[P2] Preserve primary-command keyboard accelerators** The primary-row replacement copies the command and access key, but drops KeyboardAccelerator and KeyboardAcceleratorTextOverride. The previous AppBarButton conversion registered these accelerators. After opening the menu with Shift+F10, shortcuts such as F2, Ctrl+C, and Alt+Enter therefore no longer go through this button and its Click/Hide path. Even if the global key handler runs the command, the flyout remains open; Properties is especially affected because it defers opening until Closed. Please attach the model accelerator to the new button, preserve its placement/override behavior, and cover invoking one while the menu is open.

src/Files.App/ViewModels/UserControls/Widgets/BaseWidgetViewModel.cs
```diff
@@ -62,65 +62,30 @@ widgetCardItem.DataContext is not WidgetCardItem item ||
...
+			// change item heights once the menu is visible; the loader fills their contents.
+			flyout.ConvertPlaceholderToSubMenu("OpenWithPlaceholder", Strings.OpenWith.GetLocalizedResource(), "App.ThemedIcons.OpenWith");
+			flyout.ConvertPlaceholderToSubMenu("SendToPlaceholder", Strings.SendTo.GetLocalizedResource(), null);
+
+			// Pre-add "Show more options" before showing so filling it with shell items never resizes the menu
+			var (moreOptions, moreSeparator) = flyout.AddShowMoreOptionsIfEnabled();
 
 			// Show the flyout
-			itemContextMenuFlyout.ShowAt(widgetCardItem, new() { Position = e.GetPosition(widgetCardItem) });
+			flyout.ResolvePlacement();
```
0x5bfa: **[P2] Resolve placement from the actual invocation point** This runs before ShowAt, so a newly-created flyout has no Target yet and the position from e.GetPosition(widgetCardItem) is not available to the resolver. GetInvocationReferenceScreenPoint consequently falls back to GetCursorPos. For touch long-press and keyboard invocation, the mouse cursor may be elsewhere, causing the primary row to be put on the wrong end; the Opened correction also uses that same stale reference. Please pass the anchor and invocation position into the placement calculation (the sidebar call has the same ordering), or otherwise resolve after the target is known.

src/Files.App/Helpers/MenuFlyout/FastContextFlyout.cs
```diff
@@ -0,0 +1,665 @@
...
+						items.ForEach(subItem.Items.Add);
+					}
+					else
+					{
+						onEmpty?.Invoke();
+					}
+				}
+				catch (Exception ex)
+				{
+					Debug.WriteLine(ex);
```
0x5bfa, line 455: **[P2] Restore the fallback when submenu loading throws** onEmpty is called for a successful-but-empty result, but not from this exception path. Callers have already swapped Open with / Send to from their working leaf form to an empty submenu before this task starts, so an exception from the shell loader, getter, or model conversion leaves a permanently empty submenu instead of reverting/removing it. Please invoke the fallback from the catch path as well, with the same lifetime checks used for late completion.

tests/Files.InteractionTests/Helper/AxeHelper.cs
```diff
@@ -42,6 +43,26 @@ public static void AssertNoAccessibilityErrors()
...
+
+			var parts = error.Element.Properties["BoundingRectangle"].Trim('[').Trim(']').Split(',');
+			var values = new int[4];
+			for (int index = 0; index < 4; index++)
+			{
+				if (!int.TryParse(parts[index][2..], out values[index]))
+					return false;
+			}
+
+			return (values[2] - values[0]) <= 2 || (values[3] - values[1]) <= 2;
```
0x5bfa: **[P2] Do not suppress every zero-size accessibility error** This predicate is global: every BoundingRectangleSizeReasonable violation whose width or height is at most 2 px is removed, without checking that the element is the transient sidebar scroll arrow described above. A permanently zero-sized or focusable element in the new context-menu host would now pass the Axe assertions. Please scope the exception to the known element/state, or wait for the animation and rescan.

tests/Files.InteractionTests/Tests/ContextMenuTests.cs
```diff
@@ -0,0 +1,370 @@
...
+				if (reopenMenu is not null && !TestHelper.ElementExistsByName(subMenuName, 1500))
+				{
+					reopenMenu();
+					continue;
+				}
+
+				// When the shell offers no sub-items for the item type (common on bare CI images),
+				// the app shows the leaf form of Open with / Send to instead of a submenu. There is
+				// nothing to expand then, and clicking the leaf would launch a dialog instead.
+				if (TestHelper.IsLeafMenuItem(subMenuName))
```
0x5bfa: **[P3] Limit leaf fallback to Open with and Send to** The leaf fallback is valid for those two shell-provided menus, but this helper is also used for Show more options and Set as. If either of those regresses from a submenu to a leaf item, this branch silently treats the test as successful. Please make leaf fallback an explicit option that defaults to false and enable it only for Open with / Send to.

0x5bfa (changes requested):
I found three user-facing regressions and two test gaps that can mask context-menu/accessibility failures. The build and current CI checks are green, but these paths should be addressed before merging.

src/Files.App/Helpers/MenuFlyout/FastContextFlyout.cs
yair100: We can discuss displaying keyboard accelerators in the primary button tooltips, but the current behavior matches what we have in `main`. If you see something different, please let me know.
That being said, I can reproduce the issue with the missing `KeyboardAccelerator`.

src/Files.App/ViewModels/UserControls/Widgets/BaseWidgetViewModel.cs
yair100: I can reproduce this as well.

src/Files.App/Helpers/MenuFlyout/FastContextFlyout.cs
yair100: This is resolved in the latest commit.

src/Files.App/Helpers/MenuFlyout/FastContextFlyout.cs
yair100: This should be resolved, please check.

tests/Files.InteractionTests/Helper/AxeHelper.cs
yair100: Please check the latest changes.

tests/Files.InteractionTests/Tests/ContextMenuTests.cs
yair100: Good idea, this should be resolved.

src/Files.App/ViewModels/UserControls/Widgets/BaseWidgetViewModel.cs
yair100: I believe this is resolved, please check on your end.

src/Files.App/Helpers/MenuFlyout/FastContextFlyout.cs
0x5bfa: Can confirm resolved.

src/Files.App/ViewModels/UserControls/Widgets/BaseWidgetViewModel.cs
0x5bfa: Can confirm resolved.

src/Files.App/Helpers/MenuFlyout/FastContextFlyout.cs
0x5bfa: Can confirm resolved.

tests/Files.InteractionTests/Helper/AxeHelper.cs
0x5bfa: Can confirm resolved.

tests/Files.InteractionTests/Tests/ContextMenuTests.cs
0x5bfa: Can confirm resolved.

PR #18805, opened 2026-08-09
Fix: Fixed crash when dropping items onto a folder from another window

### LINKED ISSUE #17296: Bug: Window Crashes when doing Drag and Drop
### Description

I have a dual-monitor setup, and if I open two separate file windows on each monitor and drag a file (can be anything) from one window to another, it copies/moves the file there if you drop it on empty space inside a folder.
But if you drop the file on a folder icon (where the file should go inside the folder), it crashes the whole window in which the file was dropped; all the tabs just vanish and the window disappears.
While the other window from where you dragged the file keeps working.

Crash while dropping on empty, I've experienced it once but dropping on the folder icon crashes it every time.

### Steps To Reproduce

1. Open 2 parallel windows (not tabs)
2. Drag a file from one window and drop on a folder icon in the other window.

### Files Version

3.9.12.0

### Windows Version

19045.6093

### User ID

_No response_

### Log File

[debug.log](https://github.com/user-attachments/files/21382040/debug.log)

TRUINGLol:
**Resolved / Related Issues**

- Closes #17296

**What was wrong**

`BaseLayoutPage.Item_Drop` read `e.Data`, which is only populated for drags started within the same XAML island. A drag coming from another window travels through the shell, so `e.Data` is `null` there and the getter threw a `NullReferenceException`. Since `Item_Drop` is an `async void` handler, the exception went unhandled on the UI thread and took the whole window down.

1. Opened Files and opened a second window (two windows, not two tabs).
2. Dragged a file from the first window onto a **folder icon** in the second window — the file is moved/copied and both windows stay alive. Before this change the second window crashed, losing all its tabs.
3. Verified no regressions on the other item-level drop targets: drop onto a folder within a single window, drop from File Explorer, drop onto an executable and a script (Open with), drop onto a folder shortcut, drop into a zip folder, drop onto a folder in the Grid layout with scroll-locking during drag, and drop onto empty space.

### DIFF
--- a/src/Files.App/Views/Layouts/BaseLayoutPage.cs
+++ b/src/Files.App/Views/Layouts/BaseLayoutPage.cs
@@ -1178,33 +1178,44 @@ namespace Files.App.Views.Layouts
 
 		protected virtual async void Item_Drop(object sender, DragEventArgs e)
 		{
-			var deferral = e.GetDeferral();
 			e.Handled = true;
+
+			// e.Data is only populated for drags started within this XAML island, drags coming from
+			// another window travel through the shell, so only e.DataView can be read here (#17296)
+			if (e.DataView is null)
+			{
+				e.AcceptedOperation = DataPackageOperation.None;
+				return;
+			}
+
+			DragOperationDeferral? deferral = null;
 
 			try
 			{
-				_ = e.Data.Properties;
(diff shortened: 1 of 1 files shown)


yair100:
Can you please resolve the merge conflicts? Thanks!

yair100 (approved):

d2dyno1 (approved):
lgtm

PR #18809, opened 2026-08-11
Fix: Fixed issue with filter text box being cleared

### LINKED ISSUE #18807: Bug: Filter textbox is cleared when typing right after opening a folder
### Description

This happens 1-2 seconds after the folder shows a fully loaded in the tab

### Steps To Reproduce

1. Enable filter bar
2. Open folder
3. Enter text
4. See cleared

### Files Version

4.2.2.0

### Windows Version

10.0.26200.8875

### User ID

670e1391-905a-4fe8-a9da-a22f01a4676c

### Log File

[debug.log](https://github.com/user-attachments/files/30875902/debug.log)

Rana-Shumail:
…load

This PR fixes an issue where the filter textbox is cleared 1-2 seconds after a folder loads if the user starts typing immediately.

Root cause: The desktopIniUpdateTask runs with Low priority and clears the filter box after the UI thread goes idle. Any text typed by the user before this task executes is overwritten.
Fix: Removed the delayed clearing. Instead, the filter is now cleared synchronously inside SetWorkingDirectoryAsync when the path actually changes (value != WorkingDirectory).
Note: I wasn't able to compile the app locally due to a local toolchain issue, so this fix is based on code analysis. I'm happy to iterate if any feedback or issues arise during testing!

**Resolved / Related Issues**

- Closes #18807

1. Performed code analysis on `ShellViewModel.cs` to trace the filter text binding.
2. Verified that the `desktopIniUpdateTask` was asynchronously clearing `FilesAndFoldersFilter` on a low-priority thread, which conflicted with immediate user typing.
3. Verified that `SetWorkingDirectoryAsync` correctly resets the filter synchronously when navigating between folders.

### DIFF
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -215,6 +215,7 @@ namespace Files.App.ViewModels
 		{
 			if (string.IsNullOrWhiteSpace(value))
 				return;
+			
 			if (value != WorkingDirectory)
 			{
 				FilesAndFoldersFilter = null;
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -215,6 +215,10 @@ namespace Files.App.ViewModels
 		{
 			if (string.IsNullOrWhiteSpace(value))
 				return;
+			if (value != WorkingDirectory)
+			{
+				FilesAndFoldersFilter = null;
+			}
 
 			var isLibrary = false;
 			string? name = null;
@@ -2210,7 +2214,6 @@ namespace Files.App.ViewModels
 						{
 							GetDesktopIniFileData();
 							CheckForBackgroundImage();
-							FilesAndFoldersFilter = null;
 						},
 						Microsoft.UI.Dispatching.DispatcherQueuePriority.Low);
 					});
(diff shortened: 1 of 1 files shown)


Josh65-2201 (dismissed):

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -215,6 +215,10 @@ public async Task SetWorkingDirectoryAsync(string? value)
 		{
 			if (string.IsNullOrWhiteSpace(value))
 				return;
+			if (value != WorkingDirectory)
```
yair100, line 219:
```suggestion
			
			if (value != WorkingDirectory)
```

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -215,6 +215,10 @@ public async Task SetWorkingDirectoryAsync(string? value)
...
+			{
```
yair100: Brackets can be removed

yair100: @Rana-Shumail

Rana-Shumail: Hi @yair100 — just pushed a commit removing the braces. Should be resolved now, let me know if anything else needs adjusting!

yair100 (approved):

yair100:
Thank you

Rana-Shumail:
Thank you for the guidance!

PR #18819, opened 2026-08-14
fix: avoid clearing active filter text in delayed desktop.ini task

### LINKED ISSUE #18807: Bug: Filter textbox is cleared when typing right after opening a folder

marko1olo:
## Problem

Fixes #18807.

When opening a folder with the filter bar enabled, typing into the filter textbox shortly after navigation starts gets wiped out 1–2 seconds later when the directory load finishes.

### Cause

`NavigateToPath` in `ModernShellPage.xaml.cs` already resets `ShellViewModel.FilesAndFoldersFilter = null;` when navigation begins. However, `desktopIniUpdateTask` was scheduled on `DispatcherQueuePriority.Low` and also set `FilesAndFoldersFilter = null;` upon completion, clearing whatever filter query the user had already entered while the folder was loading.

## Solution

Remove the redundant and delayed `FilesAndFoldersFilter = null;` from `desktopIniUpdateTask`.


Josh65-2201:
Duplicate of #18809 also has many unnecessary changes

PR #18825, opened 2026-08-17
Feature: Add support for Xbox gamepad controls

### LINKED ISSUE #7014: Xbox Controller Support
### What's the Problem?

I'm on sofa and want to open a file (a movie, or music folder, or whatever) without moving to pc, I have to use either wireless mouse or keyboard. But I only have controller in my hand.

While the current Files actually respond to my controller, however somehow confirm button (A) doesn't really work, and after using the context menu to get into a folder, the controller focus is somehow gone forever.

### Solution/Idea

Add controller support.

### Alternatives

wireless mouse/keyboard, or remote access through cellphone. But the mouse cursor on big screen is not as good as controller.

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow end users to browse files with game controller (typically xbox controller) | Could |

### Files Version

2.0.34

### Windows Version

Windows 10 Pro 21H1

### Additional comment

_No response_

infraredCoding:
**Resolved / Related Issues**

- Closes #7014 

1. Opened Files
2. Made sure an XBox Controller is connected (wired or wireless)
3. Used Left Joystick to move around the cursor
4. Used A to select an item, double press A to open the selected file/folder (Simulates Left Mouse Click)
5. Used B to go back to previous/parent folder or cancel dialog
6. Used X to simulate right mouse click
7. Used right joystick to Scroll. 
8. Used D-Pad buttons as just Arrow Keys

### DIFF
--- a/src/Files.App.CsWin32/NativeMethods.txt
+++ b/src/Files.App.CsWin32/NativeMethods.txt
@@ -37,6 +37,8 @@ AttachThreadInput
 SetWindowPos
 SetFocus
 SetActiveWindow
+SendInput
+SetCursorPos
 IsIconic
 CopyFileFromApp
 MoveFileFromApp
--- a/src/Files.App/Helpers/UI/XBoxGamepadController.cs
+++ b/src/Files.App/Helpers/UI/XBoxGamepadController.cs
@@ -73,6 +73,11 @@ namespace Files.App.Helpers
 		private bool _wasXPressed;
 		private bool _wasBPressed;
 
+		private bool _wasDPadDownPressed;
+		private bool _wasDPadUpPressed;
+		private bool _wasDPadLeftPressed;
+		private bool _wasDPadRightPressed;
+
 		public XBoxGamepadController(DispatcherQueue dispatcherQueue)
 		{
 			_timer = dispatcherQueue.CreateTimer();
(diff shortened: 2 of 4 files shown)


Josh65-2201 (changes requested):
I would recommend adding the below controls.

Left trigger > Left click
Right trigger > Right click
Hold left trigger + Left stick > Click and drag

If the context menu/flyout is open, B should close it and not go back a folder.

Using the mouse, no hover event's are triggered and highlights/information isn't visible

Files is goes into an unrecoverable crash where I have to restart the system, this seem to be when anything in the omnibar has it's drop down open

0x5bfa (changes requested):
Use CsWin32 instead of manually defining PInvokes. Also currently the helper does nothing?

infraredCoding:
@Josh65-2201 Should I add
`Left trigger > Left click
Right trigger > Right click
Hold left trigger + Left stick > Click and drag`
as in replace the existing controls or just add them along with the existing ones?

Josh65-2201:
I don't see a reason to remove the existing ones if they don't cause conflicts

infraredCoding:
Thanks for the clarification

Josh65-2201 (changes requested):
I can not test further due to Files going into an unrecoverable crash still being an issue

> Files is goes into an unrecoverable crash where I have to restart the system, this seem to be when anything in the omnibar has it's drop down open

There is no log for this, I will test if VS debug can get anything

infraredCoding:
@Josh65-2201 I tried recreating the crash after the changes before pushing but somehow I wasn't able to do so in VS Debug. Perhaps I had missed something. Was the crash happening on a particular release platform?

Josh65-2201:
It doesn't end up doing a full crash either stops accepting input (Seem to break all mouse input) or get stuck in not responsive after a few seconds of using a controller to navigate, VS debug isn't showing anything either.

PR #18841, opened 2026-08-19
Code Quality: Address remaining AOT issues other than XAML and automation

hez2010:
Also set `IsTrimmable` and `IsAotCompatible` to true.

We are close to finally enable NativeAOT. At least with `IsAotCompatible=true` Files shouldn't crash without running under NativeAOT now.


src/Files.App/Data/Contracts/INavigationControlItem.cs
```diff
@@ -7,10 +7,6 @@ namespace Files.App.Data.Contracts
 {
 	public interface INavigationControlItem : IComparable<INavigationControlItem>, INotifyPropertyChanged, ISidebarItemModel, ISidebarItemPresentationModel
 	{
-		public string? Text { get; }
-
-		public string? Path { get; }
-
```
hez2010, line 13: These properties are already defined by the interface it inherits, so removing them here.

yair100 (approved):

src/Files.App/Data/Contracts/INavigationControlItem.cs
yair100: 👍

PR #18847, opened 2026-08-19
Code Quality: Trim AI packages

yair100:
The AI/ML Windows App SDK components were shipping ~39 MB of unused native runtime
(`onnxruntime.dll` + `DirectML.dll`) into the self-contained package. Nothing in the
app references these APIs.

The existing `PackageReference Update="Microsoft.WindowsAppSDK.{AI,ML,Widgets}" ExcludeAssets="all"`
lines were didn't work as expected. `Update` doesn't apply asset exclusion to transitive-only packages, and
the native payload actually lives in the leaf `Microsoft.Windows.AI.MachineLearning` package.

Converted them to direct `PackageReference Include ... ExcludeAssets="all"` (which NuGet honors)
and added the leaf ML package. Restore now records `"include": "None"` for all four, so no
compile/runtime/native/build assets flow into the package.


hez2010 (approved):
LGTM!

PR #18855, opened 2026-08-20
Code Quality: Various fixes and performance improvements

hez2010:
Addresses several correctness, performance, and resource-lifetime issues found during a codebase audit.

### Path and storage safety

- Escapes paths and other values passed to PowerShell commands.
- Prevents ZIP entries from being extracted outside the destination folder.
- Runs folder flattening outside the UI thread, keeps operations within the selected root, and skips reparse points.
- Corrects FTP identities, ports, special-character paths, overwrite behavior, and stream ownership.

### Async behavior and performance

- Makes Git fetch operations awaitable, serialized, cancellable, and prevents canceled fetches from raising stale completion events.
- Moves Git metadata work outside the UI thread and limits diffs to the requested path.
- Preserves STA affinity across asynchronous `STATask` operations.
- Removes unnecessary long-running tasks, redundant PDF decoding, and inactive-pane date updates.

### Resource ownership

- Ensures timed-out file enumeration handles are eventually released through `SafeHandle`.
- Releases GDI, ACL, SID, security descriptor, PIDL, dialog filter, and COM event resources.
- Cleans up temporary dialog and alternate-stream files.
- Corrects native dialog result handling and event-sink ownership.
- Fixes a dangling pointer in `AuthenticateNetworkShare`

### Logging and synchronization

- Uses one shared asynchronous file logger while retaining an unbounded channel.
- Stops accepting and retaining messages if the log writer permanently fails.
- Uses `System.Threading.Lock` for applicable synchronization points.


src/Files.App/Services/Storage/StorageArchiveService.cs
```diff
@@ -288,7 +288,7 @@ await Task.Run(() =>
 						}
 
 						string entryFileName = zipEntry.Name;
-						string fullZipToPath = Path.Combine(destinationFolderPath, entryFileName);
+						string fullZipToPath = GetSafeExtractionPath(destinationFolderPath, entryFileName);
```
0x5bfa: The default `encoding == null` path uses `ExtractArchiveAsync`, so this check is skipped for normal extraction. SevenZipSharp accepts `..` entries and can write outside the destination. Apply the same validation to the SevenZip path.

src/Files.App/Services/Storage/StorageArchiveService.cs
```diff
@@ -358,6 +358,21 @@ await Task.Run(() =>
...
 		}
 
+		private static string GetSafeExtractionPath(string destinationFolderPath, string entryName)
+		{
+			var destinationRoot = Path.GetFullPath(destinationFolderPath);
+			var destinationPrefix = Path.EndsInDirectorySeparator(destinationRoot)
+				? destinationRoot
+				: destinationRoot + Path.DirectorySeparatorChar;
+			var normalizedEntryName = entryName.Replace('/', Path.DirectorySeparatorChar);
+			var destinationPath = Path.GetFullPath(Path.Combine(destinationPrefix, normalizedEntryName));
```
0x5bfa, line 372: This string-only check does not account for existing junctions under the destination. `File.Create` follows them and can write outside the folder. Reject reparse-point components before opening the output.

0x5bfa (changes requested):
Two archive extraction paths can still write outside the destination.

src/Files.App/Services/Storage/StorageArchiveService.cs
hez2010: Fixed.

src/Files.App/Services/Storage/StorageArchiveService.cs
hez2010: Fixed.

hez2010:
PTAL

yair100:
LGTM

PR #18856, opened 2026-08-20
Fix: Fixed an issue where right-clicking could leave the cursor stuck as a spinner or arrow

yair100:
**Resolved / Related Issues**

- Closes #

1. Right clicked item
2. Confirmed cursor didn't change

### DIFF
--- a/src/Files.App/Utils/Shell/ShellCursorScope.cs
+++ b/src/Files.App/Utils/Shell/ShellCursorScope.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
 using Windows.Win32;
+using Windows.Win32.Foundation;
 using Windows.Win32.UI.WindowsAndMessaging;
 
 namespace Files.App.Utils.Shell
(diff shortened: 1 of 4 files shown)


src/Files.App/Utils/Shell/ShellCursorScope.cs
```diff
@@ -0,0 +1,51 @@
...
+			capturedInfo = info;
+		}
+
+		public static unsafe ShellCursorScope Capture()
+		{
+			CURSORINFO info = new() { cbSize = (uint)sizeof(CURSORINFO) };
+			if (!PInvoke.GetCursorInfo(ref info))
+				info = default;
+
+			return new(info);
```
0x5bfa: Multiple shell calls can overlap. A second scope can capture the wait cursor set by the first and restore it last, leaving the cursor stuck. Keep a process-wide active count so only the first scope captures and the last restores, or serialize these shell calls.

src/Files.App/Utils/Shell/ShellCursorScope.cs
```diff
@@ -0,0 +1,51 @@
...
+		{
+			// Nothing to restore when the capture failed or the cursor was hidden/suppressed (touch input)
+			if (capturedInfo.cbSize is 0 || capturedInfo.flags is not CURSORINFO_FLAGS.CURSOR_SHOWING || capturedInfo.hCursor.IsNull)
+				return;
+
+			CURSORINFO current = new() { cbSize = (uint)sizeof(CURSORINFO) };
+			if (!PInvoke.GetCursorInfo(ref current))
+				return;
+
+			// A moved pointer has already re-resolved its cursor through WM_SETCURSOR
```
0x5bfa, line 62: The pointer can stay at the same coordinates while focus moves to another window, for example with Alt+Tab. Since SetCursor changes shared cursor state, this can overwrite another window cursor. Capture the window under the pointer and restore only if it is still the same Files window.

0x5bfa (changes requested):
The scope is clean, but cursor restoration should handle concurrent calls and window ownership.

This is Codex generated.

yair100:
I'm having trouble getting this to work properly.

PR #18857, opened 2026-08-20
Code Quality: Enable Direct P/Invoke

hez2010:
Enable Direct P/Invoke for NativeAOT to import them during linking, so that we don't pay runtime lazy initialization and resolution for some common modules.

### DIFF
--- a/src/Files.App/Files.App.csproj
+++ b/src/Files.App/Files.App.csproj
@@ -41,6 +41,25 @@
     <PropertyGroup>
         <DefineConstants>$(DefineConstants);DISABLE_XAML_GENERATED_MAIN</DefineConstants>
     </PropertyGroup>
+
+    <ItemGroup>
+        <DirectPInvoke Include="ADVAPI32" />
+        <DirectPInvoke Include="api-ms-win-core-com-l1-1-0" />
+        <DirectPInvoke Include="api-ms-win-core-com-l1-1-1" />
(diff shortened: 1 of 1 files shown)


src/Files.App/Files.App.csproj
```diff
@@ -41,6 +41,25 @@
...
+    <ItemGroup>
+        <DirectPInvoke Include="ADVAPI32" />
+        <DirectPInvoke Include="api-ms-win-core-com-l1-1-0" />
+        <DirectPInvoke Include="api-ms-win-core-com-l1-1-1" />
+        <DirectPInvoke Include="api-ms-win-core-winrt-l1-1-0" />
+        <DirectPInvoke Include="api-ms-win-core-winrt-robuffer-l1-1-0" />
+        <DirectPInvoke Include="api-ms-win-core-winrt-string-l1-1-0" />
+        <DirectPInvoke Include="CRYPT32" />
+        <DirectPInvoke Include="GDI32" />
+        <DirectPInvoke Include="KERNEL32.dll" />
```
hez2010, line 54: The `.dll` extension name here is intentional to avoid importing the lazy-bound `kernel32!LoadLibrary`, which is used by LibGit2Sharp. LibGit2Sharp imports `kernel32` without specifying the `.dll` extension name.

yair100 (approved):

0x5bfa (commented):
It'd be better to put them in DirectPInvokeList

Issue #18871, opened 2026-08-22
Bug: Files crashes when I try to create TXT file via context menu


iamrusskih:
### Description

https://github.com/user-attachments/assets/87ea3c95-4909-48e8-85a2-f177851119e9

### Steps To Reproduce

1. Open any directory via "Files"
2. Change language to some except English (e.g. Serbian or Russian) 
3. Open context menu → New (Нов/Создать) → TXT file 

### Files Version

4.2.9.0

### Windows Version

10.0.26200.9168

### User ID

6c7c87d2-8fa8-40ce-952e-45eb7fae232d

### Log File

[debug.log](https://github.com/user-attachments/files/31336462/debug.log)

Josh65-2201:
Thanks for the feedback, can you see if this is an issue in the preview version. https://files.community/download

iamrusskih:
@Josh65-2201 
I got same behavior in Files - Preview 4.2.14.0

Josh65-2201:
Can you upload the log for the preview version please.

iamrusskih:
[debug.log](https://github.com/user-attachments/files/31337602/debug.log)

Josh65-2201:
Thanks

iamrusskih:
Hope, it will be useful

harshdtripathi:
Hi! I'd like to work on this issue. Could you assign it to me?

Josh65-2201:
@harshdtripathi  This isn't ready to work on, please read the [contribution guidelines](https://github.com/files-community/Files/blob/main/.github/CONTRIBUTING.md#finding-issues-you-can-help-with), ready to work on issues are at https://github.com/orgs/files-community/projects/3/views/7

yair100:
Can anyone else reproduce this?

yair100:
A potential will be included in the next update.

yair100:
@iamrusskih can you please check if this is resolved in Preview v4.2.35?

iamrusskih:
@yair100 Sorry for late feedback, looks like fixed in the latest preview (4.2.36.0)

Josh65-2201:
Thanks for checking

PR #18874, opened 2026-08-23
Fix: Fixed the most frequent crashes in the preview release

yair100:
- Fixed a startup crash (COMException 0x8000FFFF "XamlRoot must be explicitly set for unparented popup") caused by the review/sponsor TeachingTips opening during the initial x:Bind pass, before the page had a XamlRoot; prompts now open only after MainPage is loaded, once per session
- Fixed an ObjectDisposedException when layout pages called CancelExtendedPropertiesLoading after the ShellViewModel was disposed; added isDisposed guards and switched LoadExtendedItemPropertiesAsync / LoadGitPropertiesAsync to capture the CancellationToken instead of the source
- Fixed an ObjectDisposedException when a refresh started on a disposed ShellViewModel by guarding RapidAddItemsToCollectionAsync
- Fixed a crash (ArgumentException 0x80070057) when opening shortcut properties in the NativeAOT build; ReadOnlyDictionary's value collection can't be bound to ItemsSource, replaced with a plain Dictionary

### DIFF
--- a/src/Files.App/Data/Models/SelectedItemsPropertiesViewModel.cs
+++ b/src/Files.App/Data/Models/SelectedItemsPropertiesViewModel.cs
@@ -798,9 +798,9 @@ namespace Files.App.Data.Models
 
 		private static readonly Dictionary<SHOW_WINDOW_CMD, string> showWindowCommandTypes = new()
 		{
-			{ SHOW_WINDOW_CMD.SW_NORMAL, Strings.NormalWindow.GetLocalizedResource() },
-			{ SHOW_WINDOW_CMD.SW_SHOWMINNOACTIVE, Strings.Minimized.GetLocalizedResource() },
-			{ SHOW_WINDOW_CMD.SW_MAXIMIZE, Strings.Maximized.GetLocalizedResource() }
+			[SHOW_WINDOW_CMD.SW_NORMAL] = Strings.NormalWindow.GetLocalizedResource(),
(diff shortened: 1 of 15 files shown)


hez2010 (approved):
LGTM.

hez2010:
> * Fixed a crash (ArgumentException 0x80070057) when opening shortcut properties in the NativeAOT build; ReadOnlyDictionary's value collection can't be bound to ItemsSource, replaced with a plain Dictionary

It's unfortunately to see CsWinRT's source generator doesn't work well with nested properties over collection interface properties, though. I think we might need to audit the codebase for this specific pattern to ensure we don't leave some leftovers.

Josh65-2201:
Would this also fix #18871 ?

It has the same `COMException 0x8000FFFF "XamlRoot must be explicitly set for unparented popup` but not for the startup

hez2010:
@yair100 I opened a PR (https://github.com/files-community/Files/pull/18875) against your branch to include some additional fixes for missing CsWinRT vtable lookup generation after automating a codebase audit. Please merge that PR before merging this.

yair100:
@hez2010 thanks! I merged your PR.

yair100:
> Would this also fix #18871 ?
> 
> It has the same `COMException 0x8000FFFF "XamlRoot must be explicitly set for unparented popup` but not for the startup

@Josh65-2201 I don't think so.

PR #18885, opened 2026-08-24
Feature: SSH and AOT support for git operations

hez2010:
- Maintain a forked libgit2 and libgit2sharp, so that SSH support won't open a console window
- Show git exception message directly instead of the "Pulled timeout", which is extremely misleading and helpless


yair100 (approved):
I did a complete review of the changes and nupkg packages.
The goal is to replace them with the official versions once the changes get merged upstream.

PR #18887, opened 2026-08-25
Feature: Improved scrolling performance

yair100:
**Resolved / Related Issues**

For #4180

- Move elevation check and tag I/O off the UI thread
- Defer item property loads while scrolling
- Single low-priority dispatch for item icons
- Cache selection checkbox lookups
- Lazy-load rename editors in Grid/List templates

**Details**
https://github.com/user-attachments/assets/b0395523-d35c-4e3d-a588-045d0b002129

**Grid**
https://github.com/user-attachments/assets/776ade76-0a8f-49a3-b930-ebf3b05d69da

### DIFF
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -941,6 +941,10 @@ namespace Files.App.ViewModels
 		/// </summary>
 		public void NotifyScrollStateChanged(bool isScrolling)
 		{
+			// The unsynchronized field update is only safe while all callers arrive on the UI dispatcher
+			if (!dispatcherQueue.HasThreadAccess)
+				throw new InvalidOperationException($"{nameof(NotifyScrollStateChanged)} must be called from the UI thread.");
+
 			if (isScrolling)
 				scrollSettledTcs ??= new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
 			else
@@ -1536,7 +1540,7 @@ namespace Files.App.ViewModels
 				var settledTask = scrollSettledTcs?.Task;
 				if (settledTask is not null)
 				{
-					await settledTask;
+					await settledTask.WaitAsync(token);
 					token.ThrowIfCancellationRequested();
 				}
 
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -929,6 +929,22 @@ namespace Files.App.ViewModels
 
 			loadPropsCTS.Cancel();
 			loadPropsCTS = new CancellationTokenSource();
+
+			// Release parked loads so they can observe the cancellation
+			Interlocked.Exchange(ref scrollSettledTcs, null)?.TrySetResult();
+		}
+
+		private TaskCompletionSource? scrollSettledTcs;
+
+		/// <summary>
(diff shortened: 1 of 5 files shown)


src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -1518,6 +1531,15 @@ public async Task LoadExtendedItemPropertiesAsync(ListedItem item)
 			try
 			{
 				token.ThrowIfCancellationRequested();
+
+				// Defer while a scroll gesture is in flight; flung-past items are canceled by their container's recycle before this resumes
+				var settledTask = scrollSettledTcs?.Task;
+				if (settledTask is not null)
+				{
+					await settledTask;
+					token.ThrowIfCancellationRequested();
```
hez2010:
```suggestion
					await settledTask.WaitAsync(token);
```

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -1213,6 +1214,40 @@ protected void FileList_ContainerContentChanging(ListViewBase sender, ContainerC
...
+		// Hooked lazily from the first container callback, when the list's template is guaranteed realized
+		[DynamicWindowsRuntimeCast(typeof(ScrollViewer))]
+		private void HookScrollDeferTracking()
+		{
+			if (deferScrollViewer is not null)
+				return;
+
+			deferScrollViewer = ItemsControl.FindDescendant<ScrollViewer>();
+			if (deferScrollViewer is not null)
+				deferScrollViewer.ViewChanged += DeferScrollViewer_ViewChanged;
```
hez2010, line 1229: Detach the handler on navigation and disposal so that we don't leave a handler attached after leaving the page?

src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -929,6 +929,22 @@ public void CancelExtendedPropertiesLoading()
...
+
+		private TaskCompletionSource? scrollSettledTcs;
+
+		/// <summary>
+		/// Parks extended-property loads while a scroll gesture is in flight and releases them when it settles.
+		/// </summary>
+		public void NotifyScrollStateChanged(bool isScrolling)
+		{
+			if (isScrolling)
+				scrollSettledTcs ??= new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
```
hez2010, line 949: Techincally this will lead to a race condition, but we are only accessing it on the UI dispatcher so it should be fine as the operations are serialized.
Well I suggest adding a `if (!dispatcherQueue.HasThreadAccess) throw new InvalidOperationException(...)` so that we can immediately know something went wrong from the log if the current invariants break, instead of a silent hang.

hez2010 (changes requested):
The improvement seems promising, just a few things before this can be merged.

src/Files.App/ViewModels/ShellViewModel.cs
yair100: I pushed a change

src/Files.App/Views/Layouts/BaseLayoutPage.cs
yair100: I pushed a change

src/Files.App/ViewModels/ShellViewModel.cs
yair100: I pushed a change

hez2010 (approved):

PR #18884, opened 2026-08-24
Code Quality: Removed WinUIEx

yair100:
**Resolved / Related Issues**

To prevent extra work, all changes to the Files codebase must link to an approved issue marked as `Ready to build`. Please insert the issue number following the hashtag with the issue number that this Pull Request resolves.
- Closes #

**Steps used to test these changes**

Stability is a top priority for Files and all changes are required to go through testing before being merged into the repo. Please include a list of steps that you used to test this PR.

1. Confirmed window placement is saved
2. Confirmed RTL works properly
3. Confirmed minimum window size

PR #18888, opened 2026-08-25
Feature: Performance and optimizations

yair100:
**Resolved / Related Issues**

- For #4180
- Parallelized thumbnail generation (was serialized)
- Persistent STA thread pool for shell icon/thumbnail calls
- Cached thumbnails paint at first realization (no icon-to-thumbnail flash)
- Skip redundant per-item icon fetches when the generic icon is final
- Preloaded icons fetched per layout size and DPI, shared via cache
- Release item images when scrolled out of view
- Trim memory after leaving large folders, closing tabs, or going idle
- Sorted streaming enumeration (no reshuffle when loading completes)
- Faster first paint when opening folders
- Keep previous listing visible when switching folders (no empty flash)
- Removed Shimmer loading placeholder and its package dependency

- Confirmed icons no longer shift when switching from placeholder to actual icon
- Confirmed improved responsiveness when switching locations
- Confirmed thumbnails load quicker


Josh65-2201:
Icons often load before the folder and with it not clearing the items in between it looks laggy as the reset of the folder loads.

I see this is deploy as a preview build and I will see how it looks outside the VM

src/Files.App.Storage/Windows/Managers/STATask.cs
```diff
@@ -13,6 +13,58 @@ namespace Files.App.Storage
 	/// </summary>
 	public partial class STATask
 	{
+		private const int PooledThreadCount = 12;
+
+		private static readonly BlockingCollection<Action> _pooledQueue = new();
+		private static int _poolStarted;
```
hez2010: I believe we can use a bool here.

src/Files.App.Storage/Windows/Managers/STATask.cs
```diff
@@ -13,6 +13,58 @@ namespace Files.App.Storage
 	/// </summary>
 	public partial class STATask
 	{
+		private const int PooledThreadCount = 12;
+
+		private static readonly BlockingCollection<Action> _pooledQueue = new();
```
hez2010: I would instead use `Channel` here.

src/Files.App/Services/Storage/IconCacheService.cs
```diff
@@ -12,32 +14,57 @@ internal sealed class IconCacheService : IIconCacheService
...
+				IconOptions.ReturnIconOnly | (useCurrentScale ? IconOptions.UseCurrentScale : IconOptions.None));
 
 			_cache.TryAdd(key, icon);
 			return icon;
 		}
 
+		public async Task<BitmapImage?> GetIconImageAsync(string? itemPath, string? extension, bool isFolder, uint size, bool useCurrentScale)
+		{
+			// BitmapImage has window thread affinity, so decoded images are cached per dispatcher
+			var dispatcher = DispatcherQueue.GetForCurrentThread();
```
hez2010: Do we really need a `Dispatcher` key? I think we only have one UI dispatcher per process so this dispatcher key seems redundant to me.

Also actually I'm concerned that with this cache we can display a staled icon if the icon changed even after refreshing the list. I don't see a cache invalidation path. Also this cache seems unbounded so navigating to a large folders can start taking very large memory. 

I think we would be better to stick with built-in thumbnail cache that comes with Windows.

src/Files.App/Views/Layouts/BaseLayoutPage.cs
```diff
@@ -1246,13 +1248,20 @@ private void UnhookScrollDeferTracking()
...
 			{
 				scrollSettleTimer = DispatcherQueue.CreateTimer();
 				scrollSettleTimer.Interval = TimeSpan.FromMilliseconds(200);
 				scrollSettleTimer.IsRepeating = false;
-				scrollSettleTimer.Tick += (_, _) => ParentShellPageInstance?.ShellViewModel?.NotifyScrollStateChanged(false);
+				scrollSettleTimer.Tick += (_, _) =>
+				{
+					var shellViewModel = ParentShellPageInstance?.ShellViewModel;
+					shellViewModel?.NotifyScrollStateChanged(false);
+					if (shellViewModel?.FilesAndFolders.Count >= 100)
```
hez2010: Not sure if this is intended but trimming pages like this can result in performance regression on normal use.

Josh65-2201:
Based on 4.2.19.0 preview, scrolling in grid layout will have some items thumbnail become icons and then thumbnails again. 4.2.18.0 pre will have the thumbnail always visible. I assume it's suppose to based on

> Release item images when scrolled out of view

Not seeing any other issues so far

hez2010 (changes requested):

src/Files.App/Services/Storage/IconCacheService.cs
yair100: > Also actually I'm concerned that with this cache we can display a staled icon if the icon changed even after refreshing the list. I don't see a cache invalidation path. Also this cache seems unbounded so navigating to a large folders can start taking very large memory.

This cache is only used as a placeholder while the thumbnail loads from the shell. Also, the cache is per file extension (as opposed to every single file) so it shouldn't get that large.

yair100: Removed the `Dispatcher` key.

src/Files.App.Storage/Windows/Managers/STATask.cs
yair100: Done

src/Files.App.Storage/Windows/Managers/STATask.cs
yair100: Done

hez2010 (approved):
LGTM. Can you include an additional fix I mentioned in https://github.com/files-community/Files/pull/18884#discussion_r3858978840 before merging?

yair100:
@hez2010 thank you for reviewing all these changes!

PR #18895, opened 2026-08-26
Code Quality: Fix crashes from v4.2.20

yair100:
**Resolved / Related Issues**

- Closes #

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Dialogs/DecompressArchiveDialog.xaml.cs
+++ b/src/Files.App/Dialogs/DecompressArchiveDialog.xaml.cs
@@ -16,7 +16,7 @@ namespace Files.App.Dialogs
 			get => (FrameworkElement)MainWindow.Instance.Content;
 		}
 
-		public DecompressArchiveDialogViewModel ViewModel
+		public required DecompressArchiveDialogViewModel ViewModel
 		{
 			get => (DecompressArchiveDialogViewModel)DataContext;
 			set => DataContext = value;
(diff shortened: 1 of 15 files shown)


src/Files.App/Dialogs/DecompressArchiveDialog.xaml.cs
```diff
@@ -29,7 +29,7 @@ public DecompressArchiveDialog()
 
 		private void ContentDialog_PrimaryButtonClick(ContentDialog sender, ContentDialogButtonClickEventArgs args)
 		{
-			if (ViewModel.IsArchiveEncrypted)
+			if (ViewModel is { IsArchiveEncrypted: true })
```
hez2010: Why ViewModel can even be null?

I would instead make `ViewModel` a required property so that we won't hit a null ViewModel.

hez2010 (changes requested):
Otherwise LGTM.

src/Files.App/Dialogs/DecompressArchiveDialog.xaml.cs
yair100: Good catch

hez2010: Uhh it seems that we cannot use required property in view model.
Can you revert the last commit and instead making ViewModel nullable? Then I think it will be ready to go.

PR #18911, opened 2026-09-01
Feature: WinRT out-of-proc server

hez2010:
Bring back the WinRT OOP server.

I managed to make it work under both normal and elevated environment. I have plans to offload some work to the WinRT server so that they can persist across different Files instances.


src/Files.App/Program.cs
```diff
@@ -84,54 +84,8 @@ private static void Main()
...
-					}
-				}
-			});
-
-			// NOTE:
-			//  This has been commented out since out-of-proc WinRT server seems not to support elevation.
-			//  For more info, see the GitHub issue (#15384).
 			// Now we can do the first WinRT server call
-			//Server.AppInstanceMonitor.StartMonitor(Environment.ProcessId);
+			Server.AppInstanceMonitor.StartMonitor(Environment.ProcessId);
```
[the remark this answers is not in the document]
hez2010: This can no longer happen with the latest CsWinRT and .NET.

Directory.Packages.props
```diff
@@ -32,15 +32,15 @@
     <PackageVersion Include="Sentry" Version="6.9.0" />
     <PackageVersion Include="SevenZipSharp" Version="1.0.3" />
     <PackageVersion Include="SharpZipLib" Version="1.4.2" />
-    <PackageVersion Include="SQLitePCLRaw.bundle_green" Version="2.1.11" />
     <PackageVersion Include="Microsoft.WindowsAppSDK" Version="2.4.0" />
-    <PackageVersion Include="Microsoft.Windows.AI.MachineLearning" Version="2.1.74" />
+    <PackageVersion Include="Microsoft.Windows.AI.MachineLearning" Version="2.3.42" />
```
yair100, line 36: Can this package be removed?

yair100:
Do you know how much this increases the package size?

Directory.Packages.props
hez2010: This is an unused transitive dependency coming from WASDK and here we are just bumping the version and explicitly exlucding its assets. Removing this would unnecessarily bring the AI native binaries into the final package.

hez2010:
> Do you know how much this increases the package size?

Shouldn't be large. The server itself is very compact, I forgot about the exact number but it should just be several MBs. After packaging it should be even smaller.

Directory.Packages.props
yair100: We exclude it from `Files.App.Csproj` so I'm not sure which one is being followed.

hez2010: > We exclude it from `Files.App.Csproj` so I'm not sure which one is being followed.

It follows the one in `Files.App.csproj`. I was bumping the version here just to remove the nuget package update notification from my VS 😁

yair100:
It seems to be about 4 MB, also keep in mind that #18885 added about 6 MB. It's not a blocking issue but these increases are adding up. I would like if we could reduce the size to be closer to 50 MB again.

hez2010:
> It seems to be about 4 MB, also keep in mind that #18885 added about 6 MB. It's not a blocking issue but these increases are adding up. I would like if we could reduce the size to be closer to 50 MB again.

The size can be much smaller once we migrate to CsWinRT 3.0. CsWinRT hasn't been released yet but it's close.

Issue #1834, opened 2020-08-28
Right clicking rename box closes rename mode


Jaiganeshkumaran:
**Describe the bug**
When you right-click the text box shown in rename mode it gets closed.

**To Reproduce**
1. Hit rename on any item
2. Right click or long press
3. It will get closed

**Expected behavior**
It shouldn't close until the users hits enter or space or any other standard key for closing something.

**Screenshots**
Recorded video of the problem:
https://app.box.com/s/met1ry3heo6nutt9kwdhzbcci4fclu8a

**Desktop:**
 - OS Version: Windows 10 version 2004
 - App version: v0.13.4.0

Jaiganeshkumaran:
For some reason the context menu isn't captured by the video but when I right click I does show a menu flyout for a second and then it goes away.

yaichenbaum:
@Jaiganeshkumaran This issue should be resolved with #2749.

PR #18912, opened 2026-09-01
Code Quality: Fix crashes from preview

yair100:
**Resolved / Related Issues**

- Closes #

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
+++ b/src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
@@ -4,6 +4,7 @@
 using CommunityToolkit.WinUI;
 using Files.App.Controls;
 using Files.App.UserControls.Selection;
+using Microsoft.Extensions.Logging;
 using Microsoft.UI.Input;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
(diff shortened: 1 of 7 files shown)


src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
```diff
@@ -174,22 +175,30 @@ protected override void OnNavigatedTo(NavigationEventArgs eventArgs)
...
+					ColumnsViewModel.TagColumn.Update(FolderSettings.ColumnsViewModel.TagColumn);
+					ColumnsViewModel.GitStatusColumn.Update(FolderSettings.ColumnsViewModel.GitStatusColumn);
+					ColumnsViewModel.GitLastCommitDateColumn.Update(FolderSettings.ColumnsViewModel.GitLastCommitDateColumn);
+					ColumnsViewModel.GitLastCommitMessageColumn.Update(FolderSettings.ColumnsViewModel.GitLastCommitMessageColumn);
+					ColumnsViewModel.GitCommitAuthorColumn.Update(FolderSettings.ColumnsViewModel.GitCommitAuthorColumn);
+					ColumnsViewModel.GitLastCommitShaColumn.Update(FolderSettings.ColumnsViewModel.GitLastCommitShaColumn);
+				}
+				catch (Exception ex)
+				{
+					// Same #1834 bug: the Update raises PropertyChanged into a classic Width binding whose ICustomProperty target was moved by the GC, throwing an NRE that would otherwise fail the navigation and take down the app. The persisted widths get reapplied on the next update.
```
hez2010: Do you have repro for this issue?

This doesn't make sense to me. It should always work regardless of the GC, otherwise this is a severe GC hole in the framework.

hez2010 (changes requested):

src/Files.App/Views/Layouts/DetailsLayoutPage.xaml.cs
yair100: I don't have a repro but this is the Stack Trace from Sentry.
```
System.NullReferenceException
  at ABI.Microsoft.UI.Xaml.Data.ICustomProperty.Do_Abi_GetValue_2(IntPtr, IntPtr, IntPtr*)
  at ABI.System.ComponentModel.PropertyChangedEventHandler.NativeDelegateWrapper.Invoke(object, PropertyChangedEventArgs)
  at DetailsLayoutColumnItem.set_UserLength(GridLength)
  at DetailsLayoutColumnItem.Update(DetailsLayoutColumnItem)
  at DetailsLayoutPage.OnNavigatedTo(NavigationEventArgs)
```

It seems related to this issue [CsWinRT #1834](https://github.com/microsoft/CsWinRT/issues/1834) which is still open: the `ManagedCustomProperty` CCW gets a 0 refcount, so the object behind the classic `{Binding}` can be collected out from under the binding engine while it's still calling `ICustomProperty.GetValue` during the `PropertyChanged` that `OnNavigatedTo → Update()` raises.

hez2010: This doesn't fix the problem. If we hit the exception here, the runtime state should already likely be corrupted and there's nothing we can do to workaround it. I would prefer removing this workaround and see if there're more related bug reports.

hez2010: After some investigation, the problem seems located in WinUI instead of Files or .NET or CsWinRT. There's nothing we can do to fix it without a fix from WinUI directly.

hez2010 (approved):
LGTM!

PR #18934, opened 2026-09-08
Code Quality: Improved nav perf

yair100:
**Resolved / Related Issues**

- Closes #

1. Opened Files ...
2. ...

### DIFF
--- a/src/Files.App/ViewModels/ShellViewModel.cs
+++ b/src/Files.App/ViewModels/ShellViewModel.cs
@@ -199,6 +199,12 @@ namespace Files.App.ViewModels
 		// Carries the enumeration's cloud sync result to the post-enum switch, avoiding a second query.
 		private CloudDriveSyncStatus? enumeratedCloudSyncStatus;
 
+		private static readonly ConcurrentDictionary<string, System.IO.DriveType> driveTypeCache = new();
+
+		// Speculative directory open started at navigation request time so it overlaps the frame-navigation window.
+		private sealed record DirectoryOpenPrefetch(string Path, Task<(Win32PInvoke.SafeFindHandle? Handle, WIN32_FIND_DATA Data, int ErrorCode)> Task);
+		private static DirectoryOpenPrefetch? directoryOpenPrefetch;
+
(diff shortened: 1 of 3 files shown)


src/Files.App/ViewModels/ShellViewModel.cs
```diff
@@ -199,6 +199,12 @@ public GitProperties EnabledGitProperties
 		// Carries the enumeration's cloud sync result to the post-enum switch, avoiding a second query.
 		private CloudDriveSyncStatus? enumeratedCloudSyncStatus;
 
+		private static readonly ConcurrentDictionary<string, System.IO.DriveType> driveTypeCache = new();
+
+		// Speculative directory open started at navigation request time so it overlaps the frame-navigation window.
+		private sealed record DirectoryOpenPrefetch(string Path, Task<(Win32PInvoke.SafeFindHandle? Handle, WIN32_FIND_DATA Data, int ErrorCode)> Task);
+		private static DirectoryOpenPrefetch? directoryOpenPrefetch;
```
yair100: The prefetch might seem like a small gain, but it's better than nothing. I would be happy to skip this change in favor of a better idea.

hez2010: I would rather like to remove this prefetch. Implementing a cache also needs proper cache invalidations, which would then complicate the whole thing by a lot.

For example, with this change refreshing a folder can reuse an old scan, which can then lead to missing files. And a drive letter that was previously used by a local drive could later be remapped to a network share while Files is still open. They both lead to correctness regressions.

IMO the risk here is larger than the perf gain.

yair100: Do you have an alternative solution? There is often a short gap between double clicking a folder and the items rendering. It's not a long gap but enough to be noticed.

hez2010: I need to profile this path first to identify what is blocking the navigation from happening.

yair100: I removed this change 👍

hez2010 (approved):
LGTM!

PR #18939, opened 2026-09-10
Code Quality: Optimize navigation performance

hez2010:
- Reuse row styles until the layout or size changes.
- Eliminate duplicate preference probes and unnecessary registry writes.
- Move Git discovery, Recent Items updates, and fallback access checks off the UI thread.
- Guard against stale navigation results and disposed pages.


yair100 (approved):
LGTM

PR #18950, opened 2026-09-14
Fix: Fixed crashes caused by the blocked UI thread while running in the background

yair100:
Closing to background blocked the UI thread on `Program.Pool.WaitOne()` inside `Window_Closed` (and in `OnLaunched` for the startup-task case), so the dispatcher stopped pumping for the entire background stay. WinRT wrapper finalizers stall behind the blocked thread — a captured crash event shows ~98,000 pending finalizers — which eventually crashes the ComWrappers layer under NativeAOT (`ManagedObjectWrapper.QueryInterface` NRE) and also prevented the background working-set trim from ever completing.

The semaphore wait is now awaited off the UI thread so the dispatcher keeps pumping in the background. The close is cancelled up front (`Handled` set after an `await` is read too late), and quitting from the background closes the window explicitly after teardown. No user-facing behavior change; the background instance now also releases memory properly since finalizers keep draining.

Fixes FILES-APP-6GQ.


src/Files.App/App.xaml.cs
```diff
@@ -390,19 +395,18 @@ static bool IsSameChannelInstance(Process p)
...
 				TryEmptyWorkingSetWhenIdle(cts.Token);
 
-				if (Program.Pool.WaitOne())
-				{
-					cts.Cancel();
-					// Resume the instance
-					Program.Pool.Dispose();
-					Program.Pool = null;
+				// Waiting must not block the dispatcher; WinRT wrapper finalizers stall until it pumps again
+				await WaitOneAsync(Program.Pool);
```
hez2010: This actually introduced a race condition needs to be addressed.
A rapid close → reopen → close can start another Window_Closed handler before the first wait continuation resumes. The newer handler can replace the semaphore while the older handler then disposes and clears that newer semaphore. This would lead the newer handler to throw NullReferenceException on the next resume.
Before this change the main thread was blocked so there was no such an issue.

hez2010 (changes requested):

src/Files.App/App.xaml.cs
hez2010: I think the simplest (though not complete) fix is to change the `Program.Pool.Dispose()` to `Program.Pool?.Dispose()`, but it should be enough.

yair100: 👍

hez2010 (approved):
This LGTM. Thanks!

yair100:
Thank you for reviewing

PR #18960, opened 2026-09-15
Code Quality: Use the AOT-friendly overload of `Marshal.GetFunctionPointerForDelegate`

0x5bfa:
### Resolved / Related Issues

- #4180

Its hard to validate, this change only affects on the AOT builds. Currently it seems to be working but the method we use currently is AOT unfriendly (both are but the one with generics should be safer with the presence of `T` for AOT).


yair100 (approved):
LGTM

hez2010 (approved):

PR #18963, opened 2026-09-16
Fix: Fixed folder names briefly showing unlocalized on first launch

### LINKED ISSUE #18632: Bug: Folder names show English before using the app language
### Description

Files is installed on my PCs with Windows 11 in Italian. In all of them when I launch the program, the list of folders in the system partition appears in English for a moment and then switches to Italian. I don't know the reason. Anybody else? Thank you

### Steps To Reproduce

Read above

### Files Version

4.1.3.0

### Windows Version

10.0.26200.8655

### User ID

c0b9be8e-e413-4cf4-aad1-8cc9ee9abf63

### Log File

[debug.log](https://github.com/user-attachments/files/29237835/debug.log)

yair100:
**Resolved / Related Issues**

- Closes #18632

1. Opened Files ...
2. ...


yair100:
It seems fine but it would be good if someone else can test this to make sure there are no performance regressions.

yair100:
We might want to consider persisting this cache.

src/Files.App/Services/Storage/StorageCacheService.cs
```diff
@@ -8,6 +8,8 @@ namespace Files.App.Utils.Storage
 	/// <inheritdoc cref="IStorageCacheService"/>
 	internal sealed class StorageCacheService : IStorageCacheService
```
hez2010, line 9: Does this cache really bring us meaningful improvements? I think `GetLocalizedName` should be very fast so that a cache here doesn't actually make sense despite introducing the risk of inconsistency.

yair100: No, I removed it.

hez2010:
@codex review

yair100:
Do you know if we can remove `StorageFolder.DisplayName` from `LoadExtendedItemPropertiesAsync` now that we're using `GetLocalizedName`?

hez2010 (approved):
LGTM! I have some idea around the cache if it turns out that it's better to have it in the end. But let us leave it to the future.

yair100:
Do you know if we can remove `StorageFolder.DisplayName` from `LoadExtendedItemPropertiesAsync` now that we're using `GetLocalizedName`?

hez2010:
> Do you know if we can remove `StorageFolder.DisplayName` from `LoadExtendedItemPropertiesAsync` now that we're using `GetLocalizedName`?

Not yet. Win32 and tag searches still use raw folder names and rely on this `DisplayName` update for localization. We need to use `GetLocalizedName` in those paths before removing it.

yair100:
Seems like a good optmization to do when we have some time.

